PR #11, opened 2024-06-13
Update README.md to provide more info about JDKs

edmondop:
Helping people save time instead of trying to figure out what that unreadable message of SBT means when you have JDK 21 enabled :)


rxin:
Thanks @edmondop !

PR #14, opened 2024-06-13
Change the cmd flag from --format to --output

fhuanming:
Change the cmd flag from --format to --output. Otherwise following error will raise:
```
$ bin/uc table list --catalog unity --schema default --format jsonPretty
Some of the provided parameters are not valid.
Usage: bin/uc table list [options]
Required Params:
  --catalog The name of the catalog.
  --schema The name of the schema.
Optional Params:
  --max_results The maximum number of results to return.
```


rxin:
Thanks @fhuanming !

PR #18, opened 2024-06-14
Add Docker image for Unity Catalog

pflooky:
Some small changes are included in this PR to help create a working Docker image for Unity Catalog:

- Add Dockerfile
- Add script to run build and push for Docker images
- Add GitHub action to build and upload Docker image
- Add `uc-docker` script for CLI to work straight away on the Docker image
- Allow for `server.properties` and `log4j2.properties` file pathways to be set via environment variables or system properties
- Fix banner having two `v`'s
- Change `--format` to `--output` in the documentation


tdas:
Thank you for this pr. Docker would be great. I am going to play with this thing as soon as I get a chance.

README.md
```diff
@@ -22,12 +22,21 @@ Let's take Unity Catalog for spin. In this guide, we are going to do the followi
   as well as provide a convenient way to explore the content of any UC server implementation.
 
 ### Prerequisites
+- Docker
```
tdas, line 25: You do have to clone this repo still, so the 2 options has to be under the instructions of cloning the repo.

docker-action.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
```
tdas, line 2: Could you add some comments in the file on what does this file do?

docker-action.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
tdas, line 1: Does this script have to be in the root of the repo. Keeping th root of the repo uncluttered is better

tdas:
There seems to be a more extensive docker support PR #18

README.md
pflooky: You don't need to clone the repo if running the `docker run ...` command. It will pull the image from Dockerhub and start running straight away.

pflooky:
> Would love to see a docker image for this as well but I'm wondering how `datacatering` is affiliated to `unitycatalog`? Since the image is built in this repo wouldn't it be better if it's published under `unitycatalog`?

I was using `datacatering` to test the docker image creation. Happy to change to `unitycatalog`.

When this PR is merged, to get the GitHub action to work, someone will have to add in `DOCKER_HUB_USER` and `DOCKER_HUB_TOKEN` to the GitHub action secrets that has access to upload to `unitycatalog`.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+WORKDIR /opt/app
+
+COPY server/target/unitycatalog-server-assembly.jar unitycatalog-server.jar
+COPY examples/cli/target/unitycatalog-cli-assembly.jar unitycatalog-cli.jar
+COPY etc etc
+COPY bin/uc-docker uc
+
+ENV SERVER_PROPERTIES_FILE=/opt/app/etc/conf/server.properties
+ENV SERVER_JOG4J_CONFIGURATION_FILE=/opt/app/etc/conf/server.log4j2.properties
+ENV CLI_JOG4J_CONFIGURATION_FILE=/opt/app/etc/conf/cli.log4j2.properties
```
tdas, line 12: Where are these ENV variables used? What's the point of assigning them?

README.md
tdas: Can you elaborate a bit more about this? what will be the full command of `docker run ... `
If we are adding a completely new way to run these steps, then we need to write down the exact steps that people can copy paste and run and leave nothing to assumption. 

And I am not as familiar with Docker as i want to be. I am guessing that this `docker run ...` command is going to download from dockerhub? who is going to publish docker hub?

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM openjdk:24-slim
+
+WORKDIR /opt/app
+
+COPY server/target/unitycatalog-server-assembly.jar unitycatalog-server.jar
+COPY examples/cli/target/unitycatalog-cli-assembly.jar unitycatalog-cli.jar
```
tdas, line 6: whats the purpose of the CLI jar in docker? we are not running the CLI in docker

README.md
```diff
@@ -22,12 +22,20 @@ Let's take Unity Catalog for spin. In this guide, we are going to do the followi
...
+OR
+
 You have to ensure that your local environment has the following:
 - Clone this repository.
 - Ensure the `JAVA_HOME` environment variable your terminal is configured to point to JDK11+.
 - Compile the project using `build/sbt package`
 
 ### Run the UC Server
+```shell
+docker run -d -i --name unitycatalog -p 8081:8081 unitycatalog/unitycatalog:0.1.0
```
tdas, line 36: this docker run step will only work with published versions of the server? like 0.1? 
my goal was for these steps to be runnable by anyone with the current unpublished code as soon as they clone this repo

bin/uc-docker
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
tdas, line 1: what is the point of this file

tdas (changes requested):
Overall, thank you for your attempt to dockerize the whole thing. But there are a lot of changes in this PR that I dont understand the purpose of. could you please bear with me, and answer my questions so that we understand what we are merging and what will be needed to maintain this?

build/docker-action.sh
Cupprum, line 1: Is there a reason why this is a separate script, and not part of yaml pipeline file?

README.md
Cupprum: We could document how to build the container locally, and afterwards also execute it locally.

pdbg:
+1

Nice to have dockerization version of uc.

LeonardoSanBenitez:
+1
Thanks for the initiative.
I'm getting started now with UC outside databricks and this docker image is great to have.

creechy:
This duplicates a PR https://github.com/unitycatalog/unitycatalog/pull/116  that included Docker functionality

PR #22, opened 2024-06-14
Add a Dockerfile

jaychia:
1. Updates the `.gitignore` to ignore build artifacts
2. Adds a `.dockerignore` to ignore build artifacts
3. Adds a `Dockerfile` to build a Docker container

One problem which I haven't yet found a workaround is that the test data for UC will reside inside the container instead of on the host filesystem. This does make quickstart for the `unity.default.marksheet` default table a little confusing because the data cannot found found at the storage_location on the host 😬 

I think it could be cool to maybe add an additional Docker stage called `quickstart` which installs DuckDB and we can hook it up to stdin so that users can quickly run both unitycatalog and DuckDB in the same Docker container and get started really quickly with a `bin/quickstart.sh` or something.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM openjdk:11
+
+COPY . /usr/src/myapp
+WORKDIR /usr/src/myapp
+
+RUN build/sbt clean compile
+
+CMD ["bin/start-uc-server"]
+
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+# Ignore all build artifacts
+**/target
+
+# Ignore generated sbt launch jar
+build/sbt-launch-*.jar
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,7 @@
 etc/db/h2db.mv.db
+
+# Ignore all build folders
+**/target
+
+# Ignore generated sbt launch jar
+build/sbt-launch-*.jar


tdas:
There seems to be a more extensive docker support in PR #18

tdas:
@jaychia I did a review of #18 and this PR, and I have to say I like the small and simple approach of making the quick start work easily in this form. However, we do need a solution for making the prepopulated datasets + quick start work. Could you extend this PR to make the changes you are proposing with `additional Docker stage called quickstart which installs DuckDB and we can hook it up to stdin so that users can quickly run both unitycatalog and DuckDB`

tdas:
(sorry, my fat fingers accidentally closed the PR)

jaychia:
> #18

Yeah I think it should be doable. Let me take a stab over the weekend.

MrPowers:
Where will we be publishing the Docker images?  Dockerhub?

Can you help me understand the end-user experience we want to give?

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM openjdk:11
```
Fokko, line 1: Optional: Ideally we want to have a multi-stage docker here with separate JDK for building the fat-jar. Copy the fat-jar into a JRE image for runtime.

Fokko (approved):
I think this is a great first step. We can iterate on this in a subsequent PR. For example, making sure that we make the image as small as possible (currently it is probably quite big) on publishing the image in a separate PR.

jaychia:
Sorry haven't yet had a chance to work on this! @Fokko let's hop on a call my JVM skills aren't the best :)

jaychia:
Closing in favor of https://github.com/unitycatalog/unitycatalog/pull/116 which looks way more fully featured :)

PR #28, opened 2024-06-14
addresses typos, syntax and rewordings after reading through tutorial

epps:
**PR Checklist**

**Description of changes**

This PR introduces typos fixes, edits, Markdown syntax changes, and rewordings based on my understanding of the Tutorial document. I tried to only make chances that enhanced the document without altering the original voice of the author(s).

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -167,11 +167,11 @@ bin/uc volume read --full_name unity.default.json_files --path dir1
 ```
 
 Now let's try creating a new external volume. First physically create a directory with some files in it.
-For example in the project root directory, create a directory `tmp/myVolume` and put some files in it.
+For example, create a directory `/tmp/myVolume` and put some files in it.
 Then create the volume in UC.
 
 ```sh
-bin/uc volume create --full_name unity.default.myVolume --storage_location tmp/myVolume
+bin/uc volume create --full_name unity.default.myVolume --storage_location /tmp/myVolume
 ```
 
 Now you can see the contents of this volume.
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -1,27 +1,34 @@
 # Tutorial
+
 Let's take Unity Catalog for spin. In this tutorial, we are going to do the following:
+
 - In one terminal, run the UC server.
 - In another terminal, we will explore the contents of the UC server using the UC CLI,
   which is an example UC connector provided to demonstrate how to use the UC SDK for various assets,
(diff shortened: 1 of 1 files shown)


epps:
Thank you to the creators and contributors for open sourcing Unity Catalog! I'm excited to see how my team and I will be able to incorporate it into our work!

docs/tutorial.md
```diff
@@ -117,102 +137,111 @@ To quit DuckDB, run the command `Ctrl+D` or type `.exit` in the DuckDB shell.
...
-```
+Voila! You have read the content of a file stored in a volume. We can also list the contents of any subdirectory. For example:
+
+```sh
 bin/uc volume read --full_name unity.default.json_files --path dir1
 ```
 
 Now let's try creating a new external volume. First physically create a directory with some files in it.
-For example in the project root directory, create a directory `/tmp/myVolume` and put some files in it.
+For example in the project root directory, create a directory `tmp/myVolume` and put some files in it.
```
tdas, line 170: @epps why are you changing `/tmp/myVolume` to `tmp/myVolume`. I believe the intention is to use `/tmp`

tdas (changes requested):
@epps thank you for this improvement. the only thing that needs to change is that we intentionally want to use `/tmp` and not `tmp`. so unless the current instructions are wrong and does not work, please revert that change.

docs/tutorial.md
epps: @tdas, based on the wording _"... in the project root directory ..."_ I'd understood that the `tmp` directory should be in the root of the repo, not the root of the file system. With that said, I _was_ confused because `/tmp` would also make sense. Happy to change back if the correct understanding is the latter (i.e. `/tmp/myVolume`), though I think we should also change the wording to something along the lines of _"... create `myVolume` in `/tmp` ..."_.

epps:
@tdas, happy to help! I'm really excited by UC, so I'm grateful for all the stellar work your team is doing in this area.

I left a clarifying comment and a question, so once you respond, I will make the necessary changes.

docs/tutorial.md
tdas: Yeah, that was the mistake. it should not have said `"... in the project root directory ..."`. please remove that and then keep the `/tmp`

epps:
@tdas, ready for re-review at your convenience.

tdas (approved):
Thank you so much for your interest in UC and your contribution! This is good to go. Merging it.

PR #37, opened 2024-06-14
Improving SBT configuration

edmondop:
**PR Checklist**

**Description of changes**

Over time SBT build files become difficult to maintain and when the same library is used in multiple projects, it becomes cumbersome to update it everywhere. This PR propose a modular approach for organizing dependencies into a `Dependencies.scala` and a `Version.scala`. Updates to new version can be done by touching only the Version.scala


sullis:
LGTM

build.sbt
```diff
@@ -57,21 +58,15 @@ lazy val client = (project in file("clients/java"))
...
-      "net.aichler" % "jupiter-interface" % JupiterKeys.jupiterVersion.value % Test,
-    ),
-
+    javaCheckstyleSettings(file("dev") / "checkstyle-config.xml"),
+    libraryDependencies ++=
+      Dependencies.findBugs ++
+      Dependencies.guava ++ 
+      Dependencies.hadoop ++ 
+      Dependencies.jackson(includeAnnotations = true) ++
+      Dependencies.Jakarta.annotation ++
```
tdas, line 67: While I like this clustering of libraries and simplifying the high level view of dependencies, I don't like the loss of information of which dependencies are for prod vs test. Can you somehow indicate that in this high level glance?

build.sbt
```diff
@@ -57,21 +58,15 @@ lazy val client = (project in file("clients/java"))
...
-
+    javaCheckstyleSettings(file("dev") / "checkstyle-config.xml"),
+    libraryDependencies ++=
+      Dependencies.findBugs ++
+      Dependencies.guava ++ 
+      Dependencies.hadoop ++ 
+      Dependencies.jackson(includeAnnotations = true) ++
+      Dependencies.Jakarta.annotation ++ 
+      Dependencies.junit4,
+    libraryDependencies ++= Dependencies.junit5.value,
```
tdas, line 69: Why is this dependency separately added

tdas (changes requested):

build.sbt
edmondop: This is becomes Junit5 is not a simple dependency, it's a setting that needs to be evaluated in order by SBT, see the `value` keyboard invoked to retrieve the setting value here . https://github.com/unitycatalog/unitycatalog/pull/37/files#diff-ad2642dc77b3679e4ad2c7d3b2f59a2dcf48c1bf5b93a558a4fa44828315a34cR94-R97

Unfortunately concatenating playing `Seq[ModuleID]` and `Def.Initialize[Seq[ModuleID]`` doesn't work. I haven't tried if we put junit5 as a first element on the list like so

```scala
  libraryDependencies ++= Dependencies.junit5.value ++ Dependencies.others
```

happy to try if you think that's better

build.sbt
edmondop: I think what we can do for junit4 for for example is to write something like that:

```scala
     Dependencies.Jakarta.annotation ++ 
      Dependencies.junit4.map(_ % Test),
```

if we want to ensure we don't define the scope of dependencies in `Dependencies.scala` but only in `build.sbt`.

An alternative approach could be to create an object inside Dependencies and have something like so:

```scala
     Dependencies.Compile.Jakarta.annotation ++ 
      Dependencies.Test.junit4 ++
      Dependencies.Provided.javax
```

I like this last idea, using the nested support of Scala objects as a form of namespace to be more explicit about the scope of a dependency?

tdas: While i like the structure of `Dependencies.Provided.javax`, managing it is more work. If something needs to be converting from compile to provided scope, then 2 locations need to change - Dependencies AND build.sbt. 

In general, there should be a separate of concerns
- Dependencies.scala - defines the all the dependent libraries that any module in SBT can depend on. It will ensure that all modules depend on the correct and consistent version of each library. It should not care about which SBT module is using which dependency in what way (compile or test).
- Versions.scala - defines the versions cleanly. should not care about which SBT module is using that version in what way.
- build/sbt - combines the information above to define a module. So build.sbt has the responsibilty to define whether a particular dependency is compile, or test, or provided.

So please update this PR with this approach `Dependencies.junit4.map(_ % Test),`

tdas: @edmondop any thoughts?

edmondop: That makes sense. I'll fix this today

edmondop: Fixed with [`49bf991` (#37)](https://github.com/unitycatalog/unitycatalog/pull/37/commits/49bf991ceff2a7e89e7cbf05e5829c9b2c46c458)

vikrantpuppala:
Sorry @edmondop for this PR slipping through the cracks, I will take up the review for this and get this through.

Could you please update it with the latest `main` branch so that all tests run and we can ensure no breaks. Apologies again and thank you for contributing!

PR #38, opened 2024-06-14
Storing Parent Entity Id's in All Child Entities

ravivj-db:
* Storing Parent Id's instead of names in all child entities.
* This keeps parent child relationships consistent on parent entity name updates.
* Added test cases to test out similar scenarios.
* Move all validation (entity not found, entity already existing) under a single transaction.

### Testing
Ran All Tests by running `build/sbt clean test` and they succeed.


ramesh-db (changes requested):
Please remove h2db.mv.db

ravivj-db:
Since our table/DAO definitions have changed, we need to replace h2db.mv.db file for the quickstart to work smoothly.

ramesh-db (approved):

Issue #44, opened 2024-06-15
Delete for parent entities should be handled properly.


ravivj-db:
**Describe the bug**

When deleting a parent entity like `catalog` or `schema`, we should first check if there are child entities of the one being deleted. If present, the delete should error out. If `force` flag is set to true, the entire hierarchy subtree should be deleted.
 
**To Reproduce**

1. Create a schema underneath a catalog
2. Try and delete the catalog
3.  It will succeed

**Expected behavior**

The api request should error saying child `Schema` entities are present.

PR #61, opened 2024-06-18
Implementing force delete for catalog and schema

ravivj-db:
**PR Checklist**

**Description of changes**

* Proper Implementation of delete in schema and catalog taking into account the force flag.
* If force flag is set to true, all child entities are deleted in the same transaction. 
* If force flag is set to false, and child entities are present, error is thrown

**Related Bug**

Solve the linked [issue](https://github.com/unitycatalog/unitycatalog/issues/44)

**Testing**

Test cases for the above mentioned scenarios have been added and are running fine.


server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
```diff
@@ -125,12 +126,22 @@ public CatalogInfo updateCatalog(String name, UpdateCatalog updateCatalog) {
...
     }
 
-    public void deleteCatalog(String name) {
+    public void deleteCatalog(String name, boolean force) {
         try (Session session = sessionFactory.openSession()) {
             Transaction tx = session.beginTransaction();
             try {
                 CatalogInfoDAO catalogInfo = getCatalogDAO(session, name);
                 if (catalogInfo != null) {
+                    List<SchemaInfoDAO> schemas = schemaRepository.listSchemas(session, catalogInfo.getId(), Optional.of(100));
```
haogang, line 135: Is 100 the page size or the max result? The logic is wrong if it's the max result

It's inefficient to list all schemas when `force` is not specified. Can we list one schema first to check, and only list others when `force` is specified?

server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
```diff
@@ -125,12 +126,22 @@ public CatalogInfo updateCatalog(String name, UpdateCatalog updateCatalog) {
...
+                    if (schemas != null & !schemas.isEmpty()) {
```
haogang, line 136: `&&` ?

server/src/main/java/io/unitycatalog/server/persist/SchemaRepository.java
```diff
@@ -176,22 +185,73 @@ public SchemaInfo updateSchema(String fullName, UpdateSchema updateSchema) {
...
             }
         }
     }
+
+    public void deleteSchema(Session session, UUID catalogId, String schemaName, boolean force) {
+        SchemaInfoDAO schemaInfo = getSchemaDAO(session, catalogId, schemaName);
+        if (schemaInfo != null) {
+
+            // handle tables
+            List<TableInfoDAO> tables = tableRepository.
```
haogang, line 216: It's better to unify the handling for different entity types so that we don't miss new entities types.

clients/java/src/main/java/io/unitycatalog/client/api/SchemasApi.java
```diff
@@ -217,7 +219,22 @@ private HttpRequest.Builder deleteSchemaRequestBuilder(String fullName) throws A
     String localVarPath = "/schemas/{full_name}"
         .replace("{full_name}", ApiClient.urlEncode(fullName.toString()));
 
-    localVarRequestBuilder.uri(URI.create(memberVarBaseUri + localVarPath));
+    List<Pair> localVarQueryParams = new ArrayList<>();
```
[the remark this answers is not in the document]
ravivj-db: This is autogenerated code. Currently we are not changing the client code unless there's an error in the logic.

server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
ravivj-db: Yes, you are correct. 100 currently stands for pageSize. We should leave it empty in order to get all the existing child entities.
I have implemented the similar logic for all entities (level-2, level-3) now :
* First check if a single child entity exists. 
* if not `force` throw exception
* else get a list of all child entities of that type by setting `maxResults` as empty/null and delete all of them.

server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
ravivj-db: thanks, good catch.

server/src/main/java/io/unitycatalog/server/persist/SchemaRepository.java
ravivj-db: There's no field currently in the db schema to indicate if an entity is level-3 entity (for e.g. table, function, volume). Also there is one different singleton instance of `repository` associated with each type(i.e.`TableRepository`, `FunctionRepository`, `VolumeRepository` are all different classes and we maintain singletons of each). Given this structure, I am not sure what would be a good way to unify the handling of all entity types.

haogang (approved):

Issue #66, opened 2024-06-19
Failure to build unity catalog server on Macbook


abrassel:
**Describe the bug**

The Armeria dependency does not include the netty classfiles necessary to successfully compile and run unity catalog. 
This can be hacked around by specifying the transitive dependency (using "netty-all"). It is not yet clear why we must specify the upstream dependency manually.

**To Reproduce**

1. `build/sbt package`
2. `bin/start-uc-server`

```
Exception in thread "main" java.lang.NoClassDefFoundError: io/netty/channel/ChannelPipeline
```

**Expected behavior**

I expected the script to successfully start the UC server.

**System [please complete the following information]:**

-   OS: Macos Sonoma (14.4.1 (23E224)), Apple M1 Max, Macbook Pro

rameshchandra:
@abrassel Thanks for finding the issue. 

I couldn't repro the issue on a new checkout of the repo on a Macbook Pro with MacOS Sonoma after deleting my .ivy2 and .sbt folders. Would love to see a reliable repro.

abrassel:
I'll let you know tomorrow!

rameshchandra:
@abrassel closing this issue since PR #67 is merged. Please reopen if you still see this issue.

PR #67, opened 2024-06-19
Force coursier cache to make uc server run

abrassel:
**PR Checklist**

**Description of changes**

This forces the use of the Coursier cache in all circumstances. Previously, the build was sometimes a bit unreliable. It seems as though it was pulling from multiple caches and failing to resolve all transitive dependencies.

`netty-transport` was the specific culprit.

Addresses #66 .

Tested by running `bin/start-uc-server`


tdas:
As @ramesh-db said in the issue #66, we are all running this on a macbook and we have not seen this issue. lets first dicuss on the issue and agree whether this change is needed.

abrassel:
Thank you @ravivj-db for root causing and solving this issue. 

It was due to sbt using a mix between the `.m2` and `Coursier` caches to resolve the classpath. @ravivj-db altered the `build.sbt` file to only use `Cousier`, which I was able to test and verify as an effective solution.

@ramesh-db I believe this PR is ready for merge.

tdas (approved):

tdas:
Thank you for fixing this @ramesh-db

Issue #21, opened 2024-06-14
List tables API gives one more (empty) page and terminates with empty string page_token instead of null


jaychia:
When performing a tables.list operation, it seems that the server returns the empty string `""` instead of None as documented in the spec.

<img width="884" alt="image" src="https://github.com/unitycatalog/unitycatalog/assets/17691182/c7d88f8d-5081-48f4-bdaa-fd08d054b133">

On a default run of unity catalog:

```
!pip install unitycatalog

import unitycatalog

client = unitycatalog.Unitycatalog(base_url="http://127.0.0.1:8080/api/2.1/unity-catalog/")
response = client.tables.list(catalog_name="unity", schema_name="default")

# assert response.next_page_token is None, f"This will fail, with a non-empty token: {response.next_page_token}"

response = client.tables.list(catalog_name="unity", schema_name="default", page_token=response.next_page_token)
# assert len(response.tables) > 0, "This will fail, because response has no tables"

# assert response.next_page_token is None, f"This will fail, with empty token: {response.next_page_token}"
```

This led to an unexpected infinite loop on my client-side code! Expected behavior: return `null` for the token when the page stream is exhausted.

abrassel:
Mind if I take this issue?

tdas:
@abrassel oh yes! please take this up and make a PR! thank you so much in advance!

abrassel:
thank you! @tdas can you assign it to me?

abrassel:
@jaychia https://github.com/unitycatalog/unitycatalog/pull/70

vikrantpuppala:
Fixed by #70

PR #70, opened 2024-06-19
Fix next page token

abrassel:
## PR Checklist

## Description of changes

Addresses #21 

When performing a tables.list operation, it seems that the server returns the empty string "" instead of None as documented in the spec.

Fixed this by making the paginated request return `null` instead.

## Testing

On a default run of unity catalog, did this manual test:
```
import unitycatalog

client = unitycatalog.Unitycatalog(base_url="http://127.0.0.1:8080/api/2.1/unity-catalog/")
response = client.tables.list(catalog_name="unity", schema_name="default")

assert response.next_page_token is None, f"This will fail, with a non-empty token: {response.next_page_token}"

response = client.tables.list(catalog_name="unity", schema_name="default", page_token=response.next_page_token)
assert len(response.tables) > 0, "This will fail, because response has no tables"
assert response.next_page_token is None
```

Any tips for testing this effectively?


abrassel:
@tdas can you review this PR?

ramesh-db (approved):
@abrassel thank you for fixing this.

Can you add a test for the fix? Will merge after the test.

abrassel:
@ramesh-db Thanks for the review! Do you have a code pointer for this repo's testing framework?

Fokko (approved):

ravivj-db:
Hi @abrassel , you can checkout the file BaseTableCRUDTest and SdkTableCRUDTest. Ideally we would have wanted to add this check directly to the BaseTableCRUDTest, however we have currently kept the return type of `listTables` as  List<TableInfo> (which omits the next page token)  to keep compatibility between Sdk and Cli Tests (in CLI we are not handling next page tokens). A workaround would be to create a small ListTablesTest at the level of SdkTableCRUDTest (which directly extends BaseCRUDTest or BaseServerTest) and just creates a couple of tables and Tests the output of the listTables. Then it can read the response as `ListTableResponse` and check the `nextPageToken` field.

abrassel:
Hi @ramesh-db I added a test. We should be good to go now.

tdas (approved):

Issue #62, opened 2024-06-18
While creating external tables, the CLI should initiate a delta log if the directory exists but is empty.


ravivj-db:
**Describe the bug**

Currently, the CLI only initiates the delta log if the local directory doesn't exists. It should ideally initiate the delta even for the case when the directory exists but is empty.

**To Reproduce**

1. Create the local directory matching the external table location but leave it empty.
2. Create an external table with UC OSS, specifying the storage location as a local absolute path.
3. Read Table from the CLI will error out as no delta table exists at the path.

**Expected behavior**

Delta log should be initialized and read table should succeed.

PR #72, opened 2024-06-19
EXTERNAL table creation for existing directory

ravivj-db:
**PR Checklist**

**Description of changes**

This is related to the preprocessing step of EXTERNAL table creation using CLI for local paths. Currently the CLI was checking if the directory does not exist, before proceeding to create a delta table log at the location. This was the only case for which EXTERNAL table creation succeeded. 

Ideally, new EXTERNAL table creation should succeed even for the case when there's a delta log already present at the location (although we should check if the schema confirms). If the directory does not exist or there's no delta log at the location, it should create one. The current PR implements the same.

The PR takes care of the [issue](https://github.com/unitycatalog/unitycatalog/issues/62)


tdas (changes requested):
need unit test to confirm the behavior change.

ravivj-db:
Added Tests.

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
```diff
@@ -109,11 +110,13 @@ private static void handleTableStorageLocation(String storageLocation, List<Colu
...
-                // try and initialize the directory and initiate delta log at the location
-                try {
-                    DeltaKernelUtils.createDeltaTable(path.toUri().toString(), columnInfos, null);
-                } catch (Exception e) {
+            // try and initialize the directory and initiate delta log at the location
+            try {
+                DeltaKernelUtils.createDeltaTable(path.toUri().toString(), columnInfos, null);
+            } catch (Exception e) {
+                if (e.getCause() != null && e.getCause().getClass().equals(TableAlreadyExistsException.class)) {
+                    // TODO confirm the schema of the existing table matches the schema of the new table
```
rameshchandra, line 118: @ravivj-db Can you create a GH issue for this TODO?

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
```diff
@@ -109,11 +110,13 @@ private static void handleTableStorageLocation(String storageLocation, List<Colu
...
-            if (!(Files.exists(path))) {
-                // try and initialize the directory and initiate delta log at the location
-                try {
-                    DeltaKernelUtils.createDeltaTable(path.toUri().toString(), columnInfos, null);
-                } catch (Exception e) {
+            // try and initialize the directory and initiate delta log at the location
+            try {
+                DeltaKernelUtils.createDeltaTable(path.toUri().toString(), columnInfos, null);
+            } catch (Exception e) {
+                if (e.getCause() != null && e.getCause().getClass().equals(TableAlreadyExistsException.class)) {
```
rameshchandra, line 117: Can this just be `(e.getCause() instanceof TableAlreadyExistsException)`?

examples/cli/src/test/java/io/unitycatalog/cli/table/CliExternalTableCreationTest.java
```diff
@@ -0,0 +1,155 @@
...
+    public void setUp() {
+        super.setUp();
+        catalogOperations = new CliCatalogOperations(serverConfig);
+        schemaOperations = new CliSchemaOperations(serverConfig);
+        tableOperations = new CliTableOperations(serverConfig);
+        try {
+            catalogOperations.createCatalog(CATALOG_NAME, "Test catalog");
+            schemaOperations.createSchema(new CreateSchema().name(SCHEMA_NAME).catalogName(CATALOG_NAME).comment("Test schema"));
+        } catch (Exception e) {
+            // Ignore
```
rameshchandra, line 47: Shouldn't the test fail if create catalog / create schema fail?

ravivj-db: Added the fail clause in catch.

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
ravivj-db: Yes, that takes care of the null check as well.

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
ravivj-db: Done : https://github.com/unitycatalog/unitycatalog/issues/82

examples/cli/src/test/java/io/unitycatalog/cli/table/CliExternalTableCreationTest.java
```diff
@@ -0,0 +1,155 @@
...
+        String tablePath = "/tmp/" + UUID.randomUUID();
+        TableInfo tableInfo = tableOperations.createTable(new CreateTable().catalogName(CATALOG_NAME).schemaName(SCHEMA_NAME).name(TABLE_NAME).storageLocation(tablePath).columns(
+                List.of(
+                        new ColumnInfo().name("id").typeName(ColumnTypeName.INT).comment("id"),
+                        new ColumnInfo().name("name").typeName(ColumnTypeName.STRING).comment("name"))));
+        assertNotNull(tableInfo);
+        assertNotNull(tableInfo.getTableId());
+        assertEquals(tableInfo.getTableId(), tableOperations.getTable(TABLE_FULL_NAME).getTableId());
+        assertDoesNotThrow(() -> DeltaKernelUtils.readDeltaTable(tableOperations.getTable(TABLE_FULL_NAME).getStorageLocation(), null, 100));
+        deleteDirectory(Paths.get(tablePath));
```
rameshchandra, line 98: Nit: refactor this code (it is duplicated in all the three tests).

ravivj-db: Moved common logic to a function.

examples/cli/src/test/java/io/unitycatalog/cli/table/CliTableOperations.java
```diff
@@ -26,20 +27,24 @@ public CliTableOperations(ServerConfig config) {
...
         columns.deleteCharAt(columns.length() - 1);
 
-        String [] args = addServerAndAuthParams(List.of(
-                "table", "create" ,
-                "--full_name", createTableRequest.getCatalogName() + "." + createTableRequest.getSchemaName() + "." + createTableRequest.getName(),
-                "--columns", columns.toString(),
-                "--format", createTableRequest.getDataSourceFormat() != null ? createTableRequest.getDataSourceFormat().name() : null,
-                "--storage_location", createTableRequest.getStorageLocation() != null ? createTableRequest.getStorageLocation() : null
-        ), config);
+        List<String> argsList = new ArrayList<>();
```
rameshchandra, line 36: It this related to this change?

ravivj-db: This is was an error in the earlier implementation, and would lead to a null pointer exception if data format is not set in the options (we were always setting the option in our previous tests)

rameshchandra: can this be a separate PR?

ravivj-db: There are two extra changes besides the main ones :
* Remove the null check for `typeText` while converting ColumnInfo to StructType
* If DataSourceFormat is null in CreateTable request do not pass it as a null CLI Param.

Both are necessary for the current test to pass properly as I only want to pass the minimal information to make the test run.

* typeText is currently not used at all for the conversion.
* DataSourceFormat if not specified, defaults to Delta in CLI implementation and hence not a required CLI param.

rameshchandra (approved):
LGTM

tdas (approved):

PR #78, opened 2024-06-20
fix duplicate url set call in table converter

alexreid-db:
**PR Checklist**

**Description of changes**

Removed duplicate (and non null checking) url set call in table converter code. Also, cleaned up some unused imports and exceptions thrown.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/converters/TableInfoConverter.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/converters/TableInfoConverter.java
@@ -89,9 +89,9 @@ public class TableInfoConverter {
                 .createdAt(dto.getCreatedAt() != null ? new Date(dto.getCreatedAt()) : new Date())
                 .updatedAt(dto.getUpdatedAt() != null ? new Date(dto.getUpdatedAt()) : new Date())
                 .columnCount(dto.getColumns() != null ? dto.getColumns().size() : 0)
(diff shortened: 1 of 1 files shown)


Fokko (approved):
LGTM!

server/src/main/java/io/unitycatalog/server/persist/converters/TableInfoConverter.java
```diff
@@ -94,7 +92,6 @@ public static TableInfoDAO convertToDAO(TableInfo dto) {
                 .url(dto.getStorageLocation() != null ? dto.getStorageLocation() : null)
                 .type(dto.getTableType().toString())
                 .dataSourceFormat(dto.getDataSourceFormat().toString())
-                .url(dto.getStorageLocation())
```
rameshchandra, line 97: @alexreid-db Isn't `dto.getStorageLocation() != null ? dto.getStorageLocation() : null` equivalent to `dto.getStorageLocation()`? So, shouldn't this line be left as is and line 94 be removed?

rameshchandra (changes requested):
Thank you for the fix.

server/src/main/java/io/unitycatalog/server/persist/converters/TableInfoConverter.java
alexreid-db: yeah, sorry, didn't look at the line above closely enough

rameshchandra (approved):
LGTM

PR #80, opened 2024-06-21
Remove unnecessary plugin

darklore:
**PR Checklist**

**Description of changes**

Coursier support is enabled by default in sbt 1.3 or later
https://get-coursier.io/docs/sbt-coursier
So remove coursier sbt plugin


tdas:
@ravivj-db you were mentioning something about coursier cache vs m2 cache. so isnt this plugin needed?

ravivj-db:
If I remove that line. I get the following error during sbt compile : 
```
error: not found: value CoursierPlugin
enablePlugins(CoursierPlugin)
              ^
[error] Type error in expression
```
Hence we need to keep that import in `plugins.sbt`

darklore:
Since sbt 1.3, coursier is used by default. Therefore, there is no need to `enablePlugins(CoursierPlugin)` and `useCoursier := true`.
https://www.scala-sbt.org/1.x/docs/sbt-1.3-Release-Notes.html#Library+management+with+Coursier

ravivj-db:
@darklore  If we were not using `useCoursier := true` we were getting a mix of m2 and Coursier dependencies in our classpath. To force the use of Coursier we'll have to keep just this setting `useCoursier := true` (we can remove the remaining ones). Can you please make the necessary changes and also make sure that Sbt tests are running fine. Also can you please modify the PR description to an appropriate one  (for e.g. : Remove sbt coursier plugin import as available by default)  ?

darklore:
Sorry, I couldn't resolve the sbt test failure so I withdraw this pull request.
Thank you for the review.

PR #109, opened 2024-06-29
Configure code formatter for Java code

nastra:
**PR Checklist**

**Description of changes**

This configures and applies spotless for Java code. The PR is currently WIP and requires some additional documentation so that it's easier for devs to get started and include `bin/sbt spotlessApply` into their workflow when Java code formatting is off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,19 @@ See the full [tutorial](docs/tutorial.md) for more details.
   build/sbt generate
   ``` 
   This will regenerate the OpenAPI data models in the UC server and data models + APIs in the client SDK.
+
+
+## Setting up IDE
+
+Java code adheres to the [Google style](https://google.github.io/styleguide/javaguide.html), which is verified via `build/sbt javafmtCheckAll` during builds.
+In order to automatically fix Java code style issues, please use `build/sbt javafmtAll`.
+
+### Configuring Code Formatter for Eclipse/IntelliJ
+
+Follow the instructions for [Eclipse](https://github.com/google/google-java-format#eclipse) or
+[IntelliJ](https://github.com/google/google-java-format#intellij-android-studio-and-other-jetbrains-ides) to install the **google-java-format** plugin (note the required manual actions for IntelliJ).
+
+
 ### Using more recent JDKs
 
 The build script [checks for a lower bound on the JDK](./build.sbt#L14) but the [current SBT version](./project/build.properties)
--- a/build.sbt
+++ b/build.sbt
@@ -86,6 +86,7 @@ def javaCheckstyleSettings(configLocation: File) = Seq(
 
 lazy val client = (project in file("clients/java"))
   .enablePlugins(OpenApiGeneratorPlugin)
+  .disablePlugins(JavaFormatterPlugin)
   .settings(
     name := s"$artifactNamePrefix-client",
     commonSettings,
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -30,6 +30,8 @@ addSbtPlugin("org.openapitools" % "sbt-openapi-generator" % "7.5.0")
 
 addSbtPlugin("com.github.sbt" % "sbt-license-report" % "1.6.1")
 
+addSbtPlugin("com.lightbend.sbt" % "sbt-java-formatter" % "0.8.0")
+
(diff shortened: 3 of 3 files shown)


creechy:
+1 for adding and enforcing code formatting. You don't mention the code formatting standard, but I assume its Google Java Format. Just going to throw it out there to maybe investigate Palantir Java Format. IMO Google Java Format has issues, particularly with closures/lambdas. I don't know a lot about Palantir Java Format but you can see some comparisons between the two here - https://github.com/palantir/palantir-java-format - and it seems like Spotless supports it too.

tdas:
@nastra some initial feedback
1. Its really hard to review this much amount of code difference. as I suggested in the issue, is it possible to break it down into smaller PRs. 
2. we should not modify client as its generated. so spotless should ignore client

.github/workflows/spotless.yml
```diff
@@ -0,0 +1,27 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: Check code style
```
tdas, line 6: I am trying to standardize these names to be nouns. so I suggest "Code style check"

nastra:
> @nastra some initial feedback
> 
> 1. Its really hard to review this much amount of code difference. as I suggested in the issue, is it possible to break it down into smaller PRs.
> 2. we should not modify client as its generated. so spotless should ignore client. Similar models in the server should be ignored as well as they are generated.

I included the formatting changes initially as an experimentation to give people an overview of how things would look, but I agree that this is difficult to review, so maybe as a first step we should only add the respective configuration and later apply the formatting.

@creechy just FYI I had to switch from **sbt-spotless** to https://github.com/sbt/sbt-java-formatter due to a bug in **sbt-spotless** (it's unfortunately externally maintained and the last release was in 2020). **sbt-java-formatter** produces the exact same formatting results with the one tradeoff that you can't switch the formatter as it uses **google-java-format**.

nastra:
@tdas I've updated the PR to only include the the respective configuration for auto code formatting. Code can be automatically formatted by running `build/sbt javafmtAll` (the `All` includes the `Test` configuration). 
That code style is currently not enforced, which we can later change.
I've added a note on how to configure IntelliJ so that code formatting in the IDE and on the cmd line produces the same results.

README.md
```diff
@@ -153,6 +153,17 @@ See the full [tutorial](docs/tutorial.md) for more details.
   ``` 
   This will regenerate the OpenAPI data models in the UC server and data models + APIs in the client SDK.
 
+## Setting up IDE and Code Style
```
tdas, line 156:
```suggestion
## Setting up IDE
```

tdas:
This PR looks very good! Lets just fix the conflicts, and then I will merge this. thank you @nastra for do this!

nastra:
Thanks @tdas, I've resolved the conflicts and pushed a new version

README.md
```diff
@@ -163,6 +163,19 @@ See the full [tutorial](docs/tutorial.md) for more details.
   build/sbt generate
   ``` 
   This will regenerate the OpenAPI data models in the UC server and data models + APIs in the client SDK.
+
+
+## Setting up IDE
```
tdas, line 168: I will merge this PR but I think this section needs basic instructions on how to set up IntelliJ / Eclipse. Code formatting is the not the only thing in the IDE set up.. is it?

tdas: can you improve it in a follow up PR?

tdas (approved):

README.md
nastra: @tdas just to confirm: Installing the `google-java-format` plugin is really the only step required and https://github.com/google/google-java-format#intellij-android-studio-and-other-jetbrains-ides contains all the steps required to install + enable it. I figured it should be fine to refer to the original docs for this rather than try and duplicate something here which might eventually get outdated.
Let me know if you still think that we should add some more text here

tdas: The codestyle part is fine. But since we are making a general section on "Setting up IDE", I think its good to say how to open up the project in IntelliJ or Eclipse. Just a simple thing like load the build.sbt (at least for IntelliJ that I use) is sufficient.

nastra: I've opened https://github.com/unitycatalog/unitycatalog/pull/170 to address that

PR #114, opened 2024-06-30
use LOGGER.error to output Exception stack for Problem localization

597365581:
use LOGGER.error to output Exception stack for Problem localization

**PR Checklist**

**Description of changes**

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/base/BaseCRUDTest.java
+++ b/server/src/test/java/io/unitycatalog/server/base/BaseCRUDTest.java
@@ -1,41 +1,40 @@
 package io.unitycatalog.server.base;
 
+import static io.unitycatalog.server.utils.TestUtils.CATALOG_NAME;
+import static io.unitycatalog.server.utils.TestUtils.CATALOG_NEW_NAME;
+
 import io.unitycatalog.server.base.catalog.CatalogOperations;
+import java.util.Optional;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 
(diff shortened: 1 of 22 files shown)


597365581:
@tdas 
use LOGGER.error to output Exception stack for Problem localization ，please help to review

server/src/main/java/io/unitycatalog/server/utils/JsonUtils.java
```diff
@@ -48,8 +52,10 @@ public static void printJsonAsTable(String jsonString) {
             }
 
         } catch (JsonMappingException e) {
+            LOGGER.error("printJsonAsTable happen JsonMappingException,jsonString:{}",jsonString,e);
```
vikrantpuppala, line 55: can we please format this better and add a meaningful message?

server/src/main/java/io/unitycatalog/server/utils/JsonUtils.java
```diff
@@ -48,8 +52,10 @@ public static void printJsonAsTable(String jsonString) {
...
             System.out.println(jsonString);
```
vikrantpuppala, line 56: Can we remove the system out if we add the log?

597365581: I think it's no problem

server/src/main/java/io/unitycatalog/server/utils/JsonUtils.java
597365581: Ok，it's no problem，thanks

597365581 (commented):
I will modify

Issue #105, opened 2024-06-28
Cloud Storage support in Iceberg REST catalog implementation


alexreid-db:
**Is your feature request related to a problem? Please describe.**

Currently, if Iceberg metadata for a table is stored in cloud storage (s3, gcs, adls), the catalog is not able to read it / load the table.

**Describe the solution you would like**

Handle reading metadata with cloud specific clients / fileIO.

**Additional context**

loadTable throws 500 - "Provider \"s3\" not installed"

PR #123, opened 2024-07-01
add support for tables located in S3 in Iceberg REST catalog

alexreid-db:
**PR Checklist**

**Description of changes**

This PR introduces the ability to load Iceberg tables that are located in S3. (Previously only local path was supported). This adds new dependencies on iceberg-aws module as well as aws sdk v2. This also introduces new testing frameworks - mockito and S3Mock in order to better test and simulate reading an iceberg metadata out of S3 with S3FileIO (as well as starting to use JUnit5 instead of JUnit4).

**Related Issues**

This PR covers the basic and initial S3 support addressing issue https://github.com/unitycatalog/unitycatalog/issues/105


tdas:
Thank you @alexreid-db for this! How can we test this? Could you take a look at the current unit tests and see if these code paths have enough coverage to be confident that we are not breaking local path support? If not, could you add better unit tests to ensure local path is not broken? 

For S3, have you tested manually?

alexreid-db:
Yeah, sorry, meant to create this as Draft until I added the tests and description.

This is a pre-requisite for any credential vending, etc.

Yes, I've tested this locally with OSS Spark. Caveats until credential vending support is that both the client and server need credentials in the default credentials provider chain that has access to the table location(s).

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
```diff
@@ -241,7 +241,7 @@ public void testTable() throws ApiException, IOException, URISyntaxException {
       LoadTableResponse loadTableResponse =
         RESTObjectMapper.mapper().readValue(resp.contentUtf8(), LoadTableResponse.class);
       assertThat(loadTableResponse.tableMetadata().metadataFileLocation())
-              .isEqualTo(this.getClass().getResource("/metadata.json").toURI().toString());
+              .isEqualTo(Objects.requireNonNull(this.getClass().getResource("/metadata.json")).toURI().toString());
```
alexreid-db, line 244: unrelated fix for warning

creechy (approved):
lgtm

alexreid-db:
@haogang @tdas this PR is ready for review

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -0,0 +1,35 @@
...
+    this.fileIOFactory = fileIOFactory;
+  }
+
+  public TableMetadata readTableMetadata(String metadataLocation) throws IOException {
+    URI metadataLocationUri = URI.create(metadataLocation);
+    // TODO: cache fileIO
+    FileIO fileIO = fileIOFactory.getFileIO(metadataLocationUri);
+
+    TableMetadata tableMetadata;
+    if (fileIO == null) {
```
haogang, line 26: This fallback logic looks quite ad-hoc. Can we define a subclass of FileIO that handles the local file system read, and return it in the FileIOFactory?

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -44,6 +43,7 @@ public IcebergRestCatalogService(
     this.catalogService = catalogService;
     this.schemaService = schemaService;
     this.tableService = tableService;
+    this.metadataService = new MetadataService(new FileIOFactory());
```
haogang, line 46: I feel it's better to make either the MetadataService or the FileIOFactory depdency-injectable so that we don't need to spy on them in unit tests, especially for classes that has side effect such as performing I/O.

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
alexreid-db: Seemed a little overkill to write a FileIO impl just to read a local metadata file. We can however fallback to HadoopFileIO (which handles local filesystem) -- this just introduces a dependency on hadoop-common

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
alexreid-db: I mean I can pass a metadata service into this constructor if you'd like, but there really isn't any dependency injection happening in this project as it is (and nothing else relies on metadata service currently).

I also don't really follow the argument on not spy'ing/mock'ing, are you saying we should write a whole MockFileIOFactory class just to test?

haogang: Let me rephrase. I want to minimize the use of spy() because it's often unclear which part of the implementation is mocked and which part is tested. But mocking is fine. We don't need to write a separate class just for testing.

In `MetadataServiceTest`, can we just mock `FileIOFactory.getFileIO` so that it returns different result for each test case? We should have a separate unit test for `FileIOFactory.getFileIO` if needed.

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
alexreid-db: okay, well I added a very simple fileIO impl but I'm not sure this should be the long term solution and added a TODO about it

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
alexreid-db: that's fine, I switched from spy to mock

haogang (approved):
LGTM

PR #130, opened 2024-07-02
Replace assert with assertThat in tests

nastra:
**PR Checklist**

**Description of changes**

The `assert` keyboard in Java needs to be explicitly enabled (via `-ea`), otherwise it doesn't do anything and gets optimized away. There are also a few `assert` usages in web service code and those should be replaced either with `Preconditions.checkNotNull` or with a Validation API

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/base/function/BaseFunctionCRUDTest.java
+++ b/server/src/test/java/io/unitycatalog/server/base/function/BaseFunctionCRUDTest.java
@@ -11,6 +11,7 @@ import io.unitycatalog.server.base.schema.SchemaOperations;
 
 import java.util.List;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.*;
 import static io.unitycatalog.server.utils.TestUtils.*;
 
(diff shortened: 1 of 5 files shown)


server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableCRUDTest.java
```diff
@@ -59,7 +58,11 @@ public void testListTablesWithNoNextPageTokenShouldReturnNull() throws Exception
         createCommonResources();
         TableInfo testingTable = createDefaultTestingTable();
         ListTablesResponse resp = localTablesApi.listTables(testingTable.getCatalogName(), testingTable.getSchemaName(), 100, null);
-        assert resp.getNextPageToken() == null;
-        assert Objects.equals(resp.getTables(), List.of(testingTable));
+        assertThat(resp.getNextPageToken()).isNotNull();
+        assertThat(resp.getTables()).hasSize(1)
+                .first()
+                .usingRecursiveComparison()
```
tdas, line 64: this seems to be a change in the test logic. @ravivj-db could you double check this?

nastra: this is because previously the `assert` wasn't executed and the condition was actually never met (the lists weren't equal)

server/src/test/java/io/unitycatalog/server/base/function/BaseFunctionCRUDTest.java
```diff
@@ -11,6 +11,7 @@
 
 import java.util.List;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.*;
```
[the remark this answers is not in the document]
nastra: yeah those are still used across tests until Junit4 assertions are switched to AssertJ assertions

tdas (approved):

tdas:
@nastra Thank you very much for the fixes and improvements!

PR #142, opened 2024-07-03
Output enhancements for `uc` command

creechy:
**Description of changes**

This PR addresses Issue https://github.com/unitycatalog/unitycatalog/issues/157

This change enhances the tabular output of the `uc` terminal command in a few ways
- Use the full width of the terminal for the table width
- Consider special columns like `NAME` and `ID` to be fixed with and always display their entire values
- Only fall back to equal sizes of non-special columns if any total line width is too long to fit in the terminal

The motivation is to ensure critical columns like object names are never clipped, which can make navigating entities challenging if you can't get the entire name. As an example, the sample table `marksheet_uniform` gets clipped to `marksheet..." which is not entirely useful.

Sample screenshots before this PR

image

image

And after

image

image


vikrantpuppala:
Thank you for making this change @creechy!
Could I request for a couple of additions here:
1. If you could create an issue for this change, that would be great. (If it already exists, could you please link it to this PR)
2. Could you add a few examples of what the CLI looked like without this PR vs with this PR? This would help make the improvements clear to everyone.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
```diff
@@ -167,35 +170,84 @@ public static void postProcessAndPrintOutput(CommandLine cmd, String output, Str
...
+                                }
+                            }
+                            int columnWidth = (widthRemaining) / (fields.size() - fixedWidthColumnCount);
+
+                            for (int i = 0; i < fields.size(); i++) {
+                                if (!fixedWidthColumns[i]) {
+                                    columnWidths[i] = columnWidth;
+                                }
+                            }
+                        }
```
vikrantpuppala, line 227: This looks great, could you please extract out these blocks of logic into separate functions to make this more readable, I think `postProcessAndPrintOutput` has become too big

creechy:
@vikrantpuppala
 
> 1. If you could create an issue for this change, that would be great. (If it already exists, could you please link it to this PR)

Added an issued and put it in the PR description.

> 2. Could you add a few examples of what the CLI looked like without this PR vs with this PR? This would help make the improvements clear to everyone.

Added some before and after screen shots in the description.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
```diff
@@ -151,14 +147,145 @@ public static String preprocess(String value, int length) {
...
+            }
+            at.addRow(field.getKey().toUpperCase(), valueString.toString()).setTextAlignment(TextAlignment.LEFT);
+            at.addRule();
+        });
+
+    }
+
+    private static int getOutputWidth() {
+        // TODO: maybe turn this into a command line parameter.
+        String envWidth = System.getenv().get("UC_OUTPUT_WIDTH");
```
vikrantpuppala, line 281: This is specifically to handle the failure in tutorial tests?

vikrantpuppala (commented):
Looks good overall, few minor comments

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
creechy: Yes. I didn't look into whether you could set a "terminal width" in github actions, but I doubt it. Alternatively I could change the logic to be... if we get a non-zero width from the terminal then use it, otherwise use the original 180 character width.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
```diff
@@ -151,14 +147,145 @@ public static String preprocess(String value, int length) {
...
+                }
+
+            }
+            maxLineWidth = Math.max(maxLineWidth, lineWidth);
+        }
+
+        return maxLineWidth;
+    }
+
+    private static void adjustColumnWidths(List<String> fields,
```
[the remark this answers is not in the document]
creechy: I added a little javadoc.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
```diff
@@ -151,14 +147,145 @@ public static String preprocess(String value, int length) {
...
+
+        for (int i = 0; i < fields.size(); i++) {
+            if (!fixedWidthColumns[i]) {
+                columnWidths[i] = columnWidth;
+            }
+        }
+    }
+
+    private static void processOutputAsRows(AsciiTable at, JsonNode node, int outputWidth) {
+        List<String> fields = getFields(node.get(0));
```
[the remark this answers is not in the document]
creechy: Done

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
```diff
@@ -151,14 +147,145 @@ public static String preprocess(String value, int length) {
...
+
+        int[] columnWidths = new int[fields.size()];
+        boolean[] fixedWidthColumns = new boolean[fields.size()];
```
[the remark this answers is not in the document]
creechy: Done

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
```diff
@@ -151,14 +147,145 @@ public static String preprocess(String value, int length) {
...
+                                           boolean[] fixedWidthColumns) {
+        int widthRemaining = outputWidth - fields.size();
+
+        int fixedWidthColumnCount = 0;
+        for (int i = 0; i < fields.size(); i++) {
+            widthRemaining -= fixedWidthColumns[i] ? columnWidths[i] : 0;
+            if (fixedWidthColumns[i]) {
+                fixedWidthColumnCount++;
+            }
+        }
```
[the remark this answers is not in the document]
creechy: Done. I implemented slightly different than the example because streams don't allow multiple terminating methods (in this case sum() and count()).

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
```diff
@@ -151,14 +147,145 @@ public static String preprocess(String value, int length) {
...
+        int columnWidth = (widthRemaining) / (fields.size() - fixedWidthColumnCount);
+
+        for (int i = 0; i < fields.size(); i++) {
+            if (!fixedWidthColumns[i]) {
+                columnWidths[i] = columnWidth;
+            }
+        }
```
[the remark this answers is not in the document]
creechy: Done.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
vikrantpuppala: I think it is fine for now, given we have a fallback default

.github/workflows/release-workflow.yml
```diff
@@ -15,9 +15,9 @@ jobs:
         with:
           java-version: '11'
           distribution: 'temurin'
-          cache: 'sbt'
+
       - name: Build tarball
-        run: sbt createTarball
+        run: build/sbt createTarball
```
vikrantpuppala, line 20: why is this change showing up?

vikrantpuppala:
Looks good for the most part, but I see random changes like the github workflows, tests etc showing up, can you check what's wrong?

.github/workflows/release-workflow.yml
creechy: > why is this change showing up?

I think its due to how merging/rebasing of main. I'll try to figure out how to fix this up.

creechy:
> Looks good for the most part, but I see random changes like the github workflows, tests etc showing up, can you check what's wrong?

@vikrantpuppala the other changes went away once I sync'ed the PR branch in my personal repo. Can you give it one final review now? thanks.

vikrantpuppala (approved):
lgtm, @haogang to confirm and merge, thanks!

haogang (approved):

PR #148, opened 2024-07-04
docs: fix server docs formatting

avriiil:
Minor fix to improve layout and legibility of s3 params

before:
Screenshot 2024-07-04 at 14 15 14

after:
Screenshot 2024-07-04 at 14 15 38


docs/usage/server.md
```diff
@@ -1,37 +1,40 @@
...
 ## Configuration
 
 The server config file is at the location `etc/conf/server.properties` (relative to the project root).
 
-- `server.env`: The environment in which the server is running. This can be set to `dev` or `test`. When 
- set to `test` the server will instantiate an empty in-memory h2 database for storing metadata.
- If set to `dev`, the server will use the file `etc/db/h2db.mv.db` as the metadata
- store. Any changes made to the metadata will be persisted in this file.
+- `server.env`: The environment in which the server is running. This can be set to `dev` or `test`. When
+  set to `test` the server will instantiate an empty in-memory h2 database for storing metadata.
```
MrPowers, line 18: Can we just put this all on the same line?  I don't like chopping up sentences on multiple lines.

MrPowers (commented):
LGTM, but I would prefer to put all sentences on a single line and now chop up sentences on multiple lines.

Chopping up sentences only works if everyone has exactly the same text editor settings.

Let me know what you think!

docs/usage/server.md
avriiil: Yup, good catch!

avriiil:
@MrPowers - yup good catch, not sure how that happened. Fixed it, should be good to go!

tdas (approved):

PR #149, opened 2024-07-04
docs: add tutorial to nav

avriiil:
This PR adds the tutorial to the nav side-bar


tdas:
Could you also add deployment to nav bar? then the tests will be fixed

MrPowers:
Thanks for adding this!  Looks like we need to fix a couple of links and then this PR will be good-to-go!

avriiil:
@MrPowers - this should be gtg now

MrPowers (approved):
LGTM

tdas (approved):

PR #155, opened 2024-07-05
add job to verify that generated files are not changed

vikrantpuppala:
**PR Checklist**

**Description of changes**

This PR ensures that generated files are never edited via a github action

### DIFF
--- a/.github/workflows/sbt-tasks.yml
+++ b/.github/workflows/sbt-tasks.yml
@@ -46,7 +46,7 @@ jobs:
         cache: 'sbt'
 
     - name: Generate OpenAPI Client and Server Model classes
-      run: sbt generate
+      run: build/sbt generate
 
     - name: Check Generated Files Differences
       run: |
@@ -54,9 +54,9 @@ jobs:
         SERVER_DIFF=$(git diff --name-only server/src/main/java/io/unitycatalog/server/model)
         
         if [[ -n $CLIENT_DIFF || -n $SERVER_DIFF ]]; then
-          echo "This PR contains changes to generated files. Making changes to generated files is not allowed."
-          echo "If this is intended for some reason, please speak to the project maintainers."
-          echo "Please revert or run 'sbt generate' locally and commit the changes."
+          echo "This PR contains changes to generated files that will be overridden when build/sbt generate is run again."
+          echo "If you have changed the UC REST APIs such that changes to generated files are expected, then please make sure all generated files have been properly updated in this PR."
+          echo "Otherwise, without changes to UC REST APIs, manual changes to the generated files are not acceptable."
           if [[ -n $CLIENT_DIFF ]]; then
             echo "Modified client files:"
             echo "$CLIENT_DIFF"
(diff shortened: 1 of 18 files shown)


.github/workflows/sbt-tests.yml
```diff
@@ -31,3 +31,42 @@ jobs:
...
+
+    - name: Set up JDK 11
+      uses: actions/setup-java@v4
+      with:
+        java-version: '11'
+        distribution: 'temurin'
+        cache: 'sbt'
+
+    - name: Generate OpenAPI Client and Server Model classes
+      run: sbt generate
```
tdas, line 49: `build/sbt generate`

.github/workflows/sbt-tests.yml
```diff
@@ -31,3 +31,42 @@ jobs:
...
+
+    - name: Check Generated Files Differences
+      run: |
+        CLIENT_DIFF=$(git diff --name-only clients/java)
+        SERVER_DIFF=$(git diff --name-only server/src/main/java/io/unitycatalog/server/model)
+        
+        if [[ -n $CLIENT_DIFF || -n $SERVER_DIFF ]]; then
+          echo "This PR contains changes to generated files. Making changes to generated files is not allowed."
```
tdas, line 57: this 
"This PR contains changes to generated files that will be overridden when `build/sbt generate` is run again. If you have changed the UC REST APIs such that changes to generated files are expected, then please make sure all generated files have been properly updated in this PR. Otherwise, without changes to UC REST APIs, manual changes to the generated files are not acceptable."

tdas (changes requested):
error message needs improvement

.github/workflows/sbt-tasks.yml
```diff
@@ -0,0 +1,71 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: SBT Tasks
```
tdas, line 6: I think "Tests" is fine. its nicely consistent with names of other actions.
Maybe we can call it "Unit tests" which is much meaningful

.github/workflows/sbt-tasks.yml
```diff
@@ -0,0 +1,71 @@
...
+      with:
+        java-version: '11'
+        distribution: 'temurin'
+
+    - name: Run tests
+      run: build/sbt test
+    - name: Run license check
+      run: build/sbt licenseCheck
+    - name: Run tests with coverage
+      run: build/sbt jacoco
```
tdas, line 33: does this run  all the tests all of over again? If not, then maybe we should rename this to "Generate coverage". If yes, then we should just ensure that tests run only once.

.github/workflows/sbt-tasks.yml
tdas: its fine if it does additional non-unit-test type work. dont have to call it tasks.

tdas (approved):

Issue #110, opened 2024-06-29
Updating a catalog/schema without needing to rename it


VillePuuska:
**Is your feature request related to a problem? Please describe.**

I would like to be able to update the comment on a catalog or schema. Currently the parameter `new_name` is mandatory and cannot be the current name of the catalog/schema. So to just change the comment, I currently need to update the catalog/schema twice: first with a new name, then back to the old name.

**Describe the solution you would like**

Either make the `new_name` parameter optional or allow it to be the current name of the catalog/schema.

**Additional context**

Example commands to illustrate the problem with the CLI:
```
bin/uc catalog create --name test_cat

bin/uc catalog update --name test_cat --comment foo  # fails and states that new_name is required
bin/uc catalog update --name test_cat --new_name test_cat --comment foo  # fails with error ALREADY_EXISTS

# to actually update the comment, we need to rename the catalog and then name it back
bin/uc catalog update --name test_cat --new_name test_cat2 --comment foo
bin/uc catalog update --name test_cat2 --new_name test_cat --comment foo
```

VillePuuska:
I took a stab at making `new_name` optional for updating a catalog in the CLI and server, see #158 I really don't know Java so I may very well have missed something, but this seems to work.

@tdas I didn't yet try to make it reject trying an "empty" update, e.g. `bin/uc catalog update --name unity`, since atm if you call `update catalog` without setting a comment it sets the comment to null. If this is intended, then maybe it's worth allowing an empty update like `bin/uc catalog update --name unity` so you can set a comment to null. If it's not intended, then I guess that's a separate bug to fix as well. What do you think?

vikrantpuppala:
Thanks for the change @VillePuuska! It looks good on a high level.

Regarding "empty" update: I don't think updating comment to `null` would be intended, because, there are other parameters in catalog and we wouldn't want to set all of them to null if we run `bin/uc catalog update --name unity`. I think at the CLI level we should handle this behaviour for the update command. I agree that it can be a separate bug but I'd like to see if it can be part of the same PR since the issues are similar and hopefully we can fix it in a way that it would be easily extendible as we add more parameters to `CatalogInfo`

VillePuuska:
[...]

PR #158, opened 2024-07-05
Change new_name to be optional when updating a catalog

VillePuuska:
**PR Checklist**

**Description of changes**

For the catalog part of issue #110 

Three main changes:
- Changed the parameter `new_name` to be optional in the server and CLI.
  - For the CLI, this is done by just setting the parameter to be optional instead of mandatory.
  - For the server, in the method `CatalogRepository.updateCatalog`, changed `newName` to be validated and the catalog name to be set to `newName` only if `newName` is not null.
- Changed `CatalogRepository.updateCatalog` to set the comment only if new comment is not null.
- Changed both the CLI and `CatalogRepository.updateCatalog` to not update anything if update catalog without setting any parameters to update.


docs/usage/cli.md
```diff
@@ -43,10 +43,10 @@ bin/uc catalog create --name my_catalog --comment "My First Catalog"
 
 ### Update a Catalog
 ```sh
-bin/uc catalog update --name <name> --new_name <new_name> [--comment <comment>]
+bin/uc catalog update --name <name> [--new_name <new_name>] [--comment <comment>]
 ```
 - `name` : The name of the existing catalog.
-- `new_name` : The new name of the catalog.
+- `new_name` : *[Optional]* The new name of the catalog.
 - `comment` : *[Optional]* The new description of the catalog.
```
vikrantpuppala, line 50: Could we please add a comment here mentioning that at least one optional parameter is needed?

examples/cli/src/main/java/io/unitycatalog/cli/CatalogCli.java
```diff
@@ -69,6 +70,9 @@ private static String getCatalog(CatalogsApi catalogsApi, JSONObject json) throw
     private static String updateCatalog(CatalogsApi apiClient, JSONObject json) throws JsonProcessingException, ApiException {
         String catalogName = json.getString(NAME_PARAM);
         json.remove(NAME_PARAM);
+        if (json.length() == 0) {
+            throw new CliException("Nothing to update.");
```
vikrantpuppala, line 74: Agree that this needs to be more descriptive. Something along the lines: No parameters to update, please provide one of: [[new_name, ..]](https://github.com/unitycatalog/unitycatalog/blob/main/examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java#L79) to update.

server/src/test/java/io/unitycatalog/server/base/catalog/BaseCatalogCRUDTest.java
```diff
@@ -53,10 +53,15 @@ public void testCatalogCRUD() throws ApiException {
...
-        // Update catalog
-        System.out.println("Testing update catalog..");
-        CatalogInfo updatedCatalogInfo = catalogOperations.updateCatalog(CATALOG_NAME, CATALOG_NEW_NAME, CATALOG_NEW_COMMENT);
-        assertCatalog(updatedCatalogInfo, CATALOG_NEW_NAME, CATALOG_NEW_COMMENT);
+        // Update catalog name
+        System.out.println("Testing update catalog: changing name..");
+        CatalogInfo updatedCatalogInfo = catalogOperations.updateCatalog(CATALOG_NAME, CATALOG_NEW_NAME, null);
+        assertCatalog(updatedCatalogInfo, CATALOG_NEW_NAME, COMMENT);
+
+        // Update catalog comment
```
vikrantpuppala, line 61: To make this more clear maybe: `// Update catalog comment without updating name`

server/src/test/java/io/unitycatalog/server/base/catalog/BaseCatalogCRUDTest.java
```diff
@@ -53,10 +53,15 @@ public void testCatalogCRUD() throws ApiException {
...
+        System.out.println("Testing update catalog: changing comment..");
+        CatalogInfo updatedCatalogInfo2 = catalogOperations.updateCatalog(CATALOG_NEW_NAME, null, CATALOG_NEW_COMMENT);
+        assertCatalog(updatedCatalogInfo2, CATALOG_NEW_NAME, CATALOG_NEW_COMMENT);
```
vikrantpuppala, line 64: Perhaps we can add another test case that an exception is thrown when nothing is provided to be updated?

vikrantpuppala (commented):
thanks for this change, looks good! just a few minor comments

server/src/test/java/io/unitycatalog/server/base/catalog/BaseCatalogCRUDTest.java
VillePuuska: I added a test for an exception being thrown now and changed [`CatalogRepository.updateCatalog`](https://github.com/unitycatalog/unitycatalog/blob/26657122db5e0363e4988ef9c126c541848eeb63/server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java#L102-L104) to also throw an exception if there is nothing to update.

The one thing I'm not quite happy with in my fix is that I had to change the exception that the CLI `updateCatalog` handler throws to be an `ApiException` instead of `CliException` [here](https://github.com/unitycatalog/unitycatalog/blob/26657122db5e0363e4988ef9c126c541848eeb63/examples/cli/src/main/java/io/unitycatalog/cli/CatalogCli.java#L82). This is because a `CliException` gets handled above so the test for an exception being thrown would not pass with a `CliException` 😅 This feels a bit wrong to me, but it works.

Actually, [all of this](https://github.com/unitycatalog/unitycatalog/blob/26657122db5e0363e4988ef9c126c541848eeb63/examples/cli/src/main/java/io/unitycatalog/cli/CatalogCli.java#L74-L82) is not necessarily needed now that the server itself checks if there is anything to update. I still left this in so that we can print a much more useful error message.

What do you think, is this fine or is there a cleaner way?

vikrantpuppala: Agree with everything you said. All looks good, but yeah a small nit would be the CliException, I think we should change the logic [here](https://github.com/unitycatalog/unitycatalog/blob/main/examples/cli/src/main/java/io/unitycatalog/cli/UnityCatalogCli.java#L94) and let CliException be handled the same as ApiException. Because as of now the CliException just does a print of the exception and that turns out to not be as descriptive.

vikrantpuppala:
Also, we would need to handle the same for schema, function, volume updates as well, but I would like to get this PR merged first and those can be follow up PRs

server/src/test/java/io/unitycatalog/server/base/catalog/BaseCatalogCRUDTest.java
VillePuuska: Handling a CliException the same as an ApiException sounds very good to me; added this change and changed the throwing from ApiException to CliException.

vikrantpuppala (approved):
lgtm, @haogang can we merge this please?

server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
```diff
@@ -115,7 +115,12 @@ public CatalogInfoDAO getCatalogDAO(Session session, String name) {
     }
 
     public CatalogInfo updateCatalog(String name, UpdateCatalog updateCatalog) {
-        ValidationUtils.validateSqlObjectName(updateCatalog.getNewName());
+        if (updateCatalog.getNewName() == null && updateCatalog.getComment() == null) {
```
haogang, line 118: Instead of throwing an error, the update should be a no-op with success if neither fields are specified. This will make adding new updatable fields easier later.

haogang (commented):
Thanks for addressing this. Just a minor comment before we can merge this PR.

server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
VillePuuska: Ok, done. I changed this to simply return the existing catalog info instead of throwing.

With this change, the CLI probably also should not error out. I reverted the previous change to how CliException is handled so the CLI now just prints the info message about update parameters without an error.

Also updated tests to test that an empty update does not change anything instead of testing that an exception is thrown.

haogang (approved):

PR #200, opened 2024-07-11
Set managed table's credential to table scan/write operator's options

cloud-fan:
**PR Checklist**

**Description of changes**

To support credential, the first step is to get credential of the managed table and put it into table scan/write operator's options, so that it will be part of the Hadoop conf that is used to do the actual file operations.

### DIFF
--- a/connectors/spark/etc/conf/server.properties
+++ b/connectors/spark/etc/conf/server.properties
@@ -1,0 +1,6 @@
+server.env=dev
+## temp credential config for s3 (Multiple s3 config can be added by incrementing the index)
+s3.bucketPath.0=s3://test-bucket
+s3.accessKey.0=test1
+s3.secretKey.0=test2
+s3.sessionToken.0=test3
--- a/connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SparkIntegrationTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SparkIntegrationTest.java
@@ -7,22 +7,35 @@ import io.unitycatalog.server.base.ServerConfig;
 import io.unitycatalog.server.base.catalog.CatalogOperations;
 import io.unitycatalog.server.base.schema.SchemaOperations;
 import io.unitycatalog.server.base.table.TableOperations;
+import io.unitycatalog.server.persist.dao.ColumnInfoDAO;
+import io.unitycatalog.server.persist.dao.TableInfoDAO;
+import io.unitycatalog.server.persist.utils.HibernateUtils;
 import io.unitycatalog.server.sdk.catalog.SdkCatalogOperations;
 import io.unitycatalog.server.sdk.schema.SdkSchemaOperations;
 import io.unitycatalog.server.sdk.tables.SdkTableOperations;
-import static io.unitycatalog.server.utils.TestUtils.*;
+import io.unitycatalog.server.utils.TestUtils;
+import org.apache.hadoop.fs.RawLocalFileSystem;
 import org.apache.spark.network.util.JavaUtils;
+import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.SparkSession;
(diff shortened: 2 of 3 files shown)


connectors/spark/etc/conf/server.properties
ravivj-db, line 1: Given this for tests, can we move it inside the test directory ?

ravivj-db (approved):

jaceklaskowski (approved):
LGTM

connectors/spark/etc/conf/server.properties
cloud-fan: due to the implementation of how we load property files in the UC server, I have to put it here. We should improve the implementation to make it more testable. We can follow some examples in Spark.

connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SparkIntegrationTest.java
```diff
@@ -116,6 +108,66 @@ public void testDeltaPathTable() throws IOException, ApiException {
         session.stop();
     }
 
+    @Test
+    public void testManagedParquetTable() throws ApiException, IOException, ParseException {
```
tdas, line 112: What does this have to do with "managed" table? This is only testing whether credentials vended by UC are percolated down to the relation.options. nothing more. It could be for managed tables or external tables. Table type does not matter. So the name of this test should be testCredentialUsage.

connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SparkIntegrationTest.java
```diff
@@ -116,6 +108,66 @@ public void testDeltaPathTable() throws IOException, ApiException {
...
+        SparkSession session = createSparkSessionWithCatalogs(SPARK_CATALOG);
+        // Spark only allow `spark_catalog` to return built-in file source tables.
+        setupManagedParquetTable(PARQUET_TABLE, new ArrayList<>(0), session);
+        String tableFullName = SPARK_CATALOG + "." + SCHEMA_NAME + "." + PARQUET_TABLE;
+
+        Dataset<Row> df = session.sql("SELECT * FROM " + tableFullName);
+        FileSourceScanExec scan = (FileSourceScanExec) df.queryExecution().executedPlan().collectLeaves().head();
+        Assert.assertEquals("test1", scan.relation().options().apply("fs.s3a.access.key"));
+        Assert.assertEquals("test2", scan.relation().options().apply("fs.s3a.secret.key"));
+        Assert.assertEquals("test3", scan.relation().options().apply("fs.s3a.session.token"));
```
tdas, line 123: This is not really testing whether all file system access are going to use these tokens. 
I suggest a different approach completely. 
- Make UC vend different credentials (fake ones) for different paths. 
- Create a fake FileSystem object that is used for S3 paths, but actually it's a wrapper for LocalFileSystem. With that you can intercept all of file system operations that Spark will do (including file listing, etc.)
- For each of those operations, throw error if operations on path prefix X is not configured with credentials for path X. 

If the spark read operation passes, then we will have confirmed that all file system operations have been correctly configured to use the right s3 credentials. 

Does that make sense?

cloud-fan: We probably need to update the UC sever to make it more testable. Right now the managed table location must starts with `s3://` so I can't really use a fake FileSystem to test it. Can we allow managed table location to be local file system for the ease of testing?

connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SparkIntegrationTest.java
cloud-fan: IIUC only managed table can have credentials? BTW I really want to test the real `CREATE TABLE` statement but UC doesn't support it today. This test has to use hacky way to create managed tables in the UC server.

haogang (approved):

PR #219, opened 2024-07-12
update schema: changing new_name to be optional and making empty updates change nothing

VillePuuska:
**PR Checklist**

**Description of changes**

Making `update schema` work similarly to how `update catalog` was changed to work in PR #158 

Fixes the schema part of issue #110 

Two main changes:
- Changed the parameter new_name to be optional in the server and CLI.
- Changed both the CLI and SchemaRepository.updateSchema to not update anything if no parameters are provided to update.
  - CLI just prints a help message listing all optional parameters.
  - SchemaRepository.updateSchema just returns existing SchemaInfo.


examples/cli/src/main/java/io/unitycatalog/cli/SchemaCli.java
```diff
@@ -65,11 +66,20 @@ private static String getSchema(SchemasApi schemasApi, JSONObject json) throws J
...
+        json.remove(CliParams.FULL_NAME.getServerParam());
+        if (json.length() == 0) {
+            List<CliParams> optionalParams = CliUtils.cliOptions.get(CliUtils.SCHEMA)
+                .get(CliUtils.UPDATE)
+                .getOptionalParams();
+            String errorMessage = "No parameters to update, please provide one of:";
+            for (CliParams param : optionalParams) {
+                errorMessage += "\n  --" + param.val();
+            }
+            throw new CliException(errorMessage);
```
vikrantpuppala, line 79: so, is the current model that the CLi throws an error on the client side but the server if still somehow receives such a request (from other clients) we will do a no-op?

VillePuuska: Pretty much except that the CLI does not throw the exception in `main`; I reverted that change we talked about in the catalog PR back to the original way that the CLI `main` handled `CliException`s [in this discussion](https://github.com/unitycatalog/unitycatalog/pull/158#discussion_r1670910208).

So in short:
- if the server gets an empty update, result is a no-op with success,
- if the CLI gets an empty update, the `updateXyz` handler does not call the server and instead throws a `CliException` with a help message, and the CLI `main` prints the message but result is still a success.

vikrantpuppala (approved):
Thank you for making this change! cc: @haogang

vikrantpuppala:
@haogang could you merge this please

haogang (approved):
lgtm, please resolve conflicts

VillePuuska:
> lgtm, please resolve conflicts

Done @haogang

VillePuuska:
Looks like the tutorial tests failing at the duckdb commands is unrelated to this PR as the test also failed after the latest commit.

haogang:
The test failure could be related to this PR: https://github.com/unitycatalog/unitycatalog/pull/222 cc. @vksx 

I'm going to merge this one. @vikrantpuppala can you please create a separate issue to fix the tutorial test?

vikrantpuppala:
@haogang this is fixed in #242

PR #223, opened 2024-07-12
set the licensecheck to check only compile dependencies

alexreid-db:
Maybe I'm mistaken, but think we don't have to worry about test dependency licenses? (those won't be shipped)

@Fokko thoughts?


Fokko (approved):
@alexreid-db Yes, that's correct. As far long as we don't ship the test binaries :)

alexreid-db:
@tdas @haogang can we get this merged in please?

haogang (approved):

PR #254, opened 2024-07-18
update volume: changing new_name to be optional and making empty updates change nothing

### LINKED ISSUE #110: Updating a catalog/schema without needing to rename it

VillePuuska:
**PR Checklist**

**Description of changes**

Making `update volume` work similarly to how `update catalog` was changed to work in PR #158 and `update schema`was changed to work in PR #219

Closes #110

Two main changes:
- Changed the parameter new_name to be optional in the server and CLI.
- Changed both the CLI and `VolumeRepository.updateVolume` to not update anything if no parameters are provided to update.
  - CLI just prints a help message listing all optional parameters.
  - `VolumeRepository.updateVolume` just returns existing `VolumeInfo`.

### DIFF
--- a/docs/usage/cli.md
+++ b/docs/usage/cli.md
@@ -251,8 +251,10 @@ bin/uc volume update --full_name <catalog>.<schema>.<volume> --new_name <new_nam
 - `catalog` : The name of the catalog.
 - `schema` : The name of the schema.
 - `volume` : The name of the volume.
-- `new_name` : The new name of the volume.
+- `new_name` : *[Optional]* The new name of the volume.
 - `comment` : *[Optional]* The new description of the volume.
+
+*Note:* at least one of the optional parameters must be specified.
 
 Example:
 ```sh
--- a/examples/cli/src/main/java/io/unitycatalog/cli/VolumeCli.java
+++ b/examples/cli/src/main/java/io/unitycatalog/cli/VolumeCli.java
@@ -227,8 +227,17 @@ public class VolumeCli {
       throws JsonProcessingException, ApiException {
     String volumeFullName = json.getString(CliParams.FULL_NAME.getServerParam());
     json.remove(CliParams.FULL_NAME.getServerParam());
-    UpdateVolumeRequestContent updateVolumeRequest;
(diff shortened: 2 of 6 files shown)


VillePuuska:
@vikrantpuppala Last PR for making updates work equivalently. Could you review this?

vikrantpuppala (approved):
Thank you very much @VillePuuska for seeing this through! (I remember you being hesitant early on because of not knowing Java :))
Sorry to be a buzz killer on closing the issue but I think we need to do the same for functions?
Again, thank you very much for everything you've contributed so far!

vikrantpuppala:
@haogang could we merge this please?

VillePuuska:
> Thank you very much @VillePuuska for seeing this through! (I remember you being hesitant early on because of not knowing Java :)) Sorry to be a buzz killer on closing the issue but I think we need to do the same for functions? Again, thank you very much for everything you've contributed so far!

Yeah, it actually turned out to be a good way to remember the language again after a decade of not using it 😅 

Looking at the [API](https://github.com/unitycatalog/unitycatalog/blob/main/api/Apis/FunctionsApi.md), functions don't even have an update method? Maybe there should be one?

vikrantpuppala:
Ah, makes sense! Then we're good to close the issue for now, we will make sure that this issue isn't present when update function is implemented.

haogang (approved):

PR #256, opened 2024-07-18
Migrate tests to JUnit 5

### LINKED ISSUE #202: Update Tests to use JUnit5 instead of JUnit4
JUnit5 has been out for a while now and JUnit4 hasn't been updated/supported for years. Most new extensions, etc. only support JUnit5. Don't think we are currently using any intricate junit features, so this upgrade should be pretty straight forward (mostly changing annotations).

filipovskid:
**PR Checklist**

**Description of changes**
Resolves #202.

* Migrate tests to JUnit 5 and remove dependencies to JUnit 4.
  * Remove server configuration parametrization, as discussed in the issue.
  * Use JUnit 5 annotations
  * Use JUnit 5 assertions
* Bump mocking dependencies their latest versions

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SparkIntegrationTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SparkIntegrationTest.java
@@ -26,7 +26,6 @@ import org.apache.spark.sql.execution.command.DataWritingCommandExec;
 import org.apache.spark.sql.execution.datasources.InsertIntoHadoopFsRelationCommand;
 import org.hibernate.Session;
 import org.hibernate.Transaction;
-import org.junit.Assert;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
@@ -36,6 +35,8 @@ import java.net.URI;
 import java.util.*;
 
 import static io.unitycatalog.server.utils.TestUtils.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class SparkIntegrationTest extends BaseCRUDTest {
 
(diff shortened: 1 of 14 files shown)


nastra (approved):
LGTM

connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SparkIntegrationTest.java
```diff
@@ -92,17 +93,17 @@ public void testDeltaPathTable() throws IOException, ApiException {
         String path1 = new File(dataDir, "test_delta_path1").getCanonicalPath();
         String tableName1 = String.format("delta.`%s`", path1);
         session.sql(String.format("CREATE TABLE %s(i INT) USING delta", tableName1));
-        Assert.assertTrue(
+        assertTrue(
```
nastra, line 96: stuff like this could be expressed in a more fluent way with AssertJ, such as `assertThat(session.sql("SELECT * FROM " + tableName1).collectAsList()).isEmpty()` instead of using `assertTrue()`. The big upside to this is that AssertJ will show the full content of the collection if the assertion ever fails, which is very helpful when debugging test failures.
A few tests have been already converted in https://github.com/unitycatalog/unitycatalog/commit/ca580fa648211166ad40de79971138e1927024e7. Would you be interested in converting existing assertions to AssertJ in a separate PR?

filipovskid: Makes a lot of sense. I will open an issue later today and proceed with these changes. Thank you for the feedback!

vikrantpuppala (approved):
lgtm, ~but a side question, I remember seeing a few PRs to migrate asserts to AssertJ, not sure if that applies here and could be done in this PR?~

Edit: just checked the comment by nastra

haogang (approved):

PR #211, opened 2024-07-11
Add mysql support

xx-db:
**PR Checklist**

- [ ] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**
add docker-compose.yml for mysql server
update jdbc config to connect to mysql
change docs to add instructions on how to start mysql

PR #265, opened 2024-07-22
Create a common interface to list entities

vikrantpuppala:
**PR Checklist**

**Description of changes**

Extended DAOs from a base identifiable entity to ensure we can have a common implementation for listing entities.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -8,7 +8,7 @@ import io.unitycatalog.server.persist.dao.SchemaInfoDAO;
 import io.unitycatalog.server.persist.dao.TableInfoDAO;
 import io.unitycatalog.server.persist.utils.FileUtils;
 import io.unitycatalog.server.persist.utils.HibernateUtils;
-import io.unitycatalog.server.persist.utils.PagedEntityLister;
+import io.unitycatalog.server.persist.utils.PagedListingHelper;
 import io.unitycatalog.server.persist.utils.RepositoryUtils;
 import io.unitycatalog.server.utils.Constants;
 import io.unitycatalog.server.utils.ValidationUtils;
@@ -26,7 +26,7 @@ public class TableRepository {
   private static final Logger LOGGER = LoggerFactory.getLogger(TableRepository.class);
   private static final SessionFactory SESSION_FACTORY = HibernateUtils.getSessionFactory();
   private static final SchemaRepository SCHEMA_REPOSITORY = SchemaRepository.getInstance();
-  private static final PagedEntityLister<TableInfoDAO> ENTITY_LISTER = new PagedEntityLister<>(TableInfoDAO.class);
+  private static final PagedListingHelper<TableInfoDAO> LISTING_HELPER = new PagedListingHelper<>(TableInfoDAO.class);
 
   private TableRepository() {}
 
@@ -247,8 +247,8 @@ public class TableRepository {
           Optional<String> pageToken,
           Boolean omitProperties,
           Boolean omitColumns) {
-    List<TableInfoDAO> tableInfoDAOList = ENTITY_LISTER.listEntity(session, maxResults, pageToken, schemaId);
-    String nextPageToken = ENTITY_LISTER.getNextPageToken(tableInfoDAOList, maxResults);
+    List<TableInfoDAO> tableInfoDAOList = LISTING_HELPER.listEntity(session, maxResults, pageToken, schemaId);
+    String nextPageToken = LISTING_HELPER.getNextPageToken(tableInfoDAOList, maxResults);
     List<TableInfo> result = new ArrayList<>();
     for (TableInfoDAO tableInfoDAO : tableInfoDAOList) {
       TableInfo tableInfo = tableInfoDAO.toTableInfo(!omitColumns);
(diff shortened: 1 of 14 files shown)


server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
```diff
@@ -86,14 +80,14 @@ public static TableInfoDAO from(TableInfo tableInfo) {
   public TableInfo toTableInfo(boolean fetchColumns) {
     TableInfo tableInfo =
         new TableInfo()
-            .name(name)
+            .tableId(super.getId().toString())
```
haogang, line 83: why do we call super, instead of getId/getName from the current instance? Subclasses might want to override the implementation. If not, we should make these methods `final`

server/src/main/java/io/unitycatalog/server/persist/dao/IdentifiableDAO.java
```diff
@@ -0,0 +1,27 @@
...
+
+@MappedSuperclass
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+@SuperBuilder
+public class IdentifiableDAO {
+    @Id
+    @Column(name = "id", columnDefinition = "BINARY(16)")
```
haogang, line 22: `columnDefinition` might be incompatible with mysql/postgre change in #211

server/src/main/java/io/unitycatalog/server/persist/utils/PagedEntityLister.java
```diff
@@ -0,0 +1,90 @@
...
+import jakarta.persistence.criteria.Root;
+import org.hibernate.Session;
+import org.hibernate.query.Query;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+
+public class PagedEntityLister<T extends IdentifiableDAO> {
```
haogang, line 18: shall we call it `PagedListingHelper`?

please also add javadoc for the class and methods

server/src/main/java/io/unitycatalog/server/persist/utils/PagedEntityLister.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public String getNextPageToken(List<T> entities, Optional<Integer> maxResults) {
+        if (entities == null || entities.isEmpty() || entities.size() < getPageSize(maxResults)) {
+            return null;
+        }
+        // return the last entity name
+        return entities.get(entities.size() - 1).getName();
+    }
+
+    public String getParentEntity() {
```
haogang, line 43: it's better to move this method to `IdentifiableDAO`, call it `getParentIdColumnName` and make it return an optional

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -24,9 +25,8 @@ public class TableRepository {
   private static final TableRepository INSTANCE = new TableRepository();
   private static final Logger LOGGER = LoggerFactory.getLogger(TableRepository.class);
   private static final SessionFactory SESSION_FACTORY = HibernateUtils.getSessionFactory();
-  private static final CatalogRepository CATALOG_REPOSITORY = CatalogRepository.getInstance();
   private static final SchemaRepository SCHEMA_REPOSITORY = SchemaRepository.getInstance();
-  public static final Integer DEFAULT_PAGE_SIZE = 100;
+  private static final PagedEntityLister<TableInfoDAO> ENTITY_LISTER = new PagedEntityLister<>(TableInfoDAO.class);
```
haogang, line 29: `LISTING_HELPER`?

server/src/main/java/io/unitycatalog/server/persist/dao/IdentifiableDAO.java
vikrantpuppala: updated from main

ravivj-db:
Just a general comment, please issue `build/sbt javafmtAll` and correct all formatting before submitting this PR (exclude the changes in client and server/model)

server/src/main/java/io/unitycatalog/server/persist/dao/IdentifiableDAO.java
```diff
@@ -0,0 +1,42 @@
...
+public class IdentifiableDAO {
+    @Id
+    @Column(name = "id")
+    private UUID id;
+
+    @Column(name = "name", nullable = false)
+    private String name;
+
+    public static <T extends IdentifiableDAO> Optional<String> getParentIdColumnName(Class<T> entityClass) {
+        if (TableInfoDAO.class == entityClass
```
haogang, line 32: We should just return `empty()` here and let each DAO class override the function if needed. A parent class usually should not refer to subclasses.

haogang (approved):
overall LGTM. Please address one more comment before merging

server/src/main/java/io/unitycatalog/server/persist/dao/IdentifiableDAO.java
vikrantpuppala: The issue with this is that we never actually have an object of the entityClass in `PagedListingHelper`. So the way I can do this is using reflection by creating a dummy object and then calling the overloaded function using that dummy object. let me know if that approach is fine.

haogang: Ah, didn't realize this is method is static. Probably that's the best we can do now. Later we might want to move this to an "EntityDescriptor" instance.

PR #266, opened 2024-07-22
allow configuration for server port

### LINKED ISSUE #250: Allow configuration of server port for Unity Catalog
**Description:**

Currently, the Unity Catalog server runs only on port 8080 and also some internal process uses port 8081. At least from the docs I could not find any way to configure these ports. This limits flexibility when deploying the server in environments where port 8080 or 8081 might be occupied by another application.

**Enhancement Request:**

I would like to request an enhancement that allows configuration of the server port. This would provide greater control and flexibility during deployment, especially in scenarios where port 8080 is unavailable.

**Proposed Solution:**

There are two possible approaches to achieve this:

- Configuration File: Introduce a configuration file (e.g., application.properties or server.properties) where the server port can be specified. This file would be loaded during server startup, allowing users to customize the port value.

- Command-Line Argument: Alternatively, consider adding a command-line argument (e.g., -p or --port) that accepts the desired port number. This would provide a more dynamic way to configure the port at runtime.

**Benefits:**

- Increased flexibility during deployment, allowing users to avoid conflicts with other applications using port 8080.
- Improved maintainability by separating configuration from code.
- Enhanced user experience by providing more control over server startup behavior.

vksx:
**PR Checklist**

**Description of changes**
Added checks for flags -p and --port to configure the port while starting the UC server. The default is 8080.

Usage:
```sh
bin/start-uc-server -p <port_number>
or
bin/start-uc-server --port <port_number>
```

~~The previous method of specifying the port ```bin/start-uc-server <port_number>``` will print the error message: "Invalid argument: <port_number>. Please use -p or --port to specify the port number."~~

Updated the code to utilize [Apache Commons CLI](https://commons.apache.org/proper/commons-cli/).

Fixes #250

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
+++ b/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
@@ -93,7 +93,19 @@ public class UnityCatalogServer {
   public static void main(String[] args) {
     int port = 8080;
     if (args.length > 0) {
-      port = Integer.parseInt(args[0]);
+      if ("-p".equals(args[0]) || "--port".equals(args[0])) {
+        if (args.length > 1) {
+          try {
+            port = Integer.parseInt(args[1]);
+          } catch (NumberFormatException e) {
+            System.err.println("Invalid port number: " + args[1]);
+            return;
+          }
+        } else {
+          System.err.println("No port number provided after " + args[0]);
+          return;
+        }
+      }
     }
     // Start Unity Catalog server
     UnityCatalogServer unityCatalogServer = new UnityCatalogServer(port + 1);
--- a/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
+++ b/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
@@ -26,6 +26,7 @@ import io.vertx.core.Vertx;
 import org.apache.logging.log4j.core.config.Configurator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.apache.commons.cli.*;
 
 public class UnityCatalogServer {
   private static final Logger LOGGER = LoggerFactory.getLogger(UnityCatalogServer.class);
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -93,7 +93,22 @@ private void addServices(ServerBuilder sb) {
...
+            return;
+          }
+        } else {
+          System.err.println("No port number provided after " + args[0]);
+          return;
+        }
+      } else {
+        System.err.println("Invalid argument: " + args[0] + ". Please use -p or --port to specify the port number.");
+        return;
+      }
```
vikrantpuppala, line 111: I think we can make this a bit more sophisticated by using a cli library.

vikrantpuppala (commented):
@filipovskid could you take up the review for this please?

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
vksx: Sure, we can use the same cli library as examples/cli - I will update this.

vksx: Sorry - just saw your message. I just pushed a commit with the commons-cli changes. Please let me know if this is fine and yeah meanwhile I'll check out the recommended one as well.

vksx: One question: Should we use a different library just for this part of the code? Because in examples/cli each file uses commons-cli. So, using commons-cli for consistency with this module would make sense. Thanks!

filipovskid: Let's keep it consistent and use commons-cli. If we transition to something else then we can do this here as well.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -92,8 +93,32 @@ private void addServices(ServerBuilder sb) {
...
+    // Add port option
+    options.addOption("p", "port", true, "Port number to run the server on. Default is 8080.");
+    CommandLineParser parser = new DefaultParser();
+    try {
+      // Parse the command line arguments
+      CommandLine cmd = parser.parse(options, args);
+
+      // Check if port option is provided
+      if (cmd.hasOption("p")) {
+        port = Integer.parseInt(cmd.getOptionValue("p"));
```
filipovskid, line 108: Let's avoid parsing this manually. We can define an [`Option.Builder#type(Class<?>)`](https://commons.apache.org/proper/commons-cli/apidocs/org/apache/commons/cli/Option.Builder.html#type(java.lang.Class)) and remove this parsing, along with the handling of `NumberFormatException`. Check [Converting (Parsing) Option Values](https://commons.apache.org/proper/commons-cli/usage.html) for more information.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -92,8 +93,32 @@ private void addServices(ServerBuilder sb) {
 
   public static void main(String[] args) {
     int port = 8080;
-    if (args.length > 0) {
-      port = Integer.parseInt(args[0]);
+    // Create Options object
```
filipovskid, line 96: Avoid comments on each line, the code speaks for itself :bowtie:. This applies for the comments below as well.

filipovskid (changes requested):
Minor remarks, we can stick to this approach. Thank you for your contribution!

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
vksx: Okay - I was following the same approach used [here]( https://github.com/unitycatalog/unitycatalog/blob/043dd3217fe5f6c5db36e9b4b240f3b8243ecb9c/examples/cli/src/main/java/io/unitycatalog/cli/UnityCatalogCli.java#L26). I will remove them.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
vksx: Thanks for pointing this out. Added the changes.

filipovskid (approved):
LGTM

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -92,8 +93,27 @@ private void addServices(ServerBuilder sb) {
...
+      CommandLine cmd = parser.parse(options, args);
+      if (cmd.hasOption("p")) {
+        port = cmd.getParsedOptionValue("p");
+      }
+    } catch (ParseException e) {
+      System.out.println();
+      System.out.println("Parsing Failed. Reason: " + e.getMessage());
+      System.out.println();
+      HelpFormatter formatter = new HelpFormatter();
+      formatter.printHelp("bin/start-uc-server", options);
```
vikrantpuppala, line 115: Can we please update the docs with these details? i.e. how to start the server on a specific port?

vksx: PTAL

vikrantpuppala (approved):
Thanks @vksx for making this change, @haogang to merge.

haogang (approved):

PR #270, opened 2024-07-23
Initial models API, messages and generated code

kriscon-db:
**PR Checklist**

- [x ] A description of the changes is added to the description of this PR.

**Description of changes**

Adding the initial API definition, messages, and generated code for models in UC OSS


api/all.yaml
```diff
@@ -722,6 +748,268 @@ paths:
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateModel'
+      responses:
+        '200':
+          description: The new model was successfully created.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ModelInfo'
```
kriscon-db, line 773: @haogang I see that other endpoints bypass creating dedicated response messages that only contain the info objects even tho DB UC has them.  Will that be a problem with compatability with databricks UC that *does* have things like CreateRegisteredModelResponse?

api/all.yaml
```diff
@@ -21,6 +21,14 @@ tags:
       Functions implement User-Defined Functions (UDFs) in Unity Catalog.
       The function implementation can be a SQL expression/query or a python function, and it can be invoked wherever a table reference is allowed in a query.
       In Unity Catalog, a function resides at the same level as a table, so it can be referenced with the form __catalog_name__.__schema_name__.__function_name__.
+  - name: Models
+    description: |
+      Models are machine learned models that are stored within Unity Catalog.  They are comprised of all versions of a single model that a user wishes to keep.
```
arpitjasa-db, line 26: "machine learned"? Do we want docs team to do a quick pass here as well?

api/all.yaml
```diff
@@ -21,6 +21,14 @@ tags:
...
+      A model resides in the third layer of Unity Catalog’s three-level namespace, and can be referenced with the form __catalog_name__.__schema_name__.__model_name__.
```
arpitjasa-db, line 27:
```suggestion
      A model resides in a schema of Unity Catalog’s three-level namespace, and can be referenced with the form __catalog_name__.__schema_name__.__model_name__.
```

api/all.yaml
```diff
@@ -21,6 +21,14 @@ tags:
...
+  - name: ModelVersions
+    description: |
+      Model versions contain all artifact files associated with an individual machine learned model.  They are associated with a single parent model and have an integer identifier, indicating the version number of the model.
+      Model versions are referenced via model URI's using the Unity Catalog three-level namespace of the registered model and the version identifier.
```
arpitjasa-db, line 31: Example of the URI format would help here

api/all.yaml
```diff
@@ -582,6 +590,24 @@ paths:
...
+      tags:
+        - TemporaryModelCredentials
+      summary: Generate temporary model credentials.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/GenerateTemporaryModelCredential'
+      responses:
+        '200':
```
arpitjasa-db, line 605: We don't need to worry about non-200 responses and errors in general?

api/all.yaml
```diff
@@ -722,6 +748,268 @@ paths:
           content:
             application/json:
               schema: {}
+  /models:
+    post:
+      operationId: createModel
+      summary: |
+        Create a model.
+        WARNING: This API is experimental and will change in future versions.
```
arpitjasa-db, line 756: It's still experimental? What planned changes do we have?

api/all.yaml
```diff
@@ -722,6 +748,268 @@ paths:
...
+            - when set to 0, the page length is set to a server configured value;
+            - when set to a value less than 0, an invalid parameter error is returned;
+          required: false
+          schema:
+            type: integer
+            format: int32
+            maximum: 50
+        - name: page_token
+          in: query
+          description: Opaque token to send for the next page of results (pagination).
```
arpitjasa-db, line 804: Pagination is supported in UC OSS?

api/all.yaml
```diff
@@ -722,6 +748,268 @@ paths:
...
+      summary: Delete a model version
+      description: |
+        Deletes a model version from the specified model.
+      responses:
+        '200':
+          description: The model was successfully deleted.
+          content:
+            application/json:
+              schema: {}
+  /models/{full_name}/versions/{version}/finalize:
```
arpitjasa-db, line 980: We need the finalize API in UC OSS as well? Is this mainly to maintain parity since I imagine most folks won't need this API in their custom implementations?

api/all.yaml
```diff
@@ -1414,6 +1702,39 @@ components:
           description: |
             Opaque token to retrieve the next page of results. Absent if there are no more pages.
             __page_token__ should be set to this value for the next request (for the next page of results).
+    GenerateTemporaryModelCredential:
+      example:
+        table_id: model_id
+        operation: null
+      properties:
+        table_id:
+          description: |
```
arpitjasa-db, line 1711: Would probably be helpful to have more details about when to generate these credentials, especially in the context of OSS

api/all.yaml
kriscon-db: The api is only at v0.1  This addition will roll out in v0.2 but we should keep it marked as experimental until then.

api/all.yaml
kriscon-db: Yes.

api/all.yaml
kriscon-db: Yes.  How else will you tell when the artifacts have been fully uploaded to the storage location?

api/all.yaml
kriscon-db: Probably too early for any of that.  We can fix the documentation up later.  I don't want to get too hung up or married to any documentation right now.

api/all.yaml
kriscon-db: Followed the example above for tables.

api/all.yaml
kriscon-db: No other endpoints have them defined yet.  That will probably come later.

api/all.yaml
arpitjasa-db: Yup, makes sense. Not sure how public-facing these docs will be, but I can imagine some users being confused at that, so adding that bit about `artifacts have been fully uploaded to the storage location` for this API will probably help them implement it accordingly

api/all.yaml
```diff
@@ -582,6 +590,24 @@ paths:
           content:
             application/json:
               schema: {}
+  /temporary-model-credentials:
+    post:
+      operationId: generateTemporaryModelCredentials
+      tags:
+        - TemporaryModelCredentials
+      summary: Generate temporary model credentials.
```
vladimirk-db, line 598: Should we expand this? E.g., write out why it's needed.

api/all.yaml
vladimirk-db: Curious, why is `generateTemporaryModelCredentials` above not marked experimental?

api/all.yaml
```diff
@@ -722,6 +748,268 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - name: max_results
+          in: query
+          description: |
+            Maximum number of models to return.
+            - when set to a value greater than 0, the page length is the minimum of this value and a server configured value;
+            - when set to 0, the page length is set to a server configured value;
+            - when set to a value less than 0, an invalid parameter error is returned;
```
vladimirk-db, line 796: Does the server configured value apply to UC OSS as well (or it's a databricks-only thing)?

api/all.yaml
```diff
@@ -1414,6 +1702,39 @@ components:
...
+          type: string
+        operation:
+          $ref: '#/components/schemas/ModelOperation'
+      type: object
+      required:
+        - model_id
+        - operation
+    GenerateTemporaryModelCredentialResponse:
+      properties:
+        aws_temp_credentials:
```
vladimirk-db, line 1723: Basic question, is Azure not in the picture here? Why only aws_temp_credentials?

api/all.yaml
```diff
@@ -1490,6 +1811,202 @@ components:
...
+          description: Name of model, relative to parent schema.
+          type: string
+        catalog_name:
+          description: Name of parent catalog.
+          type: string
+        schema_name:
+          description: Name of parent schema relative to its parent catalog.
+          type: string
+        storage_location:
+          description: Storage root URL for model (for **MANAGED**, **EXTERNAL** tables)
```
vladimirk-db, line 1827: What does this mean? It seems to be relevant to UC tables, not models?

api/all.yaml
```diff
@@ -1490,6 +1811,202 @@ components:
...
+          description: Name of model, relative to parent schema.
+          type: string
+        catalog_name:
+          description: Name of parent catalog.
+          type: string
+        schema_name:
+          description: Name of parent schema relative to its parent catalog.
+          type: string
+        version:
+          description: The version number of the model version
```
vladimirk-db, line 1856:
```suggestion
          description: The version number of the model version.
```

api/all.yaml
```diff
@@ -1490,6 +1811,202 @@ components:
...
+          format: int64
+        updated_at:
+          description: Time at which this model version was last modified, in epoch milliseconds.
+          type: integer
+          format: int64
+        model_version_id:
+          description: Unique identifier for the model version.
+          type: string
+    ModelVersionStatus:
+      description: All possible model version statuses
```
vladimirk-db, line 1885: Should we describe the statuses?

api/all.yaml
```diff
@@ -1490,6 +1811,202 @@ components:
...
+          description: Name of model, relative to parent schema.
+          type: string
+        catalog_name:
+          description: Name of parent catalog.
+          type: string
+        schema_name:
+          description: Name of parent schema relative to its parent catalog.
+          type: string
+        storage_location:
+          description: Storage root URL for table (for **MANAGED**, **EXTERNAL** tables)
```
vladimirk-db, line 1905: Same comment as above, why refer to tables here?

vladimirk-db (approved):
Did a pass, mainly questions for my understanding. Thanks

api/all.yaml
```diff
@@ -722,6 +748,268 @@ paths:
           content:
             application/json:
               schema: {}
+  /models:
+    post:
+      operationId: createModel
```
smurching, line 753: Should this be CreateRegisteredModel to match the name of the UC REST API + clarify that it's a registered model and not a model version? And similar comment for "summary" etc fields

api/all.yaml
```diff
@@ -1414,6 +1702,39 @@ components:
           description: |
             Opaque token to retrieve the next page of results. Absent if there are no more pages.
             __page_token__ should be set to this value for the next request (for the next page of results).
+    GenerateTemporaryModelCredential:
```
smurching, line 1705:
```suggestion
    GenerateTemporaryModelVersionCredential:
```

api/all.yaml
```diff
@@ -1414,6 +1702,39 @@ components:
...
+      example:
+        model_id: model_id
+        operation: null
+      properties:
+        model_id:
+          description: |
+            Model id for which temporary credentials are generated. 
+            Can be obtained from models/{full_name} (get model info) API.
```
smurching, line 1713: Don't think this is right

smurching: Did we copy paste from the tables API? We should also cross-reference with our own openapi spec for models in UC REST APIs to make sure we have these details correct

smurching (commented):
Leaving a few comments - @arpitjasa-db @vladimirk-db @kriscon-db could you help audit + make sure the API is consistent with our OpenAPI specs + API signatures for models in UC? Once it LGTY I can take another quick pass

api/all.yaml
kriscon-db: So this part was not clear that the operationId was anything more than the UC OSS implementation of the handler for this endpoint.  I can change it but it wasn't obvious that we wanted to keep the same nomenclature of `RegisteredModel` vs `Model`

api/all.yaml
kriscon-db: Should be model...fixing

api/all.yaml
kriscon-db: Copy paste error.  Should read models

api/all.yaml
kriscon-db: They are working on adding the other credentials.  In DB UC, this is basically a TemporaryCredentials message where the credentials field looks like:

```  oneof credentials {
    AwsCredentials aws_temp_credentials = 2;
    AzureUserDelegationSAS azure_user_delegation_sas = 3;
    GcpOauthToken gcp_oauth_token = 4;
    ForeignCredential foreign_credential = 5;
    AzureActiveDirectoryToken azure_aad = 6;
    R2Credentials r2_temp_credentials = 8;
  }```

File: api/all.yaml (hunk 11 above)
Comment (kriscon-db) [reply api/all.yaml:796]: This was copied from another paginated api.  We can update this when the other pagination descriptions are updated.

File: api/all.yaml (hunk 17 above)
Comment (kriscon-db) [reply api/all.yaml:753]: I'm going to go ahead and rename.

File: clients/java/src/main/java/io/unitycatalog/client/api/ModelApi.java [hunk 20]
```diff
@@ -0,0 +1,156 @@
+/*
```
smurching, line 1: Once this is regenerated, will it be renamed to RegisteredModelApi? One side-effect of making sure we're consistent with the Databricks OpenAPI spec is that these names should mostly become consistent

api/all.yaml
```diff
@@ -1414,6 +1702,60 @@ components:
...
+            Model name for which temporary credentials are generated. 
+            Can be obtained from models/{full_name} (get model info) API.
+          type: string
+        version:
+          description: |
+            Model version for which temporary credentials are generated.
+          type: integer
+          format: int64
+        operation:
+          $ref: '#/components/schemas/ModelOperation'
```
smurching, line 1734: I think this is called `ModelVersionOperation` in Databricks

api/all.yaml
```diff
@@ -1414,6 +1702,60 @@ components:
...
+          description: |
+            Server time when the credential will expire, in epoch milliseconds.
+            The API client is advised to cache the credential given this expiration time.
+          type: integer
+          format: int64
+      type: object
+    ModelOperation:
+      enum:
+        - UNKNOWN_MODEL_OPERATION
+        - READ
```
smurching, line 1756: Pretty sure these have different names/would collide with tables with the current name

clients/java/src/main/java/io/unitycatalog/client/api/ModelApi.java
kriscon-db: I needed to rename the tag at the top.  Fixing.

api/all.yaml
kriscon-db: So it is.  Fixing

api/all.yaml
kriscon-db: Yep....though...do we store the string value of these enums or the integer value?  regardless....fixing.

api/all.yaml
haogang: v0.1 cut was made a week ago. This change will go into v0.2

api/all.yaml
```diff
@@ -1490,6 +1832,206 @@ components:
...
+          description: Name of model, relative to parent schema.
+          type: string
+        catalog_name:
+          description: Name of parent catalog.
+          type: string
+        schema_name:
+          description: Name of parent schema relative to its parent catalog.
+          type: string
+        storage_location:
+          description: Storage root URL for registered model (for **MANAGED**, **EXTERNAL** regisetered models)
```
haogang, line 1930: Do we have managed/external concept for model version in UC OSS? What's the difference between them?

api/all.yaml
haogang: We should define a common message for credential in UC OSS, and share it cross all genreateTempCred APIs. cc. @alexreid-db

api/all.yaml
kriscon-db: We would have to.  Model versions are comprised of artifact files that make up a model version.  Those files need to be stored in a location (which is why we need a scoped token via the credential API) that we can read/write those models to/from.

api/all.yaml
kriscon-db: In looking at the DB UC, it looks like those protos were defined before the proto linter was added that required DB protos to break up responses from the request.  So currently, we have a CreateRegisteredModel message that acts as the request and there is a Response message defined inside of it (CreateRegisteredModel.Response) which only has 1 value defined, the RegisteredModelInfo object thats being returned (with that being marked as json_inline).  

The DB handler for this returns a CreateRegisteredModel.Response object and anything that uses that response references the lone value as resp.getRegisteredModelInfo().  I guess my question ultimately is, do we need to keep that definition here where I would basically define a CreateRegisteredModelResponse definition (the same way I did for ListRegisteredModels) that would only have a single info object contained in it (and i'd do that for every API).

api/all.yaml
```diff
@@ -722,6 +748,268 @@ paths:
...
+            schema:
+              $ref: '#/components/schemas/UpdateRegisteredModel'
+      responses:
+        '200':
+          description: The model was successfully updated.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RegisteredModelInfo'
+    delete:
```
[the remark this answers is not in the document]
kriscon-db: Aliases are not in scope for v0.2

api/all.yaml
```diff
@@ -722,6 +748,268 @@ paths:
...
+      description: |
+        Gets a model for a specific catalog and schema.
+      responses:
+        '200':
+          description: The model was successfully retrieved.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RegisteredModelInfo'
+    patch:
```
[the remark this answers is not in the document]
kriscon-db: ACLs are still being worked on.  I will need to add them into the handlers once I move onto coding them (and when the ACL handling code is ready).

api/all.yaml
kriscon-db: It's public undocumented currently.  Its on my list of things to track down how to handle because the credential endpoint is also the same way.  We will probably need to document it in DB UC if we are adding it into UC OSS.  I'll sort that out with @haogang and @tdas

api/all.yaml
```diff
@@ -1490,6 +1832,206 @@ components:
...
+          description: An array of model information objects.
+          type: array
+          items:
+            "$ref": "#/components/schemas/RegisteredModelInfo"
+        next_page_token:
+          type: string
+          description: |
+            Opaque token to retrieve the next page of results. Absent if there are no more pages.
+            __page_token__ should be set to this value for the next request (for the next page of results).
+    CreateModelVersion:
```
vladimirk-db, line 1961: Curious why there is no CreateModelVersion in here:
https://docs.databricks.com/api/workspace/modelversions?

api/all.yaml
```diff
@@ -1490,6 +1832,206 @@ components:
...
+      properties:
+        model_name:
+          description: Name of model, relative to parent schema.
+          type: string
+        catalog_name:
+          description: Name of parent catalog.
+          type: string
+        schema_name:
+          description: Name of parent schema relative to its parent catalog.
+          type: string
```
vladimirk-db, line 1972: nit: shouldn't the ordering be model-> schema -> catalog?

api/all.yaml
```diff
@@ -1490,6 +1832,206 @@ components:
...
+        source:
+          description: URI indicating the location of the source model artifacts.
+          type: string
+        run_id:
+          description: The run id used by the ML package that generated the model.
+          type: string
+        storage_location:
+          description: Storage root URL for model (for **MANAGED**, **EXTERNAL** models)
```
vladimirk-db, line 1980: Nit: what are managed, external models? I think this should say:
"The storage location on the cloud under which model version data files are stored." like the Databricks API.

api/all.yaml
```diff
@@ -1490,6 +1832,206 @@ components:
...
+          type: string
+        comment:
+          description: User-provided free-form text description.
+          type: string
+      required:
+        - model_name
+        - catalog_name
+        - schema_name
```
vladimirk-db, line 1988: nit: reorder, like above.

api/all.yaml
```diff
@@ -1490,6 +1832,206 @@ components:
...
+    UpdateModelVersion:
+      type: object
+      properties:
+        source:
+          description: URI indicating the location of the source model artifacts.
+          type: string
+        run_id:
+          description: The run id used by the ML package that generated the model.
+          type: string
```
vladimirk-db, line 1997: We don't allow updating source or run_id in the databricks hosted models in UC?

api/all.yaml
vladimirk-db, line 1930: typo: "regisetered"

api/all.yaml
```diff
@@ -1490,6 +1832,206 @@ components:
         - READ_VOLUME
         - WRITE_VOLUME
       type: string
+    RegisteredModelInfo:
+      type: object
+      properties:
+        name:
```
vladimirk-db, line 1838: Databricks hosted API has a `name` and a `full_name`, FWIW.

vladimirk-db: (Seems redundant.)

vladimirk-db (approved):
Overall LGTM, left a few more comments.

api/all.yaml
alexreid-db: @haogang I'd agree, but this would deviate from DB UC apis which for some reason defines separate requests/responses/components even though they technically don't differ in their definitions.

api/all.yaml
kriscon-db: I think we didn't want to expose model version creation given that the logic lives in the MLflow client for uploading artifacts and finalizing model version statuses.  With UC OSS, we now need to expose these.

api/all.yaml
kriscon-db: Confirmed that we do not so I'll remove these.  We only allow comment updates.

api/all.yaml
haogang: Why? DB UC APIs has `TemporaryCredentials` message that are shared by all the response messages.

api/all.yaml
kriscon-db: The existing documentation in databricks lists them in this order.  I think we should keep that ordering.
https://docs.databricks.com/api/workspace/registeredmodels/create#name

And the documentation's naming is incorrect for model version.....its not name, its model_name.  I'll fix when I update the databricks documentation to include the missing APIs.

api/all.yaml
kriscon-db: Same comment as above.  We list them in this order in the documentation and protos.

api/all.yaml
kriscon-db: Adding full_name.  I think this is early tech debt that was never cleaned up (why both are there).

api/all.yaml
kriscon-db: Removed the comment for managed/external because there is no notion of it.  This was a copy paste error.

api/all.yaml
rickyma: There are two spaces here, you can delete one space:
`model registry.  They are comprised`
->
`model registry. They are comprised`

Same for:
`machine learned model.  They are associated`
->
`machine learned model. They are associated`

api/all.yaml
```diff
@@ -582,6 +590,24 @@ paths:
           content:
             application/json:
               schema: {}
+  /temporary-model-version-credentials:
+    post:
+      operationId: generateTemporaryModelVersionCredential
+      tags:
+        - TemporaryModelVersionCredential
+      summary: Generate temporary model version credentials.  These credentials are used by clients to write and retrieve model artifacts from the model versions external storage location.
```
rickyma, line 598: ditto

haogang (approved):

PR #169, opened 2024-07-08
Remove generated code from repo and generate it compile time

vikrantpuppala:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [x] If there is a related issue, make sure it is linked to this PR.
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

Remove generated code from the repo and rather generate it compile time

PR #273, opened 2024-07-23
add library dependencies for slf4j in build.sbt

### LINKED ISSUE #269: SLF4J Warnings showing up when starting the UC Server in CLI
**Describe the bug**

Getting the following SLF4J warnings when starting the UC Server:
- [No SLF4J providers were found](https://www.slf4j.org/codes.html#noProviders).
- [Class path contains SLF4J bindings targeting slf4j-api versions 1.7.x or earlier](https://www.slf4j.org/codes.html#ignoredBindings).

**To Reproduce**

1. `build/sbt package`
2. `bin/start-uc-server`

**Expected behavior**

We should get rid of these warnings.

-   OS: macOS Sonoma 14.5

**Additional context**

As mentioned [here](https://logging.apache.org/log4j/2.x/log4j-slf4j-impl.html#:~:text=log4j%2Dslf4j2%2Dimpl%20should%20be%20used%20with%20SLF4J%202.0.x%20releases%20or%20newer.) we can get rid of the warnings if we replace `log4j-slf4j-impl` with `log4j-slf4j2-impl` in [`build.sbt`](https://github.com/unitycatalog/unitycatalog/blob/main/build.sbt) [here](https://github.com/unitycatalog/unitycatalog/blob/79a2ad9a9ad842f4fb82dcca3e624928a666992c/build.sbt#L178) and [here](https://github.com/unitycatalog/unitycatalog/blob/79a2ad9a9ad842f4fb82dcca3e624928a666992c/build.sbt#L275). 

Please let me know if this makes sense. Thanks!

vksx:
**PR Checklist**

**Description of changes**

Added [recommended](https://github.com/unitycatalog/unitycatalog/issues/269#issuecomment-2245915657) library dependencies for slf4j in `commonSettings` (build.sbt).

Fixes #269


build.sbt
```diff
@@ -69,6 +69,12 @@ lazy val commonSettings = Seq(
     // I think we're good with the classpath exception in there.
     case DepModuleInfo("jakarta.transaction", "jakarta.transaction-api", _) => true
   },
+  libraryDependencies ++= Seq(
+    "org.slf4j" % "slf4j-api" % "2.0.13",
+    "org.slf4j" % "slf4j-log4j12" % "2.0.13" % Test,
+    "org.apache.logging.log4j" % "log4j-slf4j2-impl" % "2.23.1",
+    "org.apache.logging.log4j" % "log4j-api" % "2.23.1",
+  ),
```
vikrantpuppala, line 77: Can we remove the log4j dependencies from [line 182](https://github.com/unitycatalog/unitycatalog/pull/273/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R182) and other places?

vikrantpuppala: Also maybe use the log4jVersion variable?

vikrantpuppala (commented):
Thank you for fixing this!

A minor comment.

Also, have we also verified that this setup respects the /etc/conf/*.log4j.properties?

vksx:
> Also, have we also verified that this setup respects the /etc/conf/*.log4j.properties?

I haven't received any further warnings or errors, during the build after making these changes. @jaceklaskowski Would you like to add something here?

build.sbt
vksx: PTAL

jaceklaskowski:
When I looked into it I came up with the list of deps, but didn't go as far as to check out whether the logging conf files are used. Let me know @vksx if you need more help here. I didn't mean to "steal" this task from you 😉

build.sbt
vikrantpuppala: I had added this change at line 31 so I think we can close this PR

vikrantpuppala:
Added this change in #169

vikrantpuppala:
Thank you Vikas and Jacek for recommending this change, full credits to you guys for solving this issue.

Issue #230, opened 2024-07-15
Support Iceberg REST catalog- Credential Vending


monamohan:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you would like**

**Additional context**

Buktoria:
Hey @monamohan I can take this

Issue #47, opened 2024-06-15
Add Support for Azure Data Lake Storage for cloud storage


zcking:
**Is your feature request related to a problem? Please describe.**

The `server.properties` and `TemporaryTableCredentialsService` only support local files and S3 for storage currently. It would be nice to bring support for ADLS Gen2 (and Google Cloud Storage probably) to accommodate the majority of cloud users and have parity with the Databricks Unity Catalog (pre-open-source). 

**Describe the solution you would like**

In `server.properties` I think we should avoid assumptions or ambiguity by not using the `s3.` prefix. Ideally I would like to see the structure mirror Databricks' External Locations more closely. Something like this is an idea:

```python
server.env=dev

# External Locations:
# 
# S3:
externalLocation.bucketPath.0=s3://unitycatalog-220412-us-east-1/
externalLocation.accessKey.0=XXXX
externalLocation.secretKey.0=XXXX
externalLocation.sessionToken.0=XXXX

# Local File System:
externalLocation.bucketPath.1=file:///data/unity-catalog/
externalLocation.accessKey.1=
externalLocation.secretKey.1=
externalLocation.sessionToken.1=

# ADLS Gen2
externalLocation.bucketPath.2=abfs://data@unitycatalog220412.dfs.core.windows.net/catalog
externalLocation.accessKey.2=XXXX
externalLocation.secretKey.2=XXXX
externalLocation.sessionToken.2=XXXX

# GCS
externalLocation.bucketPath.3=gs://unitycatalog-220412/
externalLocation.accessKey.3=XXXX
externalLocation.secretKey.3=XXXX
externalLocation.sessionToken.3=XXXX
``` 

**Additional context**

Slightly related issue to this feature request is support for the endpoint URL. That would help support more object storages (e.g. MinIO or others that are compatible with the common S3 API): https://github.com/unitycatalog/unitycatalog/issues/43

zcking:
We may want to do the same naming cleanup within the response bodies for the temporary credential APIs. 

Instead of:
```json
{
  "aws_temp_credentials": {
    "access_key_id": "string",
    "secret_access_key": "string",
    "session_token": "string"
  },
  "expiration_time": 0
}
```

Maybe just getting rid of the `aws_` prefix?
```

Datt1234:
> server.env=dev

# External Locations:
# 
# S3:
externalLocation.bucketPath.0=s3://unitycatalog-220412-us-east-1/
[...]

PR #274, opened 2024-07-24
implementation for basic credential vending

alexreid-db:
**PR Checklist**

**Description of changes**

This PR adds the ability to vend temporary credentials for S3, Azure, and Google Cloud AND is able to downscope the temporary credentials to access only certain paths and certain privileges (read vs. write, etc.). This also updates the Iceberg REST load table implementation to vend a temporary credential and supply it in the table config so client can read table without needing to have creds set on client.

AWS/S3 setup:

Bucket + Role + Creds (downscoping)
- Create S3 bucket
- Create an AWS IAM role that has access to s3 bucket
- Create user and retrieve access keys
- Give user permission to assume role
- Update IAM role trust policy to trust newly created user

Set the following properties in server.properties file

```
s3.bucketPath.X=s3://bucket-where-tables-live
s3.region.X=aws-region-of-bucket (e.g. us-east-1)
s3.awsRoleArn.X=IamRole (role to assume for access to bucket)
# Optional (If blank, it will use DefaultCredentialsProviderChain)
s3.accessKey.X=<access-key>
s3.secretKey.X=<secret-key>
```
NOTE: If you don't set keys in the server.properties, you must configure something on [DefaultCredentialsProviderChain](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html) that the sts client will use to assume the provided role.

AWS Session Creds (no downscoping)
NOTE: this should be used for test purposes only
- Create temporary session and supply access key, secret key, and session token

Azure Setup:
- Create Storage Account and container
- Create service principal 
- Update role permissions on storage account and give service principal access (e.g. Storage Blob Data Reader, Storage Blob Data Contributor, Admin, etc.)
- Retrieve creds for service principal

Set the following properties in server.properties file

```
adls.storageAccountName.X=
adls.tenantId.X=
adls.clientId.X=
adls.clientSecret.X=
```

Google Cloud Setup:
- Create GCS Bucket
- Create Service Account
- Grant access to new service account to new bucket (e.g. assign Storage Object Admin role to service account)
- Create and retrieve keys for Service Account
-- This will be a json key file, download it and put it somewhere accessible by the UC server

```
gcs.bucketPath.X=gs://bucket-name
gcs.jsonKeyFilePath.X=</path/to/json/key/file>
```

This was tested by using OSS Spark to read iceberg/uniform tables created in all 3 cloud providers.

Follow PR to come shortly that provides full documentation on setup, etc.
Better integration tests to follow as well.
Storage configurations should at some point become entities and move into the db.

Addresses Issues:
https://github.com/unitycatalog/unitycatalog/issues/230
https://github.com/unitycatalog/unitycatalog/issues/47
https://github.com/unitycatalog/unitycatalog/issues/105


clients/java/src/main/java/io/unitycatalog/client/model/AzureUserDelegationSAS.java
```diff
@@ -0,0 +1,150 @@
+/*
```
creechy, line 1: IDK if you've merged main recently, I _think_ the generated code is out of git now?

api/all.yaml
```diff
@@ -1460,6 +1464,18 @@ components:
...
+    AzureUserDelegationSAS:
+      properties:
+        sas_token:
+          description: Azure SAS Token
+          type: string
+      type: object
+    GcpOauthToken:
+      properties:
+        oauth_token:
+          description: Gcp Token
```
creechy, line 1476: nit - GCP all caps here?

server/src/main/java/io/unitycatalog/server/service/TemporaryTableCredentialsService.java
```diff
@@ -3,41 +3,29 @@
...
+  }
+
   @Post("")
   public HttpResponse generateTemporaryTableCredential(
       GenerateTemporaryTableCredential generateTemporaryTableCredential) {
     String tableId = generateTemporaryTableCredential.getTableId();
-
-    // Check if table exists
-    String tableStorageLocation = "";
     TableInfo tableInfo = TABLE_REPOSITORY.getTableById(tableId);
```
creechy, line 27: I take it, this is gonna throw if the table doesn't exist? or do you need to check that `tableInfo` is not null before proceeding?

server/src/main/java/io/unitycatalog/server/service/TemporaryVolumeCredentialsService.java
```diff
@@ -25,17 +30,7 @@ public HttpResponse generateTemporaryTableCredential(
       throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Volume ID is required.");
     }
     VolumeInfo volumeInfo = VOLUME_REPOSITORY.getVolumeById(volumeId);
```
creechy, line 32: Same question, does this throw if not exist?

creechy: nm I see the check in `vendCredentialForVolume()`

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
```diff
@@ -0,0 +1,28 @@
...
+import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
+
+import java.util.Map;
+
+public class AwsCredentialVendor {
+
+  private final Map<String, S3StorageConfig> s3Configurations;
+
+  public AwsCredentialVendor() {
+    this.s3Configurations = ServerPropertiesUtils.getInstance().getS3Configurations();
```
creechy, line 16: nit: Do we want to support dynamically adding configurations to `servers.properties` w/o restarting? If so should just do this in `vendAwsCredentials()`

server/src/main/java/io/unitycatalog/server/service/credential/azure/AzureCredential.java
```diff
@@ -0,0 +1,5 @@
+package io.unitycatalog.server.service.credential.azure;
+
+public record AzureCredential(String sasToken, long expirationTimeInEpochMillis) {
```
creechy, line 3: Am I missing something here or is this intentionally empty?

api/all.yaml
alexreid-db: this was just copied from the UC api 🤷

clients/java/src/main/java/io/unitycatalog/client/model/AzureUserDelegationSAS.java
alexreid-db: yeah, I need to pull and rebase

server/src/main/java/io/unitycatalog/server/service/TemporaryTableCredentialsService.java
alexreid-db: yeah the repository is currently taking care of throwing if it doesn't exist...

server/src/main/java/io/unitycatalog/server/service/TemporaryVolumeCredentialsService.java
alexreid-db: same as above

creechy (approved):
It all makes sense to me, although I don't have full grasp of vended credentials on Azure an GCP. Just had a few nits and questions.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
alexreid-db: yeah, maybe can do this in a follow-up PR, but also wonder if should just hold off on this since the configs _should_ be moving to database/entities/etc.

server/src/main/java/io/unitycatalog/server/persist/utils/ServerPropertiesUtils.java
```diff
@@ -28,20 +29,6 @@ private void loadProperties() {
     try (InputStream input = Files.newInputStream(Paths.get("etc/conf/server.properties"))) {
       properties.load(input);
       LOGGER.debug("Properties loaded successfully");
-      int i = 0;
-      while (true) {
```
creechy, line 32: nit: personal preference but i like `for (int i = 0;; i++) ` loops when `break`-ing to exit.

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
```diff
@@ -0,0 +1,108 @@
...
+    CredentialContext credentialContext = CredentialContext.builder().storageScheme(storageLocationUri.getScheme()).storageBase(storageLocationUri.getPath())
+      .privileges(Set.of(CredentialContext.Privilege.SELECT, CredentialContext.Privilege.UPDATE)).locations(List.of(volume.getStorageLocation())).build();
+
+    return vendCredential(credentialContext).toVolumeCredentialResponse();
+  }
+
+  public CredentialResponse vendCredential(CredentialContext context) {
+    CredentialResponse.CredentialResponseBuilder builder = new CredentialResponse.CredentialResponseBuilder();
+
+    switch (context.storageScheme) {
```
creechy, line 73: nit: direct use of field instead of getter/accessor.

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
```diff
@@ -0,0 +1,108 @@
...
+      case URI_SCHEME_ABFS, URI_SCHEME_ABFSS -> {
+        AzureCredential azureCredential = vendAzureCredential(context);
+        builder.azureUserDelegationSas(new AzureUserDelegationSAS().sasToken(azureCredential.getSasToken()))
```
creechy, line 76: nit: why not move the builder into each vend method (like `vendAzureCredential()` since those methods aren't really doing much themselves.

alexreid-db: yeah, thought about this, only thing is there are other places (specifically on the Iceberg REST side that don't need to use the Unity API specific models, but they could I guess)

server/src/main/java/io/unitycatalog/server/persist/utils/ServerPropertiesUtils.java
alexreid-db: yeah this was just copy pasting what was already there, I can update to for loop

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -18,6 +19,7 @@ public TableMetadata readTableMetadata(String metadataLocation) {
     // TODO: cache fileIO
     FileIO fileIO = fileIOFactory.getFileIO(metadataLocationUri);
 
-    return TableMetadataParser.read(fileIO, metadataLocation);
+    return CompletableFuture.supplyAsync(() -> TableMetadataParser.read(fileIO, metadataLocation))
```
creechy, line 22: qq: since you are blocking for completion anyways, what's the use of using a `CompletableFuture` ? just curious if I'm missing something.

creechy: Are you trying to throttle the number of threads that are reading the metadata?

creechy (approved):
lgtm - just had little nits and questions.

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
alexreid-db: no, armeria doesn't allow blocking io calls on main thread, this gets around that. In a follow-up PR we can look at cleaning this up a little bit.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
rameshchandra: > If so should just do this in vendAwsCredentials()

@creechy `ServerPropertiesUtils` loads the properties file when the static instance is initialized. So, moving this to `vendAwsCredentials` by itself won't support dynamically adding configs, no?

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
```diff
@@ -0,0 +1,108 @@
...
+    String tableStorageLocation = table.getStorageLocation();
+    if (tableStorageLocation == null || tableStorageLocation.isEmpty()) {
+      throw new BaseException(ErrorCode.FAILED_PRECONDITION, "Table storage location not found.");
+    }
+
+    URI storageLocationUri = URI.create(tableStorageLocation);
+
+    // FIXME!! Update privileges when access controls are implemented
+    CredentialContext credentialContext = CredentialContext.create(storageLocationUri,
+      Set.of(CredentialContext.Privilege.SELECT, CredentialContext.Privilege.UPDATE));
```
rameshchandra, line 50: Shouldn't the `TableOperation` in the `GenerateTemporaryTableCredentials` request be used here? Otherwise, this always returns a read-write credential even when the client requested a read credential.

(as pointed out in the comment, when authz is added privileges should be checked to make sure the client is authorized to perform the requested `TableOperation`)

server/src/main/java/io/unitycatalog/server/persist/utils/ServerPropertiesUtils.java
```diff
@@ -54,24 +41,83 @@ public String getProperty(String key) {
...
+    while (true) {
+      String bucketPath = properties.getProperty("s3.bucketPath." + i);
+      String region = properties.getProperty("s3.region." + i);
+      String awsRoleArn = properties.getProperty("s3.awsRoleArn." + i);
+      String accessKey = properties.getProperty("s3.accessKey." + i);
+      String secretKey = properties.getProperty("s3.secretKey." + i);
+      String sessionToken = properties.getProperty("s3.sessionToken." + i);
+      if ((bucketPath == null || region == null || awsRoleArn == null)
+          && (accessKey == null || secretKey == null || sessionToken == null)) {
+        break;
```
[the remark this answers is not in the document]
alexreid-db: Yes. I didn't try and fix/change how the server properties are handled in this PR since 1) this PR is large already and 2) at some point the storage configurations should move into the db and become first class entities so not sure how much we wanted to change this before then.

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
alexreid-db: I can switch this to use the table operations directly here. There will probably be some translation of access control privileges and table operations anyways though. 

> always returns a read-write credential even when the client requested a read credential.
I think we just come from a little bit different approaches here where when someone asks for table creds, we just look up what privileges that user/role/etc. has on the table and give back a cred with those privs instead of having the client tell the server what privs it wants/needs.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
```diff
@@ -0,0 +1,81 @@
...
+      credentialsProvider =
+          StaticCredentialsProvider.create(
+              AwsBasicCredentials.create(
+                  s3StorageConfig.getAccessKey(), s3StorageConfig.getSecretKey()));
+    } else {
+      credentialsProvider = DefaultCredentialsProvider.create();
+    }
+
+    // TODO: should we try and set the region to something configurable or specific to the server
+    // instead?
```
rameshchandra, line 75: @alexreid-db Do you mean setting the region to be based on the UC server instance instead of making it per storage config?

alexreid-db: yeah, just for the sts client

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+    Resource: []
+    """;
+
+  private static final ObjectMapper JSON_MAPPER = new ObjectMapper();
+  private static final ObjectMapper YAML_MAPPER = new ObjectMapper(new YAMLFactory());
+
+  // This can support generating a policy across multiple buckets and paths, however, the assumed role
+  // the policy is applied to for a scoped-session needs to have access across those buckets
+  @SneakyThrows
+  public static String generatePolicy(Set<CredentialContext.Privilege> privileges, List<String> locations) {
```
alexreid-db, line 54: brought to my attention that there is now an IAM policybuilder library that may make some of this easier and cleaner, maybe can look at using that in an follow-up PR (or update this PR if really necessary, though what is here has already been testing)

server/src/main/java/io/unitycatalog/server/service/credential/CredentialContext.java
```diff
@@ -0,0 +1,30 @@
...
+public class CredentialContext {
+  public enum Privilege {
+    SELECT,
+    UPDATE
+  }
+
+  private String storageScheme;
+  private String storageBase;
+  private Set<Privilege> privileges;
+  private List<String> locations;
```
rameshchandra, line 20: Is this defined as a list to be future proof since the constructor currently only sets one location corresponding to the `locationURI`?

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
```diff
@@ -0,0 +1,81 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(AwsCredentialVendor.class);
+
+  private final Map<String, S3StorageConfig> s3Configurations;
+
+  public AwsCredentialVendor() {
+    this.s3Configurations = ServerPropertiesUtils.getInstance().getS3Configurations();
+  }
+
+  public Credentials vendAwsCredentials(CredentialContext context) {
+    S3StorageConfig s3StorageConfig = s3Configurations.get(context.getStorageBase());
```
rameshchandra, line 31: Does this require the `bucketPath` in the s3 config to be a bucket root path? Users may want to use a part of a bucket (instead of the whole bucket) in the storage config (which can be supported with a prefix match).

rameshchandra: Also, the paths stored in `s3Configurations` and in the context need to be normalized for the matching to be robust. This applies to URIs in Azure and GCP as well.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+        policyStatement.add(listStatement);
+
+        ArrayNode bucketResource = (ArrayNode) listStatement.findPath("Resource");
+        ArrayNode operationsResource = (ArrayNode) operationsStatement.findPath("Resource");
+        bucketResource.add("arn:aws:s3:::%s".formatted(bucketName));
+
+        ArrayNode conditionalPrefixes = (ArrayNode) listStatement.findPath("s3:prefix");
+        paths.forEach(
+          path -> {
+            String sanitizedPath = path.replaceAll("^/+", "");
```
rameshchandra, line 87: Other special characters in policy statements should also be sanitized to be safe. I think this includes `$`, `?`, `*`, `\`. (see [this page](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html))

server/src/main/java/io/unitycatalog/server/service/credential/azure/AzureCredentialVendor.java
```diff
@@ -0,0 +1,81 @@
...
+            .build();
+    DataLakeServiceAsyncClient serviceClient =
+        new DataLakeServiceClientBuilder()
+            .httpClient(HttpClient.createDefault())
+            .endpoint("https://" + locationParts.account())
+            .credential(tokenCredential)
+            .buildAsyncClient();
+    UserDelegationKey key = serviceClient.getUserDelegationKey(start, expiry).toFuture().join();
+
+    PathSasPermission perms = resolvePrivileges(context.getPrivileges());
```
rameshchandra, line 51: This should come from the `operation` parameter in the `GenerateTemporaryTableCredential` request (similar to the comment for AWS).

server/src/main/java/io/unitycatalog/server/service/credential/azure/AzureCredentialVendor.java
```diff
@@ -0,0 +1,81 @@
...
+    DataLakeServiceSasSignatureValues sasSignatureValues =
+        new DataLakeServiceSasSignatureValues(expiry, perms).setStartTime(start);
+
+    // azure supports only downscoping to a single location for now
+    // azure wants only the path
+    String path = URI.create(context.getLocations().get(0)).getPath().substring(1);
```
rameshchandra, line 57: Why strip away the first character (`substring(1)`)?

server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import lombok.SneakyThrows;
+import org.apache.iceberg.Files;
+
+public class GcpCredentialVendor {
+
+  public static final List<String> INITIAL_SCOPES =
+      List.of("https://www.googleapis.com/auth/cloud-platform");
```
rameshchandra, line 24: Isn't it safer to start with `https://www.googleapis.com/auth/devstorage.read_write` as the initial scope (from the [cloud storage oauth scopes page](https://cloud.google.com/storage/docs/oauth-scopes))

server/src/main/java/io/unitycatalog/server/service/credential/CredentialContext.java
alexreid-db: correct

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
alexreid-db: Again, I didn't try and tackle much of anything of how the server properties, in particular the storage configurations, are stored in this PR. 

I think currently, letting people set paths beyond the bucket as the bucketPath makes it overly complicated with the way we are storing stuff in a properties file. Using a properties file and having to pattern match (what if it partially matches, etc.) is not great. I would wait to add this when 1) someone from the community requests it and 2) possibly when the storage configurations are not stored in a properties file.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
alexreid-db: this is just removing any forward slashes at the beginning or s3 paths

server/src/main/java/io/unitycatalog/server/service/credential/azure/AzureCredentialVendor.java
alexreid-db: same as the aws policy, it doesn't want/like the path to start with a forward slash. I'll update it to either the regex replace or check if there is a forward slash first.

server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java
alexreid-db: Since we are always downscoping, it doesn't really matter. At some point in the future, it's theoretically possible the server may need to do something that requires more than just read/write. Also, I don't remember if this scope is actually required when using a service account. (btw, this other oss catalog does implemented this the same way)

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
alexreid-db: So, instead here, I added code to translate the Table (and Volume) operation to access control privileges in the generate cred handlers (so stuff coming from that path will use the operation passed in). Personally, think this is better instead of requiring areas (like the iceberg rest side and credential stuff) to use the unity model. We have to look up privileges anyways and translate to verify access. We could always change this later if we need to.

server/src/main/java/io/unitycatalog/server/service/credential/CredentialContext.java
rameshchandra: Can you add a comment to explain since the rest of the code works with just a single string?

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
rameshchandra: Yes, this is removing only forward slashes. Separately, the other special characters in user-provided s3 paths need to be handled to avoid giving away more access than what the user is authorized to do. It could be deferred to a separate PR (if so, can you please create an issue)?

server/src/main/java/io/unitycatalog/server/persist/utils/ServerPropertiesUtils.java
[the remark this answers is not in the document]
rameshchandra: Yeah, it can be handled separately since it is an existing issue (perhaps when moving configs to the DB) -- can you add an issue for it to track?

rameshchandra (approved):
LGTM

Issue #229, opened 2024-07-15
Support for Identity


monamohan:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you would like**

**Additional context**

zpappa:
I'll take this one as well.

PR #277, opened 2024-07-24
Simple Token based access control and Oauth2 authentication

creechy:
**PR Checklist**

**Description of changes**

This PR adds (optional) authentication and access control to the UC server. This is the first in a series to support Issue https://github.com/unitycatalog/unitycatalog/issues/229

**Access control**
Access control is implemented using JWT Bearer tokens that are generated and verified using internal signing keys. With authorization enabled, clients accessing the UC must include an `Authorization` header with a `Bearer` token that is generated through a token exchange with the UC. When using the `uc` command, for example, an access token can be provided with the command line option `--auth_token`. (See the *Authentication* section for information on how to obtain access tokens)
```
bin/uc catalog list --auth_token <token>
```

The signing keys for creating access tokens is randomly generated on startup and stored in `etc/conf/` if they do not exist. These files are
- `private-key.der` - an RSA512, 2048-bit private key, used for token signing and validation
- `public-key.der` - a matching RSA512 public key, used for token signing and validation
- `key-id.txt` - a random 256-bit id, used for JWKS verification
- `certs.json` - a representation of all the previous files for JWKS verification

The server also generates an initial access token which is also stored in `etc/conf/`. 
- `token.txt` - the initial JWT access token

Enabling access control is done in the `server.properties` file. The property
- `server.authorization` - controls whether access control is required or not (values are `disable` or `enable`) 

**Authentication**
Authentication is implemented via OAuth2 token exchange. This allows a client to submit an identity token, from an identity provider like Google or Okta, and exchange it for a server access token.

A new endpoint for this functionality
- `/api/2.1/unity-catalog/auth/tokens` - an OAuth2 token exchange endpoint

This endpoint accepts an identity token, along with other OAuth2 parameters, validates the token using JWKS and OIDC discovery, and if validation is successful, generates and returns a new JWT access-token (generated using the previously described signing parameters)

_Token exchange with `uc` command_
The `uc` command has been enhanced with an `auth` `login` sub-command, in which you provide an identity token with the `--identity_token` parameter. As an example, you can obtain a Google Cloud identity token with
```
gcloud auth print-identity-token
```
And then perform the token exchange with
```
bin/uc auth login --identity_token <token>
```

_Identity provider login with `uc` command_
The `uc` command can also do a full OAuth2 Authorization Flow with identity providers like Google, Okta and Keycloak. In order to do that, there are additional parameters to set in the `server.properties` file
- `server.authorization-url` - The authorization URL for the identity provider
- `server.token-url` - The token URL for the identity provider
- `server.client-id` - The identity provider account Issuer client id
- `server.client-secret` - The identity provider account issuer client secret
- `server.redirect-port` - Optional port number to use for the redirect phase (Okta needs this)

With these properties configured, use the same command
```
bin/uc auth login
```
*without* the `--identity_token` parameter to initiate and Authorization Flow and token exchange with Google.

**Compute engine authentication**

_Duckdb_

The only change you need to make is to supply a valid access-token when configurating the secret:
```
CREATE SECRET (
      TYPE UC,
      TOKEN '<access-token>',
      ENDPOINT 'http://127.0.0.1:8080',
      AWS_REGION 'us-east-2'
 );
```

_Trino_

With Trino, you just need to supply an access-token when configuring the catalog `catalog/iceberg.properties`:
```
connector.name=iceberg
iceberg.catalog.type=rest
iceberg.rest-catalog.uri=http://127.0.0.1:8080/api/2.1/unity-catalog/iceberg
iceberg.rest-catalog.security=OAUTH2
iceberg.rest-catalog.oauth2.token=<access-token>
```


server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -89,6 +111,24 @@ private void addServices(ServerBuilder sb) {
...
         icebergResponseConverter);
+
+    ServerPropertiesUtils serverPropertiesUtils = ServerPropertiesUtils.getInstance();
+    String authorization = serverPropertiesUtils.getProperty("server.authorization");
+    // TODO: eventually might want to make this secure-by-default.
+    if (authorization != null && authorization.equalsIgnoreCase("enable")) {
+      LOGGER.info("Authorization enabled.");
+      AuthDecorator authDecorator = new AuthDecorator();
+      ExceptionHandlingDecorator exceptionDecorator = new ExceptionHandlingDecorator(new GlobalExceptionHandler());
+      sb.decorator(basePath + "catalogs", authDecorator)
```
alexreid-db, line 122: is there a way to do this so all paths by default have the auth decorator applied and we can just add exemptions (the token endpoint, etc.)?

server/src/main/java/io/unitycatalog/server/service/AuthDecorator.java
```diff
@@ -0,0 +1,81 @@
...
+            .filter(h -> h.getKey().equals(HttpHeaderNames.AUTHORIZATION))
+            .map(Map.Entry::getValue).findFirst()
+            .orElse(null);
+
+    if (authorization == null) {
+      throw new AuthorizationException(ErrorCode.UNAUTHENTICATED, "No authorization found.");
+    }
+
+    String[] parts = authorization.split(" ");
+    if (parts.length == 2 && parts[0].equals("Bearer")) {
```
alexreid-db, line 56: Do we want to throw if it doesn't adhere to this?

alexreid-db (approved):
lgtm as far as what I'm familiar with, just a couple questions

possible to add some unit tests for some of this?

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
creechy: I'll look again. I agree it would be nice to be "secure by default" with exceptions.

creechy: Ok was able to figure out how to secure all endpoints with exceptions.

server/src/main/java/io/unitycatalog/server/service/AuthDecorator.java
creechy: Ok throwing exception if we don't get expected Authorization header.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -0,0 +1,159 @@
...
+          ErrorCode.INVALID_ARGUMENT,
+          "Unsupported requested token type: " + request.getRequestedTokenType());
+    }
+
+    if (request.getSubjectTokenType() == null) {
+      throw new OAuthInvalidRequestException(
+          ErrorCode.INVALID_ARGUMENT, "Subject token type is required but was not specified");
+    }
+
+    if (request.getActorTokenType().isPresent()) {
```
xx-db, line 91: I think actorTokenType itself can also be null instead of Optional so this can throw NPE

server/src/main/java/io/unitycatalog/server/security/SecurityConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+      throws NoSuchAlgorithmException, InvalidKeySpecException, IOException {
+    byte[] keyBytes;
+    if (rsa512PublicKeyEncoded != null) {
+      keyBytes = Base64.getDecoder().decode(rsa512PublicKeyEncoded);
+    } else {
+      if (Files.notExists(rsa512PublicKey)) {
+        log.info("No JWT public signing key present.");
+        return null;
+      }
+      keyBytes = Files.readAllBytes(rsa512PublicKey);
```
rameshchandra, line 90: nit: couldn't lines 83-90 be simplified by moving this read (and the corresponding logic in `rsaPrivateKey`) to line 54? 

Line 53 is there to handle the case of missing key files and the invariant at line 54 is that the key files exist, otherwise the service can panic.

server/src/main/java/io/unitycatalog/server/security/SecurityConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+
+      byte[] keyIdBytes = new byte[32];
+      new SecureRandom().nextBytes(keyIdBytes);
+      Files.writeString(keyId, Hex.encodeHexString(keyIdBytes), StandardOpenOption.CREATE);
+    }
+  }
+
+  public Algorithm algorithmRSA()
+      throws NoSuchAlgorithmException, InvalidKeySpecException, IOException {
+    return Algorithm.RSA512(rsaPublicKey(), rsaPrivateKey());
```
rameshchandra, line 77: Nit: What happens if the public key and private key don't match? This can happen if there's a race in reading the key files in `rsaPublicKey` and `rsaPrivateKey` while they are being updated (see my comment below)? Arguably this will be very rare so this PR can go in without fixing it, but would like to understand the impact (and maybe create a ticket to fix later)? 

The same kind of inconsistency can also happen in lines 57-60 of `SecurityContext` when the private key, public key, and key id are being read separately from the filesystem.

server/src/main/java/io/unitycatalog/server/security/SecurityContext.java
```diff
@@ -0,0 +1,126 @@
...
+    serviceTokenFile = configurationFolder.resolve("token.txt");
+
+    rsaPublicKey = securityConfiguration.rsaPublicKey();
+    rsaPrivateKey = securityConfiguration.rsaPrivateKey();
+    algorithm = securityConfiguration.algorithmRSA();
+    keyId = securityConfiguration.getKeyId();
+
+    // Do this every time because the SecurityConfiguration parameters
+    // might have changed since last started.
+    serviceToken = createServiceToken();
```
rameshchandra, line 64: to clarify: does this mean every time the service restarts the tokens that were previously issued would be invalid?

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -0,0 +1,158 @@
...
+            .accessToken(accessToken)
+            .issuedTokenType(TokenTypes.ACCESS)
+            .tokenType(AuthTypes.BEARER)
+            .build();
+
+    return HttpResponse.ofJson(response);
+  }
+
+  // TODO: This should be probably integrated into the OpenAPI spec.
+  @ToString
```
rameshchandra, line 117: Nit: Can you better document the API -- what's being exchanged for what etc.,? 

IIUC, the `subjectToken`, which could be an OAuth access token from an external IdP (e.g., Google) is being exchanged for an access token by the local issuer for the same email subject -- is that correct?

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -0,0 +1,158 @@
...
+    DecodedJWT decodedJWT = JWT.decode(request.getSubjectToken());
+    String issuer = decodedJWT.getClaim("iss").asString();
+    String keyId = decodedJWT.getHeaderClaim("kid").asString();
+
+    LOGGER.debug("Validating token for issuer: {}", issuer);
+
+    JWTVerifier jwtVerifier = jwksOperations.verifierForIssuerAndKey(issuer, keyId);
+    decodedJWT = jwtVerifier.verify(decodedJWT);
+
+    String accessToken = securityContext.createAccessToken(decodedJWT);
```
rameshchandra, line 104: Should this access token have a limited lifetime for security?

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -0,0 +1,158 @@
...
+   *
+   * @param request The token exchange parameters
+   * @return The token exchange response
+   */
+  @Post("/tokens")
+  public HttpResponse grantToken(OAuthTokenExchangeRequest request) {
+    LOGGER.debug("Got token: {}", request);
+
+    if (request.getGrantType() != null
+        && !TokenTypes.ACCESS.equals(request.getRequestedTokenType())) {
```
rameshchandra, line 79: IIUC the returned access token only asserts the identity, so why not also support ID tokens?

server/src/main/java/io/unitycatalog/server/security/SecurityContext.java
```diff
@@ -0,0 +1,126 @@
...
+        .withIssuer(localIssuer)
+        .withIssuedAt(new Date())
+        .withKeyId(keyId)
+        .withJWTId(UUID.randomUUID().toString())
+        .withClaim(JwtClaim.TOKEN_TYPE.key(), JwtTokenType.ACCESS.name())
+        .withClaim(JwtClaim.SUBJECT.key(), decodedJWT.getClaim("email").asString())
+        .sign(algorithm);
+  }
+
+  public String createServiceToken() {
```
rameshchandra, line 85: The service token creation is not exposed via any API -- is this for the future to support admin access tokens?

server/src/main/java/io/unitycatalog/server/security/SecurityContext.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.Map;
+import java.util.UUID;
+import lombok.Getter;
+import lombok.SneakyThrows;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SecurityContext {
+
+  public interface Issuers {
```
rameshchandra, line 27: Nit: why this instead of an enum?

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -89,6 +112,21 @@ private void addServices(ServerBuilder sb) {
...
+    String authorization = serverPropertiesUtils.getProperty("server.authorization");
+    // TODO: eventually might want to make this secure-by-default.
+    if (authorization != null && authorization.equalsIgnoreCase("enable")) {
+      LOGGER.info("Authorization enabled.");
+      AuthDecorator authDecorator = new AuthDecorator();
+      ExceptionHandlingDecorator exceptionDecorator =
+          new ExceptionHandlingDecorator(new GlobalExceptionHandler());
+      sb.routeDecorator()
+          .pathPrefix(basePath)
+          .exclude(basePath + "auth/tokens")
```
rameshchandra, line 126: Isn't this `controlPath + "auth/tokens"`?

server/src/main/java/io/unitycatalog/server/service/AuthService.java
rameshchandra: Nm. Realized that this is requested token type and not the subject token type.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
creechy: Fixed.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
creechy: We are planning on addressing this in the next release along with token refresh. We agreed that for the short term, it is sufficient to invalidate existing tokens by rotating signing keys (by deleting and recreating them)

server/src/main/java/io/unitycatalog/server/security/SecurityConfiguration.java
creechy: Although there is technically a race condition, currently the keys are only created on startup, so in practicality that should never happen. If we built in a method to do manual or automatic rotation that could become a problem. I'll create an issue to keep note of it.

server/src/main/java/io/unitycatalog/server/security/SecurityContext.java
creechy: The previous token(s) only become invalid when the signing keys are rotated, but yes this does create a new token on startup.

server/src/main/java/io/unitycatalog/server/security/SecurityContext.java
creechy: Yeah this is mainly for the future. In other platforms, we had different types of tokens that are used in specific contexts so I wanted to build that here, although for now there is no practical difference between service tokens and (user) access tokens.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
creechy: good catch. I recently moved the endpoint out of the formal `/unity-catalog` namespace, but missed this.

server/src/main/java/io/unitycatalog/server/security/SecurityContext.java
creechy: Token issuers are typically URIs and expressed as a String in the JWT claim, so I made the internal ISSUER data type match.

etc/conf/server.properties
```diff
@@ -1,4 +1,10 @@
 server.env=dev
+server.authorization=disable
+server.authorization-url=
```
rameshchandra, line 3: Can you document the `authorization-url`, `token-url`, `client-id`, `client-secret` with an example?

server/src/main/java/io/unitycatalog/server/service/AuthService.java
creechy: I added some javadoc to the `grantToken()` method. Yes, you are correct, the `subjectToken` would be an id or access token from an IdP.

etc/conf/server.properties
creechy: ok added some samples for Google

server/src/main/java/io/unitycatalog/server/security/SecurityConfiguration.java
rameshchandra: If keys are only created on startup, the `rsaPublicKey`, `rsaPrivateKey`, and `getKeyId` can be simplified by reading them on startup, correct?

server/src/main/java/io/unitycatalog/server/service/AuthService.java
rameshchandra: Should we create an issue for this?

rameshchandra (approved):
LGTM

Issue #98, opened 2024-06-27
Refactor and enforce Java style check


tdas:
**Is your feature request related to a problem? Please describe.**

Currently, all the code is not style checked. The styles are added in dev/ but the build by default do not run style check - https://github.com/unitycatalog/unitycatalog/blob/main/build.sbt#L52-L53

**Describe the solution you would like**

We would like to the enable the stylecheck and refactor all the existing code to fit into that style. Since there is a lot of code already, refactoring all of that code in 1 PR is going to be a massive change. So here is my suggested approach to whoever wants to do it 

- Phase 1: Get the javastyle plugin working on the configuration. Currently it is throwing a weird error :( 

- Phase 2: Slowly refactor codes in parts over multiple PRs
      - Run the stylecheck locally to generate all the issues that need to be fixed.
      - Pick one or two directories in the source codes at a time and fix them. Make sure your generated diff is not more than 100-200 lines. This would make the PR much easier to review and quickly merge. But do not enable the style to run automatically. 
      - Keep making such size-bounded PRs until we have fixed all the source codes.

- Phase 3: Enable stylecheck to be run by default.

Multiple folks can pick up different parts of source code and start creating PRs. Before you pick up, comment on this issue to say which modules/projects/directories you are picking, so that others do not

nastra:
From previous experience using Checkstyle in Iceberg and in other projects, the main issue with Checkstyle is that it's difficult for people to (automatically) apply the style being enforced. That's why we switched over to using **spotless** in Iceberg. With **spotless** he respective code format can be enforced and automatically applied (via `./gradle spotlessApply` or via the format action in IntelliJ). 
I'm happy to help out and introduce spotless here if there's interest (I was also the main person that added spotless to the Iceberg project)

alcole:
I have found a fix for the 
[error] (client / checkstyle) java.io.FileNotFoundException: checkstyle-config.xml (No such file or directory)
errors

This enables this to run, but you can another set of errors. I've spent a bit of time on this, but nothing fixed yet. I can submit a pr so at least the config is loading

tdas:
Thank you folks for helping us. 
@nastra I see that there is a spotless sbt plugin - https://github.com/moznion/sbt-spotless . could you give that a spin (I know sbt vs gradle is a paradigm shift :) )
[...]

PR #289, opened 2024-07-26
removed checkstyle config - as conflicts with javafmt

alcole:
**PR Checklist**

**Description of changes**

Removed the configuration and plugin listing for checkstyle as javafmt is not preferred and the 2 systems had confllicting outputs. see https://github.com/unitycatalog/unitycatalog/issues/98


nastra:
I don't think we want to remove checkstyle in general but only adjust/remove the rules that conflict with the code formatter

build.sbt
```diff
@@ -88,21 +88,8 @@ resolvers ++= Seq(
...
 
-def javaCheckstyleSettings(configLocation: File) = Seq(
-  checkstyleConfigLocation := CheckstyleConfigLocation.File(configLocation.toString),
-  checkstyleSeverityLevel := Some(CheckstyleSeverityLevel.Error),
-  // (Compile / compile) := ((Compile / compile) dependsOn (Compile / checkstyle)).value,
-  // (Test / test) := ((Test / test) dependsOn (Test / checkstyle)).value,
-)
-
-// enforce java code style
-def javafmtCheckSettings() = Seq(
```
nastra, line 99: this shouldn't be changed as this is for the code formatter

build.sbt
```diff
@@ -88,21 +88,8 @@ resolvers ++= Seq(
   "Maven Central" at "https://repo1.maven.org/maven2/",
 )
 
-def javaCheckstyleSettings(configLocation: File) = Seq(
```
nastra, line 91: no need to remove this. We should rather check the conflicting checkstyle rules rather than removing everything. Having checkstyle still has value, so we need to double-check the conflicting checkstyle rules

alcole:
I can take a look and try and align the rules, I know where the configuration is to do this. I do want to check first - what would be the rationale for having 2 tools doing equivalent task?

nastra:
> I can take a look and try and align the rules, I know where the configuration is to do this. I do want to check first - what would be the rationale for having 2 tools doing equivalent task?

Checkstyle can enforce other things, like proper naming of constants, methods, ... which can't necessarily be achieved with the code formatter

alcole:
Got it. I'll try and get the import rules aligned across the 2 tools

alcole:
will start a new pr with the import work in

PR #290, opened 2024-07-26
Mark some properties in the all.yaml file as required

### LINKED ISSUE #280: Mark some properties in the all.yaml file as required
**Describe the bug**

I'm working on converting [the OpenAPI specification](https://github.com/unitycatalog/unitycatalog/blob/main/api/all.yaml) to TypeScript interfaces. I have found some properties are nullable while they should be (potentially) required, e.g., I guess the `name` and `catalog_name` of `SchemaInfo` should be required:

https://github.com/unitycatalog/unitycatalog/blob/main/api/all.yaml#L995

**To Reproduce**

I'm using [openapi-typescript](https://www.npmjs.com/package/openapi-typescript) to convert the specification.

1. npx openapi-typescript /path/to/spec/api.yaml --output /path/to/output/schema.ts

**Expected behavior**

Some output properties should not be `Optional` in a typed language like TypeScript.

**Additional context**

Basically, it’s not that problematic, but when using a typed language, it gets converted to Optional (in TypeScript, fields with a ? suffix), which can be a bit cumbersome during type checking and type guarding. It would be helpful if the OpenAPI spec could include required fields. 

Screenshot 2024-07-25 at 13 11 59
Screenshot 2024-07-25 at 13 12 19

ognis1205:
**Description of changes**
Resolves #280.

Marked some properties in the all.yaml file as required and updated the related API documents. The updating strategy is as follows:

 - Update the properties marked as `required` in the `FooInfo` schema, by referring to the properties marked as `required` in `CreateFoo` schema.
 - In addition to the above, update the property corresponding to its `id` to required in `FooInfo`.

Verified that it works in the local environment and does not break backward compatibility. It is also possible to consider making all properties defined as required nullable as another implementation approach, but this was not adopted.

### DIFF
--- a/api/all.yaml
+++ b/api/all.yaml
@@ -790,38 +790,45 @@ components:
       type: object
       properties:
         catalog_name:
-          type: string
           description: The name of the catalog where the schema and the volume are
+          type: string
         schema_name:
-          type: string
           description: The name of the schema where the volume is
-        name:
           type: string
+        name:
           description: The name of the volume
+          type: string
         comment:
+          description: The comment attached to the volume
           type: string
           maxLength: 65536
           minLength: 1
-          description: The comment attached to the volume
         created_at:
-          type: integer
-          format: int64
           description: Time at which this volume was created, in epoch milliseconds.
-        updated_at:
           type: integer
           format: int64
+        updated_at:
           description: Time at which this volume was last modified, in epoch milliseconds.
+          type: integer
+          format: int64
         volume_id:
-          type: string
           description: Unique identifier for the volume
+          type: string
         "volume_type":
           "$ref": "#/components/schemas/VolumeType"
         storage_location:
-          type: string
           description: The storage location of the volume
-        full_name:
           type: string
+        full_name:
           description: Full name of volume, in form of __catalog_name__.__schema_name__.__volume_name__.
+          type: string
+      required:
+        - catalog_name
+        - schema_name
+        - name
+        - volume_id
+        - volume_type
+        - storage_location
     ColumnTypeName:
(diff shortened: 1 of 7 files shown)


vikrantpuppala (approved):
thank you for adding this! @haogang to merge

ognis1205:
I squashed the commits into one. I will also sign-off the commit later.

api/all.yaml
```diff
@@ -1316,6 +1322,22 @@ components:
...
+        - full_data_type
+        - routine_body
+        - routine_definition
+        - parameter_style
+        - is_deterministic
+        - sql_data_access
+        - is_null_call
+        - security_type
+        - specific_name
+        - properties
```
haogang, line 1340: I believe many fields here (like properties) should be optional. Is it the case that in the current API implementation, they're required?

vikrantpuppala: I don't think they are required

ognis1205: Thank you for the comment. These lines are just reformatting the previous implementation (specifically, moving the required field to the end of the CreateFunction object). I will update this implementation to align with the current API specification. Could you please provide a reference to the current API specification or guide which fields to be optional?

vikrantpuppala: Hi Shingo, I think we should just remove properties from required for now, based on the implementation, others are required. thanks!

ognis1205: Thanks for the guide, Vikrant! I've updated the PR to make the `properties` field optional in the `CreateFunction` and `FunctionInfo` objects.

cc: @haogang @vikrantpuppala

vikrantpuppala:
@rameshchandra

PR #291, opened 2024-07-27
[FOLLOW-UP] Format Spark Integration code by javafmtAll

amaliujia:
**Description of changes**

[FOLLOW-UP] Format Spark Integration code by javafmtAll.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
@@ -2,7 +2,7 @@ package io.unitycatalog.connectors.spark
 
 import io.unitycatalog.client.{ApiClient, ApiException}
 import io.unitycatalog.client.api.{TablesApi, TemporaryTableCredentialsApi}
-import io.unitycatalog.client.model.{AwsCredentials, GenerateTemporaryTableCredential, ListTablesResponse, TableOperation, TableType}
+import io.unitycatalog.client.model.{AwsCredentials, GenerateTemporaryTableCredential, TableOperation, TableType}
 
 import java.net.URI
 import java.util
@@ -14,7 +14,6 @@ import org.apache.spark.sql.connector.expressions.Transform
 import org.apache.spark.sql.types.{DataType, StructField, StructType}
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 
-import scala.collection.convert.ImplicitConversions._
 import scala.jdk.CollectionConverters._
 
 /**
@@ -34,7 +33,7 @@ class UCSingleCatalog extends TableCatalog {
 
   override def name(): String = deltaCatalog.name()
 
-  override def listTables(namespace: Array[String]): Array[Identifier] = deltaCatalog.listTables(namespace)
+  override def listTables(namespace: Array[String]): Array[Identifier] = ???
 
   override def loadTable(ident: Identifier): Table = deltaCatalog.loadTable(ident)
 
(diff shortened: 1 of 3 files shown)


amaliujia:
I also tried to run ` build/sbt javafmtAll ` and it seemed to update other files style as well.

amaliujia:
@cloud-fan

jaceklaskowski:
Can you just leave the description of the changes alone? 🙏  Makes git log so terrible to look at 😢

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
```diff
@@ -92,7 +92,13 @@ private class UCProxy extends TableCatalog {
     this.name
   }
 
-  override def listTables(namespace: Array[String]): Array[Identifier] = ???
+  override def listTables(namespace: Array[String]): Array[Identifier] = {
+    if (namespace.length > 1) {
+      throw new ApiException("Unity Catalog does not support nested namespace.")
```
jaceklaskowski, line 97: 1. "nested namespaces" (plural)
2. Can you add what the `namespace` is in the exception

```
Nested namespaces are not supported: [namespace]
```

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
```diff
@@ -92,7 +92,13 @@ private class UCProxy extends TableCatalog {
...
+    }
+    val response: ListTablesResponse = tablesApi.listTables(this.name, namespace.head, 0, null)
```
jaceklaskowski, line 99: What do you think about introducing local variables with names for the input arguments, i.e. `catalogName`, `schemaName`, `maxResults` and `pageToken`? Java doesn't support named arguments unfortunately. I think it would make code comprehension a bit better (unless you're in IntelliJ IDEA and gets all this for free 😉 )

A bit weird to see `null` among the input arguments, isn't it?

jaceklaskowski (changes requested):

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
amaliujia: ah I see. Sure done. This seems to be more readable.

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
amaliujia: done.

amaliujia:
> Can you just leave the description of the changes alone? 🙏 Makes git log so terrible to look at 😢 
Not sure if I understand this comment but tried to address it.

connectors/spark/src/test/java/io/unitycatalog/connectors/spark/CredentialTestFileSystem.java
```diff
@@ -1,88 +1,87 @@
...
 import org.apache.hadoop.fs.*;
 import org.apache.hadoop.util.Progressable;
 
-import java.io.IOException;
-
 // A wrapper over the local file system to test UC table credentials.
 public class CredentialTestFileSystem extends RawLocalFileSystem {
 
-    @Override
-    protected void checkPath(Path path) {
```
cloud-fan, line 13: please follow the existing code style in the file.

cloud-fan:
This is hard to review because it also changes the indentation. Please don't do so as this project does not share the same code style standard with Apache Spark.

connectors/spark/src/test/java/io/unitycatalog/connectors/spark/CredentialTestFileSystem.java
amaliujia: I tried to run `build/sbt javafmtAll` and this is the result. `build/sbt javafmtAll` is from documentation. 

Are you saying we should not reply on this format tool? (I assume the previous commit didn't run this format tool and merged)

amaliujia:
Screenshot 2024-07-28 at 9 52 14 PM

@cloud-fan the main page documentation says to format the code by `build/sbt javafmtAll`

connectors/spark/src/test/java/io/unitycatalog/connectors/spark/CredentialTestFileSystem.java
cloud-fan: hmmm, seems it's not enforced...  Anyway, we should have a PR for format changes only, so that it's easier to review the actual changes.

cloud-fan (approved):

PR #295, opened 2024-07-29
Support Spark DROP TABLE command

amaliujia:
**Description of changes**

Support Spark DROP TABLE command

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
@@ -55,7 +55,7 @@ class UCSingleCatalog extends TableCatalog {
   }
   override def alterTable(ident: Identifier, changes: TableChange*): Table = ???
 
-  override def dropTable(ident: Identifier): Boolean = ???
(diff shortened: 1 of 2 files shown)


amaliujia:
@cloud-fan

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
```diff
@@ -174,7 +174,14 @@ private class UCProxy extends TableCatalog {
 
   override def alterTable(ident: Identifier, changes: TableChange*): Table = ???
 
-  override def dropTable(ident: Identifier): Boolean = ???
+  override def dropTable(ident: Identifier): Boolean = {
+    if (ident.namespace().length > 1) {
```
cloud-fan, line 178: can we add a common method to do namespace check?

cloud-fan (approved):

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
amaliujia: ah right. done

PR #309, opened 2024-07-31
Add python sdk

### LINKED ISSUE #286: Add python sdk

Just like we have the openapi generated java client sdk, there needs to be generated Python sdk for python integrations with UC.

It's important for auto-generated client SDKs to be coversioned with the server with the API spec and the server supporting the spec. Hence it's best if all the auto-generated clients ar in this repo so that all of them can released together. Otherwise it is going to be a major version hell if the server of version x is going to be compatible with java sdk version x (since java is already in this repo), python sdk version y, c++ sdk version z, etc. 

I think the best way to proceed is to add the python generation in the sbt project in the same way as that has been added for java. 
- location: clients/python 
- open API generation configuration: same as the current python sdk 0.1 released in pypi.

kevinzwang:
**PR Checklist**

**Description of changes**

This PR adds a Python client SDK generated from the OpenAPI spec. 

Unlike [unitycatalog/unitycatalog-python](https://github.com/unitycatalog/unitycatalog-python), it does not have a vendor dependency with Stainless but instead uses the [openapi-generator](https://github.com/OpenAPITools/openapi-generator/) library. As a result, the library interface is different.

This PR adds a step to `build.sbt`, which runs when `build/sbt generate` is called and generates a Python client at `target/clients/python`. 

Resolves #286


kevinzwang:
Pinging @tdas as well

tdas (commented):
This is a great start. 

1. Let me first start with things that dont need to be generated (and so should be added to .openapi-generator-ignore. 

- .github/workflows
- .gitlab-ci.yml
- .travis.yml
- README.md
- docs/**
- git_push.sh (what is this for)?
- tests/** (the generated test stubs are not really testing anything.)
Removing this will reduce a lot of the clutter. Thing we remove can always be regenerated and added later. So lets start with the absolutely essential stuff.

2.  Then, the big change I would love to see is this being integrated in the main build. we have all the generation of code (java server and java client) thats all integrated in our SBT build such that a single command `build/sbt generate` can generate everything. We want the same for all client languages. For that to work, lets try the following: 
 
- just replicate the following lines in the build.sbt file - https://github.com/unitycatalog/unitycatalog/blob/main/build.sbt#L105-L152
- name the new variable as pythonClient
- modify the openapi specific settings as needed for python client generation.

clients/README.md
```diff
@@ -0,0 +1,10 @@
+# Unity Catalog Clients
```
tdas, line 1: we need to generate all the client together in a centralized way. today it is `build/sbt generate`. for python we need to integrate with that.

kevinzwang: I'm not very familiar with Scala but I tried to add the python client generation to `build.sbt`, and when I run `build/sbt generate`, it seems to properly generate it in `target/clients/python`. 

Since it's now generated in `target/`, I removed the actual generated code from `clients/python` to match the Java client generation behavior. In addition, most of the clutter are removed by disabling tests and documentation generation. However I kept some of the files such as `git_push.sh` and `.github/workflows` since they also seem to be present in the generated Java client. 

Let me know how this looks!

kevinzwang:
@tdas just checking in, do you think this is ready to be merged?

tdas:
@vikrantpuppala can you take a look at this PR? basically, generate the python code and locally test out whether it connects to the UC server or not? 

also cc @kevinzwang let's think about how to add tests for this. today the java client gets tests via the server tests (that is, the client is used to talk to the server spun up in the local test). For python, I can imagine the following way to test this out
- start the server and wait for it to be ready - see https://github.com/unitycatalog/unitycatalog/blob/main/tests/test_tutorial.py for inspiration
- use the generated python client to run some basic operations on the server and validate the output.

we can put these tests under clients/python/

kevinzwang:
Hi @tdas, thanks for the review! As for tests, what do you think about porting the tests in `tests/test_tutorial.py` to a `tests/test_tutorial_python.py` which uses the python SDK?

tdas:
wouldnt it better to have a proper python unit tests framework for us to organize tests like a usual python library? thats what we would do if this were handwritten python code ... isn't it?

kevinzwang:
> wouldnt it be better to have a proper python unit tests framework for us to organize tests like a usual python library? thats what we would do if this were handwritten python code ... isn't it?


Yes, I intend to use pytest for proper unit testing. What I meant is that for the initial tests, I would use the ones in 
`test_tutorial.py`, translated to python sdk

tdas:
aah yes. that would be start. we can always extend to other endpoints. but lets have the basic catalog, schema, table API tests in there before we merge this PR.

kevinzwang:
@tdas added Python unit tests as well as a github action that runs them automatically in a PR! Looks like it's working well

kevinzwang:
Ok I followed the instructions from the DCO action and really messed up the commit history...

PR #320, opened 2024-08-02
Fix list max results

vksx:
**PR Checklist**

**Description of changes**

- Bugfix for `list --max_results` using PagedListingHelper for **functions**.
- Bugfix for `list --max_results` for **volumes** (without PagedListingHelper) - this might require more attention using PagedLisitngHelper - so I just added a fix for it so that it's working for now.

cc: @vikrantpuppala


server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
```diff
@@ -7,16 +7,30 @@
...
           PropertyRepository.findProperties(session, UUID.fromString(uuid), entityType);
-      Method setPropertiesMethod = entityInfo.getClass().getMethod("setProperties", Map.class);
-      setPropertiesMethod.invoke(entityInfo, PropertyDAO.toMap(propertyDAOList));
+      Class<?> entityClass = PROPERTY_TYPE_MAP.getOrDefault(entityType, Map.class);
+      Method setPropertiesMethod =
+          entityInfo.getClass().getMethod("setProperties", entityClass);
+      Map<String, String> propertyMap = PropertyDAO.toMap(propertyDAOList);
+      Object propertiesArgument = switch (entityClass.getSimpleName()) {
+          case "Map" -> propertyMap;
+          case "String" -> propertyMap.toString();
```
vikrantpuppala, line 30: So, this function now uses json style string for functions and the map for others?

vksx: Yes, because in `FunctionInfo` the argument for `setProperties` is a string unlike other entities (`Map<String, String>`). So, I took this approach to avoid "argument type mismatch".

vikrantpuppala (approved):
Thanks for making this change @vksx 
@haogang to merge

haogang (approved):

Issue #182, opened 2024-07-09
[Icerberg REST] Namespace separator mismatch


szlta:
**Describe the bug**

The current Iceberg dependency version is [1.5.2](https://github.com/unitycatalog/unitycatalog/blob/main/build.sbt#L187), which uses `%1F` as [namespace separator](https://github.com/apache/iceberg/blob/apache-iceberg-1.5.2/core/src/main/java/org/apache/iceberg/rest/RESTUtil.java#L39) character in its client implementation.

Unity however uses `.` [character](https://github.com/unitycatalog/unitycatalog/blob/main/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java#L241) which causes client calls such as `listTables()` or `loadTable()` to fail if a Namespace is defined with more levels.

**To Reproduce**

1. Create a uniform table such as `unity_test.default.mytable`
2. Use the Iceberg client and call `listTables(... , Namespace.of("unity_test", "default"));`
3. This yields to 400 Bad request

**Expected behavior**

Retrieve a TableIdentifer for `mytable`.

dennyglee:
Thanks @szlta - we will review this shortly!

alexreid-db:
Hi I will be looking at making some changes to address this

moomindani:
Still reproducible on `main` at `58d5c7b` (Iceberg is 1.11.0 now, and `RESTUtil` still joins levels with `%1F`). The mechanism is a little different from what the description says, though, so I wanted to record what I measured -- the separator is not chosen by `IcebergRestCatalogService`.

Every client request goes through `URLTranscoderVerticle`, which listens on the port clients use and forwards to Armeria on `port + 1`. It rewrites the path with `path.replace("%1F", ".")`, and query parameters by replacing the unit separator U+001F with `.`, before forwarding. So by the time any Iceberg endpoint sees the namespace, `default%1Fnested` and the literal `default.nested` are the same string:

```
# through the transcoder (port 8080)
GET /api/2.1/unity-catalog/iceberg/v1/catalogs/unity/namespaces/default%1Fnested
400 {"error":{"message":"Invalid schema full name: unity.default.nested","type":"BaseException","code":400}}

GET /api/2.1/unity-catalog/iceberg/v1/catalogs/unity/namespaces/default.nested
400 {"error":{"message":"Invalid schema full name: unity.default.nested","type":"BaseException","code":400}}

[...]

Issue #3, opened 2024-06-13
[feature request] Add additional Iceberg REST Catalog endpoints


kevinjqliu:
Unity Catalog comes pre-installed with an Iceberg REST catalog service. [1] [2] 
Currently only read endpoints are supported. 
<img src="https://github.com/unitycatalog/unitycatalog/assets/9057843/ec461a8c-3ff5-48e5-be42-a2c6dc619483" width=50% height=50%>

Opening this issue to track adding additional endpoints based on the Iceberg REST catalog OpenAPI spec [3]
This will unlock the ability to write Iceberg tables to the Unity Catalog. 

[1] https://github.com/unitycatalog/unitycatalog/blob/main/docs/tutorial.md#read-delta-uniform-tables-via-iceberg-rest-catalog
[2] https://github.com/unitycatalog/unitycatalog/blob/main/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
[3] https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml

Samrose-Ahmed:
It would excellent if `registerTable` (https://github.com/apache/iceberg/blob/b6c949cd86c372a87a4d43a557c19dd310af80d8/open-api/rest-catalog-open-api.yaml#L544) was supported.

amogh-jahagirdar:
Just a note after taking a look at the code: the reportMetrics endpoint at the moment is a no-op. https://github.com/unitycatalog/unitycatalog/blob/main/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java#L200

So we'll probably want to track that as well.

rameshchandra:
Thank you for highlighting the missing endpoints. Currently, only the read endpoints for namespaces and tables are implemented. The write (create, update, delete), view, and oauth endpoints are planned for the future.

cornelcreanga:
Hello @rameshchandra , is there a timeline for adding full support for the additional Iceberg catalog endpoints? We have a large datalake based on Iceberg (moving to Delta is not an option) and we want to move from an old Hive catalog to something better eg Unity, Polaris, Gravitino. Thanks.

rameshchandra:
@cornelcreanga thank you for highlighting the need. Adding full support for Iceberg is high priority and is targeted for v0.4.

adityakumarsinha:
[...]

PR #324, opened 2024-08-05
handle catalog and namespace properly in Iceberg REST api

alexreid-db:
**PR Checklist**

**Description of changes**

Prior to this PR, users were required to lookup namespace and tables by supplying the uc catalog name in the namespace identifier which isn't proper and lead to problems. This PR implements the REST client flow of configuring calls to the API with a defined prefix containing the catalog name. The catalog name is now part of the API paths and broken apart from the namespace.

Client (passes warehouse parameter) -> Server Config Endpoint (response has override for URL prefix) -> Client
Subsequent calls to the API now has url prefixed with `/catalogs/{warehouse/catalog-name}`

This PR "binds" the spark catalog to a UC catalog using the client config endpoint. Users must now add to their catalog configurations something like (example in spark):

`"spark.sql.catalog.<spark-catalog-name>.warehouse": "<catalog-name>"` and now address tables like `<spark-catalog-name>.<namespace-name>.<table-name>`

This should address https://github.com/unitycatalog/unitycatalog/issues/182 and parts of https://github.com/unitycatalog/unitycatalog/issues/3.

### DIFF
--- a/docs/usage/tables/uniform.md
+++ b/docs/usage/tables/uniform.md
@@ -26,8 +26,10 @@ The following is an example of the settings to configure OSS Apache Spark to rea
 "spark.sql.catalog.iceberg": "org.apache.iceberg.spark.SparkCatalog",
 "spark.sql.catalog.iceberg.catalog-impl": "org.apache.iceberg.rest.RESTCatalog",
 "spark.sql.catalog.iceberg.uri": "http://127.0.0.1:8080/api/2.1/unity-catalog/iceberg",
+"spark.sql.catalog.iceberg.warehouse": "<catalog-name>",
 "spark.sql.catalog.iceberg.token": "not_used",
 ```
 
-When querying Iceberg REST Catalog for Unity Catalog, tables are identified using the following pattern
-`iceberg.<catalog-name>.<schema-name>.<table-name>` (e.g. `iceberg.unity.default.marksheet_uniform`).
+When querying Iceberg REST Catalog for Unity Catalog, tables are identified using the following pattern `iceberg.<schema-name>.<table-name>` (e.g. `iceberg.default.marksheet_uniform`).
+
+NOTE: If you want your Spark catalog name to be the same as your UC catalog name, replace `iceberg` in the above configurations with `<catalog-name>`,  then tables are identified by the following pattern `<catalog-name>.<schema-name>.<table-name>` (e.g. `unity.default.marksheet_uniform`).
--- a/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
@@ -11,8 +11,6 @@ import com.linecorp.armeria.server.annotation.Param;
 import com.linecorp.armeria.server.annotation.Post;
 import com.linecorp.armeria.server.annotation.ProducesJson;
 import io.unitycatalog.server.exception.IcebergRestExceptionHandler;
-import io.unitycatalog.server.model.CatalogInfo;
-import io.unitycatalog.server.model.ListCatalogsResponse;
 import io.unitycatalog.server.model.ListSchemasResponse;
 import io.unitycatalog.server.model.ListTablesResponse;
 import io.unitycatalog.server.model.SchemaInfo;
@@ -23,6 +21,7 @@ import io.unitycatalog.server.service.iceberg.MetadataService;
 import io.unitycatalog.server.service.iceberg.TableConfigService;
 import io.unitycatalog.server.utils.JsonUtils;
 import java.io.IOException;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -31,9 +30,9 @@ import java.util.stream.Collectors;
 import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.exceptions.BadRequestException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
 import org.apache.iceberg.exceptions.NoSuchViewException;
-import org.apache.iceberg.relocated.com.google.common.base.Splitter;
 import org.apache.iceberg.rest.responses.ConfigResponse;
 import org.apache.iceberg.rest.responses.GetNamespaceResponse;
 import org.apache.iceberg.rest.responses.ListNamespacesResponse;
@@ -44,6 +43,8 @@ import org.hibernate.SessionFactory;
 
 @ExceptionHandler(IcebergRestExceptionHandler.class)
 public class IcebergRestCatalogService {
+
+  private static final String PREFIX_BASE = "catalogs/";
 
   private final CatalogService catalogService;
   private final SchemaService schemaService;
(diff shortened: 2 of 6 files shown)


alexreid-db:
@creechy @tdas @ramesh-db It would be good to get this one in if anyone has cycles to give this a review.

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
```diff
@@ -104,55 +112,37 @@ public void testNamespaces() throws ApiException, IOException {
...
           .isEqualTo(
               ListNamespacesResponse.builder()
-                  .add(Namespace.of(TestUtils.CATALOG_NAME))
+                  .add(Namespace.of(TestUtils.SCHEMA_NAME))
                   .build()
                   .toString());
     }
     // ListNamespaces from catalog
-    {
+    /*{
```
vikrantpuppala, line 145: should we remove this?

vikrantpuppala (approved):
minor comment but lgtm

bijanhoule (approved):

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -176,18 +137,17 @@ public HttpResponse tableExists(
     }
   }
 
-  @Get("/v1/namespaces/{namespace}/tables/{table}")
+  @Get("/v1/catalogs/{catalog}/namespaces/{namespace}/tables/{table}")
```
szlta, line 140: What is the purpose of a`catalog` part here? The Iceberg specification doesn't have such concept, and this change would further deviate from the spec defined at https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L240-L254
Parent namespaces should be nested into a `parent=` GET parameter to conform the specification.

alexreid-db: This does not deviate from the spec and is valid use of {prefix} as defined in the spec.

alexreid-db: > Well isn't catalog part of the namespace here
We are/would be saying that catalog is NOT considered part of the namespace.

> I don't think {prefix} was really meant to be used this way.
Again, this valid to the spec where operations are in the context of a catalog.

szlta: > catalog is NOT considered part of the namespace.

Then it makes sense of course.
But: catalog did use to be part of the namespace and on top of that, a client would be restricted to work with only one catalog from now on. That strikes me as a rather larger structural change compared to just a namespace separator fix, wouldn't you agree?

Also - how is this going to be backward compatible with older client versions? (Considering those that provided `.` as separator of course..)

alexreid-db: > But: catalog did use to be part of the namespace and on top of that, a client would be restricted to work with only one catalog from now on. That strikes me as a rather larger structural change compared to just a namespace separator fix, wouldn't you agree?

Not if we felt the original implementation was incorrect. Other IRC implementations (Tabular, Polaris) utilize the prefix in this way and separate out a entity like catalog/warehouse and do not consider it to be part of the namespace so this is not a new/foreign concept.

I agree that using the . was not to spec for namespace separating, but since we consider the catalog to not be part of the namespace and UC OSS only has 1 level for namespace (Schema), we don't have to worry about a separator.

> Also - how is this going to be backward compatible with older client versions? (Considering those that provided . as separator of course..)

0.2 hasn't even been released yet, so not sure why anyone would expect it to be "stable" as far as things like this. 
Also, per the spec, REST clients should be hitting the config endpoint first to get configurations.

> one more thing: I belive this also takes away the possibility for clients to discover what catalogs there are? Should we plan to substitute the lost functionality somehow?

UC OSS has other endpoints for this that are not part of the IRC spec. (so this can be done through the cli, UI, etc.)

docs/usage/tables/uniform.md
```diff
@@ -20,7 +20,10 @@ The following is an example of the settings to configure OSS Apache Spark to rea
...
 "spark.sql.catalog.iceberg.catalog-impl": "org.apache.iceberg.rest.RESTCatalog",
 "spark.sql.catalog.iceberg.uri": "http://127.0.0.1:8080/api/2.1/unity-catalog/iceberg",
+"spark.sql.catalog.iceberg.warehouse": "<catalog-name>",
 "spark.sql.catalog.iceberg.token": "not_used",
 ```
 
-When querying Iceberg REST Catalog for Unity Catalog, tables are identified using the following pattern `iceberg.<catalog-name>.<schema-name>.<table-name>` (e.g. `iceberg.unity.default.marksheet_uniform`).
+When querying Iceberg REST Catalog for Unity Catalog, tables are identified using the following pattern `iceberg.<schema-name>.<table-name>` (e.g. `iceberg.default.marksheet_uniform`).
+
+NOTE: If you want your Spark catalog name to be the same as your UC catalog name, replace `iceberg` in the above configurations with `<catalog-name>`,  then tables are identified by the following pattern `<catalog-name>.<schema-name>.<table-name>` (e.g. `unity.default.marksheet_uniform`).
```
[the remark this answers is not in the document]
bennychow: I'm confused about the difference between "Spark catalog name" and "UC catalog name".  An Iceberg Catalog doesn't actually have a name that is part of the SQL identifier.  Outside of the Iceberg Catalog, each engine will refer to the catalog using its own alias.  

The fully qualified identifier to a table in the Iceberg Catalog is engine specific.  So for engine A, the identifier to a particular table may look like:  

CatalogForEngineA.namespace.table 

And for engine B, the identifier may look like:

CatalogForEngineB.namespace.table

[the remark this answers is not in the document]
alexreid-db: Note that we would now be configuring a UC catalog per REST client through the `warehouse` config parameter.

In Spark for instance, you can set the Spark/Engine catalog name to whatever you want in the configuration. This is just saying if you want the Engine catalog name to match up to the UC catalog name you have set, then you would want to set the Engine catalog name to the UC catalog name.

Let's say I have a UC catalog named `main` and a schema named `default` that has tables in it named `table1` and `table2`. 

I could set a spark configuration to access this like:

> "spark.sql.catalog.iceberg": "org.apache.iceberg.spark.SparkCatalog",
"spark.sql.catalog.iceberg.catalog-impl": "org.apache.iceberg.rest.RESTCatalog",
"spark.sql.catalog.iceberg.uri": "http://127.0.0.1:8080/api/2.1/unity-catalog/iceberg",
"spark.sql.catalog.iceberg.warehouse": "main"

I would then in spark-sql, list tables like so:

> \>show tables from iceberg.default;
table1
table2

If I wanted to make it so I could address the schemas/tables using the actual UC catalog name for the UC catalog I configured the Spark catalog for, then I would do:

> "spark.sql.catalog.main": "org.apache.iceberg.spark.SparkCatalog",
"spark.sql.catalog.main.catalog-impl": "org.apache.iceberg.rest.RESTCatalog",
"spark.sql.catalog.main.uri": "http://127.0.0.1:8080/api/2.1/unity-catalog/iceberg",
"spark.sql.catalog.main.warehouse": "main"

> \>show tables from main.default;
table1
table2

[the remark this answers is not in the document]
bennychow: I see now.  That makes sense.  This diagram was helpful to understand the difference between the metastore vs catalog.  https://docs.databricks.com/en/data-governance/unity-catalog/index.html#the-unity-catalog-object-model  

Thanks for explaining!

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
szlta: Okay, thanks for explaining @alexreid-db

szlta (approved):

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
```diff
@@ -78,8 +80,13 @@ protected void cleanUp() {
...
   public void testConfig() {
-    AggregatedHttpResponse resp = client.get("/v1/config").aggregate().join();
-    assertThat(resp.contentUtf8()).isEqualTo("{\"defaults\":{},\"overrides\":{}}");
+    AggregatedHttpResponse resp =
+        client.get("/v1/config?warehouse=" + TestUtils.CATALOG_NAME).aggregate().join();
+    assertThat(resp.contentUtf8())
+        .isEqualTo(
+            "{\"defaults\":{},\"overrides\":{\"prefix\":\"catalogs/"
+                + TestUtils.CATALOG_NAME
+                + "\"}}");
```
tdas, line 89: can you add a test to ensure that non-warehouse fails?

tdas (approved):
thank you for this @alexreid-db this is great!

tdas:
can you fix the DCO stuff? see the failing test.

Issue #253, opened 2024-07-18
`uc` command missing command line options when displaying help


creechy:
**Describe the bug**

`uc` command is missing help text for the options `--output json` and `--output jsonPretty`

**To Reproduce**

Show the help for any command that displays tabular output:

`uc catalog list --help`

**Expected behavior**

Document the options where appropriate.

**System [please complete the following information]:**

-   OS:  macOS Sonoma

**Additional context**

Marcinx26:
Hi, i would like to contribute, can you guys assign this issue to me?

vksx:
Hi @MrPowers If no one is working/assigned to this yet - I can take this up.

vikrantpuppala:
Thanks @Marcinx26 @vksx for offering your help, I have assigned this to @Marcinx26 given they commented first. I can assign you a different issue @vksx if you're up for it :)

vksx:
There are some points I would like to add which might be relevant to this issue - and might be helpful for @Marcinx26:

Command | Observation / Issue
----------|------
`bin/uc --help` | This line: "*To connect to specific UC server, use --server https://<host>*" could be updated to include the usage related to port as well.
`bin/uc catalog get --help` | Missing details related to optional parameters for `--output json` and `--output jsonPretty`. Same as the original issue.
`bin/uc catalog list --help`|Original issue.
`bin/uc schema get --help`| Details mentioned for table and not schema in required param: "*The full name of the table. The full name is the concatenation of the catalog name, schema name, and table/volume name separated by a dot. For example, catalog_name.schema_name.table_name.*" (Also this looks verbose/redundant - we can just mention the example rather than explaining that "full name is the concatenation...". Something like `catalog_name.schema_name.entity_name` or similar should suffice - just a small suggestion) and this is also missing details related to optional parameters for `--output json` and `--output jsonPretty`. 
`bin/uc schema update --help`|Details mentioned for table and not schema in required param.
`bin/uc schema list --help`|Missing details related to optional parameters for `--output json` and `--output jsonPretty`. Same as the original issue.
`bin/uc schema delete --help`|Details mentioned for table and not schema in required param.
`bin/uc table get --help `|Missing details related to optional parameters for `--output json` and `--output jsonPretty`. Same as the original issue.
`bin/uc table list --help`|Missing details related to optional parameters for `--output json` and `--output jsonPretty`. Same as the original issue.
`bin/uc volume read --help`|Details mentioned for table and not volume in required param.
[...]

PR #327, opened 2024-08-06
adjust --help messages in uc cli

Marcinx26:
**PR Checklist**

**Description of changes**

Hi,

This is my first PR, so I would like to say hello and highlight that this PR probably needs some attention. :)

Changes:

1. Added a new entry, `CliParams`, to distinguish between `schema_full_name` for schema (which consists of two levels: `catalog.schema`) and 'full_name" for other objects (table/function/schema with a three-level name like `catalog.schema.object`).
2. `printSubCommandHelp` now prints common parameters, including `--output`, `--server`, and `--token`.
3. Minor fixes to messages used in `--help`.

The code in this PR changes `full_name` to `schema_full_name`, which may be a breaking change. We need to consider whether this is the best option or if we should stick to `full_name` in all cases and find an alternative solution.

Issue: [253](https://github.com/unitycatalog/unitycatalog/issues/253)


vikrantpuppala:
I'm not entirely convinced on why we need to change the option from `full_name` to `schema_full_name`, I feel it might add to more confusion (and more options to remember) + feels like `full_name` when used in schema context (i.e. `bin/uc schema get ...`) is quite self explanatory. Would like to know your view point.

Marcinx26:
> I'm not entirely convinced on why we need to change the option from `full_name` to `schema_full_name`, I feel it might add to more confusion (and more options to remember) + feels like `full_name` when used in schema context (i.e. `bin/uc schema get ...`) is quite self explanatory. Would like to know your view point.

Initially, my point was to differentiate between 3rd level objects and schema, which is at the 2nd level. For me, it was misleading when I tried to handle both schema and tables with the same full_name and the same description. It would also be easier to navigate in the code with separate constants for 1st level name (catalog), 2nd level (schema), and 3rd level (function/table/volume).

I understand your point, and in my opinion, we can stick to full_name to keep things simpler (as no changes in commands are required). My proposition for the full_name message is:

"The full name in the format catalog_name.schema_name for schema, or catalog_name.schema_name.table_name for table/function/volume"

Sample output after my changes:

> Usage: bin/uc schema get [options]
> Required Params:
>   --full_name The full name in the format catalog_name.schema_name for schema, or catalog_name.schema_name.table_name for table/function/volume
> Optional Params:
>   --server UC Server to connect to. Default is reference server.
>   --auth_token PAT token to authorize uc requests.
>   --output To indicate CLI output format preference. Supported values are json and jsonPretty.
>

vikrantpuppala:
Agree with changing the message 👍 
thank you

vikrantpuppala (approved):
lgtm @haogang to merge

haogang (approved):

PR #331, opened 2024-08-08
Refactoring Tests Related to Spark Integration

xuanyuanking:
To facilitate future expansion, we are splitting the current mixed Spark Integration tests into the following parts:

- BaseSparkIntegrationTest abstract class: This includes the basic Spark Catalog session creation, default setup, cleanup functions, and serves as the base class for all Spark integration tests.
- SchemaOperationsTest: This tests schema-related operations such as create, delete, and list.
- TableReadWriteTest: This tests table-related operations, including read/write operations.

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SchemaOperationsTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SchemaOperationsTest.java
@@ -58,7 +58,7 @@ public class SchemaOperationsTest extends BaseSparkIntegrationTest {
   public void testLoadSchema() {
     SparkSession session = createSparkSessionWithCatalogs(SPARK_CATALOG);
 
-    Row[] rows = session.sql("DESC NAMESPACE " + SCHEMA_NAME).collect();
+    Row[] rows = (Row[]) session.sql("DESC NAMESPACE " + SCHEMA_NAME).collect();
     assertThat(rows).hasSize(2);
     assertThat(rows[0].getString(0)).isEqualTo("Catalog Name");
     assertThat(rows[0].getString(1)).isEqualTo(SPARK_CATALOG);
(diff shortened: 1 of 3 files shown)


connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SchemaOperationsTest.java
```diff
@@ -0,0 +1,73 @@
...
+    // TODO: We need to apply a fix on Spark side to use v2 session catalog handle
+    // `setCurrentDatabase` when the catalog name is `spark_catalog`.
+    // session.catalog().setCurrentDatabase(SCHEMA_NAME);
+    session.stop();
+  }
+
+  @Test
+  public void testListSchema() {
+    SparkSession session = createSparkSessionWithCatalogs(SPARK_CATALOG);
+    Row row = session.sql("SHOW NAMESPACES").collectAsList().get(0);
```
xuanyuanking, line 49: I'm wondering if we should rename all instances of NAMESPACE to SCHEMA for consistency. Which term is preferred in Unity Catalog? @cloud-fan

cloud-fan: In Unity Catalog we prefer SCHEMA. NAMESPACE is the Spark terminology.

cloud-fan: Let's use SCHEMA consistently.

cloud-fan (approved):

cloud-fan:
also cc @amaliujia

connectors/spark/src/test/java/io/unitycatalog/connectors/spark/SchemaOperationsTest.java
xuanyuanking: Ack. Done

amaliujia:
:) I was doing something similar locally. Thanks for sending this out!

amaliujia (approved):

PR #344, opened 2024-08-13
Added RegisteredModel handlers and tests

kriscon-db:
**PR Checklist**

**Description of changes**

- Added handers and tests for CRUD operations on registered models.
- Added CLI support for CRUD operations on registered models.
- Fixed some inconsistencies with the API definition.
- Added model handling in schema deletion.


server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  private RegisteredModelInfoDAO findRegisteredModel(
+      Session session, String catalogName, String schemaName, String registeredModelName) {
+    UUID schemaId = getSchemaId(session, catalogName, schemaName);
+    return getRegisteredModelDao(session, schemaId, registeredModelName);
+  }
+
+  public RegisteredModelInfo createRegisteredModel(CreateRegisteredModel createRegisteredModel) {
+    LOGGER.info("Starting createRegisteredModel...");
```
kriscon-db, line 118: remove

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -0,0 +1,363 @@
...
+    ValidationUtils.validateSqlObjectName(createRegisteredModel.getName());
+    long createTime = System.currentTimeMillis();
+    String modelId = UUID.randomUUID().toString();
+    String storageLocation =
```
kriscon-db, line 122: Will need to change this to instead of just getting the directory, we will need to create the directory, but we will also need to carefully roll that creation back if we fail the create.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -0,0 +1,363 @@
...
+  private static final PagedListingHelper<RegisteredModelInfoDAO> LISTING_HELPER =
+      new PagedListingHelper<>(RegisteredModelInfoDAO.class);
+
+  private ModelRepository() {}
+
+  public static ModelRepository getInstance() {
+    return INSTANCE;
+  }
+
+  public RegisteredModelInfo getRegisteredModelById(String registeredModelId) {
```
kriscon-db, line 36: Unused at the moment, but will be needed once we get model version support in place.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -0,0 +1,363 @@
...
+      registeredModelInfo.setCatalogName(catalogName);
+      registeredModelInfo.setSchemaName(schemaName);
+      registeredModelInfo.setFullName(getRegisteredModelFullName(registeredModelInfo));
+      result.add(registeredModelInfo);
+    }
+    return new ListRegisteredModelsResponse().registeredModels(result).nextPageToken(nextPageToken);
+  }
+
+  public RegisteredModelInfo updateRegisteredModel(UpdateRegisteredModel updateRegisteredModel) {
+    LOGGER.info("Update Registered Model: Processing request: " + updateRegisteredModel);
```
kriscon-db, line 251: Remove

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -0,0 +1,363 @@
...
+          origRegisteredModelInfoDAO.setComment(updateRegisteredModel.getComment());
+        }
+        long updatedTime = System.currentTimeMillis();
+        origRegisteredModelInfoDAO.setUpdatedAt(new Date(updatedTime));
+        session.persist(origRegisteredModelInfoDAO);
+        registeredModelInfo = origRegisteredModelInfoDAO.toRegisteredModelInfo();
+        registeredModelInfo.setCatalogName(catalogName);
+        registeredModelInfo.setSchemaName(schemaName);
+        registeredModelInfo.setFullName(getRegisteredModelFullName(registeredModelInfo));
+        LOGGER.debug("updateRegisteredModel:  Attempting to persist " + registeredModelInfo);
```
kriscon-db, line 306: Remove

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -0,0 +1,363 @@
...
+        throw e;
+      }
+      throw new BaseException(
+          ErrorCode.INTERNAL, "Error updating registered model: " + fullName, e);
+    }
+    return registeredModelInfo;
+  }
+
+  public void deleteRegisteredModel(String fullName, boolean force) {
+    try (Session session = SESSION_FACTORY.openSession()) {
```
kriscon-db, line 325: Add logging here.

server/src/main/java/io/unitycatalog/server/persist/SchemaRepository.java
```diff
@@ -58,6 +62,7 @@ public SchemaInfo createSchema(CreateSchema createSchema) {
             .forEach(session::persist);
         session.persist(schemaInfoDAO);
         tx.commit();
+        LOGGER.info("Added schema: {}", schemaInfo.getName());
```
kriscon-db, line 65: Remove

server/src/main/java/io/unitycatalog/server/persist/utils/FileUtils.java
```diff
@@ -32,10 +32,24 @@ private static String getStorageRoot() {
     return properties.getProperty("storageRoot");
   }
 
+  // This is temporary until storageRoot is fully supported
+  private static String getModelStorageRoot() {
+    return properties.getProperty("registeredModelStorageRoot");
```
kriscon-db, line 37: I didn't want to introduce a base storage root since we have nothing that uses it yet aside from models.  Let me know if you prefer that I change this to use the `storageRoot` property now.

rickyma:
It took nearly a week for someone to review... Amazing.... _ _#

vikrantpuppala (approved):
lgtm for the most part, would like to understand a couple of items:
1. why a separate storage root?
2. relation between registered model info and model version info.

kriscon-db:
> lgtm for the most part, would like to understand a couple of items:
> 
> 1. why a separate storage root?
> 2. relation between registered model info and model version info.

So I am not married to having a separate storage root here.  I mainly wanted to make sure I was avoiding any conflict with the existing storage root and code around it vs. fixing everything that was using it for an uncertain implementation.  I can add a followup commit that unifies the storage roots, changes the existing usages to use catalogId and schemaId instead of names, and removes all references to deletion (since we punted deletion to a later version).

The registered models and model versions are related but not the same.  There is a one to many relationship with registered models to model versions.  A registered model, may contain any number of model versions.  Think of a registered model as a reserved namespace for the model in question.  As you iterate on your model, you want to keep track of the different versions of that model.  Lets say you used different input sources or you changed hyperparameters.  You would create a new version for each of those differences and the metadata w/in a model version would help you keep track of that.  That's why model versions have the notion of source (the location of the original experiments artifacts that generated that specific model version) and runId (your ML library's way of uniquely tracking the execution pipeline that created the model).  These pieces of metadata don't live on the registered model because the registered model is just the container.  In UC parlance, the registered model *is* the securable.  The model versions are simply sub-securables.  When we get ACL's into the system, then ACLs will apply to the registered model, and the registered model ACLs will transitively be applied to the model version.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
kriscon-db: Will add this in a follow on commit that adds location creation and deals with fixing all of the existing storage root touchpoint code.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -0,0 +1,361 @@
...
+      Transaction tx = session.beginTransaction();
+      try {
+        RegisteredModelInfoDAO registeredModelInfoDAO =
+            session.get(RegisteredModelInfoDAO.class, UUID.fromString(registeredModelId));
+        if (registeredModelInfoDAO == null) {
+          throw new BaseException(
+              ErrorCode.NOT_FOUND, "Registered model not found: " + registeredModelId);
+        }
+        RegisteredModelInfo registeredModelInfo = registeredModelInfoDAO.toRegisteredModelInfo();
+        SchemaInfoDAO schemaInfoDAO =
```
ravivj-db, line 49: Do we really need to check for these, unless we did some dangling/ghost deletes where we deleted parent schema/catalog without deleting the child entities ?

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
ravivj-db: +1

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -0,0 +1,361 @@
...
+            .updatedAt(createTime);
+    String fullName = getRegisteredModelFullName(registeredModelInfo);
+    registeredModelInfo.setFullName(fullName);
+    LOGGER.info("Creating Registered Model: " + fullName);
+
+    Transaction tx;
+    try (Session session = SESSION_FACTORY.openSession()) {
+      String catalogName = registeredModelInfo.getCatalogName();
+      String schemaName = registeredModelInfo.getSchemaName();
+      UUID schemaId = getSchemaId(session, catalogName, schemaName);
```
ravivj-db, line 141: Shouldn't this be inside the transaction as well ?

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -0,0 +1,361 @@
...
+    }
+    return new ListRegisteredModelsResponse().registeredModels(result).nextPageToken(nextPageToken);
+  }
+
+  public RegisteredModelInfo updateRegisteredModel(UpdateRegisteredModel updateRegisteredModel) {
+    if (updateRegisteredModel.getNewName() != null) {
+      ValidationUtils.validateSqlObjectName(updateRegisteredModel.getNewName());
+    }
+    if (updateRegisteredModel.getFullName() == null) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "No model specified.");
```
ravivj-db, line 254: Better error message ?

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -0,0 +1,361 @@
...
+        }
+        throw e;
+      }
+    }
+  }
+
+  public void deleteRegisteredModel(
+      Session session, UUID schemaId, String registeredModelName, boolean force) {
+    if (force) {
+      // Remove all model versions
```
ravivj-db, line 349: Please mention TODO : here

server/src/test/java/io/unitycatalog/server/base/model/BaseModelCRUDTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+    // Get registered model
+    System.out.println("Testing get registered model..");
+    RegisteredModelInfo retrievedRmInfo = modelOperations.getRegisteredModel(MODEL_FULL_NAME);
+    assertThat(retrievedRmInfo).isEqualTo(rmInfo);
+
+    // Calling update model with nothing to update should throw an exception
+    System.out.println("Testing updating registered model with nothing to update..");
+    UpdateRegisteredModel emptyUpdateRegisteredModel = new UpdateRegisteredModel();
+    assertThatThrownBy(
```
ravivj-db, line 85: I know we have not done this before but maybe checking of specific error messages might be a good practice going forward.

examples/cli/src/main/java/io/unitycatalog/cli/ModelCli.java
```diff
@@ -0,0 +1,115 @@
...
+      default:
+        CliUtils.printEntityHelp(CliUtils.REGISTERED_MODEL);
+    }
+    CliUtils.postProcessAndPrintOutput(cmd, output, subCommand);
+  }
+
+  private static String createRegisteredModel(
+      RegisteredModelsApi registeredModelsApi, JSONObject json)
+      throws JsonProcessingException, ApiException {
+    CreateRegisteredModel createRegisteredModel;
```
ravivj-db, line 54: We can have more detailed CUJ's here involving model path and storing the actual model at the path (mostly a version) after obtaining temporary credentials for the same.

server/src/main/java/io/unitycatalog/server/persist/utils/FileUtils.java
```diff
@@ -47,6 +61,20 @@ public static String createTableDirectory(
...
   }
 
+  public static String createRegisteredModelDirectory(
+      String catalogName, String schemaName, String modelName) {
+    String absoluteUri =
+        getModelDirectoryURI(catalogName + "." + schemaName + ".models." + modelName);
+    return createDirectory(absoluteUri).toString();
+  }
+
+  public static String createModelVersionDirectory(
```
ravivj-db, line 71: Will each version's location also be stored along with the model version entity ?

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
kriscon-db: I have a follow on PR ready to go once this lands that adds this in.

server/src/main/java/io/unitycatalog/server/persist/utils/FileUtils.java
kriscon-db: Yes.  Every version has its own storage location which is computed off of the storage location of the registered model.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
kriscon-db: I guess we don't need the check if we're positive that situation could never occur, but if this does happen we would be throwing a less specific exception/message.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
kriscon-db: I guess it depends on how session works.  If multiple sessions can be opened and other transactions can land before we begin our transaction, then yeah, we should move it in the transaction.  I can move it under, thought I think i copied this pattern from one of the other assets (either tables or schemas).

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
kriscon-db: What would be a better message here?  `No three tier full name specified.`?

server/src/test/java/io/unitycatalog/server/base/model/BaseModelCRUDTest.java
kriscon-db: I don't think matching against error message is a good idea since it would make the tests very brittle to error message changes.  I think we do error code checking in the DB UC tests, which I can add in.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
ravivj-db: Sure, we can keep it. But we are anyways handling deleting children entities properly in delete.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
ravivj-db: What if a delete happens post we get the schema Id. It would be a safer option to include this inside. And follow the same logic elsewhere.

server/src/test/java/io/unitycatalog/server/base/model/BaseModelCRUDTest.java
ravivj-db: +1, we can do an error handling overhaul in another PR.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
kriscon-db: Yep.  I'll move it in and see if I can find where else i saw this pattern.

kriscon-db: Found it: https://github.com/unitycatalog/unitycatalog/blob/main/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java#L129-L133

Let's cut a separate PR that fixes it for tables.

ravivj-db: +1

ravivj-db (approved):
LGTM

examples/cli/src/main/java/io/unitycatalog/cli/ModelCli.java
kriscon-db: So in the MLflow client, there is the abstraction of `log_model`.  This would go through the full process of:

- Create the registered model if it doesnt exist
- Create a new model version under that registered model
- Request credentials
- Upload artifacts of the model to the storage location
- Finalize the model version

We can have something similar to that in the cli but for now, and for easier testing by other devs, I'm adding all of these low level calls.  

The MLflow client is being worked on separately that will do all of those operations out of the box.

PR #345, opened 2024-08-13
Fix credential support by disabling FileSystem cache

cloud-fan:
**Description of changes**

The credential support is implemented by setting hadoop conf for different table scans. This means we can't reuse FileSystem instance as it may mess up credentials. This PR explicit disables the cache.


build.sbt
```diff
@@ -394,6 +394,8 @@ lazy val spark = (project in file("connectors/spark"))
     ),
     Compile / unmanagedJars += (serverAndClientShaded / assembly).value,
     Test / unmanagedJars += (serverAndClientShaded / assembly).value,
+    // TODO: enable the java assertion after Delta 3.2.1 is released.
+    // Test / javaOptions += "-ea",
```
tdas, line 398: @creechy @alexreid-db can you make sure that this is merged in a separate PR to fix the infra, all projects need to have this set so that we dont silently ignore test failures. 

if there are tests that are currently failing, lets mark them as ignored, and then we will fix them one by one in follow up PRs like this one.

alexreid-db: Created https://github.com/unitycatalog/unitycatalog/pull/350 to address this

amaliujia (approved):

Issue #351, opened 2024-08-13
Downgrade some modules to Java 11 for better compatibility


creechy:
**Is your feature request related to a problem? Please describe.**

For better Spark compatibility, we should downgrade certain modules to Java 11 to decrease compatibility concerns until Spark 4 comes out with newer Java compatibility.

**Describe the solution you would like**

Downgrade the Spark Connector and the Clients modules to  Java 11

**Additional context**

PR #352, opened 2024-08-13
Downgrade the clients and Spark connector to Java 11 for better Spark compatibility

creechy:
**PR Checklist**

**Description of changes**

This changes the Java language version to 11 for the Clients module and the Spark Connector. See Issue https://github.com/unitycatalog/unitycatalog/issues/351

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,8 +24,6 @@ lazy val commonSettings = Seq(
   Compile / compile / javacOptions ++= Seq(
     "-Xlint:deprecation",
     "-Xlint:unchecked",
-    "-source", "17",
-    "-target", "17",
     "-g:source,lines,vars",
   ),
   libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


alexreid-db (approved):

build.sbt
```diff
@@ -178,6 +179,11 @@ lazy val server = (project in file("server"))
     javaOnlyReleaseSettings,
     javafmtCheckSettings,
     javaCheckstyleSettings(file("dev") / "checkstyle-config.xml"),
+    Compile / compile / javacOptions ++= Seq(
+      "--release", "17",
+      "-processor",
+      "lombok.launch.AnnotationProcessorHider$AnnotationProcessor"
```
tdas, line 185: what is this line for? can write that down as a comment?

vikrantpuppala: I think @ravivj-db had added this but from ChatGPT: This is essentially instructing the Java compiler to use Lombok's annotation processor (AnnotationProcessorHider$AnnotationProcessor) to process the annotations present in the code. This allows Lombok to generate the necessary boilerplate code (like getters, setters, etc.) during the compilation phase.

Typically, we don’t need to specify this option manually as it’s automatically handled by build tools like Maven or Gradle when you include Lombok as a dependency.

However, maybe sbt doesn't handle that hence this was needed.

vikrantpuppala (approved):

PR #371, opened 2024-08-22
Initial Model Version handler code and tests

kriscon-db:
**PR Checklist**

**Description of changes**

Added the model version handlers and tests.  Updated the API for some small errors.


etc/conf/server.log4j2.properties
```diff
@@ -10,7 +10,7 @@ appender.rollingFile.layout.pattern = %d{HH:mm:ss.SSS} [%t] %-5level %logger{36}
 
 appender.rollingFile.policies.type = Policies
 appender.rollingFile.policies.time.type = TimeBasedTriggeringPolicy
-appender.rollingFile.policies.time.interval = 1
+appender.rollingFile.policies.time.interval = 100000
```
kriscon-db, line 13: Whoops.  Change this back to 1.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -345,17 +358,426 @@ public void deleteRegisteredModel(String fullName, boolean force) {
...
     RegisteredModelInfoDAO registeredModelInfoDAO =
         getRegisteredModelDao(session, schemaId, registeredModelName);
     if (registeredModelInfoDAO == null) {
       throw new BaseException(
           ErrorCode.NOT_FOUND, "Registered model not found: " + registeredModelName);
     }
+    if (force) {
+      // Remove all model versions
+      List<ModelVersionInfoDAO> versionList =
+          getModelVersionsDao(session, registeredModelInfoDAO.getId(), "0", 100);
```
kriscon-db, line 370: Not sure how we should allow the `100` here to be parameterized.  This is purely a `keep pulling back versions to delete until we have deleted everything` operation and the number shouldn't really be user controlled, but you also don't want to pull back potentially 1000s of records.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -345,17 +358,426 @@ public void deleteRegisteredModel(String fullName, boolean force) {
...
+            .createdAt(createTime)
+            .updatedAt(createTime);
+    String registeredModelFullName = getRegisteredModelFullName(catalogName, schemaName, modelName);
+    LOGGER.info("Creating Registered Model: " + registeredModelFullName);
+
+    Transaction tx;
+    try (Session session = SESSION_FACTORY.openSession()) {
+      UUID catalogId = getCatalogId(session, catalogName);
+      UUID schemaId = getSchemaId(session, catalogName, schemaName);
+      tx = session.beginTransaction();
```
kriscon-db, line 478: Move the get*Id calls to be inside the transaction.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -345,17 +358,426 @@ public void deleteRegisteredModel(String fullName, boolean force) {
...
+    LOGGER.info("Deleting model version: " + fullName + "/" + version);
+    String[] parts = fullName.split("\\.");
+    if (parts.length != 3) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT, "Invalid registered model name: " + fullName);
+    }
+    String catalogName = parts[0];
+    String schemaName = parts[1];
+    String registeredModelName = parts[2];
+    RegisteredModelInfo rmInfo = getRegisteredModel(fullName);
```
kriscon-db, line 680: Maybe we don't need this check since we would throw in the transaction below.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -345,17 +358,426 @@ public void deleteRegisteredModel(String fullName, boolean force) {
...
+    if (modelVersionInfoDAO == null) {
+      throw new BaseException(
+          ErrorCode.NOT_FOUND, "Model version not found: " + fullName + "/" + version);
+    }
+    session.remove(modelVersionInfoDAO);
+  }
+
+  public ModelVersionInfo finalizeModelVersion(FinalizeModelVersion finalizeModelVersion) {
+    if (finalizeModelVersion.getFullName() == null) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "No model specified.");
```
kriscon-db, line 719: Make this error string match the registered model error string.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -345,17 +358,426 @@ public void deleteRegisteredModel(String fullName, boolean force) {
...
+        if (existingRegisteredModel.getMaxVersionNumber() == null
+            || existingRegisteredModel.getMaxVersionNumber() < 0) {
+          throw new BaseException(
+              ErrorCode.OUT_OF_RANGE,
+              "Registered model has invalid max model version: "
+                  + existingRegisteredModel.getMaxVersionNumber());
+        }
+        UUID modelId = existingRegisteredModel.getId();
+        Long version = existingRegisteredModel.getMaxVersionNumber() + 1;
+        String storageLocation =
```
kriscon-db, line 495: Will add a follow up commit that deals with storage location creation here and in registered models, as well as rollback of the creation on failure.

server/src/main/java/io/unitycatalog/server/persist/dao/ModelVersionInfoDAO.java
```diff
@@ -0,0 +1,101 @@
...
+    indexes = {
+      @Index(name = "idx_model_version", columnList = "registered_model_id,version"),
+    })
+// Lombok annotations
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+@SuperBuilder
+public class ModelVersionInfoDAO {
```
kriscon-db, line 25: Unfortunately, because model versions don't have `name` I can't leverage the IndentifiableDao and also means I had to implement a separate page token logic for model versions up in the ModelRepository.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
ravivj-db: Let's keep the behaviour same as other places where we are deleting child entities. Also can we move `100` to a local variable.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -345,17 +358,426 @@ public void deleteRegisteredModel(String fullName, boolean force) {
...
+    }
+  }
+
+  public String getNextPageToken(List<ModelVersionInfoDAO> entities, Optional<Integer> maxResults) {
+    if (entities == null
+        || entities.isEmpty()
+        || entities.size() < PagedListingHelper.getPageSize(maxResults)) {
+      return null;
+    }
+    // return the last entity name
```
ravivj-db, line 600: nit : // return last model version.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
ravivj-db: Can we move everything to the transaction. Also since this is the repository class we should deal in DAO's rather than Info objects.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -345,17 +358,426 @@ public void deleteRegisteredModel(String fullName, boolean force) {
...
+    try (Session session = SESSION_FACTORY.openSession()) {
+      String[] parts = RepositoryUtils.parseFullName(fullName);
+      String catalogName = parts[0];
+      String schemaName = parts[1];
+      String registeredModelName = parts[2];
+      tx = session.beginTransaction();
+      try {
+        // Get the registered model record from the database
+        UUID schemaId = getSchemaId(session, catalogName, schemaName);
+        RegisteredModelInfoDAO existingRegisteredModel =
```
ravivj-db, line 740: can we move getting the parent RegisteredModelDAO and subsequently ModelVersionDAO to a common function, it's repeated everywhere. The function can just return just the ModelVersionDAO.

server/src/main/java/io/unitycatalog/server/persist/dao/ModelVersionInfoDAO.java
```diff
@@ -0,0 +1,101 @@
...
+  @Id
+  @Column(name = "id")
+  private UUID id;
+
+  @Column(name = "registered_model_id")
+  private UUID registeredModelId;
```
ravivj-db, line 31: No foreign key reference to registered model ? Might help us in simplifying some cascaded fetch / delete operations. 
ref:
https://github.com/unitycatalog/unitycatalog/blob/main/server/src/main/java/io/unitycatalog/server/persist/dao/ColumnInfoDAO.java#L28
https://github.com/unitycatalog/unitycatalog/blob/main/server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java#L53

server/src/main/java/io/unitycatalog/server/persist/dao/RegisteredModelInfoDAO.java
```diff
@@ -44,6 +44,9 @@ public class RegisteredModelInfoDAO extends IdentifiableDAO {
   @Column(name = "url", length = 2048)
   private String url;
 
+  @Column(name = "max_version_number")
+  private Long maxVersionNumber;
```
ravivj-db, line 48: `currentVersion` ?

api/all.yaml
```diff
@@ -2014,6 +2026,13 @@ components:
     UpdateModelVersion:
       type: object
       properties:
+        full_name:
```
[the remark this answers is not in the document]
kriscon-db: Our public documentation says full_name and version: https://docs.databricks.com/api/workspace/modelversions/update

I do see that the protos in UC DB use full_name_arg and version_arg.  I guess let's update the documentation but let's try to normalize all of this.  I believe UC folks did not want the `_arg` naming.....this seem to have been missed though.

examples/cli/src/main/java/io/unitycatalog/cli/ModelVersionCli.java
```diff
@@ -0,0 +1,112 @@
...
+import io.unitycatalog.client.ApiClient;
+import io.unitycatalog.client.ApiException;
+import io.unitycatalog.client.api.ModelVersionsApi;
+import io.unitycatalog.client.model.CreateModelVersion;
+import io.unitycatalog.client.model.FinalizeModelVersion;
+import io.unitycatalog.client.model.UpdateModelVersion;
+import org.apache.commons.cli.CommandLine;
+import org.json.JSONObject;
+
+public class ModelVersionCli {
```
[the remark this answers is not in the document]
kriscon-db: While it is a sub entity, I'm not sure this makes a ton of sense since model versions are mostly reprsented as full class entities to users via the API and UI.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliParams.java
```diff
@@ -4,6 +4,9 @@ public enum CliParams {
   NAME("name", "The name of the entity.", "name"),
   SCHEMA_NAME("schema", "The name of the schema.", "schema_name"),
   CATALOG_NAME("catalog", "The name of the catalog.", "catalog_name"),
+  MODEL_NAME("model_name", "The name of the registered model.", "model_name"),
```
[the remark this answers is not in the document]
kriscon-db: The existing API uses model_name.  We could just translate it to name i guess at the server layer (not sure how confusing that will be to users though).

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -201,6 +205,14 @@ public UUID getSchemaId(Session session, String catalogName, String schemaName)
     return schemaInfo.getId();
   }
 
+  public UUID getCatalogId(Session session, String catalogName) {
```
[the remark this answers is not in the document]
kriscon-db: Agree.  Let me look into reworking this.

server/src/main/java/io/unitycatalog/server/persist/dao/ModelVersionInfoDAO.java
kriscon-db: So this isn't a strict FK relationship.   We *must* keep track of the registered model id on the model version in order to be able to list the model versions for a registered model.  DB UC also has this field defined.

server/src/main/java/io/unitycatalog/server/persist/dao/RegisteredModelInfoDAO.java
kriscon-db: current version would indicate what the latest version available is.  Users could delete the latest version but we would not want to reuse that version number.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
kriscon-db: It's hard coded to 1 in other places which doesn't seem like the best way to tackle a bulk delete.  I'll update to 1, but we should really look at increasing the max limit (or at least using the default max limit) of the list operations.

api/all.yaml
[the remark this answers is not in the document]
kriscon-db: Explained offline.  Path parameter names are UC server implementation dependent and so those fields do not need to match.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
```diff
@@ -31,11 +31,13 @@ public class CliUtils {
 
   public static final String FUNCTION = "function";
   public static final String REGISTERED_MODEL = "registered_model";
+  public static final String MODEL_VERSION = "model_version";
```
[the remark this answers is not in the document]
kriscon-db: So I talked with TD about this.  I think we want to leave the model version param as we have it because there are slight differences between the flags that need to be set/specified for models and model versions.

For example, on create, you only specify the run_id and source on a model version.  On a list, you only specify catalog and schema for registered models, but must include the model name for model versions.

There are enough differences that it probably would make the CLI slightly confusing to use if you had to reason about different required flags based on the presence of a --version flag.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliParams.java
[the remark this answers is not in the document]
kriscon-db: So looking at this further, we can do that mapping between name and model name in the server layer, but this will still not allow us to use the PagedListingHelper since it uses the name as the page token and the name of all model versions under a registered model would be the same.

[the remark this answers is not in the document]
ravivj-db: I wouldn't recommend taking this to the server side .. the wiring should happen on the CLI side itself while making the request to the server.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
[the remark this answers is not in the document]
ravivj-db: sure ... let's keep it completely separate.

server/src/main/java/io/unitycatalog/server/persist/dao/RegisteredModelInfoDAO.java
ravivj-db: cool .. can we remove the `number` suffix in the variable name  .. it's redundant.

kriscon-db: So actually....this is a little interesting....in MLflow, we refer to version number as the version (I think they are used interchangeably).  I can remove this if you feel strongly on it, but that is a name that we use versus just being redundant info.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliParams.java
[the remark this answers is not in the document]
kriscon-db: I see.  Let me see about updating the CLI to do the mapping then.

server/src/main/java/io/unitycatalog/server/persist/dao/RegisteredModelInfoDAO.java
ravivj-db: Not too strongly :) . Just a suggestion.

server/src/test/java/io/unitycatalog/server/base/model/BaseModelCRUDTest.java
```diff
@@ -269,5 +269,7 @@ public void testModelCRUD() throws ApiException {
         .isInstanceOf(Exception.class);
     assertThatThrownBy(() -> modelOperations.getRegisteredModel(MODEL_FULL_NAME))
         .isInstanceOf(Exception.class);
+
+    System.out.println("Completed model tests.");
```
ravivj-db, line 273: Do we really need this ? :).

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -470,6 +470,7 @@ public ModelVersionInfo getModelVersion(String fullName, long version) {
   }
 
   public ModelVersionInfo createModelVersion(CreateModelVersion createModelVersion) {
+    LOGGER.info("Create request: " + createModelVersion.toString());
```
ravivj-db, line 473: debug ?

ravivj-db (approved):
LGTM (besides the 2 minor comments).

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
kriscon-db: Yeah.  Removing.

Issue #381, opened 2024-08-27
Upgrade sbt to 1.10.1


jaceklaskowski:
It is possible that the very recent issue of mine while building the UC Docker images could be due to zinc-related issue(s) with one similar to what sbt has fixed in 1.10.1.

The issue I hit was as follows:

```
[error] (cli / Compile / compileIncremental) Failed to find name hashes for io.unitycatalog.cli.utils.CliUtils
```

And the zinc-related issue [sbt 1.10.1](https://github.com/sbt/sbt/releases/tag/v1.10.1) fixed is _"Fixes the hash code for empty files in the classpath cache"_.

Think it's enough to upgrade to sbt 1.10.1.

MrPowers:
Should we close this one for now?  Looks like there might be issues: https://github.com/unitycatalog/unitycatalog/pull/365#issuecomment-2299329952

PR #365, opened 2024-08-20
upgrade sbt to 1.10.1

sullis:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

upgrade to latest version of SBT

https://github.com/sbt/sbt/releases

PR #382, opened 2024-08-27
[BUILD] Upgrade sbt to 1.10.1

jaceklaskowski:
Fixes https://github.com/unitycatalog/unitycatalog/issues/381

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -33,4 +33,4 @@
 #  limitations under the License.
 #
 
-sbt.version=1.9.9
+sbt.version=1.10.1


tdas:
Hey @jaceklaskowski I am worried about upgrading SBT. I mentioned it here - https://github.com/unitycatalog/unitycatalog/pull/365#issuecomment-2299329952

jaceklaskowski:
I know about when I'm thinking about the upgrade, it's possible that what I'm hitting at with the docker images is that they use "unsupported by the current sbt version" JDK 22 (`eclipse-temurin:22-jdk-alpine`). I'll downgrade the JDK to 17 (to match the JDK used in the official release build) and let you know.

Thanks for looking into it @tdas ❤️

PR #383, opened 2024-08-27
cross build for scala 2.12 and 2.13

vikrantpuppala:
**PR Checklist**

**Description of changes**

Adding cross building for scala 2.12 and 2.13 in the spark connector


build.sbt
```diff
@@ -16,6 +16,9 @@ val artifactNamePrefix = "unitycatalog"
 lazy val javacRelease11 = Seq("--release", "11")
 lazy val javacRelease17 = Seq("--release", "17")
 
+val scala212 = "2.12.15"
```
jaceklaskowski, line 19: These `val`s should be `lazy`.

build.sbt
```diff
@@ -377,7 +380,8 @@ lazy val spark = (project in file("connectors/spark"))
   .dependsOn(client)
   .settings(
     name := s"$artifactNamePrefix-spark",
-    scalaVersion := "2.12.15",
+    scalaVersion := scala212,
+    crossScalaVersions := Seq(scala212, scala213),
```
jaceklaskowski, line 384: As per https://www.scala-sbt.org/1.x/docs/Cross-Build.html:

> Note: crossScalaVersions must be set to Nil on the root project to avoid double publishing.

ravivj-db:
would test cases also run post cross building ?

jaceklaskowski:
This little '+' (plus) before `jacoco` makes it x-build. It means that sbt will execute `jacoco` for each and every Scala version in `crossScalaVersions`.

In other words, since `jacoco` (indirectly) triggers `test`, the answer to "would test cases also run post cross building ?" would be Yes 😉

ravivj-db:
Thanks @jaceklaskowski for the detailed explanation :).

rameshchandra (approved):
LGTM

PR #384, opened 2024-08-27
Model version credential handler

kriscon-db:
**PR Checklist**

**Description of changes**

Added the model version credential handler.


server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
```diff
@@ -56,6 +56,11 @@ public static String[] parseFullName(String fullName) {
     return parts;
   }
 
+  public static String getAssetFullName(
+          String catalogName, String schemaName, String assetName) {
```
jaceklaskowski, line 60: Can you make it two-indented (four spaces) 🙏

jaceklaskowski (commented):
Could we add some tests?

server/src/main/java/io/unitycatalog/server/service/TemporaryModelVersionCredentialsService.java
```diff
@@ -0,0 +1,59 @@
...
+        String catalogName = generateTemporaryModelVersionCredential.getCatalogName();
+        String schemaName = generateTemporaryModelVersionCredential.getSchemaName();
+        String modelName = generateTemporaryModelVersionCredential.getModelName();
+        String fullName = RepositoryUtils.getAssetFullName(catalogName, schemaName, modelName);
+
+        ModelVersionInfo modelVersionInfo = MODEL_REPOSITORY.getModelVersion(fullName, modelVersion);
+        ModelVersionOperation requestedOperation = generateTemporaryModelVersionCredential.getOperation();
+        // Must enforce that write credentials are not passed back if the model has been finalized
+        if (modelVersionInfo.getStatus() != ModelVersionStatus.PENDING_REGISTRATION && requestedOperation == ModelVersionOperation.READ_WRITE_MODEL_VERSION) {
+            throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Cannot request read/write credentials on a finalized model version: " + fullName + "/" + modelVersion);
```
ravivj-db, line 47: We would need to handle FAILED_REGISTRATION model version status as well.

server/src/main/java/io/unitycatalog/server/service/TemporaryModelVersionCredentialsService.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+        return HttpResponse.ofJson(credentialOps.vendCredentialForModelVersion(modelVersionInfo, modelVersionOperationToPrivileges(generateTemporaryModelVersionCredential.getOperation())));
```
ravivj-db, line 49: can pass variable `requestedOperation` here.

server/src/main/java/io/unitycatalog/server/service/TemporaryModelVersionCredentialsService.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    private Set<CredentialContext.Privilege> modelVersionOperationToPrivileges(ModelVersionOperation modelVersionOperation) {
+        return switch (modelVersionOperation) {
+            case READ_MODEL_VERSION -> Set.of(SELECT);
+            case READ_WRITE_MODEL_VERSION -> Set.of(SELECT, UPDATE);
+            case UNKNOWN_MODEL_VERSION_OPERATION -> Collections.emptySet();
```
ravivj-db, line 56: shouldn't we throw error in this case ?

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
```diff
@@ -50,6 +44,20 @@ public GenerateTemporaryTableCredentialResponse vendCredentialForTable(TableInfo
     return vendCredential(credentialContext).toTableCredentialResponse();
   }
 
+  public GenerateTemporaryModelVersionCredentialResponse vendCredentialForModelVersion(ModelVersionInfo modelVersionInfo, Set<CredentialContext.Privilege> privileges) {
+    String mvStorageLocation = modelVersionInfo.getStorageLocation();
+    if (mvStorageLocation == null || mvStorageLocation.isEmpty()) {
+      throw new BaseException(ErrorCode.FAILED_PRECONDITION, "Model version storage location not found.");
+    }
+
+    URI storageLocationUri = URI.create(mvStorageLocation);
```
ravivj-db, line 53: Will this work if mvStorageLocation is a relativePath ? we do set relative paths in our test cases/sample database.

ravivj-db:
+1 to test cases, you can use mock cloud libraries.

server/src/main/java/io/unitycatalog/server/service/TemporaryModelVersionCredentialsService.java
kriscon-db: cc @alexreid-db on this as we would need to fix it everywhere if the answer is yes.

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
kriscon-db: I don't think I currently set relative paths for this in models or model versions.

kriscon-db:
> +1 to test cases, you can use mock cloud libraries.

@alexreid-db do you have an example test around this that I could look at?

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
ravivj-db: We are setting relative paths in the quick start h2 db for external tables so that it resolves correctly in everyone's setup to `etc/data` relative to the project root and quick start commands work out of box. That's one use case.

ajreid21: think the bigger question would be why the client (or someone) is asking for temporary credentials for local filesystem/paths

server/src/main/java/io/unitycatalog/server/service/TemporaryModelVersionCredentialsService.java
ajreid21: I'm confused when a client would ever send this in the request, but sure we can throw error here. Probably should be `default -> throw new ...`

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
ravivj-db: There is a precedence in CLI : https://github.com/ravivj-db/unitycatalog/blob/main/examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java#L172. But I agree, the client can check prior to sending the credential request for path scheme(if path is available prior). 

However, in this particular case, we first fetch the ModelVersionInfo from model repository while serving the temporary credential request and use it's storageLocation field. If that's relative (which sometimes happens for pre-stored entities) `URI.create(mvStorageLocation)` will fail. In that case something like 
`Path path = Paths.get(mvStorageLocation);
URI uri = path.toUri();` 
will make more sense. 

We can also embed this conversion in `getModelVersionInfo` function itself during conversion from `DAO` to `Info` object.

server/src/main/java/io/unitycatalog/server/service/TemporaryModelVersionCredentialsService.java
ravivj-db: What if I a client is deliberately trying to be malicious. I was just worried about the case when`UnknownOperation` can by mistake translate to credentials which have the most broad access.

alexreid-db: I'm more commenting on why this value is even a part of api / spec if we are just going to always throw error if sent? (just don't include it as an enum value). If a client sends an value that isn't in the spec, it won't deserialize and the server will throw an exception on its own...

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
alexreid-db: I think CLI needs updating and possibly not ask for creds if its a local path (it also needs to update since it's expecting AwsCredentials only...). If anything in the credentials space, should we throw error or return empty creds if we see a local file path (whether we use Paths to then create the URI or not)?

server/src/main/java/io/unitycatalog/server/service/TemporaryModelVersionCredentialsService.java
alexreid-db: The other thing to mention is that once the access control story makes it into the codebase, the credential vending would always check what privileges the subject/user/principal and check against what is being asked for and never elevate privileges.

Potentially this unknown value would be more used for just having the server send back credentials for privileges that the subject/user/principal has (i.e. the client doesn't tell the server what operations it wants and the server just gives what they have access to), but no idea if that is the actual purpose. 

How the iceberg api works is that it always gives you credentials with privileges you have for a table (instead of the client asking for certain privileges), which is a little more straightforward/simpler.

alexreid-db: but all that being said, I'm fine with just throwing an error here...

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
ravivj-db: https://github.com/unitycatalog/unitycatalog/blob/main/server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java#L68 .. We are returning empty credentials here. Let's check once with UC and keep the behaviour consistent.

kriscon-db: So I don't think it makes sense to put relative paths in storage location.  I believe we treat the storage location as a URI in DB UC.  I don't think you can create a relative path file URI.

kriscon-db:
After internal discussions around the testing, I'm adding the error case tests that @ramesh-db suggested.  We will need to wait on the integration test framework before we can test the actual cloud credential vending pathways.

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
ravivj-db: It was just a hack for making the quickstart work in everyone's own environment. We can discuss it later when we are putting together the quickstart commands for models.

server/src/main/java/io/unitycatalog/server/service/TemporaryModelVersionCredentialsService.java
```diff
@@ -31,29 +31,41 @@ public TemporaryModelVersionCredentialsService(CredentialOperations credentialOp
...
-        if (modelVersionInfo.getStatus() != ModelVersionStatus.PENDING_REGISTRATION && requestedOperation == ModelVersionOperation.READ_WRITE_MODEL_VERSION) {
-            throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Cannot request read/write credentials on a finalized model version: " + fullName + "/" + modelVersion);
+        String storageLocation = modelVersionInfo.getStorageLocation();
+        if (storageLocation.toLowerCase().startsWith("file")) {
+            throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Cannot request credentials on a model version with a file based storage location: " + fullName + "/" + modelVersion);
         }
-        return HttpResponse.ofJson(credentialOps.vendCredentialForModelVersion(modelVersionInfo, modelVersionOperationToPrivileges(generateTemporaryModelVersionCredential.getOperation())));
+        ModelVersionOperation requestedOperation = generateTemporaryModelVersionCredentials.getOperation();
+        // Must enforce that the status of the model version matches the requested credential type.
+        if (modelVersionInfo.getStatus() == ModelVersionStatus.FAILED_REGISTRATION) {
```
ravivj-db, line 50: nit : we could have combined these two errors using || and converting status to string :)

server/src/main/java/io/unitycatalog/server/service/TemporaryModelVersionCredentialsService.java
```diff
@@ -31,29 +31,41 @@ public TemporaryModelVersionCredentialsService(CredentialOperations credentialOp
...
+            throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Cannot request credentials on a model version that has failed registration: " + fullName + "/" + modelVersion);
+        }
+        if (modelVersionInfo.getStatus() == ModelVersionStatus.MODEL_VERSION_STATUS_UNKNOWN) {
+            throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Cannot request credentials on a model version with an unknown status: " + fullName + "/" + modelVersion);
+        }
+        if ((modelVersionInfo.getStatus() != ModelVersionStatus.PENDING_REGISTRATION &&
```
ravivj-db, line 56: The if condition results to model version = finalized (model version != pending registration) and requestedOperation = Read_Write 
but the error message reads otherwise.

ravivj-db (approved):
LGTM

PR #387, opened 2024-08-28
Add a user-database and constrain access to known users.

creechy:
**PR Checklist**

**Description of changes**

This enhancement adds a user database to the platform, a control API for user management, CLI commands to manage users, and when authentication is enabled, will constrain access to only the users in the user database.

Related Issue https://github.com/unitycatalog/unitycatalog/issues/229

The user record is simple for this first implementation
- id - unique identifier for the user
- name - The user's full name
- email - The user's primary email address
- picture_url - A link to an avatar for the user
- external_id - An identifier for external identity providers to track the user
- created_at - the time the record was created
- updated_at - the last time the record was updated

There is the beginning of a Control API which is meant to house operations that are outside of the core of the Unity Catalog specification. 

The Control API initially provides endpoints to do user management. The endpoints are SCIM 2.0 compatible, as such they should allow integration with third party identity providers for syncing users between systems.

The CLI enhancements add commands to add, update and remove users, see command line help for details, but as an example, to add a user
```
uc user create --name "Iggy Pop" --email iggy@thestooges.com 
```

Finally, when the recently added OAuth authorization functionality is enabled, users must exist in the user database in order to gain access to the server operations.


xx-db (approved):

api/all.yaml
```diff
@@ -2082,6 +2085,75 @@ components:
       required:
         - full_name
         - version
+    CreateUser:
```
rameshchandra, line 2088: shouldn't these definitions be in `control.yaml`?

xx-db: The reason I put them in control.yaml is to keep compiling the server module relatively simple. Right now only the Client module uses both all.yaml and control.yaml. If we want to break CreateUser out of all.yaml, we would need to add a separte build task dependency for server since the generated api code can't be in the same directory.

api/all.yaml
```diff
@@ -29,6 +29,9 @@ tags:
     description: |
       Model versions contain all artifact files associated with an individual machine learning model. They are associated with a single parent model and have an integer identifier, indicating the version number of the model.
       Model versions are referenced via model URI's using the Unity Catalog three-level namespace of the registered model and the version identifier.
+  - name: Permissions
```
rameshchandra, line 32: nit: Is this for this PR or should it be moved to the access conrol PR?

api/control.yaml
```diff
@@ -0,0 +1,192 @@
...
+        - Users
+      operationId: createUser
+      summary: Create a user
+      description: |
+        Creates a new user.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/UserResource'
```
rameshchandra, line 23: shouldn't this be the `CreateUser`?

api/control.yaml
```diff
@@ -0,0 +1,192 @@
...
+        - Users
+      operationId: updateUser
+      summary: Update a user
+      description: |
+        Updates the user that matches the supplied id.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/UserResource'
```
rameshchandra, line 101: Nit: similar to create, should this be `UpdateUser`?

api/all.yaml
creechy: yep. good catch. missed that when I extracted the user database stuff from the access control work.

api/all.yaml
creechy: @xx-db is going to see if he can move this to `control.yaml`. This is kind of a special case because we've set up the repository similar to other entities, which pass in things like `CreateFoo`, even though for users the over-the-wire entity is a SCIM `UserResource`. Another option would be to get rid of these objects and just pass in the SCIM `UserResource` directly to the repository.

creechy: @xx-db refactored this into `control.yaml` 👍

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -100,6 +108,7 @@ private void addServices(ServerBuilder sb) {
         new TemporaryModelVersionCredentialsService(credentialOperations);
     sb.service("/", (ctx, req) -> HttpResponse.of("Hello, Unity Catalog!"))
         .annotatedService(controlPath + "auth", authService, unityConverterFunction)
+        .annotatedService(controlPath + "scim2/Users", Scim2UserService)
```
rameshchandra, line 111: Without the `unityConverterFunction`, doesn't the SCIM API fail on unknown properties? Is that what we want?

server/src/main/java/io/unitycatalog/server/persist/UserRepository.java
```diff
@@ -0,0 +1,191 @@
...
+  private static final PagedListingHelper<UserDAO> LISTING_HELPER =
+      new PagedListingHelper<>(UserDAO.class);
+
+  private UserRepository() {}
+
+  public static UserRepository getInstance() {
+    return INSTANCE;
+  }
+
+  public User createUser(CreateUser createUser) {
```
rameshchandra, line 35: According to line 23 in `control.yaml`, isn't the input to `createUser` a `UserResource`, which is different from `CreateUser`?

server/src/main/java/io/unitycatalog/server/persist/UserRepository.java
```diff
@@ -0,0 +1,191 @@
...
+            .name(createUser.getName())
+            .email(createUser.getEmail())
+            .externalId(createUser.getExternalId())
+            .state(User.StateEnum.ENABLED)
+            .createdAt(System.currentTimeMillis());
+
+    try (Session session = SESSION_FACTORY.openSession()) {
+      Transaction tx = session.beginTransaction();
+      try {
+        if (getUserByEmail(session, user.getEmail()) != null) {
```
rameshchandra, line 48: Should external id also be unique?

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
```diff
@@ -0,0 +1,204 @@
...
+            .filter(Email::getPrimary)
+            .findFirst()
+            .orElseThrow(
+                () ->
+                    new Scim2RuntimeException(
+                        new PreconditionFailedException("User does not have a primary email.")));
+
+    try {
+      User user = USER_REPOSITORY.getUserByEmail(primaryEmail.getValue());
+      return HttpResponse.ofJson(asUserResource(user));
```
rameshchandra, line 85: Does the SCIM create API do nothing if the user already exists?

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
```diff
@@ -0,0 +1,204 @@
...
+        if (userResource.getPhotos() != null && !userResource.getPhotos().isEmpty()) {
+          pictureUrl = userResource.getPhotos().get(0).getValue().toString();
+        }
+        User user =
+            USER_REPOSITORY.createUser(
+                new CreateUser()
+                    .name(userResource.getDisplayName())
+                    .email(primaryEmail.getValue())
+                    .externalId(userResource.getExternalId())
+                    .pictureUrl(pictureUrl));
```
rameshchandra, line 98: Are the `meta` and `active` fields of the `UserResource` that could be sent as part of a create user request ignored?

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
```diff
@@ -0,0 +1,204 @@
...
+      return HttpResponse.ofJson(asUserResource(user));
+    } catch (BaseException e) {
+      if (e.getErrorCode() == ErrorCode.NOT_FOUND) {
+        String pictureUrl = "";
+        if (userResource.getPhotos() != null && !userResource.getPhotos().isEmpty()) {
+          pictureUrl = userResource.getPhotos().get(0).getValue().toString();
+        }
+        User user =
+            USER_REPOSITORY.createUser(
+                new CreateUser()
```
rameshchandra, line 94: `CreateUser` is only used within the server code to pass the user info to the `UserRepository`. Does it need to be defined in `control.yaml` public spec? Can't it just be a class defined on the server side?

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
```diff
@@ -0,0 +1,204 @@
...
+    return HttpResponse.ofJson(asUserResource(USER_REPOSITORY.getUserByEmail(sub.asString())));
+  }
+
+  @Get("/{id}")
+  public HttpResponse getUser(@Param("id") String id) {
+    return HttpResponse.ofJson(asUserResource(USER_REPOSITORY.getUser(id)));
+  }
+
+  @Put("/{id}")
+  public HttpResponse updateUser(@Param("id") String id, UserResource userResource) {
```
rameshchandra, line 120: Nit: the update code does the read and modify in different DB transactions which can cause lost updates for concurrent updates.

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
```diff
@@ -0,0 +1,204 @@
...
+      newName = user.getDisplayName();
+    }
+
+    String externalId = userResource.getExternalId();
+    if (externalId == null) {
+      externalId = user.getExternalId();
+    }
+
+    UpdateUser updateUser =
+        new UpdateUser().newName(newName).email(primaryEmail.getValue()).externalId(externalId);
```
rameshchandra, line 143: Similar to `CreateUser`, `UpdateUser` is also used only within the server code, so is it necessary to define it in the yaml spec?

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
```diff
@@ -0,0 +1,204 @@
...
+
+    return HttpResponse.ofJson(asUserResource(USER_REPOSITORY.updateUser(id, updateUser)));
+  }
+
+  @Delete("/{id}")
+  public HttpResponse deleteUser(@Param("id") String id) {
+    User user = USER_REPOSITORY.getUser(id);
```
rameshchandra, line 150: Can the current user delete themselves?

api/control.yaml
```diff
@@ -0,0 +1,261 @@
...
+        - name: startIndex
+          in: query
+          description: Specifies the index of the first result. First item is number 1.
+          schema:
+            type: integer
+            format: int32
+          required: false
+        - name: count
+          in: query
+          description: Desired number of results per page.
```
rameshchandra, line 50:
```suggestion
          description: Desired number of results per page. If no count is provided, it defaults to 50.
```

server/src/main/java/io/unitycatalog/server/persist/UserRepository.java
```diff
@@ -0,0 +1,191 @@
...
+  public List<User> listUsers() {
+    return listUsers(false);
+  }
+
+  public List<User> listUsers(boolean includeDisabled) {
+    try (Session session = SESSION_FACTORY.openSession()) {
+      session.setDefaultReadOnly(true);
+      Transaction tx = session.beginTransaction();
+      try {
+        List<UserDAO> userDAOs =
```
rameshchandra, line 71: This is unpaginated (even though the API itself has a count), which it can be expensive on the DB. For example, if there are 100K total users and the count is 10, this would fetch all the 100K users from the DB and throw away all except 10.

Understandably, the `filter` parameter in the API complicates pagination, but there are ways to implement it that is closer to O(count) and not O(total number of users)  (e.g., by paging it from the DB, with potentially multiple small DB queries).

This doesn't need to be fixed for now, but can you create a github issue to fix it in the future?

rameshchandra: Actually, the `PagedListingHelper` defaults to a max count of 100 if `maxResultsOpt` is not present. So, I think the `listUsers` API can return fewer users than requested and isn't correct.

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
rameshchandra: Is this because deleting a user only sets the user to DISABLED and doesn't delete the DB entry? If so, shouldn't re-creating the user verify that the user's input values match and enable the user?

rameshchandra (approved):
LGTM.

Added a few comments. Please address them before merging.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
creechy: We should probably make it do the same thing. Actually there's some other special serialization for SCIM, so I'll probably make a different converter function.

server/src/main/java/io/unitycatalog/server/persist/UserRepository.java
creechy: Yeah that wouldn't hurt.

server/src/main/java/io/unitycatalog/server/persist/UserRepository.java
creechy: Most of the Databricks paginated APIs say they can return fewer users than requested, so its consistent with that. Even so, this is a little trickier for SCIM since its "pagination" works differently. We'll have to paginate internally until we get to the requested SCIM `startIndex` and then gather results until we get enough elements.

creechy: Ok this has been reworked to get a clock from `PagedListingHelper` and skip and return max as appropriate.

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
creechy: I think this is incorrect behavior (@xx-db) looks lke it should really return a 409 Conflict - https://www.rfc-editor.org/rfc/rfc7644#section-3.3 
> If the service provider determines that the creation of the requested resource conflicts with existing resources (e.g., a "User" resource with a duplicate "userName"), the service provider MUST return HTTP status code 409

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
creechy: we should process `active`, even though that would be strange to create a disabled user, but I suppose a possibility.  `meta` we do ignore.

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
creechy: Yeah it could be just internal, this was just in keeping with the current pattern used elsewhere. We'll take a look.

creechy: cc: @xx-db

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
creechy: cc: @xx-db

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
creechy: Done.

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
creechy: Done.

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
creechy: Done.

server/src/main/java/io/unitycatalog/server/persist/UserRepository.java
creechy: Done.

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
creechy: Done.

server/src/main/java/io/unitycatalog/server/service/Scim2UserService.java
creechy: the repository UpdateUser was doing a bunch of this already so I simplified the code. Done.

api/control.yaml
```diff
@@ -0,0 +1,228 @@
...
+        '200':
+          description: The user was successfully retrieved.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/UserResource'
+
+components:
+  schemas:
+    User:
```
rameshchandra, line 140: Where is this resource used?

PR #392, opened 2024-08-29
[DOCKER] UC server container displays welcome message

jaceklaskowski:
This PR changes how the UC server docker image is built (uses the faster `sbt server/package`) and copies the jars required to start it (to `jars` directory).

After https://github.com/unitycatalog/unitycatalog/commit/4b7aaa319bfad76dde4d30c9fdd5128895bf98d1, I thought I'd work on changes that would make the behaviour of the UC server when started consistent regardless whether it's in or outside Docker container (namely the welcome message at the start and no warnings).

A Dockerized UC runs in the foreground and containers are removed once stopped.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,21 +14,19 @@ ARG unitycatalog_user_basedir="${unitycatalog_home}/${unitycatalog_user_home}"
 # and will cause the process to fail when attempting to compile
 # and generate the Uber-Jar. Therefore it is important to choose
 # a size large enough for the compiler to run. 
-ARG sbt_args="-J-Xmx5G"
+ARG sbt_args="-J-Xmx2G"
 # FIXME Pass it from the outside
 ARG unitycatalog_version="0.2.0-SNAPSHOT"
 ARG server_assembly_jar="${unitycatalog_repo}/${unitycatalog_jar}/unitycatalog-server-assembly-${unitycatalog_version}.jar"
+ARG jars_directory="server/target/jars"
 
-# ###### STAGE 1 ###### #
-# Building the Uber-Jar #
-#########################
-
-FROM eclipse-temurin:22-jdk-alpine AS assemble_server
+FROM eclipse-temurin:22-jdk-alpine AS package_server
 
 ARG unitycatalog_repo
 ARG sbt_args
+ARG jars_directory
 
-# Install any required packages
+# Install required packages
 RUN <<EOF
     set -ex;
     apk update;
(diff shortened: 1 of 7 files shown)


jaceklaskowski:
How you like it too, @dennyglee 🙏

jaceklaskowski:
@tdas @MrPowers Mind if I asked you to review it (and possibly accept)?

Dockerfile
```diff
@@ -115,11 +108,12 @@ VOLUME  "${unitycatalog_home}"
 
 WORKDIR "$unitycatalog_home"
 
-# Copy the Uber-Jar from the previous stage
-COPY --from=assemble_server "${server_assembly_jar}" "${unitycatalog_jars}/"
+COPY --from=package_server "${server_jar}" "${unitycatalog_jars}/"
```
dennyglee, line 111: It appears the `${server_jar}` isn't being used - did you mean to use `${jars_directory}` which is included in line 116?

docker/README.md
```diff
@@ -108,6 +104,8 @@ unitycatalog-cli   0.2.0-SNAPSHOT   52502d16934f   About a minute ago   1.48GB
 
 > [!NOTE]
 > In case you run into the following build exception, restart the CLI docker image build.
+> This is likely due to Java 22 used to build the image (and execute the sbt build).
```
dennyglee, line 107: Minor nit, but why call out Java 22 in this scenario since it isn't supported?

dennyglee (changes requested):
Thanks @jaceklaskowski  - this is really streamlined and I really like it!  Please note a couple call outs.  Thanks!

docker/README.md
jaceklaskowski: That's what this image has always been built with, so I accepted its ups and downs. Since you brought this up, I will downgrade to the highest supported version Java 17.

Dockerfile
jaceklaskowski: It is used indirectly when the server is started up at line 129 using `jars/*`.

This "feature" makes the build `assembly`-less and more matching what the regular non-docker build gives.

docker/README.md
jaceklaskowski: It should be good now. I had to use JDK 17 (not JRE) for sbt that gives a slightly larger image, but we're "fully-compliant" ❤️

Dockerfile
dennyglee: Oh got it!

dennyglee (approved):
Thanks a bunch @jaceklaskowski

docker/README.md
dennyglee: Merci beaucoup!

jalotra:
getting this now : 
```
 1 warning found (use docker --debug to expand):
 - UndefinedVar: Usage of undefined variable '$server_jar' (line 111)
```


@jaceklaskowski can we check what this arg var should be : https://github.com/unitycatalog/unitycatalog/blame/main/Dockerfile#L111

PR #399, opened 2024-09-01
[SPARK] Support gcs reads using vended credentials

bijanhoule:
**PR Checklist**

**Description of changes**

Updates the spark connector to support vended credentials for google cloud storage buckets.  

We use the `com.google.cloud.hadoop.util.AccessTokenProvider` interface to pass the vended credential properties stored in the hadoop conf -- see [google's gcs-connector documentation](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/7257fb4de55d83f14d4820fe00672b732d959987/gcs/CONFIGURATION.md#authentication) for more details. 

Example spark conf (assumes that the [gcs connector jar](https://github.com/GoogleCloudDataproc/hadoop-connectors/releases/download/v3.0.2/gcs-connector-3.0.2-shaded.jar) is included in the spark jars/packages as well):
```
spark.hadoop.fs.AbstractFileSystem.gs.impl 	com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS
spark.hadoop.fs.gs.auth.type 			ACCESS_TOKEN_PROVIDER
spark.hadoop.fs.gs.auth.access.token.provider 	io.unitycatalog.connectors.spark.GcsVendedTokenProvider
```

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/GcsVendedTokenProvider.java
+++ b/connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/GcsVendedTokenProvider.java
@@ -1,0 +1,36 @@
+package io.unitycatalog.connectors.spark;
+
+import com.google.cloud.hadoop.util.AccessTokenProvider;
+import org.apache.hadoop.conf.Configuration;
+
+import java.io.IOException;
+import java.time.Instant;
+
+public class GcsVendedTokenProvider implements AccessTokenProvider {
+    protected static final String ACCESS_TOKEN_KEY = "fs.gs.auth.access.token.credential";
+    protected static final String ACCESS_TOKEN_EXPIRATION_KEY = "fs.gs.auth.access.token.expiration";
+    private Configuration conf;
+
+    @Override
+    public AccessToken getAccessToken() {
+        return new AccessToken(
+                conf.get(ACCESS_TOKEN_KEY),
+                Instant.ofEpochMilli(Long.parseLong(conf.get(ACCESS_TOKEN_EXPIRATION_KEY)))
+        );
+    }
+
+    @Override
+    public void refresh() throws IOException {
+        throw new IOException("Temporary token, not refreshable");
+    }
+
+    @Override
+    public void setConf(Configuration configuration) {
+        conf = configuration;
+    }
+
+    @Override
+    public Configuration getConf() {
+        return conf;
+    }
+}
--- a/connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
@@ -148,22 +148,28 @@ private class UCProxy extends TableCatalog with SupportsNamespaces {
     }.toArray
     val uri = CatalogUtils.stringToURI(t.getStorageLocation)
     val tableId = t.getTableId
-    val credential: AwsCredentials = temporaryTableCredentialsApi
+    val temporaryCredentials = temporaryTableCredentialsApi
       .generateTemporaryTableCredentials(
         // TODO: at this time, we don't know if the table will be read or written. We should get
         //       credential in a later phase.
         new GenerateTemporaryTableCredential().tableId(tableId).operation(TableOperation.READ_WRITE)
       )
-      .getAwsTempCredentials
     val extraSerdeProps = if (uri.getScheme == "s3") {
+      val awsCredentials = temporaryCredentials.getAwsTempCredentials
       Map(
         // TODO: how to support s3:// properly?
-        "fs.s3a.access.key" -> credential.getAccessKeyId,
-        "fs.s3a.secret.key" -> credential.getSecretAccessKey,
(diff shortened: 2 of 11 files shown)


connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/GcsVendedTokenProvider.java
```diff
@@ -0,0 +1,36 @@
+package io.unitycatalog.connectors.spark;
+
+import com.google.cloud.hadoop.util.AccessTokenProvider;
+import org.apache.hadoop.conf.Configuration;
+
+import java.io.IOException;
+import java.time.Instant;
+
+public class GcsVendedTokenProvider implements AccessTokenProvider {
```
cloud-fan, line 9: We created this class only for using the two static constants?

bijanhoule: Yeah, this was the only way I was able to get those creds flowing through (if there's a simpler way, LMK!).  Based on the [gcs-connector](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/CONFIGURATION.md#user-credentials) docs, authentication can be provided by a few mechanisms:

```
fs.gs.auth.type (default: COMPUTE_ENGINE)

What type of authentication mechanism to use for Google Cloud Storage access.

Valid values:

    ACCESS_TOKEN_PROVIDER - configures AccessTokenProvider authentication
    APPLICATION_DEFAULT - configures [Application Default Credentials](https://javadoc.io/doc/com.google.auth/google-auth-library-oauth2-http/latest/com/google/auth/oauth2/GoogleCredentials.html) authentication
    COMPUTE_ENGINE - configures Google Compute Engine service account authentication
    SERVICE_ACCOUNT_JSON_KEYFILE - configures JSON keyfile service account authentication
    UNAUTHENTICATED - configures unauthenticated access
    USER_CREDENTIALS - configure [user credentials](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/CONFIGURATION.md#user-credentials)
```

`ACCESS_TOKEN_PROVIDER` is the only one we can really customize -- all the other options look like they expect either json keyfile credentials, metadata server, or an id/secret pair for initiating an oauth2 flow.  

That requires a class that implements the `AccessTokenProvider` interface -- this is a simple implementation of that interface that just grabs the required values from conf.  Those values should get set after the `generateTemporaryTableCredentials` call.

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
```diff
@@ -148,23 +148,29 @@ private class UCProxy extends TableCatalog with SupportsNamespaces {
...
+        "fs.s3a.session.token" -> awsCredentials.getSessionToken,
         "fs.s3a.path.style.access" -> "true",
         "fs.s3.impl.disable.cache" -> "true",
         "fs.s3a.impl.disable.cache" -> "true"
       )
+    } else if (uri.getScheme == "gs") {
+      val gcsCredentials = temporaryCredentials.getGcpOauthToken
+      Map(
+        GcsVendedTokenProvider.ACCESS_TOKEN_KEY -> gcsCredentials.getOauthToken,
+        GcsVendedTokenProvider.ACCESS_TOKEN_EXPIRATION_KEY -> temporaryCredentials.getExpirationTime.toString
```
cloud-fan, line 172: Sorry I didn't make the question clear. Shall we just hardcode the two string constants here instead of creating `GcsVendedTokenProvider`? Unless we plan to use `GcsVendedTokenProvider` somewhere else.

cloud-fan: Or we expect end users to use `GcsVendedTokenProvider` directly?

cloud-fan: cc @bijanhoule

cloud-fan:
Shall we update `CredentialTestFileSystem` to test gcs credentials?

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
bijanhoule: We'd expect users to use `GcsVendedTokenProvider` in spark config, e.g.:

```
spark.hadoop.fs.AbstractFileSystem.gs.impl 	com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS
spark.hadoop.fs.gs.auth.type 			ACCESS_TOKEN_PROVIDER
spark.hadoop.fs.gs.auth.access.token.provider 	io.unitycatalog.connectors.spark.GcsVendedTokenProvider
```

I just want those constants so that we can easily see where they're referenced

bijanhoule:
> Shall we update `CredentialTestFileSystem` to test gcs credentials?

Added (see 83a3b0bbcb8c5dba701ffcf6dc3188ee3e1495d9)

cloud-fan (approved):

PR #400, opened 2024-09-02
Fix storage location handling

kriscon-db:
**PR Checklist**

**Description of changes**

Fix storage location handling for models.


server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  private UriUtils() {}
+
+  // Model specific storage root handlers and convenience methods
+  private static String getModelStorageRoot() {
+    String modelStorageRoot = properties.getProperty("storage-root.models");
+    if (modelStorageRoot == null) {
+      // If the model storage root is empty, use the CWD
+      modelStorageRoot = System.getProperty("user.dir");
```
kriscon-db, line 48: Also fine using https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createTempDirectory-java.lang.String-java.nio.file.attribute.FileAttribute...-
instead of the CWD

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  public static String getModelVersionStorageLocation(
+      String catalogId, String schemaId, String modelId, String versionId) {
+    return getModelDirectoryURI(
+        catalogId + "." + schemaId + ".models." + modelId + ".versions." + versionId);
+  }
+
+  public static String createStorageLocationPath(String uri) {
+    return updateDirectoryFromUri(uri, Operation.CREATE, Optional.empty()).toString();
```
kriscon-db, line 79: We'll need to thread through the credentials when this call is made.

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -0,0 +1,245 @@
...
+        listing
+            .getObjectSummaries()
+            .forEach(
+                object -> {
+                  s3Client.deleteObject(bucketName, object.getKey());
+                });
+        req.setMarker(listing.getNextMarker());
+      } while (listing.isTruncated());
+    }
+    return parsedUri;
```
ravivj-db, line 224: Note to self: Difference from File Utils: Returning `parsedUri` instead of constructing it again for both operations.

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -0,0 +1,245 @@
...
+      if (testURI.getScheme() != null && testURI.getPath() != null) {
+        return true;
+      }
+      return false;
+    } catch (URISyntaxException e) {
+      return false;
+    }
+  }
+
+  private static URI updateLocalDirectory(URI parsedUri, Operation op) throws IOException {
```
ravivj-db, line 135: Note to self : Difference from FileUtils : In FileUtils the input is `Path` object for both CREATE AND DELETE and the functions exist independently. 
In create we are returning `Path.toUri()` which is not required here as we already have the URI as input. 
Delete is not returning anything.

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -0,0 +1,245 @@
...
+              + parsedUri.toString()
+              + ": "
+              + e.getMessage());
+    }
+    throw new BaseException(
+        ErrorCode.INVALID_ARGUMENT, "Unknown scheme detected: " + parsedUri.getScheme());
+  }
+
+  private static boolean isValidURI(String uri) {
+    try {
```
ravivj-db, line 124: one thing missing here is we are not checking for malicious paths for example file://../../b which can lead to us giving clients access to storage outside storage_root. 

`URI normalized = uri.normalize();
    if (!normalized.getPath().startsWith(uri.getPath())) {
      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Normalization failed: " + uri.getPath());
    }`

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -0,0 +1,245 @@
...
+
+  // Model specific storage root handlers and convenience methods
+  private static String getModelStorageRoot() {
+    String modelStorageRoot = properties.getProperty("storage-root.models");
+    if (modelStorageRoot == null) {
+      // If the model storage root is empty, use the CWD
+      modelStorageRoot = System.getProperty("user.dir");
+    }
+    // If the model storage root is not a valid URI, make it one
+    if (!isValidURI(modelStorageRoot)) {
```
ravivj-db, line 51: So the only reason we are assuming it's not valid is because it's relative. As pointed below, there could be genuine reason of the storage root not being valid (normalized uri prefix != uri prefix) .

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  public static String deleteStorageLocationPath(String uri) {
+    return updateDirectoryFromUri(uri, Operation.DELETE, Optional.empty()).toString();
+  }
+
+  private static URI updateDirectoryFromUri(
+      String uri,
+      Operation op,
+      Optional<GenerateTemporaryModelVersionCredentialsResponse> credentials) {
```
ravivj-db, line 89: Not sure, it we want to pass credentials as `GenerateTemporaryModelVersionCredentialsResponse` here. These are more of scoped temporary credentials that we pass to the client. The server itself has access to different kind of credentials (for example an `IaM` AWS user role(just an example). We can use them here directly or do the extra step of internally calling generateTemporaryCredentials.

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -0,0 +1,245 @@
...
+    URI parsedUri = URI.create(uri);
+    validateURI(parsedUri);
```
ravivj-db, line 91: so we validate the storage Root once and then we validate the entire Uri  again (which is mostly storage root appended with entity Id's)?

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -0,0 +1,245 @@
...
+      // Convert to an absolute path
+      File file = new File(modelStorageRoot);
+      modelStorageRoot = file.toURI().toString();
+    }
+    // Check if the modelStorageRoot ends with a slash and remove it if it does
+    while (modelStorageRoot.endsWith("/")) {
+      modelStorageRoot = modelStorageRoot.substring(0, modelStorageRoot.length() - 1);
+    }
+
+    return modelStorageRoot;
```
ravivj-db, line 61: can we cache this if this being summoned again and again.

server/src/test/java/io/unitycatalog/server/base/model/BaseModelCRUDTest.java
```diff
@@ -248,6 +266,8 @@ public void testModelCRUD() throws ApiException {
     assertThat(updatedModelVersionInfo.getStatus())
         .isEqualTo(ModelVersionStatus.PENDING_REGISTRATION);
     assertThat(updatedModelVersionInfo.getVersion()).isEqualTo(3L);
+    assertThat(updatedModelVersionInfo.getStorageLocation())
```
ravivj-db, line 269: If we are allowed the update model names, can we also do a check that despite updating the model name, the storage location stays the same.

server/src/test/java/io/unitycatalog/server/utils/UriUtilsTest.java
```diff
@@ -0,0 +1,62 @@
...
+    assertThat(modelVersionPath)
+        .isEqualTo("file:///tmp/random/catalog/schema/models/my-model/versions/1");
+
+    UriUtils.deleteStorageLocationPath(modelVersionPathUri);
+    UriUtils.deleteStorageLocationPath(modelPathUri);
+
+    // cleanup the created catalog
+    UriUtils.deleteStorageLocationPath("file:/tmp/random");
+
+    assertThatThrownBy(
```
ravivj-db, line 50: not sure how is this working ... if we are not doing the normalize test in validate.

server/src/main/java/io/unitycatalog/server/persist/dao/ModelVersionInfoDAO.java
```diff
@@ -90,7 +89,7 @@ public ModelVersionInfo toModelVersionInfo() {
             .source(getSource())
             .version(getVersion())
             .status(ModelVersionStatus.valueOf(getStatus()))
-            .storageLocation(FileUtils.convertRelativePathToURI(url))
+            .storageLocation(url)
```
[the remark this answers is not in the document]
kriscon-db: So TD and Ramesh are onboard with my quickstart proposal of:

1) We have users pip install mlflow
2) Users can then use a builtin dataset inside sklearn (library dependency of mlflow) to train a toy model.
3) User can then register that toy model directly into the default catalog/schema.
4) Users can then re-load the model from UC OSS and run batch inference from the same mlflow session.

This can all be done from a notebook/code we provide with the single step of installing mlflow and will give users the whole flow for using UC OSS w/ mlflow.

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
kriscon-db: I can add some checks here for that.  I think the biggest worry would be for the File based approach that we allow users to relative directory themselves into some real danger.

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
kriscon-db: We're going to be calling this from w/in a transaction, so we should have access to the session, transaction, and user context.  We can probably build some kind of wrapper around this that accepts those, makes a call for credentials and then passes them in here.  This feels a little too low level to be making a REST API call here.

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
kriscon-db: Yep.  The endpoint handlers introduce new levels to the storage root to generate the final container path of the bucket.  It's probably overkill to do this again, but it's a lightweight check and we did mess with the URI a little before we get to it.

server/src/test/java/io/unitycatalog/server/base/model/BaseModelCRUDTest.java
kriscon-db: Yep.  Can add that.  Ids and storage locations (since they are now purely ID based) should remain consistent across updates.

server/src/test/java/io/unitycatalog/server/utils/UriUtilsTest.java
kriscon-db: There's no scheme.

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
ravivj-db: I think this check is to avert the same danger.

server/src/main/java/io/unitycatalog/server/persist/dao/ModelVersionInfoDAO.java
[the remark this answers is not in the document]
kriscon-db: So when users interact with MLflow, they will have a relative local directory that they train the model into.  Whatever we set as the default `storage-root.model` will be used as the storage location by UC OSS.

MLflow will then just make REST API calls to UC OSS, which will respond with the absolute paths that the mlflow client should use for a file based artifact upload (just a copy).

server/src/test/java/io/unitycatalog/server/utils/UriUtilsTest.java
ravivj-db: ah .. right :)

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
ravivj-db: Yes, on second thought, it's better to generate the temporary credentials for WRITE using `CredentialOperations`  and use them here. Keeps the functionality clearly separated.

server/src/main/java/io/unitycatalog/server/persist/dao/ModelVersionInfoDAO.java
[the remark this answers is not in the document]
kriscon-db: The absolute paths will be generated by UC OSS based on the default storage root, which can be relative for quickstart.  Will probably want it to be something lie `./ucroot` or something like that if we don't want to go the route of letting java create a temp directory as the root.

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
kriscon-db: That said, I don't think users ever get to specify the storage location in their calls to UC OSS.  It's a path we fully provide for them on entity creation.  So unless users are really doing something they shouldnt be doing (manually editing the database), not sure how that could happen here.

server/src/main/java/io/unitycatalog/server/persist/dao/ModelVersionInfoDAO.java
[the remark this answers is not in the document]
ravivj-db: Fine, everything is MANAGED for models. Since this path will be created by UC on the fly, we don't need to worry about relative paths.

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
kriscon-db: Yep.  If we're in control and only we will be setting this, then I think we don't need to check for bad paths.

server/src/test/java/io/unitycatalog/server/base/model/BaseModelCRUDTest.java
kriscon-db: Oh actually....thats what this check is doing.  I compare the updatedMV against the originally pulled MV

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
ravivj-db: We'll need that check once we decide to merge FileUtils with this file. but that's a TODO.

server/src/test/java/io/unitycatalog/server/base/model/BaseModelCRUDTest.java
ravivj-db: I meant change the name of the registered model. For example create a model , create a model version. Update the  registered model name. And make sure the model version is reachable from the updated model. We can add those tests later.

ravivj-db (approved):
LGTM

ravivj-db:
The unit tests are failing for some reason.

PR #428, opened 2024-09-06
Fix the `/self` endpoint.

creechy:
Add some checking in currentUser() for the token before accessing it.

**PR Checklist**

**Description of changes**

The `/self` endpoint relies on the acess-token JWT being in the request context. In order to do that, it needs to be intercepted by the `AuthDecorator` first, so adding that to the request pipeline, excluding `auth/tokens` which should be open.

Also made some changes to `/self` to make sure it found the token before trying to access it.

And the token exchange verifies the incoming token subject is in the user database.

Issue https://github.com/unitycatalog/unitycatalog/issues/429

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/AuthService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/AuthService.java
@@ -11,9 +11,11 @@ import com.linecorp.armeria.common.annotation.Nullable;
 import com.linecorp.armeria.server.annotation.ExceptionHandler;
 import com.linecorp.armeria.server.annotation.Param;
 import com.linecorp.armeria.server.annotation.Post;
+import io.unitycatalog.control.model.User;
 import io.unitycatalog.server.exception.ErrorCode;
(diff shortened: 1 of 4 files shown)


server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -146,6 +150,29 @@ public HttpResponse grantToken(OAuthTokenExchangeRequest request) {
     return HttpResponse.ofJson(response);
   }
 
+  private static void verifyPrincipal(DecodedJWT decodedJWT) {
+    String subject =
+        decodedJWT.getClaim("email").isMissing()
+            ? decodedJWT.getClaim("sub").asString()
+            : decodedJWT.getClaim("email").asString();
+
+    if (subject.equals("admin")) {
```
creechy, line 159: This is temporary... when access control is done, `admin` will be a real user in. the db and i'll remove this.

ajreid21 (approved):

alexreid-db (approved):

PR #449, opened 2024-09-09
Fix behavior of creation of managed tables

cloud-fan:
**Description of changes**

UC managed tables should use the managed locations from the UC server. We should not leave it to the underlying `DeltaCatalog` which will generate the default table location using the Spark warehouse path.


connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
```diff
@@ -48,7 +49,19 @@ class UCSingleCatalog extends TableCatalog with SupportsNamespaces {
...
-    deltaCatalog.createTable(ident, columns, partitions, properties)
+    val isExternal = properties.containsKey(TableCatalog.PROP_EXTERNAL)
+    def isPathTable = ident.namespace().length == 1 && new Path(ident.name()).isAbsolute
+    if (!isExternal && !properties.containsKey(TableCatalog.PROP_LOCATION) && !isPathTable) {
+      val newProps = new util.HashMap[String, String]
+      newProps.putAll(properties)
+      // TODO: here we use a fake location for managed table, we should generate table location
+      //       properly when Unity Catalog supports creating managed table.
+      newProps.put(TableCatalog.PROP_LOCATION, "file:///tmp/fake")
+      newProps.put(TableCatalog.PROP_IS_MANAGED_LOCATION, "true")
```
prakharjain09, line 60: Why are we passing `TableCatalog.PROP_IS_MANAGED_LOCATION -> true` instead of `TableCatalog.PROP_EXTERNAL -> false` to the underlying DeltaCatalog?

connectors/spark/src/test/java/io/unitycatalog/connectors/spark/TableReadWriteTest.java
```diff
@@ -320,6 +320,34 @@ public void testCreateExternalDeltaTable() throws ApiException, IOException {
     session.stop();
   }
 
+  @Test
+  public void testCreateManagedParquetTable() {
```
tdas, line 324: lets make this a Delta table

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
```diff
@@ -6,6 +6,7 @@ import io.unitycatalog.client.model.{ColumnInfo, ColumnTypeName, CreateSchema, C
```
tdas, line 6: why we have this PR, can we change the package names to be shorter `io.unitycatalog.spark`

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
```diff
@@ -216,8 +229,12 @@ private class UCProxy extends TableCatalog with SupportsNamespaces {
...
     createTable.setCatalogName(this.name)
+
+    val isExternal = properties.containsKey(TableCatalog.PROP_EXTERNAL)
     val storageLocation = properties.get(TableCatalog.PROP_LOCATION)
-    if (storageLocation == null) {
+    val isManagedLocation = Option(properties.get(TableCatalog.PROP_IS_MANAGED_LOCATION))
+      .exists(_.equalsIgnoreCase("true"))
+    if (!isExternal && (storageLocation == null || isManagedLocation)) {
       // TODO: Unity Catalog does not support managed tables now.
       throw new ApiException("Unity Catalog does not support managed table.")
```
tdas, line 239: managed table support in coming soon.

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
tdas: if these 2 variables are mutually exclusive to each other, then we should have an assert here as well as in Delta code that we never set them in inconsistently

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
```diff
@@ -216,8 +229,12 @@ private class UCProxy extends TableCatalog with SupportsNamespaces {
     createTable.setName(ident.name())
     createTable.setSchemaName(ident.namespace().head)
     createTable.setCatalogName(this.name)
+
+    val isExternal = properties.containsKey(TableCatalog.PROP_EXTERNAL)
     val storageLocation = properties.get(TableCatalog.PROP_LOCATION)
-    if (storageLocation == null) {
+    val isManagedLocation = Option(properties.get(TableCatalog.PROP_IS_MANAGED_LOCATION))
+      .exists(_.equalsIgnoreCase("true"))
+    if (!isExternal && (storageLocation == null || isManagedLocation)) {
```
tdas, line 237: what happens if `isExternal` is set to true, and yet `(storageLocation == null || isManagedLocation)` is true? That should be illegalstateexception instead of "managed table not supported" as the intent is not clear.

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
cloud-fan: We should respect the definition from Spark. The `PROP_EXTERNAL` should only be true if the CREATE TABLE command has the keyword `EXTERNAL` specified. For example, `CREATE EXTERNAL TABLE t ...`. It should not be true even if the keyword `LOCATION` is specified like `CREATE TABLE t ... LOCATION ...`, although most catalogs will create this table as external table since it has a user-specified location.

On the other hand, `PROP_IS_MANAGED_LOCATION` is used to indicate that the location is not user-specified by managed by the system, which is exactly the case here.

cloud-fan: that said, these two properties are not mutually exclusive to each other. @tdas

cloud-fan: I can add comments here to repeat the Spark definition for `PROP_IS_MANAGED_LOCATION`.

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
cloud-fan: From Spark's perspective, this may happen if people do `CREATE EXTERNAL TABLE t ...` without a LOCATION clause. Maybe some catalog implementations treat it as valid by generating location for it, but this will never happen in UC/HMS.

prakharjain09: What blocks this in UC? If there are no assertions for this, should we add one as part of UCSingleCatalog?

cloud-fan: makes sense, let me add this check

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -48,7 +49,24 @@ class UCSingleCatalog extends TableCatalog with SupportsNamespaces {
       columns: Array[Column],
       partitions: Array[Transform],
       properties: util.Map[String, String]): Table = {
-    deltaCatalog.createTable(ident, columns, partitions, properties)
+    val hasExternalClause = properties.containsKey(TableCatalog.PROP_EXTERNAL)
+    val hasLocationClause = properties.containsKey(TableCatalog.PROP_LOCATION)
+    def isPathTable = ident.namespace().length == 1 && new Path(ident.name()).isAbsolute
```
prakharjain09, line 54: In what scenario is `ident` a path? Is it when user says:

`df.format("delta").save("<path>")` ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -227,8 +245,17 @@ private class UCProxy extends TableCatalog with SupportsNamespaces {
     createTable.setName(ident.name())
     createTable.setSchemaName(ident.namespace().head)
     createTable.setCatalogName(this.name)
+
+    val hasExternalClause = properties.containsKey(TableCatalog.PROP_EXTERNAL)
     val storageLocation = properties.get(TableCatalog.PROP_LOCATION)
-    if (storageLocation == null) {
+    if (hasExternalClause && storageLocation == null) {
```
prakharjain09, line 251: This method is invoked after delta do commit 0 via the DeltaCatalog. Once DeltaCatalog wants to persist this table in UC, it calls this method which then uses SDK to create the table object in UC.

If the above is correct, shouldn't `storageLocation` be non-null for both managed / external?
- for external , it will be passed by user
- for managed, it will be auto generated by `UCSingleCatalog.createTable`

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
cloud-fan: it's something like `CREATE TABLE delta.'/my/path' ...`

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
cloud-fan: This is for an invalid case `CREATE EXTERNAL TABLE ... [no LOCATION clause]`, see the test in https://github.com/unitycatalog/unitycatalog/pull/449/files#diff-3fc30b0d18b7b824cbdf24035313086c9b0787ab51392c62f910400b8081067eR329

prakharjain09: The overall flow is: `UCSingleCatalog#createTable` -> `DeltaCatalog#createTable` -> `CreateDeltaTableCommand#execute` -> `UCProxy#createTable` .

If user says `CREATE EXTERNAL TABLE ... [no LOCATION clause]`, then currently we will  end up creating a default path using [spark.sessionState.catalog.defaultTablePath(id)](https://github.com/delta-io/delta/pull/3663/files#diff-e880b2b407605f237c5bb566e3e06f981a4fbbb6ff675d7832b2cf5d2daa6b3dR146), we will initialize a delta log and then we will invoke `UCProxy#createTable` where we will fail.

Should we fail early i.e. in `UCSingleCatalog#createTable` itself when EXTERNAL is mentioned but location is not.

tdas: +1 to failing early.

tdas (approved):

prakharjain09 (commented):
Late approval - PR LGTM.

PR #465, opened 2024-09-10
Introduce DEBUG env var for remote Java debugging

jaceklaskowski:
This PR introduces developer-only `DEBUG` environment variable to be the JDWP port for debugging the UC Server and CLI.

```bash
❯ DEBUG=5001 ./bin/start-uc-server
Remote Java debugging enabled on port 5001
Listening for transport dt_socket at address: 5001
...
```

```bash
❯ DEBUG=5100 ./bin/uc catalog list
Remote Java debugging enabled on port 5100
Listening for transport dt_socket at address: 5100
...
```

`DEBUG` is only the port part in the following JDWP execution. Both the server and the CLI run JDWP with `suspend=y`.

```text
-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:$DEBUG
```


jaceklaskowski:
@dennyglee, May I have your look at the changes, please? 🙏

dennyglee:
This looks awesome @jaceklaskowski  - will review shortly!

bin/start-uc-server
```diff
@@ -6,6 +6,14 @@ ROOT_DIR="$(dirname "$SCRIPT_DIR")"
 TARBALL_JAR_DIR="$ROOT_DIR/jars"
 SERVER_CLASS_FILE="$ROOT_DIR/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java"
 
+# Use DEBUG env var to turn remote Java debugging (JDWP) on
+# DEBUG is the port that the JDWP agent listens to
+DEBUG_CMD=""
+if [[ -n "$DEBUG" ]]; then
```
yaoderek, line 12: Can we name DEBUG to DEBUG_PORT to be more explicit?

jaceklaskowski: "Great minds think alike" here 😉  That's exactly what I tried the very first time but thought UC devs won't like typing so much and although not very explanatory it's not supposed to be so. It's for the devs after all, and the shorter (and even cryptic) the better.

After I ended up with `DEBUG` I liked it very much and would be against expanding it to include `_PORT` until it's really really needed (e.g., when we decide to add even more debug-related env vars).

Could this explanation be OK to convince you to accept `DEBUG` @yaoderek, please? 🙏

jaceklaskowski:
@dennyglee May I have your approval to the changes, please? 🙏

bin/start-uc-server
yaoderek: Thanks for the explanation @jaceklaskowski . If we want to use DEBUG, can we make it a flag with optional value (5001 by default). This way, the cmd can be even short like.

`DEBUG ./bin/start-uc-server`

dennyglee (approved):
Thanks @jaceklaskowski and really appreciate the review @yaoderek !

creechy:
Just a thought, why not make a command line argument (or two) instead of environment variable? I could see, for example
- `--enable-debug` to add a default debugging arguments to the command
- `--debug-args` (or `--debug-port`) to use custom debugging parameters

Or go with something like a `-J...` argument to add arbitrary JVM args. Pretty typical, but a little obtuse because one would need to know do `-J-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5001`

bin/start-uc-server
```diff
@@ -15,18 +23,19 @@ generate_jar() {
...
-  SERVER_JAVA_COMMAND="java -cp $(cat "$1") io.unitycatalog.server.UnityCatalogServer"
+  SERVER_JAVA_COMMAND="java $DEBUG_CMD -cp $(cat "$1") io.unitycatalog.server.UnityCatalogServer"
   shift
   $SERVER_JAVA_COMMAND "$@" || exit
 }
 
 # This function runs the server using the source code
 run_in_source_code() {
   SERVER_TARGET_DIR="$ROOT_DIR/server/target"
-  SERVER_JAR=$(find "$SERVER_TARGET_DIR" -name "unitycatalog-server*.jar" | head -n 1)
```
yili-db, line 26: This `find` should not be removed. Otherwise the `generate_jar` below won't be triggered.

yili-db (changes requested):

PR #467, opened 2024-09-11
UC should work without Delta Lake

cloud-fan:
**Description of changes**

If Delta Lake is not included in the class path, UC should still work and support file sources.


connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
```diff
@@ -23,65 +23,73 @@ import scala.collection.JavaConverters._
...
+  private var delegating: TableCatalog = null
 
   override def initialize(name: String, options: CaseInsensitiveStringMap): Unit = {
     val proxy = new UCProxy()
     proxy.initialize(name, options)
-    deltaCatalog = Class.forName("org.apache.spark.sql.delta.catalog.DeltaCatalog")
-      .getDeclaredConstructor().newInstance().asInstanceOf[TableCatalog]
-    deltaCatalog.asInstanceOf[DelegatingCatalogExtension].setDelegateCatalog(proxy)
+    try {
+      if (options.containsKey("__TEST_NO_DELTA__")) {
```
tdas, line 32: why pollute the options variable with such test-only stuff. isnt it cleaner to have a separate static variable that is completely isolated from the any of the production code paths like options?

cloud-fan: That static variable also needs to be in the production code so that we can access it here. I don't have a strong preference, users can set that static variable as well as this special option, if they read the source code.

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
```diff
@@ -23,65 +23,73 @@ import scala.collection.JavaConverters._
  */
 class UCSingleCatalog extends TableCatalog with SupportsNamespaces {
 
-  private var deltaCatalog: TableCatalog = null
+  private var delegating: TableCatalog = null
```
cloud-fan, line 26:
```suggestion
  @volatile private var delegating: TableCatalog = null
```

connectors/spark/src/main/scala/io/unitycatalog/connectors/spark/UCSingleCatalog.scala
cloud-fan: or I can make the delegating catalog class name as an option, so that we can set a different one other than `org.apache.spark.sql.delta.catalog.DeltaCatalog` in test.

connectors/spark/src/test/java/io/unitycatalog/connectors/spark/TableReadWriteTest.java
```diff
@@ -41,6 +41,33 @@ public class TableReadWriteTest extends BaseSparkIntegrationTest {
...
+            .config(catalogConf + ".uri", serverConfig.getServerUrl())
+            .config(catalogConf + ".token", serverConfig.getAuthToken())
+            .config(catalogConf + ".__TEST_NO_DELTA__", "true");
+    SparkSession session = builder.getOrCreate();
+
+    setupExternalParquetTable(PARQUET_TABLE, new ArrayList<>(0));
+    testTableReadWrite(SPARK_CATALOG + "." + SCHEMA_NAME + "." + PARQUET_TABLE, session);
+
+    // Since `DeltaCatalog` is not used, the following CREATE TABLE command doesn't fail while it
+    // should because the table location is empty and Delta Lake shouldn't allow it.
```
tdas, line 62: i dont get this. why should this pass?? if delta library is not in the classpath, then all queries with the format delta should fail

tdas: we cannot really test that in these unit tests since delta is in the class path. so ... what is this test really testing?

cloud-fan: `DeltaCatalog#createTable` errors out if the table to be created is external but the location is empty. This is to prove that `DeltaCatlaog` is not used in the CREATE TABLE workflow. If I remove `.config(catalogConf + ".__TEST_NO_DELTA__", "true")` in this test case, this command will fail

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
```diff
@@ -41,6 +41,33 @@ public class TableReadWriteTest extends BaseSparkIntegrationTest {
...
+            .config(catalogConf + ".uri", serverConfig.getServerUrl())
+            .config(catalogConf + ".token", serverConfig.getAuthToken())
+            .config(catalogConf + ".__TEST_NO_DELTA__", "true");
+    SparkSession session = builder.getOrCreate();
+
+    setupExternalParquetTable(PARQUET_TABLE, new ArrayList<>(0));
+    testTableReadWrite(SPARK_CATALOG + "." + SCHEMA_NAME + "." + PARQUET_TABLE, session);
+
+    // Since `DeltaCatalog` is not used, the following CREATE TABLE command doesn't fail while it
+    // should because the table location is empty and Delta Lake shouldn't allow it.
```
tdas, line 62: if this is to prove that DeltaCatalog is not being used... then i think this is a very brittle test. This assumes that there is some check somewhere in a different project that can be removed any time in the future and then this test will break in such a way that we will not know how to really fix (because we wont know what to other failure to rely on)

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -24,24 +24,32 @@ import scala.collection.JavaConverters._
  */
 class UCSingleCatalog extends TableCatalog with SupportsNamespaces {
 
-  private var deltaCatalog: TableCatalog = null
+  @volatile private var delegating: TableCatalog = null
```
tdas, line 27: delegating -> delegate or deletegateCatalog (should be a noun)

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
cloud-fan: then I don't have a good idea to test it... Delta is still in the classpath and Spark will trigger the delta scan.

Maybe we don't need to verify it as it's obvious that `DeltaCatalog` won't be used in this test.

tdas: maybe so.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -24,24 +24,32 @@ import scala.collection.JavaConverters._
...
+  @volatile private var delegate: TableCatalog = null
 
   override def initialize(name: String, options: CaseInsensitiveStringMap): Unit = {
     val proxy = new UCProxy()
     proxy.initialize(name, options)
-    deltaCatalog = Class.forName("org.apache.spark.sql.delta.catalog.DeltaCatalog")
-      .getDeclaredConstructor().newInstance().asInstanceOf[TableCatalog]
-    deltaCatalog.asInstanceOf[DelegatingCatalogExtension].setDelegateCatalog(proxy)
+    try {
+      if (options.containsKey("__TEST_NO_DELTA__")) {
```
cloud-fan, line 33: cc @tdas for this discussion: https://github.com/unitycatalog/unitycatalog/pull/467#discussion_r1755425624

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -24,24 +24,32 @@ import scala.collection.JavaConverters._
...
+        throw new ClassNotFoundException("DeltaCatalog")
+      }
+      delegate = Class.forName("org.apache.spark.sql.delta.catalog.DeltaCatalog")
+        .getDeclaredConstructor().newInstance().asInstanceOf[TableCatalog]
+      delegate.asInstanceOf[DelegatingCatalogExtension].setDelegateCatalog(proxy)
+    } catch {
+      case _: ClassNotFoundException =>
```
tdas, line 40: can you add logs so that we know what is happening? otherwise this will be impossible to debug in real deployment.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
tdas: I am okay with making class name configurable. there might be a simpler approach though. have 2 static threadlocal mutable boolean variable. 
1. one dictates whether to load delta catalog or not
2. other is set if the deltacatalog loaded or not
with that you can set the first variable to load / not load the DeltaCatalog (and reset the second one to false), and then test whether DeltaCatalog was loaded or not.

cloud-fan:
@tdas I've addressed your comment, please take another look. Thanks!

tdas (approved):

PR #473, opened 2024-09-12
support create external delta table

cloud-fan:
**Description of changes**

In `UCSingleCatalog#createTable`, we should get credentials by path and put the credential into the table properties, so that the underlying `DeltaCatalog` can access the Delta log when creating external delta tables.


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -80,7 +99,26 @@ class UCSingleCatalog extends TableCatalog with SupportsNamespaces with Logging
       // user-specified but system-generated, which is exactly the case here.
       newProps.put(TableCatalog.PROP_IS_MANAGED_LOCATION, "true")
       delegate.createTable(ident, columns, partitions, newProps)
+    } else if (hasLocationClause) {
+      val location = properties.get(TableCatalog.PROP_LOCATION)
+      assert(location != null)
+      val cred = temporaryCredentialsApi.generateTemporaryPathCredentials(
+        new GenerateTemporaryPathCredential().url(location).operation(PathOperation.PATH_READ))
```
bijanhoule, line 106: Should this be `PATH_CREATE_TABLE` here?
```suggestion
        new GenerateTemporaryPathCredential().url(location).operation(PathOperation.PATH_CREATE_TABLE))
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -122,35 +160,56 @@ class UCSingleCatalog extends TableCatalog with SupportsNamespaces with Logging
...
+        GcsVendedTokenProvider.ACCESS_TOKEN_KEY -> gcsCredentials.getOauthToken,
+        GcsVendedTokenProvider.ACCESS_TOKEN_EXPIRATION_KEY -> temporaryCredentials.getExpirationTime.toString,
+        "fs.gs.impl.disable.cache" -> "true"
+      )
+    } else if (scheme == "abfs" || scheme == "abfss") {
+      val azCredentials = temporaryCredentials.getAzureUserDelegationSas
+      Map(
+        FS_AZURE_ACCOUNT_AUTH_TYPE_PROPERTY_NAME -> "SAS",
+        FS_AZURE_ACCOUNT_IS_HNS_ENABLED -> "true",
+        FS_AZURE_SAS_TOKEN_PROVIDER_TYPE -> "io.unitycatalog.connectors.spark.AbfsVendedTokenProvider",
```
bijanhoule, line 190: not related to the refactoring, but noticed this needed migration:

```suggestion
        FS_AZURE_SAS_TOKEN_PROVIDER_TYPE -> "io.unitycatalog.spark.AbfsVendedTokenProvider",
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
cloud-fan: This is for CREATE EXTERNAL TABLE so the path already exists, Delta needs to read the delta log to do some validation, and I think this is `PATH_READ`?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
cloud-fan, line 190:
```suggestion
        FS_AZURE_SAS_TOKEN_PROVIDER_TYPE -> classOf[AbfsVendedTokenProvider].getName,
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -80,7 +99,26 @@ class UCSingleCatalog extends TableCatalog with SupportsNamespaces with Logging
...
+      newProps.putAll(credentialProps.asJava)
+      // TODO: Delta requires the options to be set twice in the properties, with and without the
+      //       `option.` prefix. We should revisit this in Delta.
+      val prefix = TableCatalog.OPTION_PREFIX
+      newProps.putAll(credentialProps.map {
+        case (k, v) => (prefix + k, v)
+      }.asJava)
+      delegate.createTable(ident, columns, partitions, newProps)
     } else {
+      // TODO: for path-based tables, Spark should generate a location property using the qualified
```
tdas, line 120: what does this TODO mean? is there some changed needed here ... or in Spark? 
will create on path based tables invoke `UCSingleCatalog`?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -122,35 +160,56 @@ class UCSingleCatalog extends TableCatalog with SupportsNamespaces with Logging
...
   val LOAD_DELTA_CATALOG = ThreadLocal.withInitial[Boolean](() => true)
   val DELTA_CATALOG_LOADED = ThreadLocal.withInitial[Boolean](() => false)
+
+  def generateCredentialProps(
+      scheme: String,
+      temporaryCredentials: TemporaryCredentials): Map[String, String] = {
+    if (scheme == "s3") {
+      val awsCredentials = temporaryCredentials.getAwsTempCredentials
+      Map(
+        // TODO: how to support s3:// properly?
```
tdas, line 170: is this TODO still relevant? if so, please clarify what does "properly" mean? what is improper?

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
```diff
@@ -173,6 +173,32 @@ public void testCredentialDelta(String scheme) throws ApiException, IOException
     session.stop();
   }
 
+  @Test
+  public void testCredentialCreateDeltaTable() throws IOException {
```
tdas, line 177: why isnt this test parameterized for s3, gs, abfs .. like the test below?

and

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
cloud-fan: In DBR, the path string in the path identifier will be qualified in the DeltaCatalog, then call UC to request credentials. Now we are requesting credentials before entering DeltaCatalog, and the path is not qualified yet. We should let Spark to qualify the path string in the path identifier, similar to how Spark handle the normal table location.

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
```diff
@@ -173,6 +173,32 @@ public void testCredentialDelta(String scheme) throws ApiException, IOException
...
+    SparkSession session = createSparkSessionWithCatalogs(SPARK_CATALOG, CATALOG_NAME);
+
+    String loc1 = "s3://test-bucket0" + generateTableLocation(SPARK_CATALOG, DELTA_TABLE);
+    setupDeltaTableLocation(session, loc1, new ArrayList<>(0));
+    String t1 = SPARK_CATALOG + "." + SCHEMA_NAME + "." + DELTA_TABLE;
+    session.sql(String.format("CREATE TABLE %s USING delta LOCATION '%s'", t1, loc1));
+    testTableReadWrite(t1, session);
+
+    String loc2 = "s3://test-bucket1" + generateTableLocation(CATALOG_NAME, DELTA_TABLE);
```
tdas, line 186: what is this second set testing beyond the first set of lies in test-bucket0?

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
```diff
@@ -173,6 +173,32 @@ public void testCredentialDelta(String scheme) throws ApiException, IOException
...
+    setupDeltaTableLocation(session, loc2, new ArrayList<>(0));
```
tdas, line 187: If the location does have a delta table, then delta should create a brand new delta log there. are we testing that?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
cloud-fan: This is kind of a hack that I don't fully understand. I have to configure `s3a` here while the user-specified path is `s3://`. Need some experts to take a look.

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
cloud-fan: do you mean empty table? let me add a test

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
cloud-fan: The test only sets up fs implementation for `s3://`: https://github.com/unitycatalog/unitycatalog/blob/ceecee1977afb04ef7b3237071346782041607a3/connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java#L50

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
cloud-fan: to perform a join and prove that we support tables with different credentials in one query.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
tdas: But can you write that down in the comment? that context needs to be written down for others to understand.

regarding the problem, why are you configuring s3a? is it because actual s3 support requires configuring s3a?

cloud-fan: I was just trying different approaches to make tests pass at that time, and using `s3a` happened to solve the issue. I can put my question in the comment `Why can't we use s3 instead of s3a?` but I really don't know the answer.

bijanhoule: I think it's pretty common to bring in `hadoop-aws` (which provides the s3a filesystem), and then just wire that up to `s3://` URIs in spark conf as well, i.e. `spark.hadoop.fs.s3.impl=org.apache.hadoop.fs.s3a.S3AFileSystem`.  In practice I don't think it's a problem to set the credential conf on both schemes just in case

cloud-fan: Cool! I can address this in a followup PR as this is unrelated to this PR.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -122,35 +160,56 @@ class UCSingleCatalog extends TableCatalog with SupportsNamespaces with Logging
...
+        "fs.s3a.session.token" -> awsCredentials.getSessionToken,
+        "fs.s3a.path.style.access" -> "true",
+        "fs.s3.impl.disable.cache" -> "true",
+        "fs.s3a.impl.disable.cache" -> "true"
+      )
+    } else if (scheme == "gs") {
+      val gcsCredentials = temporaryCredentials.getGcpOauthToken
+      Map(
+        GcsVendedTokenProvider.ACCESS_TOKEN_KEY -> gcsCredentials.getOauthToken,
+        GcsVendedTokenProvider.ACCESS_TOKEN_EXPIRATION_KEY -> temporaryCredentials.getExpirationTime.toString,
```
tdas, line 182: @cloud-fan we need to add another configuration for GCS - `fs.gs.create.items.conflict.check.enable=false`

bijan-db: Confirmed this locally - we could add the following to reduce the amount of client conf needed:
```scala
        "fs.gs.create.items.conflict.check.enable" -> "false",
        "fs.gs.auth.type" -> "ACCESS_TOKEN_PROVIDER",
        "fs.gs.auth.access.token.provider" -> classOf[GcsVendedTokenProvider].getName,
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -122,35 +160,56 @@ class UCSingleCatalog extends TableCatalog with SupportsNamespaces with Logging
...
+        "fs.gs.impl.disable.cache" -> "true"
```
cloud-fan, line 183:
```suggestion
        "fs.gs.create.items.conflict.check.enable" -> "false",
        "fs.gs.auth.type" -> "ACCESS_TOKEN_PROVIDER",
        "fs.gs.auth.access.token.provider" -> classOf[GcsVendedTokenProvider].getName,
        "fs.gs.impl.disable.cache" -> "true"
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
cloud-fan: done, thanks!

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -122,35 +160,59 @@ class UCSingleCatalog extends TableCatalog with SupportsNamespaces with Logging
...
+        "fs.s3a.path.style.access" -> "true",
+        "fs.s3.impl.disable.cache" -> "true",
+        "fs.s3a.impl.disable.cache" -> "true"
+      )
+    } else if (scheme == "gs") {
+      val gcsCredentials = temporaryCredentials.getGcpOauthToken
+      Map(
+        GcsVendedTokenProvider.ACCESS_TOKEN_KEY -> gcsCredentials.getOauthToken,
+        GcsVendedTokenProvider.ACCESS_TOKEN_EXPIRATION_KEY -> temporaryCredentials.getExpirationTime.toString,
+        "fs.gs.create.items.conflict.check.enable" -> "false",
```
tdas, line 183: I think you should add this as constants in the GcsVendedTokenProvider. follow the existing template in the 2 lines above?

tdas (approved):

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
cloud-fan: I'll leave it to @bijanhoule . For abfs these keys are defined the azurebfs library, not sure if gs has something similar.

tdas: @cloud-fan I was just talking about `GcsVendedTokenProvider.ACCESS_TOKEN_EXPIRATION_KEY ` which is defined in our file - https://github.com/unitycatalog/unitycatalog/blob/ddf5496d6c0db0e2bc8cd1bb950d641a4fadde69/connectors/spark/src/main/scala/io/unitycatalog/spark/GcsVendedTokenProvider.java#L11

so following the existing template of GCS meant adding another constant in this file.

PR #476, opened 2024-09-13
Disable failing spark test

bijanhoule:
**PR Checklist**

**Description of changes**

Temporarily disable a failing test.

This test is using a property to mock the managed table creation path.  However, that property isn't propagating through the delta table creation as intended, so the expected exception doesn't get thrown


bijanhoule:
False alarm - this was a local caching issue

PR #483, opened 2024-09-13
Enable identity logging on create/update paths

kriscon-db:
**PR Checklist**

**Description of changes**

Enable identity tracking for all create and update paths for securables in UC OSS

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
@@ -39,14 +39,17 @@ public class CatalogRepository {
   public CatalogInfo addCatalog(CreateCatalog createCatalog) {
     ValidationUtils.validateSqlObjectName(createCatalog.getName());
     String callerId = IdentityUtils.findPrincipalEmailAddress();
+    Long createTime = System.currentTimeMillis();
     CatalogInfo catalogInfo =
         new CatalogInfo()
             .id(java.util.UUID.randomUUID().toString())
             .comment(createCatalog.getComment())
             .name(createCatalog.getName())
             .owner(callerId)
-            .createdAt(System.currentTimeMillis())
+            .createdAt(createTime)
             .createdBy(callerId)
+            .updatedAt(createTime)
+            .updatedBy(callerId)
             .properties(createCatalog.getProperties());
 
     try (Session session = SESSION_FACTORY.openSession()) {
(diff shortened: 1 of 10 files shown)


server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
```diff
@@ -37,12 +38,15 @@ public static CatalogRepository getInstance() {
...
     ValidationUtils.validateSqlObjectName(createCatalog.getName());
+    String callerId = IdentityUtils.findPrincipalEmailAddress();
     CatalogInfo catalogInfo =
         new CatalogInfo()
             .id(java.util.UUID.randomUUID().toString())
             .comment(createCatalog.getComment())
             .name(createCatalog.getName())
+            .owner(callerId)
             .createdAt(System.currentTimeMillis())
+            .createdBy(callerId)
```
rameshchandra, line 49: Since create is a form of update, also set `updatedBy` to `callerId` and `updatedAt` to `currentTime`.

kriscon-db: Ah.  I had that originally in Models but then I saw that no other securable was setting update like that.  I'll fix everything to set the update on create as well.

creechy (approved):
LGTM

server/src/main/java/io/unitycatalog/server/persist/SchemaRepository.java
```diff
@@ -219,7 +227,9 @@ public SchemaInfo updateSchema(String fullName, UpdateSchema updateSchema) {
           PropertyDAO.from(updateSchema.getProperties(), schemaInfoDAO.getId(), Constants.SCHEMA)
               .forEach(session::persist);
         }
+        schemaInfoDAO.setOwner(callerId);
```
rameshchandra, line 230: This doesn't seem correct -- `updateSchema` shouldn't be updating the owner to the caller.

server/src/main/java/io/unitycatalog/server/persist/dao/ModelVersionInfoDAO.java
```diff
@@ -78,7 +78,7 @@ public static ModelVersionInfoDAO from(ModelVersionInfo modelVersionInfo) {
         .updatedAt(
             modelVersionInfo.getUpdatedAt() != null
                 ? new Date(modelVersionInfo.getUpdatedAt())
-                : new Date())
+                : null)
```
rameshchandra, line 81: Why is this different from the logic for `createdAt`?

server/src/main/java/io/unitycatalog/server/persist/SchemaRepository.java
kriscon-db: Ah.  Good catch.  Removing.

server/src/main/java/io/unitycatalog/server/persist/dao/ModelVersionInfoDAO.java
kriscon-db: This was changed to match the other DAO's filling in of this field.  I have confirmed that the field is always being filled in at creation/update, so there's little risk in having this match the other DAOs

rameshchandra (approved):
LGTM

Issue #482, opened 2024-09-13
Unify Dockerfiles


tnightengale:
**Is your feature request related to a problem? Please describe.**

The current approach of maintaining two Dockerfiles: `Dockerfile` and `uc-cli.dockerfile` seems quite cumbersome and bloated.

**Describe the solution you would like**

Combine the current Dockerfiles and separate out configuration in the clear and relevant interface. CLI or server could be an entrypoint arg, not a separate file.

**Additional context**

When publishing to DockerHub eventually, we should not push two different images that are basically the same.

tnightengale:
take

Issue #480, opened 2024-09-13
Support arm64 Docker Image


tnightengale:
**Describe the bug**

Building images on M* Macs (arm64) doesn't seem to work. As far as I can tell, this is because of the choice to use alpine distro, which does not offer an arm64 compatible image (for jdk-17 only, sigh).

See [here](https://hub.docker.com/r/arm64v8/eclipse-temurin/) for the available eclipse-temurin arm64 tags.

Note the lack of an alpine tag for jdk-17 on arm64: 
<img width="1512" alt="Screenshot 2024-09-13 at 10 07 37 AM (2)" src="https://github.com/user-attachments/assets/c1c8ca72-8091-4a38-94b2-05636d954fd9">

**To Reproduce**

```
tnightengale@Teghans-14-MBP bin % ./build-uc-server-docker   
Changing directory to /Users/tnightengale/git/unitycatalog
Building unitycatalog:0.2.0-SNAPSHOT using the build definition from Dockerfile
#0 building with "desktop-linux" instance using docker driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 4.69kB done
#1 DONE 0.0s

#2 [internal] load metadata for docker.io/library/eclipse-temurin:17-jdk-alpine
#2 ERROR: no match for platform in manifest: not found
------
 > [internal] load metadata for docker.io/library/eclipse-temurin:17-jdk-alpine:
------
Dockerfile:49
--------------------
  47 |     RUN ./docker/copy_jars_from_classpath.sh ${jars_directory}
  48 |     
  49 | >>> FROM eclipse-temurin:17-jdk-alpine AS build_uc
  50 |     
  51 |     ARG unitycatalog_uid
--------------------
ERROR: failed to solve: eclipse-temurin:17-jdk-alpine: failed to resolve source metadata for docker.io/library/eclipse-temurin:17-jdk-alpine: no match for platform in manifest: not found
Returning to base directory /Users/tnightengale/git/unitycatalog/docker/bin
tnightengale@Teghans-14-MBP bin % 
```

1. Run `$ ./build-uc-server-docker` on arm64.

**Expected behavior**

Use a supported jdk and linux distro combination with support for arm64.

**System [please complete the following information]:**

```
tnightengale@Teghans-14-MBP bin % uname -a
[...]

Issue #319, opened 2024-08-02
Unity Catalog Docker Tests


dennyglee:
**Is your feature request related to a problem? Please describe.**
With the merge of #116, we have a great jump start for the Unity Catalog docker.  But as we're still early, we should add tests to validate the build and add some additional instructions on how to use the Docker.  This would be a pre-requisite before pushing the docker up to DockerHub.

**Describe the solution you would like**
- Add tests to validate build and key functionality works after each build
- Add instructions to simplify running the docker

**Additional context**

tnightengale:
@dennyglee I am interested in contributing to this issue, as a step towards publishing the image on DockerHub. Help me get situated here: what tests are you thinking?

tnightengale:
take

tnightengale:
@dennyglee started an outline of what builds and pushes could look like here: https://github.com/unitycatalog/unitycatalog/pull/481

PR #485, opened 2024-09-16
build: Unified Docker Image

tnightengale:
**PR Checklist**

**Description of changes**

This PR simplifies the use of Docker by merging the cli and server into a unified, multi-arch compatible, Dockerfile with clean syntax, and brings the project closer to releasing an official image on DockerHub. See #319, #482, and #480.

**Additional info**
- Example of the proposed GitHub Actions workflow [successfully building and pushing images](https://github.com/tnightengale/unitycatalog/actions/runs/10930448242/job/30343362272) on my fork
- Shaved off 1GB from the `./Dockerfile` image to get it down to about 1.35GB

This PR does the following:
- Simplifies the main `Dockerfile` and enables arm64 native builds
- Add `compose.yaml` to allow running UC with one command `docker compose up -d`
- Adds standard build and push GitHub Actions workflow. Actually pushing the official image to DockerHub can be enabled in a subsequent PR.


tnightengale:
@dennyglee I have cycles this week to try and push towards releasing an official image. Would love your feedback here to collaborate on this.

dennyglee:
Thanks @tnightengale !  I do like the streamlined flow but let me also include @jaceklaskowski as he's been deeply involved with the docker creation.

ui/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+COPY package.json yarn.lock ./
+
+# TODO: configure this directly from the ui app
+# Work around to set the proxy host; required for running locally with docker-compose
+RUN <<EOF
+set -ex
+apt-get update
+apt-get install -y jq
+jq --arg proxy "http://$PROXY_HOST:8080" '.proxy = $proxy' package.json > temp.json
```
tnightengale, line 15: Hey 👋 @jamieknight-db! I am trying to parameterize the UI react app to use the api url `http://backend:8080` instead of `http:localhost:8080` when running inside of docker compose. My react knowledge is pretty dismal; is there a better way to configure the react app than just replacing the `proxy` key inside `package.json`?

tnightengale: For context, this PR enables us to spin up the frontend and backend with just `docker compose up -d`

build.sbt
```diff
@@ -270,6 +270,14 @@ lazy val server = (project in file("server"))
       "com.h2database" %  "h2" % "2.2.224",
 
       "org.hibernate.orm" % "hibernate-core" % "6.5.0.Final",
+      // For loading properties from the environment
+      "uk.org.webcompere" % "lightweight-config" % "1.3.0",
+
+      // For MySQL backend
+      "com.mysql" % "mysql-connector-j" % "8.4.0",
```
tnightengale, line 277: Hey @xx-db 👋 I saw you tried to add the mysql jdbc driver back in #211, but ultimately it wasn't merged. Currently the CI is complaining about a license issue by adding it here. Was that what you encountered? Do you know who might be able to either:

a) Approve overriding the license lint, and allow the connector to be used
b) Confirm that we can't use the connector due to a license conflict, and therefore the project should stop mentioning using MySQL

Thanks in advance!

CC: @dennyglee

Dockerfile
```diff
@@ -1,147 +1,59 @@
...
+ENV HOME=$HOME \
+    JAVA_HOME=/usr/lib/jvm/default-jvm
 
-# Define the shell used within the container
-SHELL ["/bin/bash", "-i", "-c", "-o", "pipefail"]
+# Copy Java from base
+COPY --from=base $JAVA_HOME $JAVA_HOME
 
-ENV UC_SERVER_BIN="${unitycatalog_home}/${unitycatalog_bin}/start-uc-server"
+ENV PATH="${BASH_HOME}:${JAVA_HOME}/bin:${PATH}"
```
tnightengale, line 30:
```suggestion
ENV PATH="${JAVA_HOME}/bin:${PATH}"
```

ui/Dockerfile
jamieknight-db: @tnightengale to be honest, I am unsure the best way to parameterize this. Ideally, all parameterized variables are stored in the `.env` file and could be accessed from the package.json scripts. 

```
   "proxy": "${REACT_APP_API_URL}"
```

Or we might need to utilize a setupProxy.js file and createProxyMiddleware and go that route? 

Maybe @bijanhoule or @creechy or @yc-shawn can chime in?

README.md
```diff
@@ -158,7 +163,40 @@ yarn start
...
+
+### Run with Postgres Container
+To run with a Postgres db container, instead of the default hibernate filesystem datastore, use the `compose.psql.yaml` file to [override](https://docs.docker.com/compose/multiple-compose-files/merge/#example) the default `compose.yaml` file:
+
+```sh
+docker compose -f compose.yaml -f compose.psql.yaml up -d
+```
 
+### Run with External Database
+The Unity Catalog container can be configured to connect with an external Postgres or MySQL database. Pass environment variables to the container to configure the [hibernate.properties](etc/conf/hibernate.properties) file. For example, `docker run --rm --env-file .env unitycatalog_backend_1`. See [.env.example](.env.example) for the possible variables.
```
tnightengale, line 199: TODO: get clarity on MySQL jdbc support inside the image.

server/src/main/java/io/unitycatalog/server/persist/utils/HibernateUtils.java
```diff
@@ -31,10 +30,9 @@ private static SessionFactory createSessionFactory() {
         throw new RuntimeException("PropertiesUtil instance is null in createSessionFactory");
       }
 
-      Properties hibernateProperties = new Properties();
-      try (InputStream input = Files.newInputStream(Paths.get("etc/conf/hibernate.properties"))) {
-        hibernateProperties.load(input);
-      }
+      // Load properties with environment variable substitution to configure service in Docker
+      Properties hibernateProperties =
+          ConfigLoader.loadProperties(Paths.get("etc/conf/hibernate.properties"));
```
tnightengale, line 35: Looks like this utility to read `.properties` files with env var substitution relies on https://github.com/spotbugs/spotbugs, which has a  LGPL-2.1 license. This is [failing the ci[(https://github.com/unitycatalog/unitycatalog/actions/runs/10932397607/job/30349161846).

Is this really prohibited?

Dockerfile
```diff
@@ -1,147 +1,60 @@
...
-# Note: The default allocated heam memory size is too small
-# and will cause the process to fail when attempting to compile
-# and generate the Uber-Jar. Therefore it is important to choose
-# a size large enough for the compiler to run. 
-ARG sbt_args="-J-Xmx2G"
-# FIXME Pass it from the outside
-ARG unitycatalog_version="0.2.0-SNAPSHOT"
-ARG jars_directory="server/target/jars"
+# syntax=docker.io/docker/dockerfile:1.7-labs
+ARG HOME="/unitycatalog"
```
jaceklaskowski, line 2: Why not stick to `/opt/unitycatalog`. It's shorter yet we don't really care, do we? `/opt` looks as the right place to me.

README.md
```diff
@@ -158,7 +163,40 @@ yarn start
 open localhost:3000
 ```
 
+## Docker Quickstart
+To start Unity Catalog in Docker Compose in one command, install the latest version of [Docker Desktop](https://www.docker.com/products/docker-desktop/) and run the following:
```
jaceklaskowski, line 167: nit: Add a new line to separate the title from the text.

README.md
```diff
@@ -158,7 +163,40 @@ yarn start
...
+
+```sh
+docker compose up -d
+```
+
+This starts the Unity Catalog UI and backend. You can access the UI at http://localhost:3000 and the backend at http://localhost:8080.
```
jaceklaskowski, line 173: This "backend" caught my attention. Is this "the Unity Catalog server" perhaps? What else?

You may want to write "all the Unity Catalog components" and link it to the respective architecture page where they're described (if there's any = no idea).

README.md
```diff
@@ -158,7 +163,40 @@ yarn start
...
+
+To use the Unity Catalog CLI, attach to a shell in the backend container:
+
+```sh
+docker exec -it unitycatalog-backend-1 /bin/bash
```
jaceklaskowski, line 178: Oh, no, I'm sure that `unitycatalog-backend-1` will soon be the number 1 for many to replace with something more meaningful. Can you replace `backend-1` and likely make `unitycatalog` shorter (say `uc`)? 🙏

build.sbt
```diff
@@ -270,6 +271,15 @@ lazy val server = (project in file("server"))
       "com.h2database" %  "h2" % "2.2.224",
 
       "org.hibernate.orm" % "hibernate-core" % "6.5.0.Final",
+      // For loading properties from the environment
+      "uk.org.webcompere" % "lightweight-config" % "1.3.0",
+
+      // TODO: Determine if project will support MySQL jar; linter currently states
```
jaceklaskowski, line 277: Remove it until it's requested and worked out.

build.sbt
```diff
@@ -270,6 +271,15 @@ lazy val server = (project in file("server"))
...
+      // license is imcompatible
+      // "com.mysql" % "mysql-connector-j" % "8.4.0",
+
+      // For PostgreSQL backend
+      "org.postgresql" % "postgresql" % "42.6.0",
```
jaceklaskowski, line 282: I don't think we want this to be included by default. It's one of the many hibernate-supported databases and no need to make this any better than others (but including it in the build). UC does not need it, correct? If so, whoever wants PostgreSQL, they'll download the jars and make necessary changes. WDYT?

compose.yaml
```diff
@@ -0,0 +1,25 @@
+name: unitycatalog
+
+services:
+
+  backend:
```
jaceklaskowski, line 5: s/backend/server

README.md
```diff
@@ -65,8 +72,6 @@ You have to ensure that your local environment has the following:
 - Ensure the `JAVA_HOME` environment variable your terminal is configured to point to JDK17.
 - Compile the project using `build/sbt package`
```
jaceklaskowski, line 73: Is this still required?

jaceklaskowski:
`docker/README.md` needs some love too, doesn't it? Can you review it and make it up-to-date?

jaceklaskowski:
I couldn't build the images atm as it breaks on pulling delta-spark 3.2.1 😢  I'll do it again once all the comments are resolved.

```
#12 170.5 [error] coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
#12 170.5 [error]     io.delta:delta-spark_2.12:3.2.1:
#12 170.5 [error]         not found:
```

README.md
tnightengale: To clarify, does `./bin/start-uc-server` start all the components? Because that is the CMD for the image.

README.md
tnightengale: This is default from Docker Compose, which creates containers according to the following: <docker-compose-name>_<service-name>_<service_container_cardinatlity>.

I think it's better to give the complete name: `unitycatalog` and be more clear about the container throughout its lifecycle, than to use shorthands.

build.sbt
tnightengale: I think there's arguments for both.

Considering that a real database is required to run UC in production, I think it might be good practice to include the supported backends officially, by adding them as build dependencies. It would massively simplify using their respective JDBC jars, rather than trying to mount them to the container AND update the classpath.

We could also then _test_ the backends, using docker-compose with each database. I think this relates to my earlier comments about what databases are considering "officially supported", if any.

The alternative of not including the jars still requires the project to implicitly support different databases, but with less control over testing them. Which in turn, could lead to more questions from the community, because to "official" pattern to self-hosting UC in production is less defined.

For example, not including the jars would require us to modify the image's entrypoint to handle arbitrary ENV vars for classpath modifications and require the user to correctly mount the JDBC jar. 

In my opinion, there's no need to support and document that additional interface. It would be quite brittle, and basically untested. Which means it is not as reliable for self-hosting in production, compared to using a "officially supported" and included jar.

To summarize, both approaches have merit, but I think we should hear from other maintainers if the database interface should have "officially supported" connectors.

compose.yaml
tnightengale: Can you clarify this?

tnightengale:
> `docker/README.md` needs some love too, doesn't it? Can you review it and make it up-to-date?

Basically I think we could deprecate and remove `./docker` with this PR. Thoughts?

build.sbt
xx-db: @tnightengale 👋You are right, it was ultimately not merged because they are commonly used jars supported by Hibernate. So we included documentation to download the connector jars instead.

build.sbt
tnightengale: @tdas Can we get some guidance from yourself or another maintainer on if UC should offer support for "official" backends/databases and include those JDBC connector jars in the forthcoming official image? 

I think I see two paths here. Open to any thoughts on other directions as well.

**1. UC does not include "officially supported" databases to persist its application data. Users of the official image would be responsible for the following:**
a. Mount the correct JDBC jars to the container
b. Modify the relevant `classpath` to reference those jars
c. Mount their own `./etc` config files for hibernate to the container

**Pros**: Smaller image, less project dependencies.
**Cons**: Allows ambiguity about the "best" approach to self-hosting with an external database in production. No testing of different connections and a brittle, implicit interface, which means we could break things that users rely on. For example, reorganizing the config file directories would break a users mount in their deployment scripts.

**2. UC includes "officially supported" JDBC jars (eg. MySQL & Postgres) in the official image. Users of the image would then:**
a. configure their connection by passing env vars to the container (possible vars can be added to `.properties` config files with defaults and interpolated in the app OR
b. Mount their own `./etc` config files for hibernate to the container

**Pros**: We could test the different officially supported databases in CI using docker compose. 
**Cons**: Additional jars means a slightly bigger image size for folks who would only use one of the supported databases.

README.md
jaceklaskowski: All the default components of the UC Localhost Server, yes. I think this backend term comes more from Docker Compose where it's used. Correct?

README.md
jaceklaskowski: Let's rename `backend` to be `server` to give more clarity.

compose.yaml
jaceklaskowski: I'd much appreciate if you could rename `backend` with `server` to give this service a bit more UC-specific flavour.

README.md
jaceklaskowski: Is `build/sbt package` required to have a dockerized environment? I don't think so as `sbt package` is part of building the UC images.

dennyglee: Having the `build/sbt` package would be helpful for anyone building apps though.  Not a blocker though.

dennyglee:
> I couldn't build the images atm as it breaks on pulling delta-spark 3.2.1 😢 I'll do it again once all the comments are resolved.
> 
> ```
> #12 170.5 [error] coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
> #12 170.5 [error]     io.delta:delta-spark_2.12:3.2.1:
> #12 170.5 [error]         not found:
> ```

Yeah, that should be soon :)

dennyglee:
> > `docker/README.md` needs some love too, doesn't it? Can you review it and make it up-to-date?
> 
> Is there any reason to keep `./docker/` given the simplifications in this PR and the updated `./README.md`?

Possibly though I'm not personally hard set on this.  One of the reasons it *may* make sense is in the past it was difficult to get the docker to run in DockerHub unless it had its own folder and/or root repository.  This may have changed and/or perhaps I configured it wrong so I would glad to be corrected here, eh?!

jaceklaskowski:
I'd keep README.md's separate as the main one gets fairly large and should really be part of the official docs (under `docs` ofc). I'd simply leave a note about Unity Catalog in Docker in the main `README.md` with a link to the `docker/README.md`.

I'm re-wording the main `README.md` as we speak (so perhaps it's going to be as tiny that the other `docker/README.md` becomes useless). We'll see.

jaceklaskowski:
Pulled the PR and reviewed the changes, and I'm very very sure that we should keep all docker-related files in their own `docker` directory. Having a dockerized environment is very useful, but that's not the main goal of Unity Catalog and is simply a convenience.

That's why I'm requesting the following files to be in `docker` directory:

1. compose.psql.yaml
2. compose.yaml
3. Dockerfile

(perhaps even `ui/Dockerfile` to keep everything docker-related in one directory).

jaceklaskowski:
@dennyglee @tnightengale Got a question about how to contribute to the PR. I must admit that I've never done it before (to contribute to a PR with a PR) and...could you guide me how to do it properly?

I asked the same question on slack --> https://unitycatalog.slack.com/archives/C076YREKX8W/p1727615205499499?thread_ts=1727102843.375399&cid=C076YREKX8W

dennyglee:
> @dennyglee @tnightengale Got a question about how to contribute to the PR. I must admit that I've never done it before (to contribute to a PR with a PR) and...could you guide me how to do it properly?
> 
> I asked the same question on slack --> https://unitycatalog.slack.com/archives/C076YREKX8W/p1727615205499499?thread_ts=1727102843.375399&cid=C076YREKX8W

Oh generally, I create a fork of the OPR (original PR) and then submit the PR to that one.  e.g., 
- main is `unitycatalog/unitycatalog` branch `main`
- Original PR is `foo/unitycatalog` branch `bar`
- New PR is `blah/unitycatalog` branch `bar` where this PR is submitted to `foo/unitycatalog`

tnightengale:
> > > `docker/README.md` needs some love too, doesn't it? Can you review it and make it up-to-date?
> > 
> > 
> > Is there any reason to keep `./docker/` given the simplifications in this PR and the updated `./README.md`?
> 
> Possibly though I'm not personally hard set on this. One of the reasons it _may_ make sense is in the past it was difficult to get the docker to run in DockerHub unless it had its own folder and/or root repository. This may have changed and/or perhaps I configured it wrong so I would glad to be corrected here, eh?!

@dennyglee To clarify, what do you mean by 'run in Dockerhub'? As I understand it, DockerHub is just an image registry, like any other, and we would just be pushing a built image artifact to it.

On the topic of keeping everything in `./docker/*`, I would suggest we keep the files in the `unitycatalog/` project: this is the standard for most official images because it keeps things simple. The image is an artifact of the whole project; not a subcomponent.

tnightengale:
@dennyglee @xx-db I decoupled any MySQL/Postgres dependencies from the image.

This PR now responsible has a few tightly coupled responsibilities:
1. Unifying the multiple Dockerfiles into one image
2. Deprecating the previous Docker docs
3. Creating a simple interface to run UC in Docker, using `docker compose up`
4. Adding a standard Github Actions workflow to build the image

README.md
```diff
@@ -57,6 +57,9 @@ Let's take Unity Catalog for spin. In this guide, we are going to do the followi
   An example project is provided to demonstrate how to use the UC SDK for various assets
   as well as provide a convenient way to explore the content of any UC server implementation.
 
+> If you prefer to run Unity Catalog in Docker use `docker
+> compose up`. See the [Docker Quickstart](./docs/docker_compose_quickstart.md) for more details.
```
tnightengale, line 61:
```suggestion
> compose up`. See the [Docker Quickstart](./docs/docker_quickstart.md) for more details.
```

tnightengale:
@dennyglee Hey, rewriting commit history (see above) using the suggested [DCO](https://github.com/unitycatalog/unitycatalog/pull/485/checks?check_run_id=30920376331) commands, `git rebase HEAD~43 --signoff` and `git push --force-with-lease origin teghan/build_one_dockerfile` didn't seem to work and has left the history of this branch a hot mess of garbage.

How do you feel about me opening a new branch with one new signed and squashed commit containing the changes here?

PR #489, opened 2024-09-16
Build/release fixes

bijanhoule:
**PR Checklist**

**Description of changes**

Fix a few issues in the 0.2.0 release process.
- explicitly specify release settings on a few subprojects
- fix some cross-building issues
- move `controlApi` dependency to CLI project, it doesn't need to be published / marked as a spark dependency

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@ import java.io.File
 import Tarball.createTarballSettings
 import sbt.util
 import sbtlicensereport.license.{DepModuleInfo, LicenseCategory, LicenseInfo}
-import ReleaseSettings.{javaOnlyReleaseSettings, releaseSettings, rootReleaseSettings, skipReleaseSettings}
+import ReleaseSettings.*
 
 import scala.language.implicitConversions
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -227,7 +226,7 @@ lazy val apiDocs = (project in file("api"))
   .enablePlugins(OpenApiGeneratorPlugin)
   .settings(
     name := s"$artifactNamePrefix-docs",
-
+    skipReleaseSettings,
```
tdas, line 229: are all these explicit skipRelease settings needed?

project/ReleaseSettings.scala
```diff
@@ -40,12 +40,16 @@ object ReleaseSettings {
     autoScalaLibrary := false,
   )
 
+  lazy val scalaReleaseSettings = releaseSettings ++ Seq(
+    crossPaths := true,
+    releaseCrossBuild := true,
+  )
+
   lazy val releaseSettings = Seq(
```
tdas, line 48: could this be made private. so that no one uses this directly?

build.sbt
bijanhoule: I was running into failures without having them specifically added, yeah.  It looked like it was still attempting some part of the publish procuess but had only partial config

project/ReleaseSettings.scala
bijanhoule: good call, added

tdas (approved):

PR #498, opened 2024-09-21
docs: fix Daft integration guide

peter-gy:
**Description of changes**

Updates the documentation around integration with Daft so that the guide can be followed without running into errors.

- `daft.read_delta_lake` does not exist, the right API is `daft.read_deltalake` (see the related Daft docs [here](https://www.getdaft.io/projects/docs/en/latest/api_docs/doc_gen/io_functions/daft.read_deltalake.html#daft.read_deltalake))
- Daft must be installed including extras `unity` and `deltalake`, otherwise `daft.read_deltalake` will raise `ModuleNotFoundError`


MrPowers (approved):
LGTM

PR #566, opened 2024-10-04
Make source required

kriscon-db:
**PR Checklist**

**Description of changes**

Source was erroneously left off of the required list for CreateModelVersion when checked in.  In this PR we:

- Add source as required at at the API
- Add service handler checks for source existence
- Add an explicit test in the unit test to validate that source must be filled in.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/ModelService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/ModelService.java
@@ -137,6 +137,10 @@ public class ModelService {
           @AuthorizeKey(value = REGISTERED_MODEL, key = "model_name")})
           CreateModelVersion createModelVersion) {
     assert createModelVersion != null;
+    assert createModelVersion.getModelName() != null;
+    assert createModelVersion.getCatalogName() != null;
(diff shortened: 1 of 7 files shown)


server/src/main/java/io/unitycatalog/server/service/ModelService.java
```diff
@@ -137,6 +137,10 @@ public HttpResponse createModelVersion(@AuthorizeKeys({
           @AuthorizeKey(value = REGISTERED_MODEL, key = "model_name")})
```
kriscon-db, line 137: @creechy With source being moved to required, I don't need to alter these annotations at all right?  This are purely for ACL checks on the actual securable keys?

creechy: Yeah you shouldn't need to alter the annotations at all unless you want to include source in the ACL checking somehow.

kriscon-db: Nah.  I just wanted to double check on this.

tdas (approved):

rameshchandra (commented):
LGTM

PR #589, opened 2024-10-10
Add python sdk

### LINKED ISSUE #286: Add python sdk

kevinzwang:
**PR Checklist**

**Description of changes**

This PR adds a Python client SDK generated from the OpenAPI spec. 

Unlike [unitycatalog/unitycatalog-python](https://github.com/unitycatalog/unitycatalog-python), it does not have a vendor dependency with Stainless but instead uses the [openapi-generator](https://github.com/OpenAPITools/openapi-generator/) library. As a result, the library interface is different.

This PR adds a step to `build.sbt`, which runs when `build/sbt generate` is called and generates a Python client at `target/clients/python`. 

Follow-up from #309, resolves #286


kevinzwang:
@tdas new PR with unit tests and proper commit sign-off

build.sbt
```diff
@@ -223,6 +223,32 @@ lazy val client = (project in file("target/clients/java"))
...
+    (Compile / compile) := ((Compile / compile) dependsOn generate).value,
+
+    // OpenAPI generation specs
+    openApiInputSpec := (file(".") / "api" / "all.yaml").toString,
+    openApiGeneratorName := "python",
+    openApiOutputDir := (file("target") / "clients" / "python").toString,
+    openApiPackageName := s"$artifactNamePrefix",
+    openApiAdditionalProperties := Map(
+      "packageVersion" -> s"${version.value.replace("-SNAPSHOT", ".dev0")}",
+      "library" -> "urllib3",
```
vikrantpuppala, line 240: I think `urllib3` is anyway the default library, we could remove this?

tests/python/conftest.py
```diff
@@ -0,0 +1,93 @@
...
+            print(">> Stopped the server")
+        except ProcessLookupError:
+            # Process already terminated
+            pass
+
+
+@pytest.fixture(scope="session")
+def api_client():
+    config = unitycatalog.Configuration(
+        host = "http://localhost:8080/api/2.1/unity-catalog"
```
vikrantpuppala, line 53: I think we should be hitting 8081?

vikrantpuppala, line 53: the param name is a bit misleading, reads `host` but actually expects the base url, but i don't think we can do anything about this

tests/python/test_tutorial.py
vikrantpuppala, line 1: I think this is really great but I really wish we could something like the [BaseCRUD tests](https://github.com/unitycatalog/unitycatalog/tree/main/server/src/test/java/io/unitycatalog/server/base) i.e. perform a complete CRUDL lifecycle for each asset. We could probably also split it into multiple files to keep if that makes it cleaner

.github/workflows/python-unit-tests.yml
```diff
@@ -0,0 +1,45 @@
+name: Python Unit Tests
```
vikrantpuppala, line 1:
```suggestion
name: Python Client Tests
```

vikrantpuppala, line 1: probably also better to add a new job in `unit-tests.yml` than to create a new workflow

build.sbt
```diff
@@ -223,6 +223,32 @@ lazy val client = (project in file("target/clients/java"))
...
+  .settings(
+    name := s"$artifactNamePrefix-client",
+    commonSettings,
+    (Compile / compile) := ((Compile / compile) dependsOn generate).value,
+
+    // OpenAPI generation specs
+    openApiInputSpec := (file(".") / "api" / "all.yaml").toString,
+    openApiGeneratorName := "python",
+    openApiOutputDir := (file("target") / "clients" / "python").toString,
+    openApiPackageName := s"$artifactNamePrefix",
```
vikrantpuppala, line 237: this should probably be $artifactNamePrefix_client?

build.sbt
```diff
@@ -223,6 +223,32 @@ lazy val client = (project in file("target/clients/java"))
...
+    openApiAdditionalProperties := Map(
```
vikrantpuppala, line 238: are there any more additional properties that we can set so that the generated client is more ready to be published to PyPi? perhaps to make the setup.py more PyPi ready?

vikrantpuppala: also happy to hear if there is a better way to write setup.py

build.sbt
```diff
@@ -223,6 +223,32 @@ lazy val client = (project in file("target/clients/java"))
     }
   )
 
+lazy val pythonClient = (project in file("target/clients/python"))
+  .enablePlugins(OpenApiGeneratorPlugin)
+  .settings(
+    name := s"$artifactNamePrefix-client",
```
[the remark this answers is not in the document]
kevinzwang: I'm not sure, if you think it could I can change the name though

tests/python/conftest.py
kevinzwang: It looks like both work for some reason. Either way I will change it to 8081

tests/python/test_tutorial.py
kevinzwang: Thanks. I don't think I have the bandwidth to do that at the moment, is it a blocking ask? What is currently in the PR should not break any existing workflows.

.github/workflows/python-unit-tests.yml
kevinzwang: Aside from being tests, I wouldn't say these tests are related to the unit tests, nor do they share any build artifacts. I personally feel that it would be more helpful and clear to separate them, but I am also fine with running them with the other unit tests if you still think so.

build.sbt
kevinzwang: I've named it `$artifactNamePrefix` to match the current name of the [published package](https://pypi.org/project/unitycatalog/). Should we change it?

build.sbt
kevinzwang: I couldn't find any openapi generator options to update the other `setup.py` values. I guess this is just a limitation of the tool. I imagine that a contributor would either manually do the various steps, including updating setup.py, before publishing, or we could have a github action that does it, which can be a follow-up PR.

kevinzwang:
@vikrantpuppala took a pass over your review. Hope to get this in soon!

build.sbt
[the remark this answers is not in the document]
vikrantpuppala: I think its probably a good idea to have different names, however, i'll let @tdas take the call on this one.

build.sbt
vikrantpuppala: I think it then makes sense to leave it as is. cc: @tdas

vikrantpuppala (approved):

tests/python/test_tutorial.py
vikrantpuppala: We can do this as a follow up, thank you!

tests/python/conftest.py
vikrantpuppala: Yeah, the server is actually hosted on 8080, however there's a URL transcoder on 8081 that forwards to 8080. Probably always a better idea to use via 8081

vikrantpuppala:
@kevinzwang, thanks for making the changes, there's a failure in the python tests due to the ordering in lists, once we fix that, we should be good to go!

build.sbt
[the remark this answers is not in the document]
vikrantpuppala: To come back on this, let's go ahead and change this to `unitycatalog-python-client` and maybe we can add a comment here that this is not the name that gets published to pypi

kevinzwang:
@vikrantpuppala tests fixed!

.github/workflows/python-client-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+        java-version: '17'
+        distribution: 'temurin'
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.9'
+
+    - name: Generate Python client library
+      run: build/sbt generate
```
tdas, line 30: do we need to generate all the clients for this?

tests/python/test_tutorial.py
```diff
@@ -0,0 +1,277 @@
+import pytest
```
tdas, line 1: these tests are meant to be for the python sdk, not just for the tutorial. so it does not make sense to name it based on the tutorial (btw, its also confusing to have tests/test_tutorial.py and tests/python/test_tutorial.py). So I think we should to the following:

- clients/python/tests/ - all the test files will be here
- clients/python/README - explains how where the python is generated and how to run the python client tests.

tests/python/conftest.py
```diff
@@ -0,0 +1,93 @@
+import pytest
+import subprocess
+import os
+import time
+import requests
+import signal
+
+import unitycatalog
+
```
tdas, line 9: please add docs on what is in this file. for example, I am confused, this does not have any test in it, only the utility functionalities for pytest to use .. right? if so, why is named conftest.py? is that a standard in python world?

.github/workflows/python-client-tests.yml
kevinzwang: No, but I'm not sure what the command is to only generate the python client. What would that be?

tests/python/test_tutorial.py
kevinzwang: Would it make sense to put these in `clients/`? It is not a directory in this repo right now and all of the clients are actually generated in `target/clients/` which is gitignored.

tests/python/conftest.py
tdas: oh wow. totally python noob i am :)

tests/python/test_tutorial.py
tdas: i actually want to create this directory, because otherwise, its actually very confusing to find where the clients are. unless someone really knows where to look in build.sbt, i think its hard to understand what is going on. so i actually want to make it obvious by having a top-level directory called clients, and then slowly start using them for stuff. so lets start that right now :D 

maybe we should generate the code in clients/python/target/ ? is that doable in this PR?

kevinzwang: Moved tests and changed client generation to `clients/python`!

kevinzwang: I also separated the tests out into separate files based on the part of the API

.github/workflows/python-client-tests.yml
```diff
@@ -0,0 +1,45 @@
+name: Python Client Tests
+
+on:
+  push:
+    branches: [ "main" ]
```
tdas, line 5: i think we need these tests to run on all branches. 

i see that there are other workflows with the same problem. @vikrantpuppala can you fix them?

.github/workflows/python-client-tests.yml
tdas: build/sbt pythonClient/generate

tdas: can we move all these steps into a single python script like `clients/python/run-tests.py`?
so that its easy to manually run all the tests as well just by running this?

kevinzwang: Several of these steps (creating the virtual env + installing dependencies) should only be run once for setup, and may also depend on the configuration of python environments you have locally. Because of that, I don't think it makes sense to put them into a single script.

I've already documented the steps in the readme and I believe they should be fairly straightforward to run.

.github/workflows/python-client-tests.yml
vikrantpuppala: the lines below i.e. 
```
pull_request:
    branches: [ "main" ]
```
will run these tests whenever a PR is created into main. would that not suffice?

tdas: we are backporting stuff in branch-0.2 with PRs. in those PRs, these tests will not run. that means that we are going to loose test coverage before we release 0.2.x.

.github/workflows/python-client-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+      with:
+        python-version: '3.9'
+
+    - name: Generate Python client library
+      run: build/sbt pythonClient/generate
+
+    - name: Install dependencies
+      run: |
+        python -m venv .venv
+        source .venv/bin/activate
```
tdas, line 35: why do we need this?

.github/workflows/python-client-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.9'
+
+    - name: Generate Python client library
+      run: build/sbt pythonClient/generate
+
+    - name: Install dependencies
+      run: |
```
tdas, line 33: I agree that some steps here are one time setups and should not be a in a script. thats what the `pip installs` are. but there are like 3 steps that should be done every time a test has to be run. 
- build/sbt pythonClient/generate
- pip install clients/python/target/
- pytest clients/python/tests
these can be combined into a single python script. when a developer is changing the api, then having to run 3 commands every time is annoying. iteratively running update api + run one command is better... isnt it?

build.sbt
```diff
@@ -223,6 +223,32 @@ lazy val client = (project in file("target/clients/java"))
...
+  .settings(
+    // name of the generation step. See `openApiPackageName` for the actual Python package name
+    name := s"$artifactNamePrefix-python-client",
+    commonSettings,
+    (Compile / compile) := ((Compile / compile) dependsOn generate).value,
+
+    // OpenAPI generation specs
+    openApiInputSpec := (file(".") / "api" / "all.yaml").toString,
+    openApiGeneratorName := "python",
+    openApiOutputDir := (file("clients") / "python" / "target").toString,
```
tdas, line 237: now that i think about it, this does not need to be named target at all. if we were not generating this on the fly, rather checking in the code, we would have named this directory `src` or something like that right? lets do that?

.github/workflows/python-client-tests.yml
vikrantpuppala: sure, i would say we could expand the pull request to be main + release branches

build.sbt
vikrantpuppala: The java client had a similar debate. We used to have it under src and commit to the repo but it cluttered the repo quite a bit and we ultimately decided to instead move to target since it was generated code that could be always generated on the fly. Having this generated code in the repo shows up as additional changes in PR etc and doesn't really add any value by actually being present in the repo imo.

.github/workflows/python-client-tests.yml
vikrantpuppala: #624 to track

.github/workflows/python-client-tests.yml
kevinzwang: It's not needed, but generally it's good practice to install packages into a virtual environment in Python

.github/workflows/python-client-tests.yml
kevinzwang: Sounds good, I added a simple script to run those three commands

build.sbt
tdas: yeah. i agree with not having the code checked in. that's not my proposal at all. I am just saying that do we need to generate the code inside a directory named target which typically used for generated binaries? 

either ways, i should have mentioned.. this is a nit. i dont have a strong opinion, and can be changed later as well easily (assuming dev instructions in README dont have to change for this).

clients/python/README.md
```diff
@@ -25,4 +25,6 @@ Client tests use the `pytest` library. To run them:
 3. Run tests
     ```sh
     pytest clients/python/tests
-    ```
\ No newline at end of file
+    ```
+
+You may also run the provided `run-tests.sh` script which will, generate the client library, install it, and then run the tests.
```
tdas, line 30: why not replace all the steps above with this one command?

.github/workflows/python-client-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+        python -m venv .venv
+        source .venv/bin/activate
+        pip install requests
+        pip install pytest
+    - name: Install unitycatalog Python package
```
tdas, line 38: please use the script that you have added. if we dont use the script for actual CI testing, then how are we going to find out if that script accidentally breaks?

clients/python/run-tests.sh
```diff
@@ -0,0 +1,3 @@
+build/sbt pythonClient/generate
```
tdas, line 1: dont you need `!/bin/sh` thing? or should dev run it as `sh run-tests.sh`. this should be written in the readme.

clients/python/README.md
kevinzwang: Fixed both

tdas (approved):
@kevinzwang Thank you so much for patiently working through all the review. I have been trying to get to point that is most developer friendly. Sometimes like devs are like users, e.g. non-python devs shouldnt have to care about how the python stuff works internally as long as they can easily check whether their changes in other places are breaking anything or not. And if it does break in CI, they can quick repro the issue without having to understand how the entire python stuff is set up. This means, that we have design code, scripts, everything like it's a clear user interface that abstracts out internal details.

We are good to go now :)

tdas:
MERGED! I was waiting for tests to finish :)

tdas:
Yes, the next step would be to build the set up to publish. 

But I dont think we can do 0.2, because uc 0.2 on the java side has already been released, and since then there has been api changes. so if we release python 0.2 that has different apis than java 0.2, that will be confusing to some.

But maybe we can backport this to branch-0.2, and publish from there?! I have to think about it a bit. 

In the meantime, we do need the setup to do pypi publishing.

tdas:
I had done pypi publishing setup for delta a long time back. 
maybe this can help - https://github.com/delta-io/delta/blob/master/setup.py

kevinzwang:
> In the meantime, we do need the setup to do pypi publishing. can you help with that?

I believe we can follow the instructions here: https://packaging.python.org/en/latest/tutorials/packaging-projects/

Do you know who published the current one? They probably know more about the permissions as well as what the setup was before.

kevinzwang:
> But maybe we can backport this to branch-0.2, and publish from there?! I have to think about it a bit.

Unless the next version of uc is coming soon, would be great to publish a 0.2 version of the SDK first!

BenWilson2:
With the expanding python ecosystem for unity catalog, it would be ideal to have all official integration follow a common namespace. To this end, can we look into:
1. Materializing the generated code into the repo? This way we can setup and configure the namespace sharing requirements and allow for additional functionality to be added to the Python SDK on top of the generated code based off of the OpenAPI specs.
2. Configure the SDK regeneration to populate a src/ subdirectory within the SDK repository directory to simplify the release version updates (directory replacement strategy seems less daunting than handling large-scale merge conflicts from release to release)?

cc @dennyglee @tdas

PR #590, opened 2024-10-10
Replace string concat with logger template

ebyhr:
**PR Checklist**

**Description of changes**

Replace string concat with logger template


vikrantpuppala (approved):

tdas:
@ebyhr is there any benefit to using logger template over string concal

vikrantpuppala:
@tdas i think the biggest benefit is that the logger template will not always evaluate the string i.e. if log level is error then the string expressions inside LOGGER.info will not be evaluated vs in concat, this evaluation is always done irrespective of log level.

ebyhr:
@vikrantpuppala is right, that's the main reason.

tdas (approved):

PR #607, opened 2024-10-15
Generalise server and cli for all clouds

vikrantpuppala:
**PR Checklist**

**Description of changes**

Generalise FileUtils for all clouds instead of just local FS and AWS by using the iceberg FileIO classes


server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -41,34 +41,25 @@ public TableInfo getTableById(String tableId) {
...
-        tableInfo.setSchemaName(schemaInfoDAO.getName());
-        tableInfo.setCatalogName(catalogInfoDAO.getName());
-        tx.commit();
-        return tableInfo;
-      } catch (Exception e) {
-        if (tx != null && tx.getStatus().canRollback()) {
-          tx.rollback();
-        }
-        throw e;
+      TableInfoDAO tableInfoDAO = session.get(TableInfoDAO.class, UUID.fromString(tableId));
```
vikrantpuppala, line 44: removed transaction for getTable

server/src/main/java/io/unitycatalog/server/persist/utils/FileUtils.java
```diff
@@ -181,20 +65,24 @@ private static URI adjustFileUri(URI fileUri) {
...
 
-  public static String convertRelativePathToURI(String url) {
-    if (url == null) {
-      return null;
-    }
-    if (isSupportedCloudStorageUri(url)) {
-      return url;
-    } else {
-      return adjustFileUri(createURI(url)).toString();
+  public static String toStandardizedURIString(String inputPath) {
```
sumeet-db, line 68: Add some comments for tis method

server/src/main/java/io/unitycatalog/server/persist/utils/FileUtils.java
```diff
@@ -181,20 +65,24 @@ private static URI adjustFileUri(URI fileUri) {
...
+    try {
+      // Check if the path is already a URI with a valid scheme
+      URI uri = new URI(inputPath);
+      // If it's a file URI, standardize it
+      if (uri.getScheme() != null) {
+        return switch (uri.getScheme()) {
+          case "file" -> adjustLocalFileURI(uri).toString();
+          case Constants.URI_SCHEME_S3, Constants.URI_SCHEME_ABFS, Constants.URI_SCHEME_ABFSS, Constants.URI_SCHEME_GS ->
```
sumeet-db, line 76: When would the scheme be S3? And why do we throw an exception for it?

server/src/main/java/io/unitycatalog/server/service/iceberg/FileIOFactory.java
```diff
@@ -42,6 +42,9 @@ public FileIOFactory(CredentialOperations credentialOps) {
 
   // TODO: Cache fileIOs
   public FileIO getFileIO(URI tableLocationUri) {
+    if (tableLocationUri.getScheme() == null) {
```
sumeet-db, line 45: When is it null?

sumeet-db:
Add a more elaborate PR desc. please

server/src/main/java/io/unitycatalog/server/persist/utils/FileUtils.java
```diff
@@ -1,178 +1,62 @@
...
   private static final Logger LOGGER = LoggerFactory.getLogger(FileUtils.class);
   private static final ServerProperties properties = ServerProperties.getInstance();
+  private static final CredentialOperations credentialOps = new CredentialOperations();
+  private static final FileIOFactory fileIOFactory = new FileIOFactory(credentialOps);
 
   private FileUtils() {}
 
   private static String getStorageRoot() {
-    return properties.getProperty("storageRoot");
+    // Use local tmp directory as default storage root
```
sumeet-db, line 27: Why did we change this?

server/src/main/java/io/unitycatalog/server/persist/utils/FileUtils.java
```diff
@@ -1,178 +1,62 @@
...
-                    throw new RuntimeException("Failed to delete " + path, e);
-                  }
-                });
-      }
-    } else {
-      throw new IOException("Directory does not exist: " + dirPath);
-    }
-  }
-
-  private static URI modifyS3Directory(URI parsedUri, boolean createOrDelete) {
```
sumeet-db, line 120: Did we move this function to some other file? Unable to find it in the PR

server/src/main/java/io/unitycatalog/server/persist/utils/FileUtils.java
vikrantpuppala: The scheme would be s3 when the UC OSS server is configured to work with a s3 bucket as a storage root. exception is thrown for any scheme other than file, s3, abfss, gs.

server/src/main/java/io/unitycatalog/server/service/iceberg/FileIOFactory.java
vikrantpuppala: when the path is provided without in a non URI format i.e. "/local/path/to/file" as opposed to "file:///local/path/to/file"

server/src/main/java/io/unitycatalog/server/persist/utils/FileUtils.java
vikrantpuppala: When the server storage root wasn't configured, it would just create a directory called "null" and perform all operations in that dir. This will atleast have all operations done in the tmp dir.

server/src/main/java/io/unitycatalog/server/persist/utils/FileUtils.java
vikrantpuppala: This class used to only support local and s3 file systems by having code specific to these FSs. I've made changes to instead use the iceberg FileIO that the server was already using for credential vending.

sumeet-db:
Apologies for not going through the PR deeply to find answers myself but still a bit unclear on the primary purpose of the PR.
1. Why is it needed for staging table? 
2. What was missing in the existing implementation? 
3. How does this PR address (2)? Looking to understand the `before` and `after` for this PR.

vikrantpuppala:
1. The staging table flow needs to create a directory to pass on to the delta. This directory creation flow is done by FileUtils which is being refactored in this PR.
2. The existing implementation of FileUtils only supported file operations on either local fs or s3 and had code specific to these two fs.
3. This PR adds support for azure and google cloud by using the FileIO classes from iceberg and also cleans up the code to be generic across all fs.

PR #648, opened 2024-10-22
Format top-level documentation pages via markdownlint

### LINKED ISSUE #647: Format top-level documentation pages via markdownlint

Based on the discussion in #572 the top-level documentation pages (`deployment.md`, `index.md` and `quickstart.md`) should be formatted with [markdownlint](https://github.com/DavidAnson/markdownlint/).

The following should be included:

* Formatting of the top-level documentation pages, as described above.
* A custom markdownlint configuration file to include formatting rules different from the default ones.
* Minimal changes for the user-facing documentation, when needed.

**Additional context**

See #572 for the motivation.

HorizonNet:
**PR Checklist**

**Description of changes**

Update the formatting of the top-level documentation pages by using markdownlint. Included a custom markdownlint configuration file, which adjusts the default rules to our needs. User-facing changes have been kept minimal (only in `index.md`).

Resolves #647


docs/deployment.md
```diff
@@ -7,38 +7,46 @@ This guide outlines how to deploy the Unity Catalog server.
...
+```
 
 ### Unpacking the tarball
 
 - The tarball generated in the `target` directory can be unpacked using the following command:
-  ```sh
-  tar -xvf unitycatalog-<version>.tar.gz
-  ```
+
+```sh
```
dennyglee, line 19: @HorizonNet I like the code format being indented as it falls under the bullet points.  Is it possible to modify `markdownlint.yaml` to allow for this?

HorizonNet: @dennyglee Done. It is possible by default. It just looked off to me. Changed it back to be indented. Please take a look again.

dennyglee: Sorry, allow me to clarify - the docs.unitycatalog.io deployment page has the following look:

### Option A: Code block aligned with text of bullet point
Screenshot 2024-10-23 at 22 13 48

> Notice how the code blocks are indented so they are aligned with the text after the bullet point as those code blocks are associated with that bullet point.


### Option B: Code block aligned with the bullet point itself
In the most recent pull, here is what the same page looks like.
 
Screenshot 2024-10-23 at 22 14 14

> Notice how the code block is aligned to the bullet point itself, not the text.  It implies that this is a new line vs. belonging to the previous bullet point.


### Question
1. Does it make sense to have code blocks associated/aligned to the text (Option A) vs. the bullet point (Option B)?  
2. If the former makes sense, can we configure markdownlint to accept this type of formatting? 

> Note, I don't mind if we state that we should do Option B vs. Option A.  It's just that we started with Option A and I wanted us to be deliberate with this decision.

HorizonNet: Ah, got it. Thanks for the example. Option B feels better for the author, but Option A looks better for the reader. I would propose going with Option A. I've updated the PR accordingly. I also updated the README to include this decision for later.

dennyglee (approved):
Thanks @HorizonNet !

PR #660, opened 2024-10-29
Add escape sequences for GenAI generated python code execution

BenWilson2:
**PR Checklist**

**Description of changes**

For our 'general code execution function' example:

```python
def execute_python_code_function(code: str) -> str:
    """
    Runs some code that a GenAI LLM came up with
    """
    sys_stdout = sys.stdout
    redirected_output = StringIO()
    sys.stdout = redirected_output
    exec(code)
    sys.stdout = sys_stdout
    return redirected_output.getvalue()

```

Many LLMs will create code that is in pseudo markdown format, filled with invalid python and invalid SQL reserved terms. In order to actually execute these statements as literals in the JSON-to-SQL function input, we need to escape some of these special characters and ensure that when the argument (code block) value is provided to the UC Python function, it is capable of being executed. 

Without this sanitization, simple things like single quotes, newlines, carriage returns, tabs, and other reserved terms will cause SQL parsing errors that are VERY challenging to debug. 

Added perhaps too many test conditions just to be safe.


BenWilson2:
@serena-ruan Please take a look :D

ai/core/src/unitycatalog/ai/core/databricks.py
```diff
@@ -844,6 +849,8 @@ def get_execute_function_sql_command(function: "FunctionInfo", parameters: Dict[
                         param_value = base64.b64encode(param_value).decode("utf-8")
                     # Use ubbase64 to restore binary values.
                     arg_clause += f"unbase64('{param_value}')"
+                elif param_info.type_name == ColumnTypeName.STRING:
+                    arg_clause += f"'{param_value}'"
```
serena-ruan, line 853: Is this needed? Line 871 converts everything else to string :)

ai/core/src/unitycatalog/ai/core/utils/function_processing_utils.py
```diff
@@ -282,3 +282,29 @@ def supported_function_info_types():
...
+    Args:
+        function_params: A dictionary of function parameters.
+
+    Returns:
+        A sanitized dictionary of function parameters.
+    """
+    sanitized_params = {}
+    for key, value in function_params.items():
+        if isinstance(value, str):
+            # Escape single quotes, backslashes, and control characters that would otherwise break Python code execution
```
serena-ruan, line 300: wondering if we could just use `"""{value}"""`

serena-ruan: This would convert
```
"what's databricks?" --> "what''s Databricks?" --> "'what''s Databricks?'" in the SQL command
```

BenWilson2: I just ran through testing that and anything that has a reserved character in generated markdown rendering code like a \n or a \t for formatting of generated code with long single lines will throw syntax errors or invalid line continuation errors for logical execution blocks. 
That being said, adding external triple quote wraps around the entire arg for a string provides another safeguard against some edge case syntax errors without requiring us to use something more intrusive :)

ai/core/src/unitycatalog/ai/core/databricks.py
BenWilson2: lol yeah not needed at all. oof

serena-ruan (approved):
LGTM!

BenWilson2:
@dennyglee We're good on this one. Thanks bud!

dennyglee (approved):

PR #661, opened 2024-10-29
Fix UC function schema when no params are needed

serena-ruan:
**PR Checklist**

**Description of changes**
With the latest langchain version, if the UC function has no params then the schema is not correctly parsed. We should use `create_model` instead of `BaseModel`, added test for this.

Test validation: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/11567043081/job/32198004224

### DIFF
--- a/ai/core/tests/core/databricks/test_databricks_integration_tests.py
+++ b/ai/core/tests/core/databricks/test_databricks_integration_tests.py
@@ -381,7 +381,7 @@ def test_create_function_without_replace(client: DatabricksFunctionClient):
         # Attempt to create the same function again without replace
         with pytest.raises(
             Exception,
-            match=f"Cannot create the function `{CATALOG}`.`{SCHEMA}`.`simple_func` because it already exists",
+            match=f"`{CATALOG}`.`{SCHEMA}`.`simple_func` because it already exists",
         ):
             client.create_python_function(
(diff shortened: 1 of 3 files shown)


serena-ruan:
@BenWilson2 for review :D

ai/core/tests/core/databricks/test_databricks_integration_tests.py
```diff
@@ -381,7 +381,7 @@ def simple_func(x: int) -> str:
         # Attempt to create the same function again without replace
         with pytest.raises(
             Exception,
-            match=f"Cannot create the function `{CATALOG}`.`{SCHEMA}`.`simple_func` because it already exists",
+            match=f"`{CATALOG}`.`{SCHEMA}`.`simple_func` because it already exists",
```
serena-ruan, line 384: Change this because the error message changed:
```
[ROUTINE_ALREADY_EXISTS] Cannot create the routine ...
```
This is spark error so we have no control, drop matching them

BenWilson2 (approved):
Solid change! Thanks @serena-ruan :)

BenWilson2:
@dennyglee this is good to go for a bug fix

dennyglee (approved):

PR #677, opened 2024-11-01
Add GitHub Actions workflow for documentation deployment

paulooctavio:
**PR Checklist**

**Description of changes**

Add GitHub Actions workflow for documentation deployment.


docs/requirements.txt
```diff
@@ -0,0 +1,5 @@
+mkdocs==1.5.3
```
HorizonNet, line 1: This file duplicates the existing `requirements-docx.txt` file in the root directory and also uses older versions (at least for MkDocs). I would propose to use the existing one for now.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,48 @@
...
+
+on:
+  workflow_dispatch:
+
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: gaurav-nelson/github-action-markdown-link-check@v1
```
HorizonNet, line 11: I like the inclusion of this one. I would recommend setting it to an exact version (also for the other actions) to avoid breaking changes in a minor update (happens with super-linter f.e. all the time). For this action it would be 1.0.15. An updated update could be done via Dependabot.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,48 @@
...
+          mkdocs build
+
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: JamesIves/github-pages-deploy-action@v4
+        with:
+          folder: site
+          branch: gh-pages
+          git-config-name: "Github Action"
+          commit-message: "doc update for tag `${{ github.event.client_payload.tag || github.ref_name }}`"
```
HorizonNet, line 48: NIT, please add an empty line at the end. Sometimes it messes things up or leads to additional changes if an Editorconfig gets introduced at some point.

HorizonNet (changes requested):
Looks good overall. Left two comments with suggestions and a NIT.

docs/requirements.txt
paulooctavio: Sorry, missed this one, fixed it in [843dfb6](https://github.com/unitycatalog/unitycatalog/pull/677/commits/843dfb6e441d3e635954cdd9cbaa6427bf9e47c6)

.github/workflows/docs.yml
paulooctavio: Fixed in [843dfb6](https://github.com/unitycatalog/unitycatalog/pull/677/commits/843dfb6e441d3e635954cdd9cbaa6427bf9e47c6)

.github/workflows/docs.yml
paulooctavio: Thanks! Fixed in [ac467fa](https://github.com/unitycatalog/unitycatalog/pull/677/commits/ac467fa5cf442e7150090b1d4a3e019369395b2e)

HorizonNet (approved):

dennyglee:
Thanks for your patience @paulooctavio and appreciate you reviewing it @HorizonNet.  I just wanted to confirm that this doc deployment doesn't impact the Swagger doc?  When I ran it, I was running into issues with it BUT perhaps that's on my side.   Thanks!

paulooctavio:
Hi @dennyglee! Thanks for you patience as well, how is it impacting Swagger doc? I'm asking so I can check on my side too. Thanks!

dennyglee:
Oh sorry for missing this @paulooctavio - when I ran the GitHub Action, it _seemed_ to break the swagger deployment vs. when I ran this doc flow manually.  Again, this could be very much user (i.e., me) error here.

paulooctavio:
Thanks for clarifying that, @dennyglee! Quick question though, how is the docs deployment affecting the Swagger deployment? Aren’t they separate workflows? Also, when you mention running the GitHub Action, are you referring to triggering the workflow through the Actions UI? Just want to make sure I’m clear on how to reproduce it.

dennyglee:
They are both piggybacking off docs.unitycatalog.io directly where once we manually (or use GitHub actions) to publish the docs, they will push to https://docs.unitycatalog.io/ while the swagger docs are published to https://docs.unitycatalog.io/swagger-docs/ using this [workflow](https://github.com/unitycatalog/unitycatalog/blob/main/.github/workflows/publish-swagger-docs.yml).  Saying this, this may be my own configuration so if you cannot repro, then I think we can just close my issue and use your action to start publishing docs going forward, eh?!

paulooctavio:
Hi @dennyglee! Sorry for the very late reply. Yes, unfortunately I'm not being able to reproduce the swagger deployment error, but I believe we can use the action for the docs deployment.

dennyglee:
Thanks @paulooctavio  - let me do one more run shortly to validate - much appreciated!

HorizonNet:
Sorry for the late reply. I wasn't able to reproduce the issue, but I might have an idea what could cause the issue. The new workflow deploys everything from the `site` directory onto the `gh-pages` branch and might overwrite/remove the `swagger-docs` directory created by the other workflow.

The "save" ways I currently see is either

1. Use `mkdocs gh-deploy` directly in the new workflow instead of the GitHub Pages action as this is probably the way everything has been deployed so far.
2. Combine Swagger and Mkdocs by using a [plugin](https://github.com/blueswen/mkdocs-swagger-ui-tag) and deploy both at the same time.

I would tend towards option 1 as this is closest to what have been used so far, so shouldn't contain any surprises.

PR #593, opened 2024-10-10
Enhance token endpoint by setting cookie

sudharshanraja-db:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [x] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

- Set token as only http only cookie in the response header if query param ext=cookie is set
- Cookie default age is set to 5 days and this is a configurable parameter in server.properties
- Authorize user with token in cookie if bearer token is missing

Closes #542

Issue #542, opened 2024-10-01
Enhance /tokens endpoint to optionally set cookie


creechy:
**Is your feature request related to a problem? Please describe.**

Currently the UI keeps the access-token for the logged in user as part of the application state. There are at least a couple of concerns with this
- Malicious javascript could potentially retrieve the access-token
- The application state is lost on browser refresh, requiring the user to re-authenticate

**Describe the solution you would like**

Extend the `/tokens` endpoint to also set a Cookie with the generated access-token. And extend server authorization to look at the cookie in addition to the `Authorization` header.

The pieces are as follows:

1.  Extend `AuthService.grantToken()` to look for the request query parameter `ext=cookie`. If the query parameter and value exists, the `grantToken()` should set the cookie `UC_TOKEN` with the value of the access token it has generated. This should be an `HttpOnly` token so that Javascript does not have access to it.
2. Extend `AuthDecorator.serve()` so that if it does not find an `Authorization` header, it then checks for the `UC_TOKEN` cookie to get the access-token to validate and use as for the authorized principal for the request. 

**Additional context**

PR #678, opened 2024-11-01
Expire  cookie  that was set as part of token endpoint during logout

sudharshanraja-db:
**PR Checklist**

**Description of changes**

This change is to expire cookie during logout by exposing logout endpoint  , feature that was added as part of #542  and implemented as part of PR #593


server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -167,17 +168,37 @@ public HttpResponse grantToken(
...
 
     return HttpResponse.ofJson(responseHeaders.build(), response);
   }
 
+  @Get("/logout")
+  public HttpResponse logout(HttpRequest request) {
+    String authorizationCookie =
+        request.headers().cookies().stream()
+            .filter(c -> c.name().equals(AuthDecorator.UC_TOKEN_KEY))
+            .map(Cookie::name)
```
jamieknight-db, line 188: @creechy @xx-db should this be `Cookie::value` like the change we made in AuthDecorator?

xx-db: yes, this needs to be fixed in both places

jamieknight-db: @sudharshanraja-db for reference, the change we made is in my[ UI PR here](https://github.com/unitycatalog/unitycatalog/pull/666/files#).

jamieknight-db: It also looks like there are a lot of old commits included. I got myself into a similar situation and had to sync my forked repo with main repo, then reset HEAD to last commit.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -167,17 +168,37 @@ public HttpResponse grantToken(
...
                 ServerProperties.getInstance().getProperty("server.cookie-timeout", "P5D");
             Cookie cookie =
-                Cookie.secureBuilder(AuthDecorator.UC_TOKEN_KEY, accessToken)
-                    .path("/")
-                    .maxAge(Duration.parse(cookieTimeout).getSeconds())
-                    .build();
+                createCookie(
+                    AuthDecorator.UC_TOKEN_KEY,
+                    accessToken,
+                    "/",
```
jaceklaskowski, line 174: Can we give this a name?

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -167,17 +168,37 @@ public HttpResponse grantToken(
...
+                    Duration.parse(cookieTimeout).getSeconds());
```
jaceklaskowski, line 175: Can this be moved to be part of `createCookie`?

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -167,17 +168,37 @@ public HttpResponse grantToken(
...
   }
 
+  @Get("/logout")
+  public HttpResponse logout(HttpRequest request) {
+    String authorizationCookie =
+        request.headers().cookies().stream()
+            .filter(c -> c.name().equals(AuthDecorator.UC_TOKEN_KEY))
+            .map(Cookie::name)
+            .findFirst()
+            .orElse(null);
```
jaceklaskowski, line 190: Why is this needed since you're doing `if` afterwards?

What do you think of getting rid of `orElse(null)` and chaining further to avoid `if` (that should really be called when there's non-`null` value and hence when `orElse` would not be needed)? It makes a lot of sense, doesn't it?

jaceklaskowski (changes requested):

jaceklaskowski:
I noticed you're on `sudharshanraja-db:main` branch. Create a feature branch instead.

jaceklaskowski:
And remember to sign off your work. It hit me so many times, and it takes quite some time to work out after a couple of commits.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
sudharshanraja-db: hi @jaceklaskowski  couldnt get you so  which part of the code you want to give name? is it the path parameter?

server/src/main/java/io/unitycatalog/server/service/AuthService.java
sudharshanraja-db: @jamieknight-db  yes i made the changes to use Cookie::value and also regarding reset to head noted going forward will keep this in mind

server/src/main/java/io/unitycatalog/server/service/AuthService.java
xx-db: +1, the server property seems to be used only in createCookie.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
creechy: +1 what are you suggesting to "name"?

server/src/main/java/io/unitycatalog/server/service/AuthService.java
creechy: +1

server/src/main/java/io/unitycatalog/server/service/AuthService.java
creechy: Yeah, seems reasonable to just chain and either generate a `ResponseHeaders` with a new cookie, or an empty `ResponseHeaders`. Something like 
```
     request.headers().cookies().stream()
            .filter(c -> c.name().equals(AuthDecorator.UC_TOKEN_KEY))
            .findFirst()
            .map(ResponseHeaders.of(createCookie(...))
            .orElse(ResponseHeaders.of());
```

server/src/main/java/io/unitycatalog/server/service/AuthService.java
sudharshanraja-db: main idea behind put this in a method  as it is used  during login where we  set and other during logout where we  expire , the difference is during logout we need to set age to 0  and api accepts only Long  where as during login we accept duration in days .. so we need to convert it to long and then call the  create cookie method

sudharshanraja-db:
> I noticed you're on `sudharshanraja-db:main` branch. Create a feature branch instead.

noted Thanks for suggestion  going forward for future PR will make sure this is followed

server/src/main/java/io/unitycatalog/server/service/AuthService.java
creechy: Why not put the `Duration.parse(...).getSeconds()` part into `createCookie` so that you just pass in the duration string into `createCookie`? And then when you want to set age to zero, there should be a similar Duration syntax for that? `PT0S` maybe?

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -167,17 +168,35 @@ public HttpResponse grantToken(
...
         });
 
     return HttpResponse.ofJson(responseHeaders.build(), response);
   }
 
+  @Get("/logout")
+  public HttpResponse logout(HttpRequest request) {
+    return request.headers().cookies().stream()
+        .filter(c -> c.name().equals(AuthDecorator.UC_TOKEN_KEY))
+        .map(Cookie::value)
```
creechy, line 187: You don't need this anymore.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
sudharshanraja-db: ahh got it , didnt strike that 0 can be set in the format :)

server/src/main/java/io/unitycatalog/server/service/AuthService.java
sudharshanraja-db: Updated and also changed  logout to post

sudharshanraja-db:
> @sudharshanraja-db looks like code formatting is failing, once that is fixed, this can be merged.

@jamieknight-db  there was an unused import fixed now  , didnt notice the failure email for the same

PR #687, opened 2024-11-04
Move dependency version declaration to constants

### LINKED ISSUE #622: Clean up dependency versions defined in build.sbt

Dependency versions are defined inline in many places in build.sbt, for example:
https://github.com/unitycatalog/unitycatalog/blob/77b66d58e0d078c931b7b4e5b7504c41f16a38ee/build.sbt#L44

This makes it difficult to track versions, keep dependency versions consistent across subprojects, etc

These should be parameterized/replaced with constants as much as possible, for example:
https://github.com/unitycatalog/unitycatalog/blob/77b66d58e0d078c931b7b4e5b7504c41f16a38ee/build.sbt#L130

This would help keep dependencies consistent across subprojects and make dependency versions easier to change

HorizonNet:
**PR Checklist**

**Description of changes**

Moved the declaration of dependencies into constants within `build.sbt`. Please note, to align the versions for the same dependencies, the following dependencies have been updated to a different version:

* Hadoop to use 3.4.0 (3.3.6 and 3.4.0 were both used)
* Jackson to use 2.17.0 (2.15.0 and 2.17.0 were both used)

Resolves #622

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,10 +41,10 @@ lazy val commonSettings = Seq(
     "-ea",
   ),
   libraryDependencies ++= Seq(
-    "org.slf4j" % "slf4j-api" % "2.0.13",
-    "org.slf4j" % "slf4j-log4j12" % "2.0.13" % Test,
-    "org.apache.logging.log4j" % "log4j-slf4j2-impl" % "2.23.1",
-    "org.apache.logging.log4j" % "log4j-api" % "2.23.1"
+    "org.slf4j" % "slf4j-api" % slf4jVersion,
+    "org.slf4j" % "slf4j-log4j12" % slf4jVersion % Test,
+    "org.apache.logging.log4j" % "log4j-slf4j2-impl" % log4jVersion,
+    "org.apache.logging.log4j" % "log4j-api" % log4jVersion
   ),
   resolvers += Resolver.mavenLocal,
   autoScalaLibrary := false,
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -527,35 +527,35 @@ lazy val spark = (project in file("connectors/spark"))
...
-      "com.fasterxml.jackson.core" % "jackson-databind" % "2.15.0",
-      "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.15.0",
-      "com.fasterxml.jackson.core" % "jackson-annotations" % "2.15.0",
-      "com.fasterxml.jackson.core" % "jackson-core" % "2.15.0",
-      "com.fasterxml.jackson.dataformat" % "jackson-dataformat-xml" % "2.15.0",
-      "org.antlr" % "antlr4-runtime" % "4.9.3",
-      "org.antlr" % "antlr4" % "4.9.3",
-      "com.google.cloud.bigdataoss" % "util-hadoop" % "3.0.2" % Provided,
-      "org.apache.hadoop" % "hadoop-azure" % "3.4.0" % Provided,
+      "com.fasterxml.jackson.core" % "jackson-databind" % jacksonVersion,
```
vikrantpuppala, line 530: I'm not sure if things work at runtime if we do this, we had issues with jackson version previously and had to shade the server jar to help with the test cases. see here: https://github.com/unitycatalog/unitycatalog/pull/687/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91L487-L491

HorizonNet: Do you remember which problem did you run into? Having a single version of Jackson instead of two versions (like it was before)?

PR #688, opened 2024-11-05
Make the Python SDK a namespace package

BenWilson2:
**PR Checklist**

**Description of changes**

This PR makes modifications to the generated Python SDK code so that it conforms to the PEP requirements for inferred shared namespaces https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#legacy-namespace-packages . 
In order to accomplish this, the sbt build script has been updated to create the proper directory structure, remove the root `__init__.py` from the `unitycatalog.*` namespace, and create the distributable `README.md` for PyPI, the `pyproject.toml` file for building the deployable artifact, and the `setup.py` file for local developer-mode installation of the package.


BenWilson2:
Note: I need to rewrite the Python client tests. This is WIP.

clients/python/tests/conftest.py
```diff
@@ -47,47 +61,57 @@ def uc_server():
             pass
 
 
-@pytest.fixture(scope="session")
-def api_client():
-    config = unitycatalog.Configuration(
-        host = "http://localhost:8081/api/2.1/unity-catalog"
-    )
+@pytest_asyncio.fixture()
```
BenWilson2, line 64: Note on the session scoping: pytest-asyncio has a restriction in its implementation for 0.22.0 and above where session scoped fixtures are not supported due to a decision to not support handling closures in futures. 
The recommendation is to move to function-based fixtures, which I've done throughout the suites (I validated that session based fixtures, even with setting configurations in pytest.ini, do not function correctly with the pytest execution environment).

kevinzwang:
This looks good to me. I just have one question: why did you chose to make the library async by the way? I originally kept it synchronous so that it would be easy to call outside of an async function, and it did not seem like a big issue for the calls to the API to be blocking, since I assume they generally are quick and done serially.

BenWilson2:
@kevinzwang it's mostly because we're going to be using this for GenAI tool calling for the UC OSS integration and several GenAI packages are implementing all tool calling functionality as async integrations :) For a tiny bit of headache for users (adding an `await` before calls), I think it's a pretty valuable trade-off :) 
Now, if only they built in support for httpx to the generator, we'd really be in business ;)

BenWilson2:
@tdas The additional complex logic is now migrated to a scala script :)

build.sbt
```diff
@@ -223,32 +225,41 @@ lazy val client = (project in file("target/clients/java"))
...
-    // Define the simple generate command to generate full client codes
-    generate := {
-      val _ = openApiGenerate.value
-    }
+
+    generate := Def.task {
+      val log = streams.value.log
+
+      openApiGenerate.value
+      log.info("OpenAPI client generation completed.")
```
tdas, line 251: seems like this log should be at the end after moving files?

project/PythonPostBuild.scala
```diff
@@ -0,0 +1,59 @@
+import sbt._
+import sbt.Keys._
+import java.nio.file.{Files, Paths, Path, StandardCopyOption}
+
+object PythonPostBuild {
```
tdas, line 5: PythonClient to be very specific

project/PythonPostBuild.scala
```diff
@@ -0,0 +1,59 @@
...
+  def processGeneratedFiles(
```
tdas, line 6: please add a description of the transformation that are being done.

clients/python/build/README.md
```diff
@@ -0,0 +1,160 @@
+# Unity Catalog Python SDK
```
tdas, line 1: since we are using the term "client" can we incorporate client in the names "Python Client SDK" for eg

project/PythonPostBuild.scala
tdas: actually you can just point to the readme that has the whole transformation process

clients/python/README.md
```diff
@@ -1,24 +1,79 @@
...
+
+The sbt build configuration that runs to build a deployable package for the SDK
+replaces several generic auto-generated files and performs some required cleanup actions
+on the generated directory structure to ensure that this package conforms to the
+namespace package configuration required by [this guidance](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/).
+
+This package shares the root `unitycatalog.*` namespace with the other suite of
+unitycatalog official packages (`unitycatalog-ai`, `unitycatalog-openai`, etc.).
+
+The sbt assembly process for this SDK will:
```
tdas, line 25: what is "assmebly" process. 
sbt assembly is a plugin thats for creating fat uber jars. we are not using that. so this is ambiguous. so maybe lets use "generation process"

clients/python/README.md
```diff
@@ -1,24 +1,79 @@
...
+
+1. Remove the generated `__init__.py` file from within the root `unitycatalog` directory. Having any `__init__.py` file in any of the shared namespace packages
+at this level will invalidate the inferred shared namespace logic when Python
+resolves common namespaces, effectively breaking all installed packages.
+2. Remove the `git_push.sh` script as it is not needed and should not be used for
```
tdas, line 30: why are we manually removing this file. why not add it to the open api ignore file and not generate it.

clients/python/README.md
```diff
@@ -1,24 +1,79 @@
...
+identification of a release (these packages are deployed to PyPI, not GitHub).
+3. Replace the `pyproject.toml` generated file with the deployment `pyproject.toml` file that resides within the `unitycatalog/clients/python/build` directory.
+4. Replace the `README.md` generated file with the deployment `README.md` files.
```
tdas, line 33: dont generate the readme file at all. just copy over new the file

clients/python/README.md
```diff
@@ -1,24 +1,79 @@
...
+This is to ensure that if a script-based installation from the generated package
+structure will build a valid namespace installation for local validation.
+
+## Updating deployment build files
+
+If you are updating requirements, modifying the release version, or are providing additional
+guidance within the distributed package's PyPI README.md, ensure that you make
+updates to:
+
+1. The `pyproject.toml` file located at: `unitycatalog/clients/python/build/pyproject.toml`. If you are releasing a new version of the SDK, ensure that the `version` field is updated properly to the appropriate version of Unity Catalog's release.
```
tdas, line 46: lets just make all these paths <python-client-root-dir>. just in case we have to change, it will be easier to fix stuff

clients/python/README.md
```diff
@@ -1,24 +1,79 @@
...
+2. The `setup.py` file location at: `unitycatalog/client/python/build/setup.py`. This file **must** be updated with a new version specifier when releasing a new version of the `unitycatalog-client` package so that local script-based installations of the source code will reflect the correct version.
```
tdas, line 47: please pick theh version from the version.sbt like this

https://github.com/delta-io/delta/blob/master/setup.py#L10-L17

clients/python/build/pyproject.toml
```diff
@@ -0,0 +1,103 @@
...
+    "aiohttp>=3.8.4",
+    "aiohttp-retry>=2.8.3",
+    "python-dateutil>=2.8.2",
+    "typing-extensions>=4.7.1",
+    "pydantic<3,>=2",
+]
+maintainers = [
+  {name="Denny Lee", email="denny.lee@databricks.com"},
+  {name="Ben Wilson", email="benjamin.wilson@databricks.com"},
+  {name="Tathagata Das", email="tdas@databricks.com"}
```
tdas, line 22: can you use my personal email. :)

clients/python/build/setup.py
```diff
@@ -0,0 +1,78 @@
...
+from setuptools import setup, find_packages
+from pathlib import Path
+
+# Read the long description from README.md
+this_directory = Path(__file__).parent
+long_description = (this_directory / "README.md").read_text()
+
+setup(
+    name="unitycatalog-client",
+    version="0.2.1",
```
tdas, line 18: pick the version from sbt 

https://github.com/delta-io/delta/blob/master/setup.py#L10-L17

project/PythonPostBuild.scala
```diff
@@ -0,0 +1,59 @@
...
+      if (Files.exists(source)) {
+        Files.delete(source)
+        log.info(s"Removed $description at ${source}")
+      } else {
+        sys.error(s"File $description at ${source} is not present. Please verify the path.")
+      }
+    }
+
+    val packageNameParts = openApiPackageName.split("\\.")
+    val generatedInitPath: Path = Paths.get(
```
tdas, line 32: nit: all these `: Path` does not seem necessary

clients/python/README.md
BenWilson2: Agreed on this being far easier. Updated the Script to include the logic since updating a pyproject.toml (we can't run python code in there, unlike in setup.py) is a static file update anyway and I figured that it would keep things simpler if we didn't have a mix of scala and python to construct these deployment files.

clients/python/README.md
BenWilson2: good idea!

clients/python/README.md
BenWilson2: solid idea. Changed by using the ignore file approach

clients/python/build/pyproject.toml
BenWilson2: updated!

project/PythonClientPostBuild.scala
```diff
@@ -0,0 +1,140 @@
...
+    val clientSetupPy = buildDir.resolve("setup.py")
+    val projectReadme = buildDir.resolve("README.md")
+
+    val generatedPyproject = Paths.get(openApiOutputDir, "pyproject.toml")
+    val generatedSetupPy = Paths.get(openApiOutputDir, "setup.py")
+    val generatedReadme = Paths.get(openApiOutputDir, "README.md")
+
+    copyFile(clientPyproject, generatedPyproject, "pyproject.toml")
+    copyFile(clientSetupPy, generatedSetupPy, "setup.py")
+    copyFile(projectReadme, generatedReadme, "README.md")
```
tdas, line 119: this along with copyFile function seems like a bit of bloated, repeated code. why not just make it loop 
```
Seq("pyproject.toml", ...).foreach { name =>
   Files.copy(
     buildDir.resolve(name), 
     Paths.get(openApiOutputDir, name), 
    StandardCopyOption.REPLACE_EXISTING)  
}
```

BenWilson2: I think I'm remembering how to write functional scala now ;) how's this update?

BenWilson2:
@tdas I externalized the version update stuff for python to a standalone script and just call that in the build process to ensure that the synchronized version.sbt value is always applied everywhere so that we don't have the chance of getting it wrong.

tdas (approved):

project/PythonClientPostBuild.scala
```diff
@@ -0,0 +1,173 @@
...
+      log.info("Version update script executed successfully")
+    }
+
+    Seq("setup.py", "README.md", "pyproject.toml").foreach { fileName =>
+      val sourcePath = buildDir.resolve(fileName)
+      val targetPath = Paths.get(openApiOutputDir, fileName)
+
+      if (Files.exists(sourcePath)) {
+        Try {
+          Files.copy(sourcePath, targetPath, StandardCopyOption.REPLACE_EXISTING)
```
tdas, line 101: does this command not fail if the file copy does not work before some reasone? why is the try catch needed?

BenWilson2: Purely for a nice error message :) I can remove it, though!

PR #689, opened 2024-11-05
Fix execute_function string parsing with serverless

serena-ruan:
**PR Checklist**

**Description of changes**
Update the logic of serverless execution to pass arguments using `:` prefix by SQL literals. Wrap the exception to provide better error message when failing. 
Drop the previous sanitization logic as we no longer need it :) SQL side can deal with the single+double quotes combination correctly.
Updated tests.

### DIFF
--- a/ai/core/src/unitycatalog/ai/core/databricks.py
+++ b/ai/core/src/unitycatalog/ai/core/databricks.py
@@ -657,7 +657,7 @@ class DatabricksFunctionClient(BaseFunctionClient):
         try:
             result = self.spark.sql(sqlQuery=sql_command.sql_query, args=sql_command.args or None)
         except Exception as e:
-            error = f"Failed to execute function with command {sql_command}; Error: {e}"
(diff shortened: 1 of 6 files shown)


serena-ruan:
Integration tests: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/11686476764

ai/core/src/unitycatalog/ai/core/databricks.py
```diff
@@ -146,7 +143,15 @@ def retry_on_session_expiration(func):
     def wrapper(self, *args, **kwargs):
         for attempt in range(1, max_attempts + 1):
             try:
-                return func(self, *args, **kwargs)
+                result = func(self, *args, **kwargs)
+                # for non-seession related error in the result, we should directly return the result
```
BenWilson2, line 147:
```suggestion
                # for non-session related error in the result, we should directly return the result
```

BenWilson2 (approved):
This is awesome! Great work on validating that we can use parameter calls and simplifying this :D 👍

PR #690, opened 2024-11-05
Add usage warnings for poorly constructed functions

BenWilson2:
**PR Checklist**

**Description of changes**

Introduce a warning for externally created UC functions that will be used as a GenAI tool to inform users that a lack of a function description or parameter descriptions severely impacts the viability of using UC functions as GenAI tools if these are not defined. 
There are no updates to be made for functions created with the unitycatalog-ai `create_python_function` as this is already not permitted within that API.


ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
```diff
@@ -77,3 +78,43 @@ def validate_param(param: Any, column_type: str, param_type_text: str) -> None:
...
+    """
+    params_with_no_description = []
+
+    for param_info in func_info.input_params.parameters:
+        if not param_info.comment:
+            params_with_no_description.append(param_info.name)
+
+    if params_with_no_description:
+        warnings.warn(
+            f"The following parameters do not have descriptions: {', '.join(params_with_no_description)} for the function {func_info.name}. "
```
serena-ruan, line 103:
```suggestion
            f"The following parameters do not have descriptions: {', '.join(params_with_no_description)} for the function {func_info.full_name}. "
```

serena-ruan (approved):
LGTM! Should we also warn when creating a function?

BenWilson2:
> LGTM! Should we also warn when creating a function with sql?

Definitely. I have that on our tracker as a follow up PR (since the logic will be a bit different potentially depending on when we want to warn)

ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
BenWilson2: Good call!

PR #724, opened 2024-11-13
Fix client execution issue for SparkSession references

serena-ruan:
**PR Checklist**

**Description of changes**
If the client is run on non-serverless computes, we should warn the user as some of the functionalities might not work well. 
Non-dbconnect sparkSession doesn't contain `is_stopped` attribute.
After this PR:
image
Also updated the exception catch in serverless to include result part, as the calculation happens when collect is called.

https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/11813567109 passed

### DIFF
--- a/ai/core/src/unitycatalog/ai/core/databricks.py
+++ b/ai/core/src/unitycatalog/ai/core/databricks.py
@@ -91,11 +91,12 @@ def _try_get_spark_session_in_dbr() -> Any:
 
         if not isinstance(spark, SparkSession):
             _logger.warning(
-                "Current spark session in the active Databricks runtime is not a "
-                "pyspark.sql.connect.session.SparkSession instance, it's probably "
-                "because you're not using a Serverless compute. To use the full "
-                "functionalities of this package please switch to a Serverless cluster, "
-                "check https://docs.databricks.com/en/compute/serverless/index.html#connect-to-serverless-compute "
+                "Current SparkSession in the active environment is not a "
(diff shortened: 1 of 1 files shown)


ai/core/src/unitycatalog/ai/core/databricks.py
```diff
@@ -87,7 +87,17 @@ def _try_get_spark_session_in_dbr() -> Any:
...
+        from pyspark.sql.connect.session import SparkSession
+
+        if not isinstance(spark, SparkSession):
+            _logger.warning(
+                "Current spark session in the active Databricks runtime is not a "
+                "pyspark.sql.connect.session.SparkSession instance, it's probably "
+                "because you're not using a Serverless compute. To use the full "
+                "functionalities of this package please switch to a Serverless cluster, "
+                "check https://docs.databricks.com/en/compute/serverless/index.html#connect-to-serverless-compute "
+                "for more details."
```
BenWilson2, line 99:
```suggestion
                "Current SparkSession in the active environment is not a "
                "pyspark.sql.connect.session.SparkSession instance. Classic runtime does not support "
                "all functionalities of the unitycatalog-ai framework. To use the full "
                "capabilities of unitycatalog-ai, execute your code using a client that is attached to "
                "a Serverless runtime cluster. To learn more about serverless, see the guide at: "
                "https://docs.databricks.com/en/compute/serverless/index.html#connect-to-serverless-compute "
                "for more details."
```

BenWilson2 (approved):
LGTM once the warning message is updated :)

PR #726, opened 2024-11-13
Add warning to sql_body function creation

BenWilson2:
**PR Checklist**

**Description of changes**

Adds warnings for function creation via the `create_function` API if parameter descriptions and function description is not specified to alert users that what they've just created is likely not going to be ideal for GenAI use cases. 
Decided to warn instead of raise here since there may be cases where a function without a description is inherently usable by some LLMs. 
Added tests to validate the warning logic to ensure that different conditions of FunctionInfo configuration will warn appropriately.


serena-ruan (approved):
LGTM!

PR #736, opened 2024-11-15
Send non empty data as part of expired cookie in the header

sudharshanraja-db:
**PR Checklist**

**Description of changes**

Logout endpoint was implemented  to expire cookie as part of the task #593  and changes are pushed as part of PR #678 but when the  endpoint was called it gets hung indefinitely , On further investigation non empty data needs to be send as part of the header  when we expire the cookie. This PR has the fix for the same.


sudharshanraja-db:
raising new PR as part of #738

PR #750, opened 2024-11-21
Send non empty data as part of expired cookie in the header

sudharshanraja-db:
**PR Checklist**

**Description of changes**

Logout endpoint was implemented to expire cookie as part of the task https://github.com/unitycatalog/unitycatalog/pull/593 and changes are pushed as part of PR https://github.com/unitycatalog/unitycatalog/pull/678 but when the endpoint was called it gets hung indefinitely , On further investigation non empty data needs to be send as part of the header when we expire the cookie. This PR has the fix for the same.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/AuthService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/AuthService.java
@@ -66,6 +66,7 @@ public class AuthService {
   private final JwksOperations jwksOperations;
 
   private static final String COOKIE = "cookie";
+  private static final String EMPTY_RESPONSE = "{}";
 
   public AuthService(SecurityContext securityContext) {
     this.securityContext = securityContext;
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -185,9 +185,12 @@ public HttpResponse logout(HttpRequest request) {
...
               ResponseHeaders headers =
-                  ResponseHeaders.of(
-                      HttpStatus.OK, HttpHeaderNames.SET_COOKIE, expiredCookie.toSetCookieHeader());
-              return HttpResponse.of(headers);
+                  ResponseHeaders.builder()
+                      .status(HttpStatus.OK)
+                      .add(HttpHeaderNames.SET_COOKIE, expiredCookie.toSetCookieHeader())
+                      .contentType(MediaType.JSON)
+                      .build();
+              return HttpResponse.of(headers, HttpData.ofUtf8("{}"));
```
creechy, line 193: Can you just add a comment why we're sending an empty Json for now? like - Armeria needs a response payload to complete the request, so we are sending an empty JSON.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -185,9 +185,12 @@ public HttpResponse logout(HttpRequest request) {
...
             })
         .orElse(HttpResponse.of(HttpStatus.OK));
```
creechy, line 195: Hmm.. maybe we should send an empty JSON here too, just to be consistent.

creechy (changes requested):
looks good - a couple of small suggestions.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
sudharshanraja-db: addressed

server/src/main/java/io/unitycatalog/server/service/AuthService.java
sudharshanraja-db: make sense have  made the changes

creechy (approved):

Issue #431, opened 2024-09-06
[UI] Use the official OpenAPI specification to generate types


dennyglee:
**Is your feature request related to a problem? Please describe.**
As part of the migration of the unitycatalog-ui repo to unitycatalog (main), referring the original UI issue:  [Use the official OpenAPI specification to generate types](https://github.com/unitycatalog/unitycatalog-ui/issues/57)

**Describe the solution you would like**

**Additional context**

JuanPabloDiaz:
cc @ognis1205, @everettVT

ognis1205:
Describe the solution you would like

 - Use [openapi-typescript](https://openapi-ts.dev/) to generate types (URL paths, requests, and responses)
 - Implement utility types and functions to handle generated artifacts type safely

Additional context

Update later...

ognis1205:
take

Issue #759, opened 2024-11-25
Enhance OpenAPI Specification with Detailed Error Codes.


ognis1205:
**Is your feature request related to a problem? Please describe.**

To improve the UI/UX of client applications, such as the WebUI, appropriate error codes should be added to the OpenAPI specification. For instance, in the default implementation, attempting to delete a catalog that contains a schema is treated as an error. However, the response returned to the client provides no information about the reason for the failure, currently returning only a generic 400 Bad Request.

**Describe the solution you would like**

Add appropriate error codes to the OpenAPI specification.

ognis1205:
I think we should discuss this topic @dennyglee @MrPowers .

PR #762, opened 2024-11-26
[UI] Use the official OpenAPI specification to generate types.

ognis1205:
**PR Checklist**

**Description of changes**

 - Use [openapi-typescript](https://openapi-ts.dev/) to generate types (URL paths, HTTP methods, requests, and responses)
    ***Rationale***
     - The primary need for code generated from the OpenAPI schema is type information rather than client implementations. This allows for flexibility in selecting an HTTP client and supports gradual adaptation even when the schema YAML implementation is unstable.
     - openapi-generator requires a Java runtime, while openapi-typescript does not depend on Java, node-gyp, or an API server.
     - Even if the UI project becomes independent of the unitycatalog repository, it can still generate code from a remote schema without requiring a Java runtime.
 - Implement utility types and functions to handle generated artifacts type-safely
 - Install [npm-run-all](https://www.npmjs.com/package/npm-run-all) and enhance the build script

Will close https://github.com/unitycatalog/unitycatalog/issues/431 .


ognis1205:
Move `staleTime` to `utils/constants.ts`:

https://github.com/unitycatalog/unitycatalog/blob/967151eb4211363375de2c56b0487f917e34067d/ui/src/App.tsx#L212

ognis1205:
May address this issue as well:

https://github.com/unitycatalog/unitycatalog/issues/766

ui/src/components/models/ModelVersionStatusDisplay.tsx
```diff
@@ -10,27 +10,30 @@ import React from 'react';
 export default function ModelVersionStatusDisplay({
   status,
 }: {
-  status: string;
+  status: ModelVersionStatus;
 }) {
   switch (status) {
-    case ModelVersionStatus.READY:
+    case 'READY':
```
jamieknight-db, line 16: Is there a way to keep using the `ModelVersionStatus` enum for these cases instead of the strings?

ui/src/context/control.ts
```diff
@@ -0,0 +1,12 @@
+import axios from 'axios';
+import { UC_AUTH_API_PREFIX } from '../utils/constants';
+
+/**
+ * The singleton client instance with `baseURL`.
+ */
+export const CLIENT = axios.create({
```
jamieknight-db, line 7: It looks like we are instantiating the client twice? Here and in the catalog.ts file

ui/src/context/catalog.ts
```diff
@@ -1,11 +1,12 @@
 import axios from 'axios';
 import { UC_API_PREFIX } from '../utils/constants';
 
-const client = axios.create({
+/**
+ * The singleton client instance with `baseURL`.
+ */
+export const CLIENT = axios.create({
```
jamieknight-db, line 7: I'm not sure renaming this file from `client` to `catalog` makes sense. What is your line of thinking?

ui/src/hooks/functions.tsx
```diff
@@ -4,106 +4,147 @@ import {
...
-        schema_name: schema,
+      const api = route(CLIENT, {
+        path: '/functions',
+        method: 'get',
+        params: {
+          query: {
+            catalog_name,
+            schema_name,
+            max_results,
+            page_token,
```
jamieknight-db, line 54: Not sure these are needed since we are only doing client side pagination at this time

ui/src/hooks/models.tsx
```diff
@@ -4,221 +4,259 @@ import {
...
-        schema_name: schema,
+      const api = route(CLIENT, {
+        path: '/models',
+        method: 'get',
+        params: {
+          query: {
+            catalog_name,
+            schema_name,
+            max_results,
+            page_token,
```
jamieknight-db, line 58: Same here, is the endpoint asking for these query params?

ui/src/hooks/models.tsx
```diff
@@ -4,221 +4,259 @@ import {
...
+      const api = route(CLIENT, {
+        path: '/models/{full_name}',
+        method: 'patch',
+        params: {
+          paths: {
+            full_name,
+          },
+          body: {
+            comment,
+            new_name,
```
jamieknight-db, line 196: I don't think we support `new_name` right now

ui/src/hooks/schemas.tsx
```diff
@@ -82,30 +153,58 @@ export function useCreateSchema() {
...
+        path: '/schemas/{full_name}',
+        method: 'patch',
+        params: {
+          paths: {
+            full_name,
+          },
+          body: {
+            comment,
+            properties,
+            new_name,
```
jamieknight-db, line 195: I also don't think we support `properties` or `new_name` here

ui/src/hooks/tables.tsx
```diff
@@ -4,102 +4,150 @@ import {
...
-        schema_name: schema,
+      const api = route(CLIENT, {
+        path: '/tables',
+        method: 'get',
+        params: {
+          query: {
+            catalog_name,
+            schema_name,
+            max_results,
+            page_token,
```
jamieknight-db, line 54: I think these can be removed

ui/src/hooks/user.tsx
```diff
@@ -53,41 +39,81 @@ export function useLoginWithToken() {
...
+      );
+      const response = await api.call();
+      if (response.result !== 'success') {
+        // NOTE:
+        // To the best of my knowledge, most HTTP error codes are not explicitly defined in the OpenAPI
+        // specification. I reviewed the server implementation and identified the error codes related to
+        // authentication by examining the codebase. Relevant details can be found here:
+        // - https://github.com/search?q=repo%3Aunitycatalog%2Funitycatalog%20UNAUTHENTICATED&type=code
+        // - https://github.com/search?q=repo%3Aunitycatalog%2Funitycatalog+NOT_FOUND&type=code
+        if (response.data.status === 401 || response.data.status === 404) {
```
jamieknight-db, line 111: I think this should just be checking for 401 status, the 404 was only returning when the endpoint URL was incorrect.

ui/src/hooks/volumes.tsx
```diff
@@ -4,101 +4,151 @@ import {
...
-        schema_name: schema,
+      const api = route(CLIENT, {
+        path: '/volumes',
+        method: 'get',
+        params: {
+          query: {
+            catalog_name,
+            schema_name,
+            max_results,
+            page_token,
```
jamieknight-db, line 55: remove?

ui/src/hooks/volumes.tsx
```diff
@@ -4,101 +4,151 @@ import {
...
+      const api = route(CLIENT, {
+        path: '/volumes/{name}',
+        method: 'patch',
+        params: {
+          paths: {
+            name,
+          },
+          body: {
+            comment,
+            new_name,
```
jamieknight-db, line 139: Can remove `new_name` here as well

ui/src/pages/CatalogDetails.tsx
```diff
@@ -55,9 +55,11 @@ export default function CatalogDetails() {
           <CatalogSidebar catalog={catalog} />
         </DetailsLayout.Aside>
       </DetailsLayout>
+      {/* NOTE: Due to inconsistencies in the API specification, the following workaround is necessary. 
+          This inconsistency can still be addressed at the interface level. */}
       <EditCatalogDescriptionModal
         open={open}
-        catalog={data}
+        catalog={{ new_name: data.name, ...data }}
```
jamieknight-db, line 62: Is `new_name` referring to the comment here? Can we remove `new_name` and just use `comment`

jamieknight-db:
@ognis1205 this is really fantastic! Thank you for taking the lead on this. I have left some comments and suggestions. I still need to clone your branch and test it out, I will try to get that done tomorrow. Were you able to test out error states as well?

ui/src/components/models/ModelVersionStatusDisplay.tsx
ognis1205: I generally prefer using `union` over `enum`, but considering the unusual constant values (such as `TokenType`, e.g., `urn:ietf:params:oauth:token-type:access_token`), you're absolutely right — `enum` would be more appropriate in this case. Thank you for the suggestion! Of course, we can output it as an `enum`.

ognis1205:
@jamieknight-db 

Regarding the parameters in the hooks module (e.g., max_results, page_token, new_name), all of these are destructured parameters and can therefore be optional (i.e., they can be undefined, allowing users to omit them) as defined in the OpenAPI specification. I included these parameters for clarification purposes, not as strict requirements. I was concerned that removing them might make the specification less clear. However, since this is largely a matter of personal preference, I'll leave the final decision to you as the code owner. What are your thoughts?

ognis1205:
> @ognis1205 this is really fantastic! Thank you for taking the lead on this. I have left some comments and suggestions. I still need to clone your branch and test it out, I will try to get that done tomorrow. Were you able to test out error states as well?

The current `route` function in the `utils/openapi` module is designed so that no processing is performed for error codes unless the expected error codes are explicitly specified in the `hooks` implementation and the OpenAPI specification (The current specification does not define error codes at all). As such, I am not conducting tests for specific error codes. However, in the event of a server error, the implementation ensures that the same notifications as the existing code are displayed. I have performed minimal checks, such as intentionally stopping the server to confirm that notifications are displayed. Regarding this, I plan to conduct thorough testing after resolving the following two issues:

 - https://github.com/unitycatalog/unitycatalog/issues/759
 - https://github.com/unitycatalog/unitycatalog/issues/786

ognis1205:
Regarding the issue of instantiating two Axios clients, there are two main reasons:

 - At the time I started implementing this PR, [there were parts of the Catalog API client that utilized interceptors](https://github.com/unitycatalog/unitycatalog/commit/ebe39632768112fa56e4e6d959e74ca7f5c2c5fd). Since these interceptors handled processes unnecessary for the Auth API client, a separate instance was deemed appropriate.

 - Axios client instances are typically used to reuse common configurations. However, I couldn’t find any code where the configurations were shared between the Catalog and Auth clients. If the process involves overwriting the Catalog client’s base URL with the Auth client’s base URL for each request, I decided it would be better to create a separate client to enhance the symmetry between the code for the two clients.

ognis1205:
@jamieknight-db 

Thank you for the detailed and thorough review. I have fixed the points that were straightforward and didn’t require further discussion. I’ve also left some comments, so when you have time, I’d appreciate it if you could review them again.

ognis1205:
@jamieknight-db 

 - Consolidated `/context/catalog.ts` and `/context/control.ts` back into `/context/client.ts`, as they were previously.
 - Removed unused arguments from `/hooks`.
 - For error handling, I ensured it is handled at the type level, guaranteeing at compile-time that `if (isError(response)) { ... }` will not be executed. This preserves the same behavior as the existing implementation. One exception to note is the need to specifically handle the `401` error for `useGetCurrentUser` in `/hooks/user.ts`, but for all other cases, the behavior and error messages remain unchanged.

I’d appreciate it if you could review them again when you have time. Thanks!

jamieknight-db (approved):
This all lgtm - I was not able to pull down the branch to test out CRUD operations myself but code looks good. Thank you for considering my feedback @ognis1205! I would like @yc-shawn and @dennyglee to take a look before merging.

ognis1205:
@jamieknight-db 

Thank you for your thorough review. Please feel free to let me know if there are any issues or if anything needs to be addressed.

ognis1205:
@yc-shawn @dennyglee 

Just wanted to kindly follow up on this PR to see if you’ve had a chance to take another look. Let me know if there’s anything I can clarify or adjust to help with the review. Thank you!

jamieknight-db:
@ognis1205 does this PR need to be merged first? before[ this PR](https://github.com/unitycatalog/unitycatalog/pull/771)?

ognis1205:
@jamieknight-db 

This PR and [that PR](https://github.com/unitycatalog/unitycatalog/pull/771) are not dependent on each other, but since this PR has relatively fewer impacts and stakeholders, it might be better to merge this one first and wait a bit longer for feedback on [the other PR](https://github.com/unitycatalog/unitycatalog/pull/771). Thank you for the suggestion!

jamieknight-db:
@dennyglee are you ok with me merging this?

dennyglee:
> @dennyglee are you ok with me merging this?

Absolutely, feel free to merge and/or I can take care of this later today, eh?! :)

PR #771, opened 2024-11-30
Update auth-related openapi specification.

### LINKED ISSUE #768: Add OpenAPI Specification for Auth-Related APIs.

The catalog/user-related APIs are defined in the OpenAPI specification, but the auth-related APIs are not. Adding definitions for the auth-related APIs would improve code consistency and make the codebase cleaner.

Add a YAML file for the auth-related APIs.

ognis1205:
**PR Checklist**

**Description of changes**

 - Added models and APIs related to authentication to OpenAPI specification.
 - Updated the server-side logic to align with the updated OpenAPI specifications.
 - Modified the CLI to support the changes introduced in the updated OpenAPI specification.
 - ~Fixed the server bootstrap order to ensure proper initialization.~

Will close #768 .


ognis1205:
cc: @creechy

build.sbt
```diff
@@ -137,6 +137,8 @@ lazy val controlApi = (project in file("target/control/java"))
       "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonVersion,
       "com.fasterxml.jackson.core" % "jackson-databind" % jacksonVersion,
       "com.fasterxml.jackson.datatype" % "jackson-datatype-jsr310" % jacksonVersion,
+      "org.apache.httpcomponents" % "httpclient" % orgApacheHttpVersion,
```
creechy, line 140: Feels like a big dependency just to use to use the `UrlEncoderUtils`... or were you planning to eventually refactor the current uses of `HttpURLConnection` with `HttpClient` instead? Does Armeria have any built in classes for Url Encoding?

examples/cli/src/main/java/io/unitycatalog/cli/AuthCli.java
```diff
@@ -75,24 +74,21 @@ private static String exchange(ApiClient apiClient, JSONObject json)
...
-        .append("&requested_token_type=")
-        .append(URLEncoder.encode(TokenTypes.ACCESS, StandardCharsets.UTF_8))
-        .append("&subject_token_type=")
-        .append(URLEncoder.encode(TokenTypes.ID, StandardCharsets.UTF_8))
-        .append("&subject_token=")
-        .append(URLEncoder.encode(login.get("identityToken"), StandardCharsets.UTF_8));
+    String body =
+        Oauth2CliExchange.URLEncodedForm.ofMap(
+            Map.ofEntries(
+                entry("grant_type", GrantType.TOKEN_EXCHANGE.getValue()),
```
creechy, line 82: Is there not a way to use static strings or enums instead of the strings here like `"grant_type"` ? @jaceklaskowski do you have any thoughts/opinions here?

ognis1205: I don't have a strong opinion on this matter, but as @creechy suggested in the source code comments, defining it as an enum in a shared utility could be a good approach. However, since the constant field names are not used outside of the CLI implementation, I thought implementing it as an enum might not be strictly necessary. That said, I’m happy to leave the final decision to the team.

build.sbt
ognis1205: Even when the library template is set to `native`, OpenAPI Generator seems to generate an API that depends on Apache HttpClient when handling `x-www-form-urlencoded` requests/responses:

 - https://openapi-generator.tech/docs/generators/java/
 - https://gist.github.com/ognis1205/526371ddaed10cb2e96ab83ccc24012b

Could you clarify if you are referring to the `cli` build? If so, I completely agree with your point, and I will make the necessary adjustments.

ognis1205: I have removed the Apache HttpClient dependency from the `cli` project. It would be great if you could share your thoughts on the `controlApi` project whenever you have some time.

examples/cli/src/main/java/io/unitycatalog/cli/AuthCli.java
dennyglee: While true, I can see the grant type being used outside of just the CLI implementation  and the `AuthService` especially as extend our authentication features.  While not wanting to be a blocker, IMHO, I tend to agree with @creechy that we should probably used static strings or enums if possible, eh?!

ognis1205:
@creechy @jaceklaskowski @dennyglee 

Regarding the discussion about holding the parameter names of each request and response as static strings or enums, it seems that the OpenAPI generator can also generate models that are not used in the service. Therefore, I plan to change the implementation approach to define the related requests and responses specified in the RFC in the OpenAPI spec and use it in both the CLI and Control API projects.

With this approach, there will no longer be a need to independently define common constant values in each project, which should improve maintainability. It's not necessarily the most elegant solution, but rather a workaround. Since it hasn't been implemented yet, I'm not sure if it will work as expected, but I wanted to share this for now.

ognis1205:
@creechy @jaceklaskowski @dennyglee

I have modeled all the necessary requests and responses for the OAuth token exchange flow in the OpenAPI spec. This eliminates the need to manage form keys as static strings or enums in the `cli` project, enabling centralized management of the required models for both the `controlApi` and `cli` projects:

 - [AuthCli.java#L79](https://github.com/unitycatalog/unitycatalog/blob/6212fb78379690ca26e07faec507dc7cadf27f31/examples/cli/src/main/java/io/unitycatalog/cli/AuthCli.java#L79)
 - [Oauth2CliExchange.java#L94](https://github.com/unitycatalog/unitycatalog/blob/6212fb78379690ca26e07faec507dc7cadf27f31/examples/cli/src/main/java/io/unitycatalog/cli/utils/Oauth2CliExchange.java#L94)
 - [Oauth2CliExchange.java#L137](https://github.com/unitycatalog/unitycatalog/blob/6212fb78379690ca26e07faec507dc7cadf27f31/examples/cli/src/main/java/io/unitycatalog/cli/utils/Oauth2CliExchange.java#L137)
 - [AuthService.java#L214](https://github.com/unitycatalog/unitycatalog/blob/6212fb78379690ca26e07faec507dc7cadf27f31/server/src/main/java/io/unitycatalog/server/service/AuthService.java#L214)

I’ve also left a comment above regarding the Apache HttpClient dependency, so please check that as well.
I’d appreciate it if you could review this again when you have time. Thanks!

ognis1205:
@jamieknight-db @yc-shawn 

Along with the server-side implementation, I updated the UI to post URL-encoded forms to the server. Please review it when you have time. Thanks!

jaceklaskowski:
LGTM

(I didn't know so much could be auto-generated from an openapi spec)

ognis1205:
@jaceklaskowski 

Thank you for taking the time to review!

jamieknight-db (approved):
🚀

ognis1205:
@jamieknight-db 

Thank you for taking the time to review!

ognis1205:
@creechy

Just wanted to kindly follow up on this PR to see if you’ve had a chance to take another look. Let me know if there’s anything I can clarify or adjust to help with the review. Thank you!

jamieknight-db:
Is everyone ok with me merging this PR for @ognis1205? I will plan to do so tomorrow unless anyone has any objections.

api/control.yaml
```diff
@@ -132,6 +132,94 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/OAuthTokenExchangeInfo'
+  /auth/logout:
+    post:
+      tags:
+        - OAuth2
+      operationId: logout
+      summary: Expire the access token
+      description: |
+        Expires the authorixation cookie.
```
creechy, line 202: same as previous comment.

api/control.yaml
```diff
@@ -132,6 +132,94 @@ paths:
...
+      # - https://github.com/OpenAPITools/openapi-generator/blob/v5.1.0/docs/generators/java.md#parameter-feature
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              nullable: true
+      responses:
+        '200':
+          description: The access token will be expired.
```
creechy, line 222: same as previous comment.

api/control.yaml
```diff
@@ -256,6 +344,181 @@ components:
...
+        - SAML2
+        - JWT
+    AccessTokenType:
+      description: |
+        Access token type identifier, from RFC 8693 Section 2.
+        See https://datatracker.ietf.org/doc/html/rfc8693#section-2
+      type: string
+      enum:
+        - Bearer
+        - MAC
```
creechy, line 390: Do we want to include unsupported token types?

api/control.yaml
```diff
@@ -256,6 +344,181 @@ components:
...
+      enum:
+        - code
+        - token
+    AuthorizationGrantType:
+      description: |
+        Authorization grant type, from RFC 6749 Section 4.
+        See https://datatracker.ietf.org/doc/html/rfc6749#section-4
+      type: string
+      enum:
+        - authorization_code
```
creechy, line 406: similar, do we want to include all posible grant types, or just what we support? feels a bit misleading to include types that are not usable.

examples/cli/src/main/java/io/unitycatalog/cli/AuthCli.java
```diff
@@ -75,24 +74,21 @@ private static String exchange(ApiClient apiClient, JSONObject json)
...
-        .append("grant_type=")
-        .append(URLEncoder.encode(GrantTypes.TOKEN_EXCHANGE, StandardCharsets.UTF_8))
-        .append("&requested_token_type=")
-        .append(URLEncoder.encode(TokenTypes.ACCESS, StandardCharsets.UTF_8))
-        .append("&subject_token_type=")
-        .append(URLEncoder.encode(TokenTypes.ID, StandardCharsets.UTF_8))
-        .append("&subject_token=")
-        .append(URLEncoder.encode(login.get("identityToken"), StandardCharsets.UTF_8));
+    String body =
+        Oauth2CliExchange.URLEncodedForm.of(
```
creechy, line 80: `Oauth2CliExchange.URLEncodedForm` is neat, but it looks like the generated `OAuthTokenExchangeForm` class includes a `toUrlQueryString()` method... could you just use this? And then get rid of of `Oauth2CliExchange.URLEncodedForm` ?

examples/cli/src/main/java/io/unitycatalog/cli/utils/Oauth2CliExchange.java
```diff
@@ -104,16 +92,15 @@ public String authenticate() throws IOException {
...
-            Fields.CLIENT_ID,
-            URLEncoder.encode(clientId, StandardCharsets.UTF_8),
-            Fields.REDIRECT_URL,
-            URLEncoder.encode(redirectUrl, StandardCharsets.UTF_8),
-            Fields.CODE,
-            URLEncoder.encode("openid profile email", StandardCharsets.UTF_8),
-            Hex.encodeHexString(stateBytes));
+        authorizationUrl
+            + "?"
+            + URLEncodedForm.of(
```
creechy, line 97: Same. Can you just use `OAuthAuthorizationForm.toUrlQueryString()`?

examples/cli/src/main/java/io/unitycatalog/cli/utils/Oauth2CliExchange.java
```diff
@@ -150,15 +137,15 @@ public String authenticate() throws IOException {
...
     server.stop(0);
 
-    Map<String, String> tokenParams = new HashMap<>();
-    tokenParams.put(Fields.CODE, authCode);
-    tokenParams.put(Fields.CLIENT_ID, clientId);
-    tokenParams.put(Fields.CLIENT_SECRET, clientSecret);
-    tokenParams.put(Fields.GRANT_TYPE, "authorization_code");
-    tokenParams.put(Fields.REDIRECT_URL, redirectUrl);
+    String tokenBody =
+        URLEncodedForm.of(
```
creechy, line 141: Also use `toUrlQueryString()` ?

examples/cli/src/main/java/io/unitycatalog/cli/utils/Oauth2CliExchange.java
```diff
@@ -207,23 +194,60 @@ private int findAvailablePort() throws IOException {
...
-            p -> !p.getKey().equals(Fields.CLIENT_ID) && !p.getKey().equals(Fields.CLIENT_SECRET))
-        .map(
-            p ->
-                URLEncoder.encode(p.getKey(), StandardCharsets.UTF_8)
-                    + "="
-                    + URLEncoder.encode(p.getValue(), StandardCharsets.UTF_8))
-        .reduce((p1, p2) -> p1 + "&" + p2)
-        .orElse("");
-  }
+  public static class URLEncodedForm {
```
creechy, line 197: wondering if we can get rid of all this in favor of `toUrlQueryString()` on the generated code?

creechy (changes requested):
A bunch of nits and small questions, otherwise looking good.

examples/cli/src/main/java/io/unitycatalog/cli/AuthCli.java
ognis1205: Hey @creechy ,

That’s a great point. I was aware that OpenAPI Generator outputs `toUrlQueryString()`, but the `cli` project depends on the `controlModels` project, not the `controlApi` project:

https://github.com/unitycatalog/unitycatalog/blob/a8fb85b33132ef952f1a58f112b16310981106f2/build.sbt#L477

While it’s true that `toUrlQueryString()` is available in the implementation under `target/control/java`, it’s not accessible under `server/target/controlmodels` because the library template is set as `resteasy`:

https://github.com/unitycatalog/unitycatalog/blob/a8fb85b33132ef952f1a58f112b16310981106f2/build.sbt#L460

As you mentioned, I also think we should use `toUrlQueryString()`, but to do so, we would need to change the `library` field in the `openApiAdditionalProperties` of the `controlModels` project to `native`

 Would it be okay to change this to native? What do you think?
 
 See also: https://openapi-generator.tech/docs/generators/java/

ognis1205:
@creechy @jamieknight-db 

I have addressed all of @creechy 's feedback. As a result, I made some changes to `build.sbt` and ui/src/hooks/user.ts`. I’d appreciate it if you could review them when you have time. Thank you!

ognis1205:
@creechy @jamieknight-db 

Just checking in—any updates on this?

jamieknight-db (approved):
Looks good from a UI perspective. Thanks @ognis1205!

jamieknight-db:
> @creechy @jamieknight-db
> 
> Just checking in—any updates on this?

This looks good from the UI side, it looks like @creechy will need to re-review before we can merge.

ognis1205:
@creechy

Just checking in—any updates on this?

creechy (approved):
lgtm. I did a quick Google auth test with the CLI worked fine for me.

ognis1205:
@creechy @jamieknight-db 

Thank you for patiently reviewing it!

ognis1205:
@creechy @jamieknight-db 

I think this is ready to be merged. Can we merge it?

PR #777, opened 2024-12-03
Handle invalid function definitions gracefully

BenWilson2:
**PR Checklist**

**Description of changes**

The error that is raised at a Toolkit invocation time for invalid (wildcard defined) functions within the function_names list is a very vague and difficult to troubleshoot pydantic field validation error that requires knowledge of the integration's implementation in order to troubleshoot. 
This PR fixes this for the integration toolkits by explicitly defining why the failure is occurring from within the model validation logic post-field-validation.

### DIFF
--- a/ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
+++ b/ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
@@ -1,7 +1,7 @@
 import base64
 import datetime
 import warnings
-from typing import TYPE_CHECKING, Any, NamedTuple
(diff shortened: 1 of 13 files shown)


serena-ruan:
@BenWilson2 Why do we need this? I thought we support *, as it means we should include all functions in the catalog/schema, and inside process_function_names function we deal with it.

BenWilson2:
I'll be digging in to the error that I was trying to protect against here (AutoGen validation error from an invalid function declaration) and adjust the logic in a follow-up PR :)

PR #791, opened 2024-12-08
[MINOR] New Fields.CODE and Formatting

jaceklaskowski:
Minor auth-related changes:

1. New Fields.CODE
2. Use Fields constants whenever possible
3. Formatting

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/CatalogService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/CatalogService.java
@@ -53,7 +53,7 @@ public class CatalogService {
   }
 
   @Get("")
-  @AuthorizeExpression("#defer")
+  @AuthorizeExpression
   public HttpResponse listCatalogs(
       @Param("max_results") Optional<Integer> maxResults,
       @Param("page_token") Optional<String> pageToken) {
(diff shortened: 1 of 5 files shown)


jaceklaskowski:
@creechy Please review 🙏

server/src/main/java/io/unitycatalog/server/service/CatalogService.java
```diff
@@ -53,7 +53,7 @@ public HttpResponse createCatalog(CreateCatalog createCatalog) {
   }
 
   @Get("")
-  @AuthorizeExpression("#defer")
+  @AuthorizeExpression
```
creechy, line 56: not sure what this change is about. It doesn't really matter either way but using `#defer` helps identify that the authorization is intentionally done elsewhere/later.

creechy (changes requested):
Just one question, otherwise looks good.

server/src/main/java/io/unitycatalog/server/service/CatalogService.java
jaceklaskowski: Okey, I'll revert this change. I think it was IDEA to propose it as `#defer` is the default value.

creechy (approved):

PR #817, opened 2024-12-29
Add mlflow integration test as github workflow

kriscon-db:
**PR Checklist**

**Description of changes**

Added a github workflow that starts all necessary servers and tests the mlflow integration against the existing PR.

### DIFF
--- a/.github/workflows/mlflow-integration-tests.yml
+++ b/.github/workflows/mlflow-integration-tests.yml
@@ -42,6 +42,7 @@ jobs:
           python -m venv venv
           source venv/bin/activate
           pip install mlflow
+          pip install pytest
       - name: Start uc-oss server
         run: |
           build/sbt package
--- a/.github/workflows/mlflow-integration-tests.yml
+++ b/.github/workflows/mlflow-integration-tests.yml
@@ -24,6 +24,9 @@ env:
 jobs:
   uc-oss-integration-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.9', '3.10']
     timeout-minutes: 120
     permissions: {}
     steps:
@@ -36,18 +39,19 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.9'
+          python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
           pip install mlflow
           pip install pytest
-      - name: Start uc-oss server
+      - name: Start Unity Catalog Server
         run: |
-          build/sbt package
-          nohup bin/start-uc-server > uc-server.log 2>&1 &
-
+          ai/dev/start_unitycatalog.sh
       - name: Run tests for UnityCatalog
         run: |
           echo "Running tests for UnityCatalog..."
-          export PYTHONPATH=$MLFLOW_HOME:$PYTHONPATH
           pytest tests/test_mlflow_integration.py
+      - name: Stop Unity Catalog Server
+        if: always()
+        run: |
+          ai/dev/teardown_server.sh
--- a/.github/workflows/mlflow-integration-tests.yml
+++ b/.github/workflows/mlflow-integration-tests.yml
@@ -1,0 +1,50 @@
+name: MLflow-UC OSS Integration Test
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+permissions:
+  contents: read
+
+env:
+  MLFLOW_HOME: /home/runner/work/mlflow/mlflow
+  UC_OSS_INTEGRATION: true
+
+jobs:
+  uc-oss-integration-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    permissions: {}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+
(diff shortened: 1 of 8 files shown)


.github/workflows/mlflow-integration-tests.yml
```diff
@@ -0,0 +1,57 @@
+name: MLflow-UC OSS Integration Test
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
```
BenWilson2, line 7: Should there be any path restrictions here, or do you want to ensure that this suite runs for any changes in the repo?

.github/workflows/mlflow-integration-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+          pip install pytest
+      - name: Start Unity Catalog Server
+        run: |
+          ai/dev/start_unitycatalog.sh
+      - name: Run tests for UnityCatalog
+        run: |
+          echo "Running tests for UnityCatalog..."
+          pytest tests/test_mlflow_integration.py
+      - name: Stop Unity Catalog Server
+        if: always()
```
BenWilson2, line 55: Do we want to add this? IIRC, this makes the step/run not cancellable.

.github/workflows/mlflow-integration-tests.yml
kriscon-db: I think it should run for every PR.  Changes to ACLs, underlying database connectivity, etc could all impact the integration.  I'd end up covering most of the codebase if I pattern matched against everything that could impact the integration.

.github/workflows/mlflow-integration-tests.yml
kriscon-db: We can. I do see that github actually will just kill orphaned processes after the workflow ends though, so while good practice to clean up, I don't think there is a risk of tons of orphaned servers sticking around.

tests/test_mlflow_integration.py
```diff
@@ -0,0 +1,154 @@
...
+    schema = "default"
+    registered_model_name = "iris"
+    model_name = f"{catalog}.{schema}.{registered_model_name}"
+    mlflow.set_experiment("iris-uc-oss")
+    try:
+        mlflow.MlflowClient().get_registered_model(model_name)
+    except Exception as e:
+        e.args[0].startswith("NOT_FOUND")
+    else:
+        assert False, "Expected exception for missing model not raised"
```
BenWilson2, line 41:
```suggestion
        with pytest.raises(MlflowException, match="NOT_FOUND"):
            mlflow.MlflowClient().get_registered_model(model_name)
```
nit for consistency for our other tests

BenWilson2: (not sure if that's the right Exception - looks like it might be the UC proto error that's passed through?)

tests/test_mlflow_integration.py
```diff
@@ -0,0 +1,154 @@
...
+                # The signature is automatically inferred from the input example and
+                # its predicted output.
+                input_example=input_example,
+                registered_model_name=model_name,
+            )
+
+    try:
+        build_model()
+    except Exception as e:
+        assert False, f"Unhandled exception while attempting to build a model: {e}"
```
BenWilson2, line 66: Might want to just call the internal function - the stack trace should hopefully be insightful if there is a fitting error here

tests/test_mlflow_integration.py
```diff
@@ -0,0 +1,154 @@
...
+        predictions = loaded_model.predict(X_test)
+        iris_feature_names = datasets.load_iris().feature_names
+        result = pd.DataFrame(X_test, columns=iris_feature_names)
+        result["actual_class"] = y_test
+        result["predicted_class"] = predictions
+        assert result[:4] is not None
+    except Exception as e:
+        assert False, f"Unhandled exception while loading a model: {e}"
+
+    try:
```
BenWilson2, line 87: Similar as above - we could just let any issues bubble up naturally if the call to list artifacts fails with the server

tests/test_mlflow_integration.py
```diff
@@ -0,0 +1,154 @@
...
+        mlflow.artifacts.download_artifacts(
+            artifact_uri=f"models:/{model_name}/{model_version}",
+            dst_path=path,
+        )
+        requirements_path = f"{path}/requirements.txt"
+        assert os.path.exists(requirements_path), f"File {requirements_path} does not exist."
+        with open(requirements_path) as file:
+            lines = file.readlines()
+        assert len(lines) > 0
+    except Exception as e:
```
BenWilson2, line 114: Similar comment here with removing the try catch. I feel like pytest would fairly clearly indicate what's wrong if the download fails?

tests/test_mlflow_integration.py
```diff
@@ -0,0 +1,154 @@
...
+    assert len(rms) == 1
+    mvs = mlflow.MlflowClient().search_model_versions(f"name='{model_name}'")
+    assert len(mvs) == 1
+    mlflow.MlflowClient().delete_model_version(name=model_name, version=1)
+    mvs = mlflow.MlflowClient().search_model_versions(f"name='{model_name}'")
+    assert len(mvs) == 0
+    mlflow.MlflowClient().delete_registered_model(name=model_name)
+    rms = mlflow.MlflowClient().search_registered_models()
+    assert len(rms) == 0
+    try:
```
BenWilson2, line 149: This can be converted to a with pytest.raises() call to ensure that the model version is not present any longer

tests/test_mlflow_integration.py
kriscon-db: Yes.  The UC exception is whats returned afaict.

kriscon-db:
I screwed up the rebasing on this one and am starting a new PR here: https://github.com/unitycatalog/unitycatalog/pull/834

PR #821, opened 2025-01-03
Auto-tracing for retriever

annzhang-db:
**PR Checklist**

**Description of changes**
- When a retriever tool is used, automatically trace it as a retriever span.
- Documentation updates will follow separately
- Manual testing:
Screenshot 2025-01-09 at 2 25 02 PM


ai/integrations/langchain/src/unitycatalog/ai/langchain/toolkit.py
```diff
@@ -92,6 +94,24 @@ def func(*args: Any, **kwargs: Any) -> str:
...
+
+            if result.value:
+                try:
+                    output = ast.literal_eval(result.value)
+
+                    if is_valid_retriever_output(output):
+                        import mlflow
+                        from mlflow.entities import SpanType
+
+                        with mlflow.start_span(
```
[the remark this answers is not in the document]
annzhang-db: Hmm... wondering how we could do that considering that we need to make sure it is a retriever first and currently we are doing that by checking that the output is formatted "like a retriever". We could start the span and then somehow get rid of that extra span afterwards if it is not a retriever, but I don't think we would be able to capture errors unless there is another way that a tool is marked as a retriever.

ai/integrations/autogen/src/unitycatalog/ai/autogen/toolkit.py
```diff
@@ -172,6 +173,7 @@ def func(**kwargs: Any) -> str:
             result = client.execute_function(
                 function_name=function_name,
                 parameters=args_json,
+                autologging_enabled=autologging_is_enabled("anthropic"),
```
rmalani, line 176: Perhaps you copy/pasted and forgot to update for each integration? All say "anthropic," but likely should reflect the name of each framework.

ai/integrations/langchain/src/unitycatalog/ai/langchain/toolkit.py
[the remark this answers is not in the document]
B-Step62: Ah that makes sense.. In that case, can we use [client API](https://mlflow.org/docs/latest/llms/tracing/index.html#tracing-client-apis) to create a trace with actual start time, end time? The code becomes more verbose, but I think setting right timestamp is important.

```

start_time_ns = time.time_ns()

result = None
exception = None
try:
    result = func(**parameters)
    end_time_ns = time.time_ns()
    auto_trace_retriever(function_name, parameters, result, start_time_ns, end_time_ns)
    ...

def auto_trace_retriever(function_name, parameters, result, start_time_ns, end_time_ns):
    ...
    if is_valid_retriever_output(output):
        client = MlflowClient()
        common_params = dict(
            name=function_name, 
            span_type=SpanType.RETRIEVER,
            inputs=parameters
            start_time_ns=start_time_ns
        )
        if parent_span := mlflow.get_current_active_span():
              span = client.start_span(
                  request_id=parent_span.request_id,
                  parent_span_id=parent_span.span_id,
                  **common_params
              )
              client.end_span(
                  request_id=span.request_id,
                  span_id=span.span_id,
                  outputs=output,
                  end_time_ns=end_time_ns
              )
        else:
              span = client.start_trace(**common_params)
              client.end_trace(
                  request_id=span.request_id,
                  outputs=output,
                  end_time_ns=end_time_ns
              )
```

A few key points:
* Client API requires explicit parent information unlike mlflow.start_span().
* We need to use different API depending on if it is root span or not.
   * if root -> start_trace / end_trace
   * otherwise -> start_span / end_span
   * (this is a common source of redundancy so we are going to add some utility to consolidate two, but it is not available in current or older mlflow....)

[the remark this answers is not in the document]
annzhang-db: Thanks so much for the pointers - got this to work!

B-Step62 (approved):
LGTM!

ai/core/src/unitycatalog/ai/core/utils/function_processing_utils.py
```diff
@@ -304,3 +308,48 @@ def supported_function_info_types():
...
     return types
+
+
+def auto_trace_retriever(
+    function_name: str,
+    parameters: Dict[str, Any],
+    result: str,
+    start_time_ns: int,
+    end_time_ns: int,
+):
```
BenWilson2, line 319: Can we add a docstring to this to explain to others what this is doing?

ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
BenWilson2, line 1: Let's add docstrings to these utilities

BenWilson2 (commented):
Can you add an integration test for this (mocking the actual response from the provider / service calls) functionality within the OpenAI integration suite? 
I'd like for us to verify that the traces are captured properly within MLflow just to be on the safe side and to ensure that the unitycatalog-ai implementation maintains trace collection parity with any changes on the MLflow side.

ai/core/src/unitycatalog/ai/core/utils/function_processing_utils.py
```diff
@@ -304,3 +308,59 @@ def supported_function_info_types():
...
+                    end_time_ns=end_time_ns,
+                )
+            else:
+                span = client.start_trace(**common_params)
+                client.end_trace(
+                    request_id=span.request_id, outputs=output, end_time_ns=end_time_ns
+                )
+    except Exception as e:
+        # Ignoring exceptions because auto-tracing retriever is not essential functionality
+        pass
```
serena-ruan, line 366: Let's add a DEBUG log here

ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
```diff
@@ -140,3 +140,55 @@ def validate_function_name_length(function_name: str) -> None:
...
+                return True
+            except TypeError:
+                return False
+
+        return False
+
+    return all(is_valid_retriever_item(item) for item in output)
+
+
+def autologging_is_enabled(integration_name):
```
serena-ruan, line 177: Could we rename to something like `mlflow_tracing_enabled`? (since there're flavors that don't support tracing in autologging) And we should check if `log_traces` is set to True or not right, though by default it's True.

ai/integrations/openai/tests/test_openai_utils.py
```diff
@@ -41,6 +41,41 @@ def test_generate_tool_call_messages(client: DatabricksFunctionClient):
         }
 
 
+def test_generate_tool_call_messages_with_tracing(client: DatabricksFunctionClient):
```
serena-ruan, line 44: Could we add such test for other integrations as well?

ai/integrations/openai/tests/test_openai_utils.py
```diff
@@ -41,6 +41,41 @@ def test_generate_tool_call_messages(client: DatabricksFunctionClient):
...
+            return_value=FunctionExecutionResult(format="SCALAR", value=result),
+        ),
+    ):
+        import mlflow
+
+        mlflow.openai.autolog()
+
+        generate_tool_call_messages(response=response, client=client)
+
+        trace = mlflow.get_last_active_trace()
```
serena-ruan, line 71: Could you test E2E that trace can be captured in test_openai_toolkit.py?

ai/core/src/unitycatalog/ai/core/base.py
```diff
@@ -154,7 +156,14 @@ def execute_function(
             function_info = self.get_function(function_name, **kwargs)
             parameters = parameters or {}
             self.validate_input_params(function_info.input_params, parameters)
-            return self._execute_uc_function(function_info, parameters, **kwargs)
+
+            start_time_ns = time.time_ns()
+            result = self._execute_uc_function(function_info, parameters, **kwargs)
+            end_time_ns = time.time_ns()
+
+            if kwargs.get("enable_trace_as_retriever", False):
```
serena-ruan, line 164: This param name sounds a bit weird, we're tracing retriever instead of `trace as retriever`?

ai/core/src/unitycatalog/ai/core/utils/function_processing_utils.py
```diff
@@ -304,3 +310,72 @@ def supported_function_info_types():
...
+                    end_time_ns=end_time_ns,
+                )
+            else:
+                span = client.start_trace(**common_params)
+                client.end_trace(
+                    request_id=span.request_id, outputs=output, end_time_ns=end_time_ns
+                )
+    except Exception as e:
+        # Ignoring exceptions because auto-tracing retriever is not essential functionality
+        _logger.error(
```
serena-ruan, line 378:
```suggestion
        _logger.debug(
```

ai/core/src/unitycatalog/ai/core/utils/function_processing_utils.py
```diff
@@ -304,3 +310,72 @@ def supported_function_info_types():
...
+        parameters: The input parameters to the function.
+        result: The output result of the function.
+        start_time_ns: The start time of the function in nanoseconds.
+        end_time_ns: The end time of the function in nanoseconds.
+    """
+    try:
+        if result.format == "CSV":
+            cleaned_result = result.value.encode("utf-8").decode("unicode_escape")
+            csv_buffer = StringIO(cleaned_result)
+            reader = csv.DictReader(csv_buffer)
```
serena-ruan, line 337: Is it easier if we just use `pd.read_csv(StringIO(value))`

ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
```diff
@@ -140,3 +140,60 @@ def validate_function_name_length(function_name: str) -> None:
...
+
+    def is_valid_retriever_item(item: Any) -> bool:
+        from mlflow.entities import Document
+
+        if isinstance(item, Document):
+            return True
+
+        if isinstance(item, dict):
+            try:
+                _ = Document(**item)
```
serena-ruan, line 170:
```suggestion
                Document(**item)
```

ai/integrations/autogen/tests/test_autogen_toolkit.py
```diff
@@ -210,6 +215,51 @@ def test_uc_function_to_autogen_tool(client):
...
+        import mlflow
+
+        trace = mlflow.get_last_active_trace()
+        assert trace is not None
+        assert trace.info.execution_time_ms is not None
+        assert trace.data.request == '{"x": "some input"}'
+        assert trace.data.response == RETRIEVER_OUTPUT_SCALAR
+        assert trace.data.spans[0].name == f"catalog.schema.test_{format}"
+
+        mlflow.autogen.autolog(disable=True)
```
serena-ruan, line 260: Do we need this?

ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
```diff
@@ -140,3 +140,60 @@ def validate_function_name_length(function_name: str) -> None:
...
+                return True
+            except TypeError:
+                return False
+
+        return False
+
+    return all(is_valid_retriever_item(item) for item in output)
+
+
+def mlflow_tracing_enabled(integration_name):
```
serena-ruan, line 180:
```suggestion
def mlflow_tracing_enabled(integration_name: str) -> bool:
```

serena-ruan (approved):
Overall LGTM! Left some minor comments

ai/core/src/unitycatalog/ai/core/base.py
annzhang-db: Yeah, I intended it to mean that we were enabling tracing the function as a retriever since tracing is enabled anyways but it is kinda confusing 🤔 ... let me rename it to `enable_retriever_tracing`?

shivanshuraj1333:
Is there any suggested way we should be adding in the docs to export traces? I see the `mlflow` client is built on top of `otel` and traces can be exported to otel-collector, but is there any other recommended way that we want to suggest while enabling the flag?

ai/integrations/autogen/tests/test_autogen_toolkit.py
annzhang-db: I disabled autologging at the end of each test so it doesn't affect any of the existing tests (specifically, execute_function should be called with enable_retriever_tracing = False

ai/core/src/unitycatalog/ai/core/base.py
```diff
@@ -154,7 +156,14 @@ def execute_function(
             function_info = self.get_function(function_name, **kwargs)
             parameters = parameters or {}
             self.validate_input_params(function_info.input_params, parameters)
-            return self._execute_uc_function(function_info, parameters, **kwargs)
+
+            start_time_ns = time.time_ns()
+            result = self._execute_uc_function(function_info, parameters, **kwargs)
+            end_time_ns = time.time_ns()
+
+            if kwargs.get("enable_retriever_tracing", False):
```
smurching, line 164: @serena-ruan @annzhang-db quick question, why is the default here `False`? We should have auto-tracing of retrievers that implement the MLflow retriever schema by default IMO, I don't know if people will discover/use this feature otherwise

smurching: Also, is there a way we can check signature upfront and use that to trace? It would be nice to start the span before function execution in case it fails etc

annzhang-db: @smurching The default is False because `enable_retriever_tracing` is passed in from each framework when it calls the tool, so if it's being used with `openai`, `langchain`, etc. it will be automatically traced with `mlflow.autolog()` but it won't be traced if it is called directly. We could default to True if we think that's the desired default behavior for invoking execute_function directly, but it would make us inconsistent with autologging enabled or disabled.

PR #834, opened 2025-01-07
Add mlflow integration test as github workflow

kriscon-db:
**PR Checklist**

**Description of changes**

Added a github workflow that starts all necessary servers and tests the mlflow integration against the existing PR.

### DIFF
--- a/.github/workflows/mlflow-integration-tests.yml
+++ b/.github/workflows/mlflow-integration-tests.yml
@@ -1,0 +1,55 @@
+name: MLflow-UC OSS Integration Test
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash --noprofile --norc -exo pipefail {0}
+
+env:
+  MLFLOW_HOME: /home/runner/work/mlflow/mlflow
+  UC_OSS_INTEGRATION: true
+
+jobs:
+  uc-oss-integration-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.9', '3.10']
+    timeout-minutes: 30
+    permissions: {}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          pip install mlflow
+          pip install pytest
+      - name: Start Unity Catalog Server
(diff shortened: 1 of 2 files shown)


BenWilson2 (approved):
LGTM based on review in #817

.github/workflows/mlflow-integration-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+          distribution: 'temurin'
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          pip install mlflow
+          pip install pytest
+      - name: Start Unity Catalog Server
```
tdas, line 47: nit: @kriscon-db @BenWilson2 shouldnt the test scripts like `tests/test_mlflow_integration.py` itself start and stop the UC server (just like this scripts starts mlflow server?

BenWilson2: For the integration tests in Unity Catalog AI, we have a single instance start of the server to eliminate the need to rebuild and restart the server over and over again for the 6 (soon to be 10) integrations. It shaves off ~5 minutes for each test by doing it this way. 
For MLflow's server, however, the startup time is negligible (typically <1s) and process teardown is similarly fast. This fixture is copied from how we run that server within the MLflow test suites and it is used within a few dozen of our suites. 
I feel like reusing the logic that Unity Catalog AI needs to use for its e2e tests for this suite makes sense so that we don't have multiple implementations of doing the same effective thing. Furthermore, when additional suites need to be added (for the hopeful future integration of UC as a backend tracking store for MLflow), we would likely be having many suites that would run as stages within this CI workflow (and we'd only want to have to start that server once to prevent CI runtime bloat).

shivanshuraj1333: IMO start and stoping UC server as a separate step makes sense, since say for some reason UC server fails to stop/teardown, the gh action would explicitly call it out, otherwise, one would need to check logs of the tests, if the tests is failing or there was some problem with uc itself.

That said, some other tests follows the similar pattern eg https://github.com/unitycatalog/unitycatalog/blob/main/.github/workflows/ucai-integration-tests-oss.yml#L54

wdyt @tdas ?

shivanshuraj1333 (approved):

.github/workflows/mlflow-integration-tests.yml
tdas: @BenWilson2 that makes sense. thank you for the context. I agree this is better.
@shivanshuraj1333 thats a very good point. But its always a tradeoff between make a self-contained-but-harder-to-debug blackbox script vs exposing all the steps that the caller has explicitly do. How I usually deal with the tradeoff is the chance of failure of these steps -- if the chance of UC server failing to start is practically zero, then the convenience of self-container script wins as no dev has to even care about its existence, let alone debug it. However, I can totally see that with UC server not yet being as mature as say MLFlow server, the probability of failure is non-zero.

tdas (approved):

.github/workflows/mlflow-integration-tests.yml
shivanshuraj1333: makes sense!

PR #840, opened 2025-01-09
[DRAFT] Set and check nonce during authentication sequence.

creechy:
**PR Checklist**

**Description of changes**

This fixes a potential replay attack by setting a `nonce` value and validating the `nonce` is returned in the id-token.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -184,6 +184,8 @@ lazy val client = (project in file("target/clients/java"))
       "com.google.code.findbugs" % "jsr305" % "3.0.2",
       "jakarta.annotation" % "jakarta.annotation-api" % "3.0.0" % Provided,
 
+      "com.auth0" % "java-jwt" % "4.4.0",
+
       // Test dependencies
       "org.junit.jupiter" % "junit-jupiter" % "5.10.3" % Test,
       "net.aichler" % "jupiter-interface" % JupiterKeys.jupiterVersion.value % Test,
--- a/examples/cli/src/main/java/io/unitycatalog/cli/utils/Oauth2CliExchange.java
+++ b/examples/cli/src/main/java/io/unitycatalog/cli/utils/Oauth2CliExchange.java
@@ -6,6 +6,8 @@ import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.mapping;
 import static java.util.stream.Collectors.toList;
 
+import com.auth0.jwt.JWT;
+import com.auth0.jwt.interfaces.DecodedJWT;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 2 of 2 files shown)


creechy:
@dan1elt0m - proposed enhancement to add `nonce` validation

dan1elt0m:
LGTM

dan1elt0m:
Nonce provides replay protection and according to Google's openid-connect [documentation](https://developers.google.com/identity/openid-connect/openid-connect#authenticationuriparameters), the nonce is required for both the server (auth code flow) and the implicit flow. IIRC, in the auth code flow it prevents hackers from injecting their auth response into the redirect URI of the server.

ognis1205:
The "client" mentioned above was intended to function as a client for the resource server. Strictly speaking, a CLI application is not a client of the resource server; however, since the nonce lifespan aligns with the session of the CLI application, I referred to it as a client. I understand that the nonce is associated with the client session to prevent replay attacks, but I believe it is highly unlikely for a malicious user to launch a replay attack on a CLI application.

Although I am uncertain about the rationale behind Google's documentation specifying nonce as required, [the OIDC specifications referenced in the same documentation](https://openid.net/specs/openid-connect-basic-1_0.html) appear to indicate that nonce is optional in the authorization code flow. In a standard authorization code flow, PKCE and state seem to provide sufficient security. That said, I do not have a strong objection to this draft.

creechy:
@ognis1205 @dan1elt0m I can see both your points. I don't think there is any particular not to do this, however optional and unlikely a replay attack could be.

ognis1205:
As mentioned in [the comments here](https://github.com/unitycatalog/unitycatalog/blob/6cd6a7bbdb102ad61ae39d6020e28ae8e30ee341/examples/cli/src/main/java/io/unitycatalog/cli/utils/Oauth2CliExchange.java#L74), if we continue using the current implementation of the Authorization Code Flow in AuthCli, I believe credentials such as tokens should also be handled via the back channel. However, if we replace it with the Authorization Code Flow with PKCE and remove the `client_secret` from the request, the CLI could be considered a public native application, which would conform to standard security implementations. (Even in this case, I think it would be necessary to introduce a separate property rather than reusing the server property.) That said, in Google's OIDC implementation, the `client_secret` seems to be treated as a required field, so the ideal approach would be to switch to the Authorization Code Flow with PKCE while clearly documenting this point. What do you think?

[Google OIDC Configuration Endpoint](https://accounts.google.com/.well-known/openid-configuration)
[Google OIDC Documentation](https://developers.google.com/identity/openid-connect/openid-connect)

ognis1205:
As mentioned in [the comment here](https://github.com/unitycatalog/unitycatalog/blob/6cd6a7bbdb102ad61ae39d6020e28ae8e30ee341/examples/cli/src/main/java/io/unitycatalog/cli/utils/Oauth2CliExchange.java#L74), if we continue using the current implementation of the Authorization Code Flow in AuthCli, I believe credentials such as tokens should also be handled via the back channel. 

However, if we replace it with the Authorization Code Flow with PKCE (PKCE Flow) and remove the `client_secret` from the request, the CLI could be considered a public native application, which would conform to standard security implementations (Even in this case, I think it would be necessary to introduce a separate property rather than reusing the server property).

That said, in Google's OIDC implementation, the `client_secret` seems to be treated as a required field, even if it is PKCE Flow, so the ideal approach would be to switch to the PKCE Flow while clearly documenting this point.

Additionally, since the UI implementation for both Google and Okta clients is equivalent to the Authorization Code Flow with PKCE, adopting this approach would enhance consistency across the system as a whole.

[Google OIDC Configuration Enpoint](https://accounts.google.com/.well-known/openid-configuration)
[Google OIDC Documentation](https://developers.google.com/identity/openid-connect/openid-connect#exchangecode)

PR #842, opened 2025-01-10
Improvements to the server code

vikrantpuppala:
**PR Checklist**

**Description of changes**

This PR adds code improvements to the server namely:
1. Allow injection of cloud credential vendor classes
2. Builder pattern to initialize server
3. Break down server init to smaller functions
4. Reduce staticness associated with serverProperties (future PR to remove this completely)

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
+++ b/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
@@ -27,6 +27,9 @@ import io.unitycatalog.server.security.SecurityConfiguration;
 import io.unitycatalog.server.security.SecurityContext;
 import io.unitycatalog.server.service.*;
 import io.unitycatalog.server.service.credential.CredentialOperations;
+import io.unitycatalog.server.service.credential.aws.AwsCredentialVendor;
+import io.unitycatalog.server.service.credential.azure.AzureCredentialVendor;
+import io.unitycatalog.server.service.credential.gcp.GcpCredentialVendor;
 import io.unitycatalog.server.service.iceberg.FileIOFactory;
 import io.unitycatalog.server.service.iceberg.MetadataService;
 import io.unitycatalog.server.service.iceberg.TableConfigService;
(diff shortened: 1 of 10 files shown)


server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -43,68 +46,101 @@
...
+      GcpCredentialVendor gcpCredentialVendor = new GcpCredentialVendor(builder.serverProperties);
+      CredentialOperations credentialOperations =
+          new CredentialOperations(awsCredentialVendor, azureCredentialVendor, gcpCredentialVendor);
+      builder.credentialOperations(credentialOperations);
+    }
+  }
 
-    ServerBuilder sb = Server.builder().serviceUnder("/docs", new DocService()).http(port);
-    addServices(sb);
+  private Server initializeServer(Builder unityCatalogServerBuilder) {
```
tdas, line 93: nit: this is named just builder in the constructor. lets be consistent.

tdas: i see that you are trying to differentiate from armeriaServerBuilder. thats okay then.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -43,68 +46,101 @@
...
+
+    return armeriaServerBuilder.build();
   }
 
-  private void addServices(ServerBuilder sb) {
-    ObjectMapper unityMapper =
-        JsonMapper.builder().disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES).build();
-    JacksonRequestConverterFunction unityConverterFunction =
-        new JacksonRequestConverterFunction(unityMapper);
+  private JacksonRequestConverterFunction createRequestConverterFunction() {
```
tdas, line 109: do you need these two functions in multiple places? If not, why create a separate function and not define it inline in the addApiServices

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -127,80 +163,122 @@ private void addServices(ServerBuilder sb) {
...
+    addIcebergServices(
+        armeriaServerBuilder,
+        unityCatalogServerBuilder.serverProperties,
+        unityCatalogServerBuilder.credentialOperations,
+        catalogService,
+        schemaService,
+        tableService);
+  }
+
+  private void addIcebergServices(
```
tdas, line 217: addIcebergApiServices? to mirror the addApiServices?

server/src/main/java/io/unitycatalog/server/service/credential/CredentialOperations.java
```diff
@@ -25,10 +25,10 @@ public class CredentialOperations {
   private final AzureCredentialVendor azureCredentialVendor;
   private final GcpCredentialVendor gcpCredentialVendor;
 
-  public CredentialOperations() {
-    this.awsCredentialVendor = new AwsCredentialVendor();
-    this.azureCredentialVendor = new AzureCredentialVendor();
-    this.gcpCredentialVendor = new GcpCredentialVendor();
+  public CredentialOperations(AwsCredentialVendor awsCredentialVendor, AzureCredentialVendor azureCredentialVendor, GcpCredentialVendor gcpCredentialVendor) {
```
tdas, line 28: please make this better multi line by having the one param in each line

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -43,68 +46,101 @@
...
-    securityContext =
+    this.securityContext =
         new SecurityContext(configurationFolder, securityConfiguration, "server", INTERNAL);
+    this.serverProperties = builder.serverProperties;
+    this.server = initializeServer(builder);
+  }
+
+  private void setDefaults(Builder builder) {
+    if (builder.port == 0) {
+      builder.port(8080);
```
tdas, line 77: can we put this port number as a constant somewhere?

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -237,4 +315,29 @@ private void printArt() {
             + "###################################################################\n";
     System.out.println(art);
   }
+
+  public static class Builder {
```
tdas, line 319: does the constructor for this need to be public? if not, we can make the constructor private this UnityCatalog class and just expose a public static method `builder()` that returns a new builder instance.

tdas: this mimic are armeria builder. let just borrow the same pattern :)

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -127,80 +163,122 @@ private void addServices(ServerBuilder sb) {
...
         .annotatedService(
-            basePath + "temporary-table-credentials", temporaryTableCredentialsService)
+            BASE_PATH + "temporary-table-credentials",
+            temporaryTableCredentialsService,
+            requestConverterFunction)
         .annotatedService(
-            basePath + "temporary-volume-credentials", temporaryVolumeCredentialsService)
+            BASE_PATH + "temporary-volume-credentials",
+            temporaryVolumeCredentialsService,
+            requestConverterFunction)
```
tdas, line 198: these were not passed before? why this change? what is the impact of this change?

vikrantpuppala: this change makes the service accept the request even if there are some unknown properties, this does not have an impact unless a client is built using a superset of the published openapi spec and calls this server. all of the services shouldn't be affect by this as of now.

tdas (approved):
Thank you for doing this. This will simplify a lot of future development.

PR #847, opened 2025-01-12
Move away from singleton pattern for properties and repositories 

vikrantpuppala:
**PR Checklist**

**Description of changes**

Currently, the properties and repositories classes are singleton and not tied to a UnityCatalogServer object. This makes it difficult to configure the server without using file based properties or system env variables (also difficult to have multiple instances of the server for tests). These changes uses dependency injection across classes to help tie objects of these classes together instead of depending on the singleton pattern.


server/src/main/java/io/unitycatalog/server/persist/RepositoryFactory.java
```diff
@@ -0,0 +1,36 @@
+package io.unitycatalog.server.persist;
+
+import io.unitycatalog.server.persist.utils.FileOperations;
+import io.unitycatalog.server.utils.ServerProperties;
+import java.util.HashMap;
+import java.util.Map;
+import lombok.Getter;
+import org.hibernate.SessionFactory;
+
+public class RepositoryFactory {
```
tdas, line 10: please add class docs explaining what this is

server/src/main/java/io/unitycatalog/server/persist/RepositoryFactory.java
```diff
@@ -0,0 +1,36 @@
...
+  @Getter private final SessionFactory sessionFactory;
+  @Getter private final FileOperations fileOperations;
+
+  public RepositoryFactory(SessionFactory sessionFactory, ServerProperties serverProperties) {
+    this.sessionFactory = sessionFactory;
+    this.fileOperations = new FileOperations(serverProperties);
+    initializeRepositories();
+  }
+
+  public void initializeRepositories() {
```
tdas, line 21: why is this public method? is anybody going to call this explicitly?

tdas: and why is it even a separate method .. and not just in the constructor?

server/src/main/java/io/unitycatalog/server/persist/RepositoryFactory.java
tdas: nit: is this really a "factory pattern". typically when you invoke `factory.getX(typeOfX)`, each invocation returns a new object. This class is just returning the same object every time for the same time.

tdas: maybe just call it Repositories?

server/src/main/java/io/unitycatalog/server/persist/RepositoryFactory.java
```diff
@@ -0,0 +1,36 @@
...
+    repositories.put(TableRepository.class, new TableRepository(this, sessionFactory));
+    repositories.put(VolumeRepository.class, new VolumeRepository(this, sessionFactory));
+    repositories.put(UserRepository.class, new UserRepository(this, sessionFactory));
+    repositories.put(MetastoreRepository.class, new MetastoreRepository(this, sessionFactory));
+    repositories.put(FunctionRepository.class, new FunctionRepository(this, sessionFactory));
+    repositories.put(ModelRepository.class, new ModelRepository(this, sessionFactory));
+  }
+
+  @SuppressWarnings("unchecked")
+  public <T> T getRepository(Class<T> repositoryClass) {
```
tdas, line 33: also .. i dont get this model. There is a fixed number of repositories that we are tracking within this class. why put them in a map and then make the caller specify the class explicitly and all this complex calling? why not just make simply have fields+getters for each repo? then calling reduces from 
`repositoryFactory.getRepository(UserRepository.class)` to simply `repositoryFactory.getUserRepository()`

server/src/test/java/io/unitycatalog/server/utils/FileUtilsTest.java
```diff
@@ -1,43 +0,0 @@
...
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
-
-import io.unitycatalog.server.exception.BaseException;
-import io.unitycatalog.server.persist.utils.FileUtils;
-import org.junit.jupiter.api.Test;
-
-public class FileUtilsTest {
-
-  @Test
-  public void testFileUtils() {
```
tdas, line 13: dont need this test any more? where is it moved?

server/src/main/java/io/unitycatalog/server/utils/IdentityUtils.java
```diff
@@ -23,13 +18,4 @@ public static String findPrincipalEmailAddress() {
       return null;
     }
   }
-
-  public static UUID findPrincipalId() {
```
tdas, line 27: not used any more?

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
```diff
@@ -61,22 +59,6 @@ public static String getAssetFullName(
     return catalogName + "." + schemaName + "." + assetName;
   }
 
-  public static UUID getSchemaId(Session session, String catalogName, String schemaName) {
```
tdas, line 64: where are all these deleted code gone? not needed any more?

server/src/test/java/io/unitycatalog/server/utils/FileUtilsTest.java
vikrantpuppala: This utils file itself had a lot of methods that weren't needed (were written for managed locations), both those methods and tests have been removed

server/src/main/java/io/unitycatalog/server/utils/IdentityUtils.java
vikrantpuppala: Moved to UserRepository

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
vikrantpuppala: this was needed only by the modelRepository and hence, moved it there

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -33,63 +31,6 @@ private enum Operation {
...
-   * re-instantiated with each test. Property updates shouldn't affect the instantiated class and we
-   * should require a server restart if the properties file is updated.
-   */
-  private static void reset() {
-    modelStorageRootPropertyCached = null;
-    modelStorageRootCached = null;
-  }
-
-  // Model specific storage root handlers and convenience methods
-  private static String getModelStorageRoot() {
```
vikrantpuppala, line 50: these are moved to FileOperations since these are non static methods

tdas (approved):
Thank you for doing this. This will simplify a lot of future development.

PR #849, opened 2025-01-13
Add tests for temporary credentials services using mocks

vikrantpuppala:
**PR Checklist**

**Description of changes**

Add tests for temporary credentials services using mocks


server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
```diff
@@ -0,0 +1,62 @@
...
+        AzureCredential.builder()
+            .sasToken("test-sas-token")
+            .expirationTimeInEpochMillis(System.currentTimeMillis() + 6000)
+            .build();
+    doReturn(azureCredential).when(azureCredentialVendor).vendAzureCredential(any());
+
+    gcpCredentialVendor = mock(GcpCredentialVendor.class);
+    AccessToken accessToken =
+        new AccessToken("test-token", Date.from(Instant.now().plusSeconds(10 * 60)));
+    doReturn(accessToken).when(gcpCredentialVendor).vendGcpToken(any());
```
vikrantpuppala, line 57: It would've been better if the mock does some check rather than simply returning. Something like: if credentials exist in serverProperties, only then return the temp credential. However, the current code falls back to DefaultCredentialsProvider impl when server.properties does not contain credentials. 

It is debatable if we should credentials at the system level, if they are not configured in server.properties. Should we mandatorily require the admin/user to configure the credentials in server.properties and not simply use system level credentials? @tdas

shivanshuraj1333: IIUC, you're talking about this case https://github.com/unitycatalog/unitycatalog/blob/be7b1652a2e7f9ab358220efb50ba7b2f25eea11/server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java#L56

the `gcsConfigurations` are populated by `serverProperties` if provided, but if serverProperties are not configured, the above tries to pull creds from what is being configured in the local system. (something a user would have configured using `gcloud auth`).

This looks like an acceptable behavior to me, however, I think (and you've mentioned debatable) it should be logged somewhere, so that user knows that default creds are pulled from local gcloud auth.

WDYT?

vikrantpuppala: I think we would rather not use credentials configured in the local system because it becomes difficult for UC to then federate these credentials (ensure only that the right people have access). We will remove this in future, keeping it out of scope for this PR.

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
```diff
@@ -0,0 +1,154 @@
...
+    serverProperties.put("adls.storageAccountName.0", "test-bucket0");
+    serverProperties.put("adls.tenantId.0", "tenantId0");
+    serverProperties.put("adls.clientId.0", "clientId0");
+    serverProperties.put("adls.clientSecret.0", "clientSecret0");
+  }
+
+  @Override
+  protected void setUpCredentialOperations() {
+    awsCredentialVendor = mock(AwsCredentialVendor.class);
+    Credentials credentials =
```
tdas, line 48: lets call this awsCredential

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
```diff
@@ -0,0 +1,154 @@
...
+                doReturn(azureCredential)
+                    .when(azureCredentialVendor)
+                    .vendAzureCredential(
+                        argThat(
+                            arg ->
+                                arg.getStorageScheme().equals("abfs")
+                                    && arg.getStorageBase().contains(e.getValue().toString()))));
+
+    gcpCredentialVendor = mock(GcpCredentialVendor.class);
+    AccessToken accessToken =
```
tdas, line 85: gcpCredential

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
```diff
@@ -0,0 +1,154 @@
...
+        .filter(e -> e.getKey().toString().contains("s3.bucketPath"))
+        .forEach(
+            e ->
+                doReturn(credentials)
+                    .when(awsCredentialVendor)
+                    .vendAwsCredentials(
+                        argThat(
+                            arg ->
+                                arg.getStorageScheme().equals("s3")
+                                    && arg.getStorageBase().contains(e.getValue().toString()))));
```
tdas, line 64: can this function de-triplicated?
its make this sooooo nested.

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
```diff
@@ -0,0 +1,154 @@
...
+                        argThat(
+                            arg ->
+                                arg.getStorageScheme().equals("gs")
+                                    && arg.getStorageBase().contains(e.getValue().toString()))));
+
+    credentialOperations =
+        new CredentialOperations(awsCredentialVendor, azureCredentialVendor, gcpCredentialVendor);
+  }
+
+  protected String getTestCloudPath(String scheme, boolean isConfiguredPath) {
```
tdas, line 103: what is `isConfiguredPath`? method and param docs please.

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
```diff
@@ -0,0 +1,154 @@
...
+    return switch (scheme) {
+        case "s3" -> "s3://" + bucket + "/test";
+        case "abfs", "abfss" -> "abfs://test-container@" + bucket + ".dfs.core.windows.net/test";
+        case "gs" -> "gs://" + bucket + "/test";
+        default -> throw new IllegalArgumentException("Invalid scheme");
+    };
+  }
+
+  protected void assertTemporaryCredentials(
+      TemporaryCredentials temporaryCredentials, String scheme) {
```
tdas, line 115: nit: lets try using smaller variable names to reduce bloat a bit. e.g., tempCreds

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
```diff
@@ -0,0 +1,154 @@
...
+    switch (scheme) {
+      case "s3":
+        assertThat(temporaryCredentials.getAwsTempCredentials()).isNotNull();
```
tdas, line 118: nit:  please reduce code duplication. why are we calling `temporaryCredentials.getAwsTempCredentials()` 4 times?

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
```diff
@@ -0,0 +1,154 @@
...
+        assertThat(temporaryCredentials.getGcpOauthToken().getOauthToken()).isEqualTo("test-token");
+        break;
+      default:
+        fail("Invalid scheme");
+        break;
+    }
+  }
+
+  /**
+   * @return Stream of arguments (s3, abfs, gs) x isConfiguredPath (true, false) for testing
```
tdas, line 143: what is isConfiguredPath??

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
```diff
@@ -0,0 +1,154 @@
...
+   */
+  protected static Stream<Arguments> provideTestArguments() {
```
tdas, line 145: nit: provide -> get

server/src/test/java/io/unitycatalog/server/base/BaseServerTest.java
```diff
@@ -10,19 +16,31 @@
...
 
   protected void setUpProperties() {
     serverProperties = new Properties();
     serverProperties.setProperty("server.env", "test");
   }
 
+  protected void setUpCredentialOperations() {
+    awsCredentialVendor = mock(AwsCredentialVendor.class);
+    azureCredentialVendor = mock(AzureCredentialVendor.class);
+    gcpCredentialVendor = mock(GcpCredentialVendor.class);
```
tdas, line 41: whats the point of defining these in the base class if the mocks are not even setup? 
I am guessing that any test that needs to work with mocked cred vendors will extend BaseCRUDTestWithMockCreds. So why not just move this entire thing completely in that class?

server/src/test/java/io/unitycatalog/server/sdk/tempcredential/SdkTemporaryCredentialOperations.java
```diff
@@ -0,0 +1,39 @@
+package io.unitycatalog.server.sdk.tempcredential;
```
tdas, line 1: why move everything to tempcredential package instead of keeping in credentials?

server/src/test/java/io/unitycatalog/server/sdk/tempcredential/SdkTemporaryCredentialOperations.java
```diff
@@ -0,0 +1,39 @@
...
+import io.unitycatalog.client.model.*;
+
+public class SdkTemporaryCredentialOperations {
+  private final TemporaryCredentialsApi temporaryCredentialsApi;
+
+  public SdkTemporaryCredentialOperations(ApiClient apiClient) {
+    this.temporaryCredentialsApi = new TemporaryCredentialsApi(apiClient);
+  }
+
+  public TemporaryCredentials generateTemporaryModelVersionCredentials(
```
tdas, line 15: i dont understand why this base class for cred operations is needed. 

generateTemporaryModelVersionCredentials is only going to be used only by one class SdkTemporaryModelVersionCredentialTest. Then why not define this method right in that class? 

Same for all table, volume, path?

server/src/test/java/io/unitycatalog/server/sdk/tempcredential/SdkTemporaryModelVersionCredentialTest.java
```diff
@@ -1,51 +1,88 @@
...
+      ServerConfig config) {
+    return new SdkTemporaryCredentialOperations(TestUtils.createApiClient(config));
+  }
+
+  String rootBase = "/tmp/" + UUID.randomUUID();
+
+  @Override
+  public void setUpProperties() {
+    super.setUpProperties();
+    serverProperties.setProperty("storage-root.models", rootBase);
```
tdas, line 62: `storage-root.` can we start defining this constant keys somewhere?

server/src/test/java/io/unitycatalog/server/sdk/tempcredential/SdkTemporaryModelVersionCredentialTest.java
```diff
@@ -1,51 +1,88 @@
...
+  protected SdkTemporaryCredentialOperations temporaryCredentialOperations;
 
-  protected abstract SchemaOperations createSchemaOperations(ServerConfig serverConfig);
+  @Override
+  protected CatalogOperations createCatalogOperations(ServerConfig config) {
+    return new SdkCatalogOperations(TestUtils.createApiClient(config));
+  }
 
-  protected abstract ModelOperations createModelOperations(ServerConfig serverConfig);
+  protected SchemaOperations createSchemaOperations(ServerConfig config) {
```
tdas, line 44: the implementations of all these 1 line `createXOperations` methods seems to be duplicated across all the test classes. why not put this into a base class / interace?

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
tdas: please make it getArgumentsForParameterizedTests

server/src/test/java/io/unitycatalog/server/sdk/tempcredential/SdkTemporaryModelVersionCredentialTest.java
```diff
@@ -213,9 +255,37 @@ public void testModelCRUD() throws ApiException {
...
+                temporaryCredentialOperations.generateTemporaryModelVersionCredentials(
                     generateUnknownOperation))
         .isInstanceOf(ApiException.class)
         .hasFieldOrPropertyWithValue("code", ErrorCode.INVALID_ARGUMENT.getHttpStatus().code());
   }
+
+  @ParameterizedTest
+  @MethodSource("provideTestArguments")
+  public void testGenerateTemporaryCredentialsWhereConfIsProvided(
+      String scheme, boolean isConfiguredPath) throws ApiException {
```
tdas, line 267: what isConfiguredPath

server/src/test/java/io/unitycatalog/server/sdk/tempcredential/SdkTemporaryCredentialOperations.java
vikrantpuppala: to not confuse with storage credentials

server/src/test/java/io/unitycatalog/server/sdk/tempcredential/SdkTemporaryModelVersionCredentialTest.java
vikrantpuppala: because it is not needed across all test classes like pathTests do not need schema/table operations whereas tableTests needs both

tdas (approved):

PR #853, opened 2025-01-17
Add a function cache invalidator to the UnityCatalogFunctionsClient

BenWilson2:
**PR Checklist**

**Description of changes**

Adds a public method to the UnityCatalogFunctionsClient to allow for invalidation of the execution cache for development purposes. If a function has been overwritten or modified in the current REPL scope, updates to a function will not reflect until either a new client is initialized or the REPL session has been terminated. It's an annoying devex that's fixed with the addition of this API.


serena-ruan (approved):
Should we automatically call this API when `replace=True` during function creation? Other than this LGTM!

BenWilson2:
> Should we automatically remove the function from cache when `replace=True` during function creation? Other than this LGTM!

This is a very excellent point! I'll update the logic in those APIs :)

PR #854, opened 2025-01-17
Fix integration tests

serena-ruan:
**PR Checklist**

**Description of changes**

### DIFF
--- a/ai/integrations/crewai/tests/test_crewai_toolkit.py
+++ b/ai/integrations/crewai/tests/test_crewai_toolkit.py
@@ -15,6 +15,7 @@ from unitycatalog.ai.core.base import (
     FunctionExecutionResult,
 )
 from unitycatalog.ai.test_utils.client_utils import (
+    TEST_IN_DATABRICKS,
     USE_SERVERLESS,
     client,  # noqa: F401
     get_client,
@@ -30,6 +31,12 @@ from unitycatalog.ai.test_utils.function_utils import (
 
 SCHEMA = os.environ.get("SCHEMA", "ucai_crewai_test")
 
+@pytest.fixture(autouse=True)
+def set_tracking_uri():
+    if TEST_IN_DATABRICKS:
+        import mlflow
+        
+        mlflow.set_tracking_uri("databricks-uc")
 
 @requires_databricks
 @pytest.mark.parametrize("use_serverless", [True, False])
--- a/ai/integrations/crewai/tests/test_crewai_toolkit.py
+++ b/ai/integrations/crewai/tests/test_crewai_toolkit.py
@@ -189,8 +189,7 @@ def test_crewai_tool_with_tracing_as_retriever(serverless_client, format: str, f
         import mlflow
         import mlflow.tracking._model_registry.utils
 
-        mlflow.set_tracking_uri(f"databricks://{PROFILE}")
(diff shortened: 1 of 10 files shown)


BenWilson2:
Does CrewAI need a mock for the auth layer?

ai/integrations/crewai/tests/test_crewai_toolkit.py
```diff
@@ -169,7 +170,7 @@ def test_uc_function_to_crewai_tool(client):
         ("CSV", RETRIEVER_OUTPUT_CSV),
     ],
 )
```
BenWilson2, line 172: Do we need a `@requires_databricks` here?

BenWilson2:
@serena-ruan I did a little digging on the failures on the Databricks side for the integration suites... 

Manually running the functions using the following function definition enabled me to get a correct interface. 
The newline encoding issue with those strings (there are nested double quotes in them) doesn't seem to be much of a problem on the OSS side of the house, but Databricks really doesn't like that encoding (likely due to differences in the REST interfaces that are used between OSS and Databricks UC). 

This was successful:

```python
from typing import List, Dict, Any, NamedTuple
from unitycatalog.ai.core.databricks import DatabricksFunctionClient

class FunctionInputOutput(NamedTuple):
    sql_body: str
    inputs: List[Dict[str, Any]]
    output: str

def function_with_scalar_retriever_output(func_name: str) -> FunctionInputOutput:
    retriever_output='''[{"page_content": "# Technology partners\\n## What is Databricks Partner Connect?\\n", "metadata": {"similarity_score": 0.010178182, "chunk_id": "0217a07ba2fec61865ce408043acf1cf", "url": "https://docs.databricks.com/partner-connect/walkthrough-fivetran.html"}}]'''

    sql_body=f"""CREATE FUNCTION {func_name}(query STRING)
RETURNS STRING
LANGUAGE PYTHON
AS $$
    return {retriever_output}
    $$
"""
    return FunctionInputOutput(
        sql_body=sql_body,
        inputs=[{"query": "What is Databricks Partner Connect?"}],
        output=retriever_output,
    )

CATALOG = "ben_wilson"
SCHEMA = "uc_func"

function_sample = function_with_scalar_retriever_output(f"{CATALOG}.{SCHEMA}.testing_retriever")

client = DatabricksFunctionClient()

#client.delete_function(f"{CATALOG}.{SCHEMA}.testing_retriever")  # here for debugging / replayability

client.create_function(sql_function_body=function_sample.sql_body)

client.execute_function(function_name=f"{CATALOG}.{SCHEMA}.testing_retriever", parameters={"query": "Hello"}, enable_retriever_tracing=True)
```

This generates a trace properly in the Notebook and in the artifact store for the experiment. 

However, the retriever_output string that is used within the CI suite doesn't have triple-escaped quotes and will 'silently' fail with a trace not being generated. What actually happens is that UC throws an exception about an encoding issue, preventing any of the validation checks for the presence of traces to work within the test suites.

serena-ruan:
https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/12861970790/job/35860662992 verifies this change work

BenWilson2 (approved):
Thanks for the encoding fixes and the UCMR adjustments! LGTM!

Issue #844, opened 2025-01-11
connecting with local minio


MammadTavakoli:
how can i conect to local minio and use it?

joshuagrisham:
Hi @MammadTavakoli ,

I actually looked into this a bit and got it working, but can just mention that it is not exactly trivial and will slightly distort some of the design concepts that seem to exist today (namely with this "Credential Vending" concept).

You can see an example branch I made with the comparison here for an idea on what it would look like to get this working:

https://github.com/unitycatalog/unitycatalog/compare/main...joshuagrisham:unitycatalog:add-s3-compatible-storage?expand=1

I have tested this myself using Minio and the Spark catalog class (`UCSingleCatalog`) as well as the example `uc` CLI client and it is working for me, but have not tested these changes against a real AWS bucket. I have also not tested the Iceberg support (just made some changes based on what I think probably might work...).

The biggest sticking point here is that this potential solution will essentially turn "credential vending" for AWS more into "S3 Client Connection Configuration" vending (if that makes sense?) where it will pass all of the necessary configuration settings in order to create an S3 Client Connection (including the credentials but also the endpoint, region, etc..). I also did not exactly make it a "first class citizen"; it is a semi-hack in a few places but probably "good enough" to get started.

I also cleaned up a few minor issues along the way including trying to fix support for using the file system scheme "s3://" the "right way" and not via string hacks, plus changed the config so that they will be read and parsed from properties only once instead of multiple times... 

If this seems interesting then I can try to slightly tweak what I did and open a PR?

mborodin:
Hi @MammadTavakoli 

I am currently testing my implementation that will allow usage of non-AWS S3 implementations in PR #858 
If you would like to try it you will need to use following configuration snippet:
```
    s3.bucketPath.0=s3://unity
    s3.endpoint.0=http://minio
    s3.provider.0=static

    aws.credentials.provider.static.class=io.unitycatalog.server.service.credential.aws.provider.StaticCredentialsProvider
    aws.credentials.provider.static.accessKey=access
    aws.credentials.provider.static.secretKey=secret
```

Also if you use Spark you will need to use locally built spark connector from target/clients/java/target/unitycatalog-client-0.3.0-SNAPSHOT.jar

MammadTavakoli:
Hi @joshuagrisham and @mborodin 

thanks for your recomennded . Its sound is prefect. But I want run unity catalog in docker. 

how can I deploye your solution in docker? in save it delta lake with pyspark?

PR #858, opened 2025-01-21
Allow non-AWS S3

mborodin:
**PR Checklist**

**Description of changes**
NOTE: This is a placeholder for PR until tests and documentation will be completed 

This PR reimplements handling of S3 and credentials vending to support usage outside of AWS ecosystem:
* Introduced AwsCredentialProvider that can be referenced from storage configuration
* S3 configuration now supports specifying S3 endpoints
* AwsTemporaryCredentials include endpoint information
* spark-connector configures credentials per bucket to allow different credentials for different table locations
* s3a:// scheme is properly handled in spark connector: outgoing locations send s3://, incoming are mutated to s3a://

This PR also solves #844


joshuagrisham:
Hi @mborodin , I don't want to be "that guy" but it seems like you have made some relatively big design decisions here and I assume there should be some buy-in from those who drive this project if they would even want the classes to be designed like this?

Also IMO using per-bucket config for S3AFileSystem e.g. like `fs.s3a.bucket.${bucket}.access.key` is a bit overkill and might be confusing since the use-case is that only one at a time will be used.

As mentioned in my comment in the issue here: https://github.com/unitycatalog/unitycatalog/issues/844#issuecomment-2603228411 the branch I showed where I added similar support for this (https://github.com/unitycatalog/unitycatalog/compare/main...joshuagrisham:unitycatalog:add-s3-compatible-storage?expand=1) has not really changed so much (actually up to half of the changes in this branch was were I fixed/optimized some things that don't even really impact this specific feature, e.g. change so we only read server properties from the file once instead of every time they are fetched, plus to remove this s3:// to s3a:// replace string hack and instead configure S3AFileSystem the right way) ..  

I can't say if that is a better solution or not but it definitely is a lot more of a minimal change to the design?

mborodin:
Hi @joshuagrisham,

you are correct that design changes are big, but there is a reason for that: some S3 providers might provide mechanism similar to STS to get credentials (OpenStack Keystone for instance) and this mechanism will allow to plug custom providers to allow smooth integration with non-AWS S3 providers. OpenAPI spec will be reverted back to it's original state before this PR will be ready since a) endpoint info really should not come with credentials and b) there is no easy way to tweak this kind of configuration dynamically, something more advanced should be implemented which in turn will bring even more drastic changes to architecture - reimplementation of whole S3 handling logic is neither in scope of nor intention of this PR.

I have marked this PR as a draft exactly because I want to get feedback from all interested parties and (re)implement some parts based on it before marking it as ready.

Also I disagree regarding using per-bucket config - I can came up with use cases, where such approach would be valid. For example: organization provides a bucket for Team A and a bucket for Team B to store their data. If Team C now wants to analyse data from both Team A and Team B simultaneously they can get vended credentials for both buckets and do their job.

gzusgw:
I pulled this PR and got an error when using spark-shell to create a table：
Caused by: com.amazonaws.services.s3.model.AmazonS3Exception: Forbidden (Service: Amazon S3; Status Code: 403; Error Code: 403 Forbidden; Request ID: 5C88F48QCHY9NN6S; S3 Extended Request ID: D4sxOD7tMAlafmaJW88H5uw1z4oPVFPoBlGTPrPIf/HnNddOHWrARVukaz9iFA06AkkPLO8JxzzoGx5W8M3wYc9r9UXrWoM4T8Bm2z7l308=; Proxy: 127.0.0.1), S3 Extended Request ID: D4sxOD7tMAlafmaJW88H5uw1z4oPVFPoBlGTPrPIf/HnNddOHWrARVukaz9iFA06AkkPLO8JxzzoGx5W8M3wYc9r9UXrWoM4T8Bm2z7l308=
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.handleErrorResponse(AmazonHttpClient.java:1912)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.handleServiceErrorResponse(AmazonHttpClient.java:1450)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeOneRequest(AmazonHttpClient.java:1419)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeHelper(AmazonHttpClient.java:1183)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.doExecute(AmazonHttpClient.java:838)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeWithTimer(AmazonHttpClient.java:805)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.execute(AmazonHttpClient.java:779)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.access$500(AmazonHttpClient.java:735)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutionBuilderImpl.execute(AmazonHttpClient.java:717)
	at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:581)
	at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:559)
	at com.amazonaws.services.s3.AmazonS3Client.invoke(AmazonS3Client.java:5605)
	at com.amazonaws.services.s3.AmazonS3Client.invoke(AmazonS3Client.java:5552)
	at com.amazonaws.services.s3.AmazonS3Client.getObjectMetadata(AmazonS3Client.java:1423)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.lambda$getObjectMetadata$10(S3AFileSystem.java:2545)
	at org.apache.hadoop.fs.s3a.Invoker.retryUntranslated(Invoker.java:414)
	at org.apache.hadoop.fs.s3a.Invoker.retryUntranslated(Invoker.java:377)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.getObjectMetadata(S3AFileSystem.java:2533)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.getObjectMetadata(S3AFileSystem.java:2513)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.s3GetFileStatus(S3AFileSystem.java:3776)
	... 172 more

PR #861, opened 2025-01-23
Check function signature instead of output for auto-tracing retrievers

annzhang-db:
**PR Checklist**

-> documenting this in docs for Agent Tools https://github.com/databricks-eng/docs/pull/19358

**Description of changes**
Following up from this [comment](https://github.com/unitycatalog/unitycatalog/pull/821#discussion_r1920539456), we want to modify auto-tracing for retrievers to check the function signature instead of the output format so that we can capture errors that occur for retrievers as well. Our recommended way for querying vector search in the [documentation](https://docs.databricks.com/en/generative-ai/agent-framework/unstructured-retrieval-tools.html#vector-search-retriever-tool-with-unity-catalog-functions) has data type `TABLE_TYPE` and we can match on this signature.

Manual testing:
- Langchain
Screenshot 2025-01-27 at 4 20 28 PM
Screenshot 2025-01-27 at 4 21 16 PM

- CrewAI
Screenshot 2025-01-27 at 4 53 10 PM
Screenshot 2025-01-27 at 5 21 02 PM

- LlamaIndex
Screenshot 2025-01-27 at 5 10 42 PM
Screenshot 2025-01-27 at 5 11 50 PM


ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
```diff
@@ -142,39 +147,53 @@ def validate_function_name_length(function_name: str) -> None:
...
+        bool: If the provided function has a valid retriever signature.
     """
-    if not isinstance(output, list):
+    if "TABLE_TYPE" not in str(function_info.data_type):
         return False
 
-    if len(output) < 1:
-        return False
+    full_data_type = function_info.full_data_type.strip()
+    if full_data_type.startswith("(") and full_data_type.endswith(")"):
```
smurching, line 165: Curious how robust this is, is there not another more-structured metadata field we could match against?

smurching: Looking at your example function, maybe we can use the `return_params` field?

```
"return_params": {
        "parameters": [
            {
                "name": "page_content",
                "type_text": "string",
                "type_json": "{\"name\":\"page_content\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}}",
                "type_name": "STRING",
                "type_precision": 0,
                "type_scale": 0,
                "position": 0,
                "parameter_type": "COLUMN"
            },
            {
                "name": "id",
                "type_text": "string",
                "type_json": "{\"name\":\"id\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}}",
                "type_name": "STRING",
                "type_precision": 0,
                "type_scale": 0,
                "position": 1,
                "parameter_type": "COLUMN"
            },
            {
                "name": "metadata",
                "type_text": "map<string,string>",
                "type_json": "{\"name\":\"metadata\",\"type\":{\"type\":\"map\",\"keyType\":\"string\",\"valueType\":\"string\",\"valueContainsNull\":true},\"nullable\":true,\"metadata\":{}}",
                "type_name": "MAP",
                "type_precision": 0,
                "type_scale": 0,
                "position": 2,
                "parameter_type": "COLUMN"
            }
        ]
    },
```

smurching (commented):
Looks great, just had a comment about trying to match against a more-structured field for type checking. Also we can start with this, but just curious - would it also be possible to later support UDFs that return a StructType with the appropriate schema? See e.g. https://chatgpt.com/share/6792c4fe-aed8-8009-9640-f47003b4203c

ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
```diff
@@ -158,42 +154,54 @@ def has_retriever_signature(function_info: "FunctionInfo") -> bool:
...
+        if function_info.return_params and function_info.return_params.parameters:
+            params = []
+            for param in function_info.return_params.parameters:
+                if hasattr(param, "as_dict"):
+                    params.append(param.as_dict())
+                else:
+                    params.append(dict(param))
+
+            param_dict = {param["name"]: param["type_name"] for param in params}
+            return param_dict.items() <= valid_columns_dict.items() and "page_content" in param_dict
```
smurching, line 169: Ok to change this later, but I think it's ok if the UDF includes extra columns besides the ones in MLflow's retriever schema, it should be future-proof to be permissive

ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
```diff
@@ -158,42 +154,54 @@ def has_retriever_signature(function_info: "FunctionInfo") -> bool:
...
 
-    full_data_type = function_info.full_data_type.strip()
-    if full_data_type.startswith("(") and full_data_type.endswith(")"):
-        full_data_type = full_data_type[1:-1]
+        full_data_type = function_info.full_data_type.strip()
 
-    # Split on commas but respect data types such as MAP<STRING, STRING>
-    columns = re.split(r",\s*(?![^<]*>)", full_data_type)
-    columns = [col.strip() for col in columns]
+        match = re.match(r"ARRAY<STRUCT<(.*)>>", full_data_type)
```
smurching, line 177: Ah nice - still curious how common this case is / worried about doing string matching, do you think it makes sense to remove this case for now and add it back later? If we do add it, we should check whether parsing this string representation is guaranteed to be stable

smurching (commented):
The updates for table type look great! Just one question on whether the StructType stuff is safe, if we're not sure we can ask, remove it for now, & add support for it in a followup PR instead, to unblock this one :) thanks!

ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
annzhang-db: Yeah, `return_params` isn't populated for non-TABLE types so it's difficult to match against. I can remove this logic for now and see if there's something better we can do here.

smurching: SGTM, can we ask for a recommendation from the SQL team in #eng-sql-lang for the future?

ai/core/src/unitycatalog/ai/core/utils/validation_utils.py
```diff
@@ -142,39 +142,30 @@ def validate_function_name_length(function_name: str) -> None:
         )
 
 
-def is_valid_retriever_output(output: Any) -> bool:
+def has_retriever_signature(function_info: "FunctionInfo") -> bool:
```
BenWilson2, line 145: To limit the iteration count and make the early exit a bit more readable, could we look into a refactor like this?

```python

def has_retriever_signature(function_info: "FunctionInfo") -> bool:
    """
    Checks if the given function signature follows the retriever format for MLflow, which is a
    list of Documents.

    Args:
        function_info: The function to determine if it has a valid retriever signature.

    Returns:
        bool: If the provided function has a valid retriever signature.
    """

    if "TABLE_TYPE" not in str(function_info.data_type):
        return False

    return_params = function_info.return_params

    if not (return_params and return_params.parameters):
        return False


    for param in return_params.parameters:

        param_dict = param.as_dict() if hasattr(param, "as_dict") else dict(param)
        
        # Early exit if condition matches
        if param_dict.get("name") == "page_content" and param_dict.get("type_name") == "STRING":
            return True

    return False
```

ai/core/tests/core/test_validation_utils.py
```diff
@@ -121,35 +117,116 @@ def test_check_function_info(function_info, expected_warnings):
...
+                        FunctionParameterInfo(
+                            name="extra_column",
+                            type_text="string",
+                            type_name=ColumnTypeName.STRING,
+                            position=2,
+                        ),
+                    ]
+                ),
+            ),
+            False,
```
BenWilson2, line 227: Should this be True since it as the `page_content` parameter entry with a string type?

ai/integrations/anthropic/tests/test_anthropic_utils.py
```diff
@@ -267,16 +273,32 @@ def test_generate_tool_call_messages_with_invalid_tool_use_block(mock_client, du
...
+    ],
+)
+def test_generate_tool_call_messages_with_tracing(
+    dummy_history, format, function_output, data_type, full_data_type, return_params
+):
     with mock.patch(
         "unitycatalog.ai.core.databricks.get_default_databricks_workspace_client",
         return_value=mock.Mock(),
     ):
+        function_mock = mock.MagicMock()
```
BenWilson2, line 289: Can we use a spec entry here to ensure that we're mocking the correct interface? `spec=FunctionInfo`
specifying `autospec=True` in the Mock definition should help to ensure that we're constructing the mock interface with the same properties as the underlying FunctionInfo classdef as well. 

For instance:

```python
function_mock = Mock(
    spec=FunctionInfo,
    name=f"catalog.schema.retriever_tool_{format}",
    full_name=f"catalog.schema.retriever_tool_{format}",
    data_type=data_type,
    full_data_type=full_data_type,
    return_params=return_params,
    autospec=True
)
```

ai/integrations/autogen/tests/test_autogen_toolkit.py
```diff
@@ -229,13 +232,23 @@ async def test_uc_function_to_autogen_tool(dbx_client):
...
+)
 @pytest.mark.parametrize("use_serverless", [True, False])
 def test_autogen_tool_with_tracing_as_retriever(
-    use_serverless, monkeypatch, format: str, function_output: str
+    use_serverless, monkeypatch, format, function_output, data_type, full_data_type, return_params
 ):
     monkeypatch.setenv(USE_SERVERLESS, str(use_serverless))
     client = get_client()
     mock_function_info = generate_function_info()
+    mock_function_info.data_type = data_type
```
BenWilson2, line 248: similar comment as above with the mock object creation

ai/integrations/autogen/tests/test_autogen_toolkit_oss.py
```diff
@@ -225,9 +231,25 @@ async def test_uc_function_to_autogen_tool(uc_client):
...
+        ),
+    ],
+)
 @pytest.mark.asyncio
-async def test_autogen_tool_with_tracing_as_retriever(uc_client, format: str, function_output: str):
+async def test_autogen_tool_with_tracing_as_retriever(
+    uc_client, format, function_output, data_type, full_data_type, return_params
+):
     mock_function_info = generate_function_info()
+    mock_function_info.data_type = data_type
```
BenWilson2, line 249: same here (and in all other integrations)

ai/integrations/openai/tests/test_openai_toolkit_oss.py
```diff
@@ -111,55 +111,6 @@ async def test_tool_calling(uc_client):
             )
 
 
-@pytest.mark.asyncio
```
BenWilson2, line 114: Why are we getting rid of this test?

BenWilson2 (commented):
Can support be added to Gemini (this integration merged yesterday)?

ai/core/tests/core/test_validation_utils.py
annzhang-db: Sorry yeah, this is an old test from when we wanted to ensure that we didn't have any columns that weren't specified in `mlflow.entities.Document` format. Thanks for catching this!

ai/integrations/openai/tests/test_openai_toolkit_oss.py
annzhang-db: This test relies on having the create function flow, which is not supported for functions with type `TABLE_TYPE` as far as I know. Please let me know if there is some way we can test this with the creation of a function that returns a `TABLE_TYPE`!

smurching: Without digging deeply - is it possible to test with a mock of the function result, similar to what we're doing with langchain?

annzhang-db: From separate discussion:
- Since `TABLE_TYPE` is not supported yet, we will remove this test since we have coverage for OpenAI in other tests and test the full `create_function` -> `execute_function` flow after we add support for `TABLE_TYPE` in OSS.

BenWilson2 (approved):
Looks great Ann! Thanks :D

PR #885, opened 2025-02-03
Add support for wrapping functions within the main callable function when creating a UC python function

BenWilson2:
**PR Checklist**

**Description of changes**

Add support for wrapping a list of callables into the main body of the primary function interface to support code modularity and templating.


BenWilson2:
Integration test run: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/13167190915

ai/core/src/unitycatalog/ai/core/utils/callable_utils.py
```diff
@@ -509,8 +548,80 @@ def generate_sql_function_body(
...
+        metadata.sql_return_type,
+        sql_metadata.func_comment,
+        sql_metadata.indented_body,
+        replace,
+    )
+
+    return sql_body
+
+
+def generate_wrapped_sql_function_body(
```
serena-ruan, line 567: Do we need this new function? we can make functions optional in generate_sql_function_body

BenWilson2: We don't explicitly, but my only concern was for some of the other feature request additions that the eval team will likely have in the next few weeks. Since everything they're doing for templating will require modifications specifically with function wrapping, I wanted to keep that edge logic external to the base standard path of building the function body statement. I'm simply worried about introducing too much complexity into a single utility function :)

PR #911, opened 2025-02-19
Implement callable recreation for logged python functions

BenWilson2:
**PR Checklist**

**Description of changes**

Implements the ability to recreate the original (minus the docstring) callable definition from a registered UC python function on Databricks. (This feature isn't needed in standard UC due to the existence of a helper function in the client module that already does this for local callable execution).

### DIFF
--- a/ai/core/src/unitycatalog/ai/core/databricks.py
+++ b/ai/core/src/unitycatalog/ai/core/databricks.py
@@ -20,8 +20,10 @@ from unitycatalog.ai.core.envs.databricks_env_vars import (
 from unitycatalog.ai.core.paged_list import PagedList
 from unitycatalog.ai.core.types import Variant
 from unitycatalog.ai.core.utils.callable_utils import (
+    extract_collection_types,
     generate_sql_function_body,
     generate_wrapped_sql_function_body,
+    parse_full_sql_data_type_for_return_type,
 )
 from unitycatalog.ai.core.utils.function_processing_utils import process_function_parameter_defaults
 from unitycatalog.ai.core.utils.type_utils import (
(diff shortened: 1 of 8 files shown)


BenWilson2:
e2e integration suite run (with new added validation tests) https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/13423509978

BenWilson2:
New run of integration tests after docstring adds: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/13423509978 (all green)

ai/core/src/unitycatalog/ai/core/databricks.py
```diff
@@ -742,6 +745,26 @@ def from_dict(cls, config: Dict[str, Any]):
         accept_keys = ["profile"]
         return cls(**{k: v for k, v in config.items() if k in accept_keys})
 
+    def get_python_callable(self, function_name: str) -> str:
```
serena-ruan, line 748: The name is a bit confusing since I would expect a Callable, but it's actually a string 🤔

ai/core/src/unitycatalog/ai/core/databricks.py
```diff
@@ -825,3 +848,38 @@ def get_execute_function_sql_command(
...
+            param_names.append(extract_collection_types(param))
+    return_type = parse_full_sql_data_type_for_return_type(function_info.full_data_type)
+    function_head = f"{function_info.name}({', '.join(param_names)}) -> {return_type}"
+    func_def = f"def {function_head}:\n"
+
+    docstring = reconstruct_docstring(function_info)
+    if docstring:
+        func_def += docstring
+
+    if function_info.routine_body.value == "EXTERNAL":
```
serena-ruan, line 877: Let's move this check up so we fail faster?

ai/core/src/unitycatalog/ai/core/databricks.py
BenWilson2: 100% agreed. I think `get_function_source` is a bit more expository. I've updated in #912 and I'm going to close this PR since the stacked nature of this is now out of sync and would be harder to fully review in a fragmented state.

ai/core/src/unitycatalog/ai/core/databricks.py
BenWilson2: Added in #912 :)

PR #924, opened 2025-02-28
Added an extra step to build the package if not already built

dmatrix:
**PR Checklist**

- [ x] A description of the changes is added to the description of this PR.
- [x ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

A minor change in the quickstart guide to nudge the user to make sure to build the package if not already built. This will ensure that all the jars ara available in the target directory, and the `uc cli` will not fail with classNotFound errors.


dennyglee (commented):
QQ @dmatrix  - do we need to update the `h2db.mv.db` in this case?

dmatrix:
@dennyglee Not sure what is the purpose of this `etc/db/h2db.mv.db`? It does get generated, and seems to have been modified. Do I need to commit and push it?

dennyglee:
> @dennyglee Not sure what is the purpose of this `etc/db/h2db.mv.db`? It does get generated, and seems to have been modified. Do I need to commit and push it?

It is the H2 metadata database and does not need to be committed.  Could you please remove that specific commit?  Thanks!

PR #925, opened 2025-02-28
Local exec

BenWilson2:
**PR Checklist**

**Description of changes**

Adds a local sandbox mode for execution that can protect the running REPL context (and underlying VM) from computationally complex function executions and limits the available libraries that can be utilized within the function execution to provide a modicum of security as compared to raw execution of a callable in the main process.


BenWilson2:
TODO: Docs, examples, manual tests, integration test validation (will tackle these on Monday)

BenWilson2:
e2e test run: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/13638238964

BenWilson2:
Some docs adjustments are needed to reflect the changes in moving to subprocess execution using a callable def rather than needing to go down the cloudpickle rabbithole with namespace mappings of crafted modules that subprocesses and multiple processes can 'see'. I'll file the docs updates tomorrow morning.

BenWilson2:
Additional e2e run after branch merge and patch changes: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/13638238964

ai/integrations/anthropic/tests/test_anthropic_utils.py
```diff
@@ -266,6 +266,7 @@ def test_generate_tool_call_messages_with_invalid_tool_use_block(mock_client, du
         )
 
 
+@pytest.mark.skipif(not TEST_IN_DATABRICKS, reason="Not running in Databricks environment")
```
BenWilson2, line 269:
```suggestion
```

ai/integrations/anthropic/tests/test_anthropic_utils.py
```diff
@@ -282,9 +283,15 @@ def test_generate_tool_call_messages_with_invalid_tool_use_block(mock_client, du
...
-        return_value=mock.Mock(),
+    with (
+        mock.patch(
+            "unitycatalog.ai.core.databricks.get_default_databricks_workspace_client",
+            return_value=mock.Mock(),
+        ),
+        mock.patch(
+            "unitycatalog.ai.core.databricks._validate_databricks_connect_available",
+            return_value=True,
+        ),
```
BenWilson2, line 294:
```suggestion
    with (
        mock.patch(
            "unitycatalog.ai.core.databricks.get_default_databricks_workspace_client",
            return_value=mock.Mock(),
        ),
        mock.patch(
            "unitycatalog.ai.core.databricks._validate_databricks_connect_available",
            return_value=True,
        ),
        mock.patch(
            "unitycatalog.ai.core.databricks.DatabricksFunctionClient.initialize_spark_session",
            return_value=None,
        ),
```

ai/core/src/unitycatalog/ai/core/envs/base.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def get(self) -> any:
+        raw_val = self._get_raw()
+        if raw_val is not None:
+            try:
+                if self.type is list:
+                    try:
+                        return json.loads(raw_val)
+                    except Exception:
+                        return [x.strip() for x in raw_val.split(",")]
```
serena-ruan, line 23: This seems a bit weird, if we have to support list[str] can we only support list[str] and not list of any other types?

ai/core/src/unitycatalog/ai/core/executor/local.py
```diff
@@ -0,0 +1,177 @@
...
+
+    Returns:
+        A tuple (success, result). If success is False, result contains a descriptive
+        error message or the error stack trace.
+    """
+    cpu_time_limit = EXECUTOR_MAX_CPU_TIME_LIMIT.get()
+    memory_limit = EXECUTOR_MAX_MEMORY_LIMIT.get()
+    timeout = EXECUTOR_TIMEOUT.get()
+
+    ctx = get_context("fork")
```
serena-ruan, line 123: Would `fork` be dangerous on macOS? ref: https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

ai/core/src/unitycatalog/ai/core/executor/common.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    Generate a message indicating that the import of a module is disallowed.
+
+    Parameters:
+        module_name: The name of the disallowed module.
+
+    Returns:
+        A string message indicating the disallowed import.
+    """
+    return f"The import of module '{module_name}' is restricted. {IMPORT_DISALLOWED_MESSAGE}"
```
serena-ruan, line 77: Seems we don't need a function here. We can define a string and use .format on module_name. Similar for generate_terminated_message

ai/core/src/unitycatalog/ai/core/client.py
```diff
@@ -832,36 +868,84 @@ def _validate_param_type(self, value: Any, param_info: FunctionParameterInfo) ->
...
+            if self.execution_mode == "local":
+                if asyncio.iscoroutinefunction(func):
+                    result = await func(**parameters)
+                else:
+                    loop = asyncio.get_running_loop()
+                    result = await loop.run_in_executor(None, partial(func, **parameters))
+            elif self.execution_mode == "sandbox":
+                succeeded, result = await run_in_sandbox_async(func, parameters)
+                if not succeeded:
+                    raise Exception(result)
```
serena-ruan, line 941: Should we return FunctionExecutionResult(error)?

ai/core/src/unitycatalog/ai/core/client.py
```diff
@@ -832,36 +868,84 @@ def _validate_param_type(self, value: Any, param_info: FunctionParameterInfo) ->
...
+            else:
+                raise RuntimeError(f"Unknown execution_mode: {self.execution_mode}")
+            if not result:
+                result = NO_OUTPUT_MESSAGE
+            return FunctionExecutionResult(format="SCALAR", value=str(result))
+        except Exception as e:
```
serena-ruan, line 947: BTW do we need this try catch?

ai/core/src/unitycatalog/ai/core/databricks.py
```diff
@@ -704,6 +731,29 @@ def _execute_uc_functions_with_serverless(
...
+        parameters = process_function_parameter_defaults(function_info, parameters)
+        python_def = get_callable_definition(function_info)
+        try:
+            succeeded, result = run_in_sandbox_subprocess(python_def, parameters)
+            if not succeeded:
+                raise Exception(
+                    f"Failed to execute function {function_info.name} with error: {result}"
+                )
+            return FunctionExecutionResult(format="SCALAR", value=result)
+        except Exception as e:
```
serena-ruan, line 754: Seems we don't need this try catch

ai/core/src/unitycatalog/ai/core/databricks.py
```diff
@@ -704,6 +731,29 @@ def _execute_uc_functions_with_serverless(
...
+            raise ValueError(
+                "Local sandbox execution is only supported for scalar Python functions. "
+                "Use 'serverless' execution mode for table functions."
+            )
+        _logger.info("Using local sandbox to execute functions.")
+
+        parameters = process_function_parameter_defaults(function_info, parameters)
+        python_def = get_callable_definition(function_info)
+        try:
+            succeeded, result = run_in_sandbox_subprocess(python_def, parameters)
```
serena-ruan, line 748: Why doesn't `run_in_sandbox` work for databricks?

ai/core/src/unitycatalog/ai/core/utils/callable_utils.py
```diff
@@ -810,7 +811,13 @@ def _parse_sql_data_type(type_str: str) -> str:
...
     if "<" not in type_str:
         mapped = SQL_TYPE_TO_PYTHON_TYPE_MAPPING.get(type_str.upper())
-        return mapped.__name__ if mapped else type_str.lower()
+        if mapped:
+            if isinstance(mapped, tuple):
+                return mapped[0].__name__
+            else:
+                return mapped.__name__
+        else:
+            return type_str.lower()
```
serena-ruan, line 820:
```suggestion
        if mapped:=SQL_TYPE_TO_PYTHON_TYPE_MAPPING.get(type_str.upper()):
            return mapped[0].__name__ if isinstance(mapped, tuple) else mapped.__name__
        else:
            return type_str.lower()
```

serena-ruan:
This PR is a bit too large 😅 It would be nice if we can split it into two PRs, one for client and one for databricks

ai/core/src/unitycatalog/ai/core/client.py
```diff
@@ -343,12 +349,23 @@ class UnitycatalogFunctionClient(BaseFunctionClient):
     Unity Catalog function client for managing and executing functions in Unity Catalog OSS.
     """
 
-    def __init__(self, api_client: ApiClient, **kwargs: Any) -> None:
+    def __init__(
+        self, api_client: ApiClient, execution_mode: str = "sandbox", **kwargs: Any
```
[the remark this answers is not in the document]
BenWilson2: Yeah this makes a lot of sense, particularly since we're going to be adding probably 2 more modes in the next 2 months

ai/core/src/unitycatalog/ai/core/executor/common.py
```diff
@@ -0,0 +1,77 @@
...
+    "\t2. You can increase the wall-clock time limit by setting the environment variable "
+    "EXECUTOR_TIMEOUT. The default value is 20 seconds.\n"
+)
+
+NO_OUTPUT_MESSAGE = (
+    "The function execution has completed, but no output was produced. "
+    "This may indicate that the function did not return a value or that an error occurred "
+    "without being captured. Check the function implementation and ensure it returns a value "
+    "by locally executing the function returned from calling the `get_function_source` method on the function "
+    "client object and converting your function to a callable using `unitycatalog.ai.core.utils.execution_utils.load_function_from_string`."
```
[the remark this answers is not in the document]
BenWilson2: Now that you mention it, this makes a LOT of sense. Why the 2 steps and having to call a util when we can just provide this in the client.... adding this API.

ai/core/src/unitycatalog/ai/core/databricks.py
```diff
@@ -704,6 +731,29 @@ def _execute_uc_functions_with_serverless(
             error = f"Failed to execute function with command `{sql_command_msg}`\nError: {e}"
             return FunctionExecutionResult(error=error)
 
+    @retry_on_session_expiration
```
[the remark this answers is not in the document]
BenWilson2: Good catch. No, we don't.

ai/core/src/unitycatalog/ai/core/envs/executor_env_vars.py
```diff
@@ -0,0 +1,30 @@
...
+)
+
+EXECUTOR_TIMEOUT = _EnvironmentVariable(
+    "EXECUTOR_TIMEOUT",
+    int,
+    20,
+    "Maximum timeout for the executor measured in seconds (wall clock time).",
+)
+
+EXECUTOR_DISALLOWED_MODULES = _EnvironmentVariable(
```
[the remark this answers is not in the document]
BenWilson2: whoops! I had split this out during the Databricks local execution work, then removed it, but forgot to update the multiprocessing implementation for OSS. Updated to use it properly.

ai/core/src/unitycatalog/ai/core/envs/base.py
BenWilson2: Yeah, let's just support casting collection types properly.

ai/core/src/unitycatalog/ai/core/client.py
BenWilson2: nope!

ai/core/src/unitycatalog/ai/core/databricks.py
BenWilson2: unified with the proper error handling for not successful and removed the try/catch. Thanks!

ai/core/src/unitycatalog/ai/core/databricks.py
BenWilson2: It's a long explanation - let's discuss offline :)

BenWilson2:
e2e run: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/13685748280

BenWilson2:
TODO: add guidance for using local mode in development mode to verify serverless functionality prior to deployment. (Add this to the warning message)

ai/core/src/unitycatalog/ai/core/utils/execution_utils.py
```diff
@@ -1,4 +1,50 @@
 import ast
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+class ExecutionMode:
```
serena-ruan, line 7: Actually I think we don't need to unify ExecutionMode for client/databricks, it can be separate Enums in its own namespace given that they have different logics. It would be simpler if we have 
```
class ExecutionModel(str, Enum):
    SERVERLESS = "serverless"
    LOCAL = "local"

    def __str__(self):
        return self.value

    @classmethod
    def validate(cls, value):
        return value in cls.__member__.values() ....
```

ai/core/src/unitycatalog/ai/core/executor/local_subprocess.py
```diff
@@ -0,0 +1,154 @@
...
+    except SyntaxError as e:
+        raise ValueError("Invalid Python source code") from e
+
+    for node in module_ast.body:
+        if isinstance(node, ast.FunctionDef):
+            return node.name
+    raise ValueError("No function definition found in the provided source code.")
+
+
+def _generate_runner_script(function_source: str, cpu_time_limit: int, memory_limit: int) -> str:
```
serena-ruan, line 44: Instead of this long string, I think we can create a separate python file for this, and pass those parameters as args. But this can be a follow-up :)

ai/core/src/unitycatalog/ai/core/executor/local_subprocess.py
```diff
@@ -0,0 +1,154 @@
...
+
+    The parameters are serialized via cloudpickle and passed to a temporary runner
+    script that applies resource limits and executes the function.
+    The runner returns a JSON message indicating success or error.
+
+    Returns:
+        (success, result): If success is True, result is the function's return value (or a default message if None).
+        Otherwise, result contains the error message.
+    """
+    timeout = float(EXECUTOR_TIMEOUT.get())
```
serena-ruan, line 105: Is float conversion needed here?

ai/core/src/unitycatalog/ai/core/executor/local_subprocess.py
```diff
@@ -0,0 +1,154 @@
...
+    script = _generate_runner_script(
+        function_source=function_source, cpu_time_limit=cpu_time_limit, memory_limit=memory_limit
+    )
+
+    with tempfile.NamedTemporaryFile("w", delete=False, suffix=".py") as f:
+        f.write(script)
+        script_path = f.name
+
+    env = os.environ.copy()
+    env["PYTHONPATH"] = os.pathsep.join(sys.path)
```
serena-ruan, line 122: What if PYTHONPATH already exists? Wondering if this works well on serverless cluster 🤔

ai/core/src/unitycatalog/ai/core/executor/local_subprocess.py
```diff
@@ -0,0 +1,154 @@
...
+            capture_output=True,
+            text=True,
+            timeout=timeout,
+            env=env,
+        )
+    except subprocess.TimeoutExpired:
+        return False, TIMEOUT_ERROR_MESSAGE
+    finally:
+        try:
+            os.remove(script_path)
```
serena-ruan, line 136: Why don't we move this part inside `with tempfile...` so it's auto removed?

ai/core/src/unitycatalog/ai/core/executor/local_subprocess.py
```diff
@@ -0,0 +1,154 @@
...
+    _limit_resources()
+    try:
+        b64_params = sys.argv[1]
+        params = __import__('cloudpickle').loads(__import__('base64').b64decode(b64_params))
+        result = {function_name}(**params)
+        if asyncio.iscoroutine(result):
+            result = asyncio.run(result)
+        if result is None:
+            result = "{NO_OUTPUT_MESSAGE}"
+        print(__import__('json').dumps({{"success": True, "result": result}}))
```
serena-ruan, line 77: Why not `json.dumps` since you've already imported json? Similar for traceback below

serena-ruan (approved):
Overall LGTM! Left some nits :)

ai/core/src/unitycatalog/ai/core/utils/execution_utils.py
BenWilson2: Yeah this is definitely cleaner. I'll sepearte these out and add the warning notification to the Databricks side only :)

ai/core/src/unitycatalog/ai/core/executor/local_subprocess.py
BenWilson2: I originally had this implementation, but in tempfile, when marking a temp file as deleted, a file lock is placed on it (as it's marked for deletion after the context manager exits) for subprocesses that attempt to access the file, rendering the subprocess incapable of executing the script from that file location.

ai/core/src/unitycatalog/ai/core/executor/local_subprocess.py
BenWilson2: Yeah that's a good point. They're all available directly in the subprocess environment. I'll simplify that script :D

ai/core/src/unitycatalog/ai/core/executor/local_subprocess.py
BenWilson2: It does - I tested it out yesterday :) Without it, we don't inherit some dependent packages properly.

ai/core/src/unitycatalog/ai/core/executor/local_subprocess.py
BenWilson2: nope! forgot to update that one after supporting non-string types in the env manager

BenWilson2:
final e2e run (all green): https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/13685748280

PR #927, opened 2025-03-04
silence langchain warning

BenWilson2:
**PR Checklist**

**Description of changes**

Silences a known noisy warning when using the LangChain toolkit integration.

### DIFF
--- a/ai/integrations/langchain/src/unitycatalog/ai/langchain/toolkit.py
+++ b/ai/integrations/langchain/src/unitycatalog/ai/langchain/toolkit.py
@@ -1,5 +1,10 @@
 import json
+import warnings
 from typing import Any, Dict, List, Optional
+
+from langchain_core._api.deprecation import LangChainDeprecationWarning
+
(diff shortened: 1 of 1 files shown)


serena-ruan:
What kind of warnings are we silencing here? Is it pydantic v1 deprecation warning?

BenWilson2:
> What kind of warnings are we silencing here? Is it pydantic v1 deprecation warning?

Exactly. Agent squad and some PrPr customers were bringing up the fact that this is very annoying to see with every request.

serena-ruan:
> > What kind of warnings are we silencing here? Is it pydantic v1 deprecation warning?
> 
> Exactly. Agent squad and some PrPr customers were bringing up the fact that this is very annoying to see with every request.

Could we only silent that pydantic warning then, I'm a bit worried if StructuredTool is deprecated in the future and we never capture it 🤣

serena-ruan (approved):
LGTM!

PR #936, opened 2025-03-09
fix: only login to docker hub on tag push

tnightengale:
**PR Checklist**

**Description of changes**

Fixes the `docker-build.yml` GitHub workflow to only login to DockerHub on `push` with tags, ie. a release by a maintainer. Previously was failing because repository secrets are not available in PRs from forks. Also uses Github Actions cache for image builds, because DockerHub would require login.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -33,6 +33,7 @@ jobs:
           images: ${{ env.IMAGE }}
 
       - name: Login to Docker Hub
+        if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') }}
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -52,6 +52,6 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
 
-          # Use registry caching: https://tinyurl.com/4b8mfexr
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build.yml
```diff
@@ -51,6 +52,6 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
 
-          # Use registry caching: https://tinyurl.com/4b8mfexr
-          cache-from: type=registry,ref=${{ env.IMAGE }}:buildcache
-          cache-to: type=registry,ref=${{ env.IMAGE }}:buildcache,mode=max
+          # Use github actions cache: https://docs.docker.com/build/cache/backends/gha/
```
tnightengale, line 55: Rely on Github Actions cache, instead of DockerHub, which will not be available for caching without secrets.

.github/workflows/docker-build.yml
```diff
@@ -33,6 +33,7 @@ jobs:
           images: ${{ env.IMAGE }}
 
       - name: Login to Docker Hub
+        if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') }}
```
tnightengale, line 36: Only login with secrets on pushes with tags. Same logic as in line 50.

dennyglee: Got it - thanks for the fix @tnightengale !

.github/workflows/docker-build.yml
dennyglee: Gotcha!

dennyglee (approved):
Thanks for tackling this @tnightengale !

PR #940, opened 2025-03-12
Set default client in get_uc_function_client

serena-ruan:
**PR Checklist**

**Description of changes**
We don't need a warning for get_uc_function_client when it's none. It's also confusing when it's used in validate_or_set_default_client.
This PR also moves the logic to set default client in get_uc_function_client, so users don't need to create a UCFunctionToolkit in order to trigger default client setup.

### DIFF
--- a/ai/core/src/unitycatalog/ai/core/base.py
+++ b/ai/core/src/unitycatalog/ai/core/base.py
@@ -253,6 +253,22 @@ class BaseFunctionClient(ABC):
 def get_uc_function_client() -> Optional[BaseFunctionClient]:
     global _uc_function_client
 
+    if client is None and _is_databricks_client_available():
+        try:
+            from unitycatalog.ai.core.databricks import DatabricksFunctionClient
+
+            client = DatabricksFunctionClient()
+
+        except Exception as e:
+            _logger.warning(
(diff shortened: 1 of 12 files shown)


ai/core/src/unitycatalog/ai/core/base.py
```diff
@@ -254,8 +254,6 @@ def get_uc_function_client() -> Optional[BaseFunctionClient]:
     global _uc_function_client
 
     with _client_lock:
-        if _uc_function_client is None:
-            _logger.warning("UC function client is not set.")
```
bbqiu, line 258: shouldn't we only suppress this warning in the case when we call from `validate_or_set_default_client` and we do end up setting a default client? otherwise users may think they're getting back a real client when it's actually returning None

https://github.com/unitycatalog/unitycatalog/blob/85877db477c3153a6f40df4cddd43d11bf92cbd9/ai/core/src/unitycatalog/ai/core/utils/client_utils.py#L31

serena-ruan: Well validate_or_set_default_client is the only place this function is called. If they manually call get_uc_function_client they'll see None returned :) And if client is None, validate_or_set_default_client raises ValueError

BenWilson2 (approved):
Much better location :D

PR #944, opened 2025-03-16
minor: reformatting + remove @param for a non-existent param

jaceklaskowski:
Minor changes:

1. (Re)formated a multiple-trait class definition and some other code of `UCSingleCatalog`
2. Removed `@param` tag for a non-existent param of `AuthService`

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,12 +1,23 @@
-name: Build and Push Docker Images
+name: Build and Push Docker Images - unitycatalog
 
 on:
   push:
     tags:
       - 'v*'
+    branches:
+      - main
+    paths:
+      - build/**
+      - project/**
+      - examples/**
+      - server/**
+      - api/**
+      - clients/python/**
+      - version.sbt
+      - build.sbt
 
 env:
-  IMAGE: ${{ github.repository }} # unitycatalog/unitycatalog
+  IMAGE: unitycatalog/unitycatalog
 
 jobs:
   build-and-push:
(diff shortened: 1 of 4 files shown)


jaceklaskowski:
@dennyglee @BenWilson2 @ognis1205 Can you have a look at the PR, please? 🙏  I'm very surprised that it's not been fixed yet as the following command fails just at this erroneous `@param` in `AuthService`:

```bash
./build/sbt '++2.13' clean package publishLocal
```

It's only when this `@param` is gone that it's possible to publish local(ly).

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -90,7 +90,6 @@ public AuthService(
    * the incoming identity (email, subject) matches a specific user in the system, once a user
    * management system is in place.
    *
-   * @param request The token exchange parameters
```
ognis1205, line 93: Can we add the following parameters to the API documentation?

```java
 * @param ext Specifies whether the issued token should be set as a cookie.
 * @param form The OAuth 2.0 token exchange request form.
```

ognis1205 (changes requested):
It was my mistake for forgetting to update it. Thanks for the correction. Good catch!

ognis1205 (approved):
LGTM!

jaceklaskowski:
@dennyglee you're needed here 🙏

dennyglee:
Thanks @jaceklaskowski  - yes, this looks great to me!  @tdas - can you take a look as well?  Thanks!

dennyglee (approved):

PR #952, opened 2025-03-27
Unity Catalog AI client handling of SQL NULL default parameters

BenWilson2:
**PR Checklist**

- [X] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

Currently, functions that have parameter defaults with SQL "NULL" as the default value fail validation checks with a very difficult to troubleshoot error message from pydantic validators. 
This PR introduces the fix for SQL Null handling and establishes an error handler to make it easier to understand why a Validator error happens. 
However, the larger-scale wrapping of validator checks elsewhere in the code base would make a full replacement be many thousands of lines long, so will leave usage of the decorator and the function handler to later PRs.

PR #953, opened 2025-03-28
Update the match coverage for serverless endpoint unavailability for DatabricksFunctionClient

BenWilson2:
**PR Checklist**

**Description of changes**

Update the match condition logic and expand test coverage for matching conditions due to changing error message syntax coming from the serverless RPC endpoints.


BenWilson2:
e2e run: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/14118995165 
Some tests failures are addressed in #952

BenWilson2:
Confirmed the only test failure in e2e tests is test_execute_function_with_custom_client which had a bug in its implementation, fixed in #952

serena-ruan (approved):
LGTM! This is a bit tough lol, maybe we should create an error class (instead of just string value) in the future if we need to do this again

BenWilson2:
I have a feeling that we'll need to that. This is the 3rd time these strings have been updated in the backend error reporting logic and due to the RPC generation of these errors, we can't do validation of the failure type by using instance verification. 
I'll file a follow-up ticket so that we can explore using the error message type name and ensure that there isn't any abstracted reuse of those keys in non retryable error types

PR #955, opened 2025-03-28
Increase the minimum supported version for databricks-connect for Python 3.10 compatibility issues

BenWilson2:
**PR Checklist**

**Description of changes**

Increases the minimum supported version of databricks-connect for the DatabricksFunctionClient to support serializability in Agent usage due to numpy version inclusion dropping support for Python 3.9.

### DIFF
--- a/ai/core/src/unitycatalog/ai/core/databricks.py
+++ b/ai/core/src/unitycatalog/ai/core/databricks.py
@@ -47,7 +47,7 @@ if TYPE_CHECKING:
     )
     from databricks.sdk.service.sql import StatementState
 
-DATABRICKS_CONNECT_SUPPORTED_VERSION = "15.1.0"
+DATABRICKS_CONNECT_SUPPORTED_VERSION = "16.1.1"
 DATABRICKS_CONNECT_IMPORT_ERROR_MESSAGE = (
     "Could not import databricks-connect python package. "
     "To interact with UC functions using serverless compute, install the package with "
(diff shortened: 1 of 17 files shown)


ai/core/src/unitycatalog/ai/core/databricks.py
```diff
@@ -47,7 +47,7 @@
     )
     from databricks.sdk.service.sql import StatementState
 
-DATABRICKS_CONNECT_SUPPORTED_VERSION = "15.1.0"
+DATABRICKS_CONNECT_SUPPORTED_VERSION = "16.1.1"
```
serena-ruan, line 50: Is this required?

BenWilson2: The warning message or the pinning?

serena-ruan: The pinning, doesn't 15.1.0 work?

BenWilson2: Good point. I think there's just a compatibility issue with certain runtimes, but that won't affect much as those are pre-installed with the given runtime. I'll revert this for the recommendation so as to not confuse users :)

BenWilson2, line 50:
```suggestion
DATABRICKS_CONNECT_SUPPORTED_VERSION = "15.1.0"
```

PR #958, opened 2025-03-31
Remove the string wrapping from the function clients return

BenWilson2:
**PR Checklist**

**Description of changes**

Removes the string wrapping that we've been doing for the function clients to return the native types.

### DIFF
--- a/ai/core/src/unitycatalog/ai/core/base.py
+++ b/ai/core/src/unitycatalog/ai/core/base.py
@@ -22,12 +22,11 @@ _client_lock = threading.Lock()
 class FunctionExecutionResult:
     """
     Result of executing a function.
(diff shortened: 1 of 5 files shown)


BenWilson2:
e2e test: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/14179919623 (same two core tests are failing and are fixed in #952 )

serena-ruan (approved):
LGTM! Could we file a JIRA to support pandas DataFrame as output value for TABLE functions?

BenWilson2:
> LGTM! Could we file a JIRA to support pandas DataFrame as output value for TABLE functions?

Great idea :) added!

PR #974, opened 2025-04-16
Clean up code for transaction management and auth

vikrantpuppala:
**PR Checklist**

**Description of changes**

Clean up code for transaction management and auth


server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
```diff
@@ -141,112 +127,119 @@ public CatalogInfo updateCatalog(String name, UpdateCatalog updateCatalog) {
...
+                      Optional.empty());
+
+          if (!initialSchemaCheck.getSchemas().isEmpty() && !force) {
+            throw new BaseException(
+                ErrorCode.FAILED_PRECONDITION,
+                "Cannot delete catalog with schemas. Use force=true to force deletion.");
+          }
+
+          // If we're proceeding with force=true, delete all schemas with pagination
+          if (force) {
```
tdas, line 203: why this change in logic? 
and is this logic change covered in unit test?

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
```diff
@@ -29,6 +29,9 @@ public static <T> T attachProperties(
     try {
       List<PropertyDAO> propertyDAOList =
           PropertyRepository.findProperties(session, UUID.fromString(uuid), entityType);
+      if (propertyDAOList.isEmpty()) {
```
tdas, line 32: how is this logic change tested?

tdas (approved):
LGTM

server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
vikrantpuppala: this is only a slight change in how the if conditions are written, the logic is still the same. I don't think we have tests for force specifically.

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
vikrantpuppala: I updated a test case to test this

Issue #1004, opened 2025-05-16
Support DateType


abij:
**Is your feature request related to a problem?**

Related issue: #891: Error when using type "Date".

``` 
-- Spark SQL example:

create external table unity.default.test1 (date_col date) using parquet location "/somepath/"
```

Result is error:

```
io.unitycatalog.client.ApiException: DataType not supported: date
	at io.unitycatalog.spark.UCProxy.convertDataTypeToTypeName(UCSingleCatalog.scala:365)
	at io.unitycatalog.spark.UCProxy.$anonfun$createTable$5(UCSingleCatalog.scala:327)
```

Which happens because of this mapping: 
https://github.com/unitycatalog/unitycatalog/blob/912959e79ba921f9627153c57353008218b4d0fe/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala#L359

**Describe the solution you would like**

Support the `DateType` in `io.unitycatalog.client.model.ColumnTypeName` and update the mapping.

**Additional context**

I think the default [Spark DataTypes](https://spark.apache.org/docs/latest/sql-ref-datatypes.html) should be supported.

abij:
Merged in main (#1382), not yet part of 0.4.0.

PR #1005, opened 2025-05-16
[SPARK] Support DateType in the spark connector

abij:
Improve compatibility by supporting type `Date` in the spark-connector, which allows the following:

```
-- Failed before:
create table unity.default.test1 (col1 date) using parquet location "/somelocation"
```
Issue: 
#1004 

**PR Checklist**

**Description of changes**

Add missing type `date` to the converter for Spark.

I was unable to run the test locally, I hope the build-server check the unit-tests.


abij:
@openinx or @yili-db can you have a look, since this is a small improvement and you have been working on this codebase? It's just a forgotten mapping of the date-type.

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaExternalTableReadWriteTest.java
```diff
@@ -35,6 +36,24 @@ public void testDeltaPathTable() throws IOException {
...
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG);
+
+    TableSetupOptions options = new TableSetupOptions()
+            .setCatalogName(SPARK_CATALOG)
+            .setTableName(TEST_TABLE);
+
+    sql(String.format(
+            "CREATE TABLE %s (start_date DATE) USING delta LOCATION '%s'",
+            options.fullTableName(), getLocation(options)));
+    sql(String.format("INSERT INTO %s VALUES ('2025-12-22')", options.fullTableName()));
```
yili-db: Might as well SELECT from this table to verify.

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaExternalTableReadWriteTest.java
```diff
@@ -35,6 +36,24 @@ public void testDeltaPathTable() throws IOException {
...
 
+  @Test
+  public void testExternalDeltaTableWithDateType() throws IOException {
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG);
+
+    TableSetupOptions options = new TableSetupOptions()
+            .setCatalogName(SPARK_CATALOG)
+            .setTableName(TEST_TABLE);
+
+    sql(String.format(
```
yili-db: You can remove `String.format`. `sql()` does it automatically.

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaExternalTableReadWriteTest.java
```diff
@@ -35,6 +36,24 @@ public void testDeltaPathTable() throws IOException {
     }
   }
 
+  @Test
```
yili-db, line 41: Looks like `SparkDeltaTableCRUDTest` is a better place. You can add an extra column of `Date` in the existing test table there.

abij: Updated a test in integration-test with the date type, and one unit-test in module spark-connector checking the date-type functionality.

tdas: @yili-db shouldnt we be also unit tests for managed tables? 
given that the tests for managed and external tables are not unified yet.. what is the best place to add the test for date for managed tables?

yili-db: @tdas There is `connectors/spark/src/test/java/io/unitycatalog/spark/DeltaManagedTableReadWriteTest.java` for managed tables.
For data type there shouldn't be any difference between external and managed.

tdas: @abij could update that test?

tdas:
hi @abij will you be able to update this PR based on the comments?

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaExternalTableReadWriteTest.java
```diff
@@ -35,6 +38,27 @@ public void testDeltaPathTable() throws IOException {
...
+  public void testExternalDeltaTableWithDateType() {
```
arturowczarek, line 42: I think this test is not correct. It is verifying the schema from Delta table instead of Unity Catalog. The schema is not pushed to Unity Catalog by default. This test gives the illusion of schema properly handled but we should be testing Unity Catalog, not Delta. If you go to Unity Catalog after creating the table you'll see the schema metadata is empty. See https://github.com/unitycatalog/unitycatalog/pull/1325/changes#diff-9162c2f80101299bd2b1f382257b60fd82b24eab0e819dced3013ed674ad6684R36 for the details.

tdas: you are absolutely right!

this needs to be fixed. 

since @abij is not able to respond, @arturowczarek would you mind taking a crack at this?

yili-db: @tdas 
Schema metadata is intentionally left blank on UC. Because currently Delta doesn't have the full implementation to **update** metadata on UC after the table is created. So it's better to leave it blank than stale.

arturowczarek: @tdas Sure - I already proposed an alternative approach in **#1325** 

Mechanically it’s the same conversion that @abij suggested; the main difference is **how the behavior is tested**. I think the test should **assert results via `UCSingleCatalog`**, otherwise Delta’s catalog logic can take over and effectively bypass what Unity Catalog returns.

For writes, we can still use `CREATE TABLE` as long as `spark.databricks.delta.catalog.update.enabled` is enabled.

One complication: UC’s metadata/type handling seems limited. In particular, **`VARCHAR` and `CHAR` are tricky** because UC appears to return them as `STRING`. I’m not sure what the intended behavior should be here — should we treat these as unsupported and route them through the `DataType not supported` path (at least until UC returns richer type info)? I also noticed other types (e.g. intervals) that don’t seem to be supported.

@yili-db Could you share more detail on the limitations you’re aware of (and whether they’re tied to specific type families)? In my experiments, once `spark.databricks.delta.catalog.update.enabled` is set, the flow seems to work - but if the limitations are significant, we may want to tighten the test further to validate strictly through `UCSingleCatalog` (and avoid relying on `CREATE TABLE` semantics beyond the minimum).

yili-db: @arturowczarek once `spark.databricks.delta.catalog.update.enabled` is set, the metadata is sent to UC when creating a table. But if later the metadata is changed, it isn't updated to UC.

arturowczarek: Thank you for clarification, @yili-db. Since we're testing the Unity Catalog Spark plugin here, we can rely on Delta only to the point where it is helpful. It is helpful with creating the table (it keeps the test readable and mirrors real usage). It is not helpful with accessing the Unity Catalog schema (so we should take a different approach - like using the `UCSingleCatalog` directly). And this is what I did in #1325.

arturowczarek: @tdas Any chance you can have a look at https://github.com/unitycatalog/unitycatalog/pull/1325?

abij: @arturowczarek thanks for taking the effort to have a look and make a complete implementation with the other dataTypes. For me it was hard to setup/investigate the changes I made, the sbt-test are running long and create crazy amount of output. Unable to isolate the test properly / debug in the editor. I did test the self-build jar in a project to verify that it would actually work.

I'll close this PR, since it's now in 0.4.0.

abij:
Resolved in #1382, but not yet part of release 0.4.0.

PR #1007, opened 2025-05-17
build: Move dep versions to Versions scala object

jaceklaskowski:
Introduce `project/Versions.scala` scala object with the dependencies shared across sbt modules

1. Moved the versions to `Versions` build object
2. Removed `byte-buddy` dependency (as not needed, yet caused some Spark 4 integration issues)
3. Fixed a few typos


jaceklaskowski:
@johanl-db Please review and approve 🙏

project/Versions.scala
```diff
@@ -0,0 +1,13 @@
...
+  val deltaVersion = "3.2.1"
+  val jacksonVersion = "2.15.0"
+  val log4jVersion = "2.24.3"
+  val openApiToolsJacksonBindNullableVersion = "0.2.6"
+  val orgApacheHttpVersion = "4.5.14"
+  val scala212 = "2.12.15"
+  val scala213 = "2.13.14"
+  val slf4jVersion = "2.0.13"
+  val spark3Version = "3.5.3"
+  val spark4Version = "4.0.0"
```
johanl-db, line 12: nit: can you split the versions into separate groups:
- scala versions
- delta/spark
- other dependencies

project/Versions.scala
```diff
@@ -0,0 +1,13 @@
+object Versions {
+  val antlrVersion = "4.9.3"
+  val deltaVersion = "3.2.1"
+  val jacksonVersion = "2.15.0"
```
johanl-db, line 4: There were two jackson versions used before: 2.15.0 and 2.17.0. This effectively downgrades the version for part of the targets, we'll need to define two separate jackson versions. I'd let unifying versions used for a later change, this may create more issues than it solves

project/Versions.scala
johanl-db, line 12: spark4Version isn't used, it's probably simpler to have a single `sparkVersion` since we will only support a single spark version at least for now

jaceklaskowski: Ah, right. It's for the upcoming PRs! 😉

PR #1040, opened 2025-06-25
Update deployment.yaml to reference config value

bryanterichardson:
**PR Checklist**

**Description of changes**

References the configured value for the server UI instead of hard coded default. The value in the `values.yaml` seems to be ignored without this change. There might be other values to configure as well.

### DIFF
--- a/helm/templates/ui/deployment.yaml
+++ b/helm/templates/ui/deployment.yaml
@@ -37,9 +37,9 @@ spec:
             - /bin/sh
             - -c
             - |-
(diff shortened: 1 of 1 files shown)


bryanterichardson:
I did confirm that this fixed my GCP setup that required providing a value like `my-service.my-namespace.svc.cluster.local`. Helm chart would not start otherwise. Looks like the same issue exists in the create user deployment but I can help with that in a separate PR if this one is okay.

bryanterichardson:
Ehh theres a better way to do this. Will open a diff PR

PR #1043, opened 2025-06-26
GCP Fixes: Support service annotations and increase startup probe limit

bryanterichardson:
**PR Checklist**

**Description of changes**

In GCP, service annotations are required to key certain behaviors. This change adds support for that. This also increases the time limit for the startup probe as my app was taking just over five minutes and this was causing a crash loop


helm/templates/server/service.yaml
```diff
@@ -5,8 +5,11 @@ metadata:
   name: {{ include "unitycatalog.server.fullname" . | quote }}
   labels:
     {{- include "unitycatalog.server.labels" . | nindent 4 }}
+  {{- with .Values.server.service.annotations }}
+  annotations: {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
-  type: {{ .Values.server.service.type | quote }}
+  type: {{ .Values.server.service.type }}
```
bryanterichardson, line 12: I only did this to match what is happening with the UI. Inversely, I could update the UI code to also apply quotes

helm/templates/ui/deployment.yaml
```diff
@@ -37,11 +37,15 @@ spec:
             - /bin/sh
             - -c
             - |-
-              until nc -z {{ include "unitycatalog.server.fullname" . }} {{ .Values.server.service.port }}
+              SERVER_HOST="{{ include "unitycatalog.server.fullname" . }}"
+              SERVER_PORT="{{ .Values.server.service.port }}"
```
bryanterichardson, line 41: Substitution here didnt seem to be working. The seems to fix it.

PR #1064, opened 2025-08-28
Add ability to do "client credentials" OAuth2 flow to Spark Connector

creechy:
**PR Checklist**

**Description of changes**

This change adds the ability to configure the UC Spark Connector to do an OAuth2 "client credentials" exchange to authenticate.

Since access tokens minted through a client credentials exchange generally have a limited lifetime, there is functionality here to refresh the token before the current access token expires.

This enhancment introduces two new Spark properties
- `oauth2-server-uri` - this is the URI to the token exchange endpoint
- `credential` - the principal's credential to use to exchange for an access token (in the form of client_id:client_secret)

This PR covers Issue https://github.com/unitycatalog/unitycatalog/issues/1071

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -328,8 +328,6 @@ private class UCProxy(
     checkUnsupportedNestedNamespace(ident.namespace())
     assert(properties.get("provider") != null)
 
-    System.err.println(s"XXX table properties: $properties")
-
     val createTable = new CreateTable()
     createTable.setName(ident.name())
     createTable.setSchemaName(ident.namespace().head)


connectors/spark/src/main/scala/io/unitycatalog/spark/OAuth2CredentialExchangeProvider.scala
```diff
@@ -0,0 +1,137 @@
...
+      )
+    }
+    (parts(0), parts(1))
+  }
+
+  private def exchangeCredentialsForAccessToken(): Long = {
+    val requestParameters =
+      Map(
+        "grant_type" -> "client_credentials",
+        "scope" -> "all-apis",
```
roeap, line 67: Just from reading the rest api docs, it seems we are exposing more and more scopes to read UC. Should we already make this configurable with a default value?

If we do want to do this, we would likely also need to make this an array to allow passing multiple scopes.

connectors/spark/src/main/scala/io/unitycatalog/spark/OAuth2CredentialExchangeProvider.scala
```diff
@@ -0,0 +1,137 @@
...
+    params
+      .map {
+        case (key, value) =>
+          s"${encode(key, UTF_8.toString)}=${encode(value, UTF_8.toString)}"
+      }
+      .mkString("&")
+  }
+
+  private def getClientIdAndSecret(credential: String): (String, String) = {
+    val parts = credential.split(":")
```
roeap, line 54: I tried to find a reference if there is a set of allowed characters for client credentials, but to no avail. Thus I believe technically the credential could also include `:`. Is there a similar method `.splitOnce` or `.splitN` o.a. we could use to just split on the first colon?

Or does it make more sense to just split the parameters since they are often stored in the environment separately and callers would need to concatenate the string just or us to split it again?

connectors/spark/src/main/scala/io/unitycatalog/spark/OAuth2CredentialExchangeProvider.scala
```diff
@@ -0,0 +1,137 @@
...
+        throw new OAuth2Exception(
+          s"Failed to obtain access token, status code: ${response.statusCode()}, body: ${response.body()}"
+        )
+    }
+  }
+
+  private def MAX_REFRESH_WINDOW_MILLIS = 300000L // 5 minutes
+  private def MIN_REFRESH_WAIT_MILLIS = 1000L // 1 second
+
+  private def scheduleRefresh(expiresAtMillis: Long): Unit = {
```
roeap, line 114: Just for me to understand. This will eagerly schedule a refresh to get the next token before the current token expires?

The pattern I have seen most of the time is to check the current token just before a request. The approach here does have the nice advantage that we don't need to do this on the hot path and may save some time for the individual request.

If I read this right, we might however make a few more requests to the oauth server (n_nodes request ~ every hour) that we may not need in long running jobs? Since this is not a large amount of request though this might not be a problem.

connectors/spark/src/main/scala/io/unitycatalog/spark/OAuth2CredentialExchangeProvider.scala
```diff
@@ -0,0 +1,137 @@
...
+    // attempt to schedule a refresh.
+    val tokenExpiresAtMillis = exchangeCredentialsForAccessToken()
+    scheduleRefresh(tokenExpiresAtMillis)
+  }
+
+  private def mapToUrlEncoded(params: Map[String, String]): String = {
+    params
+      .map {
+        case (key, value) =>
+          s"${encode(key, UTF_8.toString)}=${encode(value, UTF_8.toString)}"
```
roeap, line 48: since this is a private method and we know the keys do not require encoding, should we just encode the values?

connectors/spark/src/main/scala/io/unitycatalog/spark/OAuth2CredentialExchangeProvider.scala
```diff
@@ -0,0 +1,137 @@
...
+import java.net.URLEncoder.encode
+import java.nio.charset.StandardCharsets.UTF_8
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class OAuth2Exception(message: String, cause: Throwable) extends RuntimeException(message, cause) {
+  def this(message: String) = this(message, null)
+}
+
+class OAuth2CredentialExchangeProvider(
```
roeap, line 20: nit: reading "exchange" in the name somehow makes me think about federated / client assertion credentials where we exchange a foreign token for one issues by the server...

Would `OAuth2ClientCredentialProvider` also work, which (IMO) more closely identifies the credential flow we are using.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -31,22 +31,40 @@ class UCSingleCatalog
...
   @volatile private var delegate: TableCatalog = null
+  @volatile private var token: String = null
 
   override def initialize(name: String, options: CaseInsensitiveStringMap): Unit = {
     val urlStr = options.get("uri")
     if (urlStr == null) {
       throw new IllegalArgumentException(s"uri must be specified for Unity Catalog '$name'")
     }
+
+    val oauth2ServerUri = options.get("oauth2-server-uri")
```
roeap, line 45: We should definitely follow what other clients (e.g. iceberg) established, especially the spark integrated ones.

That said, personally I would have expected the `auth2-server-uri` not to have the `/token` endpoint included, since this is part of the spec, and should never be anything else for an oauth compliant server.

But again, in this case i would certainly bias for the same config potentially working for IRC and Unity clients.

connectors/spark/src/main/scala/io/unitycatalog/spark/OAuth2CredentialExchangeProvider.scala
roeap:
```suggestion
    val parts = credential.split(":", limit=2)
```
just had a quick look at the docs, i think this may also work?

PR #1065, opened 2025-08-29
Update Python version requirements and integration tests for LlamaIndex

puneet-jain159:
**PR Checklist**

- [X ] A description of the changes is added to the description of this PR.

**Description of changes**
Update Python version requirements and integration tests for LlamaIndex and correct OpenAI dependencies to resolve test failures

### DIFF
--- a/ai/integrations/openai/tests/helper_functions.py
+++ b/ai/integrations/openai/tests/helper_functions.py
@@ -5,8 +5,7 @@ from openai.types.chat.chat_completion import (
     ChatCompletionMessage,
     Choice,
 )
-from openai.types.chat.chat_completion_message import ChatCompletionMessageToolCall
-from openai.types.chat.chat_completion_message_tool_call import Function
+from openai.types.chat.chat_completion_message_tool_call import ChatCompletionMessageToolCall,Function
 from openai.types.completion_usage import CompletionTokensDetails, CompletionUsage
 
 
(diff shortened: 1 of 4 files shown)


ai/dev/start_unitycatalog.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-LOOP_COUNT=36
+LOOP_COUNT=48
```
BenWilson2, line 4: We might want to introduce a server start retry policy if this is suddenly unstable. Not a blocker for this PR, though.

ai/integrations/openai/tests/helper_functions.py
```diff
@@ -5,8 +5,10 @@
     ChatCompletionMessage,
     Choice,
 )
-from openai.types.chat.chat_completion_message import ChatCompletionMessageToolCall
-from openai.types.chat.chat_completion_message_tool_call import Function
+from openai.types.chat.chat_completion_message_tool_call import (
```
BenWilson2, line 8: Do they introduce a hard-dependency on the openai client? If so, we're good with this change. If not, we'll need a version-locked check for the import path to support older versions importing from `openai.types.chat.chat_completion_message` vs the new path.

puneet-jain159: We were pulling the dependency from the wrong directory. I check backed the version and the `ChatCompletionMessageToolCall` API is present in all the version in the openai.types.chat folder. so we are good to go

Issue #1086, opened 2025-09-26
Refresh the s3 credential automatically so that we can support long-running spark jobs.


openinx:
In here, we just assigned the fixed temprorary session token for the s3/gcp/abfs object storage reading, while for long runing job,  it is possible to expire the temporary token.  So we need a mechnism to refresh the expired session token automatically, so that we can support the long-runing spark jobs. 

https://github.com/unitycatalog/unitycatalog/blob/030a08e7fcfa3f66d99d0668089af6a99a51c1d7/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala#L185

openinx:
Essentially,  we do the following steps to get the credential for s3/gcp/azure: 

* (client) --> (uc server) :   client request to uc server to get a correct `TemporaryCredentials`. 
* (client) --> Table :  client use the `TemporaryCredentials` to configure the Hadoop configuration. 
    * fs.s3a.access.key
    * fs.s3a.secret.key
    * fs.s3a.session.token
* (spark-drvier) :  use this hadoop configuration to read s3 object.
* (spark-executor):  use this hadoop configuration to read parquet files from the underlying object storage.

So for both the `spark-driver` and `spark-executor` need to refresh the credential if there is any expiration. And the key point is :  we are using the Hadoop S3AFilesytem to auth the s3 object storage, and it's a components that we cannot change or custormize.  But the good thing is, according to the hadoop s3a document: https://hadoop.apache.org/docs/r3.4.1/hadoop-aws/tools/hadoop-aws/index.html#Using_Session_Credentials_with_TemporaryAWSCredentialsProvider

It actually provides the configure key to specify our own customized the CredentialsProvider via the `fs.s3a.aws.credentials.provider`.   So the correct approach that I will suggest is:  Just customize a new `Credentialsprovider`  and inside that credentials provider we can automatically refresh the credentials, based on all the information that we need to.

openinx:
For aws s3,  I think we need to add new implementation for the `AwsCredentialsProvider`  https://github.com/apache/hadoop/blob/d44ac28e326c786b183d1793c4c9e46d03ec2380/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/auth/CredentialProviderListFactory.java#L247. 


For azure,  there is also a similar approach for people to customize the token provider. The configure key is `fs.adl.oauth2.access.token.provider`, and the related classes is : 
[...]

PR #1087, opened 2025-09-26
[Spark] Refresh the s3 credential automatically so that we can support long-running spark jobs.

openinx:
The aws s3 credentials refresh solution for this issue: https://github.com/unitycatalog/unitycatalog/issues/1086


connectors/spark/src/main/scala/io/unitycatalog/spark/AwsVendedTokenProvider.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private AwsS3Credentials createS3Credentials() throws URISyntaxException, ApiException {
+        URI uri = new URI(conf.get("fs.s3a.unitycatalog.uri"));
+        String token = conf.get("fs.s3a.unitycatalog.token");
+
+        ApiClient apiClient = ApiClientFactory.createApiClient(uri, token);
+        TemporaryCredentialsApi tempCredApi = new TemporaryCredentialsApi(apiClient);
+
+        // Get the temp credential request.
+        String jsonRequest = conf.get("fs.s3a.unitycatalog.request");
```
tdas: this has to be `fs.s3a.unitycatalog.credential.request`

connectors/spark/src/main/scala/io/unitycatalog/spark/Constants.java
```diff
@@ -0,0 +1,20 @@
+package io.unitycatalog.spark;
+
+public class Constants {
+    public Constants() {}
+
+    public static final String UNITY_CATALOG_URI = "fs.unitycatalog.uri";
```
tdas, line 6: nit: lets shorten these constant variable by 
- either `UNITY_CATALOG_` --> `UC_`
- or ... we name the class to UCConstants and remove the UC prefix completely. everything will use UCConstants.URI

tdas: actually... on a different note. are all these constants tied to HadoopConf only? 

then it might be worthwhile to put that information in the naming somewhere. Say UCHadoopConfs instead of generic UCConstants.

connectors/spark/src/main/scala/io/unitycatalog/spark/AwsVendedTokenProvider.java
```diff
@@ -0,0 +1,125 @@
...
+    private volatile AwsS3Credentials awsS3Credentials;
+    private volatile ApiClient lazyApiClient = null;
+
+    /**
+     * Constructor for the hadoop's CredentialProviderListFactory#buildAWSProviderList to initialize.
+     */
+    public AwsVendedTokenProvider(URI ignored, Configuration conf) {
+        this.conf = conf;
+        this.uri = URI.create(conf.get(Constants.UNITY_CATALOG_URI));
+        this.token = conf.get(Constants.UNITY_CATALOG_TOKEN);
```
tdas, line 33: if these are not set, then how should we handle failures? does conf.get() return null if not set?

connectors/spark/src/main/scala/io/unitycatalog/spark/AwsVendedTokenProvider.java
```diff
@@ -0,0 +1,125 @@
...
+
+    protected TemporaryCredentialsApi temporaryCredentialsApi() {
+        return new TemporaryCredentialsApi(apiClient());
+    }
+
+    private AwsS3Credentials createS3Credentials() throws ApiException {
+        TemporaryCredentialsApi tempCredApi = temporaryCredentialsApi();
+
+        // Generate the temporary credential via requesting UnityCatalog.
+        TemporaryCredentials tempCred;
```
tdas, line 78: where are you using the initial hadoop credential acquired before job start?

connectors/spark/src/main/scala/io/unitycatalog/spark/AwsVendedTokenProvider.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        return lazyApiClient;
+    }
+
+    @Override
+    public AwsCredentials resolveCredentials() {
+        if(awsS3Credentials == null || awsS3Credentials.readyToRenew()) {
+            synchronized (this) {
+                if(awsS3Credentials == null || awsS3Credentials.readyToRenew()) {
```
tdas, line 52: this provider object will be recreated on every file access (since we disabled FS cache) ...isnt it? so we need to cache the credentials outside this objects lifecycle isnt it?

connectors/spark/src/main/scala/io/unitycatalog/spark/Constants.java
```diff
@@ -0,0 +1,20 @@
...
+
+    // Keys for table based temporary credential requests
+    public static final String UNITY_CATALOG_TABLE = "fs.unitycatalog.table";
+    public static final String UNITY_CATALOG_TABLE_OPERATION = "fs.unitycatalog.table.operation";
+
+    // Keys for path based temporary credential requests.
+    public static final String UNITY_CATALOG_PATH = "fs.unitycatalog.path";
+    public static final String UNITY_CATALOG_PATH_OPERATION = "fs.unitycatalog.path.operation";
+
+    public static final String UNITY_CATALOG_CREDENTIALS_TYPE = "fs.unitycatalog.credentials.type";
```
tdas, line 17: lets make the conf key strings have `_KEY` at the end of constant var name. for example ... `...CREDENTIALS_TYPE_KEY`. 

this is to distinguish the keys string from the value strings

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -173,32 +186,54 @@ object UCSingleCatalog {
...
+
+  private def commonProps(scheme: String, unityCatalogURL: String, unityCatalogToken: String, tempCred: TemporaryCredentials): Map[String, String] = {
     if (scheme == "s3") {
-      val awsCredentials = temporaryCredentials.getAwsTempCredentials
+      val awsCredentials = tempCred.getAwsTempCredentials
       Map(
-        // TODO: how to support s3:// properly?
         "fs.s3a.access.key" -> awsCredentials.getAccessKeyId,
         "fs.s3a.secret.key" -> awsCredentials.getSecretAccessKey,
         "fs.s3a.session.token" -> awsCredentials.getSessionToken,
```
tdas, line 217: in a separate PR, it would be good to refactor these constant key values to the same place as the new keys like URI

connectors/spark/src/test/java/io/unitycatalog/spark/AwsVendedTokenProviderTest.java
```diff
@@ -0,0 +1,138 @@
...
+import io.unitycatalog.client.model.TemporaryCredentials;
+import java.net.URI;
+import org.apache.hadoop.conf.Configuration;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
+
+public class AwsVendedTokenProviderTest {
+
+  @Test
+  public void testTableBasedTemporaryCredentialsRenew() throws Exception {
```
tdas, line 20: are you testing all the negative cases ... like url and token are not set in the hadoop conf. we should fail gracefully (e.g. keep using initial hadoop conf until expiry and fail only when renew needed).. and with good error messages (not null pointer exceptions, etc.)

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
```diff
@@ -143,7 +143,8 @@ public void testTimeTravelDeltaTable() throws ApiException, IOException {
 
     // Time-travel to before the last insert, we should only see the first inserted row.
     validateTimeTravelDeltaTable(session.sql("SELECT * FROM " + t1 + " VERSION AS OF 1"));
-    validateTimeTravelDeltaTable(session.sql("SELECT * FROM " + t1 + " TIMESTAMP AS OF '" + timestamp + "'"));
+    validateTimeTravelDeltaTable(
```
tdas, line 146: are the existing tests testing table level isolation? same query on separate tables use different credentials correctly?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -173,32 +186,54 @@ object UCSingleCatalog {
...
+        Constants.UNITY_CATALOG_URI -> unityCatalogURL,
+        Constants.UNITY_CATALOG_TOKEN -> unityCatalogToken,
+        "fs.s3a.aws.credentials.provider" -> classOf[AwsVendedTokenProvider].getName,
```
tdas, line 220: there is risk that in some environment this new stuff can cause errors while loading the provider. lets use a separate spark conf to enable the renew for now. we can always set it to be enable by default by changing the default value of the flag after we have gained confidence.

connectors/spark/src/main/scala/io/unitycatalog/spark/AwsVendedTokenProvider.java
openinx: I think here using `Preconditions.checkNotNull` should be enough, to throw the exception to the upper layer.  Let me update it in the new PR. https://github.com/unitycatalog/unitycatalog/issues/1096

connectors/spark/src/main/scala/io/unitycatalog/spark/AwsVendedTokenProvider.java
openinx: Oh,  thanks for the great point !  I think you are right, we may need a global credentials cache.  Let me take a close look at this part.

connectors/spark/src/main/scala/io/unitycatalog/spark/AwsVendedTokenProvider.java
openinx: The basic idea is that we request the credential in the spark driver side, and then set it in the `hadoopConf`, and finally pass them to the executors.  The new PR implement this: https://github.com/unitycatalog/unitycatalog/pull/1098/files#diff-9116a150ee4790736eb5b39d5ef0f4c16f40db3286a997c4676a93562196af49R40-R52

connectors/spark/src/main/scala/io/unitycatalog/spark/Constants.java
openinx: Thanks for the comments. I think it's fair,  let me reflect in the new PR.

connectors/spark/src/main/scala/io/unitycatalog/spark/Constants.java
openinx: Cool, that makes more clear !

connectors/spark/src/test/java/io/unitycatalog/spark/AwsVendedTokenProviderTest.java
openinx: Thanks for pointing this out.  Will work on the new PR to address those comments !

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
openinx: I think it's because we don't enforce the codestyle check before PR merging,  so after I use the command to format the code, then it will automatically format this unrelated code. 

```
./build/sbt javafmtAll
```

I have an issue for this before: https://github.com/unitycatalog/unitycatalog/issues/1099

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: Yes,  will do that.  Thank you for the comment !

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: > there is risk that in some environment this new stuff can cause errors while loading the provider

Agreed, there is risk.   If someone encounter this error and if he/she want to acess the UC,  then the only way that I can think of is to fix it, because finally they need this `AwsVendedTokenProvider` to create the temporary credentials.  There is no other way.  If this is correct, then seem no neccessary to introduce another switch to control enable the `AwsVendedTokenProvider` or not.  

Please correct me if I misunderstood, thanks for the comment !

openinx:
All aws s3 credentials provider with credential renewal PRs was merged now,  close this one.  Thanks @tdas for the all the reviewing work.

Issue #1103, opened 2025-10-08
Introduce the generic credential cache to reuse the table-level / path-level credential in the vended token provider.


openinx:
It's a todo task from this code review: https://github.com/unitycatalog/unitycatalog/pull/1098/files#r2412513873

Let's create separate PR for this task.

openinx:
Here is the PR for this issue : https://github.com/unitycatalog/unitycatalog/pull/1108

openinx:
Closed via https://github.com/unitycatalog/unitycatalog/commit/0257c25f7d233f0973fa54f3922e78d25519d9ea

PR #1108, opened 2025-10-09
Add GenericCredential cache to reuse the table-level / path-level credential in the vended token provider.

openinx:
**PR Checklist**

**Description of changes**

Related issue: https://github.com/unitycatalog/unitycatalog/issues/1103

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
@@ -15,8 +15,9 @@ public class UCHadoopConf {
   public static final String UC_URI_KEY = "unitycatalog.uri";
   public static final String UC_TOKEN_KEY = "unitycatalog.token";
 
-  // Key for a credential unique ID. It identifies a job-level credential for a specific table,
-  // meaning the same job and table combination will share the same credential.
+  // Key representing a unique credential ID. It identifies a job-level credential for a specific
+  // table, meaning that the same job–table combination shares the same credential. Cached
+  // credentials are indexed by this key and are not reused across different jobs.
   public static final String UC_CREDENTIALS_UID_KEY = "unitycatalog.credentials.uid";
 
   // Keys for table based temporary credential requests
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
@@ -20,10 +20,15 @@ public abstract class GenericCredentialProvider {
   private static final long DEFAULT_RENEWAL_LEAD_TIME_MILLIS = 30 * 1000;
 
   // The credential cache, for saving QPS to unity catalog server.
-  static final Cache<String, GenericCredential> CACHE = CacheBuilder
-      .newBuilder()
-      .maximumSize(1000)
-      .build();
+  static final Cache<String, GenericCredential> globalCache;
+  private static final String UC_CREDENTIAL_CACHE_MAX_SIZE =
+      "unitycatalog.credential.cache.max.size";
+  private static final long UC_CREDENTIAL_CACHE_MAX_SIZE_DEFAULT = 1024;
+
+  static {
+    long maxSize = Long.getLong(UC_CREDENTIAL_CACHE_MAX_SIZE, UC_CREDENTIAL_CACHE_MAX_SIZE_DEFAULT);
+    globalCache = CacheBuilder.newBuilder().maximumSize(maxSize).build();
+  }
 
   private final Configuration conf;
   private final URI ucUri;
@@ -99,15 +104,15 @@ public abstract class GenericCredentialProvider {
 
   private GenericCredential renewCredential() throws ApiException {
     if (credCacheEnabled) {
-      synchronized (CACHE) {
-        GenericCredential cached = CACHE.getIfPresent(credUid);
+      synchronized (globalCache) {
+        GenericCredential cached = globalCache.getIfPresent(credUid);
         // Use the cached one if existing and valid.
         if (cached != null && !cached.readyToRenew(renewalLeadTimeMillis)) {
           return cached;
         }
         // Renew the credential and update the cache.
         GenericCredential renewed = createGenericCredentials();
-        CACHE.put(credUid, renewed);
+        globalCache.put(credUid, renewed);
         return renewed;
       }
     } else {
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
@@ -32,12 +32,12 @@ public class AwsVendedTokenProviderTest {
 
   @BeforeEach
   public void before() {
-    GenericCredentialProvider.CACHE.invalidateAll();
+    GenericCredentialProvider.globalCache.invalidateAll();
   }
 
   @AfterEach
   public void after() {
-    GenericCredentialProvider.CACHE.invalidateAll();
+    GenericCredentialProvider.globalCache.invalidateAll();
   }
 
   @Test
@@ -362,9 +362,9 @@ public class AwsVendedTokenProviderTest {
 
   private static void assertGlobalCache(int expectedSize, TemporaryCredentials... creds) {
     assertThat(expectedSize).isEqualTo(creds.length);
-    assertThat(AwsVendedTokenProvider.CACHE.size()).isEqualTo(expectedSize);
+    assertThat(AwsVendedTokenProvider.globalCache.size()).isEqualTo(expectedSize);
     for (TemporaryCredentials cred : creds) {
-      assertThat(AwsVendedTokenProvider.CACHE.asMap().values())
+      assertThat(AwsVendedTokenProvider.globalCache.asMap().values())
           .contains(new GenericCredential(cred));
     }
   }


connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredKey.java
```diff
@@ -0,0 +1,96 @@
...
+import io.unitycatalog.client.model.TableOperation;
+import java.util.Objects;
+
+public interface GenericCredKey {
+
+  class TableCredKey implements GenericCredKey {
+    private final String uri;
+    private final String token;
+    private final String tableId;
+    private final TableOperation tableOp;
```
openinx: The cache key for a temporary table cred, must identify by the following 4 fields:
* uri -->  identify UC server.
* token --> identify the account.
* tableId  --> identify the table.
* tableOp --> identify the table operation.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -39,6 +46,13 @@ public GenericCredentialProvider(Configuration conf) {
         "'%s' is not set in hadoop configuration", UCHadoopConf.UC_TOKEN_KEY);
     this.ucToken = conf.get(UCHadoopConf.UC_TOKEN_KEY);
 
+    this.globalCredCacheEnabled = conf.getBoolean(
+        UCHadoopConf.UC_GLOBAL_CREDENTIAL_CACHE_ENABLED_KEY,
+        UCHadoopConf.UC_GLOBAL_CREDENTIAL_CACHE_ENABLED_VALUE);
+
+    // Init the generic credential key for caching.
+    this.genericCredKey = createGenericCredKey(conf);
```
openinx: For each `GenericCredentialProvider`, it should have only one key... for identifying the cache entry.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -27,4 +27,9 @@ private UCHadoopConf() {
   public static final String UC_CREDENTIALS_TYPE_KEY = "unitycatalog.credentials.type";
   public static final String UC_CREDENTIALS_TYPE_TABLE_VALUE = "table";
   public static final String UC_CREDENTIALS_TYPE_PATH_VALUE = "path";
+
+  // Key to enable the global credential cache.
+  public static final String UC_GLOBAL_CREDENTIAL_CACHE_ENABLED_KEY =
+      "unitycatalog.global.credential.cache.enabled";
+  public static final boolean UC_GLOBAL_CREDENTIAL_CACHE_ENABLED_VALUE = true;
```
tdas: did you meant this to be the default value? if so, then please put default in the name.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -27,4 +27,9 @@ private UCHadoopConf() {
   public static final String UC_CREDENTIALS_TYPE_KEY = "unitycatalog.credentials.type";
   public static final String UC_CREDENTIALS_TYPE_TABLE_VALUE = "table";
   public static final String UC_CREDENTIALS_TYPE_PATH_VALUE = "path";
+
+  // Key to enable the global credential cache.
+  public static final String UC_GLOBAL_CREDENTIAL_CACHE_ENABLED_KEY =
```
tdas: nit: you can remove _GLOBAL .. just CREDENTIAL_CACHE is good enough. its understood that the cache is global.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredKey.java
tdas: how are we going to make sure that the same credential is not going to be reused across jobs? we want to ensure that every job users a new set of credentials... so that if permissions are revoked the next job immediately fails.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredKey.java
```diff
@@ -0,0 +1,96 @@
...
+      this.tableOp = tableOp;
+    }
+
+    public String tableId() {
+      return tableId;
+    }
+
+    public TableOperation tableOp() {
+      return tableOp;
+    }
```
tdas: do you need these public methods? once you define the key, do you need to access the key's details?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -79,38 +89,82 @@ protected TemporaryCredentialsApi temporaryCredentialsApi() {
...
     } else {
       throw new IllegalArgumentException(String.format(
           "Unsupported unity catalog temporary credentials type '%s', please check '%s'",
           type, UCHadoopConf.UC_CREDENTIALS_TYPE_KEY));
     }
+  }
 
-    return new GenericCredential(tempCred);
+  private GenericCredential renewCredential() {
+    if (globalCredCacheEnabled) {
```
tdas: why do we need a flag for enabling? when do we want it to be disabled?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -79,38 +89,82 @@ protected TemporaryCredentialsApi temporaryCredentialsApi() {
...
+      return new GenericCredKey.PathCredKey(ucUri.toString(), ucToken, path, pathOp);
     } else if (UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE.equals(type)) {
       String tableId = conf.get(UCHadoopConf.UC_TABLE_ID_KEY);
-      String tableOperation = conf.get(UCHadoopConf.UC_TABLE_OPERATION_KEY);
+      Preconditions.checkNotNull(tableId,
+          "'%s' is not set in hadoop configuration", UCHadoopConf.UC_TABLE_ID_KEY);
+
+      String tableOpStr = conf.get(UCHadoopConf.UC_TABLE_OPERATION_KEY);
+      Preconditions.checkNotNull(tableOpStr,
+          "'%s' is not set in hadoop configuration", UCHadoopConf.UC_TABLE_ID_KEY);
```
tdas:
```suggestion
          "'%s' is not set in hadoop configuration", UCHadoopConf.UC_TABLE_OPERATION_KEY);
```
i think it might better to make a function that combines conf.get and Preconditions.checkNotNull. There are 6 instances of this pair. so a function will reduce code and also avoid mistakes like above.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
```diff
@@ -291,7 +384,15 @@ private static void assertCredentials(
     assertThat(actualSessionCred.sessionToken()).isEqualTo(expectedAwsCred.getSessionToken());
   }
 
-  private static TemporaryCredentials newAwsTempCredentials(
+  private static TemporaryCredentials newAwsTempCred(String idSuffix, long expirationTime) {
```
tdas:
```suggestion
  private static TemporaryCredentials newAwsTempCred(String idSuffix, long expirationTimeMs) {
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
openinx: Done, thanks for the comment !

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
openinx: Yes,  it's the default value. Done, thanks!

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredKey.java
openinx: Thanks for the great question. 

If the spark job is deployed in YARN, or k8s, per job per cluster. It should be fine, since for every job termination, the process level cache will be invalid. 

But if the spark job is deployed in standalone, then all those jobs will share the same executors and driver.  Then the shared cache from one job will be easily accessible for another job.  The standalone cluster usually owned by one user, then it should be fine for different jobs from the same user to share the credential cache.  I mean reusing the parallel jobs' refreshed credential for another job, to avoiding an extra credential refreshing. 

As you said, the question is: what if the credential from one job is revoked, then other related jobs need to fail immediately. I think we can think the two cases:

1.   If people start a new job, it should immediately fail, since we always need to initialize the credentials for the first initializaiton ( and then set them in the hadoop conf, finaly dispatching from driver to executors ). For the initialization,  it will fail fast. 

2.  If one job (A)'s credential is revoked, another running job (B) is reusing this credentials.  Then when the job B send http requests to the s3,  the s3 should also detect the revoked credential, and fail the job B. (Another question is:  will it make sense to fail the Job B because of the Job A's credential revocation ?).

In summary,  for yarn & k8s, that's should be fine.  for standalone cluster,   fast-failure should also be fine. Seems like have some corner cases we need to address.

openinx: Do you think we need to add the `jobId` dimenssion in those key ?  If so, how can we corrently get the `jobId` and finally set it ?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredKey.java
openinx: Yes, we need those methods to build the UC restful requests. Please see : https://github.com/unitycatalog/unitycatalog/pull/1108/files#diff-68d2e51ab5a4e6186ba02fecda9958db316feac1f8397a17c1da4cba334be188R149-R157

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
openinx: I'm considering if in future we want to enable the fs impl cache,  the filesystem is mapping to one single provider. and the filesystem have already been cached, then we finally may no need this credential.  Just like the current `S3AFileSystem`,   which have a single `credentialsProvider` with having the latest valid credential only.

openinx:
The `MLflow-UC` integration test is timeout , https://github.com/unitycatalog/unitycatalog/actions/runs/18413247521/job/52470796766?pr=1108 , which I think it's unrelated to this PR.  we can retry the CI. 

```bash
Waiting for UC server... (42/48)
Waiting for UC server... (43/48)
Waiting for UC server... (44/48)
Waiting for UC server... (45/48)
Waiting for UC server... (46/48)
Waiting for UC server... (47/48)
Waiting for UC server... (48/48)
ai/dev/start_unitycatalog.sh: line 36: kill: (2152) - No such process
UC server failed to start within 240 seconds
Error: Process completed with exit code 1.
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -15,6 +15,10 @@ private UCHadoopConf() {
   public static final String UC_URI_KEY = "unitycatalog.uri";
   public static final String UC_TOKEN_KEY = "unitycatalog.token";
 
+  // Key for a credential unique ID. It identifies a job-level credential for a specific table,
+  // meaning the same job and table combination will share the same credential.
+  public static final String UC_CREDENTIALS_UID_KEY = "unitycatalog.credentials.uid";
```
openinx, line 21: This uid will be generated when `catalog#loadTable` via uuid. So it should be always unique, which represent the `<Job, table>` combination.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
```diff
@@ -17,16 +17,29 @@
...
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
 
 public class AwsVendedTokenProviderTest {
 
+  @BeforeEach
+  public void before() {
+    GenericCredentialProvider.CACHE.invalidateAll();
```
openinx: The invalidateAll will clean all the cached entries, for preparing a clean test evironment.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -50,8 +68,8 @@ public GenericCredential accessCredentials() {
       synchronized (this) {
         if (credential == null || credential.readyToRenew(renewalLeadTimeMillis)) {
           try {
-            credential = createGenericCredentials();
-          } catch (Exception e) {
+            credential = renewCredential();
```
openinx, line 76: Different threads access the same provider, should be safe here since we've the `synchronzied(this)`.  However the `Cache` is global and process level,  so we still need to lock the `CACHE` for its renewal implementation.  Please see: https://github.com/unitycatalog/unitycatalog/pull/1108/files#diff-68d2e51ab5a4e6186ba02fecda9958db316feac1f8397a17c1da4cba334be188R102

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
tdas: Please add the reason why we are adding this. That is, so that the cached credentials are keyed by this and does not get used across multiple jobs.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -11,16 +11,25 @@
...
 public abstract class GenericCredentialProvider {
-  // The time remaining until expiration, we will try to renew the credential before the expiration
-  // time.
+  // The remaining time before expiration, used to trigger renewal in advance.
   private static final long DEFAULT_RENEWAL_LEAD_TIME_MILLIS = 30 * 1000;
 
+  // The credential cache, for saving QPS to unity catalog server.
+  static final Cache<String, GenericCredential> CACHE = CacheBuilder
+      .newBuilder()
+      .maximumSize(1000)
```
tdas: lets make this configurable in the conf.

tdas: are you sure its LRU?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -11,16 +11,25 @@
...
+import org.sparkproject.guava.cache.CacheBuilder;
 
 public abstract class GenericCredentialProvider {
-  // The time remaining until expiration, we will try to renew the credential before the expiration
-  // time.
+  // The remaining time before expiration, used to trigger renewal in advance.
   private static final long DEFAULT_RENEWAL_LEAD_TIME_MILLIS = 30 * 1000;
 
+  // The credential cache, for saving QPS to unity catalog server.
+  static final Cache<String, GenericCredential> CACHE = CacheBuilder
```
tdas: typically constants are capitalized .. isnt it?
maybe just call this `globalCache`

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
```diff
@@ -244,39 +219,156 @@ public void testPathTemporaryCredentialsRenewWithInitialCredentials() throws Exc
...
+  @Test
+  public void testGlobalCredCache() throws Exception {
+    Configuration confA = newTableBasedConf("tableA");
+    Configuration confB = newTableBasedConf("tableB");
+    Configuration confC = newPathBasedConf("pathC");
+    Configuration confD = newPathBasedConf("pathD");
+
+    TemporaryCredentialsApi tempCredApi = mock(TemporaryCredentialsApi.class);
+    // For TableA's 1st renewal
+    TemporaryCredentials cred1 = newAwsTempCred("1", System.currentTimeMillis() + 2000L);
```
tdas, line 231: the path creds and table creds are completely independent of each other. rather than duplicate the code, why not put it in a loop. something like this
```
for (String isPath: {true, false}) {
   // define confA, confB, credA1, credB1, credA2, credB2
   if (isPath) {
      when(tempCredApi.generateTemporaryPathCredentials(any())) .... 
   } else {
      when(tempCredApi.generateTemporaryTableCredentials(any()))   
   }
   // further tests
}
```

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
```diff
@@ -102,28 +109,13 @@ public void testTableTemporaryCredentialsRenewWithInitialCredentials() throws Ex
     Configuration conf = newTableBasedConf();
 
     // Use the generated credential to initialize the provider.
-    TemporaryCredentials cred0 =
-        newAwsTempCredentials(
-            "accessKeyId0",
-            "secretAccessKey0",
-            "sessionToken0",
-            System.currentTimeMillis() + 2000L);
+    TemporaryCredentials cred0 = newAwsTempCred("0", System.currentTimeMillis() + 2000L);
```
tdas, line 112: this is good code deduplication!

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
openinx: Done,  Thanks !

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
openinx: > lets make this configurable in the conf.

Sounds good to me,  let me fix this. 

> are you sure its LRU?

https://github.com/google/guava/wiki/cachesexplained#size-based-eviction

According to the document,   `The cache will try to evict entries that haven't been used recently or very often. Warning: the cache may evict entries before this limit is exceeded -- typically when the cache size is approaching the limit.` .  so by default, guava evicts entries using an approximation of LRU (Least Recently Used) policy.  I think it should be fine here.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
openinx: Yes,  make sense.  for constant values, we usually capitalize it.  while for a mutable object referenece we won't. Thanks for pointing it out !

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
openinx: > lets make this configurable in the conf.

Since the `globalCache` is a static global reference,  so we cannot use the `GenericCredentialProvider` instance's `Configuration` to set the max size, otherwise different `Configuration` will set different max size for the `globalCache`, which may lead to confusion. 

A better way is using the system's property to set the cache max size,  then it should provide a consistent global cache size.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
openinx: Hi @tdas ,  thanks for the comment.  The `path` creds and `table` creds actually are not completely independent, since the cache is a global cache,  we want to put the mixed creds with different types of path creds and table creds, into the global cache.  And for each renewal, we expect the global cache can always get the correct cred corressponding to its type ( path cred or table cred).

tdas: ahh.. good point. 

but would have been easier to read if objects were named clearly.. which is path and which is table. like pathAConf, pathBConf, pathACred1, pathACred2, ... and similarly tableAConf, tableBConf, ...

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -11,16 +11,30 @@
...
 public abstract class GenericCredentialProvider {
-  // The time remaining until expiration, we will try to renew the credential before the expiration
-  // time.
+  // The remaining time before expiration, used to trigger renewal in advance.
   private static final long DEFAULT_RENEWAL_LEAD_TIME_MILLIS = 30 * 1000;
 
+  // The credential cache, for saving QPS to unity catalog server.
+  static final Cache<String, GenericCredential> globalCache;
+  private static final String UC_CREDENTIAL_CACHE_MAX_SIZE =
+      "unitycatalog.credential.cache.max.size";
```
tdas, line 25:
```suggestion
      "unitycatalog.credential.cache.maxSize";
```
typically we each `.<string>.` is a scope  under which there can be many more confs like `<scope>.<key1>`m `<scope>.<key2>`. so its usually a component or a concept, like `cache` or `credential`, and sub keys are tied to that component or concept.


`max.` is a not a component or concept. rather "max size" is tied to the `cache` component.

tdas:
merged your PR, but please address the last 2 comments in follow up PRs. thank you!

Issue #1092, opened 2025-10-03
Add the s3 integrationt test


openinx:
Currently,  I see our `UCSingleCatalog` use a `CredentialTestFileSystem` which built on top of the `RawLocalFileSystem`  to mock the `S3AFileSystem`, `AbsFileSystem`, and `GCPFileSystem`.  

I think it's  reasonable for the unit test, but it actually not reflect the real s3/gcp/azure object storage.  I think we can add a real integration tests for that.   It's not necessary to run for every PR,  but it's good to run those test in a  real  s3/gcp/azure object storage,  for the release.

Then  we can detect the the potential issue before the release.  Open an issue here,  will try to initial a PR for that when I have the time to do it.

openinx:
Located a pontential risk in this issue, when I tried to pass the IT tests: https://github.com/unitycatalog/unitycatalog/issues/1112

openinx:
Another potential risks is: for the IT test,  seems like the OSS UC server will issue an aws credentials without any expiration time.  maybe I need to investigate more about this.  

After the investigation, so the problem is the OSS UC server will return the static `accessKey`, `secretKey`, `sessionToken` without expiration time. 

https://github.com/unitycatalog/unitycatalog/blob/a7ecc12191365d60f032cbd926231744271c89bc/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java#L32-L39

So I think we also need to handle the static `accessKey`, `secretKey`, `sessionToken` info in the credentials provider..

openinx:
Let's just make this issue to be focused on aws s3.

openinx:
I just use this command to test the integration test https://github.com/unitycatalog/unitycatalog/pull/1119

```bash
export CATALOG_URI=$UC_URI &&\
export CATALOG_AUTH_TOKEN=$UC_TOKEN &&\
export CATALOG_NAME=main &&\
export SCHEMA_NAME=demo_zh &&\
export TABLE_NAME=ittest &&\
export S3_BASE_LOCATION=$S3_LOCATION && \
SBT_OPTS="-Xmx8G -XX:+UseG1GC" \
./build/sbt "integrationTests/testOnly io.unitycatalog.integrationtests.SparkCredentialRenewalTest"
```

openinx:
When I did the integration tests, I encountered this issue: 

```java
[...]

PR #1110, opened 2025-10-10
Integrate the AwsVendedTokenProvider into the UCSingleCatalog

openinx:
**PR Checklist**

**Description of changes**

Related issue: https://github.com/unitycatalog/unitycatalog/issues/1092 

Let's add the integration test for `AwsVendedTokenProvider` in this PR.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
@@ -114,7 +114,6 @@ public class CredPropsUtil {
         .uri(uri)
         .token(token)
         .uid(UUID.randomUUID().toString())
-        .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE)
         .set(UCHadoopConf.S3A_INIT_ACCESS_KEY, awsCred.getAccessKeyId())
         .set(UCHadoopConf.S3A_INIT_SECRET_KEY, awsCred.getSecretAccessKey())
         .set(UCHadoopConf.S3A_INIT_SESSION_TOKEN, awsCred.getSessionToken());
@@ -135,6 +134,7 @@ public class CredPropsUtil {
       TableOperation tableOp,
       TemporaryCredentials tempCreds) {
     return s3TempCredPropsBuilder(uri, token, tempCreds)
+        .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE)
         .tableId(tableId)
         .tableOperation(tableOp)
         .build();
@@ -147,6 +147,7 @@ public class CredPropsUtil {
       PathOperation pathOp,
       TemporaryCredentials tempCreds) {
     return s3TempCredPropsBuilder(uri, token, tempCreds)
+        .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_PATH_VALUE)
         .path(path)
         .pathOperation(pathOp)
         .build();


connectors/spark/src/main/scala/io/unitycatalog/spark/auth/AwsVendedTokenProvider.java
```diff
@@ -28,7 +28,9 @@ public GenericCredential initGenericCredential(Configuration conf) {
       String secretKey = conf.get(UCHadoopConf.S3A_INIT_SECRET_KEY);
       String sessionToken = conf.get(UCHadoopConf.S3A_INIT_SESSION_TOKEN);
 
-      long expiredTimeMillis = conf.getLong(UCHadoopConf.S3A_INIT_CRED_EXPIRED_TIME, 0L);
+      long expiredTimeMillis = conf.getLong(
+          UCHadoopConf.S3A_INIT_CRED_EXPIRED_TIME,
+          Long.MAX_VALUE);
```
openinx, line 39: @tdas , for the fixed credentials case,  the expiration time should be infinite.  That's why I changed its default value to `Long.MAX_VALUE`.

openinx: I changed it because currently we `TestReadWriteTest` use the constant ak/sk/sessionToken to bootstrap the OSS UC server, and the uc server think it's using the fixed credentials which will never to be expired. 

See: https://github.com/unitycatalog/unitycatalog/blob/a7ecc12191365d60f032cbd926231744271c89bc/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java#L32-L39

So for the `AwsVendedTokenProvider` in the integrated unit test, it will request the DB UC server, and get the fixed and never-expire credentials ( without setting any expiration time ). 

Thrown it here for reviewing.

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
```diff
@@ -0,0 +1,262 @@
...
+import java.util.Map;
+import java.util.UUID;
+import org.sparkproject.guava.base.Preconditions;
+import org.sparkproject.guava.collect.ImmutableMap;
+
+public class CredPropsUtil {
+  private CredPropsUtil() {
+  }
+
+  public static final ThreadLocal<Boolean> LOAD_DELTA_CATALOG = ThreadLocal.withInitial(() -> true);
```
openinx: We need to move those two thread local variables out of the `CredPropsUtil`.

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
```diff
@@ -0,0 +1,262 @@
...
+      set("fs.s3a.impl.disable.cache", "true");
+    }
+
+    @Override
+    protected S3PropsBuilder self() {
+      return this;
+    }
+  }
+
+  public static class GcsPropsBuilder extends PropsBuilder<GcsPropsBuilder> {
```
openinx: `GcsPropsBuilder` should also have some common properties but I did not set them here. Since this PR is focused on aws s3. Finally when we design the `GCS` integration test, we will set those common properties here.

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
```diff
@@ -0,0 +1,262 @@
...
+        .token(token)
+        .uid(UUID.randomUUID().toString())
+        .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE)
+        .tableId(tableId)
+        .tableOperation(tableOp)
+        .set(UCHadoopConf.S3A_INIT_ACCESS_KEY, awsCred.getAccessKeyId())
+        .set(UCHadoopConf.S3A_INIT_SECRET_KEY, awsCred.getSecretAccessKey())
+        .set(UCHadoopConf.S3A_INIT_SESSION_TOKEN, awsCred.getSessionToken());
+
+    // For the static credential case, null expiration time is possible.
```
openinx: nit: `null` --> `nullabe`.

openinx: For static credential case,  we should not use the renewal provider right ? But our `AwsVendedTokenProvider` also works pretty well for the static credential case. Just as I said in the above comment: https://github.com/unitycatalog/unitycatalog/pull/1110/files#r2431163508

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
```diff
@@ -0,0 +1,262 @@
...
+    if (tempCreds.getExpirationTime() != null) {
+      builder.set(UCHadoopConf.S3A_INIT_CRED_EXPIRED_TIME,
+          String.valueOf(tempCreds.getExpirationTime()));
+    }
+
+    builder.set("fs.s3a.access.key", awsCred.getAccessKeyId());
```
openinx: We don't need to set the three keys anymore, for this provider.  I will remove the 3 lines.

connectors/spark/src/main/scala/io/unitycatalog/spark/TableProps.java
```diff
@@ -0,0 +1,20 @@
+package io.unitycatalog.spark;
+
+import java.util.Map;
+
+public class TableProps {
+  private TableProps() {
+  }
+
+  public static final String FIXED_CREDENTIAL_ENABLED = "fixed.credential.enabled";
```
openinx: The option to switch between the fixed credential and the `AwsVendedTokenProvider` ( dynamically renew the credential), should be a table property.

openinx: User can change the property value via altering the property value.

openinx: I assume we will also introduce other table properties for the OSS unity catalog, so I initiated the `TableProps` class here.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -17,29 +21,29 @@ private UCHadoopConf() {
       "fs.azure.init.sas.token.expired.time";
 
   // Keys for the UnityCatalog client.
-  public static final String UC_URI_KEY = "unitycatalog.uri";
-  public static final String UC_TOKEN_KEY = "unitycatalog.token";
+  public static final String UC_URI_KEY = "fs.unitycatalog.uri";
```
openinx, line 28: `fs.unitycatalog.uri` or `fs.<s3a/abfs/gcs>.unitycatalog.uri` ?  What's the perference ?

openinx: Anyway, we need to add the `fs.` prefix,  for reason please see: https://github.com/unitycatalog/unitycatalog/issues/1112

connectors/spark/src/test/java/io/unitycatalog/spark/S3CredentialTestFileSystem.java
```diff
@@ -2,29 +2,71 @@
...
-        assertThat(conf.get("fs.s3a.session.token")).isEqualTo("sessionToken0");
+        provider = accessProvider(conf);
+        AwsSessionCredentials credentials = (AwsSessionCredentials) provider.resolveCredentials();
+        assertThat(credentials.accessKeyId()).isEqualTo("accessKey0");
+        assertThat(credentials.secretAccessKey()).isEqualTo("secretKey0");
+        assertThat(credentials.sessionToken()).isEqualTo("sessionToken0");
+
+        // assertThat(conf.get("fs.s3a.access.key")).isEqualTo("accessKey0");
+        // assertThat(conf.get("fs.s3a.secret.key")).isEqualTo("secretKey0");
+        // assertThat(conf.get("fs.s3a.session.token")).isEqualTo("sessionToken0");
```
openinx: I think I should also design parameterized tests to cover the `useFixedCred=true` cases.  will do that later. Currently,  all tests are runing with the temporaray credentials provider enabled.

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
openinx: Done

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
openinx: Done.

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
tdas: then lets not even add the GcsPropsBuilder in this PR.
all the Gcs stuff can be added in a PR later.

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
```diff
@@ -0,0 +1,252 @@
...
+import java.util.Map;
+import java.util.UUID;
+import org.sparkproject.guava.base.Preconditions;
+import org.sparkproject.guava.collect.ImmutableMap;
+
+public class CredPropsUtil {
+  private CredPropsUtil() {
+  }
+
+  public abstract static class PropsBuilder<T extends PropsBuilder<T>> {
```
tdas: do we need this to be public?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -0,0 +1,252 @@
...
+import io.unitycatalog.client.model.TableOperation;
+import io.unitycatalog.client.model.TemporaryCredentials;
+import io.unitycatalog.spark.auth.AbfsVendedTokenProvider;
+import io.unitycatalog.spark.auth.AwsVendedTokenProvider;
+import java.util.Map;
+import java.util.UUID;
+import org.sparkproject.guava.base.Preconditions;
+import org.sparkproject.guava.collect.ImmutableMap;
+
+public class CredPropsUtil {
```
tdas, line 20: shouldnt this file be in auth package since its tied to credentials?

connectors/spark/src/main/scala/io/unitycatalog/spark/TableProps.java
```diff
@@ -0,0 +1,20 @@
+package io.unitycatalog.spark;
+
+import java.util.Map;
+
+public class TableProps {
+  private TableProps() {
+  }
+
+  public static final String FIXED_CREDENTIAL_ENABLED = "fixedCredential.enabled";
```
openinx: Do we have any preference for the table property naming ?

connectors/spark/src/main/scala/io/unitycatalog/spark/TableProps.java
```diff
@@ -0,0 +1,20 @@
+package io.unitycatalog.spark;
+
+import java.util.Map;
+
+public class TableProps {
```
tdas: why is this table props and not an overall simple spark conf or spark sql conf? 
why would anyone want to enable or disable refresh on a table-by-table basis? I expect that if someone has to set this... but they want to do it for their long-running application whatever all the tables that application is accessing.

connectors/spark/src/main/scala/io/unitycatalog/spark/TableProps.java
tdas: "fixed" does not make intuitive sense to me. can we use "refresh" or "renew"?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -0,0 +1,252 @@
...
+  public static Map<String, String> createPathCredProps(
+      boolean fixedCredEnabled,
+      String scheme,
+      String uri,
+      String token,
+      String path,
+      PathOperation pathOp,
+      TemporaryCredentials tempCreds) {
+    switch (scheme) {
+      case "s3":
```
tdas, line 215: shouldnt it cover s3a as well?

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
```diff
@@ -152,18 +154,28 @@ public void testTimeTravelDeltaTable() throws ApiException, IOException {
   }
 
   @ParameterizedTest
-  @ValueSource(strings = {"s3", "gs", "abfs"})
-  public void testCredentialParquet(String scheme) throws ApiException, IOException {
+  @CsvSource({"s3, true", "s3, false", "gs, true", "abfs, true"})
```
openinx: NOTICE: here I set the `s3`'s `fixedCredEnabled` to be false, which means we will use the `AwsVendedProvider` to access the temporay cred from the UC server, and renew it automatically. 

We don't set it for `gs` and `abfs` here. because this PR should focus on s3 e2e test.  Further PR will enable the e2e test for `gs` and `abfs`.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
tdas: understood. please write down this hard-to-remember context as a comment in this file.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -279,8 +247,21 @@ private class UCProxy(
...
+      t.getProperties,
+      TableProps.FIXED_CREDENTIAL_ENABLED,
+      TableProps.DEFAULT_FIXED_CREDENTIAL_ENABLED)
+    val extraSerdeProps = CredPropsUtil.createTableCredProps(
+      fixedCredEnabled,
+      locationUri.getScheme,
+      uri.toString,
+      token,
+      tableId,
+      TableOperation.READ_WRITE,
```
tdas: why is it always READ_WRITE? shouldnt we identify what type of credential we were successful in getting and specify that type of credential to be fetched on renew (rather than trying twice everytime)?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -0,0 +1,252 @@
...
+        if (fixedCredEnabled) {
+          return s3FixedCredProps(tempCreds);
+        } else {
+          return s3PathTempCredProps(uri, token, path, pathOp, tempCreds);
+        }
+      case "gs":
```
tdas, line 221: we have support fs and abfss before.. why not maintain that non-renew support here as well?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/AwsVendedTokenProvider.java
openinx: I'm not sure it's good to set it be `0` or `LONG.MAX_VALUE`.  Seems like the document is allowed to set `0` for the `expiration_time`.  https://docs.databricks.com/api/gcp/workspace/temporarytablecredentials/generatetemporarytablecredentials

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
```diff
@@ -0,0 +1,252 @@
...
+  public static Map<String, String> s3FixedCredProps(TemporaryCredentials tempCreds) {
+    AwsCredentials awsCred = tempCreds.getAwsTempCredentials();
+    return new S3PropsBuilder()
+        .set("fs.s3a.access.key", awsCred.getAccessKeyId())
+        .set("fs.s3a.secret.key", awsCred.getSecretAccessKey())
+        .set("fs.s3a.session.token", awsCred.getSessionToken())
+        .build();
+  }
+
+  public static Map<String, String> s3TableTempCredProps(
```
tdas: quite a bit of duplicate code between s3TableTempCredProps and s3PathTempCredProps

see if you can dedup? 

the only difference is the property ... table id or path .. isnt it? maybe that can be passed as a string with a boolean flag saying its a path or a tableid?

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
```diff
@@ -0,0 +1,252 @@
...
+        .set(GcsVendedTokenProvider.ACCESS_TOKEN_EXPIRATION_KEY,
+            String.valueOf(tempCreds.getExpirationTime()))
+        .set("fs.gs.create.items.conflict.check.enable", "false")
+        .set("fs.gs.auth.type", "ACCESS_TOKEN_PROVIDER")
+        .set("fs.gs.auth.access.token.provider", GcsVendedTokenProvider.class.getName())
+        .set("fs.gs.impl.disable.cache", "true")
+        .build();
+  }
+
+  public static Map<String, String> abfsProps(TemporaryCredentials tempCreds) {
```
tdas: why is there no path stuff for gs and abfs? did we not support those before?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/AwsVendedTokenProvider.java
tdas: let me think about it.

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
openinx: Fine, let's use the immutable map to set those static properties, without introducing the `GcsPropsBuilder`.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
openinx: Done! thanks for the comment !

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
openinx: No,  I think all those builders can be private.  Let's change it.

openinx: Done !

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
openinx: Make sense to move,  let's move it there.

connectors/spark/src/main/scala/io/unitycatalog/spark/TableProps.java
openinx: I think you are right. Seems like the real demand is setting at application level, not the table level.  I naturally thought all credentials settings are table-level, then the flag should also be table-level.  Thanks for pointing this out, then I think we will need to parse the value from the spark catalog options.

connectors/spark/src/main/scala/io/unitycatalog/spark/TableProps.java
openinx: Done ! renamed it to `renewCredential.enabled`.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
openinx: I think we should cover both `s3` and `s3a` scheme, but I think that should be out of scope for this PR. Seems in the original version, it also does not support `s3a`.  Also we need separate tests to address this one.

openinx: Filed a separate issue for it: https://github.com/unitycatalog/unitycatalog/issues/1115

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: Agreed, done !

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
openinx: Enabling the credential renewal for `gs` and `abfs`, is not the target of this PR. so for `gs` and `abfs`, let's keep it as it is.  The further e2e test PR will enable for `gs` and `abfs`.

connectors/spark/src/main/scala/io/unitycatalog/spark/GcsVendedTokenProvider.java
```diff
@@ -7,30 +7,30 @@
 import java.time.Instant;
 
 public class GcsVendedTokenProvider implements AccessTokenProvider {
-    protected static final String ACCESS_TOKEN_KEY = "fs.gs.auth.access.token.credential";
-    protected static final String ACCESS_TOKEN_EXPIRATION_KEY = "fs.gs.auth.access.token.expiration";
-    private Configuration conf;
+  public static final String ACCESS_TOKEN_KEY = "fs.gs.auth.access.token.credential";
```
openinx, line 10: Since the `CredProsUtil` moved to `auth` package, so those two static fields need to be public.

openinx: The following changes are fixing the checkstyle validation.

connectors/spark/src/main/scala/io/unitycatalog/spark/CredPropsUtil.java
tdas: sounds good. please leave TODOs.

connectors/spark/src/main/scala/io/unitycatalog/spark/OptionUtil.java
```diff
@@ -0,0 +1,24 @@
...
+public class OptionUtil {
+  private OptionUtil() {
+  }
+
+  public static final String URI = "uri";
+  public static final String TOKEN = "token";
+  public static final String WAREHOUSE = "warehouse";
+
+  public static final String RENEW_CREDENTIAL_ENABLED = "renewCredential.enabled";
+  public static final boolean DEFAULT_RENEW_CREDENTIAL_ENABLED = true;
```
tdas: please take note that if we backport this to 0.3, the default should be false.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
tdas: sounds good.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
tdas: @openinx any thoughts?
i dont want this PR to break existing funcationality.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -0,0 +1,238 @@
...
+  private static Map<String, String> s3FixedCredProps(TemporaryCredentials tempCreds) {
+    AwsCredentials awsCred = tempCreds.getAwsTempCredentials();
+    return new S3PropsBuilder()
+        .set("fs.s3a.access.key", awsCred.getAccessKeyId())
+        .set("fs.s3a.secret.key", awsCred.getSecretAccessKey())
+        .set("fs.s3a.session.token", awsCred.getSessionToken())
+        .build();
+  }
+
+  private static Map<String, String> s3TableTempCredProps(
```
tdas: referencing this comment again about dup code

https://github.com/unitycatalog/unitycatalog/pull/1110#discussion_r2433525548

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -17,29 +21,33 @@ private UCHadoopConf() {
...
+  // FileSystems such as S3AFileSystem or AbfsFileSystem, the keys must include the "fs." prefix.
+  // See also: https://github.com/unitycatalog/unitycatalog/issues/1112.
+  public static final String UC_URI_KEY = "fs.unitycatalog.uri";
+  public static final String UC_TOKEN_KEY = "fs.unitycatalog.token";
 
   // Key representing a unique credential ID. It identifies a job-level credential for a specific
   // table, meaning that the same job–table combination shares the same credential. Cached
   // credentials are indexed by this key and are not reused across different jobs.
-  public static final String UC_CREDENTIALS_UID_KEY = "unitycatalog.credentials.uid";
+  public static final String UC_CREDENTIALS_UID_KEY = "fs.unitycatalog.credentials.uid";
```
tdas, line 34: might be cleaner to actually define a prefix variable for "fs.unitycatalog." and use it everywhere.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -17,29 +21,33 @@ private UCHadoopConf() {
       "fs.azure.init.sas.token.expired.time";
 
   // Keys for the UnityCatalog client.
-  public static final String UC_URI_KEY = "unitycatalog.uri";
-  public static final String UC_TOKEN_KEY = "unitycatalog.token";
+  // Note: Delta automatically filters out configuration keys without the "fs." prefix To ensure
```
tdas, line 24: is it Delta or Spark?
if parquet tables face problems with this .. then this is a Spark issue.. not a delta issue.

connectors/spark/src/main/scala/io/unitycatalog/spark/OptionUtil.java
openinx: That's true,  let me revert this back to false.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
openinx: Hi @tdas , thanks for the comment here.  I think it may be better to separate those two method, since the two methods have two different arguments: `tableId` and `tableOp` vs `path` and `pathOp`.  

If we make them in a unfied method, such as :

```java
 public static Map<String, String> createCredProps(
      boolean renewCredEnable,
      String scheme,
      String uri,
      String token,
      String tableId,
      TableOperation tableOp,
      String path,
      PathOperation pathOp,
      TemporaryCredentials tempCreds,
      boolean useTableCred,
  ) {
   ....    
  }
```

It may make the caller feel some kind of confuse, because for a `table` temp cred,   we actually don't know how to pass the `path` and `PathOperation` arguments. and vice versa.  Although we can dup the  implementation code, but it's hard for caller to distinguish the difference.  So that's why I prefer to make them into two different methods and make them into two code paths natually.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
openinx: Hi @tdas , thanks for the comment.  I think the current PR also support the `gs` and `abfss`/`abfs`. Please see the code: 
1. https://github.com/unitycatalog/unitycatalog/pull/1110/files#diff-ab4ecc349cf85b6952657fa3266eb21da199d6460e0ccf4a95518e76a5ec5f15R229-R235
2. https://github.com/unitycatalog/unitycatalog/pull/1110/files#diff-ab4ecc349cf85b6952657fa3266eb21da199d6460e0ccf4a95518e76a5ec5f15R204-R210

And the current PR only support the non-renew cred here, for further PR we will enable the renewal when enable them in the test. 

( I'm not sure if I misunderstood the comment or not,  please feel free to correct me if there is anything wrong).

connectors/spark/src/main/scala/io/unitycatalog/spark/OptionUtil.java
openinx: Thanks for the comment !

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
openinx: You mean defining a constant like static final String UC_PREFIX = "fs.unitycatalog", and then using UC_PREFIX + "<name>", right?

From my perspective, it’s actually clearer to define each key as a complete, flat string constant. That way, readers don’t need to mentally concatenate UC_PREFIX and the suffix to figure out the full key name.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
openinx: According to this issue investigation,  I think it's delta issue. https://github.com/unitycatalog/unitycatalog/issues/1112

openinx:
The broken test is unrelated to this PR: 
```java
Downloaded https://repo1.maven.org/maven2/org/checkerframework/checker-compat-qual/2.5.3/checker-compat-qual-2.5.3.jar
Downloaded https://repo1.maven.org/maven2/org/jline/jline/3.27.1/jline-3.27.1-jdk8.jar.sha512
Downloaded https://repo1.maven.org/maven2/org/scala-lang/scala-compiler/2.13.16/scala-compiler-2.13.16.jar
Downloaded https://repo1.maven.org/maven2/org/glassfish/hk2/hk2-api/3.0.6/hk2-api-3.0.6.jar
Downloaded https://repo1.maven.org/maven2/org/jline/jline/3.27.1/jline-3.27.1-jdk8.jar
Downloaded https://repo1.maven.org/maven2/org/rocksdb/rocksdbjni/9.8.4/rocksdbjni-9.8.4.jar
[info] compiling 73 Java sources to /home/runner/work/unitycatalog/unitycatalog/server/target/models/target/classes ...
[info] compiling 96 Java sources to /home/runner/work/unitycatalog/unitycatalog/target/clients/java/target/classes ...
Generated classpath file '/home/runner/work/unitycatalog/unitycatalog/target/classpath'
Generated classpath file '/home/runner/work/unitycatalog/unitycatalog/target/control/java/target/classpath'
Generated classpath file '/home/runner/work/unitycatalog/unitycatalog/server/target/controlmodels/target/classpath'
Generated classpath file '/home/runner/work/unitycatalog/unitycatalog/clients/python/target/classpath'
[info] done compiling
Generated classpath file '/home/runner/work/unitycatalog/unitycatalog/server/target/models/target/classpath'
[info] compiling 107 Java sources to /home/runner/work/unitycatalog/unitycatalog/server/target/classes ...
[info] done compiling
Generated classpath file '/home/runner/work/unitycatalog/unitycatalog/target/clients/java/target/classpath'
[info] compiling 21 Java sources to /home/runner/work/unitycatalog/unitycatalog/examples/cli/target/classes ...
[info] done compiling
Generated classpath file '/home/runner/work/unitycatalog/unitycatalog/examples/cli/target/classpath'
[info] done compiling
Generated classpath file '/home/runner/work/unitycatalog/unitycatalog/server/target/classpath'
[error] Missing org.scala-lang:scala2-sbt-bridge:2.13.16:compile
[error] (spark / scalaCompilerBridgeBinaryJar) Missing org.scala-lang:scala2-sbt-bridge:2.13.16:compile

>> Stopped the server
Error: Process completed with exit code 1.
```

Let's try it again.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/AwsVendedTokenProvider.java
tdas: So the integration test in this PR is not really testing refresh at all... its only testing that the first initial cred received from actually gets injected .. thats all. 

isnt it?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
tdas: you could have a private method that take the s3PropsBuilder as a parameter and returns the map
```
private Map<String, String> buildCommonS3Props(
   String uri,
   String token,
   TemporaryCredentials tempCreds,
   S3PropsBuilder builder) {

  builder.set(UCHadoopConf.S3A_CREDENTIALS_PROVIDER, AwsVendedTokenProvider.class.getName())
        .uri(uri)
        .token(token)
        .uid(UUID.randomUUID().toString())
        .set(UCHadoopConf.S3A_INIT_ACCESS_KEY, awsCred.getAccessKeyId())
        .set(UCHadoopConf.S3A_INIT_SECRET_KEY, awsCred.getSecretAccessKey())
        .set(UCHadoopConf.S3A_INIT_SESSION_TOKEN, awsCred.getSessionToken());
   // For the static credential case, nullable expiration time is possible.
    if (tempCreds.getExpirationTime() != null) {
      builder.set(UCHadoopConf.S3A_INIT_CRED_EXPIRED_TIME,
          String.valueOf(tempCreds.getExpirationTime()));
    }

    return builder.build();
}
```

and then this method can simply be 

```
S3PropsBuilder builder = new S3PropsBuilder()
builder.credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE)
        .tableId(tableId)
        .tableOperation(tableOp);

return buildCommonS3Props(...builder...);
```

isnt this code deduplication without introducing confusing public methods?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
tdas: got it . thanks for confirming.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
tdas: okay. no strong opinion

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/AwsVendedTokenProvider.java
openinx: Yes, that's true.  For this PR, the focus is making the `AwsVendedTokenProvider` to go through the whole running path, I mean enabling it to the local integration test, for convincing people that after successfully configuring the `renewCrednetial.enabled=true`, it won't break the job for the starting process. 

For the credential `renewal` process,  it will be addressed in this e2e test: https://github.com/unitycatalog/unitycatalog/pull/1119

I mean we will start a long-runing job (> 30min), and continuesly run and renew the credentials whenever it's required.

openinx: Also I can try to add the renewal local unit test,  like the `TableReadWriteTest`.  All renewal credentials will be issued from the OSS UC server.  Anyway I think it's better to have separate PR to address it, since the current PR have enough complex context to address.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
openinx: Thanks for the great point which I did not think of before.  I'd love to apply it in the PR.

openinx:
```
[error] Caused by: lmcoursier.internal.shaded.coursier.cache.ArtifactError$DownloadError: download error: Caught java.nio.channels.OverlappingFileLockException while downloading https://repo1.maven.org/maven2/org/scala-lang/scala2-sbt-bridge/2.13.16/scala2-sbt-bridge-2.13.16.jar
[error] 	at lmcoursier.internal.shaded.coursier.cache.internal.Downloader$.helper$2(Downloader.scala:793)
```

The CI reports the artifact downloading error,   let's retry it again.

openinx:
4 broken tests, let me take a look:

```
2025-10-20T21:55:45.2701915Z [error] Test io.unitycatalog.spark.TableReadWriteTest.testCredentialCreateDeltaTable failed: java.lang.IllegalArgumentException: Unexpected value 'null', took 5.335s
2025-10-20T21:55:47.5310440Z [error] Test io.unitycatalog.spark.TableReadWriteTest.testCredentialCreateDeltaTable failed: io.unitycatalog.client.ApiException: java.net.ConnectException, took 2.258s
2025-10-20T21:55:49.8700854Z [error] Test io.unitycatalog.spark.TableReadWriteTest.testCredentialCreateDeltaTable failed: io.unitycatalog.client.ApiException: java.net.ConnectException, took 2.337s
2025-10-20T21:55:52.2313042Z [error] Test io.unitycatalog.spark.TableReadWriteTest.testCreateExternalDeltaTable failed: io.unitycatalog.client.ApiException: java.net.ConnectException, took 2.359s
```

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
tdas: Aah .. i see. i think i got confused with not have `Fixed` in the function name for abfs/gs when s3 has `s3FixedCredProps`

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/OptionsUtil.java
```diff
@@ -0,0 +1,24 @@
+package io.unitycatalog.spark.utils;
+
+import java.util.Map;
+
+public class OptionsUtil {
+  private OptionsUtil() {
+  }
+
+  public static final String URI = "uri";
```
tdas, line 9: good to make a note that that is is not the full configuration key

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/OptionsUtil.java
```diff
@@ -0,0 +1,24 @@
+package io.unitycatalog.spark.utils;
+
+import java.util.Map;
+
+public class OptionsUtil {
```
tdas, line 5: These are configurations.. not options. Specifically configurations of the UCSingleCatalog.. right? 
So options is very generic name.
Better to name this SparkConfs or more specifically .. UCSingleCatalogConfs.

tdas (approved):
few minor comments on internal class names. please address them in follow up PRs. 
thank you for the integration test. i hope to see the improvements with actual credential renewal being tested in the integration test with the appropriate mocking framework.

PR #1114, opened 2025-10-15
Use the mock clock to test the credential expiration.

openinx:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [x] If there is a related issue, make sure it is linked to this PR.
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

The related issue is here: https://github.com/unitycatalog/unitycatalog/issues/1113

PR #1116, opened 2025-10-17
Retry request: update HadoopConf and GenericCredentialProvider

### LINKED ISSUE #1111: Retry the request to unity catalog server when any recoverable error happen in GenericCredentialProvider

Here, https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java#L88-L107 

I add the TODO before: 

I think we need the requests with retry policy to make the unity catalog credentials creation to be more robustness. 

Add the retry policy is enough.. Maybe also add a common **TemporaryCredentialsApi** factory to provide the general and common utility for all the unity catalog client access.

TimothyW553:
**PR Checklist**

**Description of changes**
Implements automatic retry with exponential backoff and jitter for Unity Catalog credential requests to handle transient failures (503, 429, 5xx, network errors). The retry behavior is configurable via unitycatalog.retry.* configuration keys with default parameters of (4 attempts, 500ms initial delay, 1.5x multiplier).

This improves reliability by automatically recovering from temporary server or network issues without failing the entire job.

Closes #1111

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
@@ -34,7 +34,7 @@ public class AwsVendedTokenProviderTest extends BaseTokenProviderTest<AwsVendedT
 
     TestAwsVendedTokenProvider(Configuration conf, TemporaryCredentialsApi tempCredApi) {
       super(conf);
-      this.retryableApi = new RetryableTemporaryCredentialsApi(tempCredApi, conf, clock);
+      this.retryableApi = new RetryableTemporaryCredentialsApi(tempCredApi, conf);
     }
 
     @Override
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/AbfsVendedTokenProviderTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/AbfsVendedTokenProviderTest.java
@@ -28,7 +28,7 @@ public class AbfsVendedTokenProviderTest extends BaseTokenProviderTest<AbfsVende
 
     TestAbfsVendedTokenProvider(Configuration conf, TemporaryCredentialsApi mockApi) {
       initialize(conf);
-      this.retryableApi = new RetryableTemporaryCredentialsApi(mockApi, conf, clock);
+      this.retryableApi = new RetryableTemporaryCredentialsApi(mockApi, conf);
     }
 
     @Override


connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -123,31 +132,82 @@ private GenericCredential renewCredential() throws ApiException {
...
     TemporaryCredentials tempCred;
     String type = conf.get(UCHadoopConf.UC_CREDENTIALS_TYPE_KEY);
-    // TODO We will need to retry the temporary credential request if any recoverable failure, for
-    // more robustness.
     if (UCHadoopConf.UC_CREDENTIALS_TYPE_PATH_VALUE.equals(type)) {
       String path = conf.get(UCHadoopConf.UC_PATH_KEY);
       String pathOperation = conf.get(UCHadoopConf.UC_PATH_OPERATION_KEY);
 
-      tempCred = tempCredApi.generateTemporaryPathCredentials(
+      tempCred = callWithRetry(() ->
```
openinx: Do we have approach to add the retry logic inside the ApiClient,  so that we don't need to add `callWithRetry` for every api request in the upper layer code ?   I mean it will be better to hide the retry code for the api caller, otherwise for every api we need to carefully review if this need to add `callWithRetry` or not.

openinx: I just ask the GPT,  and it suggest one approach.   This could be one solution.

Java

Generated clients typically use OkHttp or Apache HttpClient.
You can:

Configure OkHttpClient with a custom Interceptor that adds retry logic, e.g.:

```
OkHttpClient client = new OkHttpClient.Builder()
    .addInterceptor(new RetryInterceptor(3)) // custom retry logic
    .build();

ApiClient apiClient = new ApiClient();
apiClient.setHttpClient(client);
```

Or use libraries like Resilience4j or Spring Retry to wrap your API calls.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -10,6 +10,7 @@
 import io.unitycatalog.spark.ApiClientFactory;
 import io.unitycatalog.spark.UCHadoopConf;
 import java.net.URI;
+import java.util.function.Supplier;
```
openinx: A key task is :  we need to carefully design the unit test for it, so that we can feel confident to say that the retry design really solve the problem and improve the robustness.  Even if there is an accidently error from the OSS UC server,  the client can successfully recover from the failure via backoff retry.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -123,31 +132,82 @@ private GenericCredential renewCredential() throws ApiException {
     }
   }
 
+  private boolean isRecoverable(Throwable e) {
+    if (e instanceof ApiException) {
```
openinx: Where can I find those recoverable codes ?  And how could we confidently say that we've already covered all the recoverable code ?

openinx: Please also consider how to design test to cover those recoverable codes, thanks .

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -123,31 +132,82 @@ private GenericCredential renewCredential() throws ApiException {
...
+        if (!isRecoverable(e) || attempt == maxAttempts) {
+          break;
+        }
+
+        long baseDelay = (long) (initialDelay * Math.pow(multiplier, attempt - 1));
+        double jitter = (Math.random() - 0.5) * 2 * UCHadoopConf.RETRY_JITTER_FACTOR;
+        long delay = (long) (baseDelay * (1 + jitter));
+
+        try {
+          Thread.sleep(delay);
```
openinx: Maybe it will be better to use the abstracted clock, so that we can easily do the clock simulation for testing purposes. 

https://github.com/unitycatalog/unitycatalog/pull/1114/files

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
timothyw553: Got it. I've added a wrapper around the base TemporaryCredentialsApi which hides retries for all credential calls (scoped to only intercept credential calls, which we could extend to a global interceptor later if needed). The wrapper also injects Clock from #1114 for testability.

I added the lightweight wrapper to avoid adding more dependencies, adding dependencies will involve understanding the client's environment.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
timothyw553: Sounds good, changed it and made it a no-op for ManualClock.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,140 @@
...
+      return false;
+    }
+    
+    if (isNetworkException(e)) {
+      return true;
+    }
+    
+    // Check if the cause is a network exception (common with wrapped exceptions)
+    Throwable cause = e.getCause();
+    return cause != null && isNetworkException(cause);
```
timothyw553: First check is if exception is directly a network exception (e.g., SocketTimeoutException) and second check is for wrapped network exception -- RuntimeException that has a SocketTimeoutException as its cause.

connectors/spark/src/test/java/io/unitycatalog/spark/RetryableTemporaryCredentialsApiTest.java
```diff
@@ -0,0 +1,247 @@
...
+    conf = new Configuration();
+    conf.setInt(UCHadoopConf.RETRY_MAX_ATTEMPTS_KEY, UCHadoopConf.RETRY_MAX_ATTEMPTS_DEFAULT);
+    conf.setLong(UCHadoopConf.RETRY_INITIAL_DELAY_KEY, UCHadoopConf.RETRY_INITIAL_DELAY_DEFAULT);
+    conf.setDouble(UCHadoopConf.RETRY_MULTIPLIER_KEY, UCHadoopConf.RETRY_MULTIPLIER_DEFAULT);
+
+    delegate = Mockito.mock(TemporaryCredentialsApi.class);
+    initRetryableApi();
+  }
+
+  @Test
```
timothyw553, line 65: Quick summary on the tests:

- `testSuccessNoRetryNeededForPath`: Tests the happy path - when the API call succeeds on the first attempt, it returns the expected result with no retries and no sleep delays.

- `testRecoverableHttp503EventuallySucceeds`: Tests that HTTP 503 errors are retried automatically. The delegate fails twice with 503, then succeeds on the third attempt. Verifies the retry count and that backoff delays are within expected bounds.

- `testRecoverableErrorCodeFromJson`: Tests that specific error codes in JSON response bodies (like `TEMPORARILY_UNAVAILABLE`) trigger retries even if the HTTP status is 500. Fails once, then succeeds.

- `testNetworkExceptionRetries`: Tests that network-level exceptions (`SocketTimeoutException` and `SocketException`) are treated as retryable errors. Fails with two different network exceptions, then succeeds on the third attempt.

- `testNonRecoverableHttpStopsImmediately`: Tests that non-retryable HTTP errors (like 400 Bad Request) don't trigger retry logic. The error is thrown immediately without any retry attempts or delays.

- `testMaxAttemptsExhausted`: Tests that when all retry attempts are exhausted (all attempts result in 503), the final exception is thrown. Verifies it retried the maximum number of times.

- `testConfigurationOverridesAffectBehaviour`: Tests that custom configuration values (max attempts, initial delay, multiplier) are properly respected. Configures 5 max attempts with custom timing, then verifies the retry behavior matches.

- `testInvalidJsonFallbacksToStatus`: Tests that when the response body is invalid JSON, the retry logic falls back to using just the HTTP status code to determine if retry is needed. Retries on 503, stops on 404.

- `testZeroMaxAttemptsDisablesRetry`: Tests that setting max attempts to 1 effectively disables retry logic. Even with a retryable 503 error, it fails immediately without any retry attempts.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -96,12 +97,13 @@ public GenericCredential accessCredentials() {
...
 
-  protected TemporaryCredentialsApi temporaryCredentialsApi() {
+  protected RetryableTemporaryCredentialsApi temporaryCredentialsApi() {
     if (tempCredApi == null) {
       synchronized (this) {
         if (tempCredApi == null) {
-          tempCredApi = new TemporaryCredentialsApi(
+          TemporaryCredentialsApi baseApi = new TemporaryCredentialsApi(
               ApiClientFactory.createApiClient(ucUri, ucToken));
+          tempCredApi = new RetryableTemporaryCredentialsApi(baseApi, conf);
```
openinx: Nit: Maybe we can use the `RetryableTemporartyCrednetialApi` to construct the `tempCredApi` directly, without exposing the `TemporaryCredentialsApi` to the upper layer. I mean hidding the `TemporaryCredentialsApi` inside the `RetryableTemporaryCredentialApi`, since for the logic, people don't care much about the inside cnstruction.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -128,13 +130,11 @@ private GenericCredential renewCredential() throws ApiException {
...
 
   private GenericCredential createGenericCredentials() throws ApiException {
-    TemporaryCredentialsApi tempCredApi = temporaryCredentialsApi();
+    RetryableTemporaryCredentialsApi tempCredApi = temporaryCredentialsApi();
 
     // Generate the temporary credential via requesting UnityCatalog.
     TemporaryCredentials tempCred;
     String type = conf.get(UCHadoopConf.UC_CREDENTIALS_TYPE_KEY);
-    // TODO We will need to retry the temporary credential request if any recoverable failure, for
-    // more robustness.
```
openinx, line 132: Thanks for removing this TODO !

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/Clock.java
```diff
@@ -15,6 +15,13 @@ public interface Clock {
    */
   void advance(Duration duration);
 
+  /**
+   * Sleeps for the specified duration.
+   * In {@link SystemClock}, blocks via {@link Thread#sleep(long)}.
+   * In {@link ManualClock}, no-op—use {@link #advance(Duration)} to simulate time progression.
+   */
+  void sleep(Duration duration) throws InterruptedException;
```
openinx, line 19: What's the difference between `advance` and `sleep` ?  In the previous test, we just use the `advance` to simulate the `sleep` so that we can advance the current timetsamp.  But that `advance` was only used for  testing before, so I don't explicitly say it's `sleep`. 

Maybe we can just replace the `advance` with `sleep` directly, so that people don't get confusing about when to use `advance` (only supported by `MannualClock`) and when to use `sleep` (only supported by `SystemClock`).

Thought?

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/Clock.java
```diff
@@ -53,5 +65,10 @@ public synchronized Instant now() {
     public synchronized void advance(Duration duration) {
       now = now.plus(duration);
     }
+
+    @Override
+    public void sleep(Duration duration) {
+      // No-op for manual clock - time is advanced explicitly via advance()
```
openinx: That is the confusing point. For the interface, people need to know what's the implementation supporting `sleep` and `advance`.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,140 @@
+package io.unitycatalog.spark;
+
+import com.fasterxml.jackson.databind.JsonNode;
```
openinx, line 3: Do we need to shade the `fasterxml` ?  Since the OSS unity catalog connector is a JAR that will be embedded in user's runtime, which may conflicts with user's jar environment, and finally break the job.

We just need to know which compononents import this dep (seems like we already have this transitive dep, right ? ), and will it be safe to external users to run the spark job, avoide bother users with the lacking of thirdpart dep.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,140 @@
...
+
+  public RetryableTemporaryCredentialsApi(TemporaryCredentialsApi delegate, Configuration conf) {
+    this(delegate, conf, Clock.systemClock());
+  }
+
+  public RetryableTemporaryCredentialsApi(
+      TemporaryCredentialsApi delegate, Configuration conf, Clock clock) {
+    this.delegate = delegate;
+    this.clock = clock;
+    this.maxAttempts = conf.getInt(
```
openinx, line 62: Nit:  Maybe we will need some kind of abstracted and common retry policy class in future ( for supporting other api, except the TemporaryCredentialsApi) , but currently I think it should be fine for us to have this customized retry code here.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,140 @@
...
+        UCHadoopConf.RETRY_INITIAL_DELAY_KEY,
+        UCHadoopConf.RETRY_INITIAL_DELAY_DEFAULT
+    );
+    this.multiplier = conf.getDouble(
+        UCHadoopConf.RETRY_MULTIPLIER_KEY,
+        UCHadoopConf.RETRY_MULTIPLIER_DEFAULT
+    );
+  }
+
+  public TemporaryCredentials generateTemporaryPathCredentials(
```
openinx, line 86: Do you think we need to extend the `TemporaryCredentialsApi` for `RetryableTemporaryCredentialsApi` (so that we can explicitly override the `generateTemporaryPathCredentials` here), or not ?

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,140 @@
...
+import io.unitycatalog.client.model.GenerateTemporaryTableCredential;
+import io.unitycatalog.client.model.TemporaryCredentials;
+import io.unitycatalog.spark.utils.Clock;
+import java.time.Duration;
+import org.apache.hadoop.conf.Configuration;
+
+public class RetryableTemporaryCredentialsApi {
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  
+  private final TemporaryCredentialsApi delegate;
```
openinx, line 39: Nit:  Could you please import the [checkstyle](https://github.com/unitycatalog/unitycatalog/blob/main/dev/checkstyle-config.xml) into you intellij IDEA, and detect the checkstyle invalidation and fix it in advance ?  I see some of the checkstyle invalidation here. 

Image

Currently, I think our OSS-unitycatalog does not support the enforcement of checkstyle validation ( I mean enforce every newly created PR to match the checkstyle rules). I think we can file a separate PR to make the checkstyle to be enforced.

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/Clock.java
```diff
@@ -15,6 +15,13 @@ public interface Clock {
    */
   void advance(Duration duration);
 
+  /**
+   * Sleeps for the specified duration.
+   * In {@link SystemClock}, blocks via {@link Thread#sleep(long)}.
```
openinx: Nit: I don't think we should let the caller percieve the different implementation for the this abstracted method.  Maybe the `advance` implementation, is also a incorrect guiadance.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,140 @@
...
+          throw new RuntimeException("Retry interrupted", interrupted);
+        }
+      }
+    }
+
+    if (lastException instanceof ApiException) {
+      throw (ApiException) lastException;
+    } else {
+      throw new RuntimeException(
+          "Failed to obtain temporary credentials after " + maxAttempts + " attempts",
```
openinx: Nit: Maybe it's also better to include the elapsed time for the retry, so that we can locate the latency issue in further.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,140 @@
...
+    if (e instanceof ApiException) {
+      ApiException apiEx = (ApiException) e;
+      int code = apiEx.getCode();
+      if (code == 429 || code == 503) {
+        return true;
+      }
+      
+      String errorCode = extractUcErrorCode(apiEx.getResponseBody());
+      if (errorCode != null) {
+        return errorCode.equals("TEMPORARILY_UNAVAILABLE")
```
openinx: Thank you so much for finding those recoverable error code !  Could you please org them as a static and immutable hash set  ?  so that here we only need to check if the error code is hit in the set.  Maybe we will need to add more error codes in this hash set.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
openinx: Do we need to check the cause chain for it ?  I mean in the exception cause chain, every cause can be the network exception that we need to address, not necessarily the directly cause. 

Maybe: ` return Throwables.getCausalChain(e).stream().anyMatch(this::isNetworkException);`

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -42,4 +42,16 @@ private UCHadoopConf() {
   public static final String UC_CREDENTIAL_CACHE_ENABLED_KEY =
       "unitycatalog.credential.cache.enabled";
   public static final boolean UC_CREDENTIAL_CACHE_ENABLED_DEFAULT_VALUE = true;
+
+  // Keys for retry configuration.
+  public static final String RETRY_MAX_ATTEMPTS_KEY = "unitycatalog.retry.max.attempts";
```
openinx: We will need the `fs.` prefix, otherwise the delta will automatically filter all those keys and the underlaying filesystem won't correctly percieve those keys. For reason please see: https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java#L23-L27

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -42,4 +42,16 @@ private UCHadoopConf() {
...
+  public static final int RETRY_MAX_ATTEMPTS_DEFAULT = 4;
+
+  public static final String RETRY_INITIAL_DELAY_KEY = "unitycatalog.retry.initial.delay.ms";
```
openinx: Please follow the guidelines, about the configure keys naming: https://github.com/unitycatalog/unitycatalog/pull/1108#discussion_r2427004787

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
openinx: Nit: Maybe you could rebase to the lastest version of the main branch.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
```diff
@@ -128,4 +137,34 @@ public void testLoadProvider() throws IOException {
     assertThat(providers).hasSize(1);
     assertThat(providers.get(0)).isInstanceOf(AwsVendedTokenProvider.class);
   }
+
+  @Test
+  public void testRetryRecoversForTableCredentials() throws Exception {
```
openinx: Could you please to make it to be an astracted test method, so that it can cover all aws/gcs/azure credential logic ?  I think it will be helpful for us to provide a robust solution for all the cloud venders.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
```diff
@@ -128,4 +137,34 @@ public void testLoadProvider() throws IOException {
...
+    Clock.ManualClock manualClock = (Clock.ManualClock) Clock.manualClock(Instant.now());
+    List<Duration> recordedSleeps = new ArrayList<>();
+    Clock manualSpy = spy(manualClock);
+    Mockito.doAnswer(invocation -> {
```
openinx: If we unify the `advance` and `sleep` into a single method, then maybe we don't need to mock the `MannualClock#sleep` again ?  ---> maybe it will be better to change the `Clock` interface`.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsVendedTokenProviderTest.java
```diff
@@ -128,4 +137,34 @@ public void testLoadProvider() throws IOException {
...
+        })
+        .when(manualSpy)
+        .sleep(Mockito.any(Duration.class));
+
+    Configuration conf = newTableBasedConf();
+    TemporaryCredentialsApi tempCredApi = mock(TemporaryCredentialsApi.class);
+    TemporaryCredentials succeeded = newTempCred("success", manualClock.now().toEpochMilli() + 4000L);
+
+    when(tempCredApi.generateTemporaryTableCredentials(any()))
+        .thenThrow(new ApiException(503, "unavailable"))
```
openinx: Could you please do some kind of abstraction so that we can cover all the error codes that we want to test, rather than the specified error ?

connectors/spark/src/test/java/io/unitycatalog/spark/RetryableTemporaryCredentialsApiTest.java
```diff
@@ -0,0 +1,247 @@
...
+    TemporaryCredentials actual = retryableApi.generateTemporaryPathCredentials(
+        new GenerateTemporaryPathCredential().url("/test").operation(null));
+
+    assertThat(actual).isSameAs(expected);
+    verify(delegate, times(1)).generateTemporaryPathCredentials(any());
+    assertThat(recordedSleeps).isEmpty();
+  }
+
+  @Test
+  public void testRecoverableHttp503EventuallySucceeds() throws Exception {
```
openinx: Please help to do some abstraction to cover all the recoverable errors, rather than listing them one by one in a separate testing method, that will be more helpful for us to cover further new codes.

openinx (changes requested):

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
timothyw553: Yes, good idea

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
timothyw553: Yeah, I thought about this before and agree but for now this targeted implementation for credential retries is ok. If we need retry logic for other APIs in the future, we can extract a common retry policy class in the future.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
timothyw553: That's true, we're only checking one level currently. Noted for future: we should definitely check the entire cause chain.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
timothyw553: Noted and updated.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
timothyw553: I think the current way is more recommended since `TemporaryCredentialsApi` is auto-generated by OpenAPI and probably not designed for inheritance. If we did end up extending it, it would couple us to the implementation details – which could change on regeneration.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
timothyw553: Do you mean like `tempCredApi = new RetryableTemporaryCredentialsApi(ucUri, ucToken, conf);`? If so, yes I agree since then the caller doesn't need to know about TemporaryCredentialsAPI (as you said)

build.sbt
```diff
@@ -560,13 +560,20 @@ lazy val spark = (project in file("connectors/spark"))
     ),
     javaCheckstyleSettings(file("dev/checkstyle-config.xml")),
     Compile / compile / javacOptions ++= javacRelease11,
+    excludeDependencies ++= Seq(
```
timothyw553: Transitively block these dependencies coming in from uc-client. Spark already provides Jackson packages.

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/Clock.java
timothyw553: Yes, I think the having the two being basically opposite of each of is a bit confusing (`advance` only does something useful in `ManualClock`, and `sleep` only does something useful in `SystemClock`). I think we can merge them into a single `sleep(Duration)` operation where it actually blocks in `SystemClock` and just simulates sleep by advancing time in `ManualClock`.

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/Clock.java
timothyw553: Yes, we will only support one sleep(Duration) which works through Thread.sleep(...) in SystemClock and currentTime.plus(d) in ManualClock, and remove advance(...)

build.sbt
timothyw553: uc-spark depends on uc-client which depends on Jackson packages and this will be used in a Spark environment, which also has Jackson packages. Therefore, we will exclude the dependency on it from uc-client, mark it as provided (from Spark env) but ensure it is included for Tests

openinx:
Hi @TimothyW553 ,  seems like the unit test was broken, is this related or unrelated ?  Please help to confirm it, thanks.
The link is here: https://productionresultssa10.blob.core.windows.net/actions-results/37f6b6f4-6eb8-4769-be76-4616406554e0/workflow-job-run-615911bf-e780-595b-98fd-87da9b8abc02/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-10-28T16%3A23%3A33Z&sig=3N%2FxulrNauz9vl%2FslXJk0TomR54ln9%2ByMCum5a5k56I%3D&ske=2025-10-29T00%3A47%3A39Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-10-28T12%3A47%3A39Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2025-10-28T16%3A13%3A28Z&sv=2025-11-05

```
2025-10-24T22:48:11.8748111Z [error] Test io.unitycatalog.cli.table.CliTableCRUDTest.testTableCRUD failed: java.lang.RuntimeException: io.unitycatalog.client.ApiException: com.fasterxml.jackson.databind.Js
onMappingException: closed (through reference chain: io.unitycatalog.client.model.ListTablesResponse["tables"]->java.util.ArrayList[16]), took 11.805s
2025-10-24T22:48:11.8751174Z [error]     at io.unitycatalog.cli.UnityCatalogCli.main(UnityCatalogCli.java:171)
2025-10-24T22:48:11.8752407Z [error]     at io.unitycatalog.cli.TestUtils.executeCLICommand(TestUtils.java:24)
2025-10-24T22:48:11.8753735Z [error]     at io.unitycatalog.cli.table.CliTableOperations.listTables(CliTableOperations.java:70)
2025-10-24T22:48:11.8755496Z [error]     at io.unitycatalog.server.base.table.BaseTableCRUDTest.verifyTablePagination(BaseTableCRUDTest.java:116)
2025-10-24T22:48:11.8757128Z [error]     at io.unitycatalog.server.base.table.BaseTableCRUDTest.testTableCRUD(BaseTableCRUDTest.java:73)
2025-10-24T22:48:11.8758488Z [error]     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
2025-10-24T22:48:11.8760186Z [error]     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
2025-10-24T22:48:11.8761745Z [error]     at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
2025-10-24T22:48:11.8762993Z [error]     at java.lang.reflect.Method.invoke(Method.java:569)
2025-10-24T22:48:11.8764023Z [error]     ...
2025-10-24T22:48:11.8766264Z [error] Caused by: io.unitycatalog.client.ApiException: com.fasterxml.jackson.databind.JsonMappingException: closed (through reference chain: io.unitycatalog.client.model.ListTablesResponse["tables"]->java.util.ArrayList[16])
2025-10-24T22:48:11.8768809Z [error]     at io.unitycatalog.client.api.TablesApi.listTablesWithHttpInfo(TablesApi.java:352)
2025-10-24T22:48:11.8770390Z [error]     at io.unitycatalog.client.api.TablesApi.listTables(TablesApi.java:317)
2025-10-24T22:48:11.8771804Z [error]     at io.unitycatalog.cli.TableCli.listTables(TableCli.java:145)
2025-10-24T22:48:11.8773078Z [error]     at io.unitycatalog.cli.TableCli.handle(TableCli.java:45)
2025-10-24T22:48:11.8774262Z [error]     at io.unitycatalog.cli.UnityCatalogCli.main(UnityCatalogCli.java:136)
2025-10-24T22:48:11.8822425Z [error]     ... 76 more
2025-10-24T22:48:11.8824280Z [error] Caused by: com.fasterxml.jackson.databind.JsonMappingException: closed (through reference chain: io.unitycatalog.client.model.ListTablesResponse["tables"]->java.util.ArrayList[16])
2025-10-24T22:48:11.8826670Z [error]     at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:401)
2025-10-24T22:48:11.8828389Z [error]     at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:372)
2025-10-24T22:48:11.8830279Z [error]     at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer._deserializeFromArray(CollectionDeserializer.java:381)
2025-10-24T22:48:11.8832196Z [error]     at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:246)
2025-10-24T22:48:11.8834086Z [error]     at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:30)
2025-10-24T22:48:11.8836542Z [error]     at com.fasterxml.jackson.databind.deser.impl.MethodProperty.deserializeAndSet(MethodProperty.java:129)
2025-10-24T22:48:11.8838305Z [error]     at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:310)
2025-10-24T22:48:11.8839992Z [error]     at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:177)
2025-10-24T22:48:11.8841872Z [error]     at com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.readRootValue(DefaultDeserializationContext.java:342)
2025-10-24T22:48:11.8843650Z [error]     at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4931)
2025-10-24T22:48:11.8845081Z [error]     at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3913)
2025-10-24T22:48:11.8886852Z [error]     at io.unitycatalog.client.api.TablesApi.listTablesWithHttpInfo(TablesApi.java:347)
2025-10-24T22:48:11.8887816Z [error]     ... 80 more
2025-10-24T22:48:11.8888406Z [error] Caused by: java.io.IOException: closed
2025-10-24T22:48:11.8889821Z [error]     at jdk.internal.net.http.ResponseSubscribers$HttpResponseInputStream.current(ResponseSubscribers.java:447)
2025-10-24T22:48:11.8891482Z [error]     at jdk.internal.net.http.ResponseSubscribers$HttpResponseInputStream.read(ResponseSubscribers.java:502)
2025-10-24T22:48:11.8893053Z [error]     at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._loadMore(UTF8StreamJsonParser.java:220)
2025-10-24T22:48:11.8894576Z [error]     at com.fasterxml.jackson.core.json.UTF8StreamJsonParser.parseEscapedName(UTF8StreamJsonParser.java:2061)
2025-10-24T22:48:11.8896494Z [error]     at com.fasterxml.jackson.core.json.UTF8StreamJsonParser.slowParseName(UTF8StreamJsonParser.java:1969)
2025-10-24T22:48:11.8898134Z [error]     at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._parseName(UTF8StreamJsonParser.java:1757)
2025-10-24T22:48:11.8899755Z [error]     at com.fasterxml.jackson.core.json.UTF8StreamJsonParser.nextFieldName(UTF8StreamJsonParser.java:1040)
2025-10-24T22:48:11.8901435Z [error]     at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:317)
2025-10-24T22:48:11.8903016Z [error]     at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:177)
2025-10-24T22:48:11.8904707Z [error]     at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer._deserializeFromArray(CollectionDeserializer.java:361)
2025-10-24T22:48:11.8906091Z [error]     ... 89 more
2025-10-24T22:48:11.8907177Z [error] Caused by: java.net.ProtocolException: protocol error: Frame type(34) length(7217722) exceeds MAX_FRAME_SIZE(16384)
2025-10-24T22:48:11.8908642Z [error]     at jdk.internal.net.http.Http2Connection.protocolError(Http2Connection.java:1174)
2025-10-24T22:48:11.8909970Z [error]     at jdk.internal.net.http.Http2Connection.processFrame(Http2Connection.java:821)
```

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,178 @@
+package io.unitycatalog.spark;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Throwables;
```
openinx: Please be careful to import the 3rd-party deps such as `guava`, since for different guava , it does not provide the sufficient compatibility. 

For the `unitycatalog-spark-connector` usually we use the `import org.sparkproject.guava` . Please change to use that one, otherwise it may conflict with the runtime's deps.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import org.apache.hadoop.conf.Configuration;
+
+public class RetryableTemporaryCredentialsApi {
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  private static final Set<Integer> RECOVERABLE_HTTP_CODES =
```
openinx: Nit:  maybe it's better to use the guava's `ImmutableSet` directly here.

```
  // 429: Too Many Requests
  // 503: Service Unavailable
  private static final Set<Integer> RECOVERABLE_HTTP_CODES = ImmutableSet.of(429, 503);
```

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,178 @@
...
+      Collections.unmodifiableSet(new HashSet<>(List.of(
+          429,  // Too Many Requests
+          503   // Service Unavailable
+      )));
+
+  private static final Set<String> RECOVERABLE_ERROR_CODES =
```
openinx: Ditto.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,178 @@
...
+    this.initialDelayMs = conf.getLong(
+        UCHadoopConf.RETRY_INITIAL_DELAY_KEY,
+        UCHadoopConf.RETRY_INITIAL_DELAY_DEFAULT
+    );
+    this.multiplier = conf.getDouble(
+        UCHadoopConf.RETRY_MULTIPLIER_KEY,
+        UCHadoopConf.RETRY_MULTIPLIER_DEFAULT
+    );
+
+    if (this.maxAttempts < 1) {
```
openinx: Nit: Maybe we can use the `guava` directly.

```
    Preconditions.checkArgument(maxAttempts > 0,
        "Retry max attempts must be at least 1, got: %d (%s=%d)",
        this.maxAttempts, UCHadoopConf.RETRY_MAX_ATTEMPTS_KEY, this.maxAttempts);
```

openinx: The following are the same..

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,178 @@
...
+      throw new IllegalArgumentException(
+          String.format("Retry max attempts must be at least 1, got: %d (%s=%d)",
+              this.maxAttempts, UCHadoopConf.RETRY_MAX_ATTEMPTS_KEY, this.maxAttempts));
+    }
+    if (this.initialDelayMs < 0) {
```
openinx: The initial delay should be greater than 0, right ?

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  public TemporaryCredentials generateTemporaryTableCredentials(
+      GenerateTemporaryTableCredential request) throws ApiException {
+    return callWithRetry(() -> delegate.generateTemporaryTableCredentials(request));
+  }
+
+  private TemporaryCredentials callWithRetry(ApiCallSupplier apiCall) throws ApiException {
+    Exception lastException = null;
+    long startTime = System.currentTimeMillis();
```
openinx: Here should we use the `Clock.now()` or the default `System.currentTimeMillis()`, I think should be the former one, right ?

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,178 @@
...
+        try {
+          clock.sleep(Duration.ofMillis(delay));
+        } catch (InterruptedException interrupted) {
+          Thread.currentThread().interrupt();
+          throw new RuntimeException("Retry interrupted", interrupted);
+        }
+      }
+    }
+
+    long elapsedMs = System.currentTimeMillis() - startTime;
```
openinx: Use the `clock.now()` here ?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/BaseTokenProviderTest.java
```diff
@@ -13,25 +15,36 @@
...
 import org.apache.hadoop.conf.Configuration;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
 
 public abstract class BaseTokenProviderTest<T extends GenericCredentialProvider> {
 
-  /** Use the {@link Configuration} and the mocked api to create a new provider. */
+  /**
```
openinx: Maybe you can use the `./build/sbt javafmtAll` to reformat it ?

openinx: I think I need to change my codestyle to adjust the `javafmtAll` command, let me think how to configure the code style correctly.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/BaseTokenProviderTest.java
```diff
@@ -285,6 +298,37 @@ public void testGlobalCredCache() throws Exception {
     assertGlobalCache(4, tableACred2, tableBCred2, pathACred2, pathBCred2);
   }
 
+  @Test
+  public void testRetryRecoversForTableCredentials() throws Exception {
+    Clock.ManualClock manualClock = (Clock.ManualClock) Clock.manualClock(Instant.now());
+    List<Duration> recordedSleeps = new ArrayList<>();
+    Clock clockSpy = spy(manualClock);
+    Mockito.doAnswer(invocation -> {
```
openinx: Why do we still need to mock the `Clock` ?  I think we can directly use the `manualClock#sleep` to advance the global time clock,right ?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/BaseTokenProviderTest.java
```diff
@@ -285,6 +298,37 @@ public void testGlobalCredCache() throws Exception {
...
+      .sleep(Mockito.any(Duration.class));
+
+    Configuration conf = newTableBasedConf();
+    TemporaryCredentialsApi tempCredApi = mock(TemporaryCredentialsApi.class);
+    TemporaryCredentials succeeded = newTempCred("success",
+        manualClock.now().toEpochMilli() + 4000L);
+
+    when(tempCredApi.generateTemporaryTableCredentials(any()))
+        .thenThrow(new ApiException(503, "unavailable"))
+        .thenThrow(new ApiException(503, "unavailable"))
```
openinx, line 333: Maybe it's better to add the `429` here also ?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/BaseTokenProviderTest.java
openinx: There is an example here: https://github.com/unitycatalog/unitycatalog/pull/1123/files#diff-fae53ed54b97a94aa0f26b1720e854c670cc4e6738641f242b981915f065c0a2R209, we actually use the `manualClock` to accurately control the global clock, so that we can confidently say the credential renwal is working.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/BaseTokenProviderTest.java
```diff
@@ -285,6 +298,37 @@ public void testGlobalCredCache() throws Exception {
...
+    })
+    .when(clockSpy)
+      .sleep(Mockito.any(Duration.class));
+
+    Configuration conf = newTableBasedConf();
+    TemporaryCredentialsApi tempCredApi = mock(TemporaryCredentialsApi.class);
+    TemporaryCredentials succeeded = newTempCred("success",
+        manualClock.now().toEpochMilli() + 4000L);
+
+    when(tempCredApi.generateTemporaryTableCredentials(any()))
```
openinx, line 314: How about the path-based temp credential ?

connectors/spark/src/test/java/io/unitycatalog/spark/RetryableTemporaryCredentialsApiTest.java
```diff
@@ -0,0 +1,311 @@
...
+          .isGreaterThanOrEqualTo(minDelay)
+          .isLessThanOrEqualTo(maxDelay);
+    }
+  }
+
+  private void initRetryableApi() throws InterruptedException {
+    manualClock = (Clock.ManualClock) Clock.manualClock(Instant.now());
+    recordedSleeps = new ArrayList<>();
+    Clock clockSpy = Mockito.spy(manualClock);
+    Mockito.doAnswer(invocation -> {
```
openinx: We don't need to mock this clock anymore.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
timothyw553: Ah got it. I wasn't familiar with that. Thanks!

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
timothyw553: Yeah that's true, we should centralize/delegate time operations to Clock

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
timothyw553: Yes, that's correct. If initial delay is 0 then it won't increase.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/BaseTokenProviderTest.java
timothyw553: yeah, chaining different requests would be better for testing.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/BaseTokenProviderTest.java
timothyw553: Ah yes

build.sbt
openinx: Could you please revert those changes firstly in this PR ? Since if the `uc-client` will finally introduce the `jackson` deps,  then I will suggest to open a separate issue to relocate all of those deps, and also help to review all those transitive packages that we may introduce, which should be seriously considered so that we won't break the user's real  jar environment.  Thanks.

openinx: Opened a new issue here: https://github.com/unitycatalog/unitycatalog/issues/1141,  @TimothyW553  feel free to assign to you.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
```diff
@@ -0,0 +1,168 @@
...
+import io.unitycatalog.client.model.GenerateTemporaryTableCredential;
+import io.unitycatalog.client.model.TemporaryCredentials;
+import io.unitycatalog.spark.utils.Clock;
+import java.net.URI;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Set;
+import org.apache.hadoop.conf.Configuration;
+
+public class RetryableTemporaryCredentialsApi {
```
openinx, line 19: Do you think we also need to use this class in the `UCSingleCatalog`  ?  So that for the credential initial,  it will also enable the retry logics.  Please help to file a separate issue for that.

https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala#L38

openinx: Or, you can just change it in this PR...

openinx (commented):
I just left two minor comments.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryableTemporaryCredentialsApi.java
timothyw553: Since this PR is quite large, I plan to make a separate PR to address the other instances of TemporaryCredentialsAPI

build.sbt
timothyw553: Yeah, sounds good. Definitely don't want run into conflicts due to some package-environment issues.

PR #1119, opened 2025-10-20
Add real production credential renwal integration test.

openinx:
**PR Checklist**

**Description of changes**

For this issue, https://github.com/unitycatalog/unitycatalog/issues/1092.  Since the credential may be expired for 30minutes,  so in this PR, I  designed a streaming job which will redirect the stream into the designated table in 40 min.  And if all the rows are correct, and the streaming job is not been broken,  then the credential renewal should be fine.

### DIFF
--- a/integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
+++ b/integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
@@ -3,7 +3,6 @@ package io.unitycatalog.integrationtests;
 import static io.unitycatalog.integrationtests.TestUtils.AUTH_TOKEN;
 import static io.unitycatalog.integrationtests.TestUtils.CATALOG_NAME;
 import static io.unitycatalog.integrationtests.TestUtils.S3_BASE_LOCATION;
-import static io.unitycatalog.integrationtests.TestUtils.SCHEMA_NAME;
 import static io.unitycatalog.integrationtests.TestUtils.SERVER_URL;
 import static io.unitycatalog.integrationtests.TestUtils.envAsBoolean;
 import static io.unitycatalog.integrationtests.TestUtils.envAsLong;
@@ -13,16 +12,19 @@ import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 import io.unitycatalog.spark.UCSingleCatalog;
 import java.util.List;
 import java.util.UUID;
+import org.apache.spark.api.java.function.MapPartitionsFunction;
+import org.apache.spark.sql.Encoders;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.SparkSession;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.sparkproject.guava.collect.Iterators;
 
 /**
  * Typically, cloud vendor credentials (used to access cloud storage) issued by the UC server expire
- * after one hour. This integration test launches a long-running write job whose duration
- * (controlled by the {@link SparkCredentialRenewalTest#ROW_COUNT} environment variable) exceeds the
+ * after one hour. This integration test launches a long-running read-write job whose duration
+ * (controlled by the 'CREDENTIAL_RENEWAL_TEST_DURATION_SECONDS' environment variable) exceeds the
  * credential’s expiration time. If the job completes successfully, it verifies that the credential
  * renewal mechanism works as expected, seamlessly refreshing credentials without interrupting the
  * ongoing job.
@@ -35,7 +37,6 @@ import org.junit.jupiter.api.Test;
  * export CATALOG_URI=...
  * export CATALOG_AUTH_TOKEN=...
  * export CATALOG_NAME=...
- * export SCHEMA_NAME=...
  * export S3_BASE_LOCATION=...
  * SBT_OPTS="-Xmx8G -XX:+UseG1GC" \
  * ./build/sbt \
@@ -48,14 +49,16 @@ public class SparkCredentialRenewalTest {
   // Define the CREDENTIAL_RENEWAL_TEST_RENEWAL_ENABLED environment variable.
   private static final boolean RENEW_CRED_ENABLED = envAsBoolean(PREFIX + "RENEWAL_ENABLED", true);
 
-  // Define the CREDENTIAL_RENEWAL_TEST_ROW_COUNT environment variable. Usually it will take
-  // about 2 hour to process 10^10 rows, which will exceeds the credential's expiration time
-  // (1hour). And finally trigger the credential renewal.
-  private static final long ROW_COUNT = envAsLong(PREFIX + "ROW_COUNT", 10_000_000_000L);
+  // Define the CREDENTIAL_RENEWAL_TEST_DURATION_SECONDS environment variable, which control how
+  // long will the test run. 90 minutes by default.
+  private static final long DURATION_SECONDS = envAsLong(PREFIX + "DURATION_SECONDS", 5600L);
 
-  // Define the table name.
-  private static final String TABLE_NAME =
-      String.format("%s.%s.%s", CATALOG_NAME, SCHEMA_NAME, randomName());
+  // Define the schema name and table name.
+  private static final String SCHEMA_NAME = randomName();
+  private static final String SRC_TABLE =
+      String.format("%s.%s.src_%s", CATALOG_NAME, SCHEMA_NAME, randomName());
+  private static final String DST_TABLE =
+      String.format("%s.%s.dst_%s", CATALOG_NAME, SCHEMA_NAME, randomName());
 
   private static SparkSession spark;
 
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/java/io/unitycatalog/integrationtests/BaseSparkTest.java
```diff
@@ -1,85 +1,90 @@
...
 import java.util.Objects;
+import lombok.Getter;
+import lombok.SneakyThrows;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import org.junit.jupiter.api.BeforeAll;
 
 public class BaseSparkTest {
-    // todo: parameterize such that catalogs can be specified per cloud provider if desired
-    private static final String ServerUrl = System.getenv().getOrDefault("CATALOG_URI", "http://localhost:8080");
```
openinx: Move those env variables into the `TestUtils` class.

integration-tests/src/test/java/io/unitycatalog/integrationtests/BaseSparkTest.java
```diff
@@ -1,85 +1,90 @@
...
-    @BeforeAll
-    public static void setup() {
-        spark = createSparkSessionWithCatalogs(CatalogName);
-    }
+  @BeforeAll
+  public static void setup() {
+    spark = createSparkSessionWithCatalogs(TestUtils.CATALOG_NAME);
+  }
 
-    protected static SparkSession createSparkSessionWithCatalogs(String... catalogs) {
```
openinx: Those changes are reformating the code and fix the checkstyle enforcement issue.  The integrationTest module does not enforce the check, after this PR https://github.com/unitycatalog/unitycatalog/pull/1128/files#diff-111489e5509a844e2edbc4634e69c496f24e54c4e844063f65717be16e538a2a. Because there's no classes in the integrationTest module,  and we don't run the integration test for this module for every PR trigger, so we have no chance to trigger the `compileJavastyle` and `testJavastyle`. 

May need a small & separate PR to enforce it , even it does not have the `main` folder.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Typically, cloud vendor credentials (used to access cloud storage) issued by the UC server
+ * expire after one hour. This integration test launches a long-running write job whose duration
+ * (controlled by the {@link SparkCredentialRenewalTest#ROW_COUNT} environment variable) exceeds
+ * the credential’s expiration time. If the job completes successfully, it verifies that the
+ * credential renewal mechanism works as expected, seamlessly refreshing credentials without
+ * interrupting the ongoing job.
+ */
+public class SparkCredentialRenewalTest {
```
openinx, line 46: This is the credential renewal long-running job. We can run this command to do the integrtation test. 

```java
env \
  CATALOG_URI="$UC_URI" \
  CATALOG_AUTH_TOKEN="$UC_TOKEN" \
  CATALOG_NAME="main" \
  S3_BASE_LOCATION="$UC_S3_LOCATION" \
  SBT_OPTS="-Xmx8G -XX:+UseG1GC" \
  "$PWD/build/sbt" "integrationTests/testOnly io.unitycatalog.integrationtests.SparkCredentialRenewalTest"
```

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -0,0 +1,87 @@
...
+  private static final String PREFIX = "CREDENTIAL_RENEWAL_TEST_";
+
+  // Define the CREDENTIAL_RENEWAL_TEST_RENEWAL_ENABLED environment variable.
+  private static final boolean RENEW_CRED_ENABLED = envBoolean(PREFIX + "RENEWAL_ENABLED", true);
+
+  // Define the CREDENTIAL_RENEWAL_TEST_ROW_COUNT environment variable.
+  private static final long ROW_COUNT = envLong(PREFIX + "ROW_COUNT", 10_000_000_000L);
```
openinx: Set this to `10^10`, the test will be running for 1.5 hours, which trigger several rounds of the credential renewal.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -0,0 +1,87 @@
...
+      CATALOG_NAME, SCHEMA_NAME, randomName());
+
+  private static SparkSession spark;
+
+  @BeforeAll
+  public static void beforeAll() {
+    String testCatalogKey = String.format("spark.sql.catalog.%s", CATALOG_NAME);
+    spark = SparkSession.builder()
+        .appName("test-credential-renewal-in-long-running-job")
+        .master("local[1]") // Make it single-threaded explicitly.
```
openinx: single-thread,  make it run slowly, without exhausting the host's CPU resources.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -0,0 +1,87 @@
...
+  @AfterAll
+  public static void afterAll() {
+    sql("DROP TABLE IF EXISTS %s", TABLE_NAME);
+    sql("DROP SCHEMA IF EXISTS %s", SCHEMA_NAME);
+    spark.stop();
+  }
+
+  @Test
+  public void testLongRunningJob() {
+    sql("INSERT INTO %s SELECT id, CONCAT('val_', id) AS val FROM range(0, %s)",
```
openinx: The whole job is a simple writing job.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkDeltaTableCRUDTest.java
```diff
@@ -10,126 +14,120 @@
...
-        spark.sql(format("UPDATE %s SET as_double = as_int * 1.5 WHERE as_int = 42", table));
-
-        assertThat(getData(table))
-                .as("Data after UPDATE")
-                .isEqualTo(List.of(
-                        "{\"as_int\":42,\"as_double\":63.0}",
-                        "{\"as_int\":539,\"as_double\":425.66102859000944}"
-                ));
-    }
+  private static final String BASE_TABLE_NAME = "numbers";
```
openinx, line 19: This file is similar to the comment, just reformat the code to fix the checkstyle report.
https://github.com/unitycatalog/unitycatalog/pull/1119/files#r2480136528

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkParquetTableCRUDTest.java
```diff
@@ -12,109 +17,103 @@
...
-    @MethodSource("locationTypes")
-    @Order(5)
-    public void update(LocationType locationType) {
-        String table = getTableName(locationType);
-
-        assertThatThrownBy(() -> spark.sql(format("UPDATE %s SET as_double = as_int * 1.5 WHERE as_int = 42", table)))
-                .isInstanceOf(SparkUnsupportedOperationException.class)
-                .hasMessageContaining("UPDATE TABLE is not supported temporarily");
-    }
+  private static final String BASE_TABLE_NAME = "numbers";
```
openinx, line 22: Ditto.

build.sbt
```diff
@@ -636,10 +638,9 @@ lazy val integrationTests = (project in file("integration-tests"))
       "org.projectlombok" % "lombok" % "1.18.32" % Provided,
       "org.apache.spark" %% "spark-sql" % sparkVersion % Test,
       "io.delta" %% "delta-spark" % deltaVersion % Test,
-      "org.apache.hadoop" % "hadoop-aws" % "3.3.6" % Test,
-      "org.apache.hadoop" % "hadoop-azure" % "3.3.6" % Test,
+      "org.apache.hadoop" % "hadoop-aws" % hadoopVersion % Test,
```
openinx, line 644: Before this PR,  we use the `hadoop 3.3.6`  to run the integration tests by default.  while actually,  I encountered a serious issue here https://github.com/unitycatalog/unitycatalog/issues/1092#issuecomment-3470427971.   

I think I will need a separate PR to address the cross-hadoop version issue.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: And if we set the `CREDENTIAL_RENEWAL_TEST_RENEWAL_ENABLED=false` in this test,  then the job will be failure.

integration-tests/src/test/java/io/unitycatalog/integrationtests/BaseSparkTest.java
tdas: then move the TODO as well.

integration-tests/src/test/java/io/unitycatalog/integrationtests/BaseSparkTest.java
tdas: please dont make format changes in this PR. if needed make a separate PR for all the formatting changes.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -0,0 +1,87 @@
...
+        .config("spark.sql.catalog.spark_catalog",
+            "org.apache.spark.sql.delta.catalog.DeltaCatalog")
+        .config(testCatalogKey, UCSingleCatalog.class.getName())
+        .config(testCatalogKey + ".uri", SERVER_URL)
+        .config(testCatalogKey + ".token", AUTH_TOKEN)
+        .config(testCatalogKey + ".warehouse", CATALOG_NAME)
+        .config(testCatalogKey + ".renewCredential.enabled", String.valueOf(RENEW_CRED_ENABLED))
+        .getOrCreate();
+
+    sql("CREATE SCHEMA %s", SCHEMA_NAME);
```
tdas: this will fail if the schema exists.. right? might not be the behavior we want when we point to a different UC.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
tdas: thsi should be in the class docs as instructions on how to run this manually. not in comments.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.List;
+import java.util.UUID;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Typically, cloud vendor credentials (used to access cloud storage) issued by the UC server
```
tdas: this test.. by default runs very fast (as it should as its always enabled) and so it really does not test anything by default... right? 

only when you manually configure it to run for long it tests something.

integration-tests/src/test/java/io/unitycatalog/integrationtests/BaseSparkTest.java
openinx: Because we've enforced the checkstyle check, so if we don't fix this, then the integration cannot run...

integration-tests/src/test/java/io/unitycatalog/integrationtests/BaseSparkTest.java
openinx: Sounds good, to me. Done

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
tdas: this is important context that needs to be written down.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
tdas: shouldnt we test both reads and writes?
basically.. why not follow the exactly strategy that you used in the earlier PR we merged? 
you just have create multiple files in the same way, it will test both reads, writes, multiple tables, etcl. 
the only thing you need to change is instead of manual clock advance.. you can simply Thread.sleep(1000). 
by default it can be 1 files with 1 rows that runs in 1 second. then the setting simply becomes how many seconds do you want the code to run

integration-tests/src/test/java/io/unitycatalog/integrationtests/BaseSparkTest.java
tdas: then please make a separate previous PR that fixes all the formatting

build.sbt
```diff
@@ -622,6 +623,7 @@ lazy val spark = (project in file("connectors/spark"))
 
 lazy val integrationTests = (project in file("integration-tests"))
   .enablePlugins(CheckstylePlugin)
+  .dependsOn(spark)
```
openinx, line 626: it will be better to use the cross module depending here than configured the static `io.unitycatalog:unitycatalog-spark` here. https://github.com/unitycatalog/unitycatalog/pull/1119/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91L645

Since for the later one, we need to manually publish the `unitycatalog-spark` into local maven repo, which will makes the `integrationTests` more hard to run. 

With the former one,  if we make any changes in the `unitycatalog-spark` connector, then we can simply use the command `./build/sbt integrationTests/test` to run the `integrationTests` without publishing.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Yes,  it's good to add in class docs. Done !

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Good point. Yes it will fail.  I think it's better to use a `random` schema name, rather than the specified name via environment, otherwise it's hard for us to determine if we need to drop it or not. 

By using the `random` schema,   we can just create our own and after the test just drop it.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -0,0 +1,112 @@
...
+    sql("CREATE SCHEMA %s", SCHEMA_NAME);
+    sql(
+        "CREATE TABLE %s (id BIGINT, val STRING) USING delta LOCATION '%s/%s'",
+        TABLE_NAME, S3_BASE_LOCATION, UUID.randomUUID());
+  }
+
+  @AfterAll
+  public static void afterAll() {
+    sql("DROP TABLE IF EXISTS %s", TABLE_NAME);
+    sql("DROP SCHEMA IF EXISTS %s", SCHEMA_NAME);
```
openinx: Related to this [comment](https://github.com/unitycatalog/unitycatalog/pull/1119/files#r2480174048)

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: By default,  we use the row count as `10^10` (https://github.com/unitycatalog/unitycatalog/pull/1119/files#diff-a961fbeeb181232588b1a8b7657dff464d114c46faf6b4fd355903115601efe5R54),  so it will take roughly 2 hours to finish all the tests.  While if people want to do some initial testing to be confident that everything for the testing environment and configuration is fine, then it will be better to specifty the environment variable `CREDENTIAL_RENEWAL_TEST_ROW_COUNT` to smaller number, such as 10.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Good point, done !

Issue #1121, opened 2025-10-20
Add local aws credential renewal integration test


openinx:
In this comment https://github.com/unitycatalog/unitycatalog/pull/1110/files#r2446034709, we left one task.  we will need to add a local integration test to address the `AwsVendedTokenProvider` will indeed refresh the credentials automatically. 

I think we will need to leverage the OSS UC server's credential issurance ability to help us to finish the credential renewal. Also we may need to improve the `S3CredentialTestFileSystem`, so that we can assert the old and renewal credential step by step.

Let me create a spareate PR for this issue.

openinx:
When I designed the streaming case, I found this issue: 

```
5/10/21 16:06:19 INFO UnityCatalogServer: Unity Catalog server stopped.

org.apache.spark.sql.AnalysisException: The input source(delta) is different from the table `renewal`.`default`.`demo`'s data source provider(DELTA).

	at org.apache.spark.sql.errors.QueryCompilationErrors$.inputSourceDiffersFromDataSourceProviderError(QueryCompilationErrors.scala:2265)
	at org.apache.spark.sql.classic.DataStreamWriter.writeToV1Table$1(DataStreamWriter.scala:197)
	at org.apache.spark.sql.classic.DataStreamWriter.toTable(DataStreamWriter.scala:208)
	at io.unitycatalog.spark.auth.AwsCredentialRenewalTest.testRenewal(AwsCredentialRenewalTest.java:144)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```

I think here we need to make it being a lower case.  

https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala#L279

openinx:
I found another issue. Currently,  If I write the following streaming job. 

```java
String location =
        "s3://test-bucket0" + (new File(dataDir, "renewal.default.demo").getCanonicalPath());
    String table = "`renewal`.`default`.`demo`";

    // Create the external delta path-based table.
    S3CredRenewFileSystem.credentialCheckEnabled = false;
    sql("CREATE TABLE delta.`%s` (id INT) USING delta", location);
    S3CredRenewFileSystem.credentialCheckEnabled = true;

    // Create the catalog table referring to external table.
[...]

PR #1123, opened 2025-10-21
Add local aws credential renewal integration test.

openinx:
**PR Checklist**

**Description of changes**

Related issue is: https://github.com/unitycatalog/unitycatalog/issues/1121

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
@@ -26,7 +26,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
@@ -35,10 +34,6 @@ import org.apache.spark.sql.Encoders;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.RowFactory;
 import org.apache.spark.sql.SparkSession;
-import org.apache.spark.sql.types.DataTypes;
-import org.apache.spark.sql.types.Metadata;
-import org.apache.spark.sql.types.StructField;
-import org.apache.spark.sql.types.StructType;
 import org.apache.spark.util.SerializableConfiguration;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
@@ -46,10 +41,23 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 import org.sparkproject.guava.collect.ImmutableList;
+import org.sparkproject.guava.collect.Iterators;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
 import software.amazon.awssdk.services.sts.model.Credentials;
 
+/**
+ * Integration test to verify that AWS credential renewal works as expected.
+ *
+ * <p>This test sets up the Unity Catalog server and a local Spark cluster, then runs a custom Spark
+ * job to validate credential renewal.
+ *
+ * <p>The approach is as follows: a testing {@link AwsCredentialsProvider} is injected into the
+ * local Unity Catalog server, issuing a new credential every 30-second interval. The Spark job uses
+ * {@link S3CredFileSystem} to check that the credential matches the current time window for each
+ * filesystem access. {@link S3CredFileSystem} also tracks the number of renewals, allowing
+ * verification that credential renewal occurs as expected.
+ */
 public class AwsCredentialRenewalTest extends BaseCRUDTest {
 
   private static final String CLOCK_NAME = AwsCredentialRenewalTest.class.getSimpleName();
@@ -78,6 +86,9 @@ public class AwsCredentialRenewalTest extends BaseCRUDTest {
     serverProperties.put("s3.accessKey.0", "accessKey0");
     serverProperties.put("s3.secretKey.0", "secretKey0");
     serverProperties.put("s3.sessionToken.0", "sessionToken0");
+    // Customize the test credential generator to issue a new credential every 30-second interval.
+    // This allows us to verify whether credential renewal is functioning correctly by checking
+    // if the current credential matches the expected time window.
     serverProperties.put("s3.credentialsGenerator.0", CREDENTIALS_GENERATOR);
   }
 
@@ -159,13 +170,8 @@ public class AwsCredentialRenewalTest extends BaseCRUDTest {
     String location = String.format("s3://%s%s/fs", BUCKET_NAME, dataDir.getCanonicalPath());
     Path locPath = new Path(location);
 
-    // Create the external delta path-based table.
-    S3CredFileSystem.credentialCheckEnabled = false;
-    sql("CREATE TABLE delta.`%s` (id INT) USING delta", location);
-    S3CredFileSystem.credentialCheckEnabled = true;
-
-    // Create the catalog table referring to external table.
-    sql("CREATE TABLE %s USING delta LOCATION '%s'", TABLE_NAME, location);
+    // Create the external delta table in catalog.
+    sql("CREATE TABLE %s (id INT) USING delta LOCATION '%s'", TABLE_NAME, location);
 
     // Insert 1 row into the table.
     sql("INSERT INTO %s VALUES (1)", TABLE_NAME);
@@ -223,26 +229,19 @@ public class AwsCredentialRenewalTest extends BaseCRUDTest {
 
   @Test
   public void testDeltaReadWriteRenewal() throws Exception {
-    String location = String.format("s3://%s%s/delta", BUCKET_NAME, dataDir.getCanonicalPath());
-    StructType schema =
-        new StructType(
-            new StructField[] {
-              new StructField("id", DataTypes.IntegerType, false, Metadata.empty()),
-              new StructField("partition", DataTypes.IntegerType, false, Metadata.empty()),
-            });
+    String srcLoc = String.format("s3://%s%s/src", BUCKET_NAME, dataDir.getCanonicalPath());
+    String dstLoc = String.format("s3://%s%s/dst", BUCKET_NAME, dataDir.getCanonicalPath());
+    String srcTable = String.format("%s.%s.src", CATALOG_NAME, SCHEMA_NAME);
+    String dstTable = String.format("%s.%s.dst", CATALOG_NAME, SCHEMA_NAME);
 
-    // Create the external delta path-based table.
-    S3CredFileSystem.credentialCheckEnabled = false;
-    sql("CREATE TABLE delta.`%s` (id INT) USING delta PARTITIONED BY (partition INT)", location);
-    S3CredFileSystem.credentialCheckEnabled = true;
+    // Create the source table referring to external table.
+    sql(
+        "CREATE TABLE %s (id INT) USING delta LOCATION '%s' PARTITIONED BY (partition INT)",
+        srcTable, srcLoc);
+    sql("CREATE TABLE %s (id INT) USING delta LOCATION '%s'", dstTable, dstLoc);
 
-    // Create the catalog table referring to external table.
-    sql("CREATE TABLE %s USING delta LOCATION '%s'", TABLE_NAME, location);
-
-    // Insert 4 rows into each partition where (id % 3) equals the partition key.
-    sql("INSERT INTO %s PARTITION (partition=1) VALUES (1), (4), (7), (10)", TABLE_NAME, 1);
-    sql("INSERT INTO %s PARTITION (partition=2) VALUES (2), (5), (8), (11)", TABLE_NAME, 2);
-    sql("INSERT INTO %s PARTITION (partition=3) VALUES (3), (6), (9), (12)", TABLE_NAME, 3);
+    // Insert 1 row into each partition where id equals the partition key.
+    sql("INSERT INTO %s VALUES (1, 1), (2, 2), (3, 3)", srcTable);
 
     // Read from the Delta table, mapping each partition to a separate task.
     // With parallelism set to 1, tasks for each partition execute sequentially.
@@ -253,28 +252,25 @@ public class AwsCredentialRenewalTest extends BaseCRUDTest {
     session
         .read()
         .format("delta")
-        .table(TABLE_NAME)
+        .table(srcTable)
         .mapPartitions(
-            (MapPartitionsFunction<Row, Row>)
+            (MapPartitionsFunction<Row, Integer>)
                 input -> {
                   // Advance the clock to trigger the credential renewal.
                   testClock().advance(Duration.ofMillis(DEFAULT_INTERVAL_MILLIS));
-                  return input;
+                  return Iterators.transform(input, row -> row.getInt(0));
                 },
-            Encoders.row(schema))
+            Encoders.INT())
+        .withColumnRenamed("value", "id")
         .write()
         .format("delta")
         .mode("append")
-        .saveAsTable(TABLE_NAME);
+        .saveAsTable(dstTable);
 
-    List<Row> rows = sql("SELECT * FROM %s ORDER BY id ASC", TABLE_NAME);
-    assertThat(rows.size()).isEqualTo(24);
+    List<Row> rows = sql("SELECT * FROM %s ORDER BY id ASC", dstTable);
+    assertThat(rows.size()).isEqualTo(3);
     assertThat(rows.stream().map(r -> r.getInt(0)).collect(Collectors.toList()))
-        .isEqualTo(
-            IntStream.concat(IntStream.rangeClosed(1, 12), IntStream.rangeClosed(1, 12))
-                .boxed()
-                .sorted()
-                .collect(Collectors.toList()));
+        .isEqualTo(ImmutableList.of(1, 2, 3));
   }
 
   private List<Row> sql(String statement, Object... args) {
--- a/server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
+++ b/server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
@@ -73,7 +73,8 @@ public class CloudCredentialVendor {
             .secretAccessKey(awsSessionCredentials.secretAccessKey())
             .sessionToken(awsSessionCredentials.sessionToken()));
 
-        // Set the expiration time.
+        // Explicitly set the expiration time for the temporary credentials if it's a non-static
+        // credential. For static credential, the expiration time can be nullable.
         if (awsSessionCredentials.expiration() != null) {
           temporaryCredentials.expirationTime(awsSessionCredentials.expiration().toEpochMilli());
         }
--- a/server/src/main/java/io/unitycatalog/server/service/credential/aws/CredentialsGenerator.java
+++ b/server/src/main/java/io/unitycatalog/server/service/credential/aws/CredentialsGenerator.java
@@ -10,6 +10,19 @@ import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.sts.StsClient;
 import software.amazon.awssdk.services.sts.model.Credentials;
 
+/**
+ * Generates credentials based on the provided {@link CredentialContext}.
+ *
+ * <p>Currently supported implementations include:
+ *
+ * <ul>
+ *   <li>{@link StaticCredentialsGenerator}: returns fixed credentials defined in configuration.
+ *   <li>{@link StsCredentialsGenerator}: retrieves temporary, expiring credentials from AWS STS.
+ * </ul>
+ *
+ * <p>Test scenarios can provide custom {@link CredentialsGenerator} implementations. For example, a
+ * time-based generator can be used in end-to-end tests to validate credential renewal behavior.
+ */
 public interface CredentialsGenerator {
   Credentials generate(CredentialContext ctx);
 


openinx:
The failed test info is here:

```
info] done compiling
[info] compiling 11 Java sources to /home/runner/work/unitycatalog/unitycatalog/connectors/spark/target/scala-2.13/test-classes ...
[error] /home/runner/work/unitycatalog/unitycatalog/connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java:189:1: cannot find symbol
[error]   symbol:   method toList()
[error]   location: interface java.util.stream.Stream<java.lang.Integer>
[error] results.stream().map(r -> r.getInt(0)).toList
[error] (spark / Test / compileIncremental) javac returned non-zero exit code
[error] Total time: 57 s, completed Oct 23, 2025, 8:36:06 PM
```

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/Clock.java
```diff
@@ -23,6 +25,18 @@ static Clock manualClock(Instant now) {
     return new ManualClock(now);
   }
 
+  Map<String, Clock> globalManualClocks = new ConcurrentHashMap<>();
```
tdas: shouldnt this be private?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,262 @@
...
+    assertThat(results.size()).isEqualTo(2);
+    assertThat(results.stream().map(r -> r.getInt(0)).collect(Collectors.toList()))
+        .isEqualTo(ImmutableList.of(1, 3));
+  }
+
+  private List<Row> sql(String statement, Object... args) {
+    return session.sql(String.format(statement, args)).collectAsList();
+  }
+
+  public static class TimeBasedCredentialsGenerator implements CredentialsGenerator {
```
tdas, line 287: docs on what does this class do.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,262 @@
...
+      return Credentials.builder()
+          .accessKeyId("accessKeyId" + startTsMillis)
+          .secretAccessKey("secretAccessKey" + startTsMillis)
+          .sessionToken("sessionToken" + startTsMillis)
+          .expiration(Instant.ofEpochMilli(startTsMillis + DEFAULT_INTERVAL_MILLIS))
+          .build();
+    }
+  }
+
+  public static class S3CredFileSystem extends CredentialTestFileSystem {
```
tdas, line 310: docs on what does this do to mimic S3AFileSystem

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+        long ts = CLOCK.now().toEpochMilli() / DEFAULT_INTERVAL_MILLIS * DEFAULT_INTERVAL_MILLIS;
+        AwsSessionCredentials cred = (AwsSessionCredentials) provider.resolveCredentials();
+        assertThat(cred.accessKeyId()).isEqualTo("accessKeyId" + ts);
+        assertThat(cred.secretAccessKey()).isEqualTo("secretAccessKey" + ts);
+        assertThat(cred.sessionToken()).isEqualTo("sessionToken" + ts);
+
+        verifiedTs.add(ts);
+      }
+    }
```
tdas, line 332: if the host is not what is expected in a test .. will the checkCredential implicitly pass? 
if yes, then that is wrong as some test may accidentally pass if someone changes the bucket name in the test and not know this internal detail.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,262 @@
...
+  private SparkSession createSparkSession() {
+    return SparkSession.builder()
+        .appName("test-credential-renewal")
+        .master("local[*]")
+        .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
+        .config(
+            "spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog")
+        .config("spark.hadoop." + UCHadoopConf.UC_TEST_CLOCK_NAME, CLOCK_NAME)
+        .config("spark.hadoop." + UCHadoopConf.UC_RENEWAL_LEAD_TIME_KEY, 0L)
+        .config("spark.default.parallelism", "1")
```
tdas: isnt this achieved by `local[1]`?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/AwsVendedTokenProvider.java
```diff
@@ -18,17 +17,11 @@ public AwsVendedTokenProvider(Configuration conf) {
...
-    initialize(conf);
-  }
-
   @Override
   public GenericCredential initGenericCredential(Configuration conf) {
     if (conf.get(UCHadoopConf.S3A_INIT_ACCESS_KEY) != null
         && conf.get(UCHadoopConf.S3A_INIT_SECRET_KEY) != null
-        && conf.get(UCHadoopConf.S3A_INIT_SESSION_TOKEN) != null
-        && conf.get(UCHadoopConf.S3A_INIT_CRED_EXPIRED_TIME) != null) {
+        && conf.get(UCHadoopConf.S3A_INIT_SESSION_TOKEN) != null) {
```
tdas, line 24: >        && conf.get(UCHadoopConf.S3A_INIT_CRED_EXPIRED_TIME) != null) {

why is this removed?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,262 @@
...
+        .config("spark.sql.shuffle.partitions", "1")
+        .config("spark.sql.catalog.renewal", UCSingleCatalog.class.getName())
```
tdas: i was stupidly a bit confused by this. `renewal` is a scope that you have introduced as part of `  public static final String UC_RENEWAL_LEAD_TIME_KEY = "fs.unitycatalog.renewal.leadTimeMillis"` ... and i was confused why `renewal` is being used here. 

can you name is something more unique.. like the name of the test. 
also the name is being used in multiple places here and beforeEach .. can you just define a const?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,262 @@
...
+    try {
+      catalogOperations.deleteCatalog("renewal", Optional.of(true));
+    } catch (Exception e) {
+      // Ignore
+    }
+    super.cleanUp();
+  }
+
+  @Test
+  public void testRenewal() throws Exception {
```
tdas: how long does this test take to run?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,262 @@
...
+    sql("CREATE TABLE %s USING delta LOCATION '%s'", TABLE_NAME, location);
+
+    // Insert the original row.
+    sql("INSERT INTO %s VALUES (1)", TABLE_NAME);
+
+    SerializableConfiguration serialConf =
+        new SerializableConfiguration(
+            DeltaTable.forName(session, TABLE_NAME).deltaLog().newDeltaHadoopConf());
+
+    JavaRDD<Row> rdd =
```
tdas: this is a fairly complicated code.. please add comments explaining what is happening in this code.. how is it testing stuff with manual clock such that credential renew is actually tested.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,262 @@
...
+                row -> {
+                  Configuration conf = serialConf.value();
+                  S3CredFileSystem fs = (S3CredFileSystem) FileSystem.get(new URI(location), conf);
+
+                  fs.getFileStatus(locPath);
+                  assertThat(fs.renewalCount()).isEqualTo(0);
+
+                  // Advance the clock to trigger the 1st renewal.
+                  CLOCK.advance(Duration.ofMillis(DEFAULT_INTERVAL_MILLIS));
+                  fs.getFileStatus(locPath);
```
tdas: you are doing fs operations yourself? then you are not testing that the Delta table read is actually doing file operations with proper credential renewal isnt it??

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+  @BeforeEach
+  public void beforeEach() throws Exception {
+    super.setUp();
+
+    session = createSparkSession();
+    catalogOperations.createCatalog(new CreateCatalog().name("renewal").comment("Spark catalog"));
+
+    SdkSchemaOperations schemaOperations = new SdkSchemaOperations(createApiClient(serverConfig));
+    schemaOperations.createSchema(new CreateSchema().name("default").catalogName("renewal"));
```
tdas: do you need to do this set up before every test.. or just once in the testsuite.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/AwsVendedTokenProvider.java
openinx: Because the `expiredTime` is allowed to be null.   When I design this PR, and the original code did not set any expiration time when generating the credential from the UC server, Please see: https://github.com/unitycatalog/unitycatalog/pull/1123/files#diff-5084dc98e0f29a934bf2486de5be2667d7c6d375beb3e23add41a3bbd42554fbR72-R74

And I found we won't correctly initialize the `GenericCredentials` because of the `conf.get(UCHadoopConf.S3A_INIT_CRED_EXPIRED_TIME) != null` checking, since it's always null.  (The UC server does not set it ).

The above is the code investigation. From another side, if the `temporaryCredentialsApi` just did not get the expiration time for credential, but actually the `credentials` is generated,  then actually we will never use it a the initial `GenericCredentials`.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -17,9 +17,6 @@
 import org.sparkproject.guava.cache.CacheBuilder;
 
 public abstract class GenericCredentialProvider {
-  // The remaining time before expiration, used to trigger renewal in advance.
-  private static final long DEFAULT_RENEWAL_LEAD_TIME_MILLIS = 30 * 1000;
```
openinx, line 21: This can be removed, because we just introduced a hadoop configuration key, so that we can also set it the test suites. That will be more convinient to do the integration testing, since for the IT, we don't have API to explicitly use the `GenericCredentialsProvider`'s constructor to pass `renewalLeadTime` to construct --- All the providers are initialized via the java reflection. 

The only approach is using the `hadoop` configuration setting.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -43,18 +39,17 @@ public abstract class GenericCredentialProvider {
...
-  GenericCredentialProvider(Clock clock, long renewalLeadTimeMillis) {
-    this.clock = clock;
-    this.renewalLeadTimeMillis = renewalLeadTimeMillis;
-  }
-
   protected void initialize(Configuration conf) {
     this.conf = conf;
 
+    // Use the test clock if one is intentionally configured for testing.
+    String clockName = conf.get(UCHadoopConf.UC_TEST_CLOCK_NAME);
```
openinx, line 46: Use the intentionally configured `clockName` to access the `manualClock` from the `globalManualClock`, so that all of those infra can share the same `manualClock`, then we can use this shared `manualClock` to control all the infra's time. 
* OSS UC server:  Need this `clock` to generate credentials. 
* `AwsVendedTokenProvider`: Need this `clock` to determine if it's time to renew.
*  IntegrationTest:  Need this `clock` to advance the time step by step. 


And considering we cannot pass the `clock` instance and make it go-through all the stack,  so introducing a configure key to selectively access the clock from global should be the correct approach.  --- But we need to remember to clean the newly allocated clock if we created one in the test. Otherwise, no big impact just accumulating some tesing clocks in the testing process.

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/Clock.java
openinx: Yes, better to be `private`.  The problem is: for the variable defined in java `interface, all will be applied the same accessibilty via `public static final`,  so we cannot add `private` modifier for the interface's level variables. 

But I think I can move the `globalManualClock` inside the class `ManualClock`, since it should be manual clock specified.  That make more sense.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -276,7 +276,7 @@ private class UCProxy(
         properties = t.getProperties.asScala.toMap ++ extraSerdeProps
       ),
       schema = StructType(fields),
-      provider = Some(t.getDataSourceFormat.getValue),
+      provider = Some(t.getDataSourceFormat.getValue.toLowerCase()),
```
openinx, line 279: If we don't lower case here,  then we will encounter another problem: https://github.com/unitycatalog/unitycatalog/issues/1121#issuecomment-3444232124

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AbfsVendedTokenProviderTest.java
```diff
@@ -19,22 +18,14 @@
 public class AbfsVendedTokenProviderTest extends BaseTokenProviderTest<AbfsVendedTokenProvider> {
   @Override
   protected AbfsVendedTokenProvider createTestProvider(
-      Clock clock,
-      long renewalLeadTimeMillis,
```
openinx, line 23: we just changed to set configuration keys in the hadoop configuration (for both `clock` and `renewLeadTimeMillis` ). So we don't need those contructors any more. 

As I saide above here: https://github.com/unitycatalog/unitycatalog/pull/1123/files#r2461390293

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Yes, read the document,  should be better to use the `local[1]` rather than setting this configure key. here is the explanation. 

* Setting `spark.default.parallelism=1` does not make Spark single-threaded — it just limits RDD parallelism by default.
* Using `local[1]` would make everything single-threaded, including I/O, scheduling, etc.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Sorry for the confusion,  let me make this more clear. .Thanks for sharing your feeling !

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: This test just use the `manualClock` to test, so it's pretty fast.  All the `Clock.advance()` will just use less than 1 ms.  The e2e test time for my local is `12sec`. 

Image

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Done, thanks for the comment !

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Done, thanks for the comment !

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: It won't success, because another host won't trigger the `verfiedTs` updating, which will finally fail the `renewalCount` assertion.   The `verifiedTs`  (exposing as `renewalCount`  is one of the intentionally designed assertion for addressing the accurate credential renewal times ).

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Updated the code and comment, please see if it make more clear for you.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Add the explanation for it here, https://github.com/unitycatalog/unitycatalog/pull/1123/files#diff-fae53ed54b97a94aa0f26b1720e854c670cc4e6738641f242b981915f065c0a2R159-R179

Please see if it make sense now, thanks for the comment !

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: I assume we may add more cases for it, so I init them in the beforeEach. I agree we should make all those operations more aligned. I just updated this.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,362 @@
...
+    super.setUp();
+    session = createSparkSession();
+
+    // Initialize the catalog in unity catalog server.
+    catalogOperations.createCatalog(
+        new CreateCatalog().name(CATALOG_NAME).comment("Spark catalog"));
+
+    // Initialize the schema in unity catalog server.
+    schemaOperations = new SdkSchemaOperations(createApiClient(serverConfig));
+    schemaOperations.createSchema(new CreateSchema().name(SCHEMA_NAME).catalogName(CATALOG_NAME));
```
tdas, line 139: cant you reuse the catalog + schema across tests with beforeAll? this is adding latency unnecessarily

tdas: then you can avoid deleting the schema as well.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,362 @@
...
+    // It is possible for a Spark job to create multiple independent filesystem instances,
+    // which may misleadingly appear to renew credentials correctly even when they do not.
+    // We adopt this simulation approach because directly accessing a Spark task’s internal
+    // filesystem instance to measure credential renewals is not feasible.
+    List<Row> rows =
+        session
+            .read()
+            .format("delta")
+            .table(TABLE_NAME)
+            .toJavaRDD()
```
tdas, line 203: if the goal is to test FS operations, then why do you need to create a table? you can just a create a RDD/task using `spark.parallelize(...)`. it will be much faster that initializing a whole table.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,362 @@
...
+    return Clock.getManualClock(CLOCK_NAME);
+  }
+
+  @Test
+  public void testFileSystemRenewal() throws Exception {
+    String location = String.format("s3://%s%s/fs", BUCKET_NAME, dataDir.getCanonicalPath());
+    Path locPath = new Path(location);
+
+    // Create the external delta path-based table.
+    S3CredFileSystem.credentialCheckEnabled = false;
```
tdas: why do you need to disable this?

tdas: and why do you need to create the path-based table separately and than again calling create to define the catalog table? what are you achieving with this two step instead of one step?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
tdas: ohh .i see .. you need the table to go through the UCSingleCatalog that configures the necessary creds + disables

then a minor feedback is to use parquet or csv tables instead of delta. testing delta is not the goal of this test.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,362 @@
...
+              new StructField("partition", DataTypes.IntegerType, false, Metadata.empty()),
+            });
+
+    // Create the external delta path-based table.
+    S3CredFileSystem.credentialCheckEnabled = false;
+    sql("CREATE TABLE delta.`%s` (id INT) USING delta PARTITIONED BY (partition INT)", location);
+    S3CredFileSystem.credentialCheckEnabled = true;
+
+    // Create the catalog table referring to external table.
+    sql("CREATE TABLE %s USING delta LOCATION '%s'", TABLE_NAME, location);
```
tdas: same comment as above test. why are you doing this into steps? why not just the second CREATE statement as `CREATE TABLE %s (id INT, partition INT) USING delta LOCATION %s PARTITIONED BY partition`

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,362 @@
...
+
+    // Insert 4 rows into each partition where (id % 3) equals the partition key.
```
tdas: why three INSERT with `INSERT INTO ... VALUES (1, 1), (1, 2), (1,3)`
that will create 3 files with 1 row each because partition column has 3 values.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,362 @@
...
+        .format("delta")
+        .table(TABLE_NAME)
+        .mapPartitions(
+            (MapPartitionsFunction<Row, Row>)
+                input -> {
+                  // Advance the clock to trigger the credential renewal.
+                  testClock().advance(Duration.ofMillis(DEFAULT_INTERVAL_MILLIS));
+                  return input;
+                },
+            Encoders.row(schema))
```
tdas: minor: probably will be simpler code to that defining schema and stuff if you use Java UDFs

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,362 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.sparkproject.guava.collect.ImmutableList;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
+import software.amazon.awssdk.services.sts.model.Credentials;
+
+public class AwsCredentialRenewalTest extends BaseCRUDTest {
```
tdas, line 61: can you document what this test does and how it does it? the overall strategy of how this test works. its pretty complicated with credential generator and S3CredsFileSystem.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,362 @@
...
+  }
+
+  @Override
+  protected void setUpProperties() {
+    super.setUpProperties();
+    serverProperties.put("s3.bucketPath.0", "s3://" + BUCKET_NAME);
+    serverProperties.put("s3.accessKey.0", "accessKey0");
+    serverProperties.put("s3.secretKey.0", "secretKey0");
+    serverProperties.put("s3.sessionToken.0", "sessionToken0");
+    serverProperties.put("s3.credentialsGenerator.0", CREDENTIALS_GENERATOR);
```
tdas, line 92: can you leave a comment here explaining what this will do?

server/src/main/java/io/unitycatalog/server/service/credential/aws/CredentialsGenerator.java
```diff
@@ -0,0 +1,77 @@
...
+import java.time.Duration;
+import java.util.UUID;
+import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.model.Credentials;
+
+public interface CredentialsGenerator {
```
tdas, line 26: docs please!

server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
```diff
@@ -72,6 +72,11 @@ public TemporaryCredentials vendCredential(CredentialContext context) {
             .accessKeyId(awsSessionCredentials.accessKeyId())
             .secretAccessKey(awsSessionCredentials.secretAccessKey())
             .sessionToken(awsSessionCredentials.sessionToken()));
+
+        // Set the expiration time.
```
tdas: add comment why this special condition is needed

server/src/main/java/io/unitycatalog/server/service/credential/aws/CredentialsGenerator.java
```diff
@@ -0,0 +1,77 @@
...
+  Credentials generate(CredentialContext ctx);
+
+  class StaticCredentialsGenerator implements CredentialsGenerator {
```
tdas, line 29: docs

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,362 @@
...
+        .saveAsTable(TABLE_NAME);
+
+    List<Row> rows = sql("SELECT * FROM %s ORDER BY id ASC", TABLE_NAME);
+    assertThat(rows.size()).isEqualTo(24);
+    assertThat(rows.stream().map(r -> r.getInt(0)).collect(Collectors.toList()))
+        .isEqualTo(
+            IntStream.concat(IntStream.rangeClosed(1, 12), IntStream.rangeClosed(1, 12))
+                .boxed()
+                .sorted()
+                .collect(Collectors.toList()));
```
tdas: general question: which test in the set of tests we have written till now ensures that the credentials are not accidentally used across queries even if they are for the same table?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,362 @@
...
+                input -> {
+                  // Advance the clock to trigger the credential renewal.
+                  testClock().advance(Duration.ofMillis(DEFAULT_INTERVAL_MILLIS));
+                  return input;
+                },
+            Encoders.row(schema))
+        .write()
+        .format("delta")
+        .mode("append")
+        .saveAsTable(TABLE_NAME);
```
tdas: shouldnt we test writing a to a different table to make sure that multiple tables in the same query work together?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
```diff
@@ -0,0 +1,362 @@
...
+
+    List<Row> rows = sql("SELECT * FROM %s ORDER BY id ASC", TABLE_NAME);
+    assertThat(rows.size()).isEqualTo(24);
+    assertThat(rows.stream().map(r -> r.getInt(0)).collect(Collectors.toList()))
+        .isEqualTo(
+            IntStream.concat(IntStream.rangeClosed(1, 12), IntStream.rangeClosed(1, 12))
+                .boxed()
+                .sorted()
```
tdas: this seems very verbose! is this the best way to do this in java?

tdas: i think if you have just 3 rows... this becomes much simpler to write this as 1, 2, 3, 1, 2, 3?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
tdas: if there are only 3 rows, 1 in each file as i suggested above ... and selected only the id, then you can just use mapFunction and Encoders.INT

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Hi @tdas ,  what you said should be right. I think the constrait is:  currently the catalog creation and schema creation depends on the OSS UC Server's [BaseCRUDITest](https://github.com/unitycatalog/unitycatalog/blob/main/server/src/test/java/io/unitycatalog/server/base/BaseCRUDTest.java#L33), however this test will intial the `catalogOperations` in the `BeforeEach` method, then we cannot make it to be part of the `beforeAll`  ( which requires to be a static method ).

openinx: There are other test cases which depend on this `BaseCRUDTest` feature ( I mean the `beforeEach` level catalog & schema initialization and destory ) ,  if  we plan to change then maybe we should change those test cases firstly.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Done, thanks for sharing your feeling !

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Done, thanks for sharing !

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Because currently we are not supporting creating UC managed table direclty. So we have to create the path-based external table in the first step, and then connect it to the catalog ( let the UC catalog know there is a table in this catalog ).

openinx: Disabling the `credentialCheckEnabled` for the first step is because the path-based table creation won't go through the `Catalog#createTable` or `Catalog#loadTable`, then we have no chance to inject the cred props , which will fail the `S3CredFileSytem`'s  credential  verification.

there are similar flag set you can see here, which I think we may need to look for a better way for this also. 
 https://github.com/unitycatalog/unitycatalog/blob/487f89fdce0428aa137fcfa59f851028f8ec35c7/connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java#L508-L514

Ideally, the correct approach should be:  supporting the credentials properties injection also for path-based table, not only catalog-based table.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: > ohh .i see .. you need the table to go through the UCSingleCatalog that configures the necessary creds + disables

Yes, that's true.  I need the delta table's hadoop configuration (with the correctly configured credential properties , so that we can do the credential renewal ).

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Left an issue here: https://github.com/unitycatalog/unitycatalog/issues/1139

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: > then a minor feedback is to use parquet or csv tables instead of delta. testing delta is not the goal of this test.

Yes,  would be simpler if we could use parquet or csv tables, however for parquet/csv,  we don't have the API to get the table-level  hadoop configuration.  Delta is a special case, since we provided the `DeltaTable` API to do this ---  Maybe I'm not so familar delta API before,  it also took me lots of time to correctly locate this API to get the hadoop configuration. 


```java
    // Generate a table level hadoop configuration, with setting the delta table's all properties.
    SerializableConfiguration serialConf =
        new SerializableConfiguration(
            DeltaTable.forName(session, TABLE_NAME).deltaLog().newDeltaHadoopConf());
```

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Oh,  sorry.  we currently only not support the path-based table creation without specified location. 

The SQL `CREATE TABLE %s (id INT) USING delta LOCATION '%s'` should be fine. Thanks for the reminder.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Thanks for the reminder,  Done !

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Em, let me simply this test.  Done !

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Done

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Done

server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
openinx: Done

server/src/main/java/io/unitycatalog/server/service/credential/aws/CredentialsGenerator.java
openinx: Done

server/src/main/java/io/unitycatalog/server/service/credential/aws/CredentialsGenerator.java
openinx: Done

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
openinx: Done

connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java
tdas: Got it! thanks for explaining.

Issue #1066, opened 2025-08-29
Spark Integration: Metadata inconsistency when schema contains hyphens


celophie:
**Describe the bug**

For Spark backticks are needed when the SQL identifier is containing hyphens.
In Unity catalog I have a compatibility problem with backticks in the SQL identifier.
Specifically, when I create a schema contains hyphens with the UC CLI or with Spark using the Spark integration, `test-schema`, and query the schema I get:
```
SHOW SCHEMAS IN `unity`;
+----------------+
|   namespace    |
+----------------+
| default        |
| `test-schema`  |
+----------------+
```
Now, I want to create a table where the table name contains hyphens, `test-table2`. 
The table will be created but there is an inconsistency now in the metadata. When Unity Catalog is trying to find the table again, it fails and throws an error. 
```
Caused by: org.apache.spark.sql.catalyst.analysis.NoSuchTableException: [TABLE_OR_VIEW_NOT_FOUND] The table or view `test-schema`.`test-table1` cannot be found. Verify the spelling and correctness of the schema and catalog.
If you did not qualify the name with a schema, verify the current_schema() output, or qualify the name with the correct schema and catalog.
To tolerate the error on drop use DROP VIEW IF EXISTS or DROP TABLE IF EXISTS.
	at io.unitycatalog.spark.UCProxy.loadTable(UCSingleCatalog.scala:240)
	at io.unitycatalog.spark.UCProxy.createTable(UCSingleCatalog.scala:336)
```
```
SHOW TABLES IN `test-schema`;
+----------------+--------------+--------------+
|   namespace    |  tableName   | isTemporary  |
+----------------+--------------+--------------+
| `test-schema`  | test-table2  | false        |
+----------------+--------------+--------------+
```
When I want to query the table I get:
```
Caused by: org.apache.spark.sql.AnalysisException: [TABLE_OR_VIEW_NOT_FOUND] The table or view `unity`.`test-schema`.`test-table2` cannot be found. Verify the spelling and correctness of the schema and catalog.
If you did not qualify the name with a schema, verify the current_schema() output, or qualify the name with the correct schema and catalog.
To tolerate the error on drop use DROP VIEW IF EXISTS or DROP TABLE IF EXISTS.; line 1 pos 15;
'DescribeRelation false, [col_name#79, data_type#80, comment#81]
+- 'UnresolvedTableOrView [unity, test-schema, test-table2], DESCRIBE TABLE, true
```
[...]

PR #1125, opened 2025-10-21
Fix hyphens in schema names (Issue #1066)

carrez:
Fixes hyphen bug in schema names when creating tables (Issue #1066)
**PR Checklist**

**Description of changes**

Once a table is created, it is loaded using `TablesAPI.getTable`. However the fullname is created using `ident.toString`, which automatically quotes names when needed (notably when a hyphen is present; [code pointer](https://github.com/apache/spark/blob/65b466c241a65f62533bc579bffd406825518429/sql/api/src/main/java/org/apache/spark/sql/connector/catalog/IdentifierImpl.java#L52)). This results in a 404 from Unity Catalog as names don't need to be quoted for this API.
The fix builds instead the table fullname from the namespace and name of the identifier.

As `loadTable` uses now only the first element of the namespace, unit tests were failing. I had to:
- add the `checkUnsupportedNestedNamespace` check to the method, which seems more aligned with other methods in the class
- change the expected exception message in `testDropTable` test as the exception is different.

Note the test was failing at executing `DROP TABLE a.b.c.d`, which would at some point try to retrieve the table with `TableRepository.getTable`. This method would throw a `BaseException` with message "Invalid table name", as it does not handle nested namespaces ([code pointer](https://github.com/unitycatalog/unitycatalog/blob/487f89fdce0428aa137fcfa59f851028f8ec35c7/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java#L76)).

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
@@ -342,7 +342,7 @@ public class TableReadWriteTest extends BaseSparkIntegrationTest {
     assertThat(session.catalog().tableExists(fullName)).isFalse();
     assertThatThrownBy(() -> session.sql("DROP TABLE a.b.c.d").collect())
         .isInstanceOf(ApiException.class)
-        .hasMessageContaining("Invalid table name");
+        .hasMessageContaining("Nested namespaces are not supported");
     session.stop();
   }
 


carrez:
Unit tests TableReadWriteTest are failing, I'll fix them.

But any feedback on the change is welcomed :)

carrez:
@tdas can I get a review?
Thanks!

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -216,9 +216,10 @@ private class UCProxy(
   }
 
   override def loadTable(ident: Identifier): Table = {
+    checkUnsupportedNestedNamespace(ident.namespace())
     val t = try {
       tablesApi.getTable(
-        name + "." + ident.toString,
+        name + "." + (if (ident.namespace().isEmpty) "" else ident.namespace().head + ".") + ident.name(),
```
carrez, line 222: Some queries might have no namespace (no schema), meaning `namespace().head` would throw an exception.

The previous behaviour was also concatenating the catalog to the table name `<catalog>.<table>` in case the schema was missing, but maybe we want to rather have only `<table>` if schema is missing.

yili-db: @carrez 
Can you please give an example on which query would result in an empty namespace()?

yili-db: Looks like this happens when a invalid table name without schema is specified in query, and no default schema is chosen.
In that case either it fails here, or it fails at UC server.

carrez: Yes, the empty namespace can happen if the SQL query is something like `CREATE TABLE table ....` with no default schema.

Without the `if` then the error is cryptic for the user:
```
Error: org.apache.hive.service.cli.HiveSQLException: Error running query: java.util.NoSuchElementException: head of empty array
```

carrez:
Closing in favor of https://github.com/unitycatalog/unitycatalog/pull/1248

PR #1126, opened 2025-10-22
[Managed table 1] Add staging table API to the spec

yili-db:
**PR Checklist**

**Description of changes**

Add staging table API to the spec

### DIFF
--- a/api/Apis/TablesApi.md
+++ b/api/Apis/TablesApi.md
@@ -4,11 +4,39 @@ All URIs are relative to *http://localhost:8080/api/2.1/unity-catalog*
 
 | Method | HTTP request | Description |
 |------------- | ------------- | -------------|
+| [**createStagingTable**](TablesApi.md#createStagingTable) | **POST** /staging-tables | Create a staging table |
 | [**createTable**](TablesApi.md#createTable) | **POST** /tables | Create a table. Only external table creation is supported. WARNING: This API is experimental and will change in future versions.  |
 | [**deleteTable**](TablesApi.md#deleteTable) | **DELETE** /tables/{full_name} | Delete a table |
 | [**getTable**](TablesApi.md#getTable) | **GET** /tables/{full_name} | Get a table |
 | [**listTables**](TablesApi.md#listTables) | **GET** /tables | List tables |
 
+
+<a name="createStagingTable"></a>
+# **createStagingTable**
+> StagingTableInfo createStagingTable(CreateStagingTable)
+
+Create a staging table
+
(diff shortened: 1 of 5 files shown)


api/Apis/TablesApi.md
```diff
@@ -4,12 +4,40 @@ All URIs are relative to *http://localhost:8080/api/2.1/unity-catalog*
...
...s a new staging table instance. Staging tables are used during managed table creation. Creating a managed table requires performing two actions – initializing the table data in cloud storage and creating the named table entry in the catalog – and these should appear as an atomic operation to other operations on catalog tables. A staging table is used to allocate storage for the managed table, and the &#x60;catalog_name.schema_name.name&#x60; parameters provided in this request are used to initialize any required storage properties and determine the storage URL that should be used for the data contained by this table.  Temporary credentials can be obtained as though the staging table were a regular table to get access to the staging table’s storage. After the table’s data is initialized, the staging table is “promoted” to a managed table by creating a managed table with the same location as the staging table. This allows for the atomic creation of a managed table that already has full data written to its storage location. Note: the name provided must match the name used to initialize the staging table originally.  WARNING: This API is experimental and will change in future versions.
```
openinx: `&#x60;catalog_name.schema_name.name&#x60;` ,  what does this mean ? Is it an encoded HTML tag ?

api/Apis/TablesApi.md
```diff
@@ -4,12 +4,40 @@ All URIs are relative to *http://localhost:8080/api/2.1/unity-catalog*
...
+|------------- | ------------- | ------------- | -------------|
+| **CreateStagingTable** | [**CreateStagingTable**](../Models/CreateStagingTable.md)|  | [optional] |
+
+### Return type
+
+[**StagingTableInfo**](../Models/StagingTableInfo.md)
+
+### Authorization
+
+No authorization required
```
openinx, line 34: What does this mean here ? `No authorization required`. What's the case that require the `Authorization` ?

api/Models/CreateStagingTable.md
```diff
@@ -0,0 +1,11 @@
+# CreateStagingTable
+## Properties
+
+| Name | Type | Description | Notes |
```
openinx, line 4: Seems like we don't have a column indicating which column is required or optional ?

yili-db: All are required. Only optional ones will be marked as [optional]. For example, check CreateModelVersion.md

api/Apis/TablesApi.md
yili-db: I guess that simply means that we don't document authorization in this all.yaml. This is also true for `createTable` in the same file.

api/Apis/TablesApi.md
yili-db: This is the encoded single quote. It appears that this openapi->md can't handle single quote gracefully. So I just removed it.

api/all.yaml
```diff
@@ -276,6 +276,39 @@ paths:
...
+        parameters provided in this request are used to initialize any required storage properties and determine the
+        storage URL that should be used for the data contained by this table.
+
+        Temporary credentials can be obtained as though the staging table were a regular table to get access to the
+        staging table’s storage. After the table’s data is initialized, the staging table is “promoted” to a managed
+        table by creating a managed table with the same location as the staging table. This allows for the atomic
+        creation of a managed table that already has full data written to its storage location. Note: the name provided
+        must match the name used to initialize the staging table originally.
+
+        WARNING: This API is experimental and will change in future versions.
```
tdas, line 299:
```suggestion
        WARNING: This API is experimental and may change in future versions.
```

PR #1127, opened 2025-10-23
Add coordinated commits API to the spec

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1127/files) to review incremental changes.
- [**commit_api**](https://github.com/unitycatalog/unitycatalog/pull/1127) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1127/files)]
  - [stack/commit_basic](https://github.com/unitycatalog/unitycatalog/pull/1166) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1166/files/725c6b686f9f169fb10d7109dfa2c5ea2cd58a4e..2ec21c04831867e54feef742774af6c7b034dea1)]
    - [stack/commit_backfill_delete_meta](https://github.com/unitycatalog/unitycatalog/pull/1169) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1169/files/6d0a2246ce9ded3408587e35355ad46deb22dd2d..711f75644a2ae2559277413d36c8e7118a9bd3ca)]
      - [stack/commit_get](https://github.com/unitycatalog/unitycatalog/pull/1170) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1170/files/333883036048eb3c5fc5defd6ad193849b2286a4..3e5b8dfe047577223b7e9decac67725e5f4d6c76)]

---------
**PR Checklist**

**Description of changes**

Add coordinated commits API to the spec

### DIFF
--- a/api/Apis/CoordinatedCommitsApi.md
+++ b/api/Apis/CoordinatedCommitsApi.md
@@ -4,7 +4,7 @@ All URIs are relative to *http://localhost:8080/api/2.1/unity-catalog*
 
 | Method | HTTP request | Description |
 |------------- | ------------- | -------------|
-| [**commit**](CoordinatedCommitsApi.md#commit) | **POST** /delta/preview/commits | Commit changes to a specified table WARNING: This API is experimental and may change in future versions.  |
+| [**commit**](CoordinatedCommitsApi.md#commit) | **POST** /delta/preview/commits | Commit changes to a specified table. The server has a limit defined in config on how many unbackfilled commits it can hold. Clients are expected to do active backfill of the commit after committing to UC. So in most cases the number of unbackfilled commits should be close to zero or one. But if clients misbehave and unbackfilled commits accumulate beyond the limit, server will reject further commits until more backfill is done. WARNING: This API is experimental and may change in future versions.  |
 | [**getCommits**](CoordinatedCommitsApi.md#getCommits) | **GET** /delta/preview/commits | List unbackfilled Delta table commits. WARNING: This API is experimental and may change in future versions.  |
 
 
@@ -12,7 +12,7 @@ All URIs are relative to *http://localhost:8080/api/2.1/unity-catalog*
 # **commit**
 > Object commit(Commit)
 
-Commit changes to a specified table WARNING: This API is experimental and may change in future versions. 
+Commit changes to a specified table. The server has a limit defined in config on how many unbackfilled commits it can hold. Clients are expected to do active backfill of the commit after committing to UC. So in most cases the number of unbackfilled commits should be close to zero or one. But if clients misbehave and unbackfilled commits accumulate beyond the limit, server will reject further commits until more backfill is done. WARNING: This API is experimental and may change in future versions. 
 
 ### Parameters
 
@@ -45,10 +45,10 @@ List unbackfilled Delta table commits. WARNING: This API is experimental and may
 
 |Name | Type | Description  | Notes |
 |------------- | ------------- | ------------- | -------------|
-| **table\_id** | **String**| The table ID provided by UC on table creation that we want to get the commits for. In the context of UC, this is the UUID that uniquely identifies a table. | [default to null] |
-| **table\_uri** | **String**| The URI of the storage location of the table. If the table_id exists but the table_uri is  different from the one previously registered (e.g., if the client moved the table), the request will fail.  | [default to null] |
-| **start\_version** | **Long**| The start version from which to retrieve commits (inclusive). | [default to null] |
-| **end\_version** | **Long**| The end version upto which to retrieve commits (inclusive). If not set, the latest version will be used as the end version. This does not affect the latest_table_version in the response. If num of commits that meet this criteria is &gt; 50, the response will be paginated to the first 50 commits.  | [optional] [default to null] |
+| **table\_id** | **String**| The ID of the table to get the commits for. This ID uniquely identifies a table. | [default to null] |
+| **table\_uri** | **String**| The URI of the storage location of the table. If the table_id exists but the table_uri is  different from the one previously registered (e.g., if the client moved the table), the request will fail. Example: s3://bucket-name/tables/some-table-id  | [default to null] |
+| **start\_version** | **Long**| The start version from which to retrieve commits (inclusive). This along with the optional end_version specifies the range of commit versions that this request wants.  | [default to null] |
+| **end\_version** | **Long**| The end version upto which to retrieve commits (inclusive). If not set, the latest version will be used as the end version. This does not affect the latest_table_version in the response. If num of commits that meet this criteria is larger than a limit set by server config, the response will be limited to the first X commits. Call can send request again with a larger start_version according to the response to get the remaining commits.  | [optional] [default to null] |
 
 ### Return type
 
--- a/api/all.yaml
+++ b/api/all.yaml
@@ -40,7 +40,7 @@ tags:
       A metastore is the top-level container of objects in Unity Catalog that stores all UC assets.
   - name: CoordinatedCommits
     description: |
-      "Coordinated commits" is a Delta tables feature that allows Delta tables to designate a coordinator to coordinate commits. This is Unity Catalog's API to allow Unity Catalog be the commit coordinator for managed tables. (See https://github.com/delta-io/delta/blob/master/PROTOCOL.md for more details)
+      "Coordinated commits" is a Delta tables feature that allows Delta tables to designate a coordinator to coordinate commits. This is Unity Catalog's API to allow Unity Catalog be the commit coordinator for managed tables. (See https://github.com/delta-io/delta/blob/master/protocol_rfcs/catalog-managed.md for more details)
 paths:
   /catalogs:
     post:
@@ -1417,7 +1417,7 @@ paths:
       parameters:
         - name: table_id
           in: query
-          description: The table ID provided by UC on table creation that we want to get the commits for. In the context of UC, this is the UUID that uniquely identifies a table.
+          description: The ID of the table to get the commits for. This ID uniquely identifies a table.
           required: true
           schema:
             type: string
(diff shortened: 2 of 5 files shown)


api/Apis/CoordinatedCommitsApi.md
```diff
@@ -0,0 +1,65 @@
+# CoordinatedCommitsApi
+
+All URIs are relative to *http://localhost:8080/api/2.1/unity-catalog*
+
+| Method | HTTP request | Description |
+|------------- | ------------- | -------------|
+| [**commit**](CoordinatedCommitsApi.md#commit) | **POST** /delta/preview/commits | Commit changes to a specified table WARNING: This API is experimental and may change in future versions.  |
```
openinx: nit: maybe we can just decalare for the first sentence that this API is expeirmental,  add this warning for every fields seems like to be a bit nosiy.

openinx: Oh,  those are API list here. not the api fields.

api/Apis/CoordinatedCommitsApi.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Parameters
+
+|Name | Type | Description  | Notes |
+|------------- | ------------- | ------------- | -------------|
+| **table\_id** | **String**| The table ID provided by UC on table creation that we want to get the commits for. In the context of UC, this is the UUID that uniquely identifies a table. | [default to null] |
+| **table\_uri** | **String**| The URI of the storage location of the table. If the table_id exists but the table_uri is  different from the one previously registered (e.g., if the client moved the table), the request will fail.  | [default to null] |
+| **start\_version** | **Long**| The start version from which to retrieve commits (inclusive). | [default to null] |
+| **end\_version** | **Long**| The end version upto which to retrieve commits (inclusive). If not set, the latest version will be used as the end version. This does not affect the latest_table_version in the response. If num of commits that meet this criteria is &gt; 50, the response will be paginated to the first 50 commits.  | [optional] [default to null] |
```
openinx: If the num of commits `>50`,  do we currently have api to fetch the next page ?   Seems like it a bit tricky for me, since it currently always get the first 50.

openinx: Oh, you mean people can specify the `start_version` to fetch the next 50 commits in a batch, right ?

yili-db: Yes. That's a good point. I just added a sentence to describe that.

api/all.yaml
```diff
@@ -2792,6 +2899,108 @@ components:
...
+          type: integer
+          format: int64
+          description: The size of the commit file in bytes.
+        file_modification_timestamp:
+          type: integer
+          format: int64
+          description: The modification time of the commit file. This is the mod time of the file as written to the file system.
+        is_disown_commit:
+          type: boolean
+          description: Should be set if this commit is a disown commit that removes UC as the commit coordinator from the table.
```
openinx: What's the meaning of `disown`  ? what's  the behavior if we set this field as true or false ? Could you please explain more ?

api/all.yaml
```diff
@@ -38,6 +38,9 @@ tags:
   - name: Metastores
     description: |
       A metastore is the top-level container of objects in Unity Catalog that stores all UC assets.
+  - name: CoordinatedCommits
+    description: |
+      "Coordinated commits" is a Delta tables feature that allows Delta tables to designate a coordinator to coordinate commits. This is Unity Catalog's API to allow Unity Catalog be the commit coordinator for managed tables. (See https://github.com/delta-io/delta/blob/master/PROTOCOL.md for more details)
```
openinx: Is the https://github.com/delta-io/delta/blob/master/PROTOCOL.md correct page for details ? Seems no mention about the `coordinator commits`.

api/all.yaml
```diff
@@ -1398,7 +1401,111 @@ paths:
...
+        WARNING: This API is experimental and may change in future versions.
+      description: |
+        List all the unbackfilled delta commits that are currently being tracked by the UC coordinator. 
+        If no commits are being tracked in the specific version range (from start_version to an optional end_version), 
+        it will return an empty list.
+        WARNING: This API is experimental and may change in future versions.
+      parameters:
+        - name: table_id
+          in: query
+          description: The table ID provided by UC on table creation that we want to get the commits for. In the context of UC, this is the UUID that uniquely identifies a table.
```
openinx: > this is the UUID that uniquely identifies a table.

For the api level,  I think it's good enough to say  `it uniquely identifies a table` , but no necessary to say it's definitely a `UUID` type.  any unique string is possible to represent the `tableId`, no necessary `UUID`, I think.

api/all.yaml
```diff
@@ -1398,7 +1401,111 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - name: table_uri
+          in: query
+          description: |
+            The URI of the storage location of the table. If the table_id exists but the table_uri is 
+            different from the one previously registered (e.g., if the client moved the table), the request will fail.
```
openinx, line 1428: Do we need to have an example here, I guess it will be `s3://bucket/key` or something like that ?

api/all.yaml
```diff
@@ -1398,7 +1401,111 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - name: start_version
```
openinx, line 1433: Is this `start_version` an auto incremental field, or will it be comparable to the `end_version` ?

api/all.yaml
```diff
@@ -1398,7 +1401,111 @@ paths:
...
+        '401':
+          description: Unauthorized. Authentication credentials are missing or incorrect.
+        '403':
+          description: Forbidden. The user does not have the necessary permissions to perform this action.
+        '404':
+          description: Not Found. The specified table does not exist.
+        '409':
+          description: Conflict. Indicates that a commit with the same version already exists.
+        '429':
+          description: The maximum number of unbackfilled commits per table has been reached. Please backfill some commits before retrying.
```
openinx, line 1512: > The maximum number of unbackfilled commits per table has been reached.

Usually, what's the threshold ?

openinx (approved):
Looks good to me overall,  just left several small comments.

api/all.yaml
yili-db: It is not published yet. I just put it here as a place holder.

api/all.yaml
yili-db: `disown` means that this table no longer wants to use UC as the commit coordinator. This releases UC the responsibility of maintaining any unbackfilled commit and makes sure that all commits are backfilled. Sort of like converting it to an EXTERNAL table. There could be use cases that want to copy the entire table files to somewhere else for example.
@tdas do we want to support this ever in the future? My current impl doesn't support that yet.

api/all.yaml
yili-db: It's basically specifying a pagination range of [start_version, end_version]. Server may also shrink the range in order not to exceeds 50 versions per response.
Added clarification here.

api/all.yaml
yili-db: It's 50. But I am not sure if this will be part of the published spec. Or we'll keep it as an internal impl detail.
Anyway it shouldn't exceed as client is expected to do backfill asap.

api/all.yaml
tdas: can you link to the RFC instead and say that it is 

so basicaly linked to both.

api/all.yaml
```diff
@@ -1398,7 +1401,114 @@ paths:
...
+            format: int64
+        - name: end_version
+          in: query
+          description: |
+            The end version upto which to retrieve commits (inclusive).
+            If not set, the latest version will be used as the end version. This does not affect the
+            latest_table_version in the response.
+            If num of commits that meet this criteria is > 50, the response will be paginated to the
+            first 50 commits. Call can send request again with a larger start_version to get the 
+            remaining commits.
```
tdas: 50 is an implementation choice the server can make based on the backend database requirements (how much store. from the catalog spec it perfectly correct to have the threshold be 1. please make that clear. 

in the implementation, this needs to be a server property

api/all.yaml
tdas: see comment below. 50 is a implementation detail

api/all.yaml
```diff
@@ -1398,7 +1401,114 @@ paths:
...
+                $ref: '#/components/schemas/GetCommitsResponse'
+        '400':
+          description: |
+            Bad request or Invalid Argument. Example: table_uri in the request doesn't match the existing table_uri for the table.
+        '401':
+          description: Unauthorized. Authentication credentials are missing or incorrect.
+        '403':
+          description: Forbidden. The user does not have the necessary permissions to perform this action.
+        '404':
+          description: Not Found. The specified table does not exist.
```
tdas, line 1470: @yili-db we need to make sure that these are response codes are tested as well

api/all.yaml
tdas: my response https://github.com/unitycatalog/unitycatalog/pull/1127/files#r2500180025

api/all.yaml
tdas: lets add this to the spec in the future. we can always add stuff as optional fields in the future when we have more clarity. lets just start with the core functionality. 
so remove it for now.

tdas (approved):
Overall looks good. please address all the comments. then i will merge.

api/all.yaml
yili-db: Changed the link to https://github.com/delta-io/delta/blob/master/protocol_rfcs/catalog-managed.md

api/all.yaml
yili-db: Yes. Will do.

openinx (approved):
LGTM, thanks @yili-db for the work.

PR #1130, opened 2025-10-24
Allow providing base path for UC API Client

urosstan-db:
**PR Checklist**
- Certain vendors may register UC API under a certain path (e.g. Microsoft Fabric API registers it under `https://onelake.table.fabric.microsoft.com/delta/<Workspace name or workspace ID>/<Item name or item ID>/api/2.1/unity-catalog/`) 
- New test suite added
- This is a behavioural change, since if someone provided a URL earlier with a specific path, then the path would be just ignored (e.g., the fabric API I provided above would be translated to `https://onelake.table.fabric.microsoft.com/api/2.1/unity-catalog/`). But probably no users provide a URL with a path, since the UC API could not be hit prior to this fix.
- Base URL prior this change was `api/2.1/unity-catalog/`, while now it is `basePathFromProvidedUrl` + `api/2.1/unity-catalog/`

**Description of changes**
- In the API Client factory, set the base path of the provided URL where the UC API is registered. This code path is hit when we initialize `UCSingleCatalog`.
- Behaviour prior this change caused failure on URIs with certain path:
image

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
@@ -16,8 +16,8 @@ public class ApiClientFactory {
     // base path from given `url` but still preserving path suffix.
     // Expected input for `url` is URL with no "/api/2.1/unity-catalog" in the path.
     String basePath = url.getPath() + BASE_PATH;
-    RetryingApiClient apiClient = (RetryingApiClient) new RetryingApiClient(clientConf, Clock.systemClock())
-        .setHost(url.getHost())
+    RetryingApiClient apiClient = new RetryingApiClient(clientConf, Clock.systemClock());
+    apiClient.setHost(url.getHost())
         .setPort(url.getPort())
         .setScheme(url.getScheme())
         .setBasePath(basePath);


connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
```diff
@@ -5,13 +5,18 @@
...
   private ApiClientFactory() {}
 
   public static ApiClient createApiClient(URI url, String token) {
+    String basePath = url.getPath() + BASE_PATH;
     ApiClient apiClient = new ApiClient()
         .setHost(url.getHost())
         .setPort(url.getPort())
-        .setScheme(url.getScheme());
+        .setScheme(url.getScheme())
+        .setBasePath(basePath);
```
urosstan-db, line 23: This base path was set in ApiClient constructor to `/api/2.1/unity-catalog`, because `ApiClient::getDefaultBaseUri` is `http://localhost:8080/api/2.1/unity-catalog`, while we invoke `updateBaseUri(getDefaultBaseUri());` in constructor, so base path was already set to `/api/2.1/unity-catalog`.

openinx:
Hi @urosstan-db , thanks for the contributionn.  Could you please help to check the `DCO` issue ?

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
openinx: If we don't set the `/api/2.1/unity-catalog`, what's the version (or base path) that the API call will go through by default ?

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
```diff
@@ -5,13 +5,18 @@
 import java.net.URI;
 
 public class ApiClientFactory {
+
+  public static final String BASE_PATH = "/api/2.1/unity-catalog";
+
   private ApiClientFactory() {}
 
   public static ApiClient createApiClient(URI url, String token) {
+    String basePath = url.getPath() + BASE_PATH;
```
openinx, line 18: Why do we need to `concat` the `url.getPath()` and `BASE_PATH` to set the `basePath` for the ApiClient ?

openinx: ( since we've already set the host and port for the `apiClient` ).

urosstan-db: `getPath` returns the part of URI after host and port.
https://docs.oracle.com/javase/8/docs/api/java/net/URI.html
https://www.geeksforgeeks.org/java/uri-getpath-method-in-java-with-examples/

Since we set host and port only currently (prior to this PR), the path is just omitted in the final URL of `ApiClient`, which prevents us from connecting to the UC API registered at some endpoint with a path.

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
urosstan-db: Before this PR, base path was `/api/2.1/unity-catalog`, while after this PR, it is path from provided URL + `/api/2.1/unity-catalog`.

## URL input 1:
https://databricks.workspace.com

Before: https://databricks.workspace.com/api/2.1/unity-catalog
After: https://databricks.workspace.com/api/2.1/unity-catalog

## URL input 2
https://onelake.table.fabric.microsoft.com/delta/<Workspace name or workspace ID>/<Item name or item ID>

Before: https://onelake.table.fabric.microsoft.com/api/2.1/unity-catalog/
After: https://onelake.table.fabric.microsoft.com/delta/<Workspace name or workspace ID>/<Item name or item ID>/api/2.1/unity-catalog/

rameshchandra (approved):
LGTM. 

@urosstan-db please test before merging

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
openinx: Okay, I think I got the point.  Thanks for the information. Could  you please help to add some unit tests to address this ?  Seems like it's hard to get this context correctly. Thanks.

urosstan-db: Thanks, I added ApiClientFactoryTest and one method to test base URI for all APIs

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
```diff
@@ -0,0 +1,26 @@
...
+public class ApiClientFactoryTest extends BaseSparkIntegrationTest {
+    @Test
+    public void testApiClientBaseUri() {
+        String token = "";
+
+        URI uriDatabricks = URI.create("https://adb-1234567890123456.3.azuredatabricks.net");
+        ApiClient apiClientDatabricks = ApiClientFactory.createApiClient(uriDatabricks, token);
+        assertThat(apiClientDatabricks.getBaseUri())
+            .isEqualTo("https://adb-1234567890123456.3.azuredatabricks.net/api/2.1/unity-catalog");
+
```
urosstan-db, line 19: I verified there is no behavioural change for existing usage, since without `main` change, test failed on second assertion when URL has some path:
image

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
```diff
@@ -0,0 +1,26 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.net.URI;
+
+public class ApiClientFactoryTest extends BaseSparkIntegrationTest {
+  @Test
+  public void testApiClientBaseUri() {
+    String token = "";
+
+    URI uriDatabricks = URI.create("https://adb-1234567890123456.3.azuredatabricks.net");
```
tdas: which of these 2 cases passed before your changes to the factory? 
I am trying to make sure that we are capturing in a test that the existing URLs are not broken.

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
tdas: can you leave this context as a comment in the code on why basePath needs to be set? this is important context that should be captured in the code.

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
urosstan-db: This URI passed, `uriDatabricks`, since it has no path in URL, while fabric failed

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
urosstan-db: @tdas Yes, that's great point

tdas (approved):

tdas:
- #1215

PR #1129, opened 2025-10-23
Fix testing failure by adding explicit dependency to commons-lang3.

yili-db:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

Fix testing failure by adding explicit dependency to commons-lang3.

PR #1134, opened 2025-10-27
Add staging table creation support

yili-db:
**PR Checklist**

    Add staging table creation support

    - Add StagingTableDAO, StagingTableRepository and StagingTableService for staging table creation
    - Enhance FileOperations with URI conversion and validation utilities
    - Add ServerProperties for managed table configuration
    - Update schema and catalog repositories with helper methods
    - Add Constants for URI scheme handling

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
+++ b/server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
@@ -65,13 +65,13 @@ public class ServerProperties {
     Map<String, S3StorageConfig> s3BucketConfigMap = new HashMap<>();
     int i = 0;
     while (true) {
-      String bucketPath = properties.getProperty("s3.bucketPath." + i);
-      String region = properties.getProperty("s3.region." + i);
-      String awsRoleArn = properties.getProperty("s3.awsRoleArn." + i);
-      String accessKey = properties.getProperty("s3.accessKey." + i);
-      String secretKey = properties.getProperty("s3.secretKey." + i);
-      String sessionToken = properties.getProperty("s3.sessionToken." + i);
-      String credentialsGenerator = properties.getProperty("s3.credentialsGenerator." + i);
+      String bucketPath = getProperty("s3.bucketPath." + i);
+      String region = getProperty("s3.region." + i);
+      String awsRoleArn = getProperty("s3.awsRoleArn." + i);
+      String accessKey = getProperty("s3.accessKey." + i);
+      String secretKey = getProperty("s3.secretKey." + i);
+      String sessionToken = getProperty("s3.sessionToken." + i);
+      String credentialsGenerator = getProperty("s3.credentialsGenerator." + i);
       if ((bucketPath == null || region == null || awsRoleArn == null)
           && (accessKey == null || secretKey == null || sessionToken == null)) {
         break;
@@ -97,8 +97,8 @@ public class ServerProperties {
     Map<String, String> gcsConfigMap = new HashMap<>();
     int i = 0;
     while (true) {
-      String bucketPath = properties.getProperty("gcs.bucketPath." + i);
-      String jsonKeyFilePath = properties.getProperty("gcs.jsonKeyFilePath." + i);
+      String bucketPath = getProperty("gcs.bucketPath." + i);
+      String jsonKeyFilePath = getProperty("gcs.jsonKeyFilePath." + i);
       if (bucketPath == null || jsonKeyFilePath == null) {
         break;
       }
@@ -114,11 +114,11 @@ public class ServerProperties {
 
     int i = 0;
     while (true) {
-      String storageAccountName = properties.getProperty("adls.storageAccountName." + i);
-      String tenantId = properties.getProperty("adls.tenantId." + i);
-      String clientId = properties.getProperty("adls.clientId." + i);
-      String clientSecret = properties.getProperty("adls.clientSecret." + i);
-      String testMode = properties.getProperty("adls.testMode." + i);
+      String storageAccountName = getProperty("adls.storageAccountName." + i);
+      String tenantId = getProperty("adls.tenantId." + i);
+      String clientId = getProperty("adls.clientId." + i);
+      String clientSecret = getProperty("adls.clientSecret." + i);
+      String testMode = getProperty("adls.testMode." + i);
       if (storageAccountName == null
           || tenantId == null
           || clientId == null
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,52 @@
+name: Backport merged pull request
+
+on:
+  pull_request_target:
+    types: [closed]
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  backport:
+    runs-on: ubuntu-latest
+    name: Backport PR
+    # Only run if the PR was merged (not just closed)
+    if: github.event.pull_request.merged == true
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.BACKPORT_ACTION_TOKEN }}
+
+      - name: Create backport PRs
+        uses: korthout/backport-action@v3
+        with:
+          # Use custom token for backport operations
+          github_token: ${{ secrets.BACKPORT_ACTION_TOKEN }}
+
+          # Match labels like "backport branch-0.3"
+          # The branch name is extracted from the capture group
+          label_pattern: '^backport ([^ ]+)$'
+
+          # Copy the reviewers from the original PR
+          copy_requested_reviewers: true
+
+          # Customize the title of the backport PR
+          pull_title: '[Backport ${target_branch}] ${pull_title}'
+
+          # Customize the description
+          pull_description: |
+            Backport of #${pull_number} to `${target_branch}`.
+
+            ---
+
+            ${pull_description}
+
+          add_author_as_assignee: true
+          experimental: >
+            {
+              "conflict_resolution": "draft_commit_conflicts"
+            }


etc/conf/server.properties
```diff
@@ -23,6 +28,15 @@ server.cookie-timeout=P5D
 #storage-root.models=gs://my-gc-bucket/root
 storage-root.models=file:/tmp/ucroot
 
+# Define the managed table storage root. Cloud storage or file based allowed.
+# If no root specified, defaults to file:///tmp
+# Managed tables will be stored under this root in the path: {storageRoot}/tables/{tableId}
+
+#storageRoot=s3://my-s3-bucket/tables
```
openinx: Thanks for adding this configuration, it's more clear.  Seems like the table does not use the same style that the `model` already used, which introduce a naming inconsistent ?

server/src/main/java/io/unitycatalog/server/persist/dao/StagingTableDAO.java
```diff
@@ -0,0 +1,76 @@
+package io.unitycatalog.server.persist.dao;
+
+import io.unitycatalog.server.model.StagingTableInfo;
+import jakarta.persistence.*;
```
openinx, line 4: Maybe I need to enforce the checkstyle rules so that we won't introduce more `*` package here.

server/src/main/java/io/unitycatalog/server/persist/dao/StagingTableDAO.java
```diff
@@ -0,0 +1,76 @@
...
+        .catalogName(catalogName)
+        .schemaName(schemaName);
+  }
+
+  public void setDefaultFields() {
+    Date now = new Date();
+    setCreatedAt(now); // Assuming current date for creation
+    setAccessedAt(now); // Assuming current date for last access
+    setStageCommitted(false);
+    setStageCommittedAt(null);
```
openinx, line 71: Seems like the `stageCommitedAt` is not nullable ? https://github.com/unitycatalog/unitycatalog/pull/1134/files#diff-aed7cf4e21b6a76835fc042779f60235d3e236398f146f9bcd112a348548c455R44

Is it able to set it to `null` here ?

openinx: Oh, sorry.  It's nullable.

server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
```diff
@@ -81,22 +81,24 @@ public static TableInfoDAO from(TableInfo tableInfo) {
...
+  public TableInfo toTableInfo(boolean fetchColumns, String catalogName, String schemaName) {
     TableInfo tableInfo =
         new TableInfo()
             .tableId(getId().toString())
             .name(getName())
+            .catalogName(catalogName)
+            .schemaName(schemaName)
             .tableType(TableType.valueOf(type))
             .dataSourceFormat(DataSourceFormat.valueOf(dataSourceFormat))
-            .storageLocation(FileOperations.convertRelativePathToURI(url))
```
openinx, line 99: What's the difference for this change ?

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
```diff
@@ -0,0 +1,204 @@
...
+   *
+   * @param createStagingTable the request containing catalog name, schema name, and table name
+   * @return StagingTableInfo containing the created staging table details including the staging
+   *     location
+   * @throws BaseException with ErrorCode.ALREADY_EXISTS if a staging table or regular table with
+   *     the same name already exists in the schema
+   * @throws BaseException with ErrorCode.NOT_FOUND if the specified catalog or schema does not
+   *     exist
+   */
+  public StagingTableInfo createStagingTable(CreateStagingTable createStagingTable) {
```
openinx, line 94: Seems no usage for this method, do we still need it ?

yili-db: The actual usage is in next PR. I have to split this to avoid a bigger than 500 lines PR.

server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
yili-db: It's about the same thing but with subtle differences:
1. the new function checks that URL is from either local or a **supported** cloud.
2. it avoids adding a trailing slash to local path

It's renamed because now it also cares about cloud urls.

etc/conf/server.properties
yili-db: Renamed to storage-root.tables

etc/conf/server.properties
```diff
@@ -23,6 +28,15 @@ server.cookie-timeout=P5D
...
 storage-root.models=file:/tmp/ucroot
 
+# Define the managed table storage root. Cloud storage or file based allowed.
+# If no root specified, defaults to file:///tmp
+# Managed tables will be stored under this root in the path: {storage-root.tables}/tables/{tableId}
+
+#storage-root.tables=s3://my-s3-bucket/tables
+#storage-root.tables=abfs://file_system@account_name.dfs.core.windows.net/tables
+#storage-root.tables=gs://my-gc-bucket/tables
+#storage-root.tables=file:///tmp
```
tdas: should we have a default local path defined like its for models? its will still be disabled by default

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -187,7 +188,7 @@ private URI modifyS3Directory(URI parsedUri, boolean createOrDelete) {
     }
   }
 
-  private static URI adjustFileUri(URI fileUri) {
+  private static URI adjustLocalFileURI(URI fileUri) {
```
tdas, line 196: can add docs on what this method does? what kind of adjustment? maybe we can name it better?

i know this function was already there.. but lets togeter make an attempt to improve the code quality along the way. 🙏

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -175,4 +177,18 @@ public boolean isAuthorizationEnabled() {
     String authorization = getProperty("server.authorization", "disable");
     return authorization.equalsIgnoreCase("enable");
   }
+
+  /**
+   * Check if experimental MANAGED table feature is enabled. This method throws BaseException with
+   * ErrorCode.INVALID_ARGUMENT if it's disabled.
+   */
+  public void checkManagedTableEnabled() {
+    String managedTableEnabled = getProperty("server.managed-table.enabled", "false");
```
tdas, line 188: can defined a VARIABLE name for this string.

tdas: and i a separate PR define constants variables for all the property strings

server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
tdas: thank you for explaining. i had the same question

build.sbt
```diff
@@ -509,6 +511,11 @@ lazy val cli = (project in file("examples") / "cli")
       "com.amazonaws" % "aws-java-sdk-core" % "1.12.728",
       "org.apache.hadoop" % "hadoop-aws" % "3.4.0",
       "org.apache.hadoop" % "hadoop-azure" % "3.4.0",
+      "com.google.cloud.bigdataoss" % "gcs-connector" % "hadoop3-2.2.8" exclude("io.grpc", "grpc-context") exclude("io.grpc", "grpc-api"),
+      "io.grpc" % "grpc-context" % "1.65.1",
+      "io.grpc" % "grpc-api" % "1.65.1",
+      "com.google.cloud" % "google-cloud-core" % "2.41.0",
+      "com.google.api" % "gax-grpc" % "2.48.0",
```
tdas: why are all these build changed needed??

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
yili-db: Will do a follow up PR for this.

build.sbt
yili-db: Removed. These are from the prototype. I guess they may be needed in order to enable the CLI tools to work with cloud files. But I am not sure as I haven't tested against cloud files.

api/Apis/TablesApi.md
```diff
@@ -17,7 +17,7 @@ All URIs are relative to *http://localhost:8080/api/2.1/unity-catalog*
...
+    Creates a new staging table instance. Staging tables are used during managed table creation. Creating a managed table requires performing two actions – initializing the table data in cloud storage and creating the named table entry in the catalog – and these should appear as an atomic operation to other operations on catalog tables. A staging table is used to allocate storage for the managed table, and the catalog_name.schema_name.name parameters provided in this request are used to initialize any required storage properties and determine the storage URL that should be used for the data contained by this table.  Temporary credentials can be obtained as though the staging table were a regular table to get access to the staging table’s storage. After the table’s data is initialized, the staging table is “promoted” to a managed table by creating a managed table with the same location as the staging table. This allows for the atomic creation of a managed table that already has full data written to its storage location. Note: the name provided must match the name used to initialize the staging table originally.  WARNING: This API is experimental and may change in future versions.
```
openinx, line 20: Nit: what's the change for this line ?

etc/conf/server.properties
```diff
@@ -15,6 +15,11 @@ server.redirect-port=
 # D-Days H-Hours M-Minutes S-Seconds (P5D = 5 days,PT5H = 5 hours, PT5M = 5 minutes, PT5S = 5 seconds)
 server.cookie-timeout=P5D
 
+## Experimental Feature Flags
+# Enable MANAGED table (experimental feature)
+# Default: false (disabled)
+server.managed-table.enabled=false
```
openinx, line 21: if I remember correctly, for the catalog owned table seems like we are using the word `catalog-owned` for the table name.  I'm not sure if we need to align with them ? 

https://github.com/search?q=repo%3Adelta-io%2Fdelta%20delta.feature.catalogOwned-preview&type=code

etc/conf/server.properties
```diff
@@ -23,6 +28,15 @@ server.cookie-timeout=P5D
...
 storage-root.models=file:/tmp/ucroot
 
+# Define the managed table storage root. Cloud storage or file based allowed.
+# If no root specified, defaults to file:///tmp
+# Managed tables will be stored under this root in the path: {storage-root.tables}/tables/{tableId}
+
+#storage-root.tables=s3://my-s3-bucket/tables
+#storage-root.tables=abfs://file_system@account_name.dfs.core.windows.net/tables
+#storage-root.tables=gs://my-gc-bucket/tables
+storage-root.tables=file:///tmp/ucroot
```
openinx, line 38: the `storage-root.tables` seems imply that we can configure the multiple paths for different tables , do I understand correctly ?

openinx: Or for the single uc server,  we can only configure one path for the process ?

server/src/main/java/io/unitycatalog/server/persist/dao/StagingTableDAO.java
openinx: Added it into this checkstyle improvement issue: https://github.com/unitycatalog/unitycatalog/issues/1142

server/src/main/java/io/unitycatalog/server/persist/dao/StagingTableDAO.java
```diff
@@ -0,0 +1,76 @@
...
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+@EqualsAndHashCode(callSuper = true)
+@SuperBuilder
+public class StagingTableDAO extends IdentifiableDAO {
+  @Column(name = "schema_id")
+  private UUID schemaId;
+
+  @Column(name = "staging_location", length = 2048, nullable = false)
```
openinx, line 29: Does the spec explicitly state that the field is limited to 2048 bytes ?

openinx: Just wanted to confirm that all the spec are matched.

server/src/main/java/io/unitycatalog/server/persist/dao/StagingTableDAO.java
```diff
@@ -0,0 +1,76 @@
...
+  private String stagingLocation;
+
+  @Column(name = "created_at", nullable = false)
+  private Date createdAt;
```
openinx, line 33: Seems like there are more fields that we did not cover in the specification.  Do we need to align them ? 

https://github.com/unitycatalog/unitycatalog/blob/b4369440ab250f4d6fbacf2d0e1b25273c4ff0d6/api/Models/StagingTableInfo.md?plain=1#L4

server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
```diff
@@ -69,7 +69,7 @@ public class TableInfoDAO extends IdentifiableDAO {
   @Column(name = "uniform_iceberg_metadata_location", length = 65535)
   private String uniformIcebergMetadataLocation;
 
-  public static TableInfoDAO from(TableInfo tableInfo) {
+  public static TableInfoDAO from(TableInfo tableInfo, UUID schemaId) {
```
openinx, line 72: Nit:  maybe it's better to use the `String` here for `schemaId`, since for the `TableInfoDAO#from`  construction, we may don't need to explicitly say that the `schemaId` must be a `UUID` identifier, it could be a just normal unique char sequence.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -196,33 +202,81 @@ private static URI adjustFileUri(URI fileUri) {
...
+   */
+  public static String toStandardizedURIString(String inputPath) {
+    try {
+      // Check if the path is already a URI with a valid scheme
+      URI uri = new URI(inputPath);
+      // If it's a file URI, standardize it
+      if (uri.getScheme() != null) {
+        if (uri.getScheme().equals(Constants.URI_SCHEME_FILE)) {
+          return adjustLocalFileURI(uri).toString();
+        } else if (Constants.SUPPORTED_CLOUD_SCHEMES.contains(uri.getScheme())) {
```
openinx: Why do we need to limit the support for only `gcs`, `s3`,  `abfs`, `abfss` here , for a general and common standardizedURI method ? 

Seems like we are mixing the standardization and cloud schemes checking in the same method.....

openinx: It may also hard to debug if we want to support the `s3a` in future, since we need to add the hard code in this `SUPPORTED_CLOUD_SCHEMES`.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -196,33 +202,81 @@ private static URI adjustFileUri(URI fileUri) {
...
+          return uri.toString();
+        } else {
+          throw new BaseException(
+              ErrorCode.INVALID_ARGUMENT, "Unsupported URI scheme: " + uri.getScheme());
+        }
+      }
+    } catch (URISyntaxException e) {
+      // Not a valid URI, treat it as a file path
```
openinx: Why need to treat it as file path ?  for this exception,  it can be any type of parse error, right ?   seems like we cannot unify all the error to the normal file path ...

openinx: Why need to treat it as file path ?  for this exception,  it can be any type of parse error, right ?   seems like we cannot unify all the error to the normal file path ...

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -196,33 +202,81 @@ private static URI adjustFileUri(URI fileUri) {
...
+      localUri = localUri.substring(0, localUri.length() - 1);
     }
+    return localUri;
   }
 
-  public static void assertValidLocation(String location) {
-    validateURI(URI.create(location));
+  private String getManagedTablesStorageRoot() {
+    // Use local tmp directory as default storage root
+    return serverProperties.getProperty("storage-root.tables", "file:///tmp");
```
openinx: Do we need to align the default value that we've declared here : https://github.com/unitycatalog/unitycatalog/pull/1134/files#diff-903b566690a83bff41179c53cc9eca441e8bd8d371e4ddfb028ce9fe0ee30af8R38 ?

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
```diff
@@ -80,4 +80,20 @@ public static CatalogInfoDAO getCatalogByIdOrThrow(Session session, UUID catalog
     }
     return catalogInfoDAO;
   }
+
+  public static Pair<String, String> getCatalogAndSchemaNames(Session session, UUID schemaId) {
+    SchemaInfoDAO schemaInfoDAO =
```
openinx: Nit:  our checkstyle seems need to make a line as long as possible , so that it reach the 100 chars. 

```java
SchemaInfoDAO schemaInfoDAO = session.get(SchemaInfoDAO.class, schemaId);
    if (schemaInfoDAO == null) {
      throw new BaseException(
          ErrorCode.NOT_FOUND, "Schema not found: " + schemaId);
    }
```

Let me consider how to enforce it.

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
```diff
@@ -80,4 +80,20 @@ public static CatalogInfoDAO getCatalogByIdOrThrow(Session session, UUID catalog
...
+            session.get(SchemaInfoDAO.class, schemaId);
+    if (schemaInfoDAO == null) {
+      throw new BaseException(
+              ErrorCode.NOT_FOUND, "Schema not found: " + schemaId);
```
openinx: Do we need to make it more clear  ?  the following schemaId is actually a schema id , not schema name or something.

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
```diff
@@ -80,4 +80,20 @@ public static CatalogInfoDAO getCatalogByIdOrThrow(Session session, UUID catalog
...
+    }
+    CatalogInfoDAO catalogInfoDAO =
+            session.get(CatalogInfoDAO.class, schemaInfoDAO.getCatalogId());
+    if (catalogInfoDAO == null) {
+      throw new BaseException(
+              ErrorCode.NOT_FOUND, "Catalog not found: " + schemaInfoDAO.getCatalogId());
```
openinx: Ditto

server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
```diff
@@ -122,6 +123,14 @@ public CatalogInfoDAO getCatalogDAO(Session session, String name) {
     return query.uniqueResult();
   }
 
+  public UUID getCatalogId(Session session, String catalogName) {
+    CatalogInfoDAO catalogInfo = getCatalogDAO(session, catalogName);
+    if (catalogInfo == null) {
+      throw new BaseException(ErrorCode.NOT_FOUND, "Catalog not found: " + catalogName);
```
openinx, line 129: Ditto.

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
openinx: Got it, thanks for the explannation.

api/Apis/TablesApi.md
yili-db: That's the will->may change to address the comment on the last PR.

etc/conf/server.properties
yili-db: catalogOwned means using UC as commit coordinator.
Here the managed table only means that the path is managed by UC.

etc/conf/server.properties
yili-db: storage-root.tables specify a storage root for **all** managed tables. Each table will be placed at `{storage-root.tables}/tables/{tableId}`
With this change only one storage root for all managed tables can be specified for a single UC server.

server/src/main/java/io/unitycatalog/server/persist/dao/StagingTableDAO.java
yili-db: Where can I find the spec?
This limit was just copied from TableInfoDAO.url

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
yili-db: This function tries to find a schema (and its name) by id. At this stage if the schema itself is missing, only the id can be known.

server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
yili-db: Here it's indeed a catalog name.

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
yili-db: This error message isn't useful to end user. This internal error reflects an inconsistent state of the database, not user error, and is only useful for debugging.

api/Apis/TablesApi.md
tdas: thank you for addressing this.

build.sbt
```diff
@@ -311,6 +311,8 @@ lazy val server = (project in file("server"))
     ) ++ javacRelease17,
     libraryDependencies ++= Seq(
       "com.linecorp.armeria" %  "armeria" % "1.28.4",
+      "org.apache.commons" % "commons-lang3" % "3.19.0",
+
       // Netty dependencies
       "io.netty" % "netty-all" % "4.1.111.Final",
       "jakarta.annotation" % "jakarta.annotation-api" % "3.0.0" % Provided,
```
tdas, line 318: there is no test for the new functionality in this PR?
can you not test this until the changes in the 3rd PR? 
how much extra LOC will be in this PR IF we merge the changes with tests in that PR into this PR?

yili-db: The majority changes of this PR is in StagingTableDAO.java and StagingTableRepository.java. They are essentially dead code until the next PR is in and until a test actually turns them on with server.managed-table.enabled=true.
In the code base there isn't usually a test for repository or dao under io/unitycatalog/server/persist. But instead the actual service handler will be tested via sdk or cli calling. In order to test these logics the next PR of adding service handler will be needed.
On the other hand I do see a small test `FileOperationsTest.java` from next PR that can be moved here if LOC of this PR isn't concern as it's already slightly above 500 lines. This small test targets `FileOperations.toStandardizedURIString` introduced here.

server/src/main/java/io/unitycatalog/server/persist/dao/StagingTableDAO.java
yili-db: This DAO is in database and is internal implementation. The fields like `createdAt` and `createdBy` are generated and changed by server automatically. Three fields `purgeState`, `numCleanupRetries` and `lastCleanupAt` are reserved for future use when/if we eventually implement managed table deletion/purging.

server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
yili-db: schema ID is **always** a UUID. The ID is never specified by user, but only generated by server. We have `IdentifiableDAO.id` for catalog, schema and table etc IDs.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: I don't have much context on `s3a` vs `s3`. Based on the existing code I just guess that server might expect s3:// URLs in general but client can feel free to use s3a to access it:
```
  public static String substituteSchemeForS3(String tablePath) {
    return tablePath.replace("s3://", "s3a://");
  }
```
Still this is a mechanism to stop other cloud usage until it's actually positively confirmed to be supported. So I think it's still beneficial. This function is just a convenient point to enforce this check. Otherwise it might be repeating everywhere to call both functions.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: Certain valid local path can't be parsed as URI. For example:
```
new URI("/def%2")
|  Exception java.net.URISyntaxException: Malformed escape pair at index 4: /def%2
```
In this case a special handling is needed:
```
jshell> Paths.get("/def%2").toUri().toString()
$15 ==> "file:///def%252"
```

yili-db: This is only a best effort guess. A valid local path can get to this code path. So all errors are treated as a local path. If it's actually intended as a malformed cloud URL, we won't know for sure.
Like for this example: "c:/bucket%2". We can't know if it's a malformed cloud URL or a valid local path.

build.sbt
yili-db: Moved FileOperationsTest.java to this PR: 
https://github.com/unitycatalog/unitycatalog/pull/1134/files#diff-d96d45ab8b0f367bdbe7d74312b6866a6832e080d9fb6ab94ee3c3d2481e14fa

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
yili-db: A follow up PR https://github.com/unitycatalog/unitycatalog/pull/1144/commits (with this extra commit https://github.com/unitycatalog/unitycatalog/commit/1f844310b0b0b3520d75d4be1ff4e234a20cc5fa ) is sent

etc/conf/server.properties
tdas: yes. exactly.

tdas: actually... slight nuance. 
- from the delta protocol perspective catalogOwned means its own by some catalog... UC or otherwise.. and changes to the table needs to coordinated through the catalog, not by the usual filesystem atomic put if absent.
- UC Managed tables is going to be designed using "catalogOwned" delta tables.

build.sbt
```diff
@@ -311,6 +311,8 @@ lazy val server = (project in file("server"))
     ) ++ javacRelease17,
     libraryDependencies ++= Seq(
       "com.linecorp.armeria" %  "armeria" % "1.28.4",
+      "org.apache.commons" % "commons-lang3" % "3.19.0",
```
tdas, line 314: what is using this?

tdas:
why are tests failing?

build.sbt
yili-db: This is not new usage. But existing test is failing on main branch. I had another PR fixing it and [this comment](https://github.com/unitycatalog/unitycatalog/pull/1129#issuecomment-3439609022) describes the failure point.
If you prefer this can be fixed in that PR.

yili-db:
> why are tests failing?

The failed test integrations/langchain/tests/test_langchain_toolkit_oss.py::test_langgraph_agents is failing to configure credential on **main branch**. I have notified the ai maintainer.

etc/conf/server.properties
yili-db: So here can I assume that this type of table, using UC assigned storage path and using catalog as commit coordinator, will be advertised as **MANAGED** table as its brand name?
If yes, this switch will be used for both code paths.

openinx: Okay, then `managed table` is a user-facing product concept, while the `catalogOwned` seems like a technical concept, which specify who will coordinate the commit process. 

Then here we defintely should use `managed table`.

etc/conf/server.properties
openinx: Make sense, thanks for the explanation !

server/src/main/java/io/unitycatalog/server/persist/dao/StagingTableDAO.java
openinx: I also don't know, just want to cofirm.  Maybe we can leave it here, and open a issue to review all of the field limitation for all API.

server/src/main/java/io/unitycatalog/server/persist/dao/StagingTableDAO.java
openinx: Got it, those are internal fields, which won't expose to user.  Thanks for the explanation !

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
openinx: In that case,  will think it should be naturelly throw it out ?  Seems like it's an invalid path string for this method to standarnize. 

Also it's one kind of the exception that you mentioned, which may is possible to fallback to local file path.  There are all kinds of exceptions that we may also don't know,  we all fallback to local file path, which seems to be incorrect.

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  /**
+   * Retrieves a staging table by its unique identifier.
+   *
+   * @param stagingTableId the UUID of the staging table to retrieve
+   * @return StagingTableInfo containing the staging table details
+   * @throws BaseException with ErrorCode.NOT_FOUND if the staging table does not exist
+   */
+  public StagingTableInfo getStagingTableById(String stagingTableId) {
```
openinx: Could you please remove this in this PR, if we neither have any usage for it or test for it ? Because it seems unrelated to the changes.

If it's for the next PR, then it will be better to introduce in the next PR. since we can have test for that.

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
```diff
@@ -0,0 +1,204 @@
...
+                      createStagingTable.getSchemaName());
+          validateIfAlreadyExists(session, schemaId, createStagingTable.getName(), stagingLocation);
+
+          StagingTableDAO stagingTableDAO = new StagingTableDAO();
+          stagingTableDAO.setDefaultFields();
+          stagingTableDAO.setId(stagingTableId);
+          stagingTableDAO.setSchemaId(schemaId);
+          stagingTableDAO.setName(createStagingTable.getName());
+          stagingTableDAO.setStagingLocation(stagingLocation);
+          stagingTableDAO.setCreatedBy(callerId);
```
openinx, line 120: Do we also need to set other non-nullable fields also ? I'm not sure, just question.

yili-db: They are set in `stagingTableDAO.setDefaultFields();`

yili-db:
I have shuffled changes between this PR and next PRs. Now this PR contains the service RPC handler of staging table creation along with RPC test of staging table creation. Please take another look.

etc/conf/server.properties
tdas: Absolutely right. "managed table" is user-facing UC-specific product concept, and catalogOwned is a internal implementation detail of how managed table implemented using delta's catalogOwned protocol. So for a UC server configuration, we should use "managed table"

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
```diff
@@ -80,4 +80,20 @@ public static CatalogInfoDAO getCatalogByIdOrThrow(Session session, UUID catalog
     }
     return catalogInfoDAO;
   }
+
+  public static Pair<String, String> getCatalogAndSchemaNames(Session session, UUID schemaId) {
```
openinx: Nit: Seems this method is useless for this PR, then I think we can remove it in this PR.

server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableCRUDTest.java
```diff
@@ -78,4 +89,160 @@ public void testListTablesWithNextPageTokenShouldReturnNextPageToken() throws Ex
...
+    assertThatExceptionOfType(ApiException.class)
+        .isThrownBy(() -> localTablesApi.createStagingTable(createStagingTableRequest))
+        .satisfies(
+            ex -> assertThat(ex.getCode()).isEqualTo(ErrorCode.NOT_FOUND.getHttpStatus().code()))
+        .withMessageContaining("not found");
+  }
+
+  /** Test that staging table creation fails when the catalog doesn't exist. */
+  @Test
+  public void testStagingTableCreationWithNonExistentCatalogShouldFail() throws Exception {
```
openinx: Nit:  we can remove this `throws Exception`, since no exception will be thrown.

openinx (approved):
Looks good to me overall,  approved.  Let two nit comments.  Thanks @yili-db for the contribution.

openinx:
Seems unrelated failure tests. 

```python
tests/test_langchain_toolkit_oss.py: 40 warnings
  /opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/pyspark/sql/pandas/utils.py:51: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    if LooseVersion(pandas.__version__) < LooseVersion(minimum_pandas_version):

tests/test_langchain_toolkit_oss.py: 40 warnings
  /opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/pyspark/sql/pandas/utils.py:85: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    if LooseVersion(pyarrow.__version__) < LooseVersion(minimum_pyarrow_version):

tests/test_langchain_toolkit_oss.py: 40 warnings
  /opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/pyspark/sql/connect/utils.py:59: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    if LooseVersion(grpc.__version__) < LooseVersion(minimum_grpc_version):

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
FAILED integrations/langchain/tests/test_langchain_toolkit_oss.py::test_langgraph_agents - ValueError: default auth: cannot configure default credentials, please check https://docs.databricks.com/en/dev-tools/auth.html#databricks-client-unified-authentication to configure credentials for your preferred authentication method.
```
https://productionresultssa9.blob.core.windows.net/actions-results/b4310efa-097d-4ec1-9533-bfaf50ae1ec7/workflow-job-run-2032c083-86ae-5af8-b2bf-6111d75c4de8/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-11-04T06%3A16%3A35Z&sig=8Ue5Jhh3EaH3D1djo5sqXENulbz%2FIfg8ZW42ILpQ0ic%3D&ske=2025-11-04T16%3A07%3A36Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-11-04T04%3A07%3A36Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2025-11-04T06%3A06%3A30Z&sv=2025-11-05

Let's try again.

PR #1135, opened 2025-10-27
Support managed table creation

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1135/files) to review incremental changes.
- [**staging_table_service**](https://github.com/unitycatalog/unitycatalog/pull/1135) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1135/files)]
  - [refactor_base_table_crud_test](https://github.com/unitycatalog/unitycatalog/pull/1153) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1153/files/90c0f2a454c595eeb19df146f38a74e404de61b9..9e82c72fb64a8f5b50a11e0558b01ebb86f906a5)]
    - [staging_table_cli_test](https://github.com/unitycatalog/unitycatalog/pull/1136) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1136/files/9e82c72fb64a8f5b50a11e0558b01ebb86f906a5..b7cbb71bbf046d9926394052258e625b9921d303)]

---------
**PR Checklist**

    Support managed table creation

    - Update TableService for creation of managed tables
    - Update API for CreateTable to make storage_location optional for managed tables
    - Enhance TemporaryTableCredentialsService with staging table support
    - Update KeyMapper to also check staging tables
    - Update repository classes to use new helper methods

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
@@ -6,7 +6,6 @@ import io.unitycatalog.server.model.CreateStagingTable;
 import io.unitycatalog.server.model.StagingTableInfo;
 import io.unitycatalog.server.persist.dao.StagingTableDAO;
 import io.unitycatalog.server.persist.dao.TableInfoDAO;
-import io.unitycatalog.server.persist.utils.RepositoryUtils;
 import io.unitycatalog.server.persist.utils.TransactionManager;
 import io.unitycatalog.server.utils.IdentityUtils;
 import io.unitycatalog.server.utils.ServerProperties;
@@ -14,7 +13,6 @@ import io.unitycatalog.server.utils.ValidationUtils;
 import java.util.Date;
 import java.util.Objects;
 import java.util.UUID;
-import org.apache.commons.lang3.tuple.Pair;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.query.Query;
@@ -32,33 +30,6 @@ public class StagingTableRepository {
     this.repositories = repositories;
     this.sessionFactory = sessionFactory;
     this.serverProperties = serverProperties;
-  }
-
-  /**
-   * Retrieves a staging table by its unique identifier.
-   *
-   * @param stagingTableId the UUID of the staging table to retrieve
-   * @return StagingTableInfo containing the staging table details
-   * @throws BaseException with ErrorCode.NOT_FOUND if the staging table does not exist
-   */
-  public StagingTableInfo getStagingTableById(String stagingTableId) {
-    LOGGER.debug("Getting staging table by id: {}", stagingTableId);
-    return TransactionManager.executeWithTransaction(
-        sessionFactory,
-        session -> {
-          StagingTableDAO stagingTableDAO =
-              session.get(StagingTableDAO.class, UUID.fromString(stagingTableId));
-          if (stagingTableDAO == null) {
-            throw new BaseException(
-                ErrorCode.NOT_FOUND, "Staging table not found: " + stagingTableId);
-          }
-          Pair<String, String> catalogAndSchemaNames =
-              RepositoryUtils.getCatalogAndSchemaNames(session, stagingTableDAO.getSchemaId());
-          return stagingTableDAO.toStagingTableInfo(
-              catalogAndSchemaNames.getLeft(), catalogAndSchemaNames.getRight());
-        },
-        "Failed to get staging table by ID",
-        /* readOnly = */ true);
   }
 
   private StagingTableDAO findBySchemaIdAndName(Session session, UUID schemaId, String name) {


api/Models/CreateTable.md
```diff
@@ -9,7 +9,7 @@
 | **table\_type** | [**TableType**](TableType.md) |  | [default to null] |
 | **data\_source\_format** | [**DataSourceFormat**](DataSourceFormat.md) |  | [default to null] |
 | **columns** | [**List**](ColumnInfo.md) | The array of __ColumnInfo__ definitions of the table&#39;s columns. | [default to null] |
-| **storage\_location** | **String** | Storage root URL for external table | [default to null] |
+| **storage\_location** | **String** | Storage root URL for external table | [optional] [default to null] |
```
openinx: Why make this explicitly to be `optional`  ?  maybe, one dumb question,  for `managed table`,   do people have the access to specify the `storage_location` as they required, since it's a `managed` table,  then people should not be able to perceive the `storage_location`.

api/all.yaml
```diff
@@ -1674,7 +1674,6 @@ components:
         - table_type
         - data_source_format
         - columns
-        - storage_location
```
openinx, line 1677: Is this auto-generated, or changed by manual ?  seems like it's easily to be broken, I mean forgot to change this, and leave an inconsistent api definition.

yili-db: This is intentional.
Previously creating an EXTERNAL always required a storage_location.
With MANAGED tables there are two new ways:
1. Create a staging table, then specify TableType=MANAGED and storage_location="staging_table_url" here to create a table
2. Skip the staging table part, just create table with TableType=MANAGED and storage_location=null.

Both ways would end up with a managed table with valid managed URL.

server/src/main/java/io/unitycatalog/server/auth/decorator/KeyMapper.java
```diff
@@ -56,20 +54,41 @@ public Map<SecurableType, Object> mapResourceKeys(Map<SecurableType, Object> res
...
+          // Check if this is actually a request for staging table. Staging tables are only
+          // addressed via their IDs.
+          StagingTableInfo stagingTable = stagingTableRepository.getStagingTableById(fullName);
+          catalogName = stagingTable.getCatalogName();
+          schemaName = stagingTable.getSchemaName();
+          tableId = stagingTable.getId();
+        } else {
+          throw e;
+        }
+      }
```
openinx: Will it be more clear if we just seperate the `id-based` selection and `name-based` selection here ? 

```java
      if (isTableName) {
        TableInfo table = tableRepository.getTable(fullName);
        catalogName = table.getCatalogName();
        schemaName = table.getSchemaName();
        tableId = table.getTableId();
      } else {
        try {
          TableInfo table = tableRepository.getTableById(fullName);
          catalogName = table.getCatalogName();
          schemaName = table.getSchemaName();
          tableId = table.getTableId();
        } catch (BaseException e) {
          if (e.getErrorCode() == ErrorCode.NOT_FOUND) {
            StagingTableInfo stagingTable = stagingTableRepository.getStagingTableById(fullName);
            catalogName = stagingTable.getCatalogName();
            schemaName = stagingTable.getSchemaName();
            tableId = stagingTable.getId();
          } else {
            throw e;
          }
        }
      }
```

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
```diff
@@ -79,4 +80,20 @@ public static CatalogInfoDAO getCatalogByIdOrThrow(Session session, UUID catalog
     }
     return catalogInfoDAO;
   }
+
+  public static Pair<String, String> getCatalogAndSchemaNames(Session session, UUID schemaId) {
+    SchemaInfoDAO schemaInfoDAO =
```
openinx: nit: could you plesae help to make it to be a full line ? 

```
SchemaInfoDAO schemaInfoDAO = session.get(SchemaInfoDAO.class, schemaId);
```

server/src/main/java/io/unitycatalog/server/persist/Repositories.java
```diff
@@ -32,7 +32,7 @@ public Repositories(SessionFactory sessionFactory, ServerProperties serverProper
 
     this.catalogRepository = new CatalogRepository(this, sessionFactory);
     this.schemaRepository = new SchemaRepository(this, sessionFactory);
-    this.tableRepository = new TableRepository(this, sessionFactory);
+    this.tableRepository = new TableRepository(this, sessionFactory, serverProperties);
```
openinx, line 35: Passing the whole `serverProperties` to the `TableRepository` seems strange to me, because the database layer should not be aware of the whole server-side properties.  A better option is:  use passing the `managedTableEnabled`  boolean switch to the database layer.

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
```diff
@@ -125,4 +156,49 @@ public StagingTableInfo createStagingTable(CreateStagingTable createStagingTable
...
+      throw new BaseException(
+          ErrorCode.FAILED_PRECONDITION, "Staging table already committed: " + storageLocation);
+    }
+
+    // Commit the staging table
+    Date now = new Date();
+    stagingTableDAO.setStageCommitted(true);
+    stagingTableDAO.setStageCommittedAt(now);
+    stagingTableDAO.setAccessedAt(now);
+    session.merge(stagingTableDAO);
```
openinx, line 172: What will happen if two contention competitors are committing the same `stagingTable` at the same time  ?

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -27,13 +28,38 @@ public class TableRepository {
...
     this.sessionFactory = sessionFactory;
     this.fileOperations = repositories.getFileOperations();
+    this.serverProperties = serverProperties;
+  }
+
+  public String getStorageLocationForTableOrStagingTable(String tableId) {
+    return TransactionManager.executeWithTransaction(
+        sessionFactory,
+        session -> {
+          LOGGER.debug("Getting storage location of table by id: {}", tableId);
```
openinx, line 71: nit:  Please left a blank line to separate the different code blocks, which make it will be more easier to read.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -27,13 +28,38 @@ public class TableRepository {
...
-  public TableRepository(Repositories repositories, SessionFactory sessionFactory) {
+  public TableRepository(
+      Repositories repositories, SessionFactory sessionFactory, ServerProperties serverProperties) {
     this.repositories = repositories;
     this.sessionFactory = sessionFactory;
     this.fileOperations = repositories.getFileOperations();
+    this.serverProperties = serverProperties;
+  }
+
+  public String getStorageLocationForTableOrStagingTable(String tableId) {
```
openinx: Why do we need to mix both the normal table access and staging table access in the same method ?  because the table id can always belongs to either specified normal table or staging table ?  

( this make the code some kind of complicated --- since someone may forget to access the staging table,  it's possible to use some abstraction to avoid other forgetting access the staging table ? )

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -45,23 +71,10 @@ public TableInfo getTableById(String tableId) {
...
-          }
-          TableInfo tableInfo =
-              tableInfoDAO.toTableInfo(true, catalogInfoDAO.getName(), schemaInfoDAO.getName());
-          tableInfo.setSchemaName(schemaInfoDAO.getName());
-          tableInfo.setCatalogName(catalogInfoDAO.getName());
-          return tableInfo;
+          Pair<String, String> catalogAndSchemaNames =
+              RepositoryUtils.getCatalogAndSchemaNames(session, tableInfoDAO.getSchemaId());
+          return tableInfoDAO.toTableInfo(
+              true, catalogAndSchemaNames.getLeft(), catalogAndSchemaNames.getRight());
```
openinx: for the `fetchColumns=true`, is it lazily fetching or sync fetching ?

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -112,49 +126,77 @@ public TableInfo createTable(CreateTable createTable) {
...
+                  ErrorCode.INVALID_ARGUMENT, "Storage location is required for external table");
+            }
+            tableID = UUID.randomUUID().toString();
+            storageLocation = createTable.getStorageLocation();
+          } else if (tableType == TableType.MANAGED) {
+            serverProperties.checkManagedTableEnabled();
+            if (createTable.getDataSourceFormat() != DataSourceFormat.DELTA) {
+              throw new BaseException(
+                  ErrorCode.INVALID_ARGUMENT,
+                  "MANAGED table creation is only supported for delta tables.");
```
openinx: Is this temporarly support delta only, or future we will support other table types ?

server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
```diff
@@ -53,34 +53,36 @@ public TemporaryCredentials vendCredential(CredentialContext context) {
...
-                .oauthToken(gcpToken.getTokenValue()))
-            .expirationTime(gcpToken.getExpirationTime().getTime());
-      }
-      case URI_SCHEME_S3 -> {
-        Credentials awsSessionCredentials = vendAwsCredential(context);
-        temporaryCredentials.awsTempCredentials(new AwsCredentials()
-            .accessKeyId(awsSessionCredentials.accessKeyId())
-            .secretAccessKey(awsSessionCredentials.secretAccessKey())
-            .sessionToken(awsSessionCredentials.sessionToken()));
+    if (storageScheme != null) {
```
openinx, line 58: When will the `storageScheme` to be nullable ?

server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableCRUDTest.java
```diff
@@ -119,8 +128,92 @@ public void testStagingTableCreationAndManagedTableFromStagingLocation() throws
     assertThat(stagingTableInfo.getStagingLocation())
         .isEqualTo("file:///tmp/ucroot/tables/" + stagingTableInfo.getId());
 
-    // Following steps: Create a managed table using the staging location
-    // TODO: This is not implemented yet.
+    // Step 2: Create a managed table using the staging location
```
openinx, line 140: Seems like few tests, compared to complicated code changes.  Maybe we need more test coverage ?

api/all.yaml
yili-db: This file is manually changed.

api/Models/CreateTable.md
yili-db: It's `optional` once it's removed from the `requried` section in `all.yaml`.
The reason is explained here: https://github.com/unitycatalog/unitycatalog/pull/1135#discussion_r2491665339

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
yili-db: The `session` is provided by TransactionManager.executeWithTransaction(... session -> {...})
So it's protected by a DB transaction. This is the same mechanism used for all DB operations in this repo.

server/src/main/java/io/unitycatalog/server/persist/Repositories.java
yili-db: It wants to call a function `checkManagedTableEnabled()` from serverProperties.
If a boolean from `serverProperties` is passed into `TableRepository`, `tableRepository` and `stagingTableRepository` will have to check the boolean by themselves.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: Do you mean before this line? Or before line 53 the next LOGGER.debug?

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: Do you meaning integrating the staging table access into `getTableById` to make it impossible to miss staging table?
The problem is that staging table and table are very different. Staging table is very incomplete that it lacks many essential metadata. In most code paths they should **not** care about staging table. There are only a limited number of scenarios that should access staging tables:
1. creating a MANAGED table
2. vending a table credential when it's still a staging table
3. auth of vending a table credential when it's still a staging table

The other code paths of normal table operations should **skip** staging tables even if the ID/name matches
Mixing up tables and staging tables will be more error prone.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: It's lazily fetched. But this part is unchanged in this PR anyway. It's just a little code refactor to make it shorter and cleaner.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: I don't know about the future. But for now we just support delta tables.

server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
yili-db: If it happens to be a local path like `/tmp/something`

server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableCRUDTest.java
yili-db: This is just the basic api test.
I have more test lined up for review in next PRs:
Adding CLI support then test with CLI:
https://github.com/unitycatalog/unitycatalog/pull/1136/commits/5f24c7bfabb83f6bdc9ad744ae6d0dab7ee3b1ed
Access control test:
https://github.com/unitycatalog/unitycatalog/pull/1138/commits/3263ae496b8aae6508c65ded32dfd5fbbc320d56

api/all.yaml
yili-db: @tdas decided that we don't want to support the 2nd code path. So I am reverting this line.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
openinx: Thanks for the context, I understand the staging table is a temporary status which is different from the normal table.  Then the question is:  how can we clearly org the code or comment , so that the following authors can easily got the points and context from your side, and finally they won't mistake when should we access the staging table and when should not.  I think that's the key point.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
openinx: Yea,  I think the intention blank line separate the code block, which makes easily to read.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
openinx: Got it, thanks.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
openinx: Okay,  I just try to understand if we want to support further tables, then how could we do the changes to support it?  Seems like currently we don't have the abstraction to support different types of managed table.....

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
```diff
@@ -79,4 +80,18 @@ public static CatalogInfoDAO getCatalogByIdOrThrow(Session session, UUID catalog
     }
     return catalogInfoDAO;
   }
+
+  public static Pair<String, String> getCatalogAndSchemaNames(Session session, UUID schemaId) {
```
tdas, line 96: this is doing two database query right?

it would be good add this to the methods docs.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: Added method doc to clarify the use cases of staging table.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: We can revise this when different types of managed table is to be supported.

server/src/main/java/io/unitycatalog/server/auth/decorator/KeyMapper.java
```diff
@@ -56,22 +57,37 @@ public Map<SecurableType, Object> mapResourceKeys(Map<SecurableType, Object> res
...
-      resourceIds.put(TABLE, UUID.fromString(table.getTableId()));
-      resourceIds.put(SCHEMA, UUID.fromString(schema.getSchemaId()));
-      resourceIds.put(CATALOG, UUID.fromString(catalog.getId()));
+      boolean isTableName = fullName.contains(".");
+
+      UUID catalogId;
+      UUID schemaId;
+      UUID tableId;
+      if (isTableName) {
+        TableInfo table = tableRepository.getTable(fullName);
```
yili-db, line 79: Line 72-78 is basically doing the same thing as old code line 67,69-74.

server/src/main/java/io/unitycatalog/server/auth/decorator/KeyMapper.java
```diff
@@ -56,22 +57,37 @@ public Map<SecurableType, Object> mapResourceKeys(Map<SecurableType, Object> res
...
+        String fullSchemaName = table.getCatalogName() + "." + table.getSchemaName();
+        SchemaInfo schema = schemaRepository.getSchema(fullSchemaName);
+        CatalogInfo catalog = catalogRepository.getCatalog(table.getCatalogName());
+        catalogId = UUID.fromString(catalog.getId());
+        schemaId = UUID.fromString(schema.getSchemaId());
+        tableId = UUID.fromString(table.getTableId());
+      } else {
+        // It may be the ID of either a table or staging table.
+        tableId = UUID.fromString(fullName);
```
yili-db, line 88: This block goes to the new function to get catalog&schema ids if the fullName is considered a table ID.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -27,43 +29,91 @@ public class TableRepository {
...
-  public TableRepository(Repositories repositories, SessionFactory sessionFactory) {
+  public TableRepository(
+      Repositories repositories, SessionFactory sessionFactory, ServerProperties serverProperties) {
     this.repositories = repositories;
     this.sessionFactory = sessionFactory;
     this.fileOperations = repositories.getFileOperations();
+    this.serverProperties = serverProperties;
   }
 
-  public TableInfo getTableById(String tableId) {
```
yili-db, line 47: `getTableById` is no longer needed.
It is replaced by `getCatalogSchemaIdsByTableOrStagingTableId` and `getStorageLocationForTableOrStagingTable`

server/src/main/java/io/unitycatalog/server/persist/FunctionRepository.java
```diff
@@ -144,7 +135,8 @@ public ListFunctionsResponse listFunctions(
     return TransactionManager.executeWithTransaction(
         sessionFactory,
         session -> {
-          UUID schemaId = getSchemaId(session, catalogName, schemaName);
+          UUID schemaId =
+              repositories.getSchemaRepository().getSchemaId(session, catalogName, schemaName);
```
tdas, line 142: super nit: not in this PR, but this seems like a repeated code pattern in many repos that can deduped by having an interface like `SchemaUtils` that every repo extends to get a simple function `getSchemaId(...)`.


can you make this a issue and mark it as a starter task label so that others can pick it up?

server/src/main/java/io/unitycatalog/server/persist/Repositories.java
tdas: i am okay with the server properties being passed as a whole. today there is one property that affects the behavior. tomorrow there may be many more properties. we dont want to keep changing the constructor signature every time. So better in the long run to send the whole serverproperties in the repo

think of it like hadoopConf. hadoopConf passed round everywhere for exactly this reason.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -27,43 +29,94 @@ public class TableRepository {
...
+          throw new BaseException(
+              ErrorCode.NOT_FOUND, "Neither table nor staging table found with id: " + tableId);
+        },
+        "Failed to get storage location of table or staging table",
+        /* readOnly = */ true);
+  }
+
+  /**
+   * Retrieves the schema ID and catalog ID for a table or staging table by its ID. First attempts
+   * to IDs associated with a regular table with the given ID, then falls back to searching for a
```
tdas:
```suggestion
   * to get IDs associated with a regular table with the given ID, then falls back to searching for a
```

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -112,49 +166,66 @@ public TableInfo createTable(CreateTable createTable) {
...
+          // on the type of table to create.
+          String tableID;
+          if (tableType == TableType.EXTERNAL) {
+            tableID = UUID.randomUUID().toString();
+          } else if (tableType == TableType.MANAGED) {
+            serverProperties.checkManagedTableEnabled();
+            if (createTable.getDataSourceFormat() != DataSourceFormat.DELTA) {
+              throw new BaseException(
+                  ErrorCode.INVALID_ARGUMENT,
+                  "MANAGED table creation is only supported for delta tables.");
```
tdas:
```suggestion
                  "Managed table creation is only supported for Delta format.");
```

server/src/main/java/io/unitycatalog/server/persist/FunctionRepository.java
yili-db: Created https://github.com/unitycatalog/unitycatalog/issues/1165

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
```diff
@@ -30,6 +34,33 @@ public StagingTableRepository(
...
   }
 
+  /**
+   * Retrieves a staging table by its unique identifier.
+   *
+   * @param stagingTableId the UUID of the staging table to retrieve
+   * @return StagingTableInfo containing the staging table details
+   * @throws BaseException with ErrorCode.NOT_FOUND if the staging table does not exist
+   */
+  public StagingTableInfo getStagingTableById(String stagingTableId) {
```
openinx: Seems like no usage in this PR ?  Could we just remove it in this PR, and if it's required in the following PR, then I think it's better to have it in the next PR.

server/src/main/java/io/unitycatalog/server/persist/Repositories.java
openinx: Okay, then let's keep it.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -27,43 +29,94 @@ public class TableRepository {
...
+   * during the short window when a staging table is just created and the initial data is being
+   * written but before the actual table is already created. Reading of a staging table is not a
+   * common supplemental of an actual table but only a special case.
+   *
+   * @param tableId the ID of the table or staging table
+   * @return the standardized URI string of the storage location
+   * @throws BaseException with ErrorCode.NOT_FOUND if neither a table nor staging table is found
+   *     with the given ID
+   */
+  public String getStorageLocationForTableOrStagingTable(UUID tableId) {
```
openinx, line 67: Thank you so much for the javadoc, which makes more clear now.

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
yili-db: Removed. I forgot this one when I introduced `getCatalogSchemaIdsByTableOrStagingTableId`

PR #1136, opened 2025-10-27
Add CLI support for MANAGED tables with CLI tests

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1136/files) to review incremental changes.
- [**staging_table_cli_test**](https://github.com/unitycatalog/unitycatalog/pull/1136) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1136/files)]

---------
**PR Checklist**

    Add CLI support for MANAGED tables and comprehensive tests

    - Enhance TableCli with MANAGED table creation
    - Update DeltaKernelUtils with staging table support and writer capabilities
    - Add comprehensive test coverage for managed and staging tables

### DIFF
--- a/examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
+++ b/examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
@@ -151,8 +151,14 @@ public class TableCli {
       DeltaKernelUtils.createDeltaTable(
           createTable.getStorageLocation(), columnInfoList, temporaryCredentials);
     } catch (Exception e) {
-      if (e instanceof TableAlreadyExistsException
-          || e.getCause() instanceof TableAlreadyExistsException) {
+      if ((e instanceof TableAlreadyExistsException
+              || e.getCause() instanceof TableAlreadyExistsException)
+          && createTable.getTableType() == TableType.EXTERNAL) {
+        // A common pattern is that many tests would test the failure cases of table creation. But
+        // they often leave the delta table on disk after failing to register a table in UC. Then
+        // they retry the table creation on the same location. This would allow those tests to
+        // continue. This will only happen to external tables as managed tables would have a unique
+        // new location allocated by UC.
         // TODO confirm the schema of the existing table matches the schema of the new table
       } else {
         throw new CliException(
--- a/examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
+++ b/examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
@@ -62,9 +62,7 @@ public class DeltaKernelUtils {
         throw new RuntimeException("Table creation failed");
       }
     } catch (Exception e) {
-      String errorMsg =
-          String.format("Failed to create Delta table at '%s': %s", tablePath, e.getMessage());
-      LOGGER.error(errorMsg, e);
+      String errorMsg = String.format("Failed to create Delta table at '%s'", tablePath);
       throw new IllegalArgumentException(errorMsg, e);
     }
   }
--- a/examples/cli/src/test/java/io/unitycatalog/cli/table/CliTableCreationTest.java
+++ b/examples/cli/src/test/java/io/unitycatalog/cli/table/CliTableCreationTest.java
@@ -19,6 +19,7 @@ import io.unitycatalog.client.model.ColumnTypeName;
 import io.unitycatalog.client.model.CreateCatalog;
 import io.unitycatalog.client.model.CreateSchema;
 import io.unitycatalog.client.model.CreateTable;
+import io.unitycatalog.client.model.DataSourceFormat;
 import io.unitycatalog.client.model.TableInfo;
 import io.unitycatalog.client.model.TableType;
 import io.unitycatalog.server.base.BaseServerTest;
@@ -42,7 +43,7 @@ public class CliTableCreationTest extends BaseServerTest {
   private SchemaOperations schemaOperations;
   private TableOperations tableOperations;
 
-  private static final List<ColumnInfo> columns =
+  private static final List<ColumnInfo> COLUMNS =
       List.of(
           new ColumnInfo().name("id").typeName(ColumnTypeName.INT).comment("id"),
           new ColumnInfo().name("name").typeName(ColumnTypeName.STRING).comment("name"));
@@ -76,7 +77,7 @@ public class CliTableCreationTest extends BaseServerTest {
   @Test
   public void testCreateTableLocalDirectoryDoesNotExist() throws IOException, ApiException {
     String tablePath = "/tmp/" + UUID.randomUUID();
-    createTableAndAssertReadTableSucceeds(TableType.EXTERNAL, Optional.of(tablePath), columns);
+    createTableAndAssertReadTableSucceeds(TableType.EXTERNAL, Optional.of(tablePath), COLUMNS);
   }
 
   @Test
@@ -85,7 +86,7 @@ public class CliTableCreationTest extends BaseServerTest {
     String tablePath = "/tmp/" + UUID.randomUUID();
     Path dir = Paths.get(tablePath);
     Files.createDirectory(dir);
-    createTableAndAssertReadTableSucceeds(TableType.EXTERNAL, Optional.of(tablePath), columns);
+    createTableAndAssertReadTableSucceeds(TableType.EXTERNAL, Optional.of(tablePath), COLUMNS);
   }
 
   @Test
@@ -95,28 +96,40 @@ public class CliTableCreationTest extends BaseServerTest {
     assertThatCode(
             () ->
                 DeltaKernelUtils.createDeltaTable(
-                    Paths.get(tablePath).toUri().toString(), columns, null))
+                    Paths.get(tablePath).toUri().toString(), COLUMNS, null))
         .doesNotThrowAnyException();
     assertThat(Paths.get(tablePath + "/_delta_log")).isDirectory();
-    createTableAndAssertReadTableSucceeds(TableType.EXTERNAL, Optional.of(tablePath), columns);
+    createTableAndAssertReadTableSucceeds(TableType.EXTERNAL, Optional.of(tablePath), COLUMNS);
   }
 
   @Test
   public void testCreateManagedTable() throws IOException, ApiException {
-    createTableAndAssertReadTableSucceeds(TableType.MANAGED, Optional.empty(), columns);
+    createTableAndAssertReadTableSucceeds(TableType.MANAGED, Optional.empty(), COLUMNS);
   }
 
   @Test
-  public void testCreateManagedTableWithLocationAndFail() throws IOException, ApiException {
+  public void testCreateManagedTableAndFail() throws IOException, ApiException {
+    // This fails due to setting a storage location for managed table.
     assertThat(
             tableOperations.createTable(
                 new CreateTable()
                     .catalogName(CATALOG_NAME)
                     .schemaName(SCHEMA_NAME)
-                    .name(TABLE_NAME)
+                    .name(TABLE_NAME + "_1")
                     .tableType(TableType.MANAGED)
                     .storageLocation("/tmp/some_path")
-                    .columns(columns)))
+                    .columns(COLUMNS)))
+        .isNull();
+    // This fails due to setting PARQUET. Only DELTA is supported for managed table right now.
+    assertThat(
+            tableOperations.createTable(
+                new CreateTable()
+                    .catalogName(CATALOG_NAME)
+                    .schemaName(SCHEMA_NAME)
+                    .name(TABLE_NAME + "_2")
+                    .tableType(TableType.MANAGED)
+                    .dataSourceFormat(DataSourceFormat.PARQUET)
+                    .columns(COLUMNS)))
         .isNull();
   }
 
--- a/server/src/test/java/io/unitycatalog/server/base/table/BaseTableCRUDTestEnv.java
+++ b/server/src/test/java/io/unitycatalog/server/base/table/BaseTableCRUDTestEnv.java
@@ -37,6 +37,25 @@ public abstract class BaseTableCRUDTestEnv extends BaseCRUDTest {
 
   protected abstract TableOperations createTableOperations(ServerConfig serverConfig);
 
+  protected static final List<ColumnInfo> COLUMNS =
+      List.of(
+          new ColumnInfo()
+              .name("as_int")
+              .typeText("INTEGER")
+              .typeJson("{\"type\": \"integer\"}")
+              .typeName(ColumnTypeName.INT)
+              .position(0)
+              .comment("Integer column")
+              .nullable(true),
+          new ColumnInfo()
+              .name("as_string")
+              .typeText("VARCHAR(255)")
+              .typeJson("{\"type\": \"string\", \"length\": \"255\"}")
+              .typeName(ColumnTypeName.STRING)
+              .position(1)
+              .comment("String column")
+              .nullable(true));
+
   @BeforeEach
   @Override
   public void setUp() {
@@ -102,32 +121,13 @@ public abstract class BaseTableCRUDTestEnv extends BaseCRUDTest {
     } else {
       assert storageLocation.isPresent();
     }
-    ColumnInfo columnInfo1 =
-        new ColumnInfo()
-            .name("as_int")
-            .typeText("INTEGER")
-            .typeJson("{\"type\": \"integer\"}")
-            .typeName(ColumnTypeName.INT)
-            .position(0)
-            .comment("Integer column")
-            .nullable(true);
-
-    ColumnInfo columnInfo2 =
-        new ColumnInfo()
-            .name("as_string")
-            .typeText("VARCHAR(255)")
-            .typeJson("{\"type\": \"string\", \"length\": \"255\"}")
-            .typeName(ColumnTypeName.STRING)
-            .position(1)
-            .comment("String column")
-            .nullable(true);
 
     CreateTable createTableRequest =
         new CreateTable()
             .name(tableName)
             .catalogName(TestUtils.CATALOG_NAME)
             .schemaName(TestUtils.SCHEMA_NAME)
-            .columns(List.of(columnInfo1, columnInfo2))
+            .columns(COLUMNS)
             .properties(TestUtils.PROPERTIES)
             .comment(TestUtils.COMMENT)
             .storageLocation(storageLocation.orElse(null))
--- a/server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableCRUDTest.java
+++ b/server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableCRUDTest.java
@@ -139,7 +139,27 @@ public class SdkTableCRUDTest extends BaseTableCRUDTest {
     assertThat(stagingTableInfo.getStagingLocation())
         .isEqualTo("file:///tmp/ucroot/tables/" + stagingTableInfo.getId());
 
-    // Step 2: Create a managed table using the staging location
+    // Step 2: Create a managed table that's not DELTA
+    CreateTable createTableRequestNotDelta =
+        new CreateTable()
+            .name(stagingTableName)
+            .catalogName(TestUtils.CATALOG_NAME)
+            .schemaName(TestUtils.SCHEMA_NAME)
+            .columns(columns)
+            .tableType(TableType.MANAGED)
+            .dataSourceFormat(DataSourceFormat.PARQUET)
+            .storageLocation(stagingTableInfo.getStagingLocation())
+            .comment("Table created from staging location");
+    // This should fail with INVALID_ARGUMENT
+    assertThatExceptionOfType(ApiException.class)
+        .isThrownBy(() -> localTablesApi.createTable(createTableRequestNotDelta))
+        .satisfies(
+            ex ->
+                assertThat(ex.getCode())
+                    .isEqualTo(ErrorCode.INVALID_ARGUMENT.getHttpStatus().code()))
+        .withMessageContaining("Managed table creation is only supported for Delta format");
+
+    // Step 3: Create a managed table using the staging location
     CreateTable createTableRequest =
         new CreateTable()
             .name(stagingTableName)


examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
```diff
@@ -90,43 +90,93 @@ private static String createTable(TablesApi apiClient, JSONObject json)
...
   }
 
-  private static void handleTableStorageLocation(
-      String storageLocation, List<ColumnInfo> columnInfos) {
-    if (!(storageLocation.startsWith("s3://")
-        || storageLocation.startsWith("file:/")
-        || storageLocation.startsWith("/"))) {
+  private static String normalizeLocalPath(String storageLocation) {
+    boolean isCloud =
+        storageLocation.startsWith("abfs://")
```
yili-db: define a list somewhere

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
```diff
@@ -163,9 +163,12 @@ public List<CliParams> getOptionalParams() {
                   put(
                       CREATE,
                       new CliOptions(
+                          List.of(CliParams.FULL_NAME, CliParams.COLUMNS),
                           List.of(
-                              CliParams.FULL_NAME, CliParams.COLUMNS, CliParams.STORAGE_LOCATION),
```
yili-db, line 178: `STORAGE_LOCATION` is now optional for MANAGED table.

examples/cli/src/main/java/io/unitycatalog/cli/utils/Constants.java
```diff
@@ -0,0 +1,13 @@
+package io.unitycatalog.cli.utils;
```
yili-db, line 1: This file is copied from `server/src/main/java/io/unitycatalog/server/utils/Constants.java`

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
```diff
@@ -90,44 +92,145 @@ private static String createTable(TablesApi apiClient, JSONObject json)
...
+      // be skipped and only local dir can work:
+      // https://github.com/unitycatalog/unitycatalog/issues/1160
+      temporaryCredentials = null;
+      // temporaryCredentials =
+      //     temporaryCredentialsApi.generateTemporaryPathCredentials(
+      //         new GenerateTemporaryPathCredential()
+      //             .url(storageLocation)
+      //             .operation(PathOperation.PATH_CREATE_TABLE));
+    } else if (createTable.getTableType() == TableType.MANAGED) {
+      // handle delta managed tables
```
yili-db, line 120: This is the major change for creating a `MANAGED` table from CLI.

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
```diff
@@ -90,44 +92,145 @@ private static String createTable(TablesApi apiClient, JSONObject json)
...
+    TableInfo tableInfo = tablesApi.createTable(createTable);
     return objectWriter.writeValueAsString(tableInfo);
   }
 
-  private static Path getLocalPath(String path) {
-    if (path.startsWith("file:")) {
-      return Paths.get(URI.create(path));
-    } else {
-      return Paths.get(path);
+  /**
```
yili-db: `adjustLocalFileURI` and `toStandardizedURIString` are copied from `server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java`

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
```diff
@@ -189,11 +295,13 @@ private static String writeTable(
     }
     String tableId = info.getTableId();
     try {
+      TemporaryCredentials temporaryCredentials =
```
yili-db, line 235: This type of change is to get away from special treatment for S3. Previously these function calls only accepted `AwsCredentials`. Now it's `TemporaryCredentials`.

server/src/test/java/io/unitycatalog/server/base/table/BaseTableCRUDTest.java
```diff
@@ -111,76 +105,6 @@ private void testDeleteTable() throws ApiException {
         .isInstanceOf(Exception.class);
   }
 
-  private void testManagedTableRetrieval() throws ApiException {
```
yili-db, line 119: This is the mock that creates a managed table for testing purpose. It's no longer needed as it's using the real managed table creation now.

server/src/test/java/io/unitycatalog/server/base/table/BaseTableCRUDTest.java
```diff
@@ -6,19 +6,13 @@
 
 import io.unitycatalog.client.ApiException;
```
yili-db, line 7: Change of this test is delay until this PR because it's shared by both SDK and CLI table tests.

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
```diff
@@ -98,44 +101,145 @@ private static String createTable(TablesApi apiClient, JSONObject json)
...
-    if (path.startsWith("file:")) {
-      return Paths.get(URI.create(path));
-    } else {
-      return Paths.get(path);
+  /**
+   * This helper function adjusts local file URI that starts with file:/ or file:// but not with
+   * file:///. This function makes sure these URIs must begin with file:/// in order to be a valid
+   * local file URI.
+   */
+  private static URI adjustLocalFileURI(URI fileUri) {
```
tdas: lets move these utils functions to a separate class / file . like it is in the server code.

server/src/test/java/io/unitycatalog/server/base/table/BaseTableCRUDTestEnv.java
```diff
@@ -59,19 +60,44 @@ private void createCommonResources() {
...
+    assertThat(managedTable.getCreatedAt()).isNotNull();
+    assertThat(managedTable.getTableId()).isNotNull();
+    return managedTable;
+  }
+
   public static TableInfo createTestingTable(
-      String tableName, String storageLocation, TableOperations tableOperations)
+      String tableName,
+      TableType tableType,
+      @Nullable String storageLocation,
```
tdas: why nullable and not optional?

tdas: lets not rely on nulls when java gives native support for optional

examples/cli/src/test/java/io/unitycatalog/cli/table/CliManagedTableCreationTest.java
```diff
@@ -0,0 +1,121 @@
...
+import java.io.IOException;
+import java.net.URI;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.Optional;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class CliManagedTableCreationTest extends BaseServerTest {
```
tdas: is there any opportunity for having a common super class (for sharing code and tests) of CliManagedTableCreationTest and CliExternalTableCreationTest. seems like we want to run almost the exact same set of tests for both external and managed.. and should be doable in a parameterized test in the same test suite.

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
yili-db: Moved to `examples/cli/src/main/java/io/unitycatalog/cli/utils/FileOperations.java`

examples/cli/src/test/java/io/unitycatalog/cli/table/CliManagedTableCreationTest.java
yili-db: Merged both into `CliTableCreationTest`.

examples/cli/src/main/java/io/unitycatalog/cli/utils/Constants.java
openinx: Do you think we need to abstract a common module, and put all the shared classes & utility there, then we don't need to copy everything from server to client.  Also the unitycatalog-spark-connector can use the common module. 

Please file a separate issue to consider this thing, thanks.

examples/cli/src/main/java/io/unitycatalog/cli/utils/FileOperations.java
```diff
@@ -0,0 +1,84 @@
...
+   * "gs://my-bucket/my-file"   -> "gs://my-bucket/my-file"
+   *
+   * // Invalid Path Example (treated as a file path):
+   * "/local/path/to/file"      -> "file:///local/path/to/file"
+   *
+   * // Unsupported Scheme Example:
+   * "ftp://example.com/file"   -> Throws BaseException with message: "Unsupported URI scheme: ftp"
+   * </pre>
+   */
+  public static String toStandardizedURIString(String inputPath) {
```
openinx, line 57: Is this the same as `FileOperation` from the server module ? I really don't want to copy everything that we need to use in client side from the server side..... which is not fixing the tech debt, instead it's producing more tech debt. 

https://github.com/unitycatalog/unitycatalog/blob/fc76ab52aa03550d6577eca7be90004a5d2913b8/server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java#L209-L269

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
```diff
@@ -98,44 +98,70 @@ private static String createTable(TablesApi apiClient, JSONObject json)
...
+      createTable.setStorageLocation(storageLocation);
+      // Currently generateTemporaryPathCredentials doesn't quite work yet. So this step has to
+      // be skipped and only local dir can work:
+      // https://github.com/unitycatalog/unitycatalog/issues/1160
+      temporaryCredentials = null;
+      // temporaryCredentials =
+      //     temporaryCredentialsApi.generateTemporaryPathCredentials(
+      //         new GenerateTemporaryPathCredential()
+      //             .url(storageLocation)
+      //             .operation(PathOperation.PATH_CREATE_TABLE));
```
openinx, line 118: Why just remove this commented code ?

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
```diff
@@ -98,44 +98,70 @@ private static String createTable(TablesApi apiClient, JSONObject json)
...
+      }
+      createTable.setStorageLocation(stagingLocation);
+      temporaryCredentials =
+          temporaryCredentialsApi.generateTemporaryTableCredentials(
+              new GenerateTemporaryTableCredential()
+                  .tableId(stagingTableId)
+                  .operation(TableOperation.READ_WRITE));
     } else {
-      return Paths.get(path);
+      throw new CliException("Unknown table type");
```
openinx: nit:  Could you please also add the `tableType` inside the error message ?

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
```diff
@@ -81,30 +86,33 @@ public static FileSystem getFileSystem(URI tablePathURI, Configuration conf) {
...
+    if (scheme.equals(Constants.URI_SCHEME_S3)) {
+      AwsCredentials awsTempCredentials = temporaryCredentials.getAwsTempCredentials();
+      if (awsTempCredentials == null) {
+        throw new IllegalArgumentException("AWS temporary credentials are missing");
+      }
       conf.set("fs.s3a.access.key", awsTempCredentials.getAccessKeyId());
       conf.set("fs.s3a.secret.key", awsTempCredentials.getSecretAccessKey());
       conf.set("fs.s3a.session.token", awsTempCredentials.getSessionToken());
       conf.set("fs.s3a.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem");
       conf.set("fs.s3a.path.style.access", "true");
```
openinx, line 102: Could you please also file a separate issue, so that we can reuse the credential properties setting with the https://github.com/unitycatalog/unitycatalog/blob/fc76ab52aa03550d6577eca7be90004a5d2913b8/connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java#L19

I think the credentials properties setting  should be same.

openinx: Not a blocking issue for this PR, but a nice to have issue.

server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableOperations.java
```diff
@@ -18,6 +21,18 @@ public SdkTableOperations(ApiClient apiClient) {
...
+    if (createTableRequest.getTableType() == TableType.MANAGED) {
+      // Caller actually should not set storage location for MANAGED tables. This function will set
+      // it to the staging location of staging table.
+      assert createTableRequest.getStorageLocation() == null;
+      CreateStagingTable createStagingTableRequest =
+          new CreateStagingTable()
+              .catalogName(createTableRequest.getCatalogName())
+              .schemaName(createTableRequest.getSchemaName())
+              .name(createTableRequest.getName());
+      StagingTableInfo stagingTableInfo = tablesApi.createStagingTable(createStagingTableRequest);
```
openinx, line 33: So for every normal table creation process, it requires us to create the staging table firstly ?  Could you please explain more about ? Or is there any reference that I can read ?

examples/cli/src/main/java/io/unitycatalog/cli/utils/FileOperations.java
yili-db: For both `FileOperation.java` and `Constants.java`, @tdas mentioned that we want the CLI to be a separate implementation that is only depending on the api models. It's not part of the server so it shouldn't depend on the server.
@tdas thinks it's OK to copy some code here in order to keep the CLI a separate package.
I am not sure how much value it brings to introduce and maintain a common package of these two files.

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
yili-db: This commented code here is to show the reader how it is supposed to work. Eventually it will be uncommented when it's actually working.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
yili-db: This is similar to https://github.com/unitycatalog/unitycatalog/pull/1136#discussion_r2501278410
@tdas how do you think about this?

server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableOperations.java
yili-db: For managed tables, yes. This has description of this process:
https://github.com/delta-io/delta/blob/master/protocol_rfcs/catalog-managed.md#creating-and-dropping-catalog-managed-tables

examples/cli/src/main/java/io/unitycatalog/cli/utils/FileOperations.java
tdas: Yes. CLI is an model "UC + Delta connector" implementation. Others will look at this implementaiton and implement their own connector. We shouldnt be showing that they need to depend on server and use server's internal classes for this.

examples/cli/src/main/java/io/unitycatalog/cli/utils/Constants.java
tdas: maybe. that also means we will have to publish that common module, and users will start using that common module as public API. which means we really need to think hard about public API stability, java docs, etc. 

I am hesitating to take on that long term burden of API guarantees for 2-3 classes being duplicated.

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
```diff
@@ -98,44 +98,70 @@ private static String createTable(TablesApi apiClient, JSONObject json)
...
-    if (path.startsWith("file:")) {
-      return Paths.get(URI.create(path));
+      if (!json.has(CliParams.STORAGE_LOCATION.getServerParam())) {
+        throw new CliException("Storage location is required for external tables");
+      }
+      String storageLocation =
+          FileOperations.toStandardizedURIString(
+              json.getString(CliParams.STORAGE_LOCATION.getServerParam()));
+      createTable.setStorageLocation(storageLocation);
+      // Currently generateTemporaryPathCredentials doesn't quite work yet. So this step has to
```
tdas: why does it not work. can you write down the reason?

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
```diff
@@ -98,44 +98,70 @@ private static String createTable(TablesApi apiClient, JSONObject json)
...
-          throw new CliException("Failed to create delta table at " + path, e);
-        }
+    // try and initialize the directory and initiate delta log at the location
+    try {
+      DeltaKernelUtils.createDeltaTable(
+          createTable.getStorageLocation(), columnInfoList, temporaryCredentials);
+    } catch (Exception e) {
+      if (e instanceof TableAlreadyExistsException
+          || e.getCause() instanceof TableAlreadyExistsException) {
+        // TODO confirm the schema of the existing table matches the schema of the new table
```
tdas, line 162: why do we need to do this? are we building `create` semantics or `create or replace` semantics?

yili-db: This behavior is unchanged by this PR.
A common pattern is that many tests would test the failure cases of table creation. But they often leave the delta table on disk after failing to register a table in UC. Then they retry the table creation on the same location. This would allow those tests to continue.

examples/cli/src/test/java/io/unitycatalog/cli/table/CliTableCreationTest.java
```diff
@@ -97,29 +98,62 @@ public void testCreateTableLocalDirectoryExistsWithDeltaLog() throws ApiExceptio
                     Paths.get(tablePath).toUri().toString(), columns, null))
         .doesNotThrowAnyException();
     assertThat(Paths.get(tablePath + "/_delta_log")).isDirectory();
-    createTableAndAssertReadTableSucceeds(tablePath, columns);
+    createTableAndAssertReadTableSucceeds(TableType.EXTERNAL, Optional.of(tablePath), columns);
   }
 
-  private void createTableAndAssertReadTableSucceeds(String tablePath, List<ColumnInfo> columns)
+  @Test
+  public void testCreateManagedTable() throws IOException, ApiException {
```
tdas, line 106: should you also test that for managed, non-delta format fails? 

think about code coverage a bit

tdas: Are there Server tests that verify non-delta format does not work?

tdas (approved):
LGTM. just a few more comments.

Can you confirm that the cli actually manually works from the command line? like you can create managed table when server is started with experimental managed table enabled? test the behavior and the output of the CLI is as expected for the scenarios that you have unit tests for (valid and invalid cases) and see if the output is  user friendly

yili-db:
> LGTM. just a few more comments.
> 
> Can you confirm that the cli actually manually works from the command line? like you can create managed table when server is started with experimental managed table enabled? test the behavior and the output of the CLI is as expected for the scenarios that you have unit tests for (valid and invalid cases) and see if the output is user friendly

@tdas 
Yes it's expected:
```
~/unitycatalog$ bin/uc table create --full_name unity.default.my_managed_table --columns "id INT, name STRING" --format DELTA --table_type MANAGED
┌────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────┐
│        KEY         │                                          VALUE                                           │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│NAME                │my_managed_table                                                                          │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│CATALOG_NAME        │unity                                                                                     │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│SCHEMA_NAME         │default                                                                                   │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│TABLE_TYPE          │MANAGED                                                                                   │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│DATA_SOURCE_FORMAT  │DELTA                                                                                     │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│COLUMNS             │{"name":"id","type_text":"int","type_json":"{\"name\":\"id\",\"type\":\"integer\",\"nullab│
│                    │le\":true,\"metadata\":{}}","type_name":"INT","type_precision":0,"type_scale":0,"type_inte│
│                    │rval_type":null,"position":0,"comment":null,"nullable":true,"partition_index":null}       │
│                    │{"name":"name","type_text":"string","type_json":"{\"name\":\"name\",\"type\":\"string\",\"│
│                    │nullable\":true,\"metadata\":{}}","type_name":"STRING","type_precision":0,"type_scale":0,"│
│                    │type_interval_type":null,"position":1,"comment":null,"nullable":true,"partition_index":nul│
│                    │l}                                                                                        │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│STORAGE_LOCATION    │file:///tmp/ucroot/tables/7a030426-315a-4f3b-9eec-b51fbc73d042                            │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│COMMENT             │null                                                                                      │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│PROPERTIES          │{}                                                                                        │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│OWNER               │null                                                                                      │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│CREATED_AT          │1762543881947                                                                             │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│CREATED_BY          │null                                                                                      │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│UPDATED_AT          │1762543881947                                                                             │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│UPDATED_BY          │null                                                                                      │
├────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│TABLE_ID            │7a030426-315a-4f3b-9eec-b51fbc73d042                                                      │
└────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────┘

~/unitycatalog$  la /tmp/ucroot/tables/7a030426-315a-4f3b-9eec-b51fbc73d042/
total 0
drwxr-xr-x 4 yi.li wheel 128 Nov  7 11:31 _delta_log
drwxr-xr-x 3 yi.li wheel  96 Nov  7 11:31 .
drwxr-xr-x 3 yi.li wheel  96 Nov  7 11:31 ..
~/unitycatalog$  la /tmp/ucroot/tables/7a030426-315a-4f3b-9eec-b51fbc73d042/_delta_log/
total 8
drwxr-xr-x 4 yi.li wheel 128 Nov  7 11:31 .
drwxr-xr-x 3 yi.li wheel  96 Nov  7 11:31 ..
-rw-r--r-- 1 yi.li wheel  16 Nov  7 11:31 .00000000000000000000.json.crc
-rw-r--r-- 1 yi.li wheel 677 Nov  7 11:31 00000000000000000000.json
```

And if I try to create PARQUET table:
```
~/unitycatalog$ bin/uc table create --full_name unity.default.my_managed_table --columns "id INT, name STRING" --format PARQUET --table_type MANAGED
Error occurred while executing the command. Only delta tables are supported for managed tables: PARQUET
```

And if it's not enabled in server properties:
```
~/unitycatalog$ bin/uc table create --full_name unity.default.my_managed_table --columns "id INT, name STRING" --format DELTA --table_type MANAGED
Exception in thread "main" java.lang.RuntimeException: io.unitycatalog.client.ApiException: createStagingTable call failed with: 400 - {"error_code":"INVALID_ARGUMENT","details":[{"reason":"INVALID_ARGUMENT","metadata":{},"@type":"google.rpc.ErrorInfo"}],"stack_trace":null,"message":"MANAGED table is an experimental feature and is currently disabled. To enable it, set 'server.managed-table.enabled=true' in server.properties"}
	at io.unitycatalog.cli.UnityCatalogCli.main(UnityCatalogCli.java:176)
Caused by: io.unitycatalog.client.ApiException: createStagingTable call failed with: 400 - {"error_code":"INVALID_ARGUMENT","details":[{"reason":"INVALID_ARGUMENT","metadata":{},"@type":"google.rpc.ErrorInfo"}],"stack_trace":null,"message":"MANAGED table is an experimental feature and is currently disabled. To enable it, set 'server.managed-table.enabled=true' in server.properties"}
	at io.unitycatalog.client.api.TablesApi.getApiException(TablesApi.java:78)
	at io.unitycatalog.client.api.TablesApi.createStagingTableWithHttpInfo(TablesApi.java:118)
	at io.unitycatalog.client.api.TablesApi.createStagingTable(TablesApi.java:96)
	at io.unitycatalog.cli.TableCli.createTable(TableCli.java:133)
	at io.unitycatalog.cli.TableCli.handle(TableCli.java:49)
	at io.unitycatalog.cli.UnityCatalogCli.main(UnityCatalogCli.java:141)
```

examples/cli/src/test/java/io/unitycatalog/cli/table/CliTableCreationTest.java
yili-db: Added a test here.
Also added a server test case in SdkTableCRUDTest.java: 
https://github.com/unitycatalog/unitycatalog/pull/1136/files#diff-9cd3f897dc685e26940078c433780dd3803392311d324afff99dcdd1441c81e3R150

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
```diff
@@ -36,12 +38,13 @@
...
 
-  public static String createDeltaTable(
-      String tablePath, List<ColumnInfo> columns, AwsCredentials awsTempCredentials) {
+  public static void createDeltaTable(
+      String tablePath, List<ColumnInfo> columns, TemporaryCredentials temporaryCredentials) {
     try {
       URI tablePathUri = URI.create(tablePath);
-      Engine engine = getEngine(tablePathUri, awsTempCredentials);
+      Engine engine = getEngine(tablePathUri, temporaryCredentials);
       Table table = Table.forPath(engine, substituteSchemeForS3(tablePath));
```
openinx, line 48: So basically, this is creating a path-based delta table right ?

openinx: Since we currenlty don't support the managed table creation.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
```diff
@@ -54,22 +57,24 @@ public static String createDeltaTable(
       // create an empty table
       TransactionCommitResult commitResult = txn.commit(engine, CloseableIterable.emptyIterable());
       if (commitResult.getVersion() >= 0) {
-        System.out.println("Table created successfully at: " + tablePath);
+        LOGGER.info("Table created successfully at: {}", tablePathUri);
```
openinx, line 60: Thanks for the change, I think we need to disable the `System.out.println()`, and enforce it in the checkstyle.

openinx: Added this https://github.com/unitycatalog/unitycatalog/issues/1142

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
```diff
@@ -54,22 +57,24 @@ public static String createDeltaTable(
...
       } else {
         throw new RuntimeException("Table creation failed");
       }
     } catch (Exception e) {
-      throw new IllegalArgumentException("Failed to create delta table", e);
+      String errorMsg =
+          String.format("Failed to create Delta table at '%s': %s", tablePath, e.getMessage());
+      LOGGER.error(errorMsg, e);
```
openinx: As we will finally throw the exception as root cause to the upper layer,  do we still need to log this redudant stacktrace via logger ?  I will assume it's not neccessary.

examples/cli/src/main/java/io/unitycatalog/cli/utils/FileOperations.java
openinx: My thought is:  it's better for us to abstract a common module, which all the `client`, `server` and `connector` can reuse the same `common` module.   the common module should meet the following requirements: 
1.  Abstracted enough,  if not applicable, then may better to keep in the individual module. 
2. Clean dependencies,  no conflicts with other.  Since all the `client`, `server`, `connector` will depend on this `common`, we don't expect the `common` will mess the runtime dependencies. 

The utilties that I can think that can be made to `common` are: 
1.  The `ApiClient` construction.  I had a comment before: https://github.com/unitycatalog/unitycatalog/pull/1138#discussion_r2511627680
2.  General cred props settings for all the 3 cloud vendor.  I will assume that lots of places will use it.  ---> for example if we want to support streaming write without catalog-based path,  then we need to inject those cred props. 
3. Other utilies, such as env var parser, properties parsers,  config keys parsers etc.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
openinx: Thanks for the links, added my comment there: https://github.com/unitycatalog/unitycatalog/pull/1136#discussion_r2501241466

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
```diff
@@ -81,30 +86,33 @@ public static FileSystem getFileSystem(URI tablePathURI, Configuration conf) {
...
       conf.set("fs.s3a.secret.key", awsTempCredentials.getSecretAccessKey());
       conf.set("fs.s3a.session.token", awsTempCredentials.getSessionToken());
       conf.set("fs.s3a.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem");
       conf.set("fs.s3a.path.style.access", "true");
-    } else if (tablePathUri.getScheme().equals("file")) {
+    } else if (scheme.equals(Constants.URI_SCHEME_FILE)) {
       conf.set("fs.file.impl", "org.apache.hadoop.fs.LocalFileSystem");
     } else {
-      throw new IllegalArgumentException("Unsupported URI scheme: " + tablePathUri.getScheme());
+      throw new IllegalArgumentException("Unsupported URI scheme: " + scheme);
```
openinx, line 106: Will we also need to consider the `gcp` and `azure abfs` ?

openinx: If so please file separate task to address them, thanks.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
yili-db: I don't know the managed vs path-based delta table in the Delta Kernel's context.
This code seems to do the work of constructing a Delta table.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
openinx: Also one more general question is:  Is it possible for us to provide an abstraction so that we don't need to use the `if schema == s3 else if schema == gcp else ` everywhere, for supporting all the 3 cloud vendors.  It make us some kind of hard to maintane and iterate new features.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
openinx: Sounds good to me, thanks for the comment.

examples/cli/src/main/java/io/unitycatalog/cli/utils/Constants.java
openinx: There are actually more and more classes need to be abstracted in the `common` module.  As I commented here: https://github.com/unitycatalog/unitycatalog/pull/1136#discussion_r2511741692

examples/cli/src/main/java/io/unitycatalog/cli/utils/FileOperations.java
yili-db: I think that we can further discuss this but it shouldn't be a blocker for this PR?

examples/cli/src/main/java/io/unitycatalog/cli/utils/Constants.java
openinx: And think from another hand,  if there are only very few API that we need to maintain in `common` module,  then we should be very easy to support the compability. but if there are more and more common classes that we need to maintain in the common module,  while we just copy them in every modules, it will be a tough path for us to maintain. 

Aslo I think we can limit those common classes to be exposed to the outside.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
yili-db: Filed https://github.com/unitycatalog/unitycatalog/issues/1182

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
yili-db: @openinx ideally yes. This also depend on the discussion of a common package to decide where it should live. For now I don't think it should block this PR though.

examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java
tdas: got it. but that makes sense of external tables where it is valid to create an external table in the catalog by pointing it to an existing file system table. however, thats not the case of managed tables. for managed tables, every attempt to create a managed table will be in a new path provided by the catalog. so it should never happen that the 

does that mean some logic here should change? i think we should allow this exception to be swallowed only when its an external table. if table-already-exists exception is ever hit for a managed table ... thats a cause for alarm.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
tdas: Yeah, we need to have updated Delta Kernel that can actually construct a Delta table where catalogManaged table feature is enabled.

examples/cli/src/main/java/io/unitycatalog/cli/utils/FileOperations.java
tdas: +1 on having a separate conversation and not blocking this PR. We should think about 
- is it only for our code deduplication, or is it for other connector developers to use. depending on that we should release the common module... or not. various different solutions are possible. 
- if we do release the common module, we need to think about API stability guarantees, whether we want to take that. Today we are providing API stability guarantees for the clients (java, python) and very loosely for the server (so that others can start the server in embedded in the jvm. do we want to expand that guarantee to this new module?

tdas (approved):
LGTM. thank you!

PR #1137, opened 2025-10-27
Add GCS token renewal support for UC-Spark connector

TimothyW553:
**PR Checklist**

**Description of changes**

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsVendedTokenProviderTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsVendedTokenProviderTest.java
@@ -116,8 +116,7 @@ public class GcsVendedTokenProviderTest extends BaseTokenProviderTest<GcsVendedT
       assertThat(fs.getClass().getName()).isEqualTo(ghfsClassName);
 
       GoogleCredentials credentials =
-          HadoopCredentialsConfiguration.getCredentials(
-              fs.getConf(), "fs.gs");
+          HadoopCredentialsConfiguration.getCredentials(fs.getConf(), "fs.gs");
 
       assertThat(credentials).isInstanceOf(AccessTokenProviderCredentials.class);
 


connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -10,7 +10,7 @@
 import io.unitycatalog.client.model.PathOperation;
 import io.unitycatalog.client.model.TableOperation;
 import io.unitycatalog.client.model.TemporaryCredentials;
-import io.unitycatalog.spark.GcsVendedTokenProvider;
+import io.unitycatalog.spark.auth.GcsVendedTokenProvider;
```
openinx: Seems like we can remove this redudent import ?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -95,6 +95,22 @@ protected S3PropsBuilder self() {
     }
   }
 
+  private static class GcsPropsBuilder extends PropsBuilder<GcsPropsBuilder> {
+
+    GcsPropsBuilder() {
+      // Common properties for GCS.
+      set("fs.gs.create.items.conflict.check.enable", "false");
```
openinx: @tdas , do you have context about what was this prop purpose before ?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
```diff
@@ -0,0 +1,85 @@
...
+  private Configuration conf;
+
+  /**
+   * Constructor for the hadoop's CredentialProviderListFactory to initialize.
+   */
+  public GcsVendedTokenProvider() {
+    super();
+  }
+
+  GcsVendedTokenProvider(Clock clock, long renewalLeadTimeMillis) {
```
openinx: After this change https://github.com/unitycatalog/unitycatalog/pull/1123/files#r2461421859, we don't need the provider with `clock` and `renewalLeadTimeMilis` now.  Please just remove it, use the hadoop configuration to set the manual clock name and `renewalLeadTimeMilis` for testing purposes.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
```diff
@@ -0,0 +1,85 @@
...
+
+public class GcsVendedTokenProvider extends GenericCredentialProvider
+    implements AccessTokenProvider {
+  public static final String ACCESS_TOKEN_KEY = "fs.gs.auth.access.token.credential";
+  public static final String ACCESS_TOKEN_EXPIRATION_KEY = "fs.gs.auth.access.token.expiration";
+
+  private Configuration conf;
+
+  /**
+   * Constructor for the hadoop's CredentialProviderListFactory to initialize.
```
openinx: Do you know how the GCS FileSystem use this constructor to dynamically load ?   I think `CredentialProviderListFactory` is the approach that the `hadoop-aws` adopt,  may not be applicable to `gcs` filesystem. Please confirm this, it's important.  Otherwise,  when design the integration test, it will fail.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  GcsVendedTokenProvider(Clock clock, long renewalLeadTimeMillis) {
+    super(clock, renewalLeadTimeMillis);
+  }
+
+  @Override
+  public GenericCredential initGenericCredential(Configuration conf) {
+    if (conf.get(UCHadoopConf.GCS_INIT_OAUTH_TOKEN) != null
+        && conf.get(UCHadoopConf.GCS_INIT_OAUTH_TOKEN_EXPIRED_TIME) != null) {
```
openinx: We don't need the expiration time nullable check here,  since it's allowed to be nullable if it's a token/credential without expiration.  I mean the static credentials. 

You can see this comment before: https://github.com/unitycatalog/unitycatalog/pull/1123/files#r2460798294

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
```diff
@@ -0,0 +1,85 @@
...
+
+      String oauthToken = conf.get(UCHadoopConf.GCS_INIT_OAUTH_TOKEN);
+      long expiredTimeMillis = conf.getLong(
+          UCHadoopConf.GCS_INIT_OAUTH_TOKEN_EXPIRED_TIME,
+          Long.MAX_VALUE);
+      Preconditions.checkState(expiredTimeMillis > 0, "Expired time %s must be greater than 0, " +
```
openinx, line 32: Nit: Please help to check the checkstyle issue.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
```diff
@@ -0,0 +1,85 @@
...
+    // Wrap the GenericCredential as an AccessToken.
+    GcpOauthToken gcpToken = generic
+        .temporaryCredentials()
+        .getGcpOauthToken();
+    Preconditions.checkNotNull(gcpToken,
+        "GCP OAuth token of generic credentials cannot be null");
+
+    return new AccessToken(
+        gcpToken.getOauthToken(),
+        Instant.ofEpochMilli(generic.temporaryCredentials().getExpirationTime())
```
openinx: Will the `expirationTime` be nullable ?   Please check.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
```diff
@@ -0,0 +1,85 @@
...
+    );
+  }
+
+  @Override
+  public void refresh() throws IOException {
+    // The refresh is handled by accessCredentials() in the parent class.
```
openinx, line 61: I think we should call the `accessCredentials`  to trigger the `refresh` , although we don't really need this credentials here, which should be fine since the newly refreshed credentials will be cached and used for the later access.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
timothyw553: Yeah, we don't need this import

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
timothyw553: Yeah that's right, GCS doesn’t use the S3A-style credential provider factory: when `fs.gs.auth.type=ACCESS_TOKEN_PROVIDER` is set, the Hadoop GCS client reads `fs.gs.auth.access.token.provider`, reflectively instantiates that class, and then calls `setConf`.

So as long as our provider has a no-arg ctor and we set that config key, it will load correctly and the integration test won’t fail.

timothyw553: Those configs are set here [connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java](https://github.com/unitycatalog/unitycatalog/pull/1137/files#diff-ab4ecc349cf85b6952657fa3266eb21da199d6460e0ccf4a95518e76a5ec5f15)

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
timothyw553: Ah got it, then I'll be removing the nullable check for GCS_INIT_OAUTH_TOKEN_EXPIRED_TIME.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
timothyw553: Relevant: https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/bcdf7c8a689eaad95f9222c60af7c41e9969f48b/gcs/CONFIGURATION.md?plain=1#L89

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
timothyw553:
```java
  /**
   * Server time when the credential will expire, in epoch milliseconds. The API client is advised to cache the credential given this expiration time. 
   * @return expirationTime
   */
  @jakarta.annotation.Nullable
  @JsonProperty(JSON_PROPERTY_EXPIRATION_TIME)
  @JsonInclude(value = JsonInclude.Include.USE_DEFAULTS)
  public Long getExpirationTime() {
    return expirationTime;
  }
  ```
  
  Yes seems like expirationTime() is nullable. I will make the correspodning check

File: connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java (hunk 2 above)
Comment (openinx) [reply connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java]: Okay,  got it.  So it's a flag to control if we need to fail the conflicting directories creation. 

Just curious why did we disable this before ? Seems like it's `true` by default.

File: connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java [hunk 9]
```diff
@@ -153,16 +168,58 @@ private static Map<String, String> s3PathTempCredProps(
...
 
-  private static Map<String, String> gsProps(TemporaryCredentials tempCreds) {
+  private static Map<String, String> gsFixedCredProps(TemporaryCredentials tempCreds) {
     GcpOauthToken gcpOauthToken = tempCreds.getGcpOauthToken();
-    return ImmutableMap.<String, String>builder()
-        .put(GcsVendedTokenProvider.ACCESS_TOKEN_KEY, gcpOauthToken.getOauthToken())
-        .put(GcsVendedTokenProvider.ACCESS_TOKEN_EXPIRATION_KEY,
+    return new GcsPropsBuilder()
+        .set(GcsVendedTokenProvider.ACCESS_TOKEN_KEY, gcpOauthToken.getOauthToken())
+        .set(GcsVendedTokenProvider.ACCESS_TOKEN_EXPIRATION_KEY,
```
openinx, line 176: Do we still need the `expiration key`  if it's a `fixedCredProps` ?   I will assume it's not.

openinx: `fs.gs.auth.access.token.expiration` this key , is it an official key from google, or our customized key ? If it's later, then I think we can remove it now.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -20,6 +20,11 @@ private UCHadoopConf() {
   public static final String AZURE_INIT_SAS_TOKEN_EXPIRED_TIME =
       "fs.azure.init.sas.token.expired.time";
 
+  // Keys for the initialized Google Cloud Storage OAuth token.
+  public static final String GCS_INIT_OAUTH_TOKEN = "fs.gs.init.oauth.token";
+  public static final String GCS_INIT_OAUTH_TOKEN_EXPIRED_TIME =
+      "fs.gs.init.oauth.token.expired.time";
```
openinx: Nit: Maybe we need to correct the name, the correct key should be `fs.gs.init.oauth.token.expiration.time`, this also applies to `fs.s3a.init.credential.expiration.time`.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
```diff
@@ -0,0 +1,83 @@
+package io.unitycatalog.spark.auth;
+
+import io.unitycatalog.spark.UCHadoopConf;
+import org.apache.hadoop.conf.Configuration;
+import org.sparkproject.guava.base.Preconditions;
+import io.unitycatalog.client.model.GcpOauthToken;
+import com.google.cloud.hadoop.util.AccessTokenProvider;
+import com.google.cloud.hadoop.util.AccessTokenProvider.AccessToken;
```
openinx: nit:  unused import, please remove it.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.io.IOException;
+import java.time.Instant;
+
+public class GcsVendedTokenProvider extends GenericCredentialProvider
+    implements AccessTokenProvider {
+
+  public static final String ACCESS_TOKEN_KEY = "fs.gs.auth.access.token.credential";
+  public static final String ACCESS_TOKEN_EXPIRATION_KEY = "fs.gs.auth.access.token.expiration";
```
openinx, line 16: What's the difference between this key and the `fs.gs.init.oauth.token.expired.time` ?  Seems it's easily to confuse people.  Maybe we can just remove this key, right ?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
```diff
@@ -0,0 +1,83 @@
...
+
+  private Configuration conf;
+
+  public GcsVendedTokenProvider() {
+  }
+
+  public GcsVendedTokenProvider(Configuration conf) {
```
openinx: Is it required or not ?  if this constructor is required, please add coment why... seems like no usage here.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
timothyw553: Oh yes that's true, I just set it to false for now, but it seems true is the default we should keep.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
timothyw553: Yeah, we can just use the one in UCHadoopConf

connectors/spark/src/main/scala/io/unitycatalog/spark/GcsVendedTokenProvider.java
timothyw553, line 1: Deleted and moved to **`spark/auth/`**`GcsVendedTokenProvider.java`

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -153,16 +166,61 @@ private static Map<String, String> s3PathTempCredProps(
...
+  }
+
+  private static GcsPropsBuilder gcsTempCredPropsBuilder(
+      String uri,
+      String token,
+      TemporaryCredentials tempCreds) {
+    GcpOauthToken gcpToken = tempCreds.getGcpOauthToken();
+    GcsPropsBuilder builder = new GcsPropsBuilder()
+        .set("fs.gs.auth.type", "ACCESS_TOKEN_PROVIDER")
+        .set("fs.gs.auth.access.token.provider", GcsVendedTokenProvider.class.getName())
```
timothyw553, line 188: We only set `fs.gs.auth.access.token.provider` for the renewal flow, where the access token provider refreshes creds, unlike the fixed credential path.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -153,16 +166,61 @@ private static Map<String, String> s3PathTempCredProps(
...
-  private static Map<String, String> gsProps(TemporaryCredentials tempCreds) {
+  private static Map<String, String> gsFixedCredProps(TemporaryCredentials tempCreds) {
     GcpOauthToken gcpOauthToken = tempCreds.getGcpOauthToken();
-    return ImmutableMap.<String, String>builder()
-        .put(GcsVendedTokenProvider.ACCESS_TOKEN_KEY, gcpOauthToken.getOauthToken())
-        .put(GcsVendedTokenProvider.ACCESS_TOKEN_EXPIRATION_KEY,
+    return new GcsPropsBuilder()
+        .set(GcsVendedTokenProvider.ACCESS_TOKEN_KEY, gcpOauthToken.getOauthToken())
+        .set(GcsVendedTokenProvider.ACCESS_TOKEN_EXPIRATION_KEY,
             String.valueOf(tempCreds.getExpirationTime()))
```
openinx: What's the reason that we finally keep this expiration time for the `fixedCredProps`, seems like can be removed now ?

openinx: And it's possible that the `TemporaryCredentials` have nullable expiration time, for example the `StaticCredentialsGenerator` won't have any expiration time https://github.com/unitycatalog/unitycatalog/blob/main/server/src/main/java/io/unitycatalog/server/service/credential/aws/CredentialsGenerator.java#L41 .  It's a credential that generated from the OSS UC server.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
openinx: So for the fixed credential path,  I think we just set the key `fs.gs.auth.access.token.credential` and it will automatically choose the default credential provider, rather than the customized `GcsVendedTokenProvider`, right ?  

i think it should be correct.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -153,16 +166,61 @@ private static Map<String, String> s3PathTempCredProps(
...
+
+  private static GcsPropsBuilder gcsTempCredPropsBuilder(
+      String uri,
+      String token,
+      TemporaryCredentials tempCreds) {
+    GcpOauthToken gcpToken = tempCreds.getGcpOauthToken();
+    GcsPropsBuilder builder = new GcsPropsBuilder()
+        .set("fs.gs.auth.type", "ACCESS_TOKEN_PROVIDER")
+        .set("fs.gs.auth.access.token.provider", GcsVendedTokenProvider.class.getName())
+        .set("fs.gs.auth.access.token.provider.impl", GcsVendedTokenProvider.class.getName())
```
openinx: Why both the two keys need to be set ?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
openinx: Then I think this one can be removed now, as we discussed in the above thread.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
openinx: I think it's okay if we think we don't need to add the `accessCredentials` in `refresh`.  not strong point.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsVendedTokenProviderTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Test
+  public void testLoadProvider() throws Exception {
+    Configuration conf = newTableBasedConf();
+    conf.set("fs.gs.auth.type", "ACCESS_TOKEN_PROVIDER");
+    conf.set("fs.gs.auth.access.token.provider.impl", GcsVendedTokenProvider.class.getName());
+
+    Class<?> providerClazz = conf.getClassByName(conf.get("fs.gs.auth.access.token.provider.impl"));
+    AccessTokenProvider provider =
+        (AccessTokenProvider) ReflectionUtils.newInstance(providerClazz, conf);
```
openinx: Is it possible to use the google filesystem's default  loader ( rather than the ReflectionUtils  ) to test the class loading process ?  we designed this test before, because we want to make sure that the S3AFileSystem/GCSFileSystem can successfully load/reflect the class.  That's why we need the buildin loader from the GCSFileSystem to test it.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
timothyw553: The reason I kept this is because the following test [/spark/GCSCredentialTestFileSystem.java](https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/test/java/io/unitycatalog/spark/GCSCredentialTestFileSystem.java) requires the expiration key.

When expiration is null should we just use `Long.MAX_VALUE`?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
timothyw553: Some sources include `.impl` and some do not. I'm not 100% sure which is correct.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
timothyw553: Yeah that's right.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
timothyw553: So actually because of existing test, it may be better to keep it.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GcsVendedTokenProvider.java
timothyw553: In some examples I found they implement `refresh` as a no-op. I don't think it will affect any downstream stuff if we handle refresh via `accessCredentials` instead of explicitly in `refresh`

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
openinx: Okay, then could you please open a separate issue to remove this key ? I think we can temporarily keep this config key. And I think we also need to change the oss uc server logic, because currently it just use the static expiration time.  https://github.com/unitycatalog/unitycatalog/blob/0127a802d23f8530fc27b33712a886962ef385a3/server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java#L48

openinx: Please link this comment to the newly created issue, thanks.

timothyw553: Done. Issue here: https://github.com/unitycatalog/unitycatalog/issues/1159

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
openinx: Seems like no usage & definition for this `fs.gs.auth.access.token.provider.impl`, https://github.com/search?q=repo%3AGoogleCloudDataproc%2Fhadoop-connectors%20fs.gs.auth.access.token.provider.impl&type=code

Then I think it should be fine to remove it.

openinx: Do you know what's the place that we will use this `impl` key ?

timothyw553: https://github.com/search?q=%22fs.gs.auth.access.token.provider.impl%22&type=code

I found some instances here. But looking deeper it seems that the one without `impl` is the latest version. Therefore, I will remove it.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsVendedTokenProviderTest.java
timothyw553: Yes, updated the test to instantiate `FileSystem` for `gs://...`, which forces `GoogleHadoopFileSystem` to run its own loader; we then assert it hands back our `GcsVendedTokenProvider`. No more direct `ReflectionUtils`.

PR #1145, opened 2025-10-29
RetryHttpClient to enable Retry for UC

TimothyW553:
**PR Checklist**

**Description of changes**

Instead of wrapping each type of Api such as TempCredsApi or TablesApi with a "Retryable" version, we instead create our own RetryableHttpClient and retry at the http level. This way, new Apis just need to pass the necessary configuration to the ApiClientFactory, which creates a RetryingApiClient, which then creates a RetryingHttpClient. The user may pass in a custom interceptor -- a function that defines under which error codes, status codes, and exception they would want the ApiClient to retry.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
@@ -90,10 +90,14 @@ public class UCHadoopConf {
     if (conf == null || apiClientConf == null) {
       return;
     }
-    conf.setInt(REQUEST_RETRY_MAX_ATTEMPTS_KEY, apiClientConf.getRequestMaxAttempts());
-    conf.setLong(REQUEST_RETRY_INITIAL_DELAY_KEY, apiClientConf.getRequestInitialDelayMs());
-    conf.setDouble(REQUEST_RETRY_DELAY_MULTIPLIER_KEY, apiClientConf.getRequestDelayMultiplier());
-    conf.setDouble(REQUEST_RETRY_DELAY_JITTER_FACTOR_KEY, apiClientConf.getRequestDelayJitterFactor());
+    conf.setInt(
+        REQUEST_RETRY_MAX_ATTEMPTS_KEY, apiClientConf.getRequestMaxAttempts());
+    conf.setLong(
+        REQUEST_RETRY_INITIAL_DELAY_KEY, apiClientConf.getRequestInitialDelayMs());
+    conf.setDouble(
+        REQUEST_RETRY_DELAY_MULTIPLIER_KEY, apiClientConf.getRequestDelayMultiplier());
+    conf.setDouble(
+        REQUEST_RETRY_DELAY_JITTER_FACTOR_KEY, apiClientConf.getRequestDelayJitterFactor());
   }
 
   public static ApiClientConf getApiClientConf(Configuration conf) {
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
@@ -126,6 +126,43 @@ public class HttpRetryHandlerTest {
 
   @Test
   @SuppressWarnings("unchecked")
+  public void testMultiplierControlsBackoffScaling() throws IOException, InterruptedException {
+    ApiClientConf conf =
+        new ApiClientConf()
+            .setRequestMaxAttempts(3)
+            .setRequestInitialDelayMs(40L)
+            .setRequestDelayMultiplier(3.0)
+            .setRequestDelayJitterFactor(0.0); // Disable jitter
+
+    HttpClient mockClient = mock(HttpClient.class);
+    HttpRequest mockRequest =
+        HttpRequest.newBuilder().uri(URI.create("http://localhost:8080/api/backoff")).build();
+    HttpResponse.BodyHandler<String> bodyHandler = HttpResponse.BodyHandlers.ofString();
+
+    HttpResponse<String> response503 = createMockResponse(503, "Service Unavailable");
+    HttpResponse<String> response502 = createMockResponse(502, "Bad Gateway");
+    HttpResponse<String> response200 = createMockResponse(200, "Recovered");
+
+    when(mockClient.send(
+            any(HttpRequest.class), ArgumentMatchers.<HttpResponse.BodyHandler<String>>any()))
+        .thenReturn(response503)
+        .thenReturn(response502)
+        .thenReturn(response200);
+
+    Instant start = clock.now();
+    HttpRetryHandler handler = new HttpRetryHandler(conf, clock);
+    HttpResponse<String> result = handler.call(mockClient, mockRequest, bodyHandler);
+
+    verify(mockClient, times(3))
+        .send(any(HttpRequest.class), ArgumentMatchers.<HttpResponse.BodyHandler<String>>any());
+    assertThat(result.statusCode()).isEqualTo(200);
+
+    // Delays: 40ms * 3^(1-1) = 40ms, then 40ms * 3^(2-1) = 120ms.
+    assertThat(clock.now()).isEqualTo(start.plus(Duration.ofMillis(160)));
+  }
+
+  @Test
+  @SuppressWarnings("unchecked")
   public void testRetriesRecoverableException() throws IOException, InterruptedException {
     ApiClientConf conf =
         new ApiClientConf()


connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingHttpClient.java
```diff
@@ -0,0 +1,268 @@
...
+    return delegate.version();
+  }
+
+  @Override
+  public Optional<Executor> executor() {
+    return delegate.executor();
+  }
+
+  @Override
+  public <T> HttpResponse<T> send(HttpRequest request, HttpResponse.BodyHandler<T> handler)
```
timothyw553, line 84: This is where the retry happens.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingHttpClient.java
```diff
@@ -0,0 +1,268 @@
...
+        lastException);
+  }
+
+  @Override
+  public <T> CompletableFuture<HttpResponse<T>> sendAsync(
+      HttpRequest request,
+      HttpResponse.BodyHandler<T> handler) {
+    // Delegate without retry for async calls
+    // Future enhancement: implement async retry logic
+    return delegate.sendAsync(request, handler);
```
openinx, line 93: Will we have `sendAsync` http call ? if so, then we may also need to add the retry logic for it, which will lots of duplicate code as the `send`.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingHttpClient.java
```diff
@@ -0,0 +1,268 @@
...
+
+  @Override
+  public <T> HttpResponse<T> send(HttpRequest request, HttpResponse.BodyHandler<T> handler)
+      throws IOException, InterruptedException {
+    Exception lastException = null;
+    var startTime = clock.now();
+
+    for (int attempt = 1; attempt <= maxAttempts; attempt++) {
+      try {
+        HttpResponse<T> response = delegate.send(request, handler);
```
openinx: This is actually some kind of intercepter that defined by our own.

openinx: Will it be more simpler for us to use the `apache-httpclient` or `OkhttpClient`'s default `Interceptor` API  do the retry ? 

The okhttpclient seems have the best API to do that, but need to introduce the extra okhttpclient. The `apache-httpclient` also have the API to do that, but not sure if it's feasible to use the `HttpRequestRetryHandler` to do this.

connectors/spark/src/main/scala/io/unitycatalog/spark/HttpRetryHandler.java
```diff
@@ -0,0 +1,38 @@
+package io.unitycatalog.spark;
+
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+
+/**
+ * Interface for determining whether an HTTP request should be retried.
+ * 
+ * <p>This interface allows users to implement custom retry logic for both
+ * exceptions and HTTP responses. Implementations should be stateless and thread-safe.
```
timothyw553: Inspiration for this interface: https://hc.apache.org/httpcomponents-client-4.5.x/current/tutorial/html/fundamentals.html

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
```diff
@@ -0,0 +1,123 @@
...
+   * @param handler The retry handler to use
+   * @return This object for method chaining
+   */
+  public RetryingApiClient setRetryHandler(HttpRetryHandler handler) {
+    this.retryHandler = handler;
+    return this;
+  }
+
+  @Override
+  public HttpClient getHttpClient() {
```
timothyw553, line 29: When the ApiClient calls getHttpClient(), we return a custom one that intercepts and retries.

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
```diff
@@ -21,4 +23,39 @@ public static ApiClient createApiClient(URI url, String token) {
...
+    // If a custom retry handler is provided, enable retrying even if retry is disabled in config
+    if (retryEnabled || retryHandler != null) {
+      RetryingApiClient client = retryHandler != null
+          ? new RetryingApiClient(conf, Clock.systemClock(), retryHandler)
+          : new RetryingApiClient(conf);
+
+      client.setHost(url.getHost())
+          .setScheme(url.getScheme())
+          .setPort(url.getPort())
+          .setBasePath(url.getPath());
```
urosstan-db: To preserve same behaviour as current `getApiClient`, basePath should have something like `api/unity-catalog/2.1` as suffix on `url.getPath`.

timothyw553: Thank you! Yes, I've removed it to comply with the behaviour of current `getApiClient`

timothyw553:
UC server failed to start -- closing and reopening...

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
openinx: Nit: Maybe we can just make this method to be private, since the upper layer shouldn't use this public creator without retry policy any more.

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
```diff
@@ -21,4 +23,38 @@ public static ApiClient createApiClient(URI url, String token) {
...
+
+  public static ApiClient createApiClient(Configuration conf, URI url, String token) {
+    return createApiClient(conf, url, token, null);
+  }
+
+  public static ApiClient createApiClient(
+      Configuration conf,
+      URI url,
+      String token,
+      HttpRetryHandler retryHandler) {
```
openinx: What's the case that we enabled the `retry` but don't have the `retryHandler` ?   I will think that people only need be aware either `retryEnabled` or `retryHandle`, but they don't need to perceive both, right ? 

Then for the above 
`
public static ApiClient createApiClient(Configuration conf, URI url, String token) {
    return createApiClient(conf, url, token, null);
  }
`

we only need to keep one public method, and remove the other one.  Does this make sense ?

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
```diff
@@ -21,4 +23,38 @@ public static ApiClient createApiClient(URI url, String token) {
...
+          : new RetryingApiClient(conf);
+
+      client.setHost(url.getHost())
+          .setScheme(url.getScheme())
+          .setPort(url.getPort());
+
+      if (token != null && !token.isEmpty()) {
+        client.setRequestInterceptor(
+            request -> request.header("Authorization", "Bearer " + token));
+      }
```
openinx: This seems like is duplicating the native `ApiClient#createApiClient`, right ?  Is it possible to dedup ?

openinx: I guess for the `OAuth` implementation, we will need to set some attribution for both of the diverse methods if we plan to keep the current divergence.   --- which is prone to be buggy maybe.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
```diff
@@ -0,0 +1,70 @@
...
+import io.unitycatalog.spark.utils.Clock;
+import org.apache.hadoop.conf.Configuration;
+
+import java.net.URI;
+import java.net.http.HttpClient;
+
+public class RetryingApiClient extends ApiClient {
+
+  private final Configuration conf;
+  private final Clock clock;
```
openinx: Do we still need this?  maybe can just remove it, seems like no useage.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
```diff
@@ -0,0 +1,70 @@
...
+  private HttpRetryHandler retryHandler;
+
+  public RetryingApiClient(Configuration conf) {
+    this(conf, Clock.systemClock(), null);
+  }
+
+  public RetryingApiClient(Configuration conf, Clock clock) {
+    this(conf, clock, null);
+  }
```
openinx: No usage ? can be removed ?

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
```diff
@@ -0,0 +1,70 @@
...
+
+  public RetryingApiClient(Configuration conf, Clock clock, HttpRetryHandler retryHandler) {
+    super();
```
openinx: it will always call the `super()` for every subclass, by default.  So it should be safe to remove it here.  Do I understand correctly ?

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  /**
+   * Sets a custom retry handler for this API client.
+   *
+   * @param handler The retry handler to use
+   * @return This object for method chaining
+   */
+  public RetryingApiClient setRetryHandler(HttpRetryHandler handler) {
```
openinx: No usage method ? Maybe we can just remove it, right ?

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
```diff
@@ -0,0 +1,70 @@
...
+    RetryingApiClient client = new RetryingApiClient(conf);
+    client.setHost(uri.getHost())
+        .setScheme(uri.getScheme())
+        .setPort(uri.getPort())
+        .setBasePath(uri.getPath());
+
+    if (token != null && !token.isEmpty()) {
+      client.setRequestInterceptor(
+          request -> request.header("Authorization", "Bearer " + token));
+    }
```
openinx: We already have 3 places to set the ApiClient attributions,  please help to dedup it, and try to simply it into one.  Thanks.

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,173 @@
...
+import java.time.Instant;
+import java.util.Set;
+import java.util.logging.Logger;
+
+/**
+ * Default implementation of {@link HttpRetryHandler} that retries on common
+ * recoverable HTTP errors and network exceptions.
+ */
+public class DefaultHttpRetryHandler implements HttpRetryHandler {
+  private static final Logger LOGGER = Logger.getLogger(DefaultHttpRetryHandler.class.getName());
```
openinx: Nit:  should be  safe to remove this, right ?

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,173 @@
...
+        UCHadoopConf.RETRY_INITIAL_DELAY_KEY,
+        UCHadoopConf.RETRY_INITIAL_DELAY_DEFAULT);
+    this.multiplier = conf.getDouble(
+        UCHadoopConf.RETRY_MULTIPLIER_KEY,
+        UCHadoopConf.RETRY_MULTIPLIER_DEFAULT);
+    this.jitterFactor = conf.getDouble(
+        UCHadoopConf.RETRY_JITTER_FACTOR_KEY,
+        UCHadoopConf.RETRY_JITTER_FACTOR_DEFAULT);
+
+    Preconditions.checkArgument(this.maxAttempts >= 1,
```
openinx: nit: usually we don't use `this.` unless we are setting value to the member of the class in the constructor or the `Setter` ( for distingushing it's an argument name or field, when reading the code ).  It seems like an implicit code style, but seems not enforce..

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,173 @@
...
+        String.format("Retry max attempts must be at least 1, got: %d (%s)",
```
openinx: Please just use the naturally supported argument format in the `Preconditions`, thanks. 

```
    Preconditions.checkArgument(maxAttempts > 0,
        "Retry max attempts must be greater than 0, but got '%s'=%s",
        UCHadoopConf.RETRY_MAX_ATTEMPTS_KEY, maxAttempts);
```

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,173 @@
...
+            this.initialDelayMs, UCHadoopConf.RETRY_INITIAL_DELAY_KEY));
+    Preconditions.checkArgument(this.multiplier > 0,
+        String.format("Retry multiplier must be positive, got: %.2f (%s)",
+            this.multiplier, UCHadoopConf.RETRY_MULTIPLIER_KEY));
+    Preconditions.checkArgument(this.jitterFactor >= 0 && this.jitterFactor < 1,
+        String.format("Retry jitter factor must be between 0 and 1 (exclusive), got: %.2f (%s)",
+            this.jitterFactor, UCHadoopConf.RETRY_JITTER_FACTOR_KEY));
+  }
+
+  public DefaultHttpRetryHandler(Configuration conf) {
```
openinx: no usage for this constructor ? can be removed maybe...

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  private boolean isRecoverableResponse(HttpResponse<?> response) {
+    int statusCode = response.statusCode();
+
+    if (RECOVERABLE_STATUS_CODES.contains(statusCode)) {
+      return true;
+    }
+
+    if (statusCode >= 500 && statusCode < 600) {
```
openinx: Why do we need to limit the `statusCode` in range `[500, 600)` ?  because the http code is always `>=500` if there is an internal error  ?  

maybe we can add a simple comment for better understanding purpose , right ?

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,173 @@
...
+  private boolean isRecoverableException(Throwable e) {
+    // Check the entire exception cause chain for recoverable exceptions
+    return Throwables.getCausalChain(e).stream()
+        .anyMatch(cause -> RECOVERABLE_EXCEPTIONS.stream()
+            .anyMatch(exceptionClass -> exceptionClass.isInstance(cause)));
+  }
+
+  private String extractErrorCode(HttpResponse<?> response) {
+    Object body = response.body();
+    if (!(body instanceof String) || ((String) body).isEmpty()) {
```
openinx: Nit: a bit strange for me here, since we need a separate `type cast`.   I just read the javadoc of the `java.net.http.HttpResponse#body()`, it says: 

```
    /**
     * Returns the body. Depending on the type of {@code T}, the returned body
     * may represent the body after it was read (such as {@code byte[]}, or
     * {@code String}, or {@code Path}) or it may represent an object with
     * which the body is read, such as an {@link java.io.InputStream}.
     *
     * <p> If this {@code HttpResponse} was returned from an invocation of
     * {@link #previousResponse()} then this method returns {@code null}
     *
     * @return the body
     */
    public T body();
```

seems like this generic type can be `byte[]`, `String`, `Path`, or `InputStream`, or anything else.  How can we ensure that all the api's error code will finally turn to be an expected `String` ?

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,173 @@
...
+    Instant startTime = clock.now();
+
+    for (int attempt = 1; attempt <= maxAttempts; attempt++) {
+      try {
+        HttpResponse<T> response = delegate.send(request, responseBodyHandler);
+        if (!isRecoverableResponse(response)) {
+          return response;
+        }
+
+        lastException = new IOException("HTTP " + response.statusCode());
```
openinx: For every retry, we need to new a separate `IOException` ?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/GenericCredentialProvider.java
```diff
@@ -126,10 +126,9 @@ private GenericCredential createGenericCredentials() throws ApiException {
     TemporaryCredentialsApi tempCredApi = temporaryCredentialsApi();
 
     // Generate the temporary credential via requesting UnityCatalog.
+    // Retry is automatically handled by RetryingHttpClient when fs.unitycatalog.retry.enabled=true
```
openinx: Nit: will it make more sense to move this comment when we actually build the api client here ? 
https://github.com/unitycatalog/unitycatalog/pull/1145/files#diff-68d2e51ab5a4e6186ba02fecda9958db316feac1f8397a17c1da4cba334be188R99

The block is acutally calling the API without any context about the `ApiClient`,  seems strange here...w

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
openinx: I say this, because there is a PR to change this api client creation. And people is easily to forget to change other places if we don't unify them in one place. 

https://github.com/unitycatalog/unitycatalog/pull/1130/files

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
timothyw553: That's a good idea, then we can enforce retry policy.

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
timothyw553: Yep, my bad. It was left in there when we were going to remove default handler outside this class.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
timothyw553: Yeah, that's true.

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
timothyw553: We only create that new `IOException("HTTP " + response.statusCode())` when a response comes back with a recoverable status, so we simply keep the latest failure around in lastException.

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
timothyw553: Yeah, 5xx for server error codes, else 4xx for client side error. I'll add a comment.

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
timothyw553: This is for the case where people have `retry` enabled but then simply want to use the `DefaultHttpRetryHandler`. If they forgot to have `retry` enabled but pass in a custom retry handler, should we still use the custom retry handler or just not use the retry mechanism?

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
timothyw553: Actually, it seems that UCSingleCatalog uses this constructor.

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
timothyw553: I think by default they return an InputStream. Should we handle that?

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingHttpClient.java
timothyw553: Async calls are a bit complicated. I think they will require a slightly different approach from what we have right now so we can create a separate PR for that.

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,293 @@
...
+    static <T> ServerErrorEvaluation<T> retry(IOException cause) {
+      return new ServerErrorEvaluation<>(true, cause, null);
+    }
+
+    static <T> ServerErrorEvaluation<T> noRetry(HttpResponse<T> response) {
+      return new ServerErrorEvaluation<>(false, null, response);
+    }
+  }
+
+  private static final class ReplayableInputStreamHttpResponse
```
timothyw553: The stream response from `HttpResponse<InputStream>` can only be consumed once, but our retry logic has to read it in order to check the `error_code`.

Once we do that the original stream is exhausted, so downstream callers would see an empty body unless we replace it. `ReplayableInputStreamHttpResponse` buffers the bytes and hands out a fresh `ByteArrayInputStream` on each `body()` call (to preserve the original response contract).

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  private boolean isRecoverableException(Throwable e) {
+    // Check the entire exception cause chain for recoverable exceptions
+    return Throwables.getCausalChain(e).stream()
+        .anyMatch(cause -> RECOVERABLE_EXCEPTIONS.stream()
+            .anyMatch(exceptionClass -> exceptionClass.isInstance(cause)));
+  }
+
+  private <T> ServerErrorEvaluation<T> evaluateServerError(HttpResponse<T> response) {
```
timothyw553: `evaluateServerError` runs only for 5xx (server error) responses. It grabs the body, and if it isn't an `InputStream` then just stops. It uses the stream buffer so we can parse the JSON `error_code` without losing the payload (see the comment below). Retry if code matches a recoverable value, otherwise replay the http response so downstream callers still get the same body.

implemented `ServerErrorEvaluation` just to keep the code cleaner. It just stores decision to retry, not retry, and response.

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,293 @@
...
+
+      boolean shouldRetry = false;
+      IOException retryReason = null;
+      int statusCode = response.statusCode();
+
+      if (RECOVERABLE_STATUS_CODES.contains(statusCode)) {
+        shouldRetry = true;
+        retryReason = new IOException("HTTP " + statusCode);
+      } else if (statusCode >= 500 && statusCode < 600) {
+        ServerErrorEvaluation<T> evaluation = evaluateServerError(response);
```
timothyw553: See comment below -- this will evaluate whether we should retry or not and also we preserve the inputstream so downstream uses the same body

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Override
+  public HttpClient getHttpClient() {
+    HttpClient baseClient = super.getHttpClient();
+    return new RetryingHttpClient(baseClient, retryHandler);
+  }
+
+  public static RetryingApiClient create(Configuration conf, URI uri, String token) {
+    return ApiClientFactory.configureClient(new RetryingApiClient(conf), uri, token);
+  }
```
openinx: Could you please remove this,  seems like no usage for this, right ?

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,142 @@
...
+      if (!(RECOVERABLE_STATUS_CODES.contains(statusCode) ||
+          (statusCode >= 500 && statusCode < 600))) {
+        return response;
+      }
+
+      lastException = new IOException("HTTP " + statusCode);
+      if (attempt == maxAttempts) {
+        break;
+      }
+      sleepWithBackoff(attempt);
```
timothyw553: We simplify the retry policy such that if the client is sending too many requests (429) or there is an internal server error (5xx) then we retry. 

Before sleep with backoff, make sure there are no more attempts left..

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,142 @@
...
+    if (lastException instanceof IOException) {
+      throw (IOException) lastException;
+    } else if (lastException instanceof InterruptedException) {
+      throw (InterruptedException) lastException;
+    }
+    throw new RuntimeException(
+        "Failed HTTP request after " + maxAttempts + " attempts" +
+            " (elapsed time: " + elapsedMs + "ms)",
+        lastException
+    );
```
timothyw553: after exhausting all attempts we log total elapsed time and throw the error up.

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
```diff
@@ -0,0 +1,142 @@
...
+        if (!isRecoverableException(e) || attempt == maxAttempts) {
+          break;
+        }
+        sleepWithBackoff(attempt);
+        continue;
+      }
+
+      int statusCode = response.statusCode();
+      if (!(RECOVERABLE_STATUS_CODES.contains(statusCode) ||
+          (statusCode >= 500 && statusCode < 600))) {
```
openinx: I think if the statusCode is in range `[500, 600)`, then we should retry rather than forwarding the response directly ?

openinx: Oh, sorry,  I misunderstood the code.  You are doing the retry.

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
```diff
@@ -1,24 +1,40 @@
...
 
 import java.net.URI;
 
 public class ApiClientFactory {
   private ApiClientFactory() {}
 
-  public static ApiClient createApiClient(URI url, String token) {
-    ApiClient apiClient = new ApiClient()
-        .setHost(url.getHost())
+  static <T extends ApiClient> T configureClient(T client, URI url, String token) {
```
tdas: why is this public function?

tdas: and this seems to be a tiny function used once.. why is this even a separate function and not embedded in the caller function?

connectors/spark/src/main/scala/io/unitycatalog/spark/DefaultHttpRetryHandler.java
timothyw553: Note: We end up removing checking the body entirely since if it's a 5xx error we know it's an internal server error and can simply retry. Observing the body can have negative affects for downstream code so just checking the status code is enough.

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
timothyw553: Ah, yes that is true. Originally we had 3 instances of setting the ApiClient attributes: https://github.com/unitycatalog/unitycatalog/pull/1145#discussion_r2487917759

Cleaned up and removed.

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
```diff
@@ -1,24 +1,35 @@
...
 import java.net.URI;
 
 public class ApiClientFactory {
   private ApiClientFactory() {}
 
-  public static ApiClient createApiClient(URI url, String token) {
-    ApiClient apiClient = new ApiClient()
-        .setHost(url.getHost())
+  public static ApiClient createApiClient(Configuration conf, URI url, String token) {
+    return createApiClient(conf, url, token, null);
```
tdas: minor: its not easy to read and understand which param is null. so let's use the `/* retryHandler = */ null` format.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
```diff
@@ -0,0 +1,32 @@
+package io.unitycatalog.spark;
+
+import io.unitycatalog.client.ApiClient;
+import io.unitycatalog.spark.utils.Clock;
+import org.apache.hadoop.conf.Configuration;
+
+import java.net.http.HttpClient;
+
+public class RetryingApiClient extends ApiClient {
```
tdas, line 18: why do we need a separate interface from ApiClient? why cant ApiClient itself support retryable?

tdas: the only difference i see that there is the private field retryHandler. why not add that to the ApiClient interface and always wrap based httpClient with RetryingHttpClient

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -63,4 +63,17 @@ private UCHadoopConf() {
...
+  public static final String RETRY_MAX_ATTEMPTS_KEY = "fs.unitycatalog.retry.maxAttempts";
+  public static final int RETRY_MAX_ATTEMPTS_DEFAULT = 3;
+
+  public static final String RETRY_INITIAL_DELAY_KEY = "fs.unitycatalog.retry.initialDelayMs";
+  public static final long RETRY_INITIAL_DELAY_DEFAULT = 500L;
+
+  public static final String RETRY_MULTIPLIER_KEY = "fs.unitycatalog.retry.multiplier";
+  public static final double RETRY_MULTIPLIER_DEFAULT = 2.0;
+
+  public static final String RETRY_JITTER_FACTOR_KEY = "fs.unitycatalog.retry.jitterFactor";
```
tdas: please add description of what each of these params mean?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -63,4 +63,17 @@ private UCHadoopConf() {
...
+  public static final double RETRY_JITTER_FACTOR_DEFAULT = 0.5;
```
tdas: this approach of defining Api client param constants in hadoopconf constants is... okay.. but not great. hadoop conf constants are just means of communicating between driver and executor the details for the executor to use. so just a communication mechanism, completely unrelated to ApiClients. so the ideal code design is ApiClient to be self-sufficient, that is, ApiClient takes a ApiClientConf, which has all the params that affect the api client. all the params are defined in that ApiClientConf. Anybody can construct a ApiClient with a ApiClientConf (that is, you dont need to create a hadoopConf to just create a ApiClient, [like you are doing here](https://github.com/unitycatalog/unitycatalog/pull/1145/files#diff-772cb160da82c9ee2d2e474f2da6a57cd89db798fdb6aeda04827181194e9d7fR40-R44)).

Then for communicate via hadoopConf, you can have additional utility functions that serializes / deserializes to a string -> string/obj map that can be put in a hadoopConf. For example
```
void UCHadoopConf.setApiClientConf(hadoopConf, apiClientConf);
ApiClientConf  UCHadoopConf.getApiClientConf(hadoopConf);
```

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
timothyw553: Noted, updated.

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
timothyw553: Created a separate interface because `ApiClient` is generated by OpenAPI generator. If we add fields to the generated file, the next regen wipes them (or we have to maintain custom templates).

Current interface leaves the generator alone

tdas: aaaah.. that makes sense. but this is important information that needs to be noted. this why we need to class docs on all classes.. most classes has such important context that need to be documented. so please add class docs and mention this point!

connectors/spark/src/main/scala/io/unitycatalog/spark/HttpRetryHandler.java
tdas: I understand why RetryingApiClient is needed. but why do we need this interface for 1 implementation? why not directly have the RetryHandler class be what you have defined as DefaultRetryHandler?

timothyw553: The original intent was to allow future API clients to use their own custom retry handler, but in most cases they would probably use the DefaultRetryHandler so I agree.

We can have just the RetryHandler class as the DefaultRetryHandler - if in the future anyone wants to use a custom one, it would be easy to refactor.

tdas: exactly. this is premature optimization. unless there is a concrete need to create an interface, its premature to create one. Especially when the functionality is so narrow that we can easily add interface when we have that concrete need for multiple implementations. until then.. unnecessary interfaces are just code bloat.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -63,4 +63,17 @@ private UCHadoopConf() {
   public static final String UC_CREDENTIAL_CACHE_ENABLED_KEY =
       "fs.unitycatalog.credential.cache.enabled";
   public static final boolean UC_CREDENTIAL_CACHE_ENABLED_DEFAULT_VALUE = true;
+
+  // Keys for HTTP client retry configuration
+  public static final String RETRY_MAX_ATTEMPTS_KEY = "fs.unitycatalog.retry.maxAttempts";
```
tdas: `retry` does not clarify retrying of what? this is about retrying requests.. right? so this suffix should be `fs.unitycatalog.request.

timothyw553: Yeah, retrying requests. I kept the `retry` part because of the backoff params like jitter factor and multiplier exist in the context of a retry not just request: `fs.unitycatalog.request.retry.*`

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
timothyw553: Makes sense, keep the retry defaults, validation, and other policies in `ApiClientConf`, and reserve `UCHadoopConf` for the transport keys only. We’ll use `setApiClientConf` to serialize those values into a Hadoop Configuration on the driver, and `getApiClientConf` to rebuild the ApiClientConf from that configuration on executors. Thank you!

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientConf.java
```diff
@@ -0,0 +1,75 @@
...
+ * <p>The defaults here are chosen to work out of the box. Callers can override them either
+ * programmatically or via {@link UCHadoopConf#setApiClientConf} when communicating through
+ * Hadoop configuration.</p>
+ */
+public class ApiClientConf {
+
+  public static final int DEFAULT_REQUEST_MAX_ATTEMPTS = 3;
+  public static final long DEFAULT_REQUEST_INITIAL_DELAY_MS = 500L;
+  public static final double DEFAULT_REQUEST_MULTIPLIER = 2.0;
+  public static final double DEFAULT_REQUEST_JITTER_FACTOR = 0.5;
```
tdas: Not obvious to a reader what each of these configurations means for the request. i understand max attempts, but what is "initial" delay? what is multiplier? multiplies what? 
what is jitter factor? jittering over what baseline? whats the unit?

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+  @Test
+  @SuppressWarnings("unchecked")
+  public void testRetriesRecoverableException() throws IOException, InterruptedException {
+    ApiClientConf conf =
+        new ApiClientConf()
+            .setRequestMaxAttempts(2)
+            .setRequestInitialDelayMs(50L)
+            .setRequestMultiplier(1.0)
+            .setRequestJitterFactor(0.0); // Disable jitter
```
tdas: there is not test testing jitter?

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
tdas: @TimothyW553 where have you documented this?? i dont see any class docs on this.

tdas: and we need the same context information on RetryingHttpClient...right? that also exists because we dont want to modify HttpClient... right?

timothyw553: Updated both retryingapiclient and retryinghttpclient with class docs specifying the need for these classes (*why* we need these classes, how they related to each other, and how retryinghttpclient uses the retryhandler)

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientConf.java
timothyw553: Yes, good point. Added explanations for these parameters, their units, and how they are used in calculating the delay.

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
```diff
@@ -0,0 +1,168 @@
...
+    // Calculated as: baseDelay * (1 ± jitterFactor). In this case, the base delay is 100ms and
+    // the jitter factor is 0.5 so the range is [50ms, 150ms].
+    long elapsedMs = Duration.between(start, clock.now()).toMillis();
+    long baseDelay = conf.getRequestInitialDelayMs();
+    long minDelay = (long) Math.floor(baseDelay * (1 - jitterFactor));
+    long maxDelay = (long) Math.ceil(baseDelay * (1 + jitterFactor));
+    assertThat(elapsedMs)
+        .as("retry delay should stay within jitter-adjusted bounds")
+        .isBetween(minDelay, maxDelay);
+  }
```
TimothyW553: testing regular retry mechanism along with jitter.

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
timothyw553: Right, just added testing with jitter. The test involves going through the normal retrying while asserting that, with a set jitter factor, the time elapsed is within the expected range.

It should be stable as it uses the `ManualClock`, so the only random part is the jitter, and we assert the elapsed time stays within the expected [baseDelay * (1 ± jitter)] window. Because the jitter factor is bounded and the clock is deterministic, the test shouldn’t flake.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -72,4 +74,47 @@ private UCHadoopConf() {
   public static final String UC_CREDENTIAL_CACHE_ENABLED_KEY =
       "fs.unitycatalog.credential.cache.enabled";
   public static final boolean UC_CREDENTIAL_CACHE_ENABLED_DEFAULT_VALUE = true;
+
+  // Keys for HTTP request configuration
```
tdas: no need to duplicate these descriptions in multiple places. recipe for these stuff diverging in the future (in fact the descriptions are already different right now). lets define the apiclientconf as the source truth, and just comment here saying see the ApiClientConf for more details.

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
```diff
@@ -0,0 +1,169 @@
...
+    // Verify the elapsed time is within the jitter-adjusted bounds of the base delay.
+    // Calculated as: baseDelay * (1 ± jitterFactor). In this case, the base delay is 100ms and
+    // the jitter factor is 0.5 so the range is [50ms, 150ms].
+    long elapsedMs = Duration.between(start, clock.now()).toMillis();
+    long baseDelay = conf.getRequestInitialDelayMs();
+    long minDelay = (long) Math.floor(baseDelay * (1 - jitterFactor));
+    long maxDelay = (long) Math.ceil(baseDelay * (1 + jitterFactor));
+    assertThat(elapsedMs)
+        .as("retry delay should stay within jitter-adjusted bounds")
+        .isBetween(minDelay, maxDelay);
```
tdas, line 124: this does not ensure that the jitter is actually being applied. for example, if the jitter was not applied and the delay is actually 100ms, then this assert will still pass right?

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
tdas: is multiplier tested?
is initial delay tested?

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
timothyw553: That's a good point. Another approach that @openinx and I thought of is to let the retry handler take an injected sequence of jitter values, so tests can assert the exact delay that gets applied and fail if jitter isn’t honored.

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
timothyw553: With jitter disabled the tests expect the manual clock to advance by the exact delays computed from the initial delay and multiplier (e.g. 100 ms + 200 ms, or 50 ms + 100 ms). If either value (multiplier or initial delay) were ignored or misapplied, those assertions would fail

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
tdas: cant you just test with and without jitter? without jitter the clock will be exactly 100ms delayed. with jitter it will be != 100ms. 

isnt that good enough?

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
tdas: for initial delay i see that different tests have different delay .. so across teh tests its getting tested that setting the initial delay to different things work correctly.

for multiplier, i see the previous test is setting it to 1. but the previous test is succeeding in only one attempt to multiplier does not kick in. so effectively you are testing multiple with only 1 value = 2? like.. if your internal code is hard coded to 2 and the conf is ignored, all your tests will pass right?

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientConf.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  public ApiClientConf setRequestInitialDelayMs(long requestInitialDelayMs) {
+    Preconditions.checkArgument(requestInitialDelayMs > 0,
+        "Retry initial delay must be positive, but got %s", requestInitialDelayMs);
+    this.requestInitialDelayMs = requestInitialDelayMs;
+    return this;
+  }
+
+  public double getRequestMultiplier() {
```
tdas: i suggest the name getRequestDelayMultiplier 
multiplier is just too opaque
please update all the confs and everything.

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientConf.java
tdas: Sorry.. one more comment.. let move these descriptions either class docs, or the method docs of this class. So that these show up the java docs in IDEs and stuff.

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientConf.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  public ApiClientConf setRequestMultiplier(double requestMultiplier) {
+    Preconditions.checkArgument(requestMultiplier > 0,
+        "Retry multiplier must be positive, but got %s", requestMultiplier);
+    this.requestMultiplier = requestMultiplier;
+    return this;
+  }
+
+  public double getRequestJitterFactor() {
```
tdas: getRequestDelayJitterFactor

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
openinx: hi @tdas ,  I think with the jiter tests we use `!=100ms` can be easily flaky, since it's possible to be 100ms if the `Math.random()` get the number. 

Maybe we can remove all the jitter retry policy in this PR,  and just use the fixed sleep time to make this PR merged. because currently,  seems like the PR involves complicated context enough. And adding the end-to-end jiter sleep test  make the retry policy and the retry behavirour even more coupled.

openinx: For this PR, let's just focus on the key things (not the jitter retry algorithm):

1.  The general retry  framework, we choose to use the `HttpRetryHandler` to handle the general http request retry.
2.  What are the recoverable codes and exceptional error that we need to cover in this case. 

For the jitter retry algorithm, let's try to address in the next PR ?  then we can try to create a clean and decoupling design  (I mean decouple the retry algorithm itself from the retry handler ) to fully address all the test cases that you concerned.

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
timothyw553: ah yes that's true, thanks for pointing that out. updated.

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
tdas: why will anything be flaky? the whole thing is with manual clock.. isnt it?

timothyw553: `Math.random()` could return exactly 0.5, which makes the jitter term zero and the delay stay at 100 ms. That’s vanishingly rare, but it means a "delay != 100 ms" assertion can flake. I guess if we agree that it's basically impossible and can ignore it, it should be totally fine and valid.

openinx: Either we use the `random` sequence or `test-customized`  sequence (some kind of mocking random sequence) to test the jitter.   If we use the `random` then it's possible to be flaky and also hard to have the accurate sleep time assertion.  If we use the `test-customized` sequence to test the jitter, then we finally need to decouple the jitter algorithm out of the http retry framework. 

That's why I suggest to make it into a separate PR to make this forward:
1.  The current PR just merge the http retry change.   ( with fixed sleep time directly).
2.  The next PR replace the `fixed sleep time` with `jitter time`, also add jitter full coverage.

tdas: I see the how flakiness can happen. basically with there is 1 in 100 chance (with 100 ms and jitterfactor = 0.5) that we get exactly 100ms. so 1 in 100 chance that the test for 100ms will fail. 

Thanks for pointing it out.

While I agree the absolutely ideal thing to do is what @openinx suggested ... i am also concerned whether it is worth the testing complexity. Lets just leave it as is in this PR. i am happy to approve. in a follow up PR, pull out the jitter stuff into a separate jitter generator class that can be independently tested, and also reused to generate jitter in other places (like leadTimeMillis). 

How does that sound?

tdas:
Created backport PR for `branch-0.3`:
- #1196 with remaining conflicts!

Please cherry-pick the changes locally and resolve any conflicts.
```bash
git fetch origin backport-1145-to-branch-0.3
git worktree add --checkout .worktree/backport-1145-to-branch-0.3 backport-1145-to-branch-0.3
cd .worktree/backport-1145-to-branch-0.3
git reset --hard HEAD^
git cherry-pick -x 71837e9d2556cc4d740b249c5dc4eb7dd574d32e
git push --force-with-lease
```

tdas:
@TimothyW553 tried to autocreate the backport PR .. but it had conflicts. so i closed that PR. can you make a 0.3 backport PR?

connectors/spark/src/test/java/io/unitycatalog/spark/HttpRetryHandlerTest.java
timothyw553: Yeah that sounds like a good idea. I have some follow up ideas/conerns that I'll leave in the next iteration.

timothyw553:
@tdas yes will do!

PR #1148, opened 2025-10-31
Github action to automatically create backport PR from merged PR's changes using labels

tdas:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

As the title says. Specifically,
- If you add a label "backport branch-0.3" to a PR, once the PR is merged, this github action will automatically create new PR with the merged + squashed changes on branch-0.3
- if there is a merge conflict, then it will create a draft pull request with the first conflict encountered committed.

It uses this market place action - https://github.com/marketplace/actions/backport-merged-pull-requests-to-selected-branches#backport-action

Here is an example usage in a project that i coped - https://github.com/camunda/camunda/blob/main/.github/workflows/backport.yml

PR #1151, opened 2025-11-04
Add sign-off to backport commits

tdas:
**PR Checklist**

  **Description of changes**
Follow up of - https://github.com/unitycatalog/unitycatalog/pull/1148
It created this backport PR - https://github.com/unitycatalog/unitycatalog/pull/1150
But the problem is that the commit in the backported PR dont have the signoff. So even if there are no conflicts, the PR cannot be readily merged.
So I modified the backport GitHub action to automatically amend backport commits with sign-off. The workflow now:
  - Extracts the original commit author's name and email from each backport branch
  - Amends the commit with `--signoff` flag using the original author's credentials
  - Force pushes the updated branch

  This ensures all backported commits include proper sign-off from the original author.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -22,6 +22,7 @@ jobs:
           token: ${{ secrets.BACKPORT_ACTION_TOKEN }}
 
       - name: Create backport PRs
+        id: backport
         uses: korthout/backport-action@v3
         with:
           # Use custom token for backport operations
@@ -50,3 +51,37 @@ jobs:
             {
               "conflict_resolution": "draft_commit_conflicts"
             }
+
+      - name: Sign-off and force push backport branches
+        if: steps.backport.outputs.created_pull_numbers != ''
+        run: |
+          # Get the PR number
+          pr_number="${{ github.event.pull_request.number }}"
+
+          # Find all backport branches for this PR from the local repo
+          # The backport action creates branches named backport-{pr-number}-to-{target-branch}
+          # See https://github.com/korthout/backport-action?tab=readme-ov-file#branch_name
+          backport_branches=$(git branch -r | grep "origin/backport-${pr_number}-to-" | sed 's/origin\///' | sed 's/^[[:space:]]*//' | xargs)
+
+          if [ -n "$backport_branches" ]; then
+            for branch in $backport_branches; do
+              echo "Processing branch: $branch"
+              git checkout "$branch"
+
+              # Get the author name and email from the last commit
+              author_name=$(git log -1 --format='%an')
+              author_email=$(git log -1 --format='%ae')
+
+              echo "Using author: $author_name <$author_email>"
+
+              # Configure git with the original commit author's information
+              git config user.name "$author_name"
+              git config user.email "$author_email"
+
+              # Amend with signoff and force push
+              git commit --amend --no-edit --signoff
+              git push --force origin "$branch"
+            done
+          else
+            echo "No backport branches found for PR #${pr_number}"
+          fi


.github/workflows/backport.yml
```diff
@@ -50,3 +51,36 @@ jobs:
...
             }
+
+      - name: Sign-off and force push backport branches
+        if: steps.backport.outputs.created_pull_numbers != ''
+        run: |
+          # Get the PR number
+          pr_number="${{ github.event.pull_request.number }}"
+
+          # Find all backport branches for this PR
+          # The backport action creates branches named backport-{pr-number}-to-{target-branch}
```
openinx, line 62: So the `backport-{pr-number}-to-{target-branch}` branch name is naturally coming from the `korthout/backport-action@v3` I think. 

https://github.com/korthout/backport-action?tab=readme-ov-file#branch_name

Could you please a comment here, for the first glance, I was not aware that and confused. After did the google search , just get it.

.github/workflows/backport.yml
```diff
@@ -50,3 +51,36 @@ jobs:
...
+              author_email=$(git log -1 --format='%ae')
+
+              echo "Using author: $author_name <$author_email>"
+
+              # Configure git with the original commit author's information
+              git config user.name "$author_name"
+              git config user.email "$author_email"
+
+              # Amend with signoff and force push
+              git commit --amend --no-edit --signoff
```
openinx, line 82: So the backport PR's credit is for author of original PR, right ?  If there is any conflict, we still require the original author to resolve them, right ?   will the one who trigger the backport workflow have the permission to resolve them.  ( I mean the one who trigger the backport, may is not exactly the one who contributed the original PR, then what's the collaboration process for them). 

Just curious.

.github/workflows/backport.yml
```diff
@@ -50,3 +51,36 @@ jobs:
...
+              git push --force origin "$branch"
```
openinx, line 83: Will the workflow have the enough permissions to force push, act as the original author ?   I'm not sure.

.github/workflows/backport.yml
```diff
@@ -50,3 +51,36 @@ jobs:
...
+
+      - name: Sign-off and force push backport branches
+        if: steps.backport.outputs.created_pull_numbers != ''
+        run: |
+          # Get the PR number
+          pr_number="${{ github.event.pull_request.number }}"
+
+          # Find all backport branches for this PR
+          # The backport action creates branches named backport-{pr-number}-to-{target-branch}
+          backport_branches=$(git branch -r | grep "origin/backport-${pr_number}-to-" | sed 's/origin\///' | sed 's/^[[:space:]]*//' | xargs)
```
openinx, line 64: A related question is:   will those backport branches be maintained in the offical  oss unitycatalog ?  After the backport PR get merged,  will those stale branched be cleaned automatically ?

openinx (approved):
Just left serveral questions, may no need to change.  Approved for later changes.

.github/workflows/backport.yml
tdas: good point!

.github/workflows/backport.yml
tdas: Good question. if there is a conflict, I think there are two possible approaches
- original author can close the auto-generated backport PR and create a new PR and proceed.
- a committer who has write access to the repo can push changes to the branch with signoff.

My goal is to make the simple, non-conflict case super easy.

.github/workflows/backport.yml
tdas: good point. i cant test it without actually merging it but i can see that this will be a problem. adding the token used earlier.

tdas: claude thinks it should work because of the token setting in the checkout@v4 action above is setting the 


---

 From the [GitHub Actions Checkout README](https://github.com/actions/checkout/blob/main/README.md):

  On the token parameter:
  "Personal access token (PAT) used to fetch the repository. The PAT is configured with the local git config, which enables your scripts to
   run authenticated git commands."

  On the persist-credentials parameter (which defaults to true):
  "Whether to configure the token or SSH key with the local git config

So yes, when you use actions/checkout@v4 with token: ${{ secrets.BACKPORT_ACTION_TOKEN }}, it automatically configures the local git
  config with that token, and subsequent git operations (like git push --force) will use that authentication. The persist-credentials
  option defaults to true, so the credentials remain available throughout your workflow.

.github/workflows/backport.yml
openinx: Sounds good, thanks for the explanation !

.github/workflows/backport.yml
openinx: So the CI directly push it to the `backport-{pr-number}-to-{target-branch}` in the official `unitycatalog/unitycatalog` repo, or a separate forked repo ?  Seems like it's the former one, right ?  Then when do we need to clean the stale branches once we merged the backport PR.

.github/workflows/backport.yml
tdas: good question. once we confirm this is working as intended, I can do a follow up PR that adds a actoin to delete backport PR branches that have been merged and are more then 30 days old. sounds good?

.github/workflows/backport.yml
tdas: yes. main repo.  once we confirm this is working as intended, I can do a follow up PR that adds a actoin to delete backport PR branches that have been merged and are more then 30 days old. sounds good?

PR #1154, opened 2025-11-05
Allow toStandardizedURIString to handle null value gracefully.

yili-db:
**PR Checklist**

**Description of changes**
Allow toStandardizedURIString to handle null value gracefully.


server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -216,7 +216,7 @@ public static void assertValidLocation(String location) {
    * the input as a local file path and converts it to a "file://" URI.
    *
    * @param inputPath the input path or URI to be standardized.
-   * @return the standardized URI string.
+   * @return the standardized URI string, or null if inputPath is null.
```
yili-db, line 219: I have no evidence on why the inputPath should be null yet. But before this change it used to allow null->null. So I am making a fix here to restore the original logic.

PR #1168, opened 2025-11-07
Update Daft URLs

desmondcheongzx:
**PR Checklist**

- (NA) If there is a related issue, make sure it is linked to this PR.
- (NA) If you've fixed a bug or added code that should be tested, add tests!
- (NA) If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

Updates Daft urls:
- getdaft.io -> daft.ai
- getdaft.io/projects/docs/en/stable/integrations/unity_catalog/ -> docs.daft.ai/en/stable/connectors/unity_catalog/

### DIFF
--- a/docs/integrations/unity-catalog-daft.md
+++ b/docs/integrations/unity-catalog-daft.md
@@ -2,7 +2,7 @@
 
 This page shows you how to use Unity Catalog with Daft.
 
-[Daft](https://getdaft.io/) is a library for parallel and distributed processing of multimodal data.
+[Daft](https://www.daft.ai/) is a library for parallel and distributed processing of multimodal data.
 
 ## Set up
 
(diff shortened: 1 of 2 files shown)


desmondcheongzx:
Hey folks! Had an additional question here too.

A user reported that the live version of the UC docs (https://docs.unitycatalog.io/integrations/unity-catalog-daft/#set-up)  uses an old pip install command `pip install -U "getdaft[unity,deltalake]"` which no longer works (we now use the `daft` project on pypi so it should be `pip install -U "daft[unity,deltalake]"`).

But I also noticed that the relevant docs page was updated 6 months ago - are we missing something to update the live version of the docs?

yili-db:
Updated by #1049

Issue #1143, opened 2025-10-29
Support UC managed tables.


yili-db:
UC managed tables is type a tables that UC manages their storage locations. The way to create a managed table is:
1. send a request to UC to create a staging table first, then use the returned staging location to construct and create a table
2. or just send a createTable request to UC with table_type=MANAGED and storage_location=null.

Either way, a UC managed location like {storage-root.tables}/tables/{tableId} will be returned.

PR #1166, opened 2025-11-06
Add basic support of Delta commits for managed tables

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1166/files) to review incremental changes.
- [**stack/commit_basic**](https://github.com/unitycatalog/unitycatalog/pull/1166) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1166/files)]
  - [stack/commit_api_2](https://github.com/unitycatalog/unitycatalog/pull/1199) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1199/files/8c6992aaad6b981ef38a4fb468ef700f083ad49a..d0d4f606101bc705a6a79514a2b4ba31b2b19b19)]
    - [stack/commit_backfill_delete_meta](https://github.com/unitycatalog/unitycatalog/pull/1169) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1169/files/d0d4f606101bc705a6a79514a2b4ba31b2b19b19..93dc85893930d1594b3a974cc48228c767f826ab)]
      - [stack/commit_get](https://github.com/unitycatalog/unitycatalog/pull/1170) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1170/files/93dc85893930d1594b3a974cc48228c767f826ab..0e7306a1c14f5476eeef78953b948fd2f7ba0453)]
        - [stack/commit_access_test](https://github.com/unitycatalog/unitycatalog/pull/1178) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1178/files/0e7306a1c14f5476eeef78953b948fd2f7ba0453..6042111d1b885fc5a9cf687b670c98b95ade14d7)]
          - [stack/ccv2_spark](https://github.com/unitycatalog/unitycatalog/pull/1186) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1186/files/6042111d1b885fc5a9cf687b670c98b95ade14d7..cc09752cb91d6c9e6bf474ebc05c82f7a0da95cc)]

---------
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [x] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

This commit adds API support for committing Delta table versions to Unity Catalog
[...]

PR #1169, opened 2025-11-07
Complete the Delta commit posting functionality

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1169/files) to review incremental changes.
- [**stack/commit_backfill_delete_meta**](https://github.com/unitycatalog/unitycatalog/pull/1169) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1169/files)]
  - [stack/commit_get](https://github.com/unitycatalog/unitycatalog/pull/1170) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1170/files/eee17f0c3f9e6a456523fb0bf502b560a31ddea7..6e1ceb0c6747f1c85bc344b25b6d982dc964a8b3)]
    - [stack/commit_access_test](https://github.com/unitycatalog/unitycatalog/pull/1178) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1178/files/6e1ceb0c6747f1c85bc344b25b6d982dc964a8b3..7cb3ace271839563449be4a900f5260633c423b5)]
      - [stack/ccv2_spark](https://github.com/unitycatalog/unitycatalog/pull/1186) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1186/files/7cb3ace271839563449be4a900f5260633c423b5..72eff6e86a48ecdcddbd407a5d7f5edabad35a7c)]

---------
**PR Checklist**

**Description of changes**

This commit implements backfill and metadata update functionality for Delta commits in Unity Catalog:
- Backfilling and deletion of old commit metadata
- Commit limit enforcement
- Table metadata updates via commits
- Comprehensive test coverage

**Related issue**

### DIFF
--- a/api/all.yaml
+++ b/api/all.yaml
@@ -2976,7 +2976,7 @@ components:
         - file_modification_timestamp
     DeltaCommitResponse:
       type: object
-      description: The response for the commit action
+      description: The response for the Delta commit action
     DeltaGetCommits:
       type: object
       properties:
--- a/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
@@ -33,23 +33,51 @@ import org.hibernate.query.Query;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** Repository for managing coordinated commits for managed Delta tables in Unity Catalog. */
+/**
+ * Repository for managing Delta commits for managed Delta tables in Unity Catalog.
+ *
+ * <p>The database table 'uc_delta_commits' has commits of all managed tables until the commits are
+ * backfilled. However, there's a special case that in order to record the last commit even after
+ * it's backfilled, UC does not delete the last commit but instead only mark it as
+ * is_backfilled_latest_commit=true. So it's guaranteed that there would be at least one record (the
+ * last commit) once the table is onboarded.
+ *
+ * <p>For example, consider the following sequence of commit operations:
+ *
+ * <ol>
+ *   <li>commit(v1). Then database has [v1]
+ *   <li>commit(v2). Then database has [v1, v2]
+ *   <li>commit(v3). Then database has [v1, v2, v3]
+ *   <li>backfill(v1). Then database has [v2, v3]. Any version <= v1 is removed.
+ *   <li>commit(v4) & backfill(v3). Then database has [v4]. Any version <= v3 is removed.
+ *   <li>commit (v5). Then database has [v4, v5]
+ *   <li>backfill(v5). Then database has [v5(is_backfilled_latest_commit=true)]. v4 is removed. But
+ *       v5 has to be kept as the last record in database.
+ *   <li>commit(v6). Then database has [v5(is_backfilled_latest_commit=true), v6]
+ * </ol>
+ */
 public class DeltaCommitRepository {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(DeltaCommitRepository.class);
 
   /**
-   * The maximum number of commits allowed per table before backfilling is required. TODO: turn this
-   * into a configurable server property.
-   */
-  private static final int MAX_NUM_COMMITS_PER_TABLE = 50;
-
-  /**
-   * The batch size limit for delete operations. Set larger than MAX_NUM_COMMITS_PER_TABLE to handle
-   * cases where tables exceed the maximum allowed commits. TODO: turn this into a configurable
+   * The maximum number of unbackfilled commits allowed per table before backfilling is required. In
+   * real life unbackfilled commits per table should remain 1 or 2 almost all the time as the client
+   * should implement proactive backfilling right after committing a version. This limit exist as a
+   * safety measure just in case there's a problem in client implementation, or the table is being
+   * committed heavily by different clients in rare cases. TODO: turn this into a configurable
    * server property.
    */
-  private static final int NUM_COMMITS_PER_BATCH = 100;
+  private static final int MAX_NUM_COMMITS_PER_TABLE = 10;
+
+  /**
+   * The batch size limit for commit delete and select operations. This limit is set to be larger
+   * than MAX_NUM_COMMITS_PER_TABLE so that it should never hit this limit at all. But it serves the
+   * purpose of another safety measure to avoid a huge query execution IF both the server and client
+   * are not implemented correct and commits per table grow unbounded. TODO: turn this into a
+   * configurable server property.
+   */
+  private static final int NUM_COMMITS_PER_BATCH = 20;
 
   private final SessionFactory sessionFactory;
   private final ServerProperties serverProperties;
@@ -59,7 +87,24 @@ public class DeltaCommitRepository {
     this.serverProperties = serverProperties;
   }
 
-  /** Commits a new version to a managed Delta table with coordinated commit semantics. */
+  /**
+   * Commits a new version to a managed Delta table with coordinated commit semantics.
+   *
+   * <p>This method handles three types of commit operations:
+   *
+   * <ul>
+   *   <li><b>Onboarding commit:</b> The first commit sent to Unity Catalog for this table
+   *   <li><b>Normal commit:</b> A new version commit with optional backfill notification
+   *   <li><b>Backfill-only commit:</b> No new version, only reports backfilled versions
+   * </ul>
+   *
+   * <p>The method validates the commit, ensures the table is a managed Delta table, and performs
+   * the appropriate commit operation within a transaction.
+   *
+   * @param commit the commit request containing version info, metadata, and backfill information
+   * @throws BaseException if the commit is invalid, table is not found, or commit limits are
+   *     exceeded
+   */
   public void postCommit(DeltaCommit commit) {
     serverProperties.checkManagedTableEnabled();
     validateCommit(commit);
@@ -76,9 +121,9 @@ public class DeltaCommitRepository {
           if (firstAndLastCommits.isEmpty()) {
             handleOnboardingCommit(session, tableId, tableInfoDAO, commit);
           } else {
-            DeltaCommitDAO firstCommit = firstAndLastCommits.get(0);
-            DeltaCommitDAO lastCommit = firstAndLastCommits.get(1);
-            assert firstCommit.getCommitVersion() <= lastCommit.getCommitVersion();
+            DeltaCommitDAO firstCommitDAO = firstAndLastCommits.get(0);
+            DeltaCommitDAO lastCommitDAO = firstAndLastCommits.get(1);
+            assert firstCommitDAO.getCommitVersion() <= lastCommitDAO.getCommitVersion();
             if (commit.getCommitInfo() == null) {
               // This is already checked in validateCommit()
               assert commit.getLatestBackfilledVersion() != null;
@@ -86,10 +131,11 @@ public class DeltaCommitRepository {
                   session,
                   tableId,
                   commit.getLatestBackfilledVersion(),
-                  firstCommit.getCommitVersion(),
-                  lastCommit.getCommitVersion());
+                  firstCommitDAO.getCommitVersion(),
+                  lastCommitDAO.getCommitVersion());
             } else {
-              handleNormalCommit(session, tableId, tableInfoDAO, commit, firstCommit, lastCommit);
+              handleNormalCommit(
+                  session, tableId, tableInfoDAO, commit, firstCommitDAO, lastCommitDAO);
             }
           }
           return null;
@@ -99,11 +145,20 @@ public class DeltaCommitRepository {
   }
 
   /**
-   * Handle an onboarding commit which is the very first commit sent to UC, and it's commiting a
-   * version. This commit does not do backfill because there is no version prior to this commit.
-   * This may be the first ever commit of the table since creation, or that the previous commits
-   * were just FS commits. In any way, the table will use UC as commit coordinator starting from
-   * this commit.
+   * Handles an onboarding commit, which is the very first commit sent to Unity Catalog for a table.
+   *
+   * <p>An onboarding commit must include commit information (version, timestamp, file details) but
+   * does not perform backfilling since there are no prior Unity Catalog-managed versions. This may
+   * be the first commit since table creation, or the table may have had previous filesystem-only
+   * commits. After this commit, Unity Catalog becomes the commit coordinator for the table.
+   *
+   * <p>The method saves the commit and optionally updates table metadata if provided.
+   *
+   * @param session the Hibernate session for database operations
+   * @param tableId the unique identifier of the table being committed to
+   * @param tableInfoDAO the table information data access object
+   * @param commit the commit request containing version info and optional metadata
+   * @throws BaseException if the commit info is null
    */
   private static void handleOnboardingCommit(
       Session session, UUID tableId, TableInfoDAO tableInfoDAO, DeltaCommit commit) {
@@ -118,8 +173,18 @@ public class DeltaCommitRepository {
   }
 
   /**
-   * Handles a commit request that only performs backfilling without adding a new commit. This
-   * validates the backfill version and delegates to the backfill logic.
+   * Handles a commit request that only performs backfilling without adding a new commit version.
+   *
+   * <p>This method is called when a commit request has no commit info but specifies a backfilled
+   * version. It validates that the backfilled version is not greater than the last committed
+   * version, then delegates to the backfill logic to remove old commits from the repository.
+   *
+   * @param session the Hibernate session for database operations
+   * @param tableId the unique identifier of the table
+   * @param latestBackfilledVersion the version up to which backfilling has already been performed
+   * @param firstCommitVersion the version number of the first commit currently in the database
+   * @param lastCommitVersion the version number of the last commit currently in the database
+   * @throws BaseException if the backfilled version is greater than the last commit version
    */
   private static void handleBackfillOnlyCommit(
       Session session,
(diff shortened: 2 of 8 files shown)


server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -9,24 +9,47 @@
...
 import org.hibernate.SessionFactory;
+import org.hibernate.query.NativeQuery;
 import org.hibernate.query.Query;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Repository for managing coordinated commits for managed Delta tables in Unity Catalog. */
 public class CommitRepository {
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(CommitRepository.class);
```
openinx: So seems like this should be the dependency PR https://github.com/unitycatalog/unitycatalog/pull/1166 right ?

yili-db: https://github.com/unitycatalog/unitycatalog/pull/1166 doesn't use the logger so it doesn't introduce it. This PR is the first time to use logger here.

api/all.yaml
```diff
@@ -2980,7 +2947,7 @@ components:
           description: The array of column descriptions.
     CommitInfo:
       type: object
-      description: Represents a coordinated commit version.
+      description: Represents a delta commit version.
```
```suggestion
      description: Represents a Delta commit version.
```

please make sure Delta is capitalized everywhere

api/all.yaml
```diff
@@ -3010,6 +2977,38 @@ components:
     CommitResponse:
       type: object
       description: The response for the commit action
```
```suggestion
      description: The response for the Delta commit action
```

api/all.yaml
```diff
@@ -1431,44 +1431,11 @@ paths:
...
-            larger start_version according to the response to get the remaining commits.
-          required: false
-          schema:
-            type: integer
-            format: int64
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/GetCommits'
```
tdas: lets call this GetDeltaCommits

api/all.yaml
tdas: can we name this DeltaCommit

tdas: DeltaCommitResponse

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -370,6 +370,9 @@ public void deleteTable(Session session, UUID schemaId, String tableName) {
       } catch (Throwable e) {
         LOGGER.error("Error deleting table directory: {}", tableInfoDAO.getUrl(), e);
       }
+      repositories
+          .getDeltaCommitRepository()
+          .permanentlyDeleteTableCommits(session, tableInfoDAO.getId());
```
tdas, line 375: good catch.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -9,25 +9,48 @@
...
 /** Repository for managing coordinated commits for managed Delta tables in Unity Catalog. */
 public class DeltaCommitRepository {
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(DeltaCommitRepository.class);
+
+  /**
+   * The maximum number of commits allowed per table before backfilling is required. TODO: turn this
+   * into a configurable server property.
+   */
+  private static final int MAX_NUM_COMMITS_PER_TABLE = 50;
```
tdas: lets keep this fairly small. say 10. because number of commits across all the table can get large. so limit of 10 means commit table size 10x # tables.

tdas: also `MAX_NUM_UNBACKFILLED_COMMITS_PER_TABLE`

btw.. is it backfilling or publishing .. what is the terminology?

tdas: if that feels too long... then you can remove NUM. MAX implies number.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -9,25 +9,48 @@
...
+
+  /**
+   * The batch size limit for delete operations. Set larger than MAX_NUM_COMMITS_PER_TABLE to handle
```
tdas: delete operations of what? how is it related to the MAX_NUM_COMMITS_PER_TABLE

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -85,16 +113,49 @@ private static void handleOnboardingCommit(
...
    * Optionally in the same commit it may also report backfilled version and/or update metadata.
    */
   private static void handleNormalCommit(
-      Session session, UUID tableId, DeltaCommit commit, DeltaCommitDAO lastCommit) {
+      Session session,
+      UUID tableId,
+      TableInfoDAO tableInfoDAO,
+      DeltaCommit commit,
+      DeltaCommitDAO firstCommit,
+      DeltaCommitDAO lastCommit) {
```
tdas: this is getting complicated. you are taking 3 different commits. this needs param docs. what does first and last mean?

tdas: also make the DAO type params into have DAO in the name to differentiate from non-DAO commit.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -109,18 +170,206 @@ private static void handleNormalCommit(
...
+    Optional<Long> latestBackfilledVersion =
+        Optional.ofNullable(commit.getLatestBackfilledVersion());
+    if (latestBackfilledVersion.filter(x -> x > lastCommitVersion).isPresent()) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT,
+          String.format(
+              "Latest backfilled version %d cannot be greater than the last commit version = %d",
+              latestBackfilledVersion.get(), lastCommitVersion));
+    }
+    long effectiveBackfilledVersion =
```
tdas: this variable is only used for checking commit limit. cant this be pushed into the checkCommitLimit function? seems like a better abstraction.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -109,18 +170,206 @@ private static void handleNormalCommit(
...
+    }
+  }
+
+  /**
+   * Validates that adding the new commit will not exceed the maximum number of commits allowed per
+   * table after the commit (and backfill if any) is finished.
+   */
+  private static void checkCommitLimit(
+      UUID tableId, long effectiveBackfilledVersion, long newCommitVersion) {
+    long expectedFirstCommitAfterBackfill = effectiveBackfilledVersion + 1L;
```
tdas: this is not 
```suggestion
    long expectedFirstVersionAfterBackfill = effectiveBackfilledVersion + 1L;
```
lets be consistent on when to use commit and when version. 
commit = entire commit object. 
version = the commit version, an integer. 

please check all the code in this file to see if this is definition is consistent.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: At this moment the RPC field is still called `latest_backfilled_version`.
It is also called "backfill" in Delta code like `UCCommitCoordinatorClient.java` and `UCTokenBasedRestClient.java`.
In some docs I've seen the word "publish" but almost everywhere in code it's called "backfill".
While the spec draft doc is being discussed this might change in future. But consider that it's already named as "backfill" in RPC field I don't think that would change easily.

yili-db: re: 50 vs 10
In real life commits per table should remain 1 or 2 almost all the time as long as the client implements correctly. It should never reach the limit regardless of being 10 or 50.
This limit only exist just in case there's a problem in client implementation, or the table is being committed heavily by different clients like in a load test. It's a safety measure.
I can add this explanation to the comment.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: It's for deletion operations of commits when backfilling or deleting a table.
In real life we should never need this limit. So in theory this is simply redundant when it's bigger than `MAX_NUM_COMMITS_PER_TABLE` because any commit deletion or selection query execution for a table shouldn't hit more than that many rows. We should feel free to delete or scan the entire 50 max rows without the `LIMIT` clause.
It's simply yet another safety measure to make sure a single query doesn't cause a huge query execution **if** something was implemented incorrectly and the number of unbackfilled commits accumulate unbounded.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
tdas: yes. thank you for adding this.

tdas: and yes. lets use backfill for now and be consistent with everything else.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -109,18 +237,288 @@ private static void handleNormalCommit(
...
+        session.createNativeQuery(
+            "UPDATE uc_delta_commits SET is_backfilled_latest_commit = true WHERE table_id = :tableId "
+                + "AND commit_version = :commitVersion");
+    query.setParameter("tableId", tableId);
+    query.setParameter("commitVersion", commitVersion);
+    query.executeUpdate();
+  }
+
+  /**
+   * Retrieves the first and last commits for a table ordered by version number.
```
tdas: These are not actually first and last commit of the table.. these are the first and last UNBACKFILLED commit of the table .. isnt it?

yili-db: Clarified a bit in the comment.
They are the the first and last commit of the table **in the database**. They may or may not be backfilled already.
For example, in the following order of operation:
1. commit v1 (in database there's [v1])
2. backfill v1 (in database there's [v1(backfilled)]
3. commit v2: `postCommit` will call `getFirstAndLastCommits` and get [v1(backfilled), v1(backfilled)]
4. commit v3: `postCommit` will call `getFirstAndLastCommits` and get [v1(backfilled), v2]

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
tdas: lets reduce the number of 10 nonetheless. our set up is much less scalable than any SAAS offering in the cloud.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -109,18 +237,293 @@ private static void handleNormalCommit(
...
+
+  /**
+   * Retrieves the first and last commits in database for a table ordered by version number.
+   *
+   * <p>The first commit is the commit in database with the lowest version number and earliest
+   * commit time.
+   *
+   * <p>The last commit is the commit in database with the highest version number and latest commit
+   * time.
+   *
```
tdas: is the monotonicity increase of commit timestamp with version number guaranteed?
if so, how? who decides the timestamp?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -109,18 +237,293 @@ private static void handleNormalCommit(
...
+   * <p>Uses a UNION ALL query to efficiently fetch both boundary commits in a single database
+   * operation. The results are sorted to ensure consistent ordering.
+   *
+   * <p>Return value interpretation:
+   *
+   * <ul>
+   *   <li>Empty list: no commits exist for this table
+   *   <li>List with two identical commits: only one commit exists (returned twice for consistency)
+   *   <li>List with two different commits: [firstCommit, lastCommit] by version number
```
tdas, line 536: once a table has been onboarded, is it guaranteed that there will be single commit always in the database? basically is it not possible for a client to report the all upto latest commit (say 5) has been backfilled (so <=5 can be removed from the database) without creating a new commit (that is, 6)?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -109,18 +237,293 @@ private static void handleNormalCommit(
...
+   */
+  private static long getEffectiveBackfilledVersion(
+      UUID tableId,
+      Optional<Long> latestBackfilledVersion,
+      DeltaCommitDAO firstCommitDAO,
+      DeltaCommitDAO lastCommitDAO) {
+    if (lastCommitDAO.getIsBackfilledLatestCommit()) {
+      // There should only be ONE commit if the last one is marked as backfilled which is the
+      // special case.
+      if (!Objects.equals(firstCommitDAO.getCommitVersion(), lastCommitDAO.getCommitVersion())) {
```
tdas: why do we have to use Objects.equals. its just long isnt it?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -109,18 +237,293 @@ private static void handleNormalCommit(
...
+      return lastCommitDAO.getCommitVersion();
+    } else if (latestBackfilledVersion
+        .filter(x -> x >= firstCommitDAO.getCommitVersion())
+        .isPresent()) {
+      // The commit request reports a valid backfilled version. We'll take that.
+      return latestBackfilledVersion.get();
+    } else if (firstCommitDAO.getIsBackfilledLatestCommit()) {
+      // The firstCommit is already backfilled. It remains only because it was the only commit
+      // left when being backfilled.
+      return firstCommitDAO.getCommitVersion();
```
tdas, line 337: isnt this an invariant that getIsBackfilledLatestCommit is true if and only if firstCommit = lastCommit? 

if so, cant this logic be simplified.

yili-db: In this example, it can happen to be `[v1(is_backfilled_latest_commit=true), v2]`
https://github.com/unitycatalog/unitycatalog/pull/1169#discussion_r2524983238

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: I had `Long` in the DAO. Now I change them into `long`.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: Actually there's no guarantee of timestamp as it's sent by client. Removing this wording from this comment.

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
```diff
@@ -214,5 +301,292 @@ public void testInvalidParameters() throws ApiException {
...
+    DeltaCommit commit =
+        createCommitObject(tableInfo.getTableId(), 1L, tableInfo.getStorageLocation())
+            .metadata(metadata);
+    deltaCommitsApi.commit(commit);
+
+    // Retrieve the table and verify properties were updated
+    TableInfo updatedTable = tableOperations.getTable(fullTableName);
+    assertNotNull(updatedTable.getProperties());
+    assertEquals("customValue", updatedTable.getProperties().get("customProperty"));
+    assertEquals("enabled", updatedTable.getProperties().get("delta.feature.test"));
```
tdas, line 510: arent you tested dropping properties?

tdas (approved):
LGTM. This was a good completion of the posting functionality.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -109,18 +259,291 @@ private static void handleNormalCommit(
...
+   *
+   * <p>The logic handles several cases:
+   *
+   * <ul>
+   *   <li>If the last commit is already marked as backfilled, it remains the backfilled version
+   *   <li>If the request specifies a valid backfilled version >= first commit, use that
+   *   <li>If the first commit is marked as backfilled, use its version
+   *   <li>Otherwise, nothing is backfilled yet (returns firstCommitVersion - 1)
+   * </ul>
+   *
```
tdas, line 301: I really liked this example - https://github.com/unitycatalog/unitycatalog/pull/1169#discussion_r2524983238

can put this example in the comment. it made it very clear.

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
yili-db: Line 579 tested with `delta.feature.test` dropped.

Issue #1175, opened 2025-11-07
Add Azure ABFS Credential renewal integration test


openinx:
This task is similar to the [S3 credential renewal integration tests](https://github.com/unitycatalog/unitycatalog/issues/1092), which means we need to add:
*  Locally run integration tests
*  Real production run integration tests

openinx:
For the abfss real production integration test,  I created this PR: https://github.com/unitycatalog/unitycatalog/pull/1202 . 

And use this command to run the integraiton test, but I encountered the following error: 

```bash
export CATALOG_URI=$UC_URI
export CATALOG_AUTH_TOKEN=$UC_TOKEN
export CATALOG_NAME=main
export ABFSS_BASE_LOCATION=$UC_ABFSS_BASE_LOCATION

SBT_OPTS="-Xmx8G -XX:+UseG1GC" \
./build/sbt "integrationTests/testOnly io.unitycatalog.integrationtests.SparkCredentialRenewalTest"
```

The error message is here: 

```java
[error] Test io.unitycatalog.integrationtests.SparkCredentialRenewalTest failed: java.nio.file.AccessDeniedException: Operation failed: "Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.", 403, HEAD, https://****** rId: c98d1d2a-d01f-0055-1672-57d1d9000000, took 5.128s
[error]     at org.apache.hadoop.fs.azurebfs.AzureBlobFileSystem.checkException(AzureBlobFileSystem.java:1526)
[error]     at org.apache.hadoop.fs.azurebfs.AzureBlobFileSystem.create(AzureBlobFileSystem.java:347)
[error]     at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:1231)
[error]     at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:1208)
[error]     at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:1089)
[error]     at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:1076)
[error]     at io.delta.storage.HadoopFileSystemLogStore.writeWithRename(HadoopFileSystemLogStore.java:107)
[error]     at io.delta.storage.AzureLogStore.write(AzureLogStore.java:54)
[error]     at org.apache.spark.sql.delta.storage.LogStoreAdaptor.write(LogStore.scala:489)
[...]

PR #1176, opened 2025-11-07
Integrate AbfsVendedTokenProvider to the unity catalog

openinx:
**PR Checklist**

**Description of changes**

Related issue: https://github.com/unitycatalog/unitycatalog/issues/1175

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/AzureCredentialTestFileSystem.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/AzureCredentialTestFileSystem.java
@@ -1,12 +1,12 @@
 package io.unitycatalog.spark;
 
-import static org.apache.hadoop.fs.azurebfs.constants.ConfigurationKeys.FS_AZURE_SAS_TOKEN_PROVIDER_TYPE;
+import static io.unitycatalog.spark.UCHadoopConf.FS_AZURE_SAS_TOKEN_PROVIDER_TYPE;
 import static org.assertj.core.api.Assertions.assertThat;
 
-import com.google.common.base.Objects;
 import io.unitycatalog.spark.auth.AbfsVendedTokenProvider;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
+import org.sparkproject.guava.base.Objects;
 
 public class AzureCredentialTestFileSystem extends CredentialTestFileSystem {
   private volatile AbfsVendedTokenProvider provider;


connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
openinx: I think we need to remove those `azurefs` packages import here, otherwise it will require use to add the `hadoop-azure` dependency even if people just access the aws s3 storage.  Which will highly confuse the end users. 

My suggestion is making those to be some static defined configuration keys in the `UCHadoopConf`, that will be better.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -1,8 +1,8 @@
 package io.unitycatalog.spark.auth;
 
-import static org.apache.hadoop.fs.azurebfs.constants.ConfigurationKeys.FS_AZURE_ACCOUNT_AUTH_TYPE_PROPERTY_NAME;
```
openinx, line 3: Since the `CredPropsUtil` is a common code path that will also be used by `AwsVendedTokenProvider`, so we definitely don't want to add the `hadoop-azure`  when runing spark job on top of the aws s3. 

That's why I removed all those hadoop-azure dependency, so that we won't mess up the cloud vendor 's runtime.

openinx: The approache is simple, we just copy all those configure keys into the `UCHadoopConf`. Please see: https://github.com/unitycatalog/unitycatalog/pull/1176/files#diff-e53d40cc08a4f4120bad08ea094a311a2e36c689b044767bc02a956e78920293R23-R30

tdas: That makes complete sense. thank you for catching that.

How do we make sure that we never break this in the future? today every test was passing using these imports. That means that we are compiling with azure libraries as a compile-time dependency... isnt it? if so, should we change it to not dependent on azure/aws/gcs libraries and everything should still compile and work?

openinx: One simple thought, is adding an checkstyle enforcement rules  in the `checkstyle-dev.xml` to explicitly forbidden any `aws`, `azure`, `gcp` filesystem imports in the production code base.

PR #1178, opened 2025-11-10
Add an access control test to verify that correct permission is enforced when committing to UC and getting commits from UC.

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1178/files) to review incremental changes.
- [**stack/commit_access_test**](https://github.com/unitycatalog/unitycatalog/pull/1178) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1178/files)]
  - [stack/ccv2_spark](https://github.com/unitycatalog/unitycatalog/pull/1186) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1186/files/cb4d52ef1a52d05e27734946528cc6c48873fd5d..c0ddf0bc7eb7cfbc63786e9fd7aa810019e4bf25)]

---------
**PR Checklist**

Add an access control test to verify that correct permission is enforced when committing to UC and getting commits from UC.

**Related issue**

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
+++ b/server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
@@ -122,8 +122,15 @@ public class SdkDeltaCommitsCRUDTest extends BaseTableCRUDTestEnv {
             .endVersion(endVersion.orElse(null)));
   }
 
-  // expectedCommits can either be a list of DeltaCommit, or a list of version numbers, or empty
-  // if it is not expected to return any commits.
+  /** Get all the commits of a table and verify that the response has the latest version as
+   * {@code expectedLatestTableVersion} and it contains all the {@code expectedCommits} in the
+   * exact order.
+   *
+   * @param expectedLatestTableVersion the expected latestTableVersion in {@code response}
+   * @param expectedCommits Can be either a list of {@code DeltaCommit}, or a list of Integer/Long
+   *                        representing the version numbers, or empty if it's not expected to
+   *                        return any commits.
+   */
   private void verifyDeltaCommits(long expectedLatestTableVersion, Object... expectedCommits)
       throws ApiException {
     DeltaGetCommitsResponse response =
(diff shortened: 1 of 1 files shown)


server/src/test/java/io/unitycatalog/server/sdk/access/SdkDeltaCommitsAccessControlCRUDTest.java
```diff
@@ -0,0 +1,213 @@
...
+ *   <li>Clients without valid authentication cannot access the API
+ *   <li>Correct HTTP error codes (401 Unauthorized, 403 Forbidden) are returned
+ * </ul>
+ */
+public class SdkDeltaCommitsAccessControlCRUDTest extends SdkAccessControlBaseCRUDTest {
+
+  private static final String READ_USER_EMAIL = "reader@example.com";
+  private static final String WRITE_USER_EMAIL = "writer@example.com";
+  private static final String NO_ACCESS_USER_EMAIL = "noaccess@example.com";
+
```
tdas, line 47: cant we share this set up code with accross AccessControl tests?

server/src/test/java/io/unitycatalog/server/sdk/access/SdkDeltaCommitsAccessControlCRUDTest.java
```diff
@@ -0,0 +1,213 @@
...
+        .latestBackfilledVersion(latestBackfilledVersion);
+  }
+
+  @SneakyThrows
+  private DeltaGetCommitsResponse getCommits(DeltaCommitsApi api) {
+    return api.getCommits(
+        new DeltaGetCommits()
+            .tableId(tableInfo.getTableId())
+            .tableUri(tableInfo.getStorageLocation())
+            .startVersion(0L));
```
tdas, line 145: no way to share these utils across the other DeltaCommit tests?

server/src/test/java/io/unitycatalog/server/sdk/access/SdkDeltaCommitsAccessControlCRUDTest.java
yili-db: It seems that other access control tests don't just need reader/writer accounts. For example `SdkStagingTableAccessControlTest` has two users with the same `USE_SCHEMA` but expect that only the owner of staging table can commit it.
If in the future we have more access control test that share this pattern, part of this test can be refactored into the base class.

server/src/test/java/io/unitycatalog/server/sdk/access/SdkDeltaCommitsAccessControlCRUDTest.java
yili-db: `getCommits` can not be shared with `SdkDeltaCommitsCRUDTest`. This test doesn't care about any of the parameters but does need a different `DeltaCommitsApi ` object to be used each time.

`createCommitObject` and `createBackfillOnlyCommitObject` are very similar in both tests. But the this test and `SdkDeltaCommitsCRUDTest` are based on totally different base classes until `BaseCRUDTest` where I don't think it should have the commit related helper functions. Moving them to a separate util class isn't ideal because they need to access the `tableInfo` in order to be convenient to use.

`createManagedTable` can be useful when there is a second SDK access control test that needs a table created. It can be the access control test version of `BaseTableCRUDTestEnv`. Or this can be an overkill. We'll know when that future access control test comes.

PR #1179, opened 2025-11-10
Allow Java SDK to report configurable client versions (e.g. Spark/Delta versions) as user-agent

tdas:
**PR Checklist**

**Description of changes**

This PR adds support for custom User-Agent headers in the Java client to enable server-side tracking of client applications and versions.

### Changes Made

1. **Java Client (`ApiClient`)**
   - Added default User-Agent header: `UnityCatalog-Java-Client/{VERSION}`
   - Added `setUserAgent(String)` method to override the entire User-Agent
   - Added `setClientVersion(String... nameVersionPairs)` method to append client application info while preserving the Unity Catalog version
   - Added `getUserAgent()` method to retrieve the current User-Agent value
   - User-Agent is automatically applied to all HTTP requests via the request interceptor

2. **Spark Connector Integration**
   - Updated `ApiClientFactory` to automatically include Spark and Delta versions in User-Agent
   - Example: `UnityCatalog-Java-Client/0.2.0 Spark/4.0.0 Delta/4.0.0`
   - Uses reflection to detect versions without hard dependencies
   - Gracefully handles when Delta is not available

3. **Testing**
   - Added `UserAgentTest.java` with 11 test cases covering all User-Agent functionality
   - Added `ApiClientFactoryTest.java` to verify Spark connector User-Agent configuration

### How This Affects Users

- **Server administrators** can now identify which client versions are accessing Unity Catalog
- **Java client users** can optionally identify their applications using `setClientVersion("MyApp", "1.0.0")`
- **Spark connector users** automatically have Spark/Delta versions included in requests
- **Backward compatible**: No breaking changes, default behavior works without any user action

### Example Usage
```
ApiClient client = new ApiClient();
// Default: UnityCatalog-Java-Client/0.2.0

client.setClientVersion("MyApp", "1.0.0", "MyWrapper", "2.5.1");
// Result: UnityCatalog-Java-Client/0.2.0 MyApp/1.0.0 MyWrapper/2.5.1
```

### DIFF
--- a/clients/java/openapi-templates/ApiClient.mustache
+++ b/clients/java/openapi-templates/ApiClient.mustache
@@ -504,10 +504,11 @@ public class ApiClient {
       throw new IllegalArgumentException("Must provide an even number of arguments (name-version pairs)");
     }
 
-    String baseUserAgent = "UnityCatalog-Java-Client/" + {{invokerPackage}}.VersionUtils.VERSION;
+    // Build user agent starting with base
+    StringBuilder userAgentBuilder = new StringBuilder("UnityCatalog-Java-Client/")
+        .append({{invokerPackage}}.VersionUtils.VERSION);
 
-    // Build application info from all pairs
-    StringBuilder appInfoBuilder = new StringBuilder();
+    // Append application info from all pairs
     for (int i = 0; i < nameVersionPairs.length; i += 2) {
       String name = nameVersionPairs[i];
       String version = nameVersionPairs[i + 1];
@@ -516,23 +517,14 @@ public class ApiClient {
         throw new IllegalArgumentException("Client name at index " + i + " cannot be null or empty");
       }
 
-      if (appInfoBuilder.length() > 0) {
-        appInfoBuilder.append(" ");
-      }
-
-      appInfoBuilder.append(name.trim());
+      userAgentBuilder.append(" ").append(name.trim());
       if (version != null && !version.trim().isEmpty()) {
-        appInfoBuilder.append("/").append(version.trim());
+        userAgentBuilder.append("/").append(version.trim());
       }
     }
 
-    // Set the user agent, replacing any previous client info
-    if (appInfoBuilder.length() > 0) {
-      this.userAgent = baseUserAgent + " " + appInfoBuilder.toString();
-    } else {
-      this.userAgent = baseUserAgent;
-    }
-
+    this.userAgent = userAgentBuilder.toString();
     return this;
   }
 }
+
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
@@ -13,7 +13,7 @@ import org.junit.jupiter.api.Test;
 public class ApiClientFactoryTest {
 
   @Test
-  public void testUserAgentContainsSpark() throws Exception {
+  public void testUserAgentContainsSparkAndDelta() throws Exception {
     ApiClientConf clientConf = new ApiClientConf();
     URI uri = new URI("http://localhost:8080");
     ApiClient client = ApiClientFactory.createApiClient(clientConf, uri, null);
@@ -23,29 +23,14 @@ public class ApiClientFactoryTest {
     // Verify the user agent contains the base Unity Catalog client
     assertThat(userAgent).startsWith("UnityCatalog-Java-Client/");
 
-    // Verify the user agent contains Spark (required)
-    assertThat(userAgent).contains("Spark");
+    // Get expected Spark version
+    String expectedSparkVersion = org.apache.spark.package$.MODULE$.SPARK_VERSION();
+    assertThat(userAgent).contains("Spark/" + expectedSparkVersion);
 
-    // Extract and verify Spark version (version may be empty but Spark must be present)
-    boolean foundSpark = false;
-    String[] parts = userAgent.split(" ");
-    for (String part : parts) {
-      if (part.startsWith("Spark")) {
-        foundSpark = true;
-        break;
-      }
-    }
-    assertThat(foundSpark).as("Spark must be present in User-Agent").isTrue();
-
-    // Delta is optional - verify it if present
-    boolean foundDelta = false;
-    for (String part : parts) {
-      if (part.startsWith("Delta")) {
-        foundDelta = true;
-        break;
-      }
-    }
-    // Note: Delta may or may not be present depending on classpath
+    // Get expected Delta version and verify
+    String expectedDeltaVersion = io.delta.package$.MODULE$.VERSION();
+    assertThat(expectedDeltaVersion).isNotNull();
+    assertThat(userAgent).contains("Delta/" + expectedDeltaVersion);
   }
 
   @Test
@@ -59,12 +44,8 @@ public class ApiClientFactoryTest {
 
     // Verify user agent is set correctly even with authentication
     assertThat(userAgent).startsWith("UnityCatalog-Java-Client/");
-
-    // Verify Spark is present (required)
     assertThat(userAgent).contains("Spark");
-
-    // Delta is optional
-    // assertThat(userAgent).contains("Delta"); // May or may not be present
+    assertThat(userAgent).contains("Delta");
 
     // Verify that the token is not part of the user agent
     assertThat(userAgent).doesNotContain(token);
@@ -94,8 +75,6 @@ public class ApiClientFactoryTest {
       }
     }
     assertThat(hasSparkOrVersion).isTrue();
-
-    // Delta is optional - we don't assert on it
   }
 
   @Test
@@ -110,7 +89,6 @@ public class ApiClientFactoryTest {
     // Verify user agent is still set with at least Spark
     assertThat(client.getUserAgent()).startsWith("UnityCatalog-Java-Client/");
     assertThat(client.getUserAgent()).contains("Spark");
-    // Delta is optional
   }
 
   @Test


clients/java/src/test/java/io/unitycatalog/client/UserAgentTest.java
```diff
@@ -0,0 +1,220 @@
...
+    public void testDefaultUserAgent() {
+        // Create a new ApiClient
+        ApiClient client = new ApiClient();
+        
+        // Verify that a default User-Agent is set
+        String userAgent = client.getUserAgent();
+        assertThat(userAgent).isNotNull();
+        assertThat(userAgent).startsWith("UnityCatalog-Java-Client/");
+        
+        System.out.println("Default User-Agent: " + userAgent);
```
tdas: remove all the println

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
```diff
@@ -13,6 +13,21 @@ public static ApiClient createApiClient(URI url, String token) {
...
+      String deltaVersion = getDeltaVersion();
+
+      if (deltaVersion != null) {
+        apiClient.setClientVersion("Spark", sparkVersion, "Delta", deltaVersion);
+      } else {
+        apiClient.setClientVersion("Spark", sparkVersion);
+      }
+    } catch (Exception e) {
+      // If we can't get versions, continue without them
+      // This shouldn't fail the client creation
```
tdas: set the clientVersion for Spark and Delta even if we could not get the versions. having Spark and Delta with no version is better than not having Spark and Delta at all

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+  public void testUserAgentMayContainDeltaVersion() throws Exception {
+    URI uri = new URI("http://localhost:8080");
+    ApiClient client = ApiClientFactory.createApiClient(uri, null);
+
+    String userAgent = client.getUserAgent();
+    System.out.println("User-Agent: " + userAgent);
+
+    // Delta version is optional (depends on classpath)
+    // If Delta is available, verify the version is present
+    if (userAgent.contains("Delta/")) {
```
tdas: verify the Delta must be present, even if the version is optional. throw error if not Delta not present

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+      String[] parts = userAgent.split(" ");
+      boolean foundDelta = false;
+      for (String part : parts) {
+        if (part.startsWith("Delta/")) {
+          String version = part.substring("Delta/".length());
+          assertThat(version).isNotEmpty();
+          foundDelta = true;
+          System.out.println("Found Delta version: " + version);
```
tdas: remove all prints.

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+    URI uri = new URI("http://localhost:8080");
+    String token = "test-token-12345";
+    ApiClient client = ApiClientFactory.createApiClient(uri, token);
+
+    String userAgent = client.getUserAgent();
+    System.out.println("User-Agent with token: " + userAgent);
+
+    // Verify user agent is set correctly even with authentication
+    assertThat(userAgent).startsWith("UnityCatalog-Java-Client/");
+    assertThat(userAgent).contains("Spark/");
```
tdas: verify it contains both Spark and Delta

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test class for ApiClientFactory to verify User-Agent configuration.
+ */
+public class ApiClientFactoryTest {
+
+  @Test
+  public void testUserAgentContainsSparkVersion() throws Exception {
```
tdas: consolidate the Spark and Delta versions tests into one.

project/JavaClientPostBuild.scala
```diff
@@ -0,0 +1,244 @@
+import sbt._
+import sbt.Keys._
+import java.nio.file.{Files, Path, Paths}
+import scala.util.{Try, Success, Failure}
+
+/**
```
tdas: This is fully LLM generated.

openinx: Hi @tdas ,  since the `ApiClient` already support `setRequestInterceptor` with interceptor to add the header for every request, such as this: https://github.com/unitycatalog/unitycatalog/pull/1179/files#diff-2a5237575234e6479dee617537056cf20c197cfec38febc6e87803abf90177a5R23

Is it possible to use the default `setRequestInterceptor` (generated by OpenAPI) to set the `User-Agent` outside the code generated by OpenAPI ?   If so, then we don't have to hack to change the generated `ApiClient`.

tdas: The problem is that the generated setRequestInterceptor() replaces the interceptor:

```
// Generated code (before modification)public ApiClient setRequestInterceptor(Consumer<HttpRequest.Builder> interceptor) {  
   this.interceptor = interceptor;  // Simple replacement, no chaining  return this;
}
```
So if you set User-Agent via an interceptor, and then later someone calls setRequestInterceptor() for auth (like in ApiClientFactory), it would wipe out the User-Agent.
this generation approach ensures User-Agent is always applied by chaining interceptors.

tdas: also, we want to provide the API as part of the SDK for all connectors to easily set their values. Its not just for UC-Spark. So to generate new APIs in ApiClient you have to generate code.

openinx: Thanks for the explanation.  I have some thoughts about the two questions: 

> Question.1 : So if you set User-Agent via an interceptor, and then later someone calls setRequestInterceptor() for auth (like in ApiClientFactory), it would wipe out the User-Agent.

I think maybe we can customized our own `UCHttpRequestInterceptor` class, so that we can handle both `token` header adding and `User-Agent` header in the same intercepter.  then we always have one `Interceptor`, which won't overrite one by another. Also if we want to add other customized headers for the general purpose (apply to all connectors),  it will be good for use to add header in this unified `UCHttpRequestInterceptor`.   Customizing the request headers via generated code approach may complicate the code, and make it hard to maintain -- as most of the code are concated string. 


> Question.2: also, we want to provide the API as part of the SDK for all connectors to easily set their values. Its not just for UC-Spark. So to generate new APIs in ApiClient you have to generate code.

That's a great point, which I totally agree. we need to apply the headers to all Java SDK. To be more general, not only general headers, we need to apply the whole `ApiClientFactory` and UC API SDK to be general, so that other projects can use the same UC API SDK  common layer ( for example, OSS-Delta will depends on UC API SDK, not the UC-Spark connector). 

Then consdering the general demands, the correct direction may be abstract the UC API SDK into a common module, then every  projects which depends on the  UC SDK would have the `User-Agent`,  also the `retry` policy etc. 


I have a simple change for this,  let me create the PR.

project/JavaClientPostBuild.scala
```diff
@@ -0,0 +1,244 @@
...
+    responseInterceptor = null;
+    asyncResponseInterceptor = null;
+  }"""
+    val constructorInit1WithUserAgent = """    updateBaseUri(getDefaultBaseUri());
+    setRequestInterceptor(null);
+    readTimeout = null;
+    connectTimeout = null;
+    responseInterceptor = null;
+    asyncResponseInterceptor = null;
+    this.userAgent = "UnityCatalog-Java-Client/" + io.unitycatalog.cli.utils.VersionUtils.VERSION;
```
openinx: I think it's better not to make our `uc-spark-connector`  depending on the `io.unitycatalog.cli.utils.VersionUtils` package, which is coming from the `example/cli` module ( https://github.com/unitycatalog/unitycatalog/blob/main/build.sbt#L483) .  Because it will introduce extra unnecessary dependencies and it's possible to conflict with user's runtime jars. 

( Currently,  our `client` module won't be imported in the user's spark job runtime,  after this change, it will be ).

tdas: ohhh this is a fantastic catch.
forget uc-spark. just using uc-client SDK will be a problem!

tdas: i dont know how tests are passing for this.

project/JavaClientPostBuild.scala
tdas: > correct direction may be abstract the UC API SDK into a common module,

what do you mean by this? do you mean that we mode the ApiClientFactory from spark module to directly in the client module such that it compiles along with the generated code? 
i like that direction, but its not a simple PR. We have to make sure to document that users should not use ApiClient directory and rather use the ApiClientFactory + ApiClientConf. In fact, we should think carefully about what is the programmatic API that we want to poeple to use. If our recommended API is not good enough or not power enough to cover all possible usage, the devs are not going to use our new API and go back to ApiClient (which gives maximum power and control), and that will bypass all the things we want the client factory to set by default.

In other worse, if we want to enforce basic stuff like client version, and retries are enabled by default.. then we have to make sure that there is (ideally) 1 method for people to create clients, and everyone uses that way. Putting functionality directly in the ApiClient is definitely one way to achieve that.

project/JavaClientPostBuild.scala
tdas: this was a simple bug in the generation code of VersionUtils

https://github.com/unitycatalog/unitycatalog/blob/main/build.sbt#L223

the package name is wrong. but its actually in the client module... not cli module.

tdas: fixed https://github.com/unitycatalog/unitycatalog/pull/1179/files/#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R236

openinx: > the package name is wrong. but its actually in the client module... not cli module.

Yes, after I take a closer look, it's indeed the `client` module, not the `example/cli` module. then it should be fine. 

I guess maybe we can move the `VersionUtils` from the `client` to a common version, so that not every module need to regenerate the `VersionUtils` ( I just saw the `server` is also generating the `VersionUtils` ).

project/JavaClientPostBuild.scala
openinx: > do you mean that we mode the ApiClientFactory from spark module to directly in the client module such that it compiles along with the generated code?

I mean we move `ApiClientFactory` from the `unity-catalog-spark connector` to the `unitycatalog-common` module, and making it to be general common unitycatalog SDK, not the spark specified connector. 

> In fact, we should think carefully about what is the programmatic API that we want to poeple to use.

Yes,  agreed.  need to carefully design and simplify the API exposing, which requires us to expose min API but provide the flexibility to implement all the functionality. 


I had the PR for injecting the `User-Agent` outside the generated API code currently : https://github.com/unitycatalog/unitycatalog/pull/1242/files .

For the next, I think we need a PR to carefully move the ApiClientFactory & retry related thing into the `unitycatalog-common`,  which shouldn't be a blocker for the comming release.   It will also address the Delta's unified UC client demands.

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
```diff
@@ -7,7 +7,112 @@
...
+    String userAgent = client.getUserAgent();
+
+    // Verify user agent is set correctly even with authentication
+    assertThat(userAgent).startsWith("UnityCatalog-Java-Client/");
+
+    // Verify Spark is present (required)
+    assertThat(userAgent).contains("Spark");
+
+    // Delta is optional
+    // assertThat(userAgent).contains("Delta"); // May or may not be present
```
tdas: uncomment this, its expected Delta will be present here because this project is compiling with Delta. but it may not be present in a real deployment. that is hard to test in a unit test.

clients/java/openapi-templates/ApiClient.mustache
```diff
@@ -0,0 +1,538 @@
+{{>licenseInfo}}
```
openinx, line 1: Is this file a fork version of this original one ? https://github.com/OpenAPITools/openapi-generator/blob/v7.9.0/modules/openapi-generator/src/main/resources/Java/libraries/native/ApiClient.mustache 

Let me see what are the diff...

openinx: Here is the diff

```diff
diff --git a/Users/zheng.hu/Downloads/ApiClient.mustache b/clients/java/openapi-templates/ApiClient.mustache
index a641525a..7b593584 100644
--- a/Users/zheng.hu/Downloads/ApiClient.mustache
+++ b/clients/java/openapi-templates/ApiClient.mustache
@@ -56,6 +56,7 @@ public class ApiClient {
   private Consumer<HttpResponse<String>> asyncResponseInterceptor;
   private Duration readTimeout;
   private Duration connectTimeout;
+  private String userAgent;
 
   public static String valueToString(Object value) {
     if (value == null) {
@@ -158,11 +159,12 @@ public class ApiClient {
     this.builder = createDefaultHttpClientBuilder();
     this.mapper = createDefaultObjectMapper();
     updateBaseUri(getDefaultBaseUri());
-    interceptor = null;
+    setRequestInterceptor(null);
     readTimeout = null;
     connectTimeout = null;
     responseInterceptor = null;
     asyncResponseInterceptor = null;
+    this.userAgent = "UnityCatalog-Java-Client/" + {{invokerPackage}}.VersionUtils.VERSION;
   }
 
   /**
@@ -176,11 +178,12 @@ public class ApiClient {
     this.builder = builder;
     this.mapper = mapper;
     updateBaseUri(baseUri != null ? baseUri : getDefaultBaseUri());
-    interceptor = null;
+    setRequestInterceptor(null);
     readTimeout = null;
     connectTimeout = null;
     responseInterceptor = null;
     asyncResponseInterceptor = null;
+    this.userAgent = "UnityCatalog-Java-Client/" + {{invokerPackage}}.VersionUtils.VERSION;
   }
 
   protected ObjectMapper createDefaultObjectMapper() {
@@ -331,7 +334,14 @@ public class ApiClient {
    * @return This object.
    */
   public ApiClient setRequestInterceptor(Consumer<HttpRequest.Builder> interceptor) {
-    this.interceptor = interceptor;
+    if (interceptor == null) {
+      this.interceptor = builder -> builder.header("User-Agent", this.userAgent);
+    } else {
+      this.interceptor = builder -> {
+        builder.header("User-Agent", this.userAgent);
+        interceptor.accept(builder);
+      };
+    }
     return this;
   }
 
@@ -446,4 +456,83 @@ public class ApiClient {
   public Duration getConnectTimeout() {
     return connectTimeout;
   }
+
+  /**
+   * Set a custom user-agent header for HTTP requests.
+   *
+   * <p>This user-agent will be included in all HTTP requests made by the API client.</p>
+   *
+   * @param userAgent The user-agent string to use. A value of null will reset to the default.
+   * @return This object.
+   */
+  public ApiClient setUserAgent(String userAgent) {
+    this.userAgent = userAgent != null ? userAgent : "UnityCatalog-Java-Client/" + {{invokerPackage}}.VersionUtils.VERSION;
+    return this;
+  }
+
+  /**
+   * Get the current user-agent header value.
+   *
+   * @return The user-agent string.
+   */
+  public String getUserAgent() {
+    return userAgent;
+  }
+
+  /**
+   * Set client application information in the user-agent header.
+   *
+   * <p>This sets client application name and version pairs that are appended to the Unity Catalog client
+   * user-agent, following the standard User-Agent format. This method accepts multiple name-version
+   * pairs and replaces any previously set client information.</p>
+   *
+   * <p>Examples:</p>
+   * <ul>
+   *   <li>{@code setClientVersion("MyApp", "1.0.0")} → "UnityCatalog-Java-Client/0.2.0 MyApp/1.0.0"</li>
+   *   <li>{@code setClientVersion("MyApp", "1.0.0", "MyWrapper", "2.5")} → "UnityCatalog-Java-Client/0.2.0 MyApp/1.0.0 MyWrapper/2.5"</li>
+   *   <li>{@code setClientVersion("MyApp", null)} → "UnityCatalog-Java-Client/0.2.0 MyApp"</li>
+   * </ul>
+   *
+   * @param nameVersionPairs Variable arguments of name-version pairs. Must contain an even number of arguments
+   *                         where each odd argument is a client application name (String) and each even argument
+   *                         is the corresponding version (String or null).
+   * @return This object.
+   * @throws IllegalArgumentException if the number of arguments is odd or if any client name is null/empty.
+   */
+  public ApiClient setClientVersion(String... nameVersionPairs) {
+    if (nameVersionPairs.length % 2 != 0) {
+      throw new IllegalArgumentException("Must provide an even number of arguments (name-version pairs)");
+    }
+
+    String baseUserAgent = "UnityCatalog-Java-Client/" + {{invokerPackage}}.VersionUtils.VERSION;
+
+    // Build application info from all pairs
+    StringBuilder appInfoBuilder = new StringBuilder();
+    for (int i = 0; i < nameVersionPairs.length; i += 2) {
+      String name = nameVersionPairs[i];
+      String version = nameVersionPairs[i + 1];
+
+      if (name == null || name.trim().isEmpty()) {
+        throw new IllegalArgumentException("Client name at index " + i + " cannot be null or empty");
+      }
+
+      if (appInfoBuilder.length() > 0) {
+        appInfoBuilder.append(" ");
+      }
+
+      appInfoBuilder.append(name.trim());
+      if (version != null && !version.trim().isEmpty()) {
+        appInfoBuilder.append("/").append(version.trim());
+      }
+    }
+
+    // Set the user agent, replacing any previous client info
+    if (appInfoBuilder.length() > 0) {
+      this.userAgent = baseUserAgent + " " + appInfoBuilder.toString();
+    } else {
+      this.userAgent = baseUserAgent;
+    }
+
+    return this;
+  }
 }
```

openinx:
```diff
+    String baseUserAgent = "UnityCatalog-Java-Client/" + {{invokerPackage}}.VersionUtils.VERSION;
+
+    // Build application info from all pairs
+    StringBuilder appInfoBuilder = new StringBuilder();
```

Here  we can pass the `baseUserAgent` directly  to the `StringBuilder`'s constructor, then we don't need to concat the string in the following part:

```diff
+    // Set the user agent, replacing any previous client info
+    if (appInfoBuilder.length() > 0) {
+      this.userAgent = baseUserAgent + " " + appInfoBuilder.toString();
+    } else {
+      this.userAgent = baseUserAgent;
+    }
```

build.sbt
```diff
@@ -224,9 +226,9 @@ lazy val client = (project in file("target/clients/java"))
     },
     // Add VersionInfo in the same way like in server
     Compile / sourceGenerators += Def.task {
-      val file = (Compile / sourceManaged).value / "io" / "unitycatalog" / "cli" / "utils" / "VersionUtils.java"
+      val file = (Compile / sourceManaged).value / "io" / "unitycatalog" / "client" / "VersionUtils.java"
       IO.write(file,
-        s"""package io.unitycatalog.cli.utils;
+        s"""package io.unitycatalog.client;
```
openinx, line 231: Nice catch !

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
openinx: Maybe we can remove it directly.

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
```diff
@@ -7,7 +7,112 @@
...
+
+    // Delta is optional - verify it if present
+    boolean foundDelta = false;
+    for (String part : parts) {
+      if (part.startsWith("Delta")) {
+        foundDelta = true;
+        break;
+      }
+    }
+    // Note: Delta may or may not be present depending on classpath
```
openinx: nit: Coud you please make this code more clean ?

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
```diff
@@ -7,7 +7,112 @@
...
+  }
+
+  @Test
+  public void testUserAgentWithToken() throws Exception {
```
openinx, line 37: Actually,  the token (is present or absent ) does not affect the `UserAgent` setting , right ?  Then I think we can just remove this redudant unit test.

openinx: Oh, it will.  since we have this change in the mustable file:

```diff
@@ -331,7 +334,14 @@ public class ApiClient {
    * @return This object.
    */
   public ApiClient setRequestInterceptor(Consumer<HttpRequest.Builder> interceptor) {
-    this.interceptor = interceptor;
+    if (interceptor == null) {
+      this.interceptor = builder -> builder.header("User-Agent", this.userAgent);
+    } else {
+      this.interceptor = builder -> {
+        builder.header("User-Agent", this.userAgent);
+        interceptor.accept(builder);
+      };
+    }
     return this;
   }
```

openinx: Then let's just ignore the above comment.

tdas: exactly.

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
tdas: i want to keep this test because this is the only way to verify that uc-spark does put in delta version when it is present.

tdas:
Created backport PR for `branch-0.3`:
- #1254 with remaining conflicts!

Please cherry-pick the changes locally and resolve any conflicts.
```bash
git fetch origin backport-1179-to-branch-0.3
git worktree add --checkout .worktree/backport-1179-to-branch-0.3 backport-1179-to-branch-0.3
cd .worktree/backport-1179-to-branch-0.3
git reset --hard HEAD^
git cherry-pick -x 907d81329cef4ebfea07941d764f97d4b41acbe0
git push --force-with-lease
```

PR #1197, opened 2025-11-12
When creating a managed table, standardize URL before looking for staging table

yili-db:
**PR Checklist**

**Description of changes**
When creating a managed table, standardize URL before looking for staging table.
This is needed because certain clients like Delta Spark may send local path URLs in non-standard form (e.g. file:/tmp) despite that UC returned a staging location in standard form (e.g. file:///tmp). UC needs to tolerate it.

**Related issue**

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -186,6 +186,8 @@ public class TableRepository {
           String schemaName = createTable.getSchemaName();
           UUID schemaId =
               repositories.getSchemaRepository().getSchemaId(session, catalogName, schemaName);
+          String storageLocation =
+              FileOperations.toStandardizedURIString(createTable.getStorageLocation());
 
           // Check if table already exists
           TableInfoDAO existingTable =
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -238,8 +240,7 @@ public TableInfo createTable(CreateTable createTable) {
                   .createdBy(callerId)
                   .updatedAt(createTime)
                   .updatedBy(callerId)
-                  .storageLocation(
-                      FileOperations.toStandardizedURIString(createTable.getStorageLocation()))
+                  .storageLocation(storageLocation)
```
yili-db, line 243: This line is still doing exactly the same as before.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -210,7 +212,7 @@ public TableInfo createTable(CreateTable createTable) {
             StagingTableDAO stagingTableDAO =
                 repositories
                     .getStagingTableRepository()
-                    .commitStagingTable(session, callerId, createTable.getStorageLocation());
+                    .commitStagingTable(session, callerId, storageLocation);
```
yili-db, line 215: This is the actual change. The location is now standardized before searching for matching staging table

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -241,6 +241,9 @@ public static void assertValidLocation(String location) {
    * </pre>
    */
   public static String toStandardizedURIString(String inputPath) {
+    if (inputPath == null) {
```
yili-db, line 244: This null checking isn't needed today. But it's here just in case in the future we'll have tables without storage path for example VIEW.

Issue #1160, opened 2025-11-05
generateTemporaryPathCredential does not do correct authorization


yili-db:
The current implementation of generateTemporaryPathCredential does not have the complete implementation of authorization. It requires the caller to be the **owner** of **metastore**. This often would cause a failure when creating a table.

```
  @Post("")
  @AuthorizeExpression("#authorize(#principal, #metastore, OWNER)")
  @AuthorizeKey(METASTORE)
  public HttpResponse generateTemporaryPathCredential(
      GenerateTemporaryPathCredential generateTemporaryPathCredential) {

```
The error message usually says:
```
: io.unitycatalog.client.ApiException: generateTemporaryPathCredentials call failed with: 403 - {"error_code":"PERMISSION_DENIED","details":[{"reason":"PERMISSION_DENIED","metadata":{},"@type":"google.rpc.ErrorInfo"}],"stack_trace":null,"message":"Access denied."}
at io.unitycatalog.client.api.TemporaryCredentialsApi.getApiException(TemporaryCredentialsApi.java:78)

```

**To Reproduce**

These issues have been reported on how to reproduce the problem:
https://github.com/unitycatalog/unitycatalog/issues/1088 https://github.com/unitycatalog/unitycatalog/issues/1034 

**Expected behavior**

A correct authorization model should be implemented for this RPC that shouldn't require ownership of metastore.

**System [please complete the following information]:**

-   OS: e.g. [Ubuntu 18.04]

**Additional context**

yili-db:
cc @tdas

1b5d:
take

Issue #1088, opened 2025-09-29
Table creation in spark fails when auth is enabled


Kavan-M:
**Describe the bug**

When running unitycatalog with auth enabled, you are not capable of creating a table via Spark. Using the same auth token, creating a table from within the CLI works as documented in the docs. Reads and writes to already existing tables work via Spark.

1. Set up auth service
2. Create user and grant the user permissions (USE CATALOG for catalog, USE SCHEMA, MODIFY, CREATE TABLE for schema)
3. Run unitycatalog with auth enabled.
4. Get auth token
5. Test table creation works via CLI (bin/uc table create \
  --full_name my_catalog.my_schema.my_table \
  --columns "id INT, name STRING" \
  --storage_location "/path/to/storage")
6. Start spark session connected to unity. Use auth token from before as token
7. within spark session, run spark.sql("""CREATE TABLE demo.mytable (id INT, desc STRING)USING deltaLOCATION '<LOCATION>'
""")

**Expected behavior**

Both spark and CLI table creation should succeed

**System [please complete the following information]:**

-   OS: MacOS Sequoia 15.7
-   UC: commit b2d072e
-   Java version 17
-   Spark version 3.5.3

**Additional context**

-   --packages io.delta:delta-spark_2.12:3.2.1,io.unitycatalog:unitycatalog-spark_2.12:0.2.0

error message:

py4j.protocol.Py4JJavaError: An error occurred while calling o47.sql.
: io.unitycatalog.client.ApiException: generateTemporaryPathCredentials call failed with: 403 - {"error_code":"PERMISSION_DENIED","details":[{"reason":"PERMISSION_DENIED","metadata":{},"@type":"google.rpc.ErrorInfo"}],"stack_trace":null,"message":"Access denied."}
	at io.unitycatalog.client.api.TemporaryCredentialsApi.getApiException(TemporaryCredentialsApi.java:78)

openinx:
Seems the error message is: `Access denied`.   Did you got the correct credentials from the GCP ?  @Kavan-M

Kavan-M:
> Seems the error message is: `Access denied`. Did you got the correct credentials from the GCP ? [@Kavan-M](https://github.com/Kavan-M)

[...]

Issue #1034, opened 2025-06-09
PySpark can't crete delta table with s3 location due to the permission of TemporaryPath


yorkdick:
I started the UC by docker and enable the authorization.
I create a new user "uctest" for spark, I add the "CREATE CATALOG" to this user and crate a catalog "myspace" using this user.
I can create schema by pyspark.

![Image](https://github.com/user-attachments/assets/d0472a96-a344-4e0e-a6a6-8093a1a20f74)

![Image](https://github.com/user-attachments/assets/aa678894-513a-41c3-9a1b-9ddec55f108e)

But I can't create table by pyspark, the error message says "PERMISSION_DENIED" for generateTemporaryPathCredentials.

![Image](https://github.com/user-attachments/assets/e535ff1d-0a7a-48d0-b1d0-a4420aeba838)
`Py4JJavaError: An error occurred while calling o48.sql.
: io.unitycatalog.client.ApiException: generateTemporaryPathCredentials call failed with: 403 - {"error_code":"PERMISSION_DENIED","details":[{"reason":"PERMISSION_DENIED","metadata":{},"@type":"google.rpc.ErrorInfo"}],"stack_trace":null,"message":"Access denied."}
	at io.unitycatalog.client.api.TemporaryCredentialsApi.getApiException(TemporaryCredentialsApi.java:78)
	at io.unitycatalog.client.api.TemporaryCredentialsApi.generateTemporaryPathCredentialsWithHttpInfo(TemporaryCredentialsApi.java:192)
	at io.unitycatalog.client.api.TemporaryCredentialsApi.generateTemporaryPathCredentials(TemporaryCredentialsApi.java:170)
	at io.unitycatalog.spark.UCSingleCatalog.createTable(UCSingleCatalog.scala:106)`

I checked the source code, only the owner of this metastore has the permission.

![Image](https://github.com/user-attachments/assets/ef2e2832-1826-412c-93a6-04f5e3f29fa0)

Does it mean that only metastore  owner  can ceate delta table using spark ? 
The metastore  owner has too many permissions. If I want to allow the catalog or schema owner to create a table, what should I do?

ray-harrison:
I'd like some insight here too, please, before we go and modify code in our particular branch. Is this something you can help point us in the right direction on @dennyglee? Was this done on purpose for reasons I don't understand (entirely possible :-)). It seems to me that if I have create table privs on a schema, I should be able to do this in Spark as I can at the command line. The CLI variation works.

[...]

PR #1203, opened 2025-11-13
Allow temporary path credentials for users granted EXTERNAL_USE_LOCATION

1b5d:
**PR Checklist**

**Description of changes**

This PR refactors the `TemporaryPathCredentialsService` to allow temporary path credentials for users granted the `EXTERNAL_USE_LOCATION` permission

Since only `OWNER` of the metastore was allowed to create temporary path credentials, it wasn't possible to create a table with any other user, even though other user types are allowed to create schema for instance see issues https://github.com/unitycatalog/unitycatalog/issues/1160, https://github.com/unitycatalog/unitycatalog/issues/1034 and https://github.com/unitycatalog/unitycatalog/issues/1088. This PR solves this limitation but implementing the new permission

Also added supported adding the new permission via the permissions API

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/service/TemporaryPathCredentialsServiceTest.java
+++ b/server/src/test/java/io/unitycatalog/server/service/TemporaryPathCredentialsServiceTest.java
@@ -28,7 +28,6 @@ import io.unitycatalog.server.sdk.storagecredential.SdkCredentialOperations;
 import io.unitycatalog.server.service.credential.CloudCredentialVendor;
 import io.unitycatalog.server.service.credential.CredentialContext;
 import io.unitycatalog.server.service.credential.aws.AwsCredentialVendor;
-import io.unitycatalog.server.utils.ServerProperties;
 import io.unitycatalog.server.utils.TestUtils;
 import java.util.stream.Stream;
 import lombok.SneakyThrows;
@@ -127,9 +126,7 @@ public class TemporaryPathCredentialsServiceTest extends SdkAccessControlBaseCRU
   protected void setUpCredentialOperations() {
     setupAwsCredentials();
 
-    cloudCredentialVendor =
-        new CloudCredentialVendor(
-            awsCredentialVendor, null, null, new ServerProperties(serverProperties));
+    cloudCredentialVendor = new CloudCredentialVendor(awsCredentialVendor, null, null);
   }
 
   private void setupAwsCredentials() {


server/src/main/java/io/unitycatalog/server/service/TemporaryPathCredentialsService.java
```diff
@@ -1,36 +1,58 @@
...
+      CloudCredentialVendor cloudCredentialVendor,
+      Repositories repositories) {
     this.cloudCredentialVendor = cloudCredentialVendor;
+    this.userRepository = repositories.getUserRepository();
+    this.keyMapper = new KeyMapper(repositories);
+    this.evaluator = new UnityAccessEvaluator(authorizer);
   }
 
   @Post("")
-  @AuthorizeExpression("#authorize(#principal, #metastore, OWNER)")
```
openinx: Why remove this ?

server/src/main/java/io/unitycatalog/server/service/TemporaryPathCredentialsService.java
```diff
@@ -44,4 +66,59 @@ private Set<CredentialContext.Privilege> pathOperationToPrivileges(PathOperation
...
+      return;
+    }
+
+    evaluateArbitraryPathOperation();
+  }
+
+  private void evaluateCreateTableForSchema(PathComponents pathComponents) {
+    // CREATE_TABLE requires either OWNER or USE_CATALOG on the catalog AND
+    // either OWNER or USE_SCHEMA on the schema.
+    String authorizeExpression =
```
openinx: Because we move the check from [this](https://github.com/unitycatalog/unitycatalog/pull/1203/files#r2524236759) to here ?

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -95,6 +96,25 @@ private UnityCatalogServer(UnityCatalogServer.Builder unityCatalogServerBuilder)
     this.server = initializeServer(unityCatalogServerBuilder);
   }
 
+  public static void main(String[] args) {
+    OptionParser options = new OptionParser();
+    options.parse(args);
+    // Start Unity Catalog server
+    UnityCatalogServer unityCatalogServer =
+        UnityCatalogServer.builder().port(options.getPort() + 1).build();
```
openinx, line 104: Why use the `options.getPort() + 1 `  rather than the `options.getPort()`?

server/src/main/java/io/unitycatalog/server/service/TemporaryPathCredentialsService.java
1b5d: Yes exactly, because we make the expression conditional on the content of the path we're trying to authorize, then we revert back to the original exp when no match

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
1b5d: I didn't really change this, I merely just formatted the file, see https://github.com/unitycatalog/unitycatalog/pull/1203/files#diff-fcd3330038de1005eca051c7dfb37a467d727c2ac4625ae62759fc81fc04b339L338

It's a good question though!

server/src/main/java/io/unitycatalog/server/service/TemporaryPathCredentialsService.java
1b5d: answered in https://github.com/unitycatalog/unitycatalog/pull/1203#discussion_r2524353672

yili-db:
@1b5d FYI we have built a new managed table creation feature. The table path is allocated by server and in form of `.../tables/<table-id>/...`
https://github.com/unitycatalog/unitycatalog/blob/b33279336000d0d315e667d3825155183c04b2f4/examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java#L120

The solution in this PR seems to enforce the new EXTERNAL table creation must be done in a path like this:
`.../catalog_name/schema_name/...`:
1. It's less useful as it forces users to follow a path convention that may or may not work for them
2. It uses name instead of id. But schema and catalog can be renamed.

I think ultimately the solution should authorize against ExternalLocation (see `ExternalLocationRepository.java`). Besides catalog and schema permission check, we can build a new external location permission check on a new permission `EXTERNAL_USE_LOCATION`. The flow would be like:
1. search for a matching external location by the URL path
2. check if the user has `EXTERNAL_USE_LOCATION` permission on that external location

The idea is from here: https://docs.databricks.com/api/workspace/temporarypathcredentials/generatetemporarypathcredentials

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
yili-db: From the code it seems that the server listens on 8081 while this `URLTranscoderVerticle` listens on 8080 and works as a proxy to 8081 with some URL manipulation.
This `URLTranscoderVerticle` has been there since the first commit. I don't know why it is needed. It seems all it cares about is the 0x1f special character.

1b5d:
> @1b5d FYI we have built a new managed table creation feature. The table path is allocated by server and in form of `.../tables/<table-id>/...`
> 
> https://github.com/unitycatalog/unitycatalog/blob/b33279336000d0d315e667d3825155183c04b2f4/examples/cli/src/main/java/io/unitycatalog/cli/TableCli.java#L120
> 
> The solution in this PR seems to enforce the new EXTERNAL table creation must be done in a path like this: `.../catalog_name/schema_name/...`:
> 
> 1. It's less useful as it forces users to follow a path convention that may or may not work for them
> 2. It uses name instead of id. But schema and catalog can be renamed.
> 
> I think ultimately the solution should authorize against ExternalLocation (see `ExternalLocationRepository.java`). Besides catalog and schema permission check, we can build a new external location permission check on a new permission `EXTERNAL_USE_LOCATION`. The flow would be like:
> 
> 1. search for a matching external location by the URL path
> 2. check if the user has `EXTERNAL_USE_LOCATION` permission on that external location
> 
> The idea is from here: https://docs.databricks.com/api/workspace/temporarypathcredentials/generatetemporarypathcredentials

That's interesting, thanks for the quick feedback! we have implemented this PR largely due to the lack of managed table previously, we basically worked around that by constructing an external location for every table we need to create, however we were blocked by the restriction to the metastore owner, however managed tables would actually be a much better solution if it works already!

1b5d:
@yili-db I have a few questions it would be great if you can help me adapt this implementation to match your feedback:
- How can we still do catalog and schema permission check by using the URL then ? in `ExternalLocationDAO` there is no catalog / schema reference
- Maybe I'm missing sth obvious, but it's not clear to me how this works when I want to create a table and an external location is not there yet?
- When matching a URL against an external location, should it be a partial match? in that case I assume it should match the longest possible location yeah? if so, wouldn't that be a performance concern?

yili-db:
> @yili-db I have a few questions it would be great if you can help me adapt this implementation to match your feedback:
> 
> * How can we still do catalog and schema permission check by using the URL then ? in `ExternalLocationDAO` there is no catalog / schema reference

External location does not belong to any schema or catalog. It's just a registered storage location and can be used by any catalog or schema to create table. If we borrow the concept of Databricks external location, then even managed tables should eventually be created on the storage location specified by external location.
In the context of external tables, any external table in any schema can live under any external location since it's **external**.

> * Maybe I'm missing sth obvious, but it's not clear to me how this works when I want to create a table and an external location is not there yet?

In this case we can make it fail due to lack of external location.
We are not in the state that every table is created with a registered external location yet. But my idea is that if we want to make the temp path credential work with proper authorization, the Databricks model is a good reference to move forward to.

> * When matching a URL against an external location, should it be a partial match? in that case I assume it should match the longest possible location yeah? if so, wouldn't that be a performance concern?

A partial matching is needed. An external location of ("s3://bucket/ucroot") covers all paths start with it. So creating a table under ("s3://bucket/ucroot/mytable") should authorize against it. This needs to be done by searching external location with URL prefixes of all levels.
Ideally we should also make sure that external locations (and tables) do not have overlapping path.

Overall this isn't something trivial to implement. I'll let @tdas to comment on how we should proceed.

1b5d:
@yili-db @openinx @tdas I have attempted to implement the EXTERNAL_USE_LOCATION permission in this PR based on the feedback I received from @yili-db

1b5d:
@yili-db btw I tried to create a managed table (from Spark 4.0) after setting `server.managed-table.enabled=true` and `storage-root.tables`, I get a `Table not found` error no matter what I do.

Then I noticed that nothing changed in UCSingleCatalog, is managed table still needs to be implemented in the connector?

yili-db:
> @yili-db btw I tried to create a managed table (from Spark 4.0) after setting `server.managed-table.enabled=true` and `storage-root.tables`, I get a `Table not found` error no matter what I do.
> 
> Then I noticed that nothing changed in UCSingleCatalog, is managed table still needs to be implemented in the connector?

I am still implementing in UCSingleCatalog. For now you can use the command line to create one if you want to try it:
```
bin/uc table create --full_name unity.default.my_managed_table --columns "id INT, name STRING" --format DELTA --table_type MANAGED
```

PR #1193, opened 2025-11-11
Add local Abfs credential renewal integration test

openinx:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [x] If there is a related issue, make sure it is linked to this PR.
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

The related issue is : https://github.com/unitycatalog/unitycatalog/issues/1175,   this PR achieved  the similar local integration tests that we've did in aws s3. https://github.com/unitycatalog/unitycatalog/pull/1123

PR #1204, opened 2025-11-13
Add GCS credential renewal integration test (local)

TimothyW553:
**PR Checklist**

**Description of changes**

Add local integration tests for GCS credential renewal.

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredentialRenewalTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredentialRenewalTest.java
@@ -324,30 +324,6 @@ public class GcsCredentialRenewalTest extends BaseCRUDTest {
   }
 
   /**
-   * Returns a fixed, non-expiring access token for simple integration tests. This allows validating
-   * plumbing and downscoping without exercising refresh logic or service-account IO.
-   */
-  public static class StaticTestingCredentialsGenerator
-      implements io.unitycatalog.server.service.credential.gcp.GcpCredentialVendor
-          .GcpCredentialsGenerator {
-    private final AccessToken staticToken;
-
-    public StaticTestingCredentialsGenerator() {
-      this.staticToken =
-          AccessToken.newBuilder()
-              .setTokenValue("testing://static-token")
-              .setExpirationTime(Date.from(Instant.ofEpochMilli(Long.MAX_VALUE)))
-              .build();
-    }
-
-    @Override
-    public AccessToken generate(
-        io.unitycatalog.server.service.credential.CredentialContext context) {
-      return staticToken;
-    }
-  }
-
-  /**
    * Issues short-lived tokens whose expiration aligns with the shared manual clock. Each token
    * encodes the active renewal window and bucket so tests can assert rotation frequency. This
    * generator is dynamically loaded by the Unity Catalog server and serves credential generation
--- a/server/src/test/java/io/unitycatalog/server/service/credential/CloudCredentialVendorTest.java
+++ b/server/src/test/java/io/unitycatalog/server/service/credential/CloudCredentialVendorTest.java
@@ -123,7 +123,7 @@ public class CloudCredentialVendorTest {
                     .bucketPath("gs://uctest")
                     .jsonKeyFilePath("")
                     .credentialsGenerator(
-                        "io.unitycatalog.spark.auth.GcsCredentialRenewalTest$StaticTestingCredentialsGenerator")
+                        "io.unitycatalog.server.service.credential.gcp.StaticTestingCredentialsGenerator")
                     .build()));
     GcpCredentialVendor gcpCredentialVendor = new GcpCredentialVendor(serverProperties);
     credentialsOperations = new CloudCredentialVendor(null, null, gcpCredentialVendor);
--- a/server/src/test/java/io/unitycatalog/server/service/credential/gcp/StaticTestingCredentialsGenerator.java
+++ b/server/src/test/java/io/unitycatalog/server/service/credential/gcp/StaticTestingCredentialsGenerator.java
@@ -1,0 +1,30 @@
+package io.unitycatalog.server.service.credential.gcp;
+
+import com.google.auth.oauth2.AccessToken;
+import io.unitycatalog.server.service.credential.CredentialContext;
+import java.time.Instant;
+import java.util.Date;
+
+/**
+ * Test-only generator that issues a non-expiring static token.
+ *
+ * <p>Used by {@link io.unitycatalog.server.service.credential.CloudCredentialVendorTest} to verify
+ * basic GCS credential vending without requiring real GCP service accounts.
+ */
+public class StaticTestingCredentialsGenerator
+    implements GcpCredentialVendor.GcpCredentialsGenerator {
+  private final AccessToken staticToken;
+
+  public StaticTestingCredentialsGenerator() {
+    this.staticToken =
+        AccessToken.newBuilder()
+            .setTokenValue("testing://static-token")
+            .setExpirationTime(Date.from(Instant.ofEpochMilli(Long.MAX_VALUE)))
+            .build();
+  }
+
+  @Override
+  public AccessToken generate(CredentialContext context) {
+    return staticToken;
+  }
+}


connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredentialRenewalTest.java
```diff
@@ -0,0 +1,275 @@
...
+  private SparkSession createSparkSession() {
+    String testCatalogKey = String.format("spark.sql.catalog.%s", CATALOG_NAME);
+    return SparkSession.builder()
+        .appName("test-gcs-credential-renewal")
+        .master("local[1]")
+        .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
+        .config(
+            "spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog")
+        .config("spark.hadoop." + UCHadoopConf.UC_TEST_CLOCK_NAME, CLOCK_NAME)
+        .config("spark.hadoop." + UCHadoopConf.UC_RENEWAL_LEAD_TIME_KEY, 0L)
```
timothyw553, line 103: inject manual clock and turns on credential renewal.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredentialRenewalTest.java
```diff
@@ -0,0 +1,275 @@
...
+import org.apache.spark.util.SerializableConfiguration;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.sparkproject.guava.collect.ImmutableList;
+import org.sparkproject.guava.collect.Iterators;
+
+public class GcsCredentialRenewalTest extends BaseCRUDTest {
```
timothyw553, line 48: Following the same template for AWS: https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/test/java/io/unitycatalog/spark/auth/AwsCredentialRenewalTest.java

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredentialRenewalTest.java
```diff
@@ -0,0 +1,374 @@
...
+    }
+  }
+
+  /**
+   * Issues short-lived tokens whose expiration aligns with the shared manual clock. Each token
+   * encodes the active renewal window and bucket so tests can assert rotation frequency. This
+   * generator is dynamically loaded by the Unity Catalog server and serves credential generation
+   * requests from client REST API calls.
+   */
+  public static class TestingRenewalCredentialsGenerator
```
timothyw553, line 332: Here we create the generators in connectors/spark instead of the server side because then we can manually control the clock.

server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java
```diff
@@ -15,52 +15,74 @@
...
-    String serviceAccountKeyJsonFilePath =
-        gcsConfigurations.get(credentialContext.getStorageBase());
-
-    GoogleCredentials creds;
-    if (serviceAccountKeyJsonFilePath != null && !serviceAccountKeyJsonFilePath.isEmpty()) {
-      if (serviceAccountKeyJsonFilePath.startsWith("testing://")) {
-        // allow pass-through of a dummy value for integration testing
-        return AccessToken.newBuilder()
-            .setTokenValue(serviceAccountKeyJsonFilePath)
-            .setExpirationTime(Date.from(Instant.ofEpochMilli(253370790000000L)))
```
timothyw553, line 48: We moved this out to [StaticTestingCredentialsGenerator.java](https://github.com/unitycatalog/unitycatalog/pull/1204/files#diff-04f2e6221f04dccab970deb6948dff8d2343e325da8f16655a2365adda3779db) because we want to decouple GcpCredentialVendor from just the static test as we added the renewal test.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredentialRenewalTest.java
```diff
@@ -0,0 +1,374 @@
...
+   * Configure the server to use the test-specific credential generator that issues time-bucketed
+   * tokens aligned with the manual clock. This matches the AWS renewal test approach.
+   */
+  @Override
+  protected void setUpProperties() {
+    super.setUpProperties();
+    serverProperties.put("gcs.bucketPath.0", "gs://" + BUCKET_NAME);
+    serverProperties.put("gcs.jsonKeyFilePath.0", "");
+    serverProperties.put(
+        "gcs.credentialsGenerator.0", TestingRenewalCredentialsGenerator.class.getName());
```
timothyw553, line 86: Here we configure gcsstorageconfig to point to the testingrenewealcredentialsgenerator

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredentialRenewalTest.java
timothyw553: This renewal credentials generator is used when the storagebase is `testing-renew://`. It is loaded dynamically loaded by `GcpCredentialsGenerator createGenerator(...)` as it sees 

```
storageConfig.getCredentialsGenerator() = "io.unitycatalog.spark.auth.GcsCredentialRenewalTest$TestingRenewalCredentialsGenerator"
```

This was configured when we set the properties upon server startup for testing in [`GcsCredentialRenewalTest#setUpProperties`](https://github.com/unitycatalog/unitycatalog/pull/1204/files#r2525205739)

So the flow is:
```
server startup
  |     ↳ configure credentialsGenerator in setUpProperties()
  |         ↳ point to TestingRenewalCredentialsGenerator class name
  v 
first table read
  |     ↳ getAccessToken()
  |         ↳ accessCredentials() → credential is null (first time)
  |         ↳ renewCredential()
  |             ↳ vendGcpToken (UC server REST call)
  |                 ↳ createGenerator (ONLY HAPPENS ONCE)
  |                |    ↳ Class.forName(...TestingRenewalCredentialsGenerator).newInstance()
  |                |    ↳ CACHE generator in credentialGenerators map
  |                 ↳ generator.generate(context)
  |                     ↳ TestingRenewalCredentialsGenerator.generate()
  |                         ↳ returns AccessToken with expiry = 30000ms
  v
advance time by expiration delta (30s)
  |
  v
next table read
        ↳ getAccessToken() called
            ↳ accessCredentials() → check readyToRenew(clock, 0)
           |    ↳ 30000 <= clock.now() (60000) + 0 ? → TRUE (expired!)
            ↳ renewCredential()
                ↳ vendGcpToken (UC server REST call)
                    ↳ credentialGenerators.get("gs://test-bucket")
                   |    ↳ REUSE cached generator (no createGenerator call!)
                    ↳ generator.generate(context)
                        ↳ TestingRenewalCredentialsGenerator.generate()
                            ↳ returns NEW AccessToken with expiry = 60000ms
```

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredentialRenewalTest.java
openinx: Hi @TimothyW553 ,  we are not planning to copy another `CredentialRenewalTest` for all the three cloud vendor.  we need to abstract the credential renewal IT test into a `BasedCredRenewalITTest`, and each cloud storage will just extend the `BaseCredRenewalITTest`, and provide the method implementation,  that should be good since we don't want to modify each test if there is any change for those it tests.

openinx: I have this PR and get approval from @tdas , but still not merged yet.  Maybe you can try to port this PR to your local, and add the extra changes on top of the PR. and then try the PR change again. I guess it will be much simple to add the `GcpCredITTest`, since we only need to add few implementations. 

(Once the PR get merged, you may try rebase to the `main`  branch ).

https://github.com/unitycatalog/unitycatalog/pull/1193/files#diff-6de4f05da3644b318732958da80a4df47b8dcac8792ec9423f2a581b8f42831fR58

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredentialRenewalTest.java
openinx: Yes, I think your understanding is correct. That's why we can use the mock clock to successfully test the credentials renwal, which was hard at the first glance when we started thinking about the integration test.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -120,16 +121,23 @@ public Map<String, S3StorageConfig> getS3Configurations() {
...
     int i = 0;
     while (true) {
       String bucketPath = getProperty("gcs.bucketPath." + i);
       String jsonKeyFilePath = getProperty("gcs.jsonKeyFilePath." + i);
       if (bucketPath == null || jsonKeyFilePath == null) {
         break;
       }
-      gcsConfigMap.put(bucketPath, jsonKeyFilePath);
+      String generatorClass = getProperty("gcs.credentialsGenerator." + i);
+      gcsConfigMap.put(
```
openinx, line 134: The original design is hard to maintan multipe gcs storage config with multiple propertes,  so I think we change to use `Map<String, GcsStorageConfig>` should be the correct approach.

server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java
```diff
@@ -15,52 +15,74 @@
...
-  private final Map<String, String> gcsConfigurations;
+  private final Map<String, GcsStorageConfig> gcsConfigurations;
+  private final Map<String, GcpCredentialsGenerator> credentialGenerators =
+      new ConcurrentHashMap<>();
+
+  /**
+   * Pluggable contract for generating GCS OAuth tokens. This is used to allow for custom credential
+   * generation for local integration tests.
+   */
+  public interface GcpCredentialsGenerator {
```
openinx, line 44: Why not just move the `GcpCredentialsGenerator` interface into a separate java file ?   Then we can make all the generator implementations inside that class, which actually make the code more clear?

PR #1209, opened 2025-11-14
Add GCS credential renewal integration test (local)

TimothyW553:
**PR Checklist**

**Description of changes**
This branch is currently based on @openinx's PR https://github.com/unitycatalog/unitycatalog/pull/1193 which creates an abstraction for integration tests. 

**TODO:** update this branch once his PR is merged to main. The delta for this PR can be found here: https://github.com/openinx/unitycatalog/pull/1/files

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -117,7 +117,6 @@ useCoursier := true
 
 // Configure resolvers
 resolvers ++= Seq(
-  "Sonatype OSS Releases" at "https://oss.sonatype.org/content/repositories/releases/",
   "Maven Central" at "https://repo1.maven.org/maven2/",
 )
 
--- a/project/ReleaseSettings.scala
+++ b/project/ReleaseSettings.scala
@@ -53,17 +53,17 @@ object ReleaseSettings {
     pgpPassphrase := sys.env.get("PGP_PASSPHRASE").map(_.toArray),
     sonatypeProfileName := "io.unitycatalog",
     credentials += Credentials(
-      "Sonatype Nexus Repository Manager",
-      "s01.oss.sonatype.org",
+      "OSSRH Staging API Service",
+      "ossrh-staging-api.central.sonatype.com",
       sys.env.getOrElse("SONATYPE_USERNAME", ""),
       sys.env.getOrElse("SONATYPE_PASSWORD", "")
     ),
     publishTo := {
-      val nexus = "https://s01.oss.sonatype.org/"
+      val ossrhBase = "https://ossrh-staging-api.central.sonatype.com/"
       if (isSnapshot.value) {
-        Some("snapshots" at nexus + "content/repositories/snapshots")
+        Some("snapshots" at ossrhBase + "content/repositories/snapshots")
       } else {
-        Some("releases" at nexus + "service/local/staging/deploy/maven2")
+        Some("releases" at ossrhBase + "service/local/staging/deploy/maven2")
       }
     },
     licenses += ("Apache-2.0", url("http://www.apache.org/licenses/LICENSE-2.0")),
@@ -100,7 +100,7 @@ object ReleaseSettings {
   lazy val rootReleaseSettings = Seq(
     publishArtifact := false,
     publish / skip := true,
-    publishTo := Some("snapshots" at "https://oss.sonatype.org/content/repositories/snapshots"),
+    publishTo := Some("snapshots" at "https://ossrh-staging-api.central.sonatype.com/content/repositories/snapshots"),
     releaseCrossBuild := false,
     crossScalaVersions := Nil,
     releaseProcess := Seq[ReleaseStep](
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -22,7 +22,7 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "2.2.0")
 
 addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.3")
 
-addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "2.3")
+addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "3.11.3")
 
 addSbtPlugin("net.aichler" % "sbt-jupiter-interface" % "0.11.1")
 
--- a/server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
@@ -16,11 +16,8 @@ import java.util.UUID;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.query.Query;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class StagingTableRepository {
-  private static final Logger LOGGER = LoggerFactory.getLogger(StagingTableRepository.class);
   private final SessionFactory sessionFactory;
   private final Repositories repositories;
   private final ServerProperties serverProperties;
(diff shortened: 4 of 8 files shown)


timothyw553:
Some maven download issues... its a bit flaky sometimes -- closing and reopening

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredRenewITTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+public class GcsCredRenewITTest extends BaseCredRenewITTest {
+  private static final String SCHEME = "gs";
+  private static final String CREDENTIALS_GENERATOR_CLASS = GcsCredGenerator.class.getName();
+
+  @Override
+  protected void setUpProperties() {
+    super.setUpProperties();
+    serverProperties.put("gcs.bucketPath.0", "gs://" + BUCKET_NAME);
+    serverProperties.put("gcs.jsonKeyFilePath.0", "");
```
openinx: What's the meaning of this field,  if it's empty, is it necessary ?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredRenewITTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Override
+  protected String scheme() {
+    return SCHEME;
+  }
+
+  @Override
+  protected Map<String, String> catalogExtraProps() {
+    return Map.of(
+        "fs.gs.impl", GcsCredFileSystem.class.getName(), "fs.gs.impl.disable.cache", "true");
```
openinx: we don't need to set the `fs.gs.impl.disable.cache` any more here, because we already set it in the unity catalog.  It will automatically propagate to hadoop configuration.

timothyw553: Ah I see, gotcha. Thanks! So the server already installs that flag in the propagated Hadoop configuration, so the test only needs to wire up the custom filesystem implementation if my understanding is correct.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredRenewITTest.java
timothyw553: Oops yes, it was left there originally from when I was moving stuff over from the original PR.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredRenewITTest.java
openinx: > the server already installs that flag

Not the server,  it's the `uc-spark-connector`,  which is the connector will set those configuration keys. 
https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java#L102

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredRenewITTest.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    protected AccessTokenProvider createProvider() {
+      String clazz = getConf().get("fs.gs.auth.access.token.provider");
+      assertThat(clazz).isEqualTo(GcsVendedTokenProvider.class.getName());
+
+      try {
+        AccessTokenProvider provider =
+            (AccessTokenProvider) Class.forName(clazz).getDeclaredConstructor().newInstance();
```
openinx: Why not just `new GcsVendedTokenProvider` directly ?  I think it has the same effect right ?

server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java
```diff
@@ -15,52 +15,60 @@
...
+    String storageBase = credentialContext.getStorageBase();
+    GcsStorageConfig storageConfig = gcsConfigurations.get(storageBase);
+
+    GcpCredentialsGenerator generator =
+        credentialGenerators.computeIfAbsent(storageBase, key -> createGenerator(storageConfig));
+    return generator.generate(credentialContext);
+  }
+
+  private GcpCredentialsGenerator createGenerator(GcsStorageConfig storageConfig) {
+    if (storageConfig != null) {
```
openinx: When will the `storageConfig` to be null ?  If the `storageBase` is existing in the `gcsConfigurations` map, then the `storageConfig` should always be non-null, right ? while if `storageBase` is not existing,  then it should reports as an invalid `storageBase`, right ?

server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java
```diff
@@ -15,52 +15,60 @@
...
+        }
       }
     }
 
-    return downscopeGcpCreds(creds.createScoped(INITIAL_SCOPES), credentialContext)
-        .refreshAccessToken();
+    String jsonKeyFilePath = storageConfig != null ? storageConfig.getJsonKeyFilePath() : null;
+    if (jsonKeyFilePath != null && jsonKeyFilePath.isEmpty()) {
+      jsonKeyFilePath = null;
+    } else if (jsonKeyFilePath != null && jsonKeyFilePath.startsWith(TESTING_TOKEN_PREFIX)) {
```
openinx: Maybe we can remove this special case for testing, and just configure the `generatorClass` to be `TestingCredentialsGenerator`, then it will automatically load the `TestingCredentialsGenerator` in here https://github.com/unitycatalog/unitycatalog/pull/1209/files#diff-259f12dc6991de29d1a6d85eeac6a7faa919a2a713ec703e8fb75a5befb88de1R55-R56, for testing. 

Right ?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/GcsCredRenewITTest.java
timothyw553: Ah yes I relocated this previously.

server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java
timothyw553: Ah yes, we should check before even trying to create generator.

server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java
timothyw553: Yeah, we should decouple the test from the gcpcredentialvendor

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
```diff
@@ -18,8 +17,8 @@ public void testApiClientBaseUri() {
         .isEqualTo("https://localhost:8080/api/2.1/unity-catalog");
 
     URI uriWithSuffix = URI.create("https://localhost:8080/path/to/uc/api");
-    ApiClient apiClientWithSuffix = ApiClientFactory
-        .createApiClient(clientConf, uriWithSuffix, token);
+    ApiClient apiClientWithSuffix =
```
timothyw553, line 20: this file was caught when i ran javafmtAll

openinx (approved):
LGTM

tdas (approved):
LGTM

Issue #1200, opened 2025-11-12
OAuth2 support for oss-unitycatalog spark connector


openinx:
In this issue,  I'm trying to add the oauth2 in oss-unitycatalog spark connector. 

There are following things that we may need to consider in this issue:
*  M2M client credentials.
* workload identity federation.
* Token roation.
* JWKS endpoints.

Some related issues and PRs are:
* https://github.com/unitycatalog/unitycatalog/pull/1201
* https://github.com/unitycatalog/unitycatalog/issues/1071
* https://github.com/unitycatalog/unitycatalog/pull/1064

openinx:
After I run the test in this PR,  I use the following command to run:

```
export UC_URI=<UC-SERVER-URL>
export OAUTH_URI=<url> #same as the above curl command.
export OAUTH_CLIENT_ID=******
export OAUTH_CLIENT_SECRET=******
./build/sbt -J-Xmx8G -J-XX:+UseG1GC "spark/testOnly io.unitycatalog.spark.oauth.TestOAuth"
```

It reports the following error:

```java
[error] 25/11/13 23:09:02 WARN UCCommitCoordinatorBuilder: MessageWithContext(Skipping catalog main as it does ,{})not have both uri and token configured in Spark Session.
[error] Test io.unitycatalog.spark.oauth.TestOAuth.testOAuth failed: java.lang.IllegalArgumentException: Catalog main not found in the provided SparkSession configurations., took 7.254s
[error]     at org.apache.spark.sql.delta.coordinatedcommits.UCCommitCoordinatorBuilder$.buildForCatalog(UCCommitCoordinatorBuilder.scala:89)
[error]     at org.apache.spark.sql.delta.coordinatedcommits.CatalogOwnedTableUtils$.$anonfun$getCommitCoordinator$2(CoordinatedCommitsUtils.scala:104)
[error]     at scala.Option.orElse(Option.scala:477)
[error]     at org.apache.spark.sql.delta.coordinatedcommits.CatalogOwnedTableUtils$.getCommitCoordinator(CoordinatedCommitsUtils.scala:102)
[error]     at org.apache.spark.sql.delta.coordinatedcommits.CatalogOwnedTableUtils$.$anonfun$populateTableCommitCoordinatorFromCatalog$1(CoordinatedCommitsUtils.scala:57)
[error]     at scala.Option.map(Option.scala:242)
[...]

PR #1220, opened 2025-11-19
Support OAuth UCTokenProvider

openinx:
**PR Checklist**

**Description of changes**

The related issue is here: https://github.com/unitycatalog/unitycatalog/issues/1200

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/catalog/UCTokenProviderTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/catalog/UCTokenProviderTest.java
@@ -46,7 +46,7 @@ public class UCTokenProviderTest {
     // Test with no valid config - should throw
     assertThatThrownBy(() -> UCTokenProvider.create(Map.of()))
         .isInstanceOf(IllegalArgumentException.class)
-        .hasMessageContaining("Cannot determine UCTokenProvider from options");
+        .hasMessageContaining("Cannot determine UC authentication configuration from options");
 
     // Test token takes precedence over OAuth when both are present
     Map<String, String> bothOptions = new HashMap<>();
@@ -91,7 +91,7 @@ public class UCTokenProviderTest {
     Configuration emptyConf = new Configuration();
     assertThatThrownBy(() -> UCTokenProvider.create(emptyConf))
         .isInstanceOf(IllegalArgumentException.class)
-        .hasMessageContaining("Cannot determine UCTokenProvider from options");
+        .hasMessageContaining("Cannot determine UC authentication configuration from options");
 
     // Test token takes precedence over OAuth in Configuration
     Configuration bothConf = new Configuration();
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AbfsVendedTokenProviderTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AbfsVendedTokenProviderTest.java
@@ -79,7 +79,7 @@ public class AbfsVendedTokenProviderTest extends BaseTokenProviderTest<AbfsVende
     conf.set(UCHadoopConf.UC_URI_KEY, "http://localhost:8080");
     assertThatThrownBy(() -> provider.initialize(conf))
         .isInstanceOf(IllegalArgumentException.class)
-        .hasMessage("Cannot determine UC authentication configuration from options");
+        .hasMessageContaining("Cannot determine UC authentication configuration from options");
 
     // Verify the UID validation error message.
     conf.set(UCHadoopConf.UC_TOKEN_KEY, "unity-catalog-token");
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AwsVendedTokenProviderTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AwsVendedTokenProviderTest.java
@@ -92,7 +92,7 @@ public class AwsVendedTokenProviderTest extends BaseTokenProviderTest<AwsVendedT
     conf.set(UCHadoopConf.UC_URI_KEY, "http://localhost:8080");
     assertThatThrownBy(() -> new AwsVendedTokenProvider(conf))
         .isInstanceOf(IllegalArgumentException.class)
-        .hasMessage("Cannot determine UC authentication configuration from options");
+        .hasMessageContaining("Cannot determine UC authentication configuration from options");
 
     // Verify the UID validation error message.
     conf.set(UCHadoopConf.UC_TOKEN_KEY, "unity-catalog-token");
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/GcsVendedTokenProviderTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/GcsVendedTokenProviderTest.java
@@ -89,7 +89,7 @@ public class GcsVendedTokenProviderTest extends BaseTokenProviderTest<GcsVendedT
     conf.set(UCHadoopConf.UC_URI_KEY, "http://localhost:8080");
     assertThatThrownBy(() -> provider.setConf(conf))
         .isInstanceOf(IllegalArgumentException.class)
-        .hasMessage("Cannot determine UC authentication configuration from options");
+        .hasMessageContaining("Cannot determine UC authentication configuration from options");
 
     conf.set(UCHadoopConf.UC_TOKEN_KEY, "unity-catalog-token");
     assertThatThrownBy(() -> provider.setConf(conf))


connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -33,8 +34,8 @@ public T uri(String uri) {
       return self();
     }
 
-    public T token(String token) {
-      builder.put(UCHadoopConf.UC_TOKEN_KEY, token);
+    public T ucTokenProvider(UCTokenProvider ucTokenProvider) {
+      builder.putAll(ucTokenProvider.properties());
```
openinx, line 41: Now, we need the `UCTokenProvider` to generate the hadoop configuration keys and values, so that it can  propagate to the spark driver and spark executors.  And then the spark driver and the spark executors's cloud vendor filesystem will finally use the `fs.unitycatalog.*`  keys and values to reconstruct the `UCTokenProvider`. 

Notice:  we set the spark **catalog properties** when we launch the spark job, but those properties won't be propagate to driver & executor's `S3AFileSystem`, `ABFSFileSystem` etc. Since those catalog properties are not prefixed with `fs.*`, and the OSS-Delta will automatically filter all those catalog properties without the correct prefix. 

Appendix:

1. The **catalog properties** we need to set are:

```bash
spark.sql.catalog.<catalogName>.oauthUri
spark.sql.catalog.<catalogName>.oauthClientId
spark.sql.catalog.<catalogName>.oauthClientSecret.
```

2. The **hadoop configuration keys** that the UCTokenProvide will set are:

```
fs.unitycatalog.oauthUri
fs.unitycatalog.oauthClientId
fs.unitycatalog.oauthClientSecret
```

Both the catalog properties and hadoop configured keys should share the same suffix components: `oauthUri`, `oauthClientId`, `oauthClientSecret`.

And also the delta's OAuth properties also need share the same suffix: https://github.com/delta-io/delta/pull/5511/files#diff-e713dd4904f1e14bc5a90fc8517d818c6aa12655799e935e6f7a020d950c47ceR194-R196

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/catalog/UCTokenProvider.java
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Creates a token provider from Hadoop configuration by extracting Unity Catalog properties
+   * and delegating to {@link #create(Map)}.
+   *
+   * @param conf Hadoop configuration containing Unity Catalog properties
+   * @return a token provider instance
+   * @throws IllegalArgumentException if configuration is missing or incomplete
+   */
+  static UCTokenProvider create(Configuration conf) {
```
openinx: We need this hadoopConfiguration based `create` to parse all the `UCTokenProvider` properties from the hadoop configuration, and finally construct a new `UCTokenProvider` instance.

This is critical important, because once we launch the Spark Job at the client side. The Spark driver and executors will deserialize the binary bytes into hadoop configuration instance, and the driver&executors need to use this hadoop configuration to construct `GenericCredentialProvider`, where will finally construct the `UCTokenProvider` in its implementation.

openinx: Combined with this context (https://github.com/unitycatalog/unitycatalog/pull/1220/files#r2543301751), it may make thing and understanding more clear.

connectors/spark/src/main/scala/io/unitycatalog/spark/token/FixedUCTokenProvider.java
```diff
@@ -0,0 +1,26 @@
+package io.unitycatalog.spark.token;
```
tdas: arent all of these related to auth? yeah, its not storage authentication, but UC authentication, but still its related auth right?
if so, should we move it to `auth` dir? 
`token`is a very generic term and by itself confusing and ambiguous. "auth token" makes sense. but at that point, might as well put it under `auth` package. 

if you want to be really structured, then all storage auth can go into `auth/storage` and this UC auth stuff and go into `auth/rest` or `auth/catalog`. but i am okay to keep things simple for now.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/catalog/UCTokenProvider.java
```diff
@@ -0,0 +1,70 @@
...
+   *
+   * @return a map of configuration key-value pairs
+   */
+  Map<String, String> properties();
+
+  /**
+   * Creates a token provider from configuration options. Returns {@link FixedUCTokenProvider} if
+   * a token is provided, or {@link OAuthUCTokenProvider} if OAuth credentials are provided.
+   *
+   * @param options containing authentication credentials
```
tdas: which options are these? `spark.sql.catalog.<catalogName>.` or `fs.unitycatalog.`?? its not obvious from neither the name or method description.

tdas: i suggest clarifying that this method assumes that there is no prefix in th keys.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -41,8 +41,12 @@ private UCHadoopConf() {
   // custom configuration keys (e.g., AWS, ABFS, or GCS credentials) propagate correctly into Hadoop
   // FileSystems such as S3AFileSystem or AbfsFileSystem, the keys must include the "fs." prefix.
   // See also: https://github.com/unitycatalog/unitycatalog/issues/1112.
+  public static final String FS_UC_PREFIX = "fs.unitycatalog.";
   public static final String UC_URI_KEY = "fs.unitycatalog.uri";
   public static final String UC_TOKEN_KEY = "fs.unitycatalog.token";
+  public static final String UC_OAUTH_URI = "fs.unitycatalog.oauthUri";
```
tdas: should we just put all of these auth related confgs under `fs.unitycatalog.auth.*`
like `fs.unitycatalog.auth.uri|token|oauthUri|...`

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/catalog/UCTokenProvider.java
```diff
@@ -0,0 +1,70 @@
+package io.unitycatalog.spark.token;
+
+import static org.sparkproject.guava.base.Preconditions.checkArgument;
+
+import io.unitycatalog.spark.UCHadoopConf;
+import io.unitycatalog.spark.utils.OptionsUtil;
+import java.util.Map;
+import org.apache.hadoop.conf.Configuration;
+
+public interface UCTokenProvider {
```
tdas, line 19: class docs

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/OptionsUtil.java
```diff
@@ -9,6 +9,9 @@ private OptionsUtil() {
   public static final String URI = "uri";
   public static final String TOKEN = "token";
   public static final String WAREHOUSE = "warehouse";
+  public static final String OAUTH_URI = "oauthUri";
```
tdas: should we do `oauth.[uri | clientId | clientSecret]` ? clearly oauth is a namespace with 3 params in it.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -41,8 +41,12 @@ private UCHadoopConf() {
   // custom configuration keys (e.g., AWS, ABFS, or GCS credentials) propagate correctly into Hadoop
   // FileSystems such as S3AFileSystem or AbfsFileSystem, the keys must include the "fs." prefix.
   // See also: https://github.com/unitycatalog/unitycatalog/issues/1112.
+  public static final String FS_UC_PREFIX = "fs.unitycatalog.";
```
tdas, line 44: In general, i dont think this is the right abstraction. we are using hadoopconf as way to communicate conf. we should be consistent with how and where we set and get configs. 

- Approach 1: get/setApiClientConf has the abstraction that ApiClientConf does not have to care or understand what property names are being used. But UCHadoopConf still has to understand every detail of ApiClientConf. So its half-way abstracted out, only in 1 direction.
- Approach 2: every other conf, all usages need to understand the configurations. Thats why we are having to expose these internal details like prefix. And UCHadoopConf needs to understand every parameter inside other stuff. So not abstracted in both directions, worse than Approach 1. 

What are you thoughts on a better approach? Here is an idea. 

A trait UCHadoopConf that has
- `prefix(): String` -> `fs.unitycatalog.<prefix>`
- `getConfs(): Map<String, String>` --> called when serializing to hadoop key-value pairs that will be put in hadoopconf as `fs.unitycatalog.<prefix>.<key> = <value>`
- `setConfs(params:  Map<String, String>)` --> called when 

Other confs can extend UCHadoopConf by implementing the methods. Any class that extends UCHadoopConf can be passed into the static methods 
```
class UCHadoopConf {
  void setInHadoopConf(hadoopConf: Configuration, ucConf: UCHadoopConf) // will call ucConf.getConfs()
  void getFromHadoopConf(hadoopConf: Configuration, ucConf: UCHadoopConf) // will call ucConf.setConfs()
} 
```

tdas: then every component can define their own confs keys in its own namespace without having to understand how it will be written hadoopconf, and hadoopconf does not need to understand all the conf keys. 

isnt that a good abstraction?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/catalog/UCTokenProvider.java
```diff
@@ -0,0 +1,70 @@
...
+    String oauthUri = options.get(OptionsUtil.OAUTH_URI);
+    String oauthClientId = options.get(OptionsUtil.OAUTH_CLIENT_ID);
+    String oauthClientSecret = options.get(OptionsUtil.OAUTH_CLIENT_SECRET);
+    if (oauthUri != null || oauthClientId != null || oauthClientSecret != null) {
+      checkArgument(oauthUri != null && oauthClientId != null && oauthClientSecret != null,
+          "Incomplete OAuth configuration detected. All of the keys are required: " +
+              "spark.sql.catalog.<catalogName>.oauthUri, " +
+              "spark.sql.catalog.<catalogName>.oauthClientId, " +
+              "spark.sql.catalog.<catalogName>.oauthClientSecret. " +
+              "Please ensure they are all set.");
```
tdas: this error message does not make sense when this method has been called from `UCTokenProvider create(Configuration conf)` where the problem is not user not setting these confs, but somehow hadoop conf being incorrect. 


so the error message needs to be different based on the caller of this generic method

connectors/spark/src/main/scala/io/unitycatalog/spark/token/UCTokenProvider.java
```diff
@@ -0,0 +1,70 @@
...
+          "Incomplete OAuth configuration detected. All of the keys are required: " +
+              "spark.sql.catalog.<catalogName>.oauthUri, " +
+              "spark.sql.catalog.<catalogName>.oauthClientId, " +
+              "spark.sql.catalog.<catalogName>.oauthClientSecret. " +
+              "Please ensure they are all set.");
+
+      return new OAuthUCTokenProvider(oauthUri, oauthClientId, oauthClientSecret);
+    }
+
+    throw new IllegalArgumentException("Cannot determine UCTokenProvider from options");
```
tdas: `UCTokenProvider` is an internal class name. this is not useful to show as a message to the user. rather better to say "Cannot determine UC authentication configuration from options". 

additionally can you print out the configurations that are indeed set (except the token and secret which can be printed as `***`)

connectors/spark/src/main/scala/io/unitycatalog/spark/token/UCTokenProvider.java
```diff
@@ -0,0 +1,70 @@
...
+    // If OAuth options is available, use OAuthUCTokenProvider.
+    String oauthUri = options.get(OptionsUtil.OAUTH_URI);
+    String oauthClientId = options.get(OptionsUtil.OAUTH_CLIENT_ID);
+    String oauthClientSecret = options.get(OptionsUtil.OAUTH_CLIENT_SECRET);
+    if (oauthUri != null || oauthClientId != null || oauthClientSecret != null) {
+      checkArgument(oauthUri != null && oauthClientId != null && oauthClientSecret != null,
+          "Incomplete OAuth configuration detected. All of the keys are required: " +
+              "spark.sql.catalog.<catalogName>.oauthUri, " +
+              "spark.sql.catalog.<catalogName>.oauthClientId, " +
+              "spark.sql.catalog.<catalogName>.oauthClientSecret. " +
```
tdas: where are these confs keys defined?

connectors/spark/src/main/scala/io/unitycatalog/spark/token/FixedUCTokenProvider.java
openinx: Yes,  I want to separate those **cloud vendor** token providers and **catalog** token providers into two different package, so that we can easily to locate and distinguish between them, and the maintainers can easily understand them. 

Originally,  I thought maybe we can maintain them into the `io.unitycatalog.spark.uctoken` and `io.unitycatalog.spark.auth` packages.  But seem the `io.unitycatalog` already implies the unitycatalog, which have a duplicated meaning in `uctoken`.  so I chose the `io.unitycatalog.spark.token`.

After reading the above comment,  the most meaningful packages should be `auth/storage` and `auth/catalog`, which precisely describe the **cloud vendor** token providers **catalog** token providers, and also express the meaning that all of them are the `auth` classes.

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/OptionsUtil.java
openinx: Okay, that sounds good to me, let me follow the rules that we've agreed.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
openinx: the `fs.unitycatalog.uri` should be a separate one,  which don't have to include `auth`. Since it's a general API endpoint for all the requests. 

For others ( including `token`, `oauthUri`, `oauthClientId`, `oauthClientSecret` ),  adding a `auth` make sense to me.

openinx: Reconsidered this,  for the `oauthXX`,  which the `oauth` already implies it's a key used for `auth`,  do we still need to add the extra `auth` component in the key ?  Maybe we don't have to.

And for the `fs.unitycatalog.token`,  it's naturally a key used for unitycatalog auth also, which means we may also don't have to add an extra `auth`.

openinx: Since we've agreed following this guidance https://github.com/unitycatalog/unitycatalog/pull/1220/files#r2558284019 for the catalog properties naming,  the filesystem keys should also follow the same naming. then it will be:

* `fs.unitycatalog.token`
* `fs.unitycatalog.oauth.uri`
* `fs.unitycatalog.oauth.clientId`
* `fs.unitycatalog.oauth.clientSecret`

tdas: I am good with this. token stays same. oauth is a namespace. And both catalog props and Hadoop conf props have same pattern.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
openinx: Essentially, we need to consistently handle how the catalog properties go through to the hadoop filesystem hadoop configuration, so that the unitycatalog auth can correctly and consistently access those keys. 

I think it's great point.  Let me file a separate issue to do that, rather than a block for this PR.

tdas: Yes I am good with refactoring this after this PR. This is purely internal and can be changed anytime for general internal code cleanliness.

tdas: Yes I am good with refactoring this after this PR. This is purely internal and can be changed anytime for general internal code cleanliness.

openinx: Here is the issue: https://github.com/unitycatalog/unitycatalog/issues/1245

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/catalog/UCTokenProvider.java
openinx: Yes, agreed.  Let me separate the error message, and make them more clear.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/catalog/UCTokenProvider.java
openinx: Done

connectors/spark/src/main/scala/io/unitycatalog/spark/token/UCTokenProvider.java
openinx: I only defined those suffix in the https://github.com/unitycatalog/unitycatalog/pull/1220/files#diff-ee85a43625f6709e9ad3e2e2629e6716fa6a968cfd1b94d7b19d711d8d322d0fR12-R14. 

And with the `<catalogName>`, we can easily infer those catalog config keys that needed to be configured in the spark session.

connectors/spark/src/main/scala/io/unitycatalog/spark/token/UCTokenProvider.java
openinx: Make sense, let me make it more clear.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/catalog/UCTokenProvider.java
openinx: Done

connectors/spark/src/main/scala/io/unitycatalog/spark/token/UCTokenProvider.java
openinx: Done

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/catalog/OAuthUCTokenProvider.java
```diff
@@ -0,0 +1,147 @@
...
+      // Build HTTP request
+      HttpRequest request = HttpRequest.newBuilder()
+          .uri(URI.create(oauthUri))
+          .header("Authorization", "Basic " + encodedCredentials)
+          .header("Content-Type", "application/x-www-form-urlencoded")
+          .POST(HttpRequest.BodyPublishers.ofString(formData))
+          .build();
+
+      // Send request with retry support
+      HttpResponse<String> response = httpClient.send(request,
```
tdas, line 109: where is the retry support here? is it automatic in the httpClient.send()?

tdas:
- #1253

PR #1223, opened 2025-11-19
Refactor ExternalTableReadWriteTest

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1223/files) to review incremental changes.
- [**stack/refactor_table_test**](https://github.com/unitycatalog/unitycatalog/pull/1223) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1223/files)]
  - [stack/uc_spark_managed](https://github.com/unitycatalog/unitycatalog/pull/1225) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1225/files/ed13a31c4e08f24e30f4140af58761107b849c17..ec928dc9307cb084fd638b5471a16534318fc1f8)]

---------
**PR Checklist**

**Description of changes**
Refactor ExternalTableReadWriteTest
- Move the test that can be shared between external and managed tables to BaseTableReadWriteTest
- Make the functions that creates table base class functions
- Make the cloud test parameters configurable in subclasses

All 42 tests are still running in the same way.

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
@@ -16,7 +16,9 @@ import io.unitycatalog.server.sdk.schema.SdkSchemaOperations;
 import io.unitycatalog.server.service.credential.gcp.TestingCredentialsGenerator;
 import io.unitycatalog.server.utils.TestUtils;
 import io.unitycatalog.spark.utils.OptionsUtil;
+import java.util.List;
 import java.util.Optional;
+import org.apache.spark.sql.Row;
 import org.apache.spark.sql.SparkSession;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
@@ -67,11 +69,17 @@ public abstract class BaseSparkIntegrationTest extends BaseCRUDTest {
     return builder.getOrCreate();
   }
 
+  protected List<Row> sql(String statement, Object... args) {
+    return session.sql(String.format(statement, args)).collectAsList();
+  }
+
   @BeforeEach
   @Override
   public void setUp() {
     super.setUp();
     // Some Delta Spark functionalities needs testing mode to be turned on so that we can test.
+    // Specifically the file CreateDeltaTableCommand.scala in Delta checks for Utils.isTesting
+    // before allowing catalog owned table creation.
     System.setProperty("spark.testing", "true");
     schemaOperations = new SdkSchemaOperations(createApiClient(serverConfig));
     try {
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
@@ -10,9 +10,8 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.List;
-import org.apache.spark.sql.Dataset;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.spark.sql.Row;
-import org.apache.spark.sql.SparkSession;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
@@ -24,23 +23,21 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
   private static final String DELTA_TABLE_PARTITIONED = "test_delta_partitioned";
 
   /**
-   * Set up a delta table and returns the table full name. Subclasses need to override it
-   * accordingly.
+   * Set up a delta table and returns the table full name. This function is used by tests that
+   * aren't cloud aware. It simply uses the same cloud as configured for managed storage.
    */
-  protected abstract String setupDeltaTable(
-      String catalogName, String tableName, List<String> partitionColumns, SparkSession session)
-      throws IOException, ApiException;
+  protected final String setupDeltaTable(
+      String catalogName, String tableName, List<String> partitionColumns)
+      throws IOException, ApiException {
+    return setupDeltaTable(managedStorageCloudScheme(), catalogName, tableName, partitionColumns);
+  }
 
   /**
    * Set up a delta table with storage on emulated cloud and returns the table full name. Subclasses
    * need to override it accordingly.
    */
-  protected abstract String setupDeltaTableForCloud(
-      String scheme,
-      String catalogName,
-      String tableName,
-      List<String> partitionColumns,
-      SparkSession session)
+  protected abstract String setupDeltaTable(
+      String cloudScheme, String catalogName, String tableName, List<String> partitionColumns)
       throws IOException, ApiException;
 
   @Test
@@ -48,43 +45,65 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
     // Test both `spark_catalog` and other catalog names.
     session = createSparkSessionWithCatalogs(SPARK_CATALOG, CATALOG_NAME);
 
-    String t1 = setupDeltaTable(SPARK_CATALOG, DELTA_TABLE, List.of(), session);
-    testTableReadWrite(t1, session);
+    String t1 = setupDeltaTable(SPARK_CATALOG, DELTA_TABLE, List.of());
+    testTableReadWrite(t1);
 
-    String t2 =
-        setupDeltaTable(SPARK_CATALOG, DELTA_TABLE_PARTITIONED, Arrays.asList("s"), session);
-    testTableReadWrite(t2, session);
+    String t2 = setupDeltaTable(SPARK_CATALOG, DELTA_TABLE_PARTITIONED, Arrays.asList("s"));
+    testTableReadWrite(t2);
 
-    String t3 = setupDeltaTable(CATALOG_NAME, DELTA_TABLE, List.of(), session);
-    testTableReadWrite(t3, session);
+    String t3 = setupDeltaTable(CATALOG_NAME, DELTA_TABLE, List.of());
+    testTableReadWrite(t3);
 
-    String t4 = setupDeltaTable(CATALOG_NAME, DELTA_TABLE_PARTITIONED, Arrays.asList("s"), session);
-    testTableReadWrite(t4, session);
+    String t4 = setupDeltaTable(CATALOG_NAME, DELTA_TABLE_PARTITIONED, Arrays.asList("s"));
+    testTableReadWrite(t4);
   }
 
-  private void validateTimeTravelDeltaTable(Dataset<Row> df) {
-    List<Row> rows = df.collectAsList();
-    assertThat(rows).hasSize(1);
-    assertThat(rows.get(0).getInt(0)).isEqualTo(1);
+  /**
+   * This function checks that the first column of each row in {@code rows} matches {@code expected}
+   * in both length and value.
+   *
+   * @param rows Rows to be checked
+   * @param expected The expected integer values of first column
+   */
+  protected void validateRows(List<Row> rows, Integer... expected) {
+    assertThat(rows).hasSize(expected.length);
+    for (int i = 0; i < expected.length; i++) {
+      assertThat(rows.get(i).getInt(0)).isEqualTo(expected[i]);
+    }
+  }
+
+  /**
+   * This function checks that the first and second column of each row in {@code rows} matches
+   * {@code expected} in both length and value.
+   *
+   * @param rows Rows to be checked
+   * @param expected The expected integer and string values of first two columns
+   */
+  protected void validateRows(List<Row> rows, Pair<Integer, String>... expected) {
+    assertThat(rows).hasSize(expected.length);
+    for (int i = 0; i < expected.length; i++) {
+      Row row = rows.get(i);
+      assertThat(row.getInt(0)).isEqualTo(expected[i].getLeft());
+      assertThat(row.getString(1)).isEqualTo(expected[i].getRight());
+    }
   }
 
   @Test
   public void testTimeTravelDeltaTable() throws ApiException, IOException {
     session = createSparkSessionWithCatalogs(SPARK_CATALOG);
 
-    String t1 = setupDeltaTable(SPARK_CATALOG, DELTA_TABLE, List.of(), session);
-    session.sql("INSERT INTO " + t1 + " SELECT 1, 'a'");
+    String t1 = setupDeltaTable(SPARK_CATALOG, DELTA_TABLE, List.of());
+    sql("INSERT INTO %s SELECT 1, 'a'", t1);
 
     String timestamp = Instant.now().toString();
 
-    session.sql("INSERT INTO " + t1 + " SELECT 2, 'b'");
+    sql("INSERT INTO %s SELECT 2, 'b'", t1);
 
     // Time-travel to before the last insert, we should only see the first inserted row.
-    validateTimeTravelDeltaTable(session.sql("SELECT * FROM " + t1 + " VERSION AS OF 1"));
-    validateTimeTravelDeltaTable(
-        session.sql("SELECT * FROM " + t1 + " TIMESTAMP AS OF '" + timestamp + "'"));
-    validateTimeTravelDeltaTable(session.read().option("versionAsOf", 1).table(t1));
-    validateTimeTravelDeltaTable(session.read().option("timestampAsOf", timestamp).table(t1));
+    validateRows(sql("SELECT * FROM %s VERSION AS OF 1", t1), 1);
+    validateRows(sql("SELECT * FROM %s TIMESTAMP AS OF '%s'", t1, timestamp), 1);
+    validateRows(session.read().option("versionAsOf", 1).table(t1).collectAsList(), 1);
+    validateRows(session.read().option("timestampAsOf", timestamp).table(t1).collectAsList(), 1);
   }
 
   @ParameterizedTest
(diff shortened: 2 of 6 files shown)


connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
```diff
@@ -67,8 +70,9 @@ protected SparkSession createSparkSessionWithCatalogs(boolean renewCred, String.
   @Override
   public void setUp() {
     super.setUp();
+    // Some Delta Spark functionalities needs testing mode to be turned on so that we can test.
+    System.setProperty("spark.testing", "true");
```
openinx, line 83: Why we need to set this `spark.testing` property for every `setUp`?   What are the cases that we will use this ?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
```diff
@@ -110,13 +114,22 @@ protected CatalogOperations createCatalogOperations(ServerConfig serverConfig) {
...
+  @AfterEach
   @Override
   public void cleanUp() {
     try {
       catalogOperations.deleteCatalog(SPARK_CATALOG, Optional.of(true));
     } catch (Exception e) {
       // Ignore
     }
+    try {
+      if (session != null) {
```
openinx, line 137: So we will create the spark session for every method testing, rather than the class testing ? Typically, it's better to maintain the `session` in the class-level lifecycle, which will save the spark catalog initialization/destory time.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -0,0 +1,195 @@
...
+
+  private static final String DELTA_TABLE = "test_delta";
+  private static final String ANOTHER_DELTA_TABLE = "test_delta_another";
+  private static final String DELTA_TABLE_PARTITIONED = "test_delta_partitioned";
+
+  /**
+   * Set up a delta table and returns the table full name. Subclasses need to override it
+   * accordingly.
+   */
+  protected abstract String setupDeltaTable(
```
openinx: what's the difference between `setupDeltaTable` and `setupDeltaTableForCloud` ? What are the cases that we will setup a delta table without could support ?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
yili-db: Currently there's a line of code in Delta that prevents CCv2 table creation:
```
    if (!Utils.isTesting && (tableFeatures.contains(CatalogOwnedTableFeature) ||
      CatalogOwnedTableUtils.defaultCatalogOwnedEnabled(spark = sparkSession))) {
      throw DeltaErrors.deltaCannotCreateCatalogOwnedTable()
    }
```
Yes it can be setup just once in a `@BeforeAll` function. But it's done here since there's no existing function of `@BeforeAll` and it's super lightweight operation anyway.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
yili-db: The tests want to create session differently each time:
```
SparkSession session = createSparkSessionWithCatalogs(renewCredEnabled, SPARK_CATALOG);
```
or
```
SparkSession session = createSparkSessionWithCatalogs(SPARK_CATALOG, CATALOG_NAME);
```
etc

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: There are tests that don't have cloud parameters. They just test with local file path and they don't care credential renew either.
For example:
```
  @Test
  public void testDeltaReadWrite() throws IOException, ApiException {
```

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.Arrays;
+import java.util.List;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
```
openinx, line 19: The intention that you make this abstraction is ?  I guess you want to extend this `BaseTableReadWriteTest` to test the managed table ?  But for managed table creation,  it shouldn't test the external table creation cases ( I mean all those cases are non-managed table testing, then why need to abstract them).

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -0,0 +1,195 @@
...
+    // Time-travel to before the last insert, we should only see the first inserted row.
+    validateTimeTravelDeltaTable(session.sql("SELECT * FROM " + t1 + " VERSION AS OF 1"));
+    validateTimeTravelDeltaTable(
+        session.sql("SELECT * FROM " + t1 + " TIMESTAMP AS OF '" + timestamp + "'"));
+    validateTimeTravelDeltaTable(session.read().option("versionAsOf", 1).table(t1));
+    validateTimeTravelDeltaTable(session.read().option("timestampAsOf", timestamp).table(t1));
+  }
+
+  @ParameterizedTest
+  @MethodSource("cloudParameters")
```
openinx, line 110: That's a good abstraction, thanks !

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: Some, not all, of the tests in ‎`TableReadWriteTest.java` are moved to `BaseTableReadWriteTest.java` . These tests **can be shared with managed tables**. They used to test external tables only, but from now on they can also be used for testing managed tables as these table type is now abstracted.
The remaining tests that are only related to external tables are moved to ExternalTableReadWriteTest.java.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -0,0 +1,195 @@
...
+    assertThat(rows).hasSize(1);
+    assertThat(rows.get(0).getInt(0)).isEqualTo(1);
+  }
+
+  @Test
+  public void testTimeTravelDeltaTable() throws ApiException, IOException {
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG);
+
+    String t1 = setupDeltaTable(SPARK_CATALOG, DELTA_TABLE, List.of(), session);
+    session.sql("INSERT INTO " + t1 + " SELECT 1, 'a'");
```
openinx: Since we've already planed to add the spark `session` variable in `BaseSparkIntegrationTest`,  then  I will suggest adding the  `List<Row> sql(String statement, Object...args)` in the `BaseSparkIntegrationTest`. https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/test/java/io/unitycatalog/spark/auth/BaseCredRenewITTest.java#L272-L274

There here and all the sql calling place, we can directly use:

```
sql("INSERT INTO %s SELECT 1, 'a'", t1);
```

That will make all the SQL more clear.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
tdas: can you leave a comment to the Delta file path for this context?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -0,0 +1,195 @@
...
+    assertThat(rows).hasSize(1);
+    assertThat(rows.get(0).getInt(0)).isEqualTo(2);
+  }
+
+  @Test
+  public void testShowTables() throws ApiException, IOException {
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG);
+    setupDeltaTable(SPARK_CATALOG, DELTA_TABLE, List.of(), session);
+
+    Row[] tables = (Row[]) session.sql("SHOW TABLES in " + SCHEMA_NAME).collect();
```
openinx: nit: please remove this redundant `(Row[])`.

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -19,34 +19,49 @@
...
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.CsvSource;
-
-public class TableReadWriteTest extends BaseSparkIntegrationTest {
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
+public class ExternalTableReadWriteTest extends BaseTableReadWriteTest {
```
openinx, line 38: Oh, I think I got.  You just abstracted all the common table operation integration tests into the `BaseTableReadWriteTest`, and exclude all the external table read write tests, which actually be moved into this separate classes.

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -19,34 +19,49 @@
...
   private static final String ANOTHER_PARQUET_TABLE = "test_parquet_another";
   private static final String PARQUET_TABLE_PARTITIONED = "test_parquet_partitioned";
   private static final String DELTA_TABLE = "test_delta";
   private static final String PARQUET_TABLE = "test_parquet";
   private static final String ANOTHER_DELTA_TABLE = "test_delta_another";
-  private static final String DELTA_TABLE_PARTITIONED = "test_delta_partitioned";
 
-  private final File dataDir = new File(System.getProperty("java.io.tmpdir"), "spark_test");
+  private File dataDir;
```
openinx: Maybe you can use `@TempDir` annoation directly,  like this: https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/test/java/io/unitycatalog/spark/auth/BaseCredRenewITTest.java#L66

Then you don't have to create an unique `dataDir` for every test method, since it's naturally unique generated after this annotation.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -0,0 +1,195 @@
...
+  private static final String DELTA_TABLE = "test_delta";
+  private static final String ANOTHER_DELTA_TABLE = "test_delta_another";
+  private static final String DELTA_TABLE_PARTITIONED = "test_delta_partitioned";
+
+  /**
+   * Set up a delta table and returns the table full name. Subclasses need to override it
+   * accordingly.
+   */
+  protected abstract String setupDeltaTable(
+      String catalogName, String tableName, List<String> partitionColumns, SparkSession session)
```
tdas: where is the schema name?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
tdas: the only difference is scheme as the parameter, right? why not unify these with scheme as an optional parameter?

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -19,34 +19,49 @@
...
+  // 0 and 1 to construct the bucket name.
+  private int bucketIndex = 0;
 
   private TableOperations tableOperations;
 
+  /**
+   * This function provides a set of test parameters that cloud-aware tests should run for this
+   * class.
+   */
+  protected static Stream<Arguments> cloudParameters() {
```
openinx, line 60: Nit: could you please add explannation for each field ? since it's a separate method, seems hard to understand what're the field meanings at the first glance.

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
```diff
@@ -62,53 +77,29 @@ public void testNoDeltaCatalog() throws IOException, ApiException {
             .config(catalogConf, UCSingleCatalog.class.getName())
             .config(catalogConf + "." + OptionsUtil.URI, serverConfig.getServerUrl())
             .config(catalogConf + "." + OptionsUtil.TOKEN, serverConfig.getAuthToken())
-            .config(catalogConf + "." + OptionsUtil.WAREHOUSE, CATALOG_NAME)
-            .config(catalogConf + ".__TEST_NO_DELTA__", "true");
```
openinx, line 66: Any usage for this key before ?

openinx, line 66: Any usage for this configure key before ?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -0,0 +1,195 @@
...
+
+    String t2 =
+        setupDeltaTable(SPARK_CATALOG, DELTA_TABLE_PARTITIONED, Arrays.asList("s"), session);
+    testTableReadWrite(t2, session);
+
+    String t3 = setupDeltaTable(CATALOG_NAME, DELTA_TABLE, List.of(), session);
+    testTableReadWrite(t3, session);
+
+    String t4 = setupDeltaTable(CATALOG_NAME, DELTA_TABLE_PARTITIONED, Arrays.asList("s"), session);
+    testTableReadWrite(t4, session);
```
tdas: can this be combined together into a for loop to reduce code?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: Despite what Intellij says, it's actually not redundant:
```
[error] /home/yi.li/unitycatalog/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java:165:1: incompatible types: java.lang.Object cannot be converted to org.apache.spark.sql.Row[]
[error] session.sql("SHOW TABLES in " + SCHEMA_NAME).collect()
```

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: All these tests don't care about schema. So they always use `TestUtils.SCHEMA_NAME` hardcoded.

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
yili-db: No. It's just a residual from the original PR. The test actually uses `UCSingleCatalog.LOAD_DELTA_CATALOG().set(false);` but this is forgotten to remove.

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -522,23 +349,29 @@ private void setupDeltaTableLocation(
     CredentialTestFileSystem.credentialCheckEnabled = true;
   }
 
-  private void setupExternalDeltaTable(
+  /**
+   * Returns an emulated cloud bucket to be used for creating external table. Alternating between
+   * two configured test buckets
+   */
+  private String testBucket(String scheme) {
```
openinx, line 325: This seems weird and hard to read ?  As we configured the bucket in the `serverProperties` in `BaseSparkIntegrationTest`,  why just add a `protected  rootUri(String scheme, int id)` in the ` BaseSparkIntegrationTest`, and then each subclass can specify the `scheme` and `id` separately in their test ?

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -19,34 +19,49 @@
...
 
   private TableOperations tableOperations;
 
+  /**
+   * This function provides a set of test parameters that cloud-aware tests should run for this
+   * class.
+   */
+  protected static Stream<Arguments> cloudParameters() {
+    return Stream.of(
+        Arguments.of("s3", false),
```
tdas, line 62: what is the boolean? comment please.

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -522,23 +349,29 @@ private void setupDeltaTableLocation(
...
+  @Override
+  protected String setupDeltaTableForCloud(
+      String scheme,
       String catalogName,
       String tableName,
-      String location,
       List<String> partitionColumns,
       SparkSession session)
       throws IOException, ApiException {
+    String cloudPrefix = scheme.equals("file") ? "" : testBucket(scheme);
```
openinx: I hope we can have an abstraction to hide all the filesystem differences between the local file and cloud storage,  so that we won't need to do any special check for the local file any more. 

But shouldn't be a block for this PR.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: Yes. But it doesn't save much code though. It will look a little more complicated.
Also I prefer to leave it as is in order to make the diff smaller and easier to review as this is an existing test.
Please check the list of commits and if you only look at the 2nd commit you can see better diff view.

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -603,18 +436,15 @@ private void setupTables(
   public void setUp() {
     super.setUp();
     tableOperations = new SdkTableOperations(createApiClient(serverConfig));
+    dataDir = testDirectoryRoot.resolve("spark_test").toFile();
```
openinx: After use the `@TempDir` annoation, then we don't need this any more.  

Or we must use the server-side root dir to do the managed table test ?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: I see. It should be `collectAsList` then the conversion is not needed.

server/src/test/java/io/unitycatalog/server/base/BaseServerTest.java
```diff
@@ -21,15 +27,45 @@ public abstract class BaseServerTest {
...
+  protected String getManagedStorageCloudPath(Path localPath) {
+    String localPathString;
+    localPathString = localPath.toAbsolutePath().normalize().toString();
+    String scheme = managedStorageCloudScheme();
+    if (scheme.equals("file")) {
+      return "file://" + localPathString;
+    } else {
+      return scheme + "://test-bucket0" + localPathString;
+    }
+  }
```
openinx, line 53: Please consider how to abstract the local file and cloud storage fs into a general path in the further work,  because I think it will be much easier for the downstream to do the  integration tests if they don't need to percieve the filesystem difference.

Otherwise, it's a high threshold for others to do the integration tests, because there are so many details that we need to touch if we want to make the integration run.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  @ParameterizedTest
+  @MethodSource("cloudParameters")
+  public void testDeleteFromDeltaTable(String scheme, boolean renewCredEnabled)
+      throws ApiException, IOException {
+    session = createSparkSessionWithCatalogs(renewCredEnabled, SPARK_CATALOG);
+
+    String t1 = setupDeltaTableForCloud(scheme, SPARK_CATALOG, DELTA_TABLE, List.of(), session);
+    testTableReadWrite(t1, session);
```
tdas: why are you do the whole testTableReadWrite test when the purpose of this test is only to delete?

server/src/test/java/io/unitycatalog/server/base/BaseServerTest.java
```diff
@@ -81,6 +125,21 @@ public void tearDown() {
       session.close();
 
       unityCatalogServer.stop();
+      if (testDirectoryRoot != null && Files.isDirectory(testDirectoryRoot)) {
```
openinx: Maybe we can use the `@TempDir` annoation in this class, then we don't need to worry about root path deletion.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
tdas: Why dont we abstract out such that it can be reused across formats as well. basically the same table CRUD ops need to work in absolutely the same way with "external parquet", "external delta", "managed delta" ... right? so why make the abstraction independent of the file format?

tdas: Basically, think about a better hierarchy.
base classes / interfaces
- BaseTableReadWriteTest (works for all tables formats and types)
- AdditionalDeltaTests (extra tests applicable Delta tables only, but both external and managed) 

real tests
- ParquetTableReadWriteTest extends BaseTableReadWriteTest
- ExternalDeltaTableReadWriteTest extend BaseTableReadWriteTest with AdditionalDeltaTests
- ManagedDeltaTableReadWriteTest extend BaseTableReadWriteTest with AdditionalDeltaTests

each test can have additional tests specific for that case.

and any additional tests that is independent of the format at all (like list tables) should be in a test.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: Java doesn't have optional parameter like Scala does.
I can rename this function `setupDeltaTableForCloud` to `setupDeltaTable` with the extra parameter. So we'll have two `setupDeltaTable` functions. Only one of them will be overriden by subclasses.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: To simplify the discussion, let me assume that we want to test three types of tables:
1. managed Delta tables
2. external Delta tables
3. external parquet tables (as an example)

The tests can be classified as 5 (2*2+1) types according to two dimensions:
1. run with external tables or external&managed tables.
2. run with only Delta tables or Delta&Parquet tables.
3. plus the tests run with only managed Delta tables.

There are tests that only run with external tables but they don't care Delta or parquet.
There are tests that only run with Delta but they don't care external or managed.

It's hard to make a class hierarchy like this in Java because Java doesn't allow multiple inheritance of abstract classes.

yili-db: An easier approach is to stick with the current structure of:
base, delta_external(base), delta_managed(base)
then later it be further refactored into:
base, external(base), delta(external), parquet(external), delta_managed(base)

Now the problem is that some of the test in external class can not be run on parquet tables. This is an easier problem to solve with `@EnabledIf` on the individual Delta only tests, instead of using class inheritance.

And when this is done it would introduce additional test coverage. I want to keep this PR a pure refactor PR. This can be done in a follow up PR.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+    sql("UPDATE %s SET i = 2 WHERE i = 1", t1);
+    List<Row> rows = sql("SELECT * FROM %s", t1);
+    assertThat(rows).hasSize(1);
+    assertThat(rows.get(0).getInt(0)).isEqualTo(2);
+  }
+
+  @Test
+  public void testShowTables() throws ApiException, IOException {
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG);
```
tdas, line 151: this not tied to any particular type of table .. isnt it? should this be in the BaseTableReadWriteTest that is meant to be replicated for multiple types of table?

connectors/spark/src/test/java/io/unitycatalog/spark/TableReadWriteTest.java
yili-db: No. It's just a residual from the original PR. The test actually uses UCSingleCatalog.LOAD_DELTA_CATALOG().set(false); but this is forgotten to remove.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -0,0 +1,187 @@
...
+  @MethodSource("cloudParameters")
+  public void testUpdateDeltaTable(String scheme, boolean renewCredEnabled)
+      throws ApiException, IOException {
+    session = createSparkSessionWithCatalogs(renewCredEnabled, SPARK_CATALOG);
+
+    String t1 = setupDeltaTable(scheme, SPARK_CATALOG, DELTA_TABLE, List.of());
+    sql("INSERT INTO %s SELECT 1, 'a'", t1);
+
+    sql("UPDATE %s SET i = 2 WHERE i = 1", t1);
+    List<Row> rows = sql("SELECT * FROM %s", t1);
```
tdas: given that each test's setup is heavy .. we could combine the Update/Delete/Merge together into one test. especially since its going to be a only for Delta.

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
yili-db: The reason is that this function is called by `setupDeltaTable` which is shared by external and managed tables. In order to make the interface simpler and uniform between external and managed tables, it can not let caller to decide which test bucket to use because most likely the caller is testing both external and managed tables.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -0,0 +1,187 @@
...
+    testTableReadWrite(t2);
+
+    String t3 = setupDeltaTable(CATALOG_NAME, DELTA_TABLE, List.of());
+    testTableReadWrite(t3);
+
+    String t4 = setupDeltaTable(CATALOG_NAME, DELTA_TABLE_PARTITIONED, Arrays.asList("s"));
+    testTableReadWrite(t4);
+  }
+
+  private void validateTimeTravelDeltaTable(List<Row> rows) {
```
tdas: nothing in this method is specific to time travel.
i think you should make a simple method that take 2 params, sql query (that produces integers) and expected sequence of integers

then even testTableReadWrite can use it. common reusable code. and easier to read when used as the query and expected result is in the same line.. unlike this functions usage.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
tdas: let me think about it a little. i understand java's limitation makes this hard.

server/src/test/java/io/unitycatalog/server/base/BaseServerTest.java
yili-db: This IS part of the abstraction. This file is `BaseServerTest.java` and it handles most of the abstraction already. Most individual tests based on this base class don't need to care about cloud anymore except a few external table tests.
If you mean "globally", then it's out of scope of this PR.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
tdas: okay.. i actually like this - base, external(base), delta(external), parquet(external), delta_managed(base)

but for that isnt better to name all the methods to be independent of delta?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: I didn't actually change this test. It's still doing exactly the same thing as before this PR.
But from what I can see it's reasonable. This test wants to check that "DELETE FROM" is actually working. Then it must make sure there's something in the table before the deletion. `testTableReadWrite` is a convenient and lightweight way to insert some data and verify that data is indeed inserted into the table.
Added one line comment here.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: Right it's not tied to any particular type of table. I was just thinking maybe it's useful to test against both external and managed table. Just in case if there's a bug to prevent external or managed table from being shown in the `SHOW TABLES` result?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: Yes. **Most** test methods can be renamed independent of Delta, unless they are actually only useful for testing Delta tables.
Still, I would prefer to do it one step at a time. If we do the entire refactor in a single PR, it will look completely different and it's hard to tell what has changed. Also it will be time consuming and the `UCSingleCatalog` change is still stacked on top of this PR.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
openinx: Okay, got. it's from oss-delta.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: > okay.. i actually like this - base, external(base), delta(external), parquet(external), delta_managed(base)
> 
> but for that isnt better to name all the methods to be independent of delta?

Sending out https://github.com/unitycatalog/unitycatalog/pull/1244 to implement this.

PR #1225, opened 2025-11-19
Change UCSingleCatalog to support creation of managed Delta tables with UC coordinated commits enabled

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1225/files) to review incremental changes.
- [**stack/uc_spark_managed**](https://github.com/unitycatalog/unitycatalog/pull/1225) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1225/files)]
  - [stack/table_test_refactor_2](https://github.com/unitycatalog/unitycatalog/pull/1244) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1244/files/a7abd51801da3312c234fddf906bd16ad68543b0..e439e3231adef3ac8a3e1e435f4aa4ebe16b16b0)]

---------
**PR Checklist**

**Description of changes**

Change UCSingleCatalog to support creation of managed Delta tables with UC coordinated commits enabled.
Also added integration test to exercise managed tables.

**Related issue**

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
@@ -9,13 +9,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.connector.catalog.TableCatalog;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
 /**
@@ -29,25 +27,8 @@ import org.junit.jupiter.params.provider.MethodSource;
  * <p>This test needs to start server with a single managed root location on a single emulated cloud
  * so it can not test all emulated clouds yet.
  */
-public class ManagedTableReadWriteTest extends BaseTableReadWriteTest {
+public abstract class ManagedTableReadWriteTest extends BaseTableReadWriteTest {
   private static final String DELTA_TABLE = "test_delta";
-
-  /**
-   * This function provides a set of test parameters that cloud-aware tests should run for this
-   * class.
-   *
-   * @return A stream of Arguments.of(String scheme, boolean renewCredEnabled)
-   */
-  protected static Stream<Arguments> cloudParameters() {
-    // Right now this test suite only support testing with s3. In the future we'll expand this
-    // test to all supported clouds when UC supports setting up multiple managed storage locations.
-    return Stream.of(Arguments.of("s3", false), Arguments.of("s3", true));
-  }
-
-  @Override
-  protected String managedStorageCloudScheme() {
-    return "s3";
-  }
 
   @Test
   public void testCreateManagedTableErrors() {
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
@@ -61,6 +61,7 @@ public class ExternalTableReadWriteTest extends BaseTableReadWriteTest {
    */
   protected static Stream<Arguments> cloudParameters() {
     return Stream.of(
+        Arguments.of("file", false),
         Arguments.of("s3", false),
         Arguments.of("s3", true),
         Arguments.of("gs", false),
@@ -126,13 +127,13 @@ public class ExternalTableReadWriteTest extends BaseTableReadWriteTest {
       throws ApiException, IOException {
     session = createSparkSessionWithCatalogs(renewCredEnabled, SPARK_CATALOG);
 
-    String loc1 = scheme + "://test-bucket0" + generateTableLocation(SPARK_CATALOG, PARQUET_TABLE);
+    String loc1 = cloudPathPrefix(scheme) + generateTableLocation(SPARK_CATALOG, PARQUET_TABLE);
     setupExternalParquetTable(PARQUET_TABLE, loc1, new ArrayList<>(0));
     String t1 = SPARK_CATALOG + "." + SCHEMA_NAME + "." + PARQUET_TABLE;
     testTableReadWrite(t1);
 
     String loc2 =
-        scheme + "://test-bucket1" + generateTableLocation(SPARK_CATALOG, ANOTHER_PARQUET_TABLE);
+        cloudPathPrefix(scheme) + generateTableLocation(SPARK_CATALOG, ANOTHER_PARQUET_TABLE);
     setupExternalParquetTable(ANOTHER_PARQUET_TABLE, loc2, new ArrayList<>(0));
     String t2 = SPARK_CATALOG + "." + SCHEMA_NAME + "." + ANOTHER_PARQUET_TABLE;
     testTableReadWrite(t2);
@@ -146,13 +147,13 @@ public class ExternalTableReadWriteTest extends BaseTableReadWriteTest {
       throws IOException {
     session = createSparkSessionWithCatalogs(renewCredEnabled, SPARK_CATALOG, CATALOG_NAME);
 
-    String loc1 = scheme + "://test-bucket0" + generateTableLocation(SPARK_CATALOG, DELTA_TABLE);
+    String loc1 = cloudPathPrefix(scheme) + generateTableLocation(SPARK_CATALOG, DELTA_TABLE);
     setupDeltaTableLocation(loc1, new ArrayList<>(0));
     String t1 = SPARK_CATALOG + "." + SCHEMA_NAME + "." + DELTA_TABLE;
     sql("CREATE TABLE %s USING delta LOCATION '%s'", t1, loc1);
     testTableReadWrite(t1);
 
-    String loc2 = scheme + "://test-bucket1" + generateTableLocation(CATALOG_NAME, DELTA_TABLE);
+    String loc2 = cloudPathPrefix(scheme) + generateTableLocation(CATALOG_NAME, DELTA_TABLE);
     setupDeltaTableLocation(loc2, new ArrayList<>(0));
     String t2 = CATALOG_NAME + "." + SCHEMA_NAME + "." + DELTA_TABLE;
     sql("CREATE TABLE %s USING delta LOCATION '%s'", t2, loc2);
@@ -162,7 +163,7 @@ public class ExternalTableReadWriteTest extends BaseTableReadWriteTest {
 
     // Path that does not exist
     String loc3 =
-        scheme + "://test-bucket1" + generateTableLocation(CATALOG_NAME, ANOTHER_DELTA_TABLE);
+        cloudPathPrefix(scheme) + generateTableLocation(CATALOG_NAME, ANOTHER_DELTA_TABLE);
     String t3 = CATALOG_NAME + "." + SCHEMA_NAME + "." + ANOTHER_DELTA_TABLE;
     sql("CREATE TABLE %s(i INT) USING delta LOCATION '%s'", t3, loc3);
     List<Row> rows = session.table(t3).collectAsList();
@@ -295,11 +296,15 @@ public class ExternalTableReadWriteTest extends BaseTableReadWriteTest {
     return String.format("%s://test-bucket%d", scheme, bucketIndex);
   }
 
+  protected String cloudPathPrefix(String cloudScheme) {
+    return cloudScheme.equals("file") ? "" : testBucket(cloudScheme);
+  }
+
   @Override
   protected String setupDeltaTable(
       String cloudScheme, String catalogName, String tableName, List<String> partitionColumns)
       throws IOException, ApiException {
-    String cloudPrefix = cloudScheme.equals("file") ? "" : testBucket(cloudScheme);
+    String cloudPrefix = cloudPathPrefix(cloudScheme);
     String location = cloudPrefix + generateTableLocation(catalogName, tableName);
     setupDeltaTableLocation(location, partitionColumns);
     setupTables(catalogName, tableName, DataSourceFormat.DELTA, location, partitionColumns, false);
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/LocalManagedTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/LocalManagedTableReadWriteTest.java
@@ -1,0 +1,24 @@
+package io.unitycatalog.spark;
+
+import java.util.stream.Stream;
+import org.junit.jupiter.params.provider.Arguments;
+
+/**
+ * This test suite starts UC server with managed storage root on local path and exercise the tests.
+ */
+public class LocalManagedTableReadWriteTest extends ManagedTableReadWriteTest {
+  /**
+   * This function provides a set of test parameters that cloud-aware tests should run for this
+   * class.
+   *
+   * @return A stream of Arguments.of(String scheme, boolean renewCredEnabled)
+   */
+  protected static Stream<Arguments> cloudParameters() {
+    return Stream.of(Arguments.of("file", false));
+  }
+
+  @Override
+  protected String managedStorageCloudScheme() {
+    return "file";
+  }
+}
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/S3ManagedTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/S3ManagedTableReadWriteTest.java
@@ -1,0 +1,25 @@
+package io.unitycatalog.spark;
+
+import java.util.stream.Stream;
+import org.junit.jupiter.params.provider.Arguments;
+
+/**
+ * This test suite starts UC server with managed storage root on emulated S3 path and exercise the
+ * tests.
+ */
+public class S3ManagedTableReadWriteTest extends ManagedTableReadWriteTest {
+  /**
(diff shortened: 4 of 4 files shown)


connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class ManagedTableReadWriteTest extends BaseTableReadWriteTest {
```
yili-db: This test runs all test cases in BaseTableReadWriteTest, plus two extra tests for managed tables only.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -178,8 +209,25 @@ class UCSingleCatalog
 }
 
 object UCSingleCatalog {
+  // These properties are required to have UC as Delta commit coordinator
+  val UC_TABLE_ID_KEY = "ucTableId"
```
yili-db: Property rename isn't part of this PR yet.

openinx: Hi @yili-db , Since we've already moved all the hadoop properties in the `UChadoopConf`, for this non-hadoop properties, but UC properties, maybe it's also good to have them into a separate `CatalogProps` java class.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -94,17 +97,51 @@ class UCSingleCatalog
...
     if (!hasExternalClause && !hasLocationClause && !isPathTable) {
+      // Check that caller shouldn't set some properties
+      List(UCSingleCatalog.UC_TABLE_ID_KEY, TableCatalog.PROP_IS_MANAGED_LOCATION)
+        .filter(properties.containsKey(_))
+        .foreach(p => throw new ApiException(s"Cannot specify property $p."))
+      // It's OK to set delta.feature.catalogOwned-preview if it matches exactly what we need
+      Option(properties.get(UCSingleCatalog.CATALOG_OWNED_KEY))
+        .filter(_ != UCSingleCatalog.CATALOG_OWNED_VALUE)
+        .foreach(v => throw new ApiException(
+          s"Cannot specify property ${UCSingleCatalog.CATALOG_OWNED_KEY}=$v."))
```
openinx: Could we make this error message to be more clear ?  Seems like we are not allowed to set the `delta.feature.catalogOwned-preview` to be other value except `supported`, for the managed table creation.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -94,17 +97,51 @@ class UCSingleCatalog
...
+
+      // Get staging table location and table id from UC
+      val createStagingTable = new CreateStagingTable()
+        .catalogName(name())
+        .schemaName(ident.namespace().head)
```
openinx, line 126: When will the namespace have multiple elements ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
tdas: +1 to @openinx

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -178,8 +209,25 @@ class UCSingleCatalog
 }
 
 object UCSingleCatalog {
+  // These properties are required to have UC as Delta commit coordinator
+  val UC_TABLE_ID_KEY = "ucTableId"
+  val CATALOG_OWNED_KEY = "delta.feature.catalogOwned-preview"
```
tdas: lets start using CATALOG_MANAGED_* to optimize for the what the name is going to be.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
yili-db: Changed the error message and comment.
In general, we do not expect this table property to be set at all as we always set it automatically to `supported`.
This `.filter` allowed setting it to `supported` because some document asks setting it already. This is just to avoid unnecessary errors.
But after all, caller should not set it.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
yili-db: Added `UCTableProperties.java`

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
yili-db: Actually I don't know when it will be multiple. But in this file only single level is supported. There's a check in UCProxy below in this file. I just modified this function to do that check as well.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -88,23 +91,64 @@ class UCSingleCatalog
...
+
       val newProps = new util.HashMap[String, String]
       newProps.putAll(properties)
-      // TODO: here we use a fake location for managed table, we should generate table location
-      //       properly when Unity Catalog supports creating managed table.
-      newProps.put(TableCatalog.PROP_LOCATION, properties.get("__FAKE_PATH__"))
+      newProps.put(TableCatalog.PROP_LOCATION, stagingTableInfo.getStagingLocation)
+      // Sets both the new and old table ID property while it's being renamed.
+      newProps.put(UCTableProperties.UC_TABLE_ID_KEY, stagingTableInfo.getId)
+      newProps.put(UCTableProperties.UC_TABLE_ID_KEY_OLD, stagingTableInfo.getId)
```
yili-db, line 137: Setting both the old and new table id property for now. The old one will be removed in next release after 0.3.1.
The same thing can not be done for `CATALOG_MANAGED_KEY` though as Delta checks for validity of table feature property.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
tdas: Actually it is currently expected that in spark create table, users will set the catalogManaged table feature directly to enable the feature. So it's expected that it maybe set. 

TableId is not expected to be set.

Why are we not allowing them to pass through they are set to expected values??

tdas: Actually it is currently expected that in spark create table, users will set the catalogManaged table feature directly to enable the feature. So it's expected that it maybe set. 

TableId is not expected to be set.

Why are we not allowing them to pass through they are set to expected values??

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -88,23 +91,64 @@ class UCSingleCatalog
...
+      val stagingTableInfo = tablesApi.createStagingTable(createStagingTable)
+      val stagingLocation = stagingTableInfo.getStagingLocation
+      val stagingTableId = stagingTableInfo.getId
+
       val newProps = new util.HashMap[String, String]
       newProps.putAll(properties)
-      // TODO: here we use a fake location for managed table, we should generate table location
-      //       properly when Unity Catalog supports creating managed table.
-      newProps.put(TableCatalog.PROP_LOCATION, properties.get("__FAKE_PATH__"))
+      newProps.put(TableCatalog.PROP_LOCATION, stagingTableInfo.getStagingLocation)
```
tdas, line 134: Is this a special property define by spark as the target location?

connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
```diff
@@ -0,0 +1,158 @@
...
+              .contains("delta.feature.catalogOwned-preview=supported");
+          assertThat(describeResult.get("Table Properties")).contains("ucTableId=");
+
+          // Make sure the table has commit under /_delta_log/_staged_commits
+          assertThat(describeResult).containsKey("Location");
+          String location = describeResult.get("Location");
+          URI uri = new URI(location);
+          assertThat(uri.getScheme()).isEqualTo(scheme);
+          String path = uri.getPath();
+          String stagedCommitPath = path + "/_delta_log/_staged_commits";
```
tdas: Why do we need to test this?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
yili-db: It's needed by Delta. In `AbstractDeltaCatalog.createDeltaTable`:
```
    val location = if (isByPath) {
      Option(ident.name())
    } else {
      Option(allTableProperties.get("location"))
    }
```
It's defined by Spark as:
```
  /**
   * A reserved property to specify the location of the table. The files of the table
   * should be under this location. The location is a Hadoop Path string.
   */
  String PROP_LOCATION = "location";
```

connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
yili-db: This is an effective way to make sure that the table is indeed using CC. Delta should write json version files under this directory before publishing to the upper `/_delta_log/`

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
yili-db: This error only pops up if it is set to a **different** value than expected value.
If it's indeed set as the expected value, it will "pass through" (the following line will overwrite it with the expected value, which is the same value, anyway)

connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
tdas: this feels very invasive. ideally we want to write blackbox tests that do not require understanding the internals of the box. so while this may be a nice trick, it assumes that CC will always use _staged_commits. Tomorrow if this name changes.. the test will fail.. right? 

it might be cleaner to just check whether the corresponding table feature has been enabled or not. it should be in the property. That is a clear interface that delta maintains, enabled table features show up in the properties that delta itself use in their code (prod and test) as a way to verify if CC is enabled. So lets respect that abstraction?

yili-db: Sounds good. I'll remove this section then.
A few lines above I have this check. Is it good enough?
```
      assertThat(describeResult.get("Table Properties"))
              .contains("delta.feature.catalogOwned-preview=supported");
```

tdas: yeah that looks good. but shouldnt we make it more robust by testing that either of the two feature names should exist?

similarly i see the property is ucTableId embedded in the line below - https://github.com/unitycatalog/unitycatalog/pull/1225/files
shouldnt we test for either property?

yili-db: Now testing both properties.
However only the old feature name can be actually set before Delta does the rename and a release.

yili-db: Also removed the reference of the raw strings of these property names. Changed them to refer to the defined constants in UCTableProperties.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
tdas: @yili-db do we need to fix this for backticks? is the new backtick test testing this code path?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -182,6 +225,17 @@ object UCSingleCatalog {
   val LOAD_DELTA_CATALOG = ThreadLocal.withInitial[Boolean](() => true)
   val DELTA_CATALOG_LOADED = ThreadLocal.withInitial[Boolean](() => false)
 
+  def setCredentialProps(props: util.HashMap[String, String],
+                         credentialProps: util.Map[String, String]): Unit = {
+    props.putAll(credentialProps)
+    // TODO: Delta requires the options to be set twice in the properties, with and without the
+    //       `option.` prefix. We should revisit this in Delta.
+    val prefix = TableCatalog.OPTION_PREFIX
```
tdas, line 233: Is it a Delta requirement or Spark requirement?

tdas: not a blocker for this PR, since this existed.. just curious if you understand why this is the case.

yili-db: No I have no idea about this.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
yili-db: No we don't need any fix here.
But there isn't test coverage for this code path. I'll quickly add one.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
tdas: Okay .. so this is the way Spark wants its connector to deal with locations.

connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
```diff
@@ -0,0 +1,168 @@
...
+            () ->
+                sql(
+                    "CREATE TABLE %s(name STRING) USING delta " + "TBLPROPERTIES ('%s' = 'false')",
+                    fullTableName, TableCatalog.PROP_IS_MANAGED_LOCATION))
+        .hasMessageContaining("is_managed_location");
+  }
+
+  @ParameterizedTest
+  @MethodSource("cloudParameters")
+  public void testCreateManagedDeltaTable(String scheme, boolean renewCredEnabled)
```
tdas: add a note saying why we need to test create of managed tables separately and not in the common TableReadWriteTest. Basically, we need to start the server with different root locations and do additional managed-table specific tests

connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
```diff
@@ -0,0 +1,168 @@
...
+                    fullTableName, UCTableProperties.UC_TABLE_ID_KEY_OLD))
+        .hasMessageContaining(UCTableProperties.UC_TABLE_ID_KEY_OLD);
+    assertThatThrownBy(
+            () ->
+                sql(
+                    "CREATE TABLE %s(name STRING) USING delta " + "TBLPROPERTIES ('%s' = 'false')",
+                    fullTableName, TableCatalog.PROP_IS_MANAGED_LOCATION))
+        .hasMessageContaining("is_managed_location");
+  }
+
```
tdas, line 73: is create managed table on local file system already tested via common test in TableReadWriteTest?

connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
yili-db: Added class doc.

connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
yili-db: No. This class can not do that as it starts the server with "s3" as managed root location.
I can further subclass this class into `S3ManagedTableReadWriteTest` and `FileManagedTableReadWriteTest` (which just need to override one function `managedStorageCloudScheme`) if that's desired. But in reality there shouldn't be any difference.

tdas: why dont we generalize this parameterized test to cover all 4 - local, s3, azure, gcs?

yili-db: For external table test, we can do that. But the parameterized tests are meant to test the credential vending and renew of different clouds. Testing local with those tests may not bring extra coverage.
For managed table, this has to be done by different sub classes like `S3ManagedTableReadWriteTest` with minimum override. Because we need to start UC server with the correct managed location of the chosen cloud. But when it starts there doesn't seem to be a **clean way** to know the parameter of the test that is to be run.

tdas (approved):
LGTM.

tdas:
- #1255

PR #1230, opened 2025-11-19
Allow staging table creation with the same name.

yili-db:
**PR Checklist**

**Description of changes**

This is to fix a bug that UC didn't allow a 2nd staging table to be created using the same name. However this is actually a valid use case. When the table creation failed after the staging table is created but before the table creation, the dangling staging table would prevent retrying effort to create the table using the same name again. We have to allow the staging table to be created again using the same name to get it through. The new staging table will have new ID and location so it won't be a problem anyway.

**Related issue**

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
@@ -16,11 +16,8 @@ import java.util.UUID;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.query.Query;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class StagingTableRepository {
-  private static final Logger LOGGER = LoggerFactory.getLogger(StagingTableRepository.class);
   private final SessionFactory sessionFactory;
   private final Repositories repositories;
   private final ServerProperties serverProperties;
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/unitycatalog/server/persist/dao/StagingTableDAO.java
```diff
@@ -21,7 +21,6 @@
 @Table(
     name = "uc_staging_tables",
     indexes = {
-      @Index(name = "idx_name", columnList = "name"),
```
yili-db, line 24: Name no longer matters. We no longer query by name. It's OK to have multiple entries with the same name too.

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
```diff
@@ -32,15 +29,6 @@ public StagingTableRepository(
     this.serverProperties = serverProperties;
   }
 
-  private StagingTableDAO findBySchemaIdAndName(Session session, UUID schemaId, String name) {
```
yili-db, line 35: No longer need to query by name.

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
```diff
@@ -50,18 +38,15 @@ private StagingTableDAO findByStagingLocation(Session session, String stagingLoc
 
   private void validateIfAlreadyExists(
       Session session, UUID schemaId, String tableName, String stagingLocation) {
-    // check if staging table or table by the same name already exists
-    // Also ensure that no staging table exists at the same location
-    StagingTableDAO existingStagingTable = findBySchemaIdAndName(session, schemaId, tableName);
```
yili-db, line 55: This is the actual change. Skip searching for staging table by name. Allow duplicate names to be used again.

Issue #1218, opened 2025-11-19
Add a real production OAuth integration tests.


openinx:
This is a sub task of this issue https://github.com/unitycatalog/unitycatalog/issues/1200

Finally,  we need to add an integration test which just run agaist the real UC server and OAuth server, so that we can be confident that all things are worked as expected in the real production environment. 

I will try to create a pull request for this issue.

openinx:
Added a PR to address this issue, and I used two different test command to test, all seems good:

* Test the Personal Access Token (PAT) integration test

```bash
export CATALOG_URI=$UC_URI
export CATALOG_AUTH_TOKEN=$UC_TOKEN
export CATALOG_NAME=main
export S3_BASE_LOCATION=$UC_S3_LOCATION
export CREDENTIAL_RENEWAL_TEST_DURATION_SECONDS=200
SBT_OPTS="-Xmx8G -XX:+UseG1GC" \
./build/sbt \
"integrationTests/testOnly io.unitycatalog.integrationtests.SparkCredentialRenewalTest"
```

* Test the OAuth integration tests.

```bash
export CATALOG_URI=$UC_URI
export CATALOG_OAUTH_URI=$OAUTH_URI
export CATALOG_OAUTH_CLIENT_ID=$OAUTH_CLIENT_ID
export CATALOG_OAUTH_CLIENT_SECRET=$OAUTH_CLIENT_SECRET
export CATALOG_NAME=main
export S3_BASE_LOCATION=$UC_S3_LOCATION
export CREDENTIAL_RENEWAL_TEST_DURATION_SECONDS=200
SBT_OPTS="-Xmx8G -XX:+UseG1GC" \
./build/sbt \
"integrationTests/testOnly io.unitycatalog.integrationtests.SparkCredentialRenewalTest"
```

openinx:
If people want to run all the integration tests via one command, then it should be:

```
export CATALOG_NAME=main
export SCHEMA_NAME=demo_zh
export CATALOG_URI=$UC_URI

# Enable PAT test
export CATALOG_AUTH_TOKEN=$UC_TOKEN

# Enable OAuth test
export CATALOG_OAUTH_URI=$OAUTH_URI
export CATALOG_OAUTH_CLIENT_ID=$OAUTH_CLIENT_ID
export CATALOG_OAUTH_CLIENT_SECRET=$OAUTH_CLIENT_SECRET

# Enable S3 test.
[...]

PR #1239, opened 2025-11-23
Add OAuth real production integration test

openinx:
**PR Checklist**

**Description of changes**

Related issue is: https://github.com/unitycatalog/unitycatalog/issues/1218

### DIFF
--- a/integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
+++ b/integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
@@ -202,8 +202,9 @@ public class SparkCredentialRenewalTest {
    * contains:
    *
    * <ul>
-   *   <li><b>managedTable</b> - true indicate covering the managed table long-running test,
-   *       otherwise the external table long-running test.
+   *   <li><b>tableType</b> - {@link TestUtils#MANAGED_TABLE_TYPE} or {@link
+   *       TestUtils#EXTERNAL_TABLE_TYPE}, which indicates covering managed table tests or external
+   *       table tests.
    *   <li><b>baseLocation</b> – The base cloud storage path used during testing. External tables
    *       created in the tests will be placed under this location.
    *   <li><b>catalogProps</b> – A {@code Map<String, String>} of catalog configuration properties.


integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -51,21 +60,38 @@ public class SparkCredentialRenewalTest {
 
   // Define the CREDENTIAL_RENEWAL_TEST_DURATION_SECONDS environment variable, which control how
   // long will the test run. 90 minutes by default.
-  private static final long DURATION_SECONDS = envAsLong(PREFIX + "DURATION_SECONDS", 5600L);
+  private static final long DURATION_SECONDS = envAsLong(PREFIX + "DURATION_SECONDS", 5400L);
+
+  // Define the CREDENTIAL_RENEWAL_TEST_MANAGED_TABLE_ENABLED environment variable, which decides
```
tdas: this seems wrong.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -51,21 +60,38 @@ public class SparkCredentialRenewalTest {
...
+  // if enable the managed table or not.
+  private static final boolean MANAGED_TABLE_ENABLED =
```
tdas: do we need to explicitly enable this with env variable? why not always enable it?

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -51,21 +60,38 @@ public class SparkCredentialRenewalTest {
...
+      envAsBoolean(PREFIX + "MANAGED_TABLE_ENABLED", true);
+
+  // Define managed table name.
+  private static final String MANAGED_SRC_TABLE =
```
tdas, line 69: why do we need different names for managed and external. why can we reuse names by dropping the table after each test?

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -74,37 +100,59 @@ public static void beforeAll() {
...
-
-    sql("CREATE SCHEMA %s.%s", CATALOG_NAME, SCHEMA_NAME);
-    sql(
-        "CREATE TABLE %s (id INT) USING delta LOCATION '%s/%s' PARTITIONED BY (partition INT)",
-        SRC_TABLE, S3_BASE_LOCATION, UUID.randomUUID());
-    sql(
-        "CREATE TABLE %s (id INT) USING delta LOCATION '%s/%s'",
-        DST_TABLE, S3_BASE_LOCATION, UUID.randomUUID());
+            .config("spark.sql.adaptive.enabled", "false")
+            .config("spark.sql.adaptive.coalescePartitions.enabled", "false");
```
tdas, line 103: why do you need this. drop a comment?

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -127,14 +176,71 @@ public void testLongRunningJob() {
...
   private static List<Row> sql(String statement, Object... args) {
     return spark.sql(String.format(statement, args)).collectAsList();
   }
+
+  /**
+   * Generates argument combinations for long-running tests. Each {@link Arguments} instance
+   * contains:
+   *
+   * <ul>
+   *   <li><b>managedTable</b> - true indicate covering the managed table long-running test,
```
tdas: can you name this isManagedTable

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -74,37 +100,59 @@ public static void beforeAll() {
...
+      spark = null;
+    }
   }
 
-  @Test
-  public void testLongRunningJob() {
+  @ParameterizedTest
+  @MethodSource("credentialRenewalArguments")
+  public void testLongRunningJob(
+      boolean managedTable, String baseLocation, Map<String, String> catalogProps) {
```
tdas: isManagedTable

more obvious that its a boolean.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -74,37 +100,59 @@ public static void beforeAll() {
...
+    // Use the UCSingleCatalog by default.
+    builder.config(testCatalogKey, UCSingleCatalog.class.getName());
+    // Set all the specified catalog props.
+    catalogProps.forEach((k, v) -> builder.config(testCatalogKey + "." + k, v));
+    spark = builder.getOrCreate();
+
+    sql("CREATE SCHEMA IF NOT EXISTS %s.%s", CATALOG_NAME, SCHEMA_NAME);
+    if (managedTable) {
+      // TODO: SQL-managed table creation is not supported yet, so we create the table manually.
+      // Once SQL support is added, use the SQL to create tables.
```
tdas: where are you creating the table?

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
tdas: aftereach function is always deleting the table. so why use different names?

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Thanks for the comment @tdas ,  Could you say it more clearly ? I don't understand your point.  Thanks.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Because I hope we can have a switch to control either optional combination of the tests or just run all combination of the tests. 

Currently,  we have `<managedTable,externalTable>`, `<s3, azure, gcs>`, `<PAT, OAuth>`,  `<credentialRenewalEnabled, credentialsRenewalDisabled>`.  

for the pre-release testing,  we absolutely need to run all the combination.  But if I just want to do some light-weight verfication,  for example just run the `<managedTable>` + `<s3>`  + `<PAT>` + `<credentialRenewalEnabled>`.   Then having this switch will helps me a lot since I don't need to change the code and only specify the env var, then  I can selectively choose the test that I want to run.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: For external tables, it's fine.  But for managed table,  we did not support the table creation that's why we need to create the managed table manually, as I said in this [TODO comment](https://github.com/unitycatalog/unitycatalog/pull/1239/files#diff-a961fbeeb181232588b1a8b7657dff464d114c46faf6b4fd355903115601efe5R115-R116). 

And recently @yili-db  helped to support the managed table creation,  maybe  I can unify this table creation for both external table and managed table now. 

Let me try to update the PR and simplify those now.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Because the Spark AQE will coalesce the small partitions by default,  which is possible for the job run less than the expected time.  And finally have no chance to trigger the credential renewal because the job stops in advance.

openinx: In the core [test](https://github.com/unitycatalog/unitycatalog/pull/1239/files#diff-a961fbeeb181232588b1a8b7657dff464d114c46faf6b4fd355903115601efe5R168-R174), we do require each parititon will execute serially and every one sleep 100seconds.  While if there is any partition coalesce happen,  then we won't have the expected partitions to run and its execution time will be less than the expected time.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Before we supported the managed table,  I just used the following SQL to create those tables in the external UC server. 

```sql
CREATE TABLE main.demo_zh.managedSrcCredRenewal (id INT) USING delta
PARTITIONED BY (partition INT)
TBLPROPERTIES ('delta.feature.catalogOwned-preview' = 'supported');

CREATE TABLE main.demo_zh.managedDstCredRenewal (id INT) USING delta
TBLPROPERTIES ('delta.feature.catalogOwned-preview' = 'supported');
```

But currently, since we've already support the managed table creation, I think maybe I can try to create them directly via the SQL in this tests.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Thanks for the comment,  let me refine them and make them more clear.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Will do,  thanks.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Hi @tdas , I removed this `CREDENTIAL_RENEWAL_TEST_MANAGED_TABLE_ENABLED`, instead  I introduce a more reasonable and general `TABLE_TYPE` environment to specify the `ExternalTable` or `ManagedTable`.  Please see: https://github.com/unitycatalog/unitycatalog/pull/1239/files#diff-3458ce6d27d9e1a8964b23bd83c6758a920ca43308b1ef5ac19123ce3836637dR20-R21

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: I confirmed and talked with @yili-db . Currently,  it still have some problem when creating managed table in external UC server, which will report the error.  So for the short term, we still need to separate the managed table creation and external table creation into two cases:
* For external table ,  we can automatically handle the whole table lifecycle,  from `create` -> `insert` -> `drop`. 
* For managed table, we need to create them manually and then access them in the tests.  Since there are many cases in the test, we don't expect to create them manually for every round.  That's why we separate the external test and managed tests into two table sets. 

Once we fully supported the managed table creation,  I think we can unify the table lifecycle for both.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Done

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Done

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Done

PR #1244, opened 2025-11-24
Refactor the table read/write integration tests.

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1244/files) to review incremental changes.
- [**stack/table_test_refactor_2**](https://github.com/unitycatalog/unitycatalog/pull/1244) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1244/files)]

---------
**PR Checklist**

**Description of changes**

Refactor the table read/write integration tests
Refactor the tests so that most tests are made abstract of managed/external table and table formats. These tests then can run against managed/external tables with different table formats.
This PR is a follow up of https://github.com/unitycatalog/unitycatalog/pull/1223#discussion_r2548113952

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
@@ -34,12 +34,16 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
           UCTableProperties.DELTA_CATALOG_MANAGED_KEY,
           UCTableProperties.DELTA_CATALOG_MANAGED_VALUE);
 
+  /**
+   * This class is used for control various options for table creation during test. The tableFormat
+   * is controlled at the testsuite level by tableFormat().
+   */
   @NoArgsConstructor
   @AllArgsConstructor
   @Accessors(chain = true)
   @Getter
   @Setter
-  protected class SetupTableOptions {
+  protected class TableSetupOptions {
     private String catalogName;
     private String schemaName = SCHEMA_NAME;
     private String tableName;
@@ -47,34 +51,35 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
     private Optional<String> partitionColumn = Optional.empty();
     @With private Optional<Pair<Integer, String>> asSelect = Optional.empty();
     private Optional<String> comment = Optional.empty();
+    private boolean replaceTable = false;
 
-    public SetupTableOptions setCatalogName(String name) {
+    public TableSetupOptions setCatalogName(String name) {
       catalogName = quoteEntityName(name);
       return this;
     }
 
-    public SetupTableOptions setSchemaName(String name) {
+    public TableSetupOptions setSchemaName(String name) {
       schemaName = quoteEntityName(name);
       return this;
     }
 
-    public SetupTableOptions setTableName(String name) {
+    public TableSetupOptions setTableName(String name) {
       tableName = quoteEntityName(name);
       return this;
     }
 
-    public SetupTableOptions setPartitionColumn(String column) {
+    public TableSetupOptions setPartitionColumn(String column) {
       assert List.of("i", "s").contains(column);
       partitionColumn = Optional.of(column);
       return this;
     }
 
-    public SetupTableOptions setAsSelect(int i, String s) {
+    public TableSetupOptions setAsSelect(int i, String s) {
       asSelect = Optional.of(Pair.of(i, s));
       return this;
     }
 
-    public SetupTableOptions setComment(String c) {
+    public TableSetupOptions setComment(String c) {
       comment = Optional.of(c);
       return this;
     }
@@ -102,9 +107,14 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
       return comment.map(c -> String.format("COMMENT '%s'", c)).orElse("");
     }
 
+    public String ddlCommand() {
+      return replaceTable ? "REPLACE" : "CREATE";
+    }
+
     public String createManagedTableSql() {
       return String.format(
-          "CREATE TABLE %s.%s.%s %s USING %s %s %s %s %s",
+          "%s TABLE %s.%s.%s %s USING %s %s %s %s %s",
+          ddlCommand(),
           catalogName,
           schemaName,
           tableName,
@@ -118,7 +128,8 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
 
     public String createExternalTableSql(String location) {
       return String.format(
-          "CREATE TABLE %s.%s.%s%s USING %s %s %s LOCATION '%s' %s",
+          "%s TABLE %s.%s.%s%s USING %s %s %s LOCATION '%s' %s",
+          ddlCommand(),
           catalogName,
           schemaName,
           tableName,
@@ -133,7 +144,8 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
     public String createDeltaPathTableSql(String location) {
       assert testingDelta();
       return String.format(
-          "CREATE TABLE %s.`%s`%s USING %s %s %s %s",
+          "%s TABLE %s.`%s`%s USING %s %s %s %s",
+          ddlCommand(),
           tableFormat(),
           location,
           columnsClause(),
@@ -149,35 +161,54 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
   }
 
   /**
-   * Set up a delta table and returns the table full name. This function is used by tests that
-   * aren't cloud aware. It simply uses the same cloud as configured for managed storage.
+   * Set up a table and returns the table full name. This function is used by tests that aren't
+   * cloud aware. It simply uses the same cloud as configured for managed storage. The table format
+   * is controlled by tableFormat().
    */
   protected final String setupTable(String catalogName, String tableName) {
-    return setupTable(new SetupTableOptions().setCatalogName(catalogName).setTableName(tableName));
+    return setupTable(new TableSetupOptions().setCatalogName(catalogName).setTableName(tableName));
   }
 
-  /** Set up a delta table with storage on emulated cloud and returns the table full name. */
+  /**
+   * Set up a table with storage on emulated cloud and returns the table full name. The table format
+   * is controlled by tableFormat().
+   */
   protected final String setupTable(String cloudScheme, String catalogName, String tableName) {
     return setupTable(
-        new SetupTableOptions()
+        new TableSetupOptions()
             .setCloudScheme(cloudScheme)
             .setCatalogName(catalogName)
             .setTableName(tableName));
   }
 
   /**
-   * Set up a delta table with storage on emulated cloud and returns the table full name. Subclasses
-   * need to override it accordingly.
+   * Set up a table with storage on emulated cloud and returns the table full name. The table format
+   * is controlled by tableFormat(). Subclasses need to override this function accordingly.
    */
-  protected abstract String setupTable(SetupTableOptions options);
+  protected abstract String setupTable(TableSetupOptions options);
 
-  /** The table format to be tested */
+  /**
+   * The table format to be tested. Subclasses override this function to test different table
+   * formats.
+   */
   protected abstract String tableFormat();
 
   protected final boolean testingDelta() {
     return tableFormat().equalsIgnoreCase("DELTA");
   }
 
+  /**
+   * This test creates table in different ways:
+   *
+   * <ul>
+   *   <li>in both SPARK_CATALOG and CATALOG_NAME
+   *   <li>with and without partition columns
+   *   <li>with and without CTAS
+   *   <li>using UPDATE TABLE and CREATE TABLE
+   * </ul>
+   *
+   * For all the 16 (2^4) tables it does a simple read and write test to make sure they all work.
+   */
   @Test
   public void testTableCreateReadWrite() {
     // Test both `spark_catalog` and other catalog names.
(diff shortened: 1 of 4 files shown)


connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -70,194 +55,86 @@ protected static Stream<Arguments> cloudParameters() {
...
-            .config(catalogConf + "." + OptionsUtil.TOKEN, serverConfig.getAuthToken())
-            .config(catalogConf + "." + OptionsUtil.WAREHOUSE, CATALOG_NAME);
-    session = builder.getOrCreate();
-    setupExternalParquetTable(PARQUET_TABLE, new ArrayList<>(0));
-    testTableReadWrite("spark_catalog." + SCHEMA_NAME + "." + PARQUET_TABLE);
-    assertThat(UCSingleCatalog.DELTA_CATALOG_LOADED().get()).isEqualTo(false);
-  }
-
-  @Test
-  public void testParquetReadWrite() throws IOException, ApiException {
```
yili-db, line 95: covered by testTableCreateReadWrite in BaseTableReadWriteTest.java

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -70,194 +55,86 @@ protected static Stream<Arguments> cloudParameters() {
...
-    // Test CTAS
-    String path2 = new File(dataDir, "test_delta_path2").getCanonicalPath();
-    String tableName2 = String.format("delta.`%s`", path2);
-    sql("CREATE TABLE %s USING delta AS SELECT 1 AS i", tableName2);
-    validateRows(sql("SELECT * FROM %s", tableName2), 1);
-  }
-
-  @ParameterizedTest
-  @MethodSource("cloudParameters")
-  public void testCredentialParquet(String scheme, boolean renewCredEnabled)
```
yili-db, line 126: covered by testCredentialReadWrite in BaseTableReadWriteTest.java

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -70,194 +55,86 @@ protected static Stream<Arguments> cloudParameters() {
...
-    setupExternalParquetTable(ANOTHER_PARQUET_TABLE, loc2, new ArrayList<>(0));
-    String t2 = SPARK_CATALOG + "." + SCHEMA_NAME + "." + ANOTHER_PARQUET_TABLE;
-    testTableReadWrite(t2);
-
-    validateRows(sql("SELECT l.i FROM %s l JOIN %s r ON l.i = r.i", t1, t2), 1);
-  }
-
-  @ParameterizedTest
-  @MethodSource("cloudParameters")
-  public void testCredentialCreateDeltaTable(String scheme, boolean renewCredEnabled)
```
yili-db, line 146: covered by testCredentialReadWrite in BaseTableReadWriteTest.java

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -70,194 +55,86 @@ protected static Stream<Arguments> cloudParameters() {
...
+                Pair.of("i", DataTypes.IntegerType),
+                Pair.of("s", DataTypes.StringType));
+          }
+        }
+      }
     }
   }
 
-  @Test
-  public void testCreateExternalDeltaTable() throws ApiException, IOException {
```
yili-db, line 215: This is also covered by testCreateExternalTable

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -70,194 +55,86 @@ protected static Stream<Arguments> cloudParameters() {
...
-    // Spark only allow `spark_catalog` to return built-in file source tables.
-    setupExternalParquetTable(PARQUET_TABLE, new ArrayList<>(0));
-    testTableReadWrite(SPARK_CATALOG + "." + SCHEMA_NAME + "." + PARQUET_TABLE);
-
-    setupExternalParquetTable(PARQUET_TABLE_PARTITIONED, List.of("s"));
-    testTableReadWrite(SPARK_CATALOG + "." + SCHEMA_NAME + "." + PARQUET_TABLE_PARTITIONED);
-  }
-
-  @Test
-  public void testDeltaPathTable() throws IOException {
```
yili-db, line 106: Moved to DeltaExternalTableReadWriteTest.java

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -6,58 +6,187 @@
...
    * Set up a delta table with storage on emulated cloud and returns the table full name. Subclasses
    * need to override it accordingly.
    */
-  protected abstract String setupDeltaTable(
-      String cloudScheme, String catalogName, String tableName, List<String> partitionColumns)
-      throws IOException, ApiException;
+  protected abstract String setupTable(SetupTableOptions options);
+
+  /** The table format to be tested */
+  protected abstract String tableFormat();
```
yili-db, line 194: This is where to specify table format to test.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -110,67 +252,57 @@ public void testTimeTravelDeltaTable() throws ApiException, IOException {
 
   @ParameterizedTest
   @MethodSource("cloudParameters")
-  public void testCredentialDelta(String scheme, boolean renewCredEnabled)
-      throws ApiException, IOException {
+  public void testTableOperations(String scheme, boolean renewCredEnabled) {
```
yili-db, line 320: This test is a umbrella test that covers CREATE, INSERT, SELECT, JOIN, MERGE INTO, DELETE FROM, UPDATE, SHOW TABLES, DROP TABLE

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaExternalTableReadWriteTest.java
```diff
@@ -0,0 +1,38 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+import org.apache.commons.lang3.tuple.Pair;
+import org.junit.jupiter.api.Test;
+
+public class DeltaExternalTableReadWriteTest extends ExternalTableReadWriteTest {
+
+  @Test
+  public void testDeltaPathTable() throws IOException {
```
yili-db, line 16: This is an **extra** test only for Delta external table.

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaManagedTableReadWriteTest.java
```diff
@@ -0,0 +1,175 @@
+package io.unitycatalog.spark;
```
yili-db, line 1: This file is renamed from ManagedTableReadWriteTest.java. But adding an extra dimention of `withPartition` plus the new usage of `SetupTableOptions` produced too much difference so that git is no longer recognizing it as a rename.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
```diff
@@ -129,13 +130,10 @@ protected CatalogOperations createCatalogOperations(ServerConfig serverConfig) {
     return new SdkCatalogOperations(createApiClient(serverConfig));
   }
 
+  @SneakyThrows
```
tdas, line 133: what does this do?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -6,58 +6,206 @@
...
+          "TBLPROPERTIES ('%s'='%s')",
+          UCTableProperties.DELTA_CATALOG_MANAGED_KEY,
+          UCTableProperties.DELTA_CATALOG_MANAGED_VALUE);
+
+  @NoArgsConstructor
+  @AllArgsConstructor
+  @Accessors(chain = true)
+  @Getter
+  @Setter
+  protected class SetupTableOptions {
```
tdas: @openinx this looks like an interesting pattern to follow in the delta repo as well

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -6,58 +6,206 @@
...
    */
-  protected final String setupDeltaTable(
-      String catalogName, String tableName, List<String> partitionColumns)
-      throws IOException, ApiException {
-    return setupDeltaTable(managedStorageCloudScheme(), catalogName, tableName, partitionColumns);
+  protected final String setupTable(String catalogName, String tableName) {
+    return setupTable(new SetupTableOptions().setCatalogName(catalogName).setTableName(tableName));
+  }
+
+  /** Set up a delta table with storage on emulated cloud and returns the table full name. */
```
tdas: is this still delta?
the method name does not clarify. if its to be used only for Delta then please keep it in the method name. if its for any format, how does this specify the format?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -6,58 +6,206 @@
...
+  /** The table format to be tested */
+  protected abstract String tableFormat();
+
+  protected final boolean testingDelta() {
+    return tableFormat().equalsIgnoreCase("DELTA");
+  }
 
   @Test
-  public void testDeltaTableReadWrite() throws IOException, ApiException {
+  public void testTableCreateReadWrite() {
```
tdas, line 213: this is not a simple create/read/write. please expand the comment to explain the entire scope.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -6,58 +6,206 @@
...
-
-    String t3 = setupDeltaTable(CATALOG_NAME, DELTA_TABLE, List.of());
-    testTableReadWrite(t3);
-
-    String t4 = setupDeltaTable(CATALOG_NAME, DELTA_TABLE_PARTITIONED, Arrays.asList("s"));
-    testTableReadWrite(t4);
+    int tableNameCounter = 0;
+    for (String catalogName : List.of(SPARK_CATALOG, CATALOG_NAME)) {
+      for (boolean withPartitionColumns : List.of(true, false)) {
+        for (boolean withCtas : List.of(true, false)) {
```
tdas, line 220: should we test replace?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
tdas: no need for join. thats not very different from select. 
you should add few variations of alter table

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaExternalTableReadWriteTest.java
tdas: add this as a class docs 
there are already tests with in Base which is Delta-only.. when should we add tests there vs here.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
tdas: should add docs to explain that tableFormat is controlled at the testsuite level.

tdas: nit: `Setup` is a verb in this case. .. so are you setting up "table options"?
i think grammatically it makes more sense to say "TableSetupOptions" where the whole thing is a noun.. a type of "options"

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: ALTER TABLE is not supported yet:
```
  override def alterTable(ident: Identifier, changes: TableChange*): Table = {
    throw new UnsupportedOperationException("Altering a table is not supported yet")
  }
```

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaExternalTableReadWriteTest.java
yili-db: This table has a special test that is only for **external** **Delta** tables.

Some tests in Base are also Delta-only but they are for both external and managed tables.
Because we already have `ExternalTableReadWriteTest`. We can not split the tests in another dimension to have a `DeltaTableReadWriteTest` for both managed and external tables. Java only allows single inheritance.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
yili-db: Just to remove the boilerplate:
```
    try {
...
    } catch (ApiException e) {
      throw new RuntimeException(e);
    }
```
Previously this test package doesn't depend on `lombok` so it wasn't used. Now we can.

connectors/spark/src/test/java/io/unitycatalog/spark/LocalDeltaManagedTableReadWriteTest.java
```diff
@@ -6,7 +6,7 @@
 /**
  * This test suite starts UC server with managed storage root on local path and exercise the tests.
  */
-public class LocalManagedTableReadWriteTest extends ManagedTableReadWriteTest {
+public class LocalDeltaManagedTableReadWriteTest extends DeltaManagedTableReadWriteTest {
```
tdas, line 9: remind me why this test exists and cloudParameters cannot handled as cloudParameter?

build.sbt
```diff
@@ -574,6 +574,19 @@ lazy val spark = (project in file("connectors/spark"))
     ),
     javaCheckstyleSettings("dev/checkstyle-config.xml"),
     Compile / compile / javacOptions ++= javacRelease11,
+    Test / compile / javacOptions := {
+      // lombok is only a dependency of test. So its path needs to be added explicitly.
```
tdas, line 578: i dont get this. isnt lombok used in any other tests in any other module? 
if not, is there documentation what is needed for lombok to work with javac? this seems very esoteric for a library that just adds annotations.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
tdas: understood.

connectors/spark/src/test/java/io/unitycatalog/spark/LocalDeltaManagedTableReadWriteTest.java
yili-db: Ideally we should have all clouds in `cloudParameters` and don't need to have a separate test suite class to provide different cloud parameter.
The reason we need it is actually `managedStorageCloudScheme`. When the UC server starts **before** each test run it needs to know the cloud for managed tables already. However there isn't a clean way to know that before the test actually runs. So here's the comprise that for managed table tests I just split into different test suites according to cloud and each test suite only deal with one single cloud.

tdas: aah i get it again. 
see this is convoluted enough that I am having to ask this again within 2 weeks. 
so i am looking forward to managed location support that simplifies this.

tdas (approved):
LGTM

build.sbt
yili-db: lombok is used in other modules in this file. And they just have a simpler config to make it work:
```
    Compile / compile / javacOptions ++= Seq(
      "-processor",
      "lombok.launch.AnnotationProcessorHider$AnnotationProcessor"
    ) ++ javacRelease17,
``` 

 However it just doesn't work here unless I add this `lombokPath` hack.
I think this module has a unique situation here:
1. the production code doesn't use lombok. Being a scala module, its dependencies likely don't use it neither.
2. however the test is in java and it wants to use lombok to avoid a lot of boilerplate.
3. I want to use lombok for test only and avoid polluting the UC-Spark package (which is mostly scala) with unnecessary dependency.

I don't have 100% understanding of this `lombokPath` hack as it is the best answer given by AI. But it works and it's safe as a test only thing.

PR #1248, opened 2025-11-25
Fix the name quote bug in UCSingleCatalog.

yili-db:
**PR Checklist**

Fix the name quote bug in UCSingleCatalog.
UCSingleCatalog should not quote schema and table names with backtick before sending out an RPC to UC server.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -187,6 +187,25 @@ object UCSingleCatalog {
     }
   }
 
+  /**
+   * Constructs a fully qualified table name for Unity Catalog API calls.
+   *
+   * This method creates a three-part name in the format `catalog.schema.table` by combining
+   * the catalog name with the schema name (from the identifier's namespace) and table name.
+   * It is NOT backtick quoted like what is usually used in SQL statements even if the names have
+   * special characters like hyphens.
+   *
+   * Example:
+   * catalogName=catalog, ident=(schema, table): it returns "catalog.schema.table"
+   * catalogName=cata-log, ident=(sche-ma, ta-ble): it returns "cata-log.sche-ma.ta-ble" (no quote)
+   * catalogName=catalog, ident=((schema1, schema2), table): it throws
+   *   ApiException(Nested namespace not supported)
+   *
+   * @param catalogName the name of the catalog
+   * @param ident the table identifier containing the namespace (schema) and table name
+   * @return a fully qualified table name in the format "catalog.schema.table"
+   * @throws ApiException if the identifier contains nested namespaces (more than one level)
+   */
   def fullTableNameForApi(catalogName: String, ident: Identifier): String = {
     checkUnsupportedNestedNamespace(ident.namespace())
     Seq(catalogName, ident.namespace()(0), ident.name()).mkString(".")


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -180,6 +180,17 @@ class UCSingleCatalog
 object UCSingleCatalog {
   val LOAD_DELTA_CATALOG = ThreadLocal.withInitial[Boolean](() => true)
   val DELTA_CATALOG_LOADED = ThreadLocal.withInitial[Boolean](() => false)
+
+  def checkUnsupportedNestedNamespace(namespace: Array[String]) = {
```
openinx: nit: My intellij idea suggested to `Type annotation required for Unit definition `.  Maybe we can add a `Unit` for the method return type ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -180,6 +180,17 @@ class UCSingleCatalog
...
+    if (namespace.length > 1) {
+      throw new ApiException("Nested namespaces are not supported: " + namespace.mkString("."))
+    }
+  }
+
+  def fullTableNameForApi(catalogName: String, ident: Identifier) = {
```
openinx: Similar tips from my idea:

```
  def fullTableNameForApi(catalogName: String, ident: Identifier): String = {
    checkUnsupportedNestedNamespace(ident.namespace())
    Seq(catalogName, ident.namespace()(0), ident.name()).mkString(".")
  }
```

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -299,6 +299,32 @@ public void testCreateManagedDeltaTable() throws IOException {
         .hasMessageContaining("not support managed table");
   }
 
+  // TODO: move this into BaseTableReadWriteTest
+  @Test
+  public void hyphenInTableName() throws ApiException, IOException {
```
openinx, line 304: Thanks for the test coverage, basically we covered the cases in this test:
*  Schema: create
*  Table:  create, SELECT, INSERT, SHOW, DROP

The cases that we may not cover are:
*  Catalog:  create / drop / show
*  Schema:  drop/show.

The ideal goal for us, is have an abstracted test suites which cover all the tests, and people can set different `<catalog>, <schema>, <table>` combination to have the tests. But seems like the current test infra is lag behind the goal.  Maybe we need to think about how to do the abstraction in futher. 

The abstraction work not block this PR, while maybe we need more coverage for this one currently.

yili-db: In general we are lacking test coverage. but this PR is only changing `loadTable`. This test should be sufficient to cover just that.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -180,6 +180,17 @@ class UCSingleCatalog
...
   val LOAD_DELTA_CATALOG = ThreadLocal.withInitial[Boolean](() => true)
   val DELTA_CATALOG_LOADED = ThreadLocal.withInitial[Boolean](() => false)
+
+  def checkUnsupportedNestedNamespace(namespace: Array[String]): Unit = {
+    if (namespace.length > 1) {
+      throw new ApiException("Nested namespaces are not supported: " + namespace.mkString("."))
+    }
+  }
+
+  def fullTableNameForApi(catalogName: String, ident: Identifier): String = {
```
tdas, line 209: please add docs explaining what this is doing with examples. what can be the various fields in ident, and what does it finally produce.

Issue #1259, opened 2025-12-02
Abstract the general UC Client to the unitycatalog-client module


openinx:
There are several factors that make us to unify the uc-client into the unitycatalog-client modules

1.  When I developed the OAuth feature for the unitycatalog, we find the OSS-Delta also have the UCClient modules, which means we need to implemement the OAuth in two different projects.  https://github.com/unitycatalog/unitycatalog/issues/1200#issuecomment-3531279892 .  A better approach for us is: maintaing the Java SDK into an unified module, so that different projects can share the same module. 

2.  UC client need the following funtionalities:
* The unified retry policy framework. 
* The different unitycatalog authentication methods:  
    - Personal Access Token (PAT) 
    - OAuth support. 
*  Customize the User-Agent. 
*  Well-defined the abstracted API etc. 

So let me open a separate issue for this whole stuff, and I will try to define the API in the `unitycatalog-client` module.

PR #1261, opened 2025-12-02
Abstract the general UC Client to the unitycatalog-client module

openinx:
**PR Checklist**

**Description of changes**

The related issue is: https://github.com/unitycatalog/unitycatalog/issues/1259

### DIFF
--- a/clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
+++ b/clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
@@ -15,7 +15,7 @@ import java.net.URI;
  * <pre>{@code
  * ApiClient client = ApiClientBuilder.create()
  *     .url("http://localhost:8080")
- *     .tokenProvider(TokenProvider.builder().token("my-token").build())
+ *     .tokenProvider(TokenProvider.create("my-token"))
  *     .retryPolicy(JitterDelayRetryPolicy.builder().maxAttempts(5).build())
  *     .clientVersion("MyApp", "1.0.0", "Java", "11")
  *     .build();
@@ -28,7 +28,7 @@ import java.net.URI;
 public class ApiClientBuilder {
   private static final String BASE_PATH = "/api/2.1/unity-catalog";
 
-  private URI url = null;
+  private URI uri = null;
   private TokenProvider tokenProvider = null;
   private String[] nameVersionPairs = null;
   private RetryPolicy retryPolicy = JitterDelayRetryPolicy.builder().build();
@@ -43,27 +43,27 @@ public class ApiClientBuilder {
   }
 
   /**
-   * Sets the Unity Catalog server URL.
+   * Sets the Unity Catalog server URI.
    *
    * <p>The base path {@code /api/2.1/unity-catalog} will be automatically appended.
    *
-   * @param url the Unity Catalog server URL, must not be null
+   * @param uri the Unity Catalog server URI, must not be null
    * @return this builder instance for method chaining
    */
-  public ApiClientBuilder url(URI url) {
-    this.url = url;
+  public ApiClientBuilder uri(URI uri) {
+    this.uri = uri;
     return this;
   }
 
   /**
-   * Sets the Unity Catalog server URL from a string.
+   * Sets the Unity Catalog server URI from a string.
    *
-   * @param url the Unity Catalog server URL, must not be null
+   * @param uri the Unity Catalog server URI, must not be null
    * @return this builder instance for method chaining
-   * @see #url(URI)
+   * @see #uri(URI)
    */
-  public ApiClientBuilder url(String url) {
-    return url(URI.create(url));
+  public ApiClientBuilder uri(String uri) {
+    return uri(URI.create(uri));
   }
 
   /**
@@ -81,7 +81,7 @@ public class ApiClientBuilder {
   }
 
   /**
-   * Sets client version metadata as name-version pairs.
+   * Sets application version metadata as name-version pairs.
    *
    * <p>Arguments must be provided in alternating name-version order with an even count, e.g.,
    * {@code "MyApp1", "0.1.0", "MyApp2", "1.0.2"}.
@@ -90,7 +90,7 @@ public class ApiClientBuilder {
    * @return this builder instance for method chaining
    * @throws IllegalArgumentException if an odd number of arguments is provided
    */
-  public ApiClientBuilder clientVersion(String... nameVersionPairs) {
+  public ApiClientBuilder appVersion(String... nameVersionPairs) {
     Preconditions.checkArgument(
         nameVersionPairs.length % 2 == 0,
         "Must provide an even number of arguments for the name-version pairs.");
@@ -117,16 +117,16 @@ public class ApiClientBuilder {
    * Builds and returns a configured {@link ApiClient} instance.
    *
    * @return a configured {@link ApiClient} instance
-   * @throws NullPointerException if {@code url} or {@code ucTokenProvider} is null
+   * @throws NullPointerException if {@code uri} or {@code tokenProvider} is null
    */
   public ApiClient build() {
     // Set the scheme, host, port and base path, for the unity catalog client.
-    Preconditions.checkNotNull(url, "The unitycatalog url cannot be null");
+    Preconditions.checkNotNull(uri, "The unitycatalog uri cannot be null");
     ApiClient apiClient = new RetryingApiClient(retryPolicy);
-    apiClient.setScheme(url.getScheme());
-    apiClient.setHost(url.getHost());
-    apiClient.setPort(url.getPort());
-    apiClient.setBasePath(url.getPath() + BASE_PATH);
+    apiClient.setScheme(uri.getScheme());
+    apiClient.setHost(uri.getHost());
+    apiClient.setPort(uri.getPort());
+    apiClient.setBasePath(uri.getPath() + BASE_PATH);
 
     // Set the unity catalog token provider.
     Preconditions.checkNotNull(tokenProvider, "The unitycatalog token provider cannot be null");
--- a/clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
+++ b/clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
@@ -86,7 +86,7 @@ class OAuthTokenProvider implements TokenProvider {
   }
 
   @Override
-  public Map<String, String> properties() {
+  public Map<String, String> getConfigs() {
     return Map.of(
         AuthProps.OAUTH_URI, oauthUri,
         AuthProps.OAUTH_CLIENT_ID, oauthClientId,
--- a/clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
+++ b/clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
@@ -22,11 +22,11 @@ public interface TokenProvider {
   String accessToken();
 
   /**
-   * Returns the configuration properties associated with this token provider.
+   * Returns the configuration associated with this token provider.
    *
    * @return a map of configuration key-value pairs
    */
-  Map<String, String> properties();
+  Map<String, String> getConfigs();
 
   /**
    * Creates a token provider using a static access token.
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
@@ -12,21 +12,21 @@ public class ApiClientFactory {
   }
 
   public static ApiClient createApiClient(
-      RetryPolicy retryPolicy, URI url, TokenProvider ucTokenProvider) {
+      RetryPolicy retryPolicy, URI uri, TokenProvider tokenProvider) {
 
     // Create a new ApiClient Builder.
     ApiClientBuilder builder = ApiClientBuilder.create()
-        .url(url)
-        .tokenProvider(ucTokenProvider)
+        .uri(uri)
+        .tokenProvider(tokenProvider)
         .retryPolicy(retryPolicy);
 
     // Add Spark to User-Agent, and Delta if available
     String sparkVersion = getSparkVersion();
     String deltaVersion = getDeltaVersion();
     if (deltaVersion != null) {
-      builder.clientVersion("Spark", sparkVersion, "Delta", deltaVersion);
+      builder.appVersion("Spark", sparkVersion, "Delta", deltaVersion);
     } else {
-      builder.clientVersion("Spark", sparkVersion);
+      builder.appVersion("Spark", sparkVersion);
     }
 
     return builder.build();
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
@@ -37,10 +37,10 @@ public class CredPropsUtil {
       return self();
     }
 
-    public T ucTokenProvider(TokenProvider ucTokenProvider) {
+    public T tokenProvider(TokenProvider tokenProvider) {
       // As we can only propagate the properties with prefix 'fs.*' to the FileSystem
       // implementation. So let's add the prefix here.
-      ucTokenProvider.properties()
+      tokenProvider.getConfigs()
           .forEach((key, value) -> builder.put(UCHadoopConf.FS_UC_PREFIX + key, value));
       return self();
     }
@@ -141,13 +141,13 @@ public class CredPropsUtil {
 
   private static S3PropsBuilder s3TempCredPropsBuilder(
       String uri,
-      TokenProvider ucTokenProvider,
+      TokenProvider tokenProvider,
       TemporaryCredentials tempCreds) {
     AwsCredentials awsCred = tempCreds.getAwsTempCredentials();
     S3PropsBuilder builder = new S3PropsBuilder()
         .set(UCHadoopConf.S3A_CREDENTIALS_PROVIDER, AwsVendedTokenProvider.class.getName())
         .uri(uri)
-        .ucTokenProvider(ucTokenProvider)
+        .tokenProvider(tokenProvider)
         .uid(UUID.randomUUID().toString())
         .set(UCHadoopConf.S3A_INIT_ACCESS_KEY, awsCred.getAccessKeyId())
         .set(UCHadoopConf.S3A_INIT_SECRET_KEY, awsCred.getSecretAccessKey())
@@ -164,11 +164,11 @@ public class CredPropsUtil {
 
   private static Map<String, String> s3TableTempCredProps(
       String uri,
-      TokenProvider ucTokenProvider,
+      TokenProvider tokenProvider,
       String tableId,
       TableOperation tableOp,
       TemporaryCredentials tempCreds) {
-    return s3TempCredPropsBuilder(uri, ucTokenProvider, tempCreds)
+    return s3TempCredPropsBuilder(uri, tokenProvider, tempCreds)
         .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE)
         .tableId(tableId)
         .tableOperation(tableOp)
@@ -177,11 +177,11 @@ public class CredPropsUtil {
 
   private static Map<String, String> s3PathTempCredProps(
       String uri,
-      TokenProvider ucTokenProvider,
+      TokenProvider tokenProvider,
       String path,
       PathOperation pathOp,
       TemporaryCredentials tempCreds) {
-    return s3TempCredPropsBuilder(uri, ucTokenProvider, tempCreds)
+    return s3TempCredPropsBuilder(uri, tokenProvider, tempCreds)
         .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_PATH_VALUE)
         .path(path)
         .pathOperation(pathOp)
@@ -202,14 +202,14 @@ public class CredPropsUtil {
 
   private static GcsPropsBuilder gcsTempCredPropsBuilder(
       String uri,
-      TokenProvider ucTokenProvider,
+      TokenProvider tokenProvider,
       TemporaryCredentials tempCreds) {
     GcpOauthToken gcpToken = tempCreds.getGcpOauthToken();
     GcsPropsBuilder builder = new GcsPropsBuilder()
         .set("fs.gs.auth.type", "ACCESS_TOKEN_PROVIDER")
         .set("fs.gs.auth.access.token.provider", GcsVendedTokenProvider.class.getName())
         .uri(uri)
-        .ucTokenProvider(ucTokenProvider)
+        .tokenProvider(tokenProvider)
         .uid(UUID.randomUUID().toString())
         .set(UCHadoopConf.GCS_INIT_OAUTH_TOKEN, gcpToken.getOauthToken());
 
@@ -224,11 +224,11 @@ public class CredPropsUtil {
 
   private static Map<String, String> gsTableTempCredProps(
       String uri,
-      TokenProvider ucTokenProvider,
+      TokenProvider tokenProvider,
       String tableId,
       TableOperation tableOp,
       TemporaryCredentials tempCreds) {
-    return gcsTempCredPropsBuilder(uri, ucTokenProvider, tempCreds)
+    return gcsTempCredPropsBuilder(uri, tokenProvider, tempCreds)
         .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE)
         .tableId(tableId)
         .tableOperation(tableOp)
@@ -237,11 +237,11 @@ public class CredPropsUtil {
 
   private static Map<String, String> gsPathTempCredProps(
       String uri,
-      TokenProvider ucTokenProvider,
+      TokenProvider tokenProvider,
       String path,
       PathOperation pathOp,
       TemporaryCredentials tempCreds) {
-    return gcsTempCredPropsBuilder(uri, ucTokenProvider, tempCreds)
+    return gcsTempCredPropsBuilder(uri, tokenProvider, tempCreds)
         .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_PATH_VALUE)
         .path(path)
         .pathOperation(pathOp)
@@ -257,13 +257,13 @@ public class CredPropsUtil {
 
   private static AbfsPropsBuilder abfsTempCredPropsBuilder(
       String uri,
-      TokenProvider ucTokenProvider,
+      TokenProvider tokenProvider,
       TemporaryCredentials tempCreds) {
     AzureUserDelegationSAS azureSas = tempCreds.getAzureUserDelegationSas();
     AbfsPropsBuilder builder = new AbfsPropsBuilder()
         .set(FS_AZURE_SAS_TOKEN_PROVIDER_TYPE, AbfsVendedTokenProvider.class.getName())
         .uri(uri)
-        .ucTokenProvider(ucTokenProvider)
+        .tokenProvider(tokenProvider)
         .uid(UUID.randomUUID().toString())
         .set(UCHadoopConf.AZURE_INIT_SAS_TOKEN, azureSas.getSasToken());
 
@@ -278,11 +278,11 @@ public class CredPropsUtil {
 
   private static Map<String, String> abfsTableTempCredProps(
       String uri,
-      TokenProvider ucTokenProvider,
+      TokenProvider tokenProvider,
       String tableId,
       TableOperation tableOp,
       TemporaryCredentials tempCreds) {
-    return abfsTempCredPropsBuilder(uri, ucTokenProvider, tempCreds)
+    return abfsTempCredPropsBuilder(uri, tokenProvider, tempCreds)
         .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE)
         .tableId(tableId)
         .tableOperation(tableOp)
@@ -291,11 +291,11 @@ public class CredPropsUtil {
 
   private static Map<String, String> abfsPathTempCredProps(
       String uri,
-      TokenProvider ucTokenProvider,
+      TokenProvider tokenProvider,
       String path,
       PathOperation pathOp,
       TemporaryCredentials tempCreds) {
-    return abfsTempCredPropsBuilder(uri, ucTokenProvider, tempCreds)
+    return abfsTempCredPropsBuilder(uri, tokenProvider, tempCreds)
         .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_PATH_VALUE)
         .path(path)
         .pathOperation(pathOp)
--- a/clients/java/src/main/java/io/unitycatalog/client/auth/FixedTokenProvider.java
+++ b/clients/java/src/main/java/io/unitycatalog/client/auth/FixedTokenProvider.java
@@ -15,7 +15,7 @@ class FixedTokenProvider implements TokenProvider {
   }
 
   @Override
-  public Map<String, String> properties() {
+  public Map<String, String> getConfigs() {
     return Map.of(AuthProps.TOKEN, token);
   }
 }
--- a/clients/java/src/test/java/io/unitycatalog/client/ApiClientBuilderTest.java
+++ b/clients/java/src/test/java/io/unitycatalog/client/ApiClientBuilderTest.java
@@ -21,57 +21,57 @@ import org.mockito.ArgumentCaptor;
  */
 public class ApiClientBuilderTest {
 
-  private static final String TEST_URL = "http://localhost:8080";
+  private static final String TEST_URI = "http://localhost:8080";
   private static final String TEST_TOKEN = "test-token";
 
   @Test
-  public void testBasicBuildAndUrlConfiguration() {
+  public void testBasicBuildAndUriConfiguration() {
     TokenProvider tokenProvider = TokenProvider.create(TEST_TOKEN);
 
     // Test basic build with minimal configuration
     ApiClient client1 =
-        ApiClientBuilder.create().url(TEST_URL).tokenProvider(tokenProvider).build();
+        ApiClientBuilder.create().uri(TEST_URI).tokenProvider(tokenProvider).build();
     assertThat(client1).isNotNull();
     assertThat(client1.getRequestInterceptor()).isNotNull();
 
-    // Test URL with String parameter (different host and port)
+    // Test URI with String parameter (different host and port)
     ApiClient client2 =
         ApiClientBuilder.create()
-            .url("https://example.com:9090")
+            .uri("https://example.com:9090")
             .tokenProvider(tokenProvider)
             .build();
     assertThat(client2).isNotNull();
 
-    // Test URL with URI parameter
+    // Test URI with URI parameter
     ApiClient client3 =
         ApiClientBuilder.create()
-            .url(URI.create("https://catalog.example.com:8443"))
+            .uri(URI.create("https://catalog.example.com:8443"))
             .tokenProvider(tokenProvider)
             .build();
     assertThat(client3).isNotNull();
 
-    // Test URL with custom path
+    // Test URI with custom path
     ApiClient client4 =
         ApiClientBuilder.create()
-            .url("http://localhost:8080/custom/path")
+            .uri("http://localhost:8080/custom/path")
             .tokenProvider(tokenProvider)
             .build();
     assertThat(client4).isNotNull();
 
-    // Test URL without explicit port
+    // Test URI without explicit port
     ApiClient client5 =
-        ApiClientBuilder.create().url("https://example.com").tokenProvider(tokenProvider).build();
+        ApiClientBuilder.create().uri("https://example.com").tokenProvider(tokenProvider).build();
     assertThat(client5).isNotNull();
 
     // Verify multiple builds produce different instances
-    ApiClientBuilder builder = ApiClientBuilder.create().url(TEST_URL).tokenProvider(tokenProvider);
+    ApiClientBuilder builder = ApiClientBuilder.create().uri(TEST_URI).tokenProvider(tokenProvider);
     assertThat(builder.build()).isNotSameAs(builder.build());
   }
 
   @Test
   public void testTokenProviderConfiguration() {
     TokenProvider tokenProvider = TokenProvider.create(TEST_TOKEN);
-    ApiClient client = ApiClientBuilder.create().url(TEST_URL).tokenProvider(tokenProvider).build();
+    ApiClient client = ApiClientBuilder.create().uri(TEST_URI).tokenProvider(tokenProvider).build();
 
     assertThat(client.getRequestInterceptor()).isNotNull();
 
@@ -90,38 +90,34 @@ public class ApiClientBuilderTest {
   }
 
   @Test
-  public void testClientVersionConfiguration() {
+  public void testAppVersionConfiguration() {
     TokenProvider tokenProvider = TokenProvider.create(TEST_TOKEN);
 
     // Test single client version pair
     ApiClient client1 =
         ApiClientBuilder.create()
-            .url(TEST_URL)
+            .uri(TEST_URI)
             .tokenProvider(tokenProvider)
-            .clientVersion("MyApp", "1.0.0", "Java", "11")
+            .appVersion("MyApp", "1.0.0", "Java", "11")
             .build();
     assertThat(client1.getUserAgent()).contains("MyApp/1.0.0", "Java/11");
 
     // Test multiple client version pairs
     ApiClient client2 =
         ApiClientBuilder.create()
-            .url(TEST_URL)
+            .uri(TEST_URI)
             .tokenProvider(tokenProvider)
-            .clientVersion("App1", "1.0", "App2", "2.0", "App3", "3.0")
+            .appVersion("App1", "1.0", "App2", "2.0", "App3", "3.0")
             .build();
     assertThat(client2.getUserAgent()).contains("App1/1.0", "App2/2.0", "App3/3.0");
 
     // Test empty client version (no-op)
     ApiClient client3 =
-        ApiClientBuilder.create()
-            .url(TEST_URL)
-            .tokenProvider(tokenProvider)
-            .clientVersion()
-            .build();
+        ApiClientBuilder.create().uri(TEST_URI).tokenProvider(tokenProvider).appVersion().build();
     assertThat(client3).isNotNull();
 
     // Test validation: odd number of arguments
-    assertThatThrownBy(() -> ApiClientBuilder.create().clientVersion("MyApp", "1.0.0", "Java"))
+    assertThatThrownBy(() -> ApiClientBuilder.create().appVersion("MyApp", "1.0.0", "Java"))
         .isInstanceOf(IllegalArgumentException.class)
         .hasMessageContaining("Must provide an even number of arguments");
   }
@@ -134,7 +130,7 @@ public class ApiClientBuilderTest {
     RetryPolicy customRetryPolicy = JitterDelayRetryPolicy.builder().maxAttempts(10).build();
     ApiClient client1 =
         ApiClientBuilder.create()
-            .url(TEST_URL)
+            .uri(TEST_URI)
             .tokenProvider(tokenProvider)
             .retryPolicy(customRetryPolicy)
             .build();
@@ -142,7 +138,7 @@ public class ApiClientBuilderTest {
 
     // Test with default retry policy (no explicit setting)
     ApiClient client2 =
-        ApiClientBuilder.create().url(TEST_URL).tokenProvider(tokenProvider).build();
+        ApiClientBuilder.create().uri(TEST_URI).tokenProvider(tokenProvider).build();
     assertThat(client2).isNotNull();
   }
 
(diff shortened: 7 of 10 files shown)


clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
```diff
@@ -1,29 +1,28 @@
-package io.unitycatalog.spark.auth.catalog;
+package io.unitycatalog.client.auth;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import io.unitycatalog.spark.ApiClientConf;
-import io.unitycatalog.spark.RetryingApiClient;
-import io.unitycatalog.spark.UCHadoopConf;
-import io.unitycatalog.spark.utils.Clock;
+import com.google.common.base.Preconditions;
```
openinx: We have to use this `guava` for all those preconditions checks and [Throwables](https://github.com/unitycatalog/unitycatalog/pull/1261/files#diff-4f42fa8947c6b10f473c01306c7a841fb7f41ab880fea086ff8494fdc8c3fb1aR87) usage.,  maybe we need to relocate it.

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
```diff
@@ -1,29 +1,28 @@
...
-import org.sparkproject.guava.base.Preconditions;
 
 /**
  * OAuth-based token provider that fetches and automatically renews access tokens.
  *
- * <p>This provider uses the OAuth 2.0 client credentials flow to obtain access tokens.
- * It automatically renews tokens before they expire (default: 30 seconds before expiration).
- * Token requests are retried using {@link RetryingApiClient} for resilience against
- * transient network failures.</p>
+ * <p>This provider uses the OAuth 2.0 client credentials flow to obtain access tokens. It
```
openinx, line 23: Those changes are automatically reformatted via the `./build/sbt client/javafmtAll`, after moved codes from the `unitycatalog-spark` to the `unitycatalog-client` module, which I think should be reasonable.

clients/java/src/main/java/io/unitycatalog/client/auth/AuthProps.java
```diff
@@ -0,0 +1,10 @@
+package io.unitycatalog.client.auth;
+
+public class UCAuthProps {
+  private UCAuthProps() {}
+
+  public static final String TOKEN = "token";
```
openinx: We have to move those auth properties from the `unitycatalog-spark` to the `unitycatalog-client` module, since those are for the `UCTokenProvider`'s properties. 

We need to use those properties to construct and de-construct the `UCTokenProvider` implementation, including `FixedUCTokenProvider` and `OAuthUCTokenProvider`.

clients/java/src/main/java/io/unitycatalog/client/auth/UCTokenProvider.java
```diff
@@ -0,0 +1,99 @@
...
+ * Interface for providing access tokens to authenticate with Unity Catalog.
+ *
+ * <p>Implementations include:
+ *
+ * <ul>
+ *   <li>{@link FixedUCTokenProvider} - uses a pre-configured static token
+ *   <li>{@link OAuthUCTokenProvider} - obtains tokens via OAuth 2.0 client credentials flow
+ * </ul>
+ */
+public interface UCTokenProvider {
```
openinx: Public API for  `unitycatalog-client`.

clients/java/src/main/java/io/unitycatalog/client/retry/JitterDelayRetryPolicy.java
```diff
@@ -0,0 +1,172 @@
...
+ *     .maxAttempts(5)
+ *     .initDelayMs(100L)
+ *     .delayMultiplier(2.0)
+ *     .delayJitterFactor(0.2)
+ *     .build();
+ * }</pre>
+ *
+ * @see RetryPolicy
+ */
+public class JitterDelayRetryPolicy implements RetryPolicy {
```
openinx, line 43: Public API for `unitycatalog-client`.

clients/java/src/main/java/io/unitycatalog/client/auth/UCTokenProvider.java
openinx: Do we need some kind of annotation to make whether it's public API or not ?  But I don't find a approperiate annotation tools for this.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private void setIfPresent(Map<String, String> options, String key, Consumer<String> setter) {
+      String value = options.get(key);
+      if (value != null) {
+        setter.accept(value);
+      }
+    }
+
+    public Builder options(Map<String, String> options) {
```
openinx: Set all the attributes via parsing the auth properties from the map.

clients/java/src/main/java/io/unitycatalog/client/retry/FixedDelayRetryPolicy.java
```diff
@@ -0,0 +1,51 @@
+package io.unitycatalog.client.retry;
+
+import java.time.Duration;
+
+public class FixedDelayRetryPolicy implements RetryPolicy {
```
openinx: I just keep it here for another kind of implementation for `RetryPolicy`, although we did not really use it in the production code.

clients/java/src/main/java/io/unitycatalog/client/retry/JitterDelayRetryPolicy.java
```diff
@@ -0,0 +1,172 @@
...
+ *       computed delay (e.g., 0.5 randomizes by ±50%).
+ * </ul>
+ *
+ * <p>Example usage:
+ *
+ * <pre>{@code
+ * RetryPolicy policy = JitterDelayRetryPolicy.builder()
+ *     .maxAttempts(5)
+ *     .initDelayMs(100L)
+ *     .delayMultiplier(2.0)
```
openinx, line 36: It's unlikely for people to tuning this `delayMultiplier`.  Maybe we can make it as a constant value, with no allowrance for people to configure this, which simplify the API setting.

openinx: Do people really care about the `delayMultiper` and `delayJitterFactor` ?   I guess exposing `maxAttempts` and `initDelayMs` is good enough for people to tune the jitter retry policy.

clients/java/src/main/java/io/unitycatalog/client/retry/JitterDelayRetryPolicy.java
```diff
@@ -0,0 +1,172 @@
...
+     */
+    public Builder delayMultiplier(double delayMultiplier) {
+      this.delayMultiple = delayMultiplier;
+      return this;
+    }
+
+    /**
+     * Sets the jitter factor to randomize delays and prevent synchronized retries.
+     *
+     * <p>The jitter factor should be in the range [0, 1). A factor of 0.5 means the actual delay
```
openinx, line 160: We need to add the `Preconditions.checkArgument`.

clients/java/src/main/java/io/unitycatalog/client/retry/RetryPolicy.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * <p>Example usage:
+ *
+ * <pre>{@code
+ * RetryPolicy policy = JitterDelayRetryPolicy.builder().build();
+ * int attempt = 1;
+ * if (policy.allowRetry(attempt)) {
+ *   Thread.sleep(policy.sleepTime(attempt).toMillis());
+ *   // Retry the operation
+ * }
```
openinx, line 22: Maybe use a better example?

clients/java/src/main/java/io/unitycatalog/client/utils/Clock.java
```diff
@@ -1,21 +1,19 @@
-package io.unitycatalog.spark.utils;
+package io.unitycatalog.client.utils;
```
openinx: No code change,  Automatically reformatted after applied the `build/sbt client/javafmtAll`.

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
```diff
@@ -0,0 +1,142 @@
...
+ *     .retryPolicy(JitterDelayRetryPolicy.builder().maxAttempts(5).build())
+ *     .clientVersion("MyApp", "1.0.0", "Java", "11")
+ *     .build();
+ * }</pre>
+ *
+ * @see ApiClient
+ * @see UCTokenProvider
+ * @see RetryPolicy
+ */
+public class ApiClientBuilder {
```
openinx, line 28: Public API for `unitycatalog-client`.

clients/java/src/test/java/io/unitycatalog/client/EnumDeserializationTest.java
```diff
@@ -23,18 +23,16 @@
 import org.junit.jupiter.api.Test;
```
openinx, line 23: No code change,  automatically reformatted after applied the `build/sbt client/javafmtAll`.

clients/java/src/test/java/io/unitycatalog/client/UserAgentTest.java
```diff
@@ -1,16 +1,14 @@
 package io.unitycatalog.client;
```
openinx, line 1: No code change,  auto reformatted after applied `build/sbt client/javafmtAll`.

connectors/spark/src/main/scala/io/unitycatalog/spark/utils/OptionsUtil.java
```diff
@@ -9,9 +9,6 @@ private OptionsUtil() {
   public static final String URI = "uri";
   public static final String TOKEN = "token";
   public static final String WAREHOUSE = "warehouse";
-  public static final String OAUTH_URI = "oauth.uri";
```
openinx, line 12: Move from here to the https://github.com/unitycatalog/unitycatalog/pull/1261/files#diff-62ea646a74fab831be8a8d08ff4a3086e2de16fb4f2f9d344617bc39bd8e9185R3, which make more sense because those properties are used for unitycatalog client authentication.

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientConf.java
```diff
@@ -1,89 +0,0 @@
...
- *   <li>{@link #DEFAULT_REQUEST_INITIAL_DELAY_MS}: initial backoff delay in milliseconds; later
- *       attempts scale this using {@code initialDelayMs * multiplier ^ (attempt - 1) *
- *       (1 ± jitterFactor)}.</li>
- *   <li>{@link #DEFAULT_REQUEST_DELAY_MULTIPLIER}: exponential multiplier (e.g. {@code 2.0} doubles
- *       the wait each retry).</li>
- *   <li>{@link #DEFAULT_REQUEST_DELAY_JITTER_FACTOR}: jitter fraction in {@code [0, 1)} applied to
- *       the computed delay (e.g. {@code 0.5} randomises by ±50%).</li>
- * </ul>
- */
-public class ApiClientConf {
```
openinx, line 26: No need any more since we turn to use the `JitterDelayRetryPolicy.builder().xxx.build()`.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/catalog/UCTokenProviderTest.java
```diff
@@ -1,126 +0,0 @@
...
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
-
-import io.unitycatalog.spark.UCHadoopConf;
-import io.unitycatalog.spark.utils.OptionsUtil;
-import java.util.HashMap;
-import java.util.Map;
-import org.apache.hadoop.conf.Configuration;
-import org.junit.jupiter.api.Test;
-
-public class UCTokenProviderTest {
```
openinx, line 13: Moved to `unitycatalog-client` test module .

connectors/spark/src/main/scala/io/unitycatalog/spark/RetryingApiClient.java
```diff
@@ -1,34 +0,0 @@
...
- *
- * <p>The base {@link ApiClient} comes from the OpenAPI generator and is overwritten on every regen.
- * To keep that file untouched, this subclass overrides {@link #getHttpClient()} and wraps the
- * generated {@link HttpClient} in a {@link RetryingHttpClient} configured through
- * {@link ApiClientConf} and {@link HttpRetryHandler} (which defines the retry mechanism). Adding
- * retries at the HTTP layer means every generated API surface (for example {@code
- * TemporaryCredentialsApi}, {@code TablesApi}, etc.) inherits the same policy
- * without extra wrappers.</p>
- */
-public class RetryingApiClient extends ApiClient {
```
openinx, line 18: Moved to `unitycatalog-client` test module.

connectors/spark/src/main/scala/io/unitycatalog/spark/HttpRetryHandler.java
```diff
@@ -1,112 +0,0 @@
...
-import java.net.http.HttpRequest;
-import java.net.http.HttpResponse;
-import java.time.Duration;
-import java.time.Instant;
-import java.util.Set;
-
-/**
- * Retry handler that retries on common recoverable HTTP errors and network exceptions.
- */
-public class HttpRetryHandler {
```
openinx, line 18: Moved to `unitycatalog-client` module.

build.sbt
```diff
@@ -199,7 +206,7 @@ lazy val client = (project in file("target/clients/java"))
     // OpenAPI generation specs
     openApiInputSpec := (file(".") / "api" / "all.yaml").toString,
     openApiGeneratorName := "java",
-    openApiOutputDir := (file("target") / "clients" / "java").toString,
```
openinx, line 202: We move the openapi generated code from the project root directory to the `clients/java/target`, which make more sense because it's a Java SDK only.   it won't make other git changes, since all the generated code are excluded from git management.

build.sbt
```diff
@@ -185,8 +187,13 @@ lazy val client = (project in file("target/clients/java"))
       "org.openapitools" % "jackson-databind-nullable" % openApiToolsJacksonBindNullableVersion,
       "com.google.code.findbugs" % "jsr305" % "3.0.2",
       "jakarta.annotation" % "jakarta.annotation-api" % "3.0.0" % Provided,
+      // TODO: Be careful to introduce this guava, since it's easily to conflict.
```
openinx: This is a great point we need to take care of. My plan is: for the first step,  we can make it align with the spark version's guava.  And finally we will need to relocate it.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -89,36 +88,30 @@ private UCHadoopConf() {
...
-    conf.setInt(
-        REQUEST_RETRY_MAX_ATTEMPTS_KEY, apiClientConf.getRequestMaxAttempts());
-    conf.setLong(
-        REQUEST_RETRY_INITIAL_DELAY_KEY, apiClientConf.getRequestInitialDelayMs());
-    conf.setDouble(
-        REQUEST_RETRY_DELAY_MULTIPLIER_KEY, apiClientConf.getRequestDelayMultiplier());
-    conf.setDouble(
-        REQUEST_RETRY_DELAY_JITTER_FACTOR_KEY, apiClientConf.getRequestDelayJitterFactor());
-  }
+  public static RetryPolicy createRetryPolicy(Configuration conf) {
```
tdas: retry of what? needs docs.

clients/java/src/main/java/io/unitycatalog/client/auth/UCTokenProvider.java
tdas: do connectors need to access this class directly? if not, then this is internal.. and therefore put this class in a package named "internal".  `io.unitycatalog.client.internal.auth`

tdas: same for the implementations

tdas: This entire module is for UC. so putting UC in the name does not make sense. its ApiClient .. not UCApiClient. So lets just make this TokenProvider.
And remove UC from the implementations as well

clients/java/src/test/java/io/unitycatalog/client/auth/UCTokenProviderTest.java
```diff
@@ -0,0 +1,80 @@
+package io.unitycatalog.client.auth;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public class UCTokenProviderTest {
```
tdas: remove UC

clients/java/src/test/java/io/unitycatalog/client/auth/UCTokenProviderTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+  private static final String OAUTH_URI = "https://oauth.example.com/token";
+  private static final String CLIENT_ID = "test-client-id";
+  private static final String CLIENT_SECRET = "test-client-secret";
+
+  @Test
+  public void testCreateUCTokenProviderViaOptions() {
```
tdas: remove UC

clients/java/src/main/java/io/unitycatalog/client/auth/AuthProps.java
tdas: The TokenProvider does not need to be public class .. right? 
If yes, then UCAuthProps should not be public API as well ... isnt it? what is UCAuthProps really used for?

clients/java/src/main/java/io/unitycatalog/client/retry/RetryPolicy.java
```diff
@@ -0,0 +1,55 @@
...
+ * }</pre>
+ */
+public interface RetryPolicy {
```
tdas, line 25: "retry" of what?
name it RequestRetryPolicy.

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * Sets the token provider for authentication.
+   *
+   * <p>The token will be included in the Authorization header as a Bearer token.
+   *
+   * @param ucTokenProvider the token provider implementation, must not be null
+   * @return this builder instance for method chaining
+   * @see UCTokenProvider
+   */
+  public ApiClientBuilder ucTokenProvider(UCTokenProvider ucTokenProvider) {
```
tdas: do we need to make this public now? this is forces the UCTokenProvider to become public API

clients/java/src/main/java/io/unitycatalog/client/auth/UCAuthProps.java
```diff
@@ -0,0 +1,10 @@
+package io.unitycatalog.client.auth;
+
+public class UCAuthProps {
```
tdas: same comment as other places.. does not make sense to have "UC" in the name.

clients/java/src/main/java/io/unitycatalog/client/auth/AuthProps.java
openinx: > The TokenProvider does not need to be public class .. right?

I think the `UCTokenProvider`s are required to be public, since when we build the `ApiClient`,  it's actually the following example:

```java
  ApiClient client = ApiClientBuilder.create()
      .url("http://localhost:8080")
      .ucTokenProvider(UCTokenProvider.builder().token("my-token").build())
      .retryPolicy(JitterDelayRetryPolicy.builder().maxAttempts(5).build())
      .clientVersion("MyApp", "1.0.0", "Java", "11")
      .build();
```

Since we will need the `UCTokenProvider.builder()` to build a provider implementation to provide the valid unitycatalog token. 

I also considered not exposing the `UCTokenProvider` to public before,  which means exposing the specified methods in the `ApiClientBuilder`

```java
public ApiClientBuilder  token(String token) {
     // set the uc token provider internally.
     return this;
}

public ApiClientBuilder  oauthUri(String oauthUri){
   // ...
   return this;
}

public ApiClientBuilder  oauthClientId(String oauthClientId){
   // ...
   return this;
}


public ApiClientBuilder  oauthClientSecret(String oauthClientSecret){
   // ...
   return this;
}
```

But that seems will lack the flexibility for us to customizing the `UCTokenProvider` in the future, for example if we want to support the federal OAuth authentication, then we have no change to do that.

openinx: > what is UCAuthProps really used for?

Those uc authentication properties are used to generate the downstream properties.   For example,

* **Catalog level properties**:   we need to use those keys as the suffix to generate the catalog level properties

```sql
# For Personal Access Token (PAT)
spark.sql.catalog.<catalog-name>.token

# For OAuth.
spark.sql.catalog.<catalog-name>.oauth.uri
spark.sql.catalog.<catalog-name>.oauth.clientId
spark.sql.catalog.<catalog-name>.oauth.clientSecret
```


* **hadoop configuration**:  we want to use this `suffix` to propragate the UCTokenProvider properties to the underlying FileSystem instances, with `fs.unitycatalog.` prefix added.

```sql
# For Personal Access Token (PAT)
fs.unitycatalog.token

# For OAuth
fs.unitycatalog.oauth.uri
fs.unitycatalog.oauth.clientId
fs.unitycatalog.oauth.clientSecret
```

You can see, with the `Map<String, String> UCTokenProvider#properties()`,  people can easily use the abstracted approach to generate those catalog-level properties and hadoop configuration, but we may still need those public properties to set in the unit tests and integration tests.

clients/java/src/main/java/io/unitycatalog/client/auth/UCTokenProvider.java
openinx: Hi @tdas ,  I think it's the same discussion as this one: 
* https://github.com/unitycatalog/unitycatalog/pull/1261#discussion_r2599440427

My thought will incline towards exposing it as public.

clients/java/src/main/java/io/unitycatalog/client/HttpRetryHandler.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.IOException;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Set;
+
+/** Retry handler that retries on common recoverable HTTP errors and network exceptions. */
+public class HttpRetryHandler {
```
tdas: does this need to be public API. if not move to internal package.

clients/java/src/main/java/io/unitycatalog/client/internal/RetryingApiClient.java
```diff
@@ -0,0 +1,30 @@
...
+ * Adds retry handling on top of the generated {@link ApiClient}.
+ *
+ * <p>The base {@link ApiClient} comes from the OpenAPI generator and is overwritten on every regen.
+ * To keep that file untouched, this subclass overrides {@link #getHttpClient()} and wraps the
+ * generated {@link HttpClient} in a {@link RetryingHttpClient} configured through {@link
+ * RetryPolicy} and {@link HttpRetryHandler} (which defines the retry mechanism). Adding retries at
+ * the HTTP layer means every generated API surface (for example {@code TemporaryCredentialsApi},
+ * {@code TablesApi}, etc.) inherits the same policy without extra wrappers.
+ */
+public class RetryingApiClient extends ApiClient {
```
tdas, line 17: again.. does this need to be public class? if not move to internal package.

clients/java/src/main/java/io/unitycatalog/client/RetryingHttpClient.java
```diff
@@ -1,4 +1,4 @@
-package io.unitycatalog.spark;
+package io.unitycatalog.client;
```
tdas: same as above. move to internal package.

clients/java/src/main/java/io/unitycatalog/client/auth/UCTokenProvider.java
openinx: Okay, that make sense to me. Let me simplify those naming.

clients/java/src/main/java/io/unitycatalog/client/utils/Clock.java
tdas: move the utils package inside the internal package.

clients/java/src/main/java/io/unitycatalog/client/auth/AuthProps.java
tdas: - I understand TokenProvider needs to be public. but please put all implementations in internal package.

> Catalog level properties: we need to use those keys as the suffix to generate the catalog level properties

This is a very Spark-specific conf-based mechanism. Some other engine may already define a different way and different names for finding out user configs. Furthermore, the connector is configuring the TokenProvider by specifying the parameters directory, not by property name. So the most connectors will not need the property names.. and the property names can be private. 

> hadoop configuration: we want to use this suffix to propragate the UCTokenProvider properties to the underlying FileSystem instances, with fs.unitycatalog. prefix added.

Again, this is a very spark-specific detail that we are using hadoopconf to transfer this information to executors. but i agree that other connectors may want to do something similar. but in that case, the connector dont really care what are those properties as long as they can get a map of KV pairs that they need to serialize/deserialize. So they do not need to know the actually property names ... if `UCTokenProvider#properties()` return `k --> v` then we can just blindly append `fs.unitycatalog.auth.` to every key and set in hadoop. And then in the executor, blindly extracting all the ``fs.unitycatalog.auth.*` key-values from hadoopConf, and remove `fs.unitycatalog.auth.` from the keys, and build a new client with these kv pairs. 

In other words, I dont think UCAuthProps should be public API as connector should not have to know about the key names, that just token provider implementation detail.
And uc-spark should not reuse those fields. uc-spark should continue defining its own properties that makes sense of spark-based configuration mechanisms.

clients/java/src/main/java/io/unitycatalog/client/retry/RetryPolicy.java
openinx: Hi @tdas , here for the `io.unitycatalog.client.retry.RetryPolicy`, I think it's a general and common `RetryPolicy`  interface, which is not binded to http `request`.   

Essentially,  the `RetryPolicy` does not define any retry actions, it simply defined the `maxAttempts`, `sleepTime`, `allowRetry`, while how to retry the action is still defined in `HttpRetryHandler`.

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
openinx: I think it's the same discussion from here: https://github.com/unitycatalog/unitycatalog/pull/1261/files#r2599440427

clients/java/src/test/java/io/unitycatalog/client/auth/UCTokenProviderTest.java
openinx: Done

clients/java/src/test/java/io/unitycatalog/client/auth/UCTokenProviderTest.java
openinx: Done

clients/java/src/main/java/io/unitycatalog/client/auth/UCTokenProvider.java
tdas: discussed in that thread. i agree this should be public. but not the implementations. resolving this thread.

clients/java/src/main/java/io/unitycatalog/client/auth/AuthProps.java
openinx: Yes, agreed ! 

I think the downstream usages don't need to perceive those detailed properties,  what they need is: those are abstracted properties, but they shouldn't know what exactly are those keys. 

Just remove the `public` modifier for the `AuthProps` classes.

clients/java/src/test/java/io/unitycatalog/client/UserAgentTest.java
tdas: This code is supposed to change right? we need to remove the setAppVersion API from the APIClient. 
That can be a follow up PR though.. this PR is already big.

clients/java/src/main/java/io/unitycatalog/client/utils/Clock.java
openinx: Done

clients/java/src/main/java/io/unitycatalog/client/RetryingHttpClient.java
openinx: Done

clients/java/src/main/java/io/unitycatalog/client/internal/RetryingApiClient.java
openinx: The problem is:  currently the `RetryingApiClient` was depended by both `io.unitycatalog.client.OAuthTokenProvider` and `io.unitycatalog.client.ApiClientBuilder`.  So have to be exposed as cross-package access. 

I just moved it to the `internal` package now, I think we can refactor this in further PR so that the code can be more clear. 

Done.

clients/java/src/main/java/io/unitycatalog/client/HttpRetryHandler.java
openinx: Removed the `public` modifier, and also moved to `internal` package.

clients/java/src/main/java/io/unitycatalog/client/auth/UCAuthProps.java
openinx: Done, removed all the extra `UC` prefixed naming.

clients/java/src/main/java/io/unitycatalog/client/auth/UCTokenProvider.java
openinx: Done

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
openinx: Done

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
```diff
@@ -0,0 +1,99 @@
...
+  String accessToken();
+
+  /**
+   * Returns the configuration properties associated with this token provider.
+   *
+   * @return a map of configuration key-value pairs
+   */
+  Map<String, String> properties();
+
+  static Builder builder() {
```
tdas: didnt we decide offline that TokenProvider does not need a builder?

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
```diff
@@ -0,0 +1,99 @@
...
+public interface TokenProvider {
+  /**
+   * Returns the access token for Unity Catalog authentication.
+   *
+   * @return the access token string
+   */
+  String accessToken();
+
+  /**
+   * Returns the configuration properties associated with this token provider.
```
tdas: there is more context here. these are all the properties needed to initialize a token provider

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
```diff
@@ -0,0 +1,99 @@
...
+      setIfPresent(options, AuthProps.TOKEN, this::token);
+      setIfPresent(options, AuthProps.OAUTH_URI, this::oauthUri);
+      setIfPresent(options, AuthProps.OAUTH_CLIENT_ID, this::oauthClientId);
+      setIfPresent(options, AuthProps.OAUTH_CLIENT_SECRET, this::oauthClientSecret);
+      return this;
+    }
+
+    public TokenProvider build() {
+      if (token != null) {
+        return new FixedTokenProvider(token);
```
tdas: the user may have provided both token as well as oauth stuff. thats a clearly a bad usages of the builder and we sohuld block those kind of bad usage. not blocking bad usage can cause surprising bugs in the future.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public Builder oauthClientId(String oauthClientId) {
+      this.oauthClientId = oauthClientId;
+      return this;
+    }
+
+    public Builder oauthClientSecret(String oauthClientSecret) {
+      this.oauthClientSecret = oauthClientSecret;
+      return this;
+    }
```
tdas: why do you need three methods for taking 3 parameter that either always need to be provided .. or never need to be provided? having three methods does not ensure at compile time that the user is calling all 3 of them, thus causing runtime failure instead of compile time failures. having a single method that takes all 3 params ensure compile time safety, no one can accidentally specify only 1 or 2 of the params and fail later at runtime.

connectors/spark/src/test/java/io/unitycatalog/spark/ApiClientFactoryTest.java
```diff
@@ -3,110 +3,79 @@
...
-    // Verify the client is configured with the correct URI components
-    assertThat(client.getBaseUri()).contains("https://example.com:8443");
-
-    // Verify user agent is still set with at least Spark
-    assertThat(client.getUserAgent()).startsWith("UnityCatalog-Java-Client/");
-    assertThat(client.getUserAgent()).contains("Spark");
-  }
-
-  @Test
-  public void testApiClientBaseUri() {
```
openinx, line 95: Those two tests are moved to `ApiClientBuilderTest`.  Since there are common tests for the general builder, not spark-specific tests.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
openinx: For this abstracted `properties()`,  I  incline to not explain every key here, because we don't expect downstream need to perceive those properties. 

But for `createFromConfigs`,  since people need to know how to set them individually. so I added the detailed doc there.

clients/java/src/main/java/io/unitycatalog/client/internal/HttpRetryHandler.java
```diff
@@ -0,0 +1,98 @@
...
+import java.io.IOException;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Set;
+
+/** Retry handler that retries on common recoverable HTTP errors and network exceptions. */
+class HttpRetryHandler {
```
openinx, line 13: Moved from the original uc-spark connector to this `unitycatalog-client` module.

clients/java/src/main/java/io/unitycatalog/client/internal/HttpRetryHandler.java
```diff
@@ -0,0 +1,98 @@
...
+    if (isRecoverableException(e)) {
+      return true;
+    }
+    for (Throwable cause = e.getCause(); cause != e && cause != null; cause = cause.getCause()) {
+      if (isRecoverableException(cause)) {
+        return true;
+      }
+    }
+    return false;
+  }
```
openinx, line 97: Remove the `guava` dependency here,  just go-through the cause chain and see if there is any expected error.

clients/java/src/test/java/io/unitycatalog/client/UserAgentTest.java
openinx: Yes, I think it can be a separate following PR.

clients/java/src/main/java/io/unitycatalog/client/internal/HttpRetryHandler.java
openinx: https://github.com/unitycatalog/unitycatalog/pull/1261/files#diff-fd7a1421dd2d751f9d980bb7a5ad66276e6d92d4b3ee1e97c6368d223c0a6d9bL18

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+   * Sets client version metadata as name-version pairs.
+   *
+   * <p>Arguments must be provided in alternating name-version order with an even count, e.g.,
+   * {@code "MyApp1", "0.1.0", "MyApp2", "1.0.2"}.
+   *
+   * @param nameVersionPairs the metadata as alternating name-version pairs
+   * @return this builder instance for method chaining
+   * @throws IllegalArgumentException if an odd number of arguments is provided
+   */
+  public ApiClientBuilder clientVersion(String... nameVersionPairs) {
```
tdas: This is not client version. APIClient version is 0.3.  .. thats not what you want users to set. why not use AppVersion as i had used?

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -37,8 +37,11 @@ public T uri(String uri) {
       return self();
     }
 
-    public T ucTokenProvider(UCTokenProvider ucTokenProvider) {
-      builder.putAll(ucTokenProvider.properties());
+    public T ucTokenProvider(TokenProvider ucTokenProvider) {
```
tdas: maybe follow up PR to clean this ... shouldnt be using ucTokenProvider, just tokenProvider

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
```diff
@@ -1,44 +1,35 @@
...
-    // Base path in ApiClient is already set to `BASE_PATH`, so we override it to provide
-    // base path from given `url` but still preserving path suffix.
-    // Expected input for `url` is URL with no "/api/2.1/unity-catalog" in the path.
-    String basePath = url.getPath() + BASE_PATH;
-    RetryingApiClient apiClient = new RetryingApiClient(clientConf, Clock.systemClock());
-    apiClient.setHost(url.getHost())
-        .setPort(url.getPort())
-        .setScheme(url.getScheme())
-        .setBasePath(basePath);
+      RetryPolicy retryPolicy, URI url, TokenProvider ucTokenProvider) {
```
tdas: tokenProvider

tdas: nit: since we are changing it .. its weird that we are using `URI` as type and `url` as variable name. why not make it `URI uri`

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
```diff
@@ -82,37 +87,39 @@ public String accessToken() {
...
+              .header("Authorization", "Basic " + encodedCredentials)
+              .header("Content-Type", "application/x-www-form-urlencoded")
+              .POST(HttpRequest.BodyPublishers.ofString(formData))
+              .build();
 
       // Send request with retry support
-      HttpResponse<String> response = httpClient.send(request,
-          HttpResponse.BodyHandlers.ofString());
+      HttpResponse<String> response =
+          apiClient.getHttpClient().send(request, HttpResponse.BodyHandlers.ofString());
```
openinx: Oh, key mistake here:   We cannot call the `apiClient.getHttpClient()` for every request, since it will build a new http client.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
tdas: i think the only reason any one should use createFromConfigs() method is if they have gotten the configurations from another `tokenProvider.getConfigs()`. These configuration should be treated as a black box as they are really implementation details how implementations of TokenProvider interface have chosen to represent its configs in a map. 

In fact, I am of the opinion that we shoudl explicitly write it in the method docs of `createFromConfigs` and `getConfigs` that do not assume the configuration keys to be stable.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
```diff
@@ -0,0 +1,149 @@
...
+   * @return the access token string
+   */
+  String accessToken();
+
+  /**
+   * Returns the configuration properties associated with this token provider.
+   *
+   * @return a map of configuration key-value pairs
+   */
+  Map<String, String> properties();
```
tdas: why is this not getConfigs()? the output of this will be sent to the function `createFromConfigs` right? so lets be consistent with our names

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Builder to create configured {@link ApiClient} instances for Unity Catalog operations.
+ *
+ * <p>Example usage:
+ *
+ * <pre>{@code
+ * ApiClient client = ApiClientBuilder.create()
+ *     .url("http://localhost:8080")
+ *     .tokenProvider(TokenProvider.builder().token("my-token").build())
```
openinx: Oh, we need to use `TokenProvider.create("my-token")` here.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
```diff
@@ -0,0 +1,149 @@
...
+   * }</pre>
+   *
+   * @param configs a map containing authentication configuration parameters
+   * @return a token provider configured according to the provided parameters
+   * @throws IllegalArgumentException if both token and OAuth configs are provided, if OAuth configs
+   *     are incomplete (only some OAuth parameters are set), or if no valid authentication
+   *     configuration is found
+   * @throws NullPointerException if configs is null
+   */
+  static TokenProvider createFromConfigs(Map<String, String> configs) {
```
tdas, line 119: a more well-abstracted version of this implementation is as follows:
- each provider implementation has a createFromConfigs() method which returns `Optional<TokenProvider>`. What props each provider uses is completely contained within the provider. 
- TokenProvider.createFromConfigs() calles every providers.createFromConfig() and gets either None or Some(tokenProvider)
- if only one tokenProvider was found, return that. if none or multiple found, then then throw error due to incomplete or ambiguous configs.

Then the details of the what the props are, are completely contained within each implementation. No need for a central place like to understand the details of every implementation. And with that you can simply avoid these sort of quadratic-complexity checks - config-set-1 is set and config-set-2...N are all not set. 

wdyt?

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
openinx: Done,  I added the check in `createFromConfigs` to forbide set both keys.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/CredPropsUtil.java
openinx: Done, renamed all the `ucTokenProvider` to `tokenProvider`.

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
openinx: Make sense to me, done !

connectors/spark/src/main/scala/io/unitycatalog/spark/ApiClientFactory.java
openinx: Done

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+  public ApiClientBuilder appVersion(String... nameVersionPairs) {
+    Preconditions.checkArgument(
+        nameVersionPairs.length % 2 == 0,
+        "Must provide an even number of arguments for the name-version pairs.");
+    this.nameVersionPairs = nameVersionPairs;
+    return this;
+  }
+
+  /**
+   * Sets a custom retry policy for handling transient failures.
```
tdas, line 102: request retry policy

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+   * @return this builder instance for method chaining
+   * @see TokenProvider
+   */
+  public ApiClientBuilder tokenProvider(TokenProvider tokenProvider) {
+    this.tokenProvider = tokenProvider;
+    return this;
+  }
+
+  /**
+   * Sets application version metadata as name-version pairs.
```
tdas, line 84: add to the docs that this appversion is set as the user agent using this format - https://www.rfc-editor.org/rfc/rfc9110#name-user-agent

tdas (approved):
LGTM. but please address the last set of comments.

tdas:
since there is going to be a follow up PR.. please address comments in that PR. merging this to unblock that PR.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
openinx: I think iceberg already have the mature solution to solve this before. I mean if there are different implementations that need to be loaded from a `Map<String, String> conf`.  And how to abstract to load the implementation without perceive those conf keys. 

https://github.com/apache/iceberg/blob/11f3dcb3011fe5e37a4ec57fbc7a0c859ad5f24e/api/src/main/java/org/apache/iceberg/io/FileIO.java#L110-L115

The idea is simple and similar to yours idea.  The interface needs a `void initialize(Map<String, String> options)`,  and the factory will load the implementation based on the configured options. And the whole abstraction layer does not need to see and perceive those detailed keys. 

Let's adopt that approach.  Update in the following PR.

PR #1264, opened 2025-12-04
Require setting delta.feature.catalogOwned-preview for managed table creation by user.

yili-db:
**PR Checklist**

**Description of changes**

Require setting delta.feature.catalogOwned-preview for managed table creation by user.

Stop automatically setting this feature property in UCSingleCatalog when creating a UC managed table. Instead require user to specify it explicitly.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -112,17 +112,16 @@ class UCSingleCatalog
       // Setting the catalogManaged table feature is required for creating a managed table.
       if (!properties.containsKey(UCTableProperties.DELTA_CATALOG_MANAGED_KEY) &&
         !properties.containsKey(UCTableProperties.DELTA_CATALOG_MANAGED_KEY_NEW)) {
-        // Setting the old property name is accepted during the feature rename transition period.
-        // So in this error message only the current feature name is included.
         throw new ApiException(
           s"Managed table creation requires table property " +
-            s"'${UCTableProperties.DELTA_CATALOG_MANAGED_KEY}'=" +
+            s"'${UCTableProperties.DELTA_CATALOG_MANAGED_KEY_NEW}'=" +
             s"'${UCTableProperties.DELTA_CATALOG_MANAGED_VALUE}'" +
             s" to be set.")
       }
       // Caller should not set these two table properties to values other than "supported". This is
       // the only documented value.
-      List(UCTableProperties.DELTA_CATALOG_MANAGED_KEY, UCTableProperties.DELTA_CATALOG_MANAGED_KEY_NEW)
+      List(UCTableProperties.DELTA_CATALOG_MANAGED_KEY,
+        UCTableProperties.DELTA_CATALOG_MANAGED_KEY_NEW)
         .foreach(k => {
           Option(properties.get(k))
             .filter(_ != UCTableProperties.DELTA_CATALOG_MANAGED_VALUE)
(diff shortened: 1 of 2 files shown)


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -136,9 +145,6 @@ class UCSingleCatalog
       // Sets both the new and old table ID property while it's being renamed.
       newProps.put(UCTableProperties.UC_TABLE_ID_KEY, stagingTableInfo.getId)
       newProps.put(UCTableProperties.UC_TABLE_ID_KEY_OLD, stagingTableInfo.getId)
-      // Only set the existing feature name. When Delta renames it, Delta needs to handle it
```
yili-db, line 140: Stop setting it automatically.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -108,17 +108,26 @@ class UCSingleCatalog
...
         TableCatalog.PROP_IS_MANAGED_LOCATION)
         .filter(properties.containsKey(_))
-        .foreach(p => throw new ApiException(s"Cannot specify property $p."))
-      // Caller should not set this table property directly. But it's OK if it happens to set
-      // this property to the exactly value of what we need.
-      // This is because some document may have mentioned setting it to enable UC as commit
-      // coordinator. But we don't actually need that as we always set it automatically.
+        .foreach(p => throw new ApiException(s"Cannot specify property '$p'."))
+      // Setting the catalogManaged table feature is required for creating a managed table.
+      if (!properties.containsKey(UCTableProperties.CATALOG_MANAGED_KEY) &&
```
yili-db: Require the user to set it (either the current one or the new one)

connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
```diff
@@ -29,46 +29,52 @@
  */
 public abstract class ManagedTableReadWriteTest extends BaseTableReadWriteTest {
   private static final String DELTA_TABLE = "test_delta";
+  protected static final String TBLPROPERTIES_CLAUSE =
+      String.format(
+          "TBLPROPERTIES ('%s'='%s')",
+          UCTableProperties.CATALOG_MANAGED_KEY, UCTableProperties.CATALOG_MANAGED_VALUE);
```
yili-db: In these integration test only the old feature name can be used because it's testing against Delta 4.0.0.

connectors/spark/src/test/java/io/unitycatalog/spark/ManagedTableReadWriteTest.java
```diff
@@ -29,46 +29,52 @@
...
-        .hasMessageContaining(UCTableProperties.UC_TABLE_ID_KEY_OLD);
+                    "CREATE TABLE %s(name STRING) USING parquet %s",
+                    fullTableName, TBLPROPERTIES_CLAUSE))
+        .hasMessageContaining("not support non-Delta managed table");
+    for (String featureProperty :
+        List.of(UCTableProperties.CATALOG_MANAGED_KEY, UCTableProperties.CATALOG_MANAGED_KEY_NEW)) {
+      assertThatThrownBy(
+              () ->
+                  sql(
+                      "CREATE TABLE %s(name STRING) USING delta TBLPROPERTIES ('%s' = 'disabled')",
```
tdas, line 62: can be a follow up PR in the interest of 0.3.1, but it seems create table SQL command is showing up in a lot of places and it might be a good idea to consolidate all of managed tables ones since they have this common theme of setting a property. might be easier in the future to fix stuff when we wont want the table property to be set.

yili-db: I have a PR to consolidate the SQL commands and make the table format/managed/external abstract:
https://github.com/unitycatalog/unitycatalog/pull/1244

However this test is special as it does not exercise the table creation in the normal way but instead it tries all error cases. So it has to run its own hand written sql statements.

PR #1268, opened 2025-12-08
Validation for ServerProperties.

yili-db:
**PR Checklist**

**Description of changes**
Validation for ServerProperties.

ServerProperties is implemented as Map<String, String> to provide server the essential settings. Before this PR if a invalid value is set for a property it will only be checked and fail when that settings is actually used. This PR would proactively validate property values to report errors and crash when the server starts.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
+++ b/server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
@@ -2,18 +2,27 @@ package io.unitycatalog.server.utils;
 
 import io.unitycatalog.server.exception.BaseException;
 import io.unitycatalog.server.exception.ErrorCode;
+import io.unitycatalog.server.persist.utils.FileOperations;
 import io.unitycatalog.server.service.credential.aws.S3StorageConfig;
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -23,19 +31,153 @@ public class ServerProperties {
...
+        throw new BaseException(
+            ErrorCode.INVALID_ARGUMENT,
+            String.format(
+                "Invalid storage path '%s' for property '%s': %s", value, key, e.getMessage()));
+      }
+    }
+  }
+
+  /** Validator for positive integer properties */
+  private static class PositiveIntegerValidator implements PropertyValidator {
```
tdas: why cant this be just a integer validator .. which a integer condition. 
i can imagine we will need a param with 0 as valid. so a base class IntegerValidator and field POSITIVE_INTEGER_VALIDATOR = new IntegerValidator(<<condition > 0>>) makes it more extensible

tdas (approved):

PR #1269, opened 2025-12-09
Improve the abstracted unitycatalog-client API.

openinx:
**PR Checklist**

**Description of changes**

The related issue is : https://github.com/unitycatalog/unitycatalog/issues/1259

This PR is the follow up of this one:  https://github.com/unitycatalog/unitycatalog/pull/1261

In this PR I did the following things:

*  rename the `appVersion` to `addAppVersion`, which can append the app version for multiple calls. 
*  Rename the `FixedTokenProvider` to `StaticTokenProvider`,which make more sense. 
*  Add the `initializate(Map<String, String> configs>` in the abstracted TokenProvider interface to initialize the provider instance.   The core problem is:  if we use the abstraction way to initialize the provider, then we have to introduce the `auth.type`, which can be `static`, `oauth`, `xxxx.CustomizedTokenProvider`. With this, the abstracted `TokenProvider#createFromConfigs` can dynamically load the provider instance from the `Map<String, String>`.  Otherwise, the `TokenProvider` can still have no abstracted approach to load the instance, because it doesnot know which to load.  Another point is:  we've already have released the version, which support `uri` and `token` , so we will use the `auth.type=static` by default if people did not specify the `auth.type`. With this way, we can be compatible to the old version. 
* Removed the original customized ApiClient template generation, since currently we can customize the requestInteceptor directly to inject those `User-Agent` information.


clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
```diff
@@ -9,11 +8,14 @@
...
  *
  * <ul>
- *   <li>{@link FixedTokenProvider} - uses a pre-configured static token
+ *   <li>{@link StaticTokenProvider} - uses a pre-configured static token
  *   <li>{@link OAuthTokenProvider} - obtains tokens via OAuth 2.0 client credentials flow
  * </ul>
  */
 public interface TokenProvider {
+
+  void initialize(Map<String, String> configs);
```
openinx, line 24: Add the `initializate(Map<String, String> configs>` in the abstracted TokenProvider interface to initialize the provider instance.   The core problem is:  

* if we use the abstraction way to initialize the provider, then we have to introduce the `auth.type`, which can be `static`, `oauth`, `xxxx.CustomizedTokenProvider`. With this, the abstracted `TokenProvider#createFromConfigs` can dynamically load the provider instance from the `Map<String, String>`.  Otherwise, the `TokenProvider` can still have no abstracted approach to load the instance, because it doesnot know which to load.  
 

Another point is:  we've already have released the version, which support `uri` and `token` , so we will use the `auth.type=static` by default if people did not specify the `auth.type`. With this way, we can be compatible to the old version.

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
```diff
@@ -26,12 +31,14 @@
  * @see RetryPolicy
  */
 public class ApiClientBuilder {
+  private static final String DEFAULT_APP_NAME = "UnityCatalog-Java-Client";
   private static final String BASE_PATH = "/api/2.1/unity-catalog";
 
   private URI uri = null;
   private TokenProvider tokenProvider = null;
-  private String[] nameVersionPairs = null;
+  private final Map<String, String> appVersionMap = new LinkedHashMap<>();
```
openinx, line 40: I need to use a linked hashMap here to keep the order of insertion. Also use a `map<String, String>` rather than the `String[]`, because map is more friendly to do java streaming operation.

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
```diff
@@ -84,22 +96,24 @@ public ApiClientBuilder tokenProvider(TokenProvider tokenProvider) {
...
    *
-   * @param nameVersionPairs the metadata as alternating name-version pairs
+   * @param name the application name.
+   * @param version the application version.
    * @return this builder instance for method chaining
-   * @throws IllegalArgumentException if an odd number of arguments is provided
+   * @throws IllegalArgumentException if any argument is null or empty.
    */
-  public ApiClientBuilder appVersion(String... nameVersionPairs) {
+  public ApiClientBuilder addAppVersion(String name, String version) {
```
openinx, line 107: The `addAppVersion(name, version)` is more simple than the `appVersion(String...nameVersionPairs)`, since we can easily to check every `name` and `version` when call addAppVersion method, report the exact error message for every call. 

While using the `String[]`, is actually hard to report the clear error message if there is any argument checking invalidation.

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
```diff
@@ -113,30 +127,47 @@ public ApiClientBuilder retryPolicy(RetryPolicy retryPolicy) {
     return this;
   }
 
+  public ApiClientBuilder addRequestInterceptor(Consumer<HttpRequest.Builder> interceptor) {
```
openinx, line 156: Organizing the `interceptor` as a queue with the insertion order, and then intercepting one by one.  This will help a lot for multiple interceptions.

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
tdas: the downside of this style of API is that we can never remove anything once added. 
which is totally okay in builders..... but otherwise maybe problematic.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
tdas: you have not added the docs we talkd about

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
```diff
@@ -117,33 +134,26 @@ static TokenProvider createFromOAuthConfigs(
    * @throws NullPointerException if configs is null
```
tdas: you havent added the docs we talked about.

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
```diff
@@ -26,12 +31,14 @@
  * @see RetryPolicy
  */
 public class ApiClientBuilder {
+  private static final String DEFAULT_APP_NAME = "UnityCatalog-Java-Client";
```
tdas: nit: technically this is not AppName. its client name.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
tdas: okay. as long as the Spark users dont have to specify another configuration for type. internally we can make it as clean as we want as long as we hide the complexity from the user. for the Spark user setting spark configuration, the behavior should be intuitive 
- if only token is specified, use static provider
- if only the all three oauth is specified, use oauth
- any other combination throw an error.

tdas: do you have a test with Spark for the 3rd case already? can you point me to it?

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -233,6 +234,7 @@ public static Stream<Arguments> credentialRenewalArguments() {
                     baseLocation,
                     Map.of(
                         OptionsUtil.URI, SERVER_URL,
+                        UC_PROPS_AUTH_TYPE, "static",
```
tdas: why does the credential renewal test need to change? are you requiring that the user specifies this type as well?

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
openinx: Yes,   let me rename it to `DEFAULT_CLIENT_NAME`.

clients/java/src/main/java/io/unitycatalog/client/ApiClientBuilder.java
openinx: Agreed, for builder we don't need to remove anything.  But for other cases,  it may not the best choice.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Yes, for this version. We still need the user to specify the `auth.type` in the spark job. Because without this key, the `TokenProvider#createFromConfigs` does not know which is the correct `TokenProvider` to choose under the non TokeProvider's properties perceiving limitation.

clients/java/src/main/java/io/unitycatalog/client/auth/AuthConfigs.java
```diff
@@ -0,0 +1,21 @@
+package io.unitycatalog.client.auth;
+
+class AuthConfigs {
```
openinx, line 4: I think it's more intuitive to rename the original `AuthProps` to `AuthConfigs`, since those are the config keys to use.

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/AuthConfigUtils.java
```diff
@@ -0,0 +1,36 @@
...
+        newConfigs.put(newKey, e.getValue());
+      }
+    }
+
+    // Unity Catalog versions 0.3.0 and earlier did not use the 'auth.token' key. To maintain
+    // backward compatibility, we also copy the legacy 'token' key directly into the new config map.
+    String token = configs.get(AuthConfigUtils.STATIC_TOKEN);
+    if (token != null) {
+      newConfigs.put(TYPE, STATIC_TYPE);
+      newConfigs.put(STATIC_TOKEN, token);
```
tdas, line 37: should we throw error if both token and auth.token is specified? 
i think we should.. because otherwise its confusing which takes precendence.
can be a follow up PR, its a minor fix.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
```diff
@@ -9,141 +9,135 @@
...
-    if (oauthUri != null || oauthClientId != null || oauthClientSecret != null) {
-      Preconditions.checkArgument(
-          oauthUri != null && oauthClientId != null && oauthClientSecret != null,
-          "Incomplete OAuth configuration detected. All of the keys are required: "
-              + "oauthUri, oauthClientId, oauthClientSecret. Please ensure they are "
-              + "all set.");
+      case AuthConfigs.OAUTH_TYPE:
+        OAuthTokenProvider oauthTokenProvider = new OAuthTokenProvider();
+        oauthTokenProvider.initialize(configs);
+        return oauthTokenProvider;
```
tdas: why not move the initialize to outside this switch case? its common to all cases.

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
```diff
@@ -86,11 +106,12 @@ public String accessToken() {
   }
 
   @Override
-  public Map<String, String> getConfigs() {
+  public Map<String, String> configs() {
     return Map.of(
-        AuthProps.OAUTH_URI, oauthUri,
-        AuthProps.OAUTH_CLIENT_ID, oauthClientId,
-        AuthProps.OAUTH_CLIENT_SECRET, oauthClientSecret);
+        AuthConfigs.TYPE, AuthConfigs.OAUTH_TYPE,
```
openinx: It must return the `type` to reconstruct the token provider.

clients/java/src/test/java/io/unitycatalog/client/auth/TokenProviderUtils.java
```diff
@@ -0,0 +1,20 @@
+package io.unitycatalog.client.auth;
+
+import java.util.Map;
+
+public class TokenProviderUtils {
```
tdas, line 5: why is this public? 
also please document in the docs that this is internal class not meant for public consumption

openinx: This is for testing purpose,  maintained in `test` module, not production code. Just used for easiliy constructing the  TokenProvider.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
openinx: As  we discussed online,  we are aligned on the new approach, but also keep the compatibility for the old version usages. The new approach we will support the following keys to configure the spark:

```sql
# For static token
spark.sql.catalog.<name>.auth.type=static
spark.sql.catalog.<name>.auth.token=my-token

# For old version backward compatibily with static token
spark.sql.catalog.<name>.token=my-token

# For OAuth token
spark.sql.catalog.<name>.auth.type=oauth
spark.sql.catalog.<name>.auth.oauth.uri=http://example.com
spark.sql.catalog.<name>.auth.oauth.clientId=my-client-id
spark.sql.catalog.<name>.auth.oauth.clientSecret=my-client-secret

For custom token provider
spark.sql.catalog.<name>.auth.type=com.CustomTokenProvider
spark.sql.catalog.<name>.auth.<key>=xxxxx
```

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
openinx: Done, add the doc for the connection between `TokenProvider#configs` and `TokenProvider#create`.

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
openinx: Done, add the doc for the connection.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Done

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/AuthConfigUtils.java
openinx: Yes,  I think you are right !  Updated the PR, and add the precondition check, along with the test coverage.

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
```diff
@@ -28,24 +28,16 @@ class OAuthTokenProvider implements TokenProvider {
   private static final long DEFAULT_LEAD_RENEWAL_TIME_SECONDS = 30L;
```
tdas, line 30: for all the package private classes, can you add in the docs that this is a internal class, not to be used directly.

clients/java/src/main/java/io/unitycatalog/client/auth/AuthConfigs.java
```diff
@@ -0,0 +1,21 @@
...
+  private AuthConfigs() {}
+
+  // Define the authentication type, where the type can be:
+  // 1. static: which uses the StaticTokenProvider to generate the static token.
+  // 2. oauth: which uses the OAuthTokenProvider to generate the oauth token.
+  // 3. fully qualified class name of the custom TokenProvider implementation.
+  static final String TYPE = "type";
+
+  // Configure keys for static token provider.
+  static final String STATIC_TYPE = "static";
```
tdas: `STATIC_TYPE_VALUE`. to differentiate from `TYPE` which is the key.

clients/java/src/main/java/io/unitycatalog/client/auth/AuthConfigs.java
```diff
@@ -0,0 +1,21 @@
...
+  static final String STATIC_TOKEN = "token";
+
+  // Configure keys for oauth token provider.
+  static final String OAUTH_TYPE = "oauth";
```
tdas: OAUTH_TYPE_VALUE

clients/java/src/main/java/io/unitycatalog/client/auth/AuthConfigs.java
tdas: https://github.com/unitycatalog/unitycatalog/pull/1269/files#r2604771088

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
tdas: Setting this to static should not be needed any more right. then lets not do it. this makes sense sure that we havent broken any user CUJ.

clients/java/src/test/java/io/unitycatalog/client/auth/TokenProviderTest.java
```diff
@@ -14,68 +14,147 @@ public class TokenProviderTest {
...
+  public void testCreateTokenProviderViaConfigs() {
+    // Test with valid token - should create StaticTokenProvider
+    TokenProvider tokenProvider = TokenProviderUtils.create("test-token");
+    assertThat(tokenProvider).isInstanceOf(StaticTokenProvider.class);
     assertThat(tokenProvider.accessToken()).isEqualTo("test-token");
 
     // Test with complete OAuth config - should create OAuthUCTokenProvider
     TokenProvider oauthProvider =
-        TokenProvider.createFromConfigs(
+        TokenProvider.create(
```
openinx: I think here we can use the testing `TokenProviderUtils` to simplfy the construction.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
```diff
@@ -43,9 +43,10 @@ private UCHadoopConf() {
   // custom configuration keys (e.g., AWS, ABFS, or GCS credentials) propagate correctly into Hadoop
   // FileSystems such as S3AFileSystem or AbfsFileSystem, the keys must include the "fs." prefix.
   // See also: https://github.com/unitycatalog/unitycatalog/issues/1112.
-  public static final String FS_UC_PREFIX = "fs.unitycatalog.";
+  public static final String FS_UC_AUTH_PREFIX = "fs.unitycatalog.auth.";
```
openinx: Seems like it make more sense to remove the `FS_` prefix for the naming.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Yes, that totally makes sense to me.  We just leverage the test to see the backward-compatibilty really works well !

clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
openinx: Done

clients/java/src/test/java/io/unitycatalog/client/auth/TokenProviderTest.java
openinx: Done

connectors/spark/src/main/scala/io/unitycatalog/spark/UCHadoopConf.java
openinx: Done

connectors/spark/src/main/scala/io/unitycatalog/spark/auth/AuthConfigUtils.java
openinx: Done, see line33.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
openinx: Done

clients/java/src/main/java/io/unitycatalog/client/auth/AuthConfigs.java
openinx: Done

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
openinx: Done for all classes.

clients/java/src/main/java/io/unitycatalog/client/auth/AuthConfigs.java
openinx: Done

clients/java/src/main/java/io/unitycatalog/client/auth/AuthConfigs.java
openinx: Done

PR #1284, opened 2025-12-29
Add URL overlap validation for external locations

yili-db:
**PR Checklist**

**Description of changes**

Add URL overlap validation for external locations

This commit implements comprehensive URL overlap prevention for external locations to ensure that no two external locations can have URLs that overlap in the directory hierarchy.

Key Changes:

1. URL Overlap Validation
   - Prevent creating external locations with URLs that overlap with existing ones
   - Check for exact matches, parent directories, and subdirectories
   - Apply validation on both create and update operations
   - Normalize all URLs before validation using FileOperations.toStandardizedURIString()

2. New ExternalLocationUtils Utility Class
   - Provides path-based database queries for Unity Catalog entities
   - Uses Hadoop's Path class for proper URI manipulation across schemes
   - Supports finding entities with overlapping URLs (self, parent, or subdir)

Testing:
- All existing tests pass
- Added comprehensive tests for URL overlap validation
- Added unit tests for ExternalLocationUtils helper methods
- Added tests for FileOperations.removeExtraSlashes()

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/ExternalLocationRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/ExternalLocationRepository.java
@@ -74,7 +74,7 @@ public class ExternalLocationRepository {
           }
 
           String url = FileOperations.toStandardizedURIString(createExternalLocation.getUrl());
-          validateUrlNotUsedByAnyExternalLocation(session, url);
+          validateUrlNotUsedByAnyExternalLocation(session, url, Optional.empty());
 
           CredentialDAO credentialDAO =
               validateAndGetCredentialDAO(session, createExternalLocation.getCredentialName());
@@ -172,7 +172,8 @@ public class ExternalLocationRepository {
           if (updateExternalLocation.getUrl() != null) {
             existingLocation.setUrl(
                 FileOperations.toStandardizedURIString(updateExternalLocation.getUrl()));
-            validateUrlNotUsedByAnyOtherExternalLocation(session, existingLocation);
+            validateUrlNotUsedByAnyExternalLocation(
+                session, existingLocation.getUrl(), Optional.of(existingLocation.getId()));
           }
           if (updateExternalLocation.getComment() != null) {
             existingLocation.setComment(updateExternalLocation.getComment());
@@ -211,7 +212,7 @@ public class ExternalLocationRepository {
   }
 
   /**
-   * Validates that no existing external location has a URL that overlaps with the given URL.
+   * Validates that no other external location has a URL that overlaps with the given URL.
    *
    * <p>This method ensures that external locations have non-overlapping URL hierarchies. It checks
    * for three types of overlap:
@@ -227,68 +228,40 @@ public class ExternalLocationRepository {
    *
    * @param session The Hibernate session for database access
    * @param url The URL to validate (must be standardized)
+   * @param currentExternalLocationId The ID of the current external location it's validating. This
+   *     is only set for updating an existing external location. For creating new ones it's empty.
    * @throws BaseException if an overlapping external location exists
    */
-  private void validateUrlNotUsedByAnyExternalLocation(Session session, String url) {
+  private void validateUrlNotUsedByAnyExternalLocation(
+      Session session, String url, Optional<UUID> currentExternalLocationId) {
     ExternalLocationUtils.<ExternalLocationDAO>getEntitiesDAOsOverlapUrl(
             session,
             url,
             SecurableType.EXTERNAL_LOCATION,
-            /* limit= */ 1,
-            /* includeParent= */ true,
-            /* includeSelf= */ true,
-            /* includeSubdir= */ true)
-        .stream()
-        .findAny()
-        .ifPresent(
-            existingExternalLocation -> {
-              throw new BaseException(
-                  ErrorCode.INVALID_ARGUMENT,
-                  "An existing external location with URL '"
-                      + existingExternalLocation.getUrl()
-                      + "' already exists. A new external location should not duplicate or"
-                      + " overlap with existing external locations.");
-            });
-  }
-
-  /**
-   * Validates that no other external location has a URL that overlaps with the new URL.
-   *
-   * <p>This method is similar to {@link #validateUrlNotUsedByAnyExternalLocation}, but is used when
-   * updating an existing external location's URL. It filters out the location being updated itself,
-   * only checking for overlaps with other external locations.
-   *
-   * <p>This allows an external location to be updated to a URL that may technically "overlap" with
-   * its old URL (since they're the same entity), but prevents overlapping with any other external
-   * location.
-   *
-   * @param session The Hibernate session for database access
-   * @param locationWithNewUrl The external location DAO with the new URL to validate
-   * @throws BaseException if an overlapping external location exists (other than itself)
-   */
-  private void validateUrlNotUsedByAnyOtherExternalLocation(
-      Session session, ExternalLocationDAO locationWithNewUrl) {
-    ExternalLocationUtils.<ExternalLocationDAO>getEntitiesDAOsOverlapUrl(
-            session,
-            locationWithNewUrl.getUrl(),
-            SecurableType.EXTERNAL_LOCATION,
-            /* limit= */ 2,
+            // If it needs to find *another* besides the current one, it has to find at least 2.
+            /* limit= */ currentExternalLocationId.map(current -> 2).orElse(1),
             /* includeParent= */ true,
             /* includeSelf= */ true,
             /* includeSubdir= */ true)
         .stream()
         .filter(
             otherExternalLocation ->
-                !locationWithNewUrl.getId().equals(otherExternalLocation.getId()))
+                currentExternalLocationId
+                    .map(id -> !id.equals(otherExternalLocation.getId()))
+                    .orElse(true))
         .findAny()
         .ifPresent(
-            existingExternalLocation -> {
+            otherExternalLocation -> {
+              LOGGER.error(
+                  "Cannot have external location with a URL {} that overlaps/duplicates with "
+                      + "existing external location {} with url {}",
+                  url,
+                  otherExternalLocation.getName(),
+                  otherExternalLocation.getUrl());
               throw new BaseException(
                   ErrorCode.INVALID_ARGUMENT,
-                  "An existing external location with URL '"
-                      + existingExternalLocation.getUrl()
-                      + "' already exists. Cannot change the URL to duplicate or overlap with "
-                      + "existing external locations.");
+                  "Cannot accept an external location that duplicates or overlaps with existing "
+                      + "external location");
             });
   }
 }
--- a/server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
+++ b/server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
@@ -1,10 +1,7 @@
 package io.unitycatalog.server.base.externallocation;
 
-import static io.unitycatalog.server.utils.TestUtils.COMMENT;
-import static io.unitycatalog.server.utils.TestUtils.COMMENT2;
 import static io.unitycatalog.server.utils.TestUtils.assertApiException;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import io.unitycatalog.client.ApiException;
 import io.unitycatalog.client.model.AwsIamRoleRequest;
@@ -18,7 +15,9 @@ import io.unitycatalog.server.base.BaseCRUDTest;
 import io.unitycatalog.server.base.ServerConfig;
 import io.unitycatalog.server.base.credential.CredentialOperations;
 import io.unitycatalog.server.exception.ErrorCode;
+import java.util.HashSet;
 import java.util.Optional;
+import java.util.Set;
 import lombok.SneakyThrows;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
@@ -41,6 +40,7 @@ public abstract class BaseExternalLocationCRUDTest extends BaseCRUDTest {
   protected abstract CredentialOperations createCredentialOperations(ServerConfig config);
 
   protected CredentialInfo credentialInfo = null;
+  protected Set<String> externalLocationsToDelete = new HashSet<>();
 
   @SneakyThrows
   @BeforeEach
@@ -53,7 +53,6 @@ public abstract class BaseExternalLocationCRUDTest extends BaseCRUDTest {
     CreateCredentialRequest createCredentialRequest =
         new CreateCredentialRequest()
             .name(CREDENTIAL_NAME)
-            .comment(COMMENT)
             .purpose(CredentialPurpose.STORAGE)
             .awsIamRole(new AwsIamRoleRequest().roleArn(DUMMY_ROLE_ARN));
     credentialInfo = credentialOperations.createCredential(createCredentialRequest);
@@ -63,9 +62,24 @@ public abstract class BaseExternalLocationCRUDTest extends BaseCRUDTest {
   @AfterEach
   @Override
   public void tearDown() {
+    for (String externalLocationName : externalLocationsToDelete) {
+      externalLocationOperations.deleteExternalLocation(externalLocationName);
+    }
+    externalLocationsToDelete.clear();
     credentialOperations.deleteCredential(credentialInfo.getName());
     credentialInfo = null;
     super.tearDown();
+  }
+
+  protected ExternalLocationInfo create(String name, String url) throws ApiException {
+    ExternalLocationInfo externalLocationInfo =
+        externalLocationOperations.createExternalLocation(
+            new CreateExternalLocation().name(name).url(url).credentialName(CREDENTIAL_NAME));
+    externalLocationsToDelete.add(name);
+    assertThat(externalLocationInfo.getName()).isEqualTo(name);
+    assertThat(externalLocationInfo.getUrl()).isEqualTo(url);
+    assertThat(externalLocationInfo.getCredentialId()).isEqualTo(credentialInfo.getId());
+    return externalLocationInfo;
   }
 
   @Test
@@ -74,7 +88,6 @@ public abstract class BaseExternalLocationCRUDTest extends BaseCRUDTest {
     CreateExternalLocation createExternalLocation =
         new CreateExternalLocation()
             .name(EXTERNAL_LOCATION_NAME)
-            .comment(COMMENT)
             .url(URL)
             .credentialName("not_exist");
     // Fails as the credential does not exist
@@ -87,13 +100,7 @@ public abstract class BaseExternalLocationCRUDTest extends BaseCRUDTest {
         .noneMatch(
             externalLocationInfo -> externalLocationInfo.getName().equals(EXTERNAL_LOCATION_NAME));
 
-    createExternalLocation.setCredentialName(CREDENTIAL_NAME);
-    ExternalLocationInfo externalLocationInfo =
-        externalLocationOperations.createExternalLocation(createExternalLocation);
-    assertThat(externalLocationInfo.getName()).isEqualTo(EXTERNAL_LOCATION_NAME);
-    assertThat(externalLocationInfo.getComment()).isEqualTo(COMMENT);
-    assertThat(externalLocationInfo.getUrl()).isEqualTo(URL);
-    assertThat(externalLocationInfo.getCredentialId()).isEqualTo(credentialInfo.getId());
+    ExternalLocationInfo externalLocationInfo = create(EXTERNAL_LOCATION_NAME, URL);
 
     // List external locations
     assertThat(externalLocationOperations.listExternalLocations(Optional.empty()))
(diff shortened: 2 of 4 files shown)


server/src/main/java/io/unitycatalog/server/persist/dao/ExternalLocationDAO.java
```diff
@@ -49,29 +54,6 @@ public class ExternalLocationDAO extends IdentifiableDAO {
   @Column(name = "updated_by")
   private String updatedBy;
 
-  public static ExternalLocationDAO from(ExternalLocationInfo externalLocationInfo) {
```
yili-db, line 52: This isn't used anyway.

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -0,0 +1,213 @@
...
+   * @param includeSubdir If true, include entities whose URL is a subdirectory of the given URL
+   * @return List of matching entity DAOs, ordered by URL length descending
+   * @throws IllegalArgumentException if the securableType is not supported for URL overlap checks
+   */
+  public static <T extends IdentifiableDAO> List<T> getEntitiesDAOsOverlapUrl(
+      Session session,
+      String url,
+      SecurableType securableType,
+      int limit,
+      boolean includeParent,
```
yili-db, line 72: This function is called only with these 3 parameters all being true at this point.
Later it will be called with different values for these 3 parameters to achieve different purposes.

server/src/main/java/io/unitycatalog/server/persist/ExternalLocationRepository.java
```diff
@@ -197,4 +209,86 @@ public ExternalLocationDAO deleteExternalLocation(String name) {
...
+   * @param locationWithNewUrl The external location DAO with the new URL to validate
+   * @throws BaseException if an overlapping external location exists (other than itself)
+   */
+  private void validateUrlNotUsedByAnyOtherExternalLocation(
+      Session session, ExternalLocationDAO locationWithNewUrl) {
+    PathBasedRpcUtils.<ExternalLocationDAO>getEntitiesDAOsOverlapUrl(
+            session,
+            locationWithNewUrl.getUrl(),
+            SecurableType.EXTERNAL_LOCATION,
+            /* limit= */ 2,
```
yili-db: limit=2 so that it returns at least one other external location other than itself.

server/src/test/java/io/unitycatalog/server/persist/utils/FileOperationsTest.java
```diff
@@ -20,21 +17,52 @@ public void testToStandardizedURIString() {
             FileOperations.toStandardizedURIString(
                 "abfs://my-container@my-storage.dfs.core.windows.net/my-file"))
         .isEqualTo("abfs://my-container@my-storage.dfs.core.windows.net/my-file");
+    assertThat(FileOperations.toStandardizedURIString("gs://my-bucket///"))
```
tdas, line 58: why are these very detailed tests only for gs? shouldnt we test such "///" cases for all the schemes?

tdas: also .. i see that now there is a new test for all the slash handling 

https://github.com/unitycatalog/unitycatalog/pull/1284/changes#diff-d96d45ab8b0f367bdbe7d74312b6866a6832e080d9fb6ab94ee3c3d2481e14faR58

then do we need these tests as well?

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Stream;
+import org.apache.hadoop.fs.Path;
+import org.hibernate.Session;
+import org.hibernate.query.Query;
+
+/**
+ * Utility class for performing path-based database queries on Unity Catalog entities.
```
tdas, line 20: this seems like a very specific singular functionality that applies only to external locations. then why not name it ExternalLocationUtils?

yili-db: At this point it only searches for external locations by path. Later it will also search for tables, volumes, etc.
For example, when deleting an external location, it needs to find out if there's any table created under this external location and deny the deletion if any.

tdas: okay.. i see that its not tied to just ExternalLocation. Then, how about PathValidationUtils?
in fact, why not add this FileOperations instead of a new class? i am just trying to understand the justification of a new class, given FileOperations already has some functionality relevant for normalizing/validating. if there needs to be a new class, then we we should clarify (in the class docs of bth classes) what semantic differences are, so that we can reason in the future where new functionalities should go to.

server/src/main/java/io/unitycatalog/server/persist/utils/PathBasedRpcUtils.java
```diff
@@ -0,0 +1,213 @@
...
+            daoClassInfo.urlFieldName,
+            limit,
+            includeParent,
+            includeSelf,
+            includeSubdir)
+        .map(entity -> (T) entity)
+        .toList();
+  }
+
+  private static <T extends IdentifiableDAO> Stream<T> getEntitiesDAOsOverlapUrlHelper(
```
tdas: this seems like a single use very shallow function? why do we need a separate function at all?

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -0,0 +1,213 @@
...
+      boolean includeSelf,
+      boolean includeSubdir) {
+    Query<T> query =
+        generateEntitiesDAOsOverlapUrlQuery(
+            session, url, daoClass, urlFieldName, limit, includeParent, includeSelf, includeSubdir);
+    return query.stream();
+  }
+
+  @VisibleForTesting
+  static <T extends IdentifiableDAO> Query<T> generateEntitiesDAOsOverlapUrlQuery(
```
tdas, line 110: add docs please. non-trivial function.

server/src/test/java/io/unitycatalog/server/persist/utils/FileOperationsTest.java
yili-db: Added more tests.

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -0,0 +1,213 @@
...
+      queryConditions.add(String.format("%s LIKE :likePattern ESCAPE '\\'", urlFieldName));
+    } else {
+      likePattern = "";
+    }
+
+    String queryString =
+        String.format(
+            "FROM %s WHERE %s ORDER BY LENGTH(%s) DESC",
+            daoClass.getSimpleName(), String.join(" OR ", queryConditions), urlFieldName);
+
```
tdas, line 160: is this a safe way to construct queries .. that guards against sql injection, etc.?

server/src/main/java/io/unitycatalog/server/persist/utils/PathBasedRpcUtils.java
yili-db: Removed.

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
yili-db: Doc added. I moved the method doc of getEntitiesDAOsOverlapUrl here to avoid duplication. The docs of both methods will be almost same.

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
yili-db: Yes. Please refer to the unit tests: https://github.com/unitycatalog/unitycatalog/pull/1284/changes#diff-0c65f7ba81891972a3b7f8e9d92ffef46433dc9d917511af0782a1f54ae69219R198
The query is never inserting actual values as the condition directly. Instead the string manipulation only deals with constant string pieces like "url IN (:matchPaths)" and "url LIKE :likePattern ESCAPE '\'".
The actual URL values are still part of the ParameterValue, just like any other queries.

yili-db: I just refactored this code a bit to make it clear that query construction isn't taking any URL as part of the string.

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
yili-db: Technically it can be merged with FileOperations. However there are clear distinctions between these two classes:
`FileOperations`: operations on **cloud storage files**. Also does some path normalization.
`PathBasedRpcUtils`: operations on the **database**. searching for entities that has certain URLs. Uses hibernate session to do its job.
There is a little bit cross over between the two as both involve working with URLs. But majority of the scope is different.

As for `PathValidationUtils`: at this moment it's only doing validation for external locations. Later it will also be used for searching for an entity for other purposes. For example in any `Temporary*CredentialsService` they need to search for an external location along with its storage credential in order to properly vend a credential.

tdas: okay. i like that clear distinction. now i am thinking the names should 
- FileOperations (maybe rename to PathUtils) - utilities common to all paths, unrelated to exernal locations or not.
- ExternalLocationUtils - utils specific to searching through paths under registered external locations (ignore the fact that root location stuff, we will eventually deprecate that).

wdyt?

server/src/test/java/io/unitycatalog/server/persist/utils/PathBasedRpcUtilsTest.java
```diff
@@ -0,0 +1,359 @@
...
+    assertThat(s3BucketRootResult).isEmpty();
+
+    // Test S3 single path level
+    List<String> s3SingleLevelResult = PathBasedRpcUtils.getParentPathsList("s3://bucket/path");
+    assertThat(s3SingleLevelResult).containsExactly("s3://bucket");
+
+    // Test Azure Blob Storage path
+    List<String> azureResult =
+        PathBasedRpcUtils.getParentPathsList(
+            "abfs://container@storage.dfs.core.windows.net/path/to/file");
```
tdas: no test for trailing / ?

tdas: or multiple slashes all anywhere?

server/src/test/java/io/unitycatalog/server/persist/utils/ExternalLocationUtilsTest.java
```diff
@@ -0,0 +1,359 @@
...
+                    SecurableType.SCHEMA,
+                    1,
+                    /* includeSelf= */ true,
+                    /* includeParent= */ true,
+                    /* includeSubdir= */ true))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessageContaining("Unsupported securable type for URL overlap check");
+  }
+
+  private void validateQuery(
```
tdas, line 271: add docs on what about the query does it validate?

server/src/test/java/io/unitycatalog/server/persist/utils/PathBasedRpcUtilsTest.java
```diff
@@ -0,0 +1,359 @@
...
+
+  @Test
+  public void testGenerateEntitiesDAOsOverlapUrlQueryWithRealSession() {
+    // Test 1: includeSelf only
+    Query<ExternalLocationDAO> query1 =
+        PathBasedRpcUtils.generateEntitiesDAOsOverlapUrlQuery(
+            session,
+            "s3://bucket/path",
+            SecurableType.EXTERNAL_LOCATION,
+            10,
```
tdas: what is 10? use the /* name= */ here as well.

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
```diff
@@ -53,26 +76,18 @@ public void testExternalLocationCRUD() throws ApiException {
             .name(EXTERNAL_LOCATION_NAME)
             .comment(COMMENT)
             .url(URL)
-            .credentialName(CREDENTIAL_NAME);
+            .credentialName(CREDENTIAL_NAME + "_not_exist");
```
tdas: why do you need `CREDENTIAL_NAME + "_not_exist"` and not just `does_not_exist`?

tdas: actually.. why are you changing this? the exists test was correct.. `CREDENTIAL_NAME` does not exist until created.

yili-db: I have changed this because I changed the `CREDENTIAL_NAME` to be created in `setUp` and deleted in `tearDown`. Individual tests no longer need to create it.

server/src/test/java/io/unitycatalog/server/persist/utils/PathBasedRpcUtilsTest.java
yili-db: Added more tests for now.
I'll improve this in a follow up PR so that `getParentPathsList` will no longer need to worry about non-standardized URLs. Right now the actual caller of `getParentPathsList` standardizes the URL already but there's no enforcement. I'll introduce a new wrapper class `NormalizedURL` that are a string URL but guaranteed to be standardized.

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
```diff
@@ -108,4 +123,143 @@ public void testExternalLocationCRUD() throws ApiException {
...
+            .comment(COMMENT)
+            .url(URL)
+            .credentialName(CREDENTIAL_NAME);
+    externalLocationOperations.createExternalLocation(firstLocation);
+
+    // Test 1: Duplicate URL should be rejected
+    CreateExternalLocation duplicateLocation =
+        new CreateExternalLocation()
+            .name(EXTERNAL_LOCATION_NAME + "_duplicate")
+            .comment(COMMENT)
```
tdas: why do we need to pass comment for all these invalid test cases? seems redundant and code bloat?

tdas: in fact, for the valid cases as well.. if you are not validating whether comment has been set, there isnt much of a point to set a comment.

server/src/main/java/io/unitycatalog/server/persist/ExternalLocationRepository.java
```diff
@@ -197,4 +209,86 @@ public ExternalLocationDAO deleteExternalLocation(String name) {
...
+            otherExternalLocation ->
+                !locationWithNewUrl.getId().equals(otherExternalLocation.getId()))
+        .findAny()
+        .ifPresent(
+            existingExternalLocation -> {
+              throw new BaseException(
+                  ErrorCode.INVALID_ARGUMENT,
+                  "An existing external location with URL '"
+                      + existingExternalLocation.getUrl()
+                      + "' already exists. Cannot change the URL to duplicate or overlap with "
```
tdas: redundant to say "existing ... " and "already exists" in the same sentence.

tdas: also ... you can merge this into one sentence. `"Cannot accept an external location that duplicates or overlaps with existing external location " + existingExternalLocation.getUrl()`

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
yili-db: Sounds good. I'll rename it to `ExternalLocationUtils`.
As for `FileOperations`, I want to keep it as is but move the path normalization part into a separate class `NormalizedURL` in a follow up PR

server/src/main/java/io/unitycatalog/server/persist/ExternalLocationRepository.java
tdas: is it secure to respond with `existingExternalLocation` in the error message? what if the caller does not have the permissions to list external locations? then we would be giving out the existence of this location .. isnt it? 

I think its fair to log it .. but not return in the error message.

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
```diff
@@ -108,4 +123,143 @@ public void testExternalLocationCRUD() throws ApiException {
...
+    CreateExternalLocation parentLocation =
+        new CreateExternalLocation()
+            .name(EXTERNAL_LOCATION_NAME + "_parent")
+            .comment(COMMENT)
+            .url(URL)
+            .credentialName(CREDENTIAL_NAME);
+    assertApiException(
+        () -> externalLocationOperations.createExternalLocation(parentLocation),
+        ErrorCode.INVALID_ARGUMENT,
+        "An existing external location with URL");
```
tdas: make sure the error message says something about overlap.

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
```diff
@@ -108,4 +123,143 @@ public void testExternalLocationCRUD() throws ApiException {
...
+
+    // Clean up for next test
+    externalLocationOperations.deleteExternalLocation(EXTERNAL_LOCATION_NAME);
+
+    // Create an external location with subdir first
+    CreateExternalLocation childFirst =
+        new CreateExternalLocation()
+            .name(EXTERNAL_LOCATION_NAME)
+            .comment(COMMENT)
+            .url(URL + "/subpath/deeper")
```
tdas: nit: since you are making you are going to use this path multiple times.. make it a variable

tdas: nvm.. its only once.

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
```diff
@@ -108,4 +123,143 @@ public void testExternalLocationCRUD() throws ApiException {
...
+            .name(EXTERNAL_LOCATION_NAME + "_sibling2")
+            .comment(COMMENT)
+            .url(URL + "/data2")
+            .credentialName(CREDENTIAL_NAME);
+    ExternalLocationInfo sibling2Info = externalLocationOperations.createExternalLocation(sibling2);
+    assertThat(sibling2Info.getName()).isEqualTo(EXTERNAL_LOCATION_NAME + "_sibling2");
+
+    // Cleanup
+    externalLocationOperations.deleteExternalLocation(EXTERNAL_LOCATION_NAME + "_sibling1");
+    externalLocationOperations.deleteExternalLocation(EXTERNAL_LOCATION_NAME + "_sibling2");
```
tdas: shouldnt all the clean up be more rigorous. if any of the above assert fails, then you are leaving behind the location causing which may cause cascading failures?

might be clear to track all created locations in the test and in finally, clean them up.

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
```diff
@@ -108,4 +123,143 @@ public void testExternalLocationCRUD() throws ApiException {
...
+            .url("s3://bucket1/path")
+            .credentialName(CREDENTIAL_NAME);
+    externalLocationOperations.createExternalLocation(bucket1Location);
+
+    CreateExternalLocation bucket2Location =
+        new CreateExternalLocation()
+            .name(EXTERNAL_LOCATION_NAME + "_bucket2")
+            .comment(COMMENT)
+            .url("s3://bucket2/path")
+            .credentialName(CREDENTIAL_NAME);
```
tdas: out of these 5 lines.. only 2 needs to be updated in every unit test. in fact the only one that matters is url; as long as the location name is unique, it does not matter. cant you dedup all this code?

server/src/main/java/io/unitycatalog/server/persist/ExternalLocationRepository.java
```diff
@@ -197,4 +209,86 @@ public ExternalLocationDAO deleteExternalLocation(String name) {
...
+   *
+   * <p>This allows an external location to be updated to a URL that may technically "overlap" with
+   * its old URL (since they're the same entity), but prevents overlapping with any other external
+   * location.
+   *
+   * @param session The Hibernate session for database access
+   * @param locationWithNewUrl The external location DAO with the new URL to validate
+   * @throws BaseException if an overlapping external location exists (other than itself)
+   */
+  private void validateUrlNotUsedByAnyOtherExternalLocation(
```
tdas: its not clear from the name whats the difference between these 2 `validateUrlNotUsed...` functions. Any vs AnyOther? The first line in the doc does clarify the difference. 

my guess is that one is to be used for new locations, and another to be used for updating existin location? why cant they be merged together in a single function?

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
yili-db: Added a function do dedup all the lines. The name does matter though as some test case would try to update the external location.

server/src/main/java/io/unitycatalog/server/persist/ExternalLocationRepository.java
yili-db: Merged into a single function with a Optional<UUID> parameter.

tdas (approved):

PR #1285, opened 2026-01-01
Add force deletion support for external locations

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1285/files) to review incremental changes.
- [**stack/el_delete_force**](https://github.com/unitycatalog/unitycatalog/pull/1285) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1285/files)]
  - [stack/el_info_cred_name](https://github.com/unitycatalog/unitycatalog/pull/1287) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1287/files/cc0ac9c80f8fda49027d857ac71635e9063ffdfe..ab9e5513959aabc794d5587a36c8a0dd065a20c9)]
    - [stack/el_crud_perm](https://github.com/unitycatalog/unitycatalog/pull/1288) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1288/files/ab9e5513959aabc794d5587a36c8a0dd065a20c9..dfbeea242a21629630ad7584c6cdf14b0c9fd602)]
---------
**PR Checklist**

**Description of changes**
Add force deletion support for external locations

Adds an optional `force` parameter to the external location deletion API to allow deletion of external locations that are still in use by other entities.

Changes:
- When force=false (default), deletion fails if the external location is still referenced by any tables, volumes, or registered models
- When force=true, deletion proceeds regardless of dependent entities
- Add PathBasedRpcUtils.getAllEntitiesDAOsOverlapUrl() helper method to efficiently check multiple securable types for URL overlap

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/ExternalLocationRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/ExternalLocationRepository.java
@@ -269,7 +269,7 @@ public class ExternalLocationRepository {
    */
   private void validateUrlNotUsedByAnyExternalLocation(
       Session session, String url, Optional<UUID> currentExternalLocationId) {
-    ExternalLocationUtils.<ExternalLocationDAO>getEntitiesDAOsOverlapUrl(
+    ExternalLocationUtils.<ExternalLocationDAO>getEntityDAOsOverlapUrl(
             session,
             url,
             SecurableType.EXTERNAL_LOCATION,
--- a/server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
@@ -60,7 +60,11 @@ public class ExternalLocationUtils {
    *
    * <p>This is a convenience method that queries multiple securable types at once and returns
    * results as pairs of (SecurableType, IdentifiableDAO). Results are limited by 'limit' globally
-   * across all types, not per type.
+   * across all types, not per type. For each SecurableType, multiple IdentifiableDAO objects can be
+   * returned. For example, if it's called with securableTypes=[TABLE, VOLUME] and limit=3, and
+   * there are a total of 2 matching tables and 2 matching volumes in the database, it returns:
+   *
+   * <p>[(TABLE, table1), (TABLE, table2), (VOLUME, volume1)] // volume2 isn't returned
    *
    * @param session The Hibernate session for database access
    * @param url The URL to check for overlaps
@@ -117,7 +121,7 @@ public class ExternalLocationUtils {
    * @return List of matching entity DAOs, ordered by URL length descending
    * @throws IllegalArgumentException if the securableType is not supported for URL overlap checks
    */
-  public static <T extends IdentifiableDAO> List<T> getEntitiesDAOsOverlapUrl(
+  public static <T extends IdentifiableDAO> List<T> getEntityDAOsOverlapUrl(
       Session session,
       String url,
       SecurableType securableType,
--- a/server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
+++ b/server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
@@ -228,7 +228,9 @@ public abstract class BaseExternalLocationCRUDTest extends BaseCRUDTest {
   @Test
   public void testExternalLocationDeletion() throws ApiException, IOException {
     // Test 1: External location with volume - delete without force should fail
-    create(EXTERNAL_LOCATION_NAME + "_volume", URL + "/volumes");
+    String volumeExternalLocationName = EXTERNAL_LOCATION_NAME + "_volume";
+    String volumeUrlRoot = URL + "/volumes";
+    create(volumeExternalLocationName, volumeUrlRoot);
 
     CreateVolumeRequestContent createVolumeRequest =
         new CreateVolumeRequestContent()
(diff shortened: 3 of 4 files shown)


server/src/main/java/io/unitycatalog/server/persist/ExternalLocationRepository.java
```diff
@@ -195,14 +195,49 @@ public ExternalLocationInfo updateExternalLocation(
...
+                    /* includeSelf= */ true,
+                    /* includeSubdir= */ true)
+                .stream()
+                .findAny()
+                .ifPresent(
+                    pair -> {
+                      throw new BaseException(
+                          ErrorCode.INVALID_ARGUMENT,
+                          String.format(
+                              "External location still used by %s '%s'",
```
yili-db, line 237: In this case it should be fair to reply in error message, as the user is owner of external location.

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
```diff
@@ -180,4 +224,61 @@ public void testExternalLocationUrlOverlapPrevention() throws ApiException {
     create(EXTERNAL_LOCATION_NAME + "_sibling1", URL + "/data1");
     create(EXTERNAL_LOCATION_NAME + "_sibling2", URL + "/data2");
   }
+
+  @Test
+  public void testExternalLocationDeletion() throws ApiException, IOException {
+    // Test 1: External location with volume - delete without force should fail
+    create(EXTERNAL_LOCATION_NAME + "_volume", URL + "/volumes");
```
tdas: nit: please use the define a variable for this  `EXTERNAL_LOCATION_NAME + "_volume"` and reuse it

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
```diff
@@ -180,4 +224,61 @@ public void testExternalLocationUrlOverlapPrevention() throws ApiException {
...
+    assertApiException(
+        () -> delete(EXTERNAL_LOCATION_NAME + "_volume", Optional.empty()),
+        ErrorCode.INVALID_ARGUMENT,
+        "External location still used by");
+
+    // Delete with force should succeed
+    delete(EXTERNAL_LOCATION_NAME + "_volume", Optional.of(true));
+
+    // Test 2: External location with table - delete without force should fail
+    create(EXTERNAL_LOCATION_NAME + "_table", URL + "/tables");
```
tdas: same comment as above

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
```diff
@@ -180,4 +224,61 @@ public void testExternalLocationUrlOverlapPrevention() throws ApiException {
...
+                        .position(0)));
+    tableOperations.createTable(createTableRequest);
+
+    assertApiException(
+        () -> delete(EXTERNAL_LOCATION_NAME + "_table", Optional.of(false)),
+        ErrorCode.INVALID_ARGUMENT,
+        "External location still used by");
+
+    // Delete with force should succeed
+    delete(EXTERNAL_LOCATION_NAME + "_table", Optional.of(true));
```
tdas: shouldnt we test for models as well?

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
tdas: nit: shouldnt this be 

```suggestion
  public static <T extends IdentifiableDAO> List<T> getEntityDAOsOverlapUrl(
```

tdas: also .. isnt one SecurableType associated with only one IdentifiableDAO? thats why you are able to define Pair<SecurableType, IdentifiableDAO> in the new method.. isnt it? 

so why should this single securableType version return a list?

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
yili-db: Today models are created under `storage-root.models` specified in server.properties. We can consider the `storage-root.models` as one of the implicit  fallback External Location.
In this case it doesn't need to check for models as they will still be covered by that fallback External Location.

Later when the managed location for catalog&schema is implemented, it will be created under that instead. This code will be changed to check for catalog and schema locations instead (but still not models).

To summarize, this code only checks for entities that can be created using the actual External Locations directly. Today it's just **external** tables  and volumes.

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
yili-db: Added an example in the comment in code:
> For each SecurableType, multiple IdentifiableDAO objects can
> be returned. For example, if it's called with securableTypes=[TABLE, VOLUME] and limit=3,
> and there are a total of 2 matching tables and 2 matching volumes in the database, it returns:
> <p>[(TABLE, table1), (TABLE, table2), (VOLUME, volume1)]  // volume2 isn't returned
In this case, `getEntityDAOsOverlapUrl(TABLE)` returns `[table1, table2]`, and `getEntityDAOsOverlapUrl(VOLUME)` returns `[volume1, volume2]`.
`getAllEntityDAOsOverlapUrl` combines the results and returns `[(TABLE, table1), (TABLE, table2), (VOLUME, volume1)]`

server/src/test/java/io/unitycatalog/server/base/externallocation/BaseExternalLocationCRUDTest.java
tdas: Lets put it in the plan (after the core infra is done) to harmonize models as well. does not block this PR.

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -64,7 +121,7 @@ SecurableType.REGISTERED_MODEL, new DaoClassInfo(RegisteredModelInfoDAO.class, "
    * @return List of matching entity DAOs, ordered by URL length descending
    * @throws IllegalArgumentException if the securableType is not supported for URL overlap checks
    */
-  public static <T extends IdentifiableDAO> List<T> getEntitiesDAOsOverlapUrl(
+  public static <T extends IdentifiableDAO> List<T> getEntityDAOsOverlapUrl(
```
tdas, line 124: nit: can you rename both these functions to `get***DAOsWithURLOverlap`? the `with` is important to make it obvious what its returning. usually getX means it returns X and `getEntityDAOsOverlapUrl` sounds like its getting some kind of URL. 

this can be done in a later PR in the stack.

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
tdas: thanks for explaining.

Issue #1279, opened 2025-12-15
View Support for UC


TheThow:
**Is your feature request related to a problem? Please describe.**

At Celonis, we would like the Unity Catalog to support views, similar to the [Iceberg Spec](https://iceberg.apache.org/view-spec/). Especially via Iceberg to connect other query engines that don't support UC.

**Describe the solution you would like**

I'd like to create / load / list / delete views via the CLI / Spark Connect / Iceberg.
In the future, update would also be nice.

I've been [playing around locally](https://github.com/unitycatalog/unitycatalog/compare/main...TheThow:unitycatalog:view-support) adding another field to the TableDAO that contains a List of ViewRepresentations and piped that up to all 3 APIs as View. 
With Uniform tables and Spark 3.5, Iceberg it works out of the box. For Spark 4.0 Iceberg and the delta extension some extra work is needed.

For a first implementation I would not support versioning as even without the feature would be very useful. (And if we also handle update as a follow-up there would be no need).

I'd be happy to contribute this and discuss with someone how a suitable architecture would look like

**Additional context**

Iceberg View Architecture

<img width="1194" height="751" alt="Image" src="https://github.com/user-attachments/assets/fd879348-cb62-41bb-8cc6-2d0e47b32ae7" />

arturowczarek:
Great work, @TheThow. There is a couple of challenges to address before your views support could succeed.

1. Databricks Unity Catalog must adapt it. Now it doesn't expose endpoint `/views`. This is important for adoption, because for many users Databricks Unity Catalog remains the main source of their data (for example when you mix Databricks/Kubernetes Spark runtimes and using Unity Catalog OSS to just connect to Databricks Unity Catalog).

@openinx Maybe you have some connections to make it happen?

2. Spark doesn't have the implementation using the `ViewCatalog` yet. See: https://issues.apache.org/jira/browse/SPARK-31357. So your implementation in Spark catalog plugin will simply not work yet. I think, since the API is already established, Unity Catalog could already have it if had some decent unit tests. This way, when the time is right, someone could add integration tests using it in Spark.

[...]

PR #1286, opened 2026-01-01
Add support for loading views in Spark catalog

arturowczarek:
**PR Checklist**

- [X] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

The change introduces support for views represented as tables (tables of type `VIEW` with extra fields describing the view).

The Data Source V2 interface `org.apache.spark.sql.connector.catalog.ViewCatalog` doesn't have a proper implementation yet (see https://issues.apache.org/jira/browse/SPARK-31357). Before it is complete and changes from https://github.com/unitycatalog/unitycatalog/issues/1279 can be used, we can still use the Views as `org.apache.spark.sql.catalyst.catalog.CatalogTable`. The class `org.apache.spark.sql.catalyst.analysis.RelationResolution` from Catalyst responsible for resolving the view can properly use it.

The integration test proves that it will work. In addition, I've built the Spark plugin and tested it with Databricks Unity catalog. Everything works.

The integration test uses a mock HTTP server, because the Unity Catalog OSS can't store the view definitions yet. Once it does, the mock HTTP server can be replaced with the actual server.

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/ViewsTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/ViewsTest.java
@@ -1,0 +1,162 @@
+package io.unitycatalog.spark;
+
+import static io.unitycatalog.server.base.BaseServerTest.serverConfig;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.unitycatalog.server.model.ColumnInfo;
+import io.unitycatalog.server.model.ColumnTypeName;
+import io.unitycatalog.server.model.TableInfo;
+import io.unitycatalog.server.model.TableInfoViewDependencies;
+import io.unitycatalog.server.model.TableType;
+import io.unitycatalog.server.utils.JsonUtils;
+import io.unitycatalog.spark.utils.OptionsUtil;
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import lombok.AllArgsConstructor;
+import lombok.Data;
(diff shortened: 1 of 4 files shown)


arturowczarek:
@dennyglee @tdas Do you have time to have a look? It would be great to have support for loading views using `CatalogTable`

janresiak:
awesome, waiting for that!

janresiak (approved):

connectors/spark/src/test/java/io/unitycatalog/spark/ViewsTest.java
```diff
@@ -0,0 +1,162 @@
...
+                        .typeJson(
+                            "{\"name\":\"age\",\"type\":\"integer\",\"nullable\":false,\"metadata\":{}}")
+                        .nullable(false)))
+            .owner("tom@example.com")
+            .createdAt(1767261705238L)
+            .createdBy("tom@example.com")
+            .updatedAt(1767261705238L)
+            .updatedBy("tom@example.com")
+            .tableId("AAAA-BBBB-CCCC-DDDD-1234")
+            .viewDefinition(viewDefinition)
```
TheThow, line 106: Rel: https://github.com/unitycatalog/unitycatalog/issues/1279
If we are defining the API here can we make the definitions a list of objects with [dialect, sqlstring]?

There will be additional work needed to make that storable but it will be very helpful for multiple engines

yili-db:
Implemented by #1639

PR #1289, opened 2026-01-06
Add descriptions for Privilege in all.yaml

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1289/files) to review incremental changes.
- [**stack/privilege_enum**](https://github.com/unitycatalog/unitycatalog/pull/1289) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1289/files)]

---------
**PR Checklist**

**Description of changes**
Add descriptions for Privilege in all.yaml

### DIFF
--- a/api/all.yaml
+++ b/api/all.yaml
@@ -2673,6 +2673,7 @@ components:
         - Enables creating external tables through external locations.
         - Enables creating external volumes through external locations.
         - Allows designating managed storage locations at catalog or schema levels.
+        - Required privilege on the metastore to create storage credentials.
     UpdatePermissions:
       type: object
       properties:


yili-db:
@tdas By adding these in all.yaml, it would generate comments in java code:
```
/**
 * The privilege to grant.
 */
public enum Privilege {
  
  /**
   * Enables users to create new catalogs in a metastore.
   */
  CREATE_CATALOG("CREATE CATALOG"),
```

However as you can see it does not generate any doc changes. Because it doesn't generate anything in Privilege.md:
```
$ cat ./api/Models/Privilege.md
# Privilege
## Properties

[[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
```

This requires modifying build script to make it happen. It will be a next PR if it's desired.

tdas:
can you update the PR since the base PR was merged.

api/all.yaml
```diff
@@ -2653,6 +2653,25 @@ components:
         - CREATE EXTERNAL VOLUME
         - CREATE MANAGED STORAGE
       description: The privilege to grant.
+      x-enum-descriptions:
```
tdas, line 2657: are these basically aligned by order with the list of enums above?

yili-db: Yes. Unfortunately there's nothing to enforce the order and size of the list. This is the limitation of x-enum-descriptions.

tdas: actually.. what hapepns if someone adds another enum.. but not the corresponding description? does anything fail? if not, there is no forcing function to keep these two independent lists updated .. right?

yili-db: It doesn't fail. It would generate the new enum without comment assuming the new one is only added at the end. In this example I just added `SOMETHING ELSE` at the end without adding a description.
```
  /**
   * Required privilege on the metastore to create storage credentials.
   */
  CREATE_STORAGE_CREDENTIAL("CREATE STORAGE CREDENTIAL"),
  
  SOMETHING_ELSE("SOMETHING ELSE");
```
If a new enum is added in the **middle**, then the descriptions will be shifted and wrong.
A proper enforcement has to be done by build.sbt calling some custom scala script.

yili-db: Another PR https://github.com/unitycatalog/unitycatalog/pull/1298 to solve all these problems by adding a custom .md generator that **patches** the generated .md files.

tdas: but does the custom generator prevent (by failing) change that causes the descriptions<>enum value association to be wrong? how do we prevent that?

yili-db: Yes it does. Because they are just two lists, it does so by enforcing that every description must be like `ENUM VALUE: This is a description`.

tdas: okay.. merging this and then reviewing the next PR to discuss this further.

tdas (approved):

PR #1290, opened 2026-01-07
Introducing NormalizedURL to handle URL normalization.

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1290/files) to review incremental changes.
- [**stack/norm_url**](https://github.com/unitycatalog/unitycatalog/pull/1290) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1290/files)]

---------
**PR Checklist**

**Description of changes**
Introducing `NormalizedURL` to handle URL normalization.

This commit introduces a new `NormalizedURL` class that provides type-safe URL handling throughout the Unity Catalog codebase, replacing string-based URL manipulations.

Key changes:
- Add `NormalizedURL` class as an immutable wrapper that automatically normalizes URLs at construction time
- Move URL normalization logic from `FileOperations` to `NormalizedURL`
- `NormalizedURL.normalize` replaces all usages of `FileOperations.toStandardizedURIString()` if it has to provide a normalized URL String instead of a `NormalizedURL`.
- Update method signatures to accept `NormalizedURL` instead of String where appropriate for better type safety

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
@@ -581,7 +581,7 @@ public class ModelRepository {
           } catch (Exception deleteErr) {
             LOGGER.error(
                 "Unable to delete storage location {} during rollback: {}",
-                storageLocation == null ? "null" : storageLocation.toString(),
+                storageLocation,
                 deleteErr.getMessage());
           }
           tx.rollback();
--- a/server/src/main/java/io/unitycatalog/server/utils/NormalizedURL.java
+++ b/server/src/main/java/io/unitycatalog/server/utils/NormalizedURL.java
@@ -1,6 +1,5 @@
 package io.unitycatalog.server.utils;
 
-import com.google.common.annotations.VisibleForTesting;
 import io.unitycatalog.server.exception.BaseException;
 import io.unitycatalog.server.exception.ErrorCode;
 import java.net.URI;
@@ -188,8 +187,7 @@ public final class NormalizedURL {
    * @param path The path string to normalize, may be null
    * @return The normalized path, or null if input is null
    */
-  @VisibleForTesting
-  static String removeExtraSlashes(String path) {
+  private static String removeExtraSlashes(String path) {
     if (path == null || path.length() <= 1) {
       return path;
     }
--- a/server/src/test/java/io/unitycatalog/server/utils/NormalizedURLTest.java
+++ b/server/src/test/java/io/unitycatalog/server/utils/NormalizedURLTest.java
@@ -83,16 +83,4 @@ public class NormalizedURLTest {
     String uuid = UUID.randomUUID().toString();
     assertNormalizedURL("/tmp/tables/" + uuid, "file:///tmp/tables/" + uuid);
   }
-
-  @Test
-  public void testRemoveExtraSlashes() {
-    assertThat(NormalizedURL.removeExtraSlashes(null)).isEqualTo(null);
-    assertThat(NormalizedURL.removeExtraSlashes("")).isEqualTo("");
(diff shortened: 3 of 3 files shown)


server/src/test/java/io/unitycatalog/server/utils/NormalizedURLTest.java
yili-db, line 1: Renamed and modified from [FileOperationsTest.java‎](https://github.com/unitycatalog/unitycatalog/pull/1290/changes/7cd6eaea5fa62d2fc0c85e2810a71b46e4f644f0..59cd1c729810b1917944823ea6faab28715174a9#diff-d96d45ab8b0f367bdbe7d74312b6866a6832e080d9fb6ab94ee3c3d2481e14fa)

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -190,136 +188,10 @@ private URI modifyS3Directory(URI parsedUri, boolean createOrDelete) {
     }
   }
 
-  /**
```
yili-db, line 193: Moved to NormalizedURL.java

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -580,7 +581,7 @@ public ModelVersionInfo createModelVersion(CreateModelVersion createModelVersion
           } catch (Exception deleteErr) {
             LOGGER.error(
                 "Unable to delete storage location {} during rollback: {}",
-                storageLocation,
+                storageLocation == null ? "null" : storageLocation.toString(),
```
tdas: what is this change for? seems unrelated to normalizedurl

server/src/main/java/io/unitycatalog/server/utils/NormalizedURL.java
```diff
@@ -0,0 +1,218 @@
...
+   *   <li>"a/b////" -> "a/b"
+   *   <li>"" -> ""
+   *   <li>null -> null
+   * </ul>
+   *
+   * @param path The path string to normalize, may be null
+   * @return The normalized path, or null if input is null
+   */
+  @VisibleForTesting
+  static String removeExtraSlashes(String path) {
```
tdas: why does this need to explicitly tested?
removeExtraSlashes does not seem to be used by any production code directly. So then this can be private method in NormalizedURL, and just use NormalizedURL tests to test out all the cases of extra slashes.

isnt it?

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
yili-db: I see. This change isn't needed as the logger would print out "null" for null anyway.
I changed this line because earlier it was `String storageLocation = "";` so that when it fails it remains an empty string and is safe to print. But `NormalizedURL` can't begin as `""`. It has to be `null` if it fails.
I worried that it may cause NPE in the failure case. But actually it shouldn't.

server/src/main/java/io/unitycatalog/server/utils/NormalizedURL.java
yili-db: Removed.

tdas:
I see you have remove the removedExtraSlashes code.. but can you confirm that the existing tests for NormalizeUrl has sufficient code coverage.

yili-db:
> I see you have remove the removedExtraSlashes code.. but can you confirm that the existing tests for NormalizeUrl has sufficient code coverage.

@tdas Yes. In the `NormalizedURLTest` there are still tests against URLs with extra slashes in the front, end, and middle.

PR #1293, opened 2026-01-07
Add force deletion support for credential

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1293/files) to review incremental changes.
- [**stack/sc_delete_force**](https://github.com/unitycatalog/unitycatalog/pull/1293) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1293/files)]
  - [stack/path_cred_perm](https://github.com/unitycatalog/unitycatalog/pull/1299) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1299/files/59f667d79470d11bbfc251457b1425777940f5e8..aa13b53f63db60e31cb45aeb59028b14f51dd498)]
    - stack/create_ext_table
      - stack/sc_vend
        - stack/azure_sp
          - stack/cli

---------
**PR Checklist**

**Description of changes**
Add force deletion support for credential
Adds an optional `force` parameter to the credential deletion API to allow deletion of credential that are still in use by external locations.

### DIFF
--- a/examples/cli/src/test/java/io/unitycatalog/cli/credential/CliCredentialCRUDTest.java
+++ b/examples/cli/src/test/java/io/unitycatalog/cli/credential/CliCredentialCRUDTest.java
@@ -1,10 +1,12 @@
 package io.unitycatalog.cli.credential;
 
 import io.unitycatalog.cli.catalog.CliCatalogOperations;
+import io.unitycatalog.cli.externallocation.CliExternalLocationOperations;
 import io.unitycatalog.server.base.ServerConfig;
 import io.unitycatalog.server.base.catalog.CatalogOperations;
 import io.unitycatalog.server.base.credential.BaseCredentialCRUDTest;
 import io.unitycatalog.server.base.credential.CredentialOperations;
+import io.unitycatalog.server.base.externallocation.ExternalLocationOperations;
 
 public class CliCredentialCRUDTest extends BaseCredentialCRUDTest {
 
@@ -16,5 +18,10 @@ public class CliCredentialCRUDTest extends BaseCredentialCRUDTest {
   @Override
   protected CatalogOperations createCatalogOperations(ServerConfig config) {
     return new CliCatalogOperations(config);
+  }
+
+  @Override
+  protected ExternalLocationOperations createExternalLocationOperations(ServerConfig config) {
+    return new CliExternalLocationOperations(config);
   }
 }
--- a/examples/cli/src/test/java/io/unitycatalog/cli/credential/CliCredentialOperations.java
+++ b/examples/cli/src/test/java/io/unitycatalog/cli/credential/CliCredentialOperations.java
@@ -117,10 +117,14 @@ public class CliCredentialOperations implements CredentialOperations {
   }
 
   @Override
-  public void deleteCredential(String name) throws ApiException {
+  public void deleteCredential(String name, Optional<Boolean> force) throws ApiException {
     try {
-      String[] args =
-          addServerAndAuthParams(List.of("credential", "delete", "--name", name), config);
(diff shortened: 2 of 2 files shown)


server/src/main/java/io/unitycatalog/server/persist/dao/ExternalLocationDAO.java
```diff
@@ -19,6 +19,7 @@
     name = "uc_external_locations",
     indexes = {
       @Index(name = "idx_url", columnList = "url"),
+      @Index(name = "idx_credential_id", columnList = "credential_id"),
```
tdas, line 22: adding a index like .. can it break any existing backend database after a server restart with new version?

server/src/main/java/io/unitycatalog/server/service/CredentialService.java
```diff
@@ -92,8 +92,10 @@ public HttpResponse updateCredential(
       """)
   @AuthorizeKey(METASTORE)
   public HttpResponse deleteCredential(
-      @Param("name") @AuthorizeKey(CREDENTIAL) String name) {
-    CredentialInfo credentialInfo = credentialRepository.deleteCredential(name);
+      @Param("name") @AuthorizeKey(CREDENTIAL) String name,
+      @Param("force") Optional<Boolean> force) {
```
tdas, line 96: where is this ` @Param("force")` used? why add this?

server/src/main/java/io/unitycatalog/server/persist/dao/ExternalLocationDAO.java
yili-db: That is OK. It has happened many times in previous PRs, like the "idx_url" added in an earlier PR.
Quoting from Google:
> Adding a new @Index annotation to your Hibernate DAO (Entity) does not break or corrupt existing data in your database. 
>  1. Behavior Based on hbm2ddl.auto Settings 
> update: Hibernate will attempt to create the index automatically upon application startup. It generally adds the index without dropping existing tables or columns.

And we have: `etc/conf/hibernate.properties: hibernate.hbm2ddl.auto=update`

server/src/main/java/io/unitycatalog/server/service/CredentialService.java
yili-db: It's already part of `all.yaml` but was not implemented:
```
    delete:
      tags:
        - Credentials
      operationId: deleteCredential
...
      parameters:
        - name: force
          in: query
          description: Force deletion even if there are dependent external locations.
          required: false
          schema:
            type: boolean
```

tdas: then this is a very naive question. what uses this annotation ` @Param("x")`? clearly there is some sort of binding involved. whats component is using this annotation to bind and call this method with the right parameter values.

server/src/main/java/io/unitycatalog/server/persist/dao/ExternalLocationDAO.java
tdas: great. thanks for confirming.

server/src/main/java/io/unitycatalog/server/service/CredentialService.java
yili-db: @tdas This is handled by `UnityAccessDecorator.java`:
```
          // No key defined so implicitly referencing an (annotated) (query) parameter
          Param param = parameter.getAnnotation(Param.class);
```

yili-db: Ah actually I was wrong. This is only for authorization. This param binding must be part of armeria server code.

PR #1294, opened 2026-01-07
Handle credential generation failures gracefully in UCSingleCatalog

murali-db:
## Summary

This PR enables Unity Catalog Spark connector to handle tables where credentials cannot be generated gracefully. Currently, when credentials cannot be generated for such tables, the connector throws an exception. With this change, the connector returns the table metadata without credentials (when a catalog conf is set), allowing downstream code the opportunity to obtain credentials from elsewhere (e.g., a different endpoint).

## Changes

```mermaid
sequenceDiagram
    participant User
    participant UCSingleCatalog
    participant UCProxy
    participant SparkSession
    participant Delta

    User->>UCSingleCatalog: initialize(options with serverSidePlanning.enabled=true)
    UCSingleCatalog->>UCProxy: new UCProxy(sspEnabled=true)
    UCProxy->>UCProxy: Store sspEnabled
    
    User->>UCSingleCatalog: loadTable()
    UCSingleCatalog->>UCProxy: loadTable()
    UCProxy->>UCProxy: Try credentials (fail)
    UCProxy->>UCProxy: Check catalog sspEnabled
    alt SSP Enabled
        UCProxy->>SparkSession: "set(spark.databricks.delta.catalog.enableServerSidePlanning, true)"
        UCProxy->>Delta: Return table with empty credentials
        Delta->>UCProxy: Uses SSP for file access
    else SSP Disabled
        UCProxy-->>User: Throw ApiException
    end
```

- **Changed `temporaryCredentials`** from a required value to `Option[TemporaryCredentials]`
- **Wrapped credential generation** in nested try-catch blocks:
  - First attempts READ_WRITE credentials
  - Falls back to READ credentials if READ_WRITE fails
  - If SSP catalog conf is enabled -> Returns `None` if both fail (indicates user is not authorized for direct credentials) 
  - Else, same exception thrown as before
- **Updated `extraSerdeProps` handling** to return empty map when credentials are unavailable

### Before
- Tables where credentials cannot be generated would fail to load with an exception
- Query would fail completely

### After
- Tables where credentials cannot be generated load successfully without credentials
- Downstream code can handle these tables using alternative approaches
- Regular tables continue to receive credentials as before

## Testing

```mermaid
sequenceDiagram
    participant Test
    participant SparkSession
    participant UCSingleCatalog
    participant UCServer
    
    Note over Test,UCServer: Scenario 1: SSP Disabled
    Test->>SparkSession: create (SSP=false)
    Test->>SparkSession: CREATE EXTERNAL TABLE (unconfigured location)
    Test->>SparkSession: load table
    SparkSession->>UCSingleCatalog: loadTable()
    UCSingleCatalog->>UCServer: generateTemporaryTableCredentials()
    UCServer-->>UCSingleCatalog: 404 (no credentials configured)
    UCSingleCatalog->>UCSingleCatalog: Check serverSidePlanningEnabled=false
    UCSingleCatalog-->>Test: Throw ApiException
    
    Note over Test,UCServer: Scenario 2: SSP Enabled
    Test->>SparkSession: create (SSP=true)
    Test->>SparkSession: CREATE EXTERNAL TABLE (unconfigured location)
    Test->>SparkSession: load table
    SparkSession->>UCSingleCatalog: loadTable()
    UCSingleCatalog->>UCServer: generateTemporaryTableCredentials()
    UCServer-->>UCSingleCatalog: 404 (no credentials configured)
    UCSingleCatalog->>UCSingleCatalog: Check serverSidePlanningEnabled=true
    UCSingleCatalog->>SparkSession: "set(spark.databricks.delta.catalog.enableServerSidePlanning, true)"
    UCSingleCatalog-->>SparkSession: Return table with empty credentials
    SparkSession-->>Test: Table loaded successfully
```

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
@@ -74,6 +74,33 @@ public abstract class BaseSparkIntegrationTest extends BaseCRUDTest {
     return builder.getOrCreate();
   }
 
+  protected SparkSession createSparkSessionWithSSP(String... catalogs) {
+    SparkSession.Builder builder =
+        SparkSession.builder()
+            .appName("test-ssp")
+            .master("local[*]")
+            .config("spark.sql.shuffle.partitions", "4")
+            .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension");
+    for (String catalog : catalogs) {
+      String catalogConf = "spark.sql.catalog." + catalog;
+      builder =
+          builder
+              .config(catalogConf, UCSingleCatalog.class.getName())
+              .config(catalogConf + "." + OptionsUtil.URI, serverConfig.getServerUrl())
+              .config(catalogConf + "." + OptionsUtil.TOKEN, serverConfig.getAuthToken())
+              .config(catalogConf + "." + OptionsUtil.WAREHOUSE, catalog)
+              .config(catalogConf + "." + OptionsUtil.SERVER_SIDE_PLANNING_ENABLED, "true");
+      if (!List.of(SPARK_CATALOG, CATALOG_NAME).contains(catalog)) {
+        createTestCatalog(catalog);
+      }
+    }
+    // Use fake file system for cloud storage so that we can test credentials.
+    builder.config("fs.s3.impl", S3CredentialTestFileSystem.class.getName());
+    builder.config("fs.gs.impl", GCSCredentialTestFileSystem.class.getName());
+    builder.config("fs.abfs.impl", AzureCredentialTestFileSystem.class.getName());
+    return builder.getOrCreate();
+  }
+
   protected List<Row> sql(String statement, Object... args) {
     return session.sql(String.format(statement, args)).collectAsList();
   }
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
@@ -25,7 +25,9 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledIf;
 import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.ValueSource;
 
 public abstract class ExternalTableReadWriteTest extends BaseTableReadWriteTest {
   @TempDir protected File dataDir;
@@ -172,6 +174,44 @@ public abstract class ExternalTableReadWriteTest extends BaseTableReadWriteTest 
     TableSetupOptions optionsWithoutAsSelect = options.withAsSelect(Optional.empty());
     sql(optionsWithoutAsSelect.createExternalTableSql(location));
     return options.fullTableName();
+  }
+
+  @ParameterizedTest
+  @ValueSource(booleans = {false, true})
+  public void testServerSidePlanningCredentialFallback(boolean sspEnabled) {
+    // Recreate session with appropriate SSP configuration
+    if (session != null) {
+      session.stop();
+    }
+
+    session =
+        sspEnabled
+            ? createSparkSessionWithSSP(CATALOG_NAME)
+            : createSparkSessionWithCatalogs(CATALOG_NAME);
+
+    // Use unconfigured storage location to trigger credential failure
+    String tableName = "test_ssp_fallback";
+    String unconfiguredLocation = "s3://unconfigured-bucket-no-creds/data";
+    String fullTableName = String.format("%s.%s.%s", CATALOG_NAME, SCHEMA_NAME, tableName);
+
+    // Create external table pointing to unconfigured location
+    sql(
+        "CREATE EXTERNAL TABLE %s (id INT, name STRING) LOCATION '%s'",
+        fullTableName, unconfiguredLocation);
+
+    if (sspEnabled) {
+      // SSP enabled: should succeed with empty credentials
+      assertThat(session.table(fullTableName)).isNotNull();
+
+      // Verify Spark config was set by UC catalog
+      assertThat(session.conf().get("spark.databricks.delta.catalog.enableServerSidePlanning"))
+          .isEqualTo("true");
+    } else {
+      // SSP disabled (default): should throw exception
+      assertThatThrownBy(() -> session.table(fullTableName))
+          .hasCauseInstanceOf(ApiException.class)
+          .hasMessageContaining("generateTemporaryTableCredentials failed");
+    }
   }
 
   @BeforeEach


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -328,36 +328,53 @@ private class UCProxy(
...
+          tableId,
+          TableOperation.READ_WRITE,
+          creds,
+        )
+      case None =>
+        java.util.Collections.emptyMap[String, String]()
+    }
 
     val sparkTable = CatalogTable(
       identifier,
```
tdas, line 408: no tests for this??

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -328,36 +328,53 @@ private class UCProxy(
...
+              .generateTemporaryTableCredentials(
+                new GenerateTemporaryTableCredential().tableId(tableId).operation(TableOperation.READ)
+              )
+            Some(creds)
+          } catch {
+            case _: ApiException =>
+              // Credentials not available (e.g., user not authorized due to FGAC row filter/column mask)
+              // Return None to allow downstream code the opportunity to obtain credentials from elsewhere
+              // (e.g., a different endpoint) rather than throwing an exception
+              None
```
tdas: do know what will happen if you use UC 0.4 (with this change that gives the control back to delta-spark) and Delta-Spark 4.0 (which obviously does not have any code to throw error on missing credential)?

docs/integrations/unity-catalog-spark.md
```diff
@@ -472,6 +472,139 @@ Drop Table
 !!! warning
     Note, this action will only drop the table from UC, it will not remove the data from the file system
 
+## Credential Handling and Access Control
```
tdas: is this an intentional change in the docs?

connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogCredentialFallbackTest.java
```diff
@@ -0,0 +1,423 @@
+package io.unitycatalog.spark;
```
murali-db: this test is still todo for me. will take a pass to improve this. but pls take a look at the prod code approach. @tdas

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -328,36 +329,82 @@ private class UCProxy(
...
+          TableOperation.READ_WRITE,
+          creds,
+        )
+      case None =>
+        // Determine if server-side planning is enabled by checking SparkSession config
+        val sspEnabled = try {
+          SparkSession.getActiveSession match {
+            case Some(spark) =>
+              val configValue = spark.conf.get(
+                "spark.databricks.delta.catalog.enableServerSidePlanning",
```
tdas: this is okay for this PR, but can we make this a catalog-specific configuration .. rather than a delta configuration? in a follow up PR. 

basically UCSingleCatalog... based on its own configuration (new option like `serverSidePlanning` along with the `url`, `token`, etc.) ... will set tell DeltaCatalog the details of whether to enable SSP and other configs like URL.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -328,36 +329,82 @@ private class UCProxy(
...
+          locationUri.getScheme,
+          uri.toString,
+          tokenProvider,
+          tableId,
+          TableOperation.READ_WRITE,
+          creds,
+        )
+      case None =>
+        // Determine if server-side planning is enabled by checking SparkSession config
+        val sspEnabled = try {
```
tdas: can we move this case into a separate method. its very nested and hard to read.

connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogCredentialFallbackTest.java
tdas: is it still a todo?

connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogCredentialFallbackTest.java
```diff
@@ -0,0 +1,535 @@
...
+ *   <li>Parameterized tests to ensure complete matrix coverage
+ *   <li>Reflection-based mock injection for internal API testing
+ * </ul>
+ *
+ * @see UCSingleCatalog
+ * @see io.unitycatalog.spark.auth.CredPropsUtil
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@Execution(ExecutionMode.SAME_THREAD)
+public class UCSingleCatalogCredentialFallbackTest {
```
tdas: are you sure this is the right name? see the naming pattern in other classes in the directory?

connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogCredentialFallbackTest.java
```diff
@@ -0,0 +1,535 @@
...
+      // Verify expected number of API calls
+      verify(fixture.mockCredentialsApi, times(scenario.expectedApiCalls))
+          .generateTemporaryTableCredentials(any());
+    } finally {
+      session.stop();
+    }
+  }
+
+  static Stream<Arguments> tableTypeScenarios() {
+    return Stream.of(TableTypeScenario.values())
```
tdas: why so nested.

connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogCredentialFallbackTest.java
```diff
@@ -0,0 +1,535 @@
...
+
+  /**
+   * Parameterized test covering credential fallback for all table types: - 2 table types (MANAGED,
+   * EXTERNAL) - 3 credential scenarios (RW succeeds, RW fails/R succeeds, both fail) - 2 SSP modes
+   * (enabled, disabled) = 12 test combinations
+   */
+  @ParameterizedTest(name = "{0} table, {1}, {2}")
+  @MethodSource("tableTypeScenarios")
+  @DisplayName("Credential fallback works for all table types")
+  void testCredentialFallbackByTableType(
```
tdas: why do we need to test for different table types?

connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogCredentialFallbackTest.java
```diff
@@ -0,0 +1,535 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.MethodSource;
+
+/**
```
tdas: this really feels like massive amount of llm-generated code to test something very small thats not worth this much code. and these tests are heavy. 

are you sure this is the best way to test the new logic you have added?

connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogCredentialFallbackTest.java
murali-db: yep, don't bother looking at the tests yet

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -56,8 +57,8 @@ class UCSingleCatalog
       JitterDelayRetryPolicy.builder().build(),uri, tokenProvider)
     temporaryCredentialsApi = new TemporaryCredentialsApi(apiClient)
     tablesApi = new TablesApi(apiClient)
-    val proxy = new UCProxy(uri, tokenProvider, renewCredEnabled, apiClient, tablesApi,
```
yili-db, line 58: This line seems unchanged?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
murali-db: 1. client should read the url from a catalog level conf.
2. enable ssp should be a catalog level conf. 

for now, the client just reads the url conf and continues. the url conf has a default hardcoded value (instead of hardcoded url in code). in the future, when server sets this, client automatically gets the signal from uc. 
todo:
url change can be in a different pr. 
this pr just move the conf to be catalog level. 

for now, the uc can set the session option if the catalog option is set. 
and the catalog option is the only user visible conf. 


@murali-db

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
murali-db: the query will fail with a credential access error at a later stage. but in practice, the intended behavior is that people on older clients will not set the catalog conf to true. so the 2 tests we have added should have coverage: - one when the enableSSP catalog conf is set to true, and another when it is unset (default to false).

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
murali-db: it it for for the scala style line limit after we've added a new param

murali-db:
moved to https://github.com/unitycatalog/unitycatalog/pull/1334

PR #1298, opened 2026-01-11
Tune enum descriptions add a comment

yili-db:
**PR Checklist**

**Description of changes**
Tune enum descriptions add a comment

This commit turns the description of enum values in all.yaml file to have a prefix of enum value. This is to help developers to identify mismatched descriptions in enum definition.
It also adds a comment to reminding adding a description in the same order whenever adding a new value.

### DIFF
--- a/api/all.yaml
+++ b/api/all.yaml
@@ -2476,7 +2476,9 @@ components:
         - PENDING_REGISTRATION
         - FAILED_REGISTRATION
         - READY
-      # NOTE: When adding a new enum, make sure add a new description in the same order.
+      # NOTE: When adding a new enum, always add a new enum to the end of enum list, and
+      #  corresponding description at the end of the description list. Both list must have the same
+      #  order.
       x-enum-descriptions:
         - 'MODEL_VERSION_STATUS_UNKNOWN: Unknown'
         - 'PENDING_REGISTRATION: Indicates that the client has not completely written all model artifacts to external storage.'
@@ -2656,7 +2658,9 @@ components:
         - CREATE EXTERNAL VOLUME
         - CREATE MANAGED STORAGE
         - CREATE STORAGE CREDENTIAL
-      # NOTE: When adding a new enum, make sure add a new description in the same order.
+      # NOTE: When adding a new enum, always add a new enum to the end of enum list, and
+      #  corresponding description at the end of the description list. Both list must have the same
+      #  order.
       x-enum-descriptions:
         - 'CREATE CATALOG: Enables users to create new catalogs in a metastore.'
         - 'USE CATALOG: Required to interact with any object within a catalog.'


api/all.yaml
```diff
@@ -2654,6 +2655,26 @@ components:
         - CREATE MANAGED STORAGE
         - CREATE STORAGE CREDENTIAL
       description: The privilege to grant.
+      x-enum-descriptions:
+        - 'CREATE CATALOG: Enables users to create new catalogs in a metastore.'
```
yili-db, line 2665: Adding a prefix `CREATE CATALOG:` to help the custom `EnumDocProcessor` to enforce the ordering of comments while still being compatible with `x-enum-descriptions`.

project/EnumDocProcessor.java
```diff
@@ -0,0 +1,233 @@
+import sbt.util.Logger;
+import org.yaml.snakeyaml.Yaml;
+import java.io.*;
+import java.nio.file.*;
+import java.util.*;
+import java.util.function.Supplier;
+
+/**
```
tdas: this is a lot of code for adding docs!
in order to minimize all this custom code, can you show the code for verifying # enums = # descriptions?

tdas: basically i really want to minimize any logic taht require reading generated files and parsing its content. is there a way around it?

yili-db:
```
    if (descriptionList.size() != enumValues.size()) {
      throw new RuntimeException(
          String.format(
              "%s has %d enum values but %d descriptions.",
              schemaName, enumValues.size(), descriptionList.size()));
    }
```
And
```
      // Validate that description starts with enum value. This is to make sure that the
      // descriptions are in proper order.
      if (!desc.startsWith(expectedPrefix)) {
        throw new RuntimeException(
            String.format(
                "%s.%s: Description must start with '%s' but got: '%s'",
                schemaName, enumValue, expectedPrefix, desc));
      }
```

yili-db: This script needs to parse the `all.yaml` file in order to know the enums and descriptions.
It can be made slightly simpler and smaller if we don't need it to add the enum values and descriptions in the .md files (currently they are with empty tables). But it still needs to parse `all.yaml` file anyway.

tdas: thats the part i dont like. this post-processing parsing stuff is a ugly hack.
I see there are 2 approaches possible
1. Update the openAPI mustache files (templates for code/file generation) to put the doc. That's the principled way. See if you can make it work. See my past PR  - https://github.com/unitycatalog/unitycatalog/pull/1179/changes#diff-99f074fad01fd6dd5533925fb3a3245b23ec61712a0a4181e7098baa909bbc57
2. if the above gets too complicated and hard to maintain, then lets just drop this, and leave comments in the yaml saying to always add enum to the end  (maybe the comment is at the end of the list) and to update the descriptions below. no enforcement, but whoever reviews has to verify that both are updated consistently.

yili-db: I tried to use a OpenAPI mustache file to generate the right `.md` file. But it turns out there's a bug in the generator we use so that the enum values and descriptions is never passed in to mustache files.
This is a bug in OpenAPI generator that we can't fix here without a custom generator.

I have removed the generator and the .md file content it generates. Instead I have just added NOTE in the `all.yaml` file to remind adding descriptions.

api/all.yaml
```diff
@@ -2469,17 +2469,19 @@ components:
...
-        FAILED_REGISTRATION status indicates that the client has failed to write all model artifacts to external storage.
-        READY status indicates that the client has successfully written all model artifacts to external storage.
+      description: All possible model version statuses.
       type: string
       enum:
         - MODEL_VERSION_STATUS_UNKNOWN
         - PENDING_REGISTRATION
         - FAILED_REGISTRATION
         - READY
+      # NOTE: When adding a new enum, make sure add a new description in the same order.
```
```suggestion
      # NOTE: When adding a new enum, always add a new enum to the end of enum list, and corresponding description at the end of the description list. Both list must have the same order. 
      Also keep this note at the end of the list so that reviewers see this when reviewing.
      
```

api/all.yaml
tdas: same comment as above.

tdas (approved):

PR #1302, opened 2026-01-13
Fix the static members in BaseServerTest

yili-db:
**PR Checklist**

**Description of changes**
Fix the static members in BaseServerTest

The static member objects, especially cloudCredentialVendor, would cause a problem when multiple tests accesses it. They should not be static. This PR fixes them.

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/base/BaseServerTest.java
+++ b/server/src/test/java/io/unitycatalog/server/base/BaseServerTest.java
@@ -20,11 +20,11 @@ import org.junit.jupiter.api.io.TempDir;
 
 public abstract class BaseServerTest {
 
-  public static ServerConfig serverConfig = new ServerConfig("http://localhost", "");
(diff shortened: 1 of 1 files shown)


server/src/test/java/io/unitycatalog/server/base/BaseServerTest.java
```diff
@@ -20,11 +20,11 @@
 
 public abstract class BaseServerTest {
 
-  public static ServerConfig serverConfig = new ServerConfig("http://localhost", "");
-  protected static UnityCatalogServer unityCatalogServer;
-  protected static Properties serverProperties;
-  protected static HibernateConfigurator hibernateConfigurator;
-  protected static CloudCredentialVendor cloudCredentialVendor;
+  public static final ServerConfig serverConfig = new ServerConfig("http://localhost", "");
```
yili-db, line 23: having `static final` is OK. Otherwise non-final static fields can cause trouble because they reused across different tests.

PR #1304, opened 2026-01-14
Introduce a new non-resource @AuthorizeKey annotation for conditional auth expressions

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1304/files) to review incremental changes.
- [**stack/auth_param**](https://github.com/unitycatalog/unitycatalog/pull/1304) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1304/files)]
  - [stack/create_ext_table](https://github.com/unitycatalog/unitycatalog/pull/1306) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1306/files/dcdb42f1f57a923ad79a7c145f6b86095f9edc49..f3b8a884ac04662078cae3ba085656a4b10377e7)]
---------
**PR Checklist**

**Description of changes**
Introduce a new non-resource `@AuthorizeKey` annotation for conditional auth expressions

This change involves the following classes:
- `@AuthorizeResourceKey`: Was renamed from existing `@AuthorizeKey` in earlier PR. Annotation for resource keys.
- `@AuthorizeKey`: A new annotation for any keys from request. This annotation does not invovle resource ID mapping regardless if the key is about a resource or not.
- `KeyLocator`: This class is extended to cover BOTH annotations.

With this new annotation, service methods can introduce request parameters that are not a resource into the auth expression as well.
For example:
```java
  @AuthorizeExpression("""
      #authorizeAny(#principal, #schema, OWNER, USE_SCHEMA) &&
      #authorizeAny(#principal, #catalog, OWNER, USE_CATALOG) &&
      (#operation == 'READ_VOLUME'
        ? #authorizeAny(#principal, #volume, OWNER, READ_VOLUME)
        : #authorize(#principal, #volume, OWNER))
      """)
  public HttpResponse generateTemporaryTableCredential(
      @AuthorizeResourceKey(value = VOLUME, key = "volume_id")
      @AuthorizeKey(key = "operation")
      GenerateTemporaryVolumeCredential generateTemporaryVolumeCredential) {
```
This expression requires OWNER or READ_VOLUME for read operations, but OWNER for write operations. The ``#operation` variable in the auth expression is picked up from the request payload by the authorization framework.

Credential services that previously have to do a manual authorization are migrated to use this new annotation and conditional auth expression:
   - TemporaryTableCredentialsService
   - TemporaryVolumeCredentialsService
   - TemporaryModelVersionCredentialsService
   - TemporaryPathCredentialsService

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/auth/annotation/AuthorizeKey.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/annotation/AuthorizeKey.java
@@ -8,7 +8,7 @@ import java.lang.annotation.Target;
 /**
  * Expose a request payload field as a variable in authorization expressions.
  *
- * <p>Unlike {@code @AuthorizeResourceKey}, which annotates payload fields that reference to
+ * <p>Unlike {@link AuthorizeResourceKey}, which annotates payload fields that reference to
  * resources and maps them to resource identifiers (UUIDs), this class can annotate ANY payload
  * fields whether they are resource or not. It does NOT do the resource ID mapping but exposes the
  * raw payload value directly to the SpEL expression context, even if the payload field references a
--- a/server/src/main/java/io/unitycatalog/server/auth/annotation/AuthorizeResourceKey.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/annotation/AuthorizeResourceKey.java
@@ -10,12 +10,13 @@ import java.lang.annotation.Target;
 /**
  * Map a request parameter to a unity catalog resource key.
  *
- * <p>This annotation is used to map a request parameter to a unity catalog resource key. The
- * resource key is used to retrieve the resource identifier, which is then used to authorize the
- * request. As an example, suppose you are making a request the retrieve a schema, the parameter
- * that contains the schema name might be defined in the request as
+ * <p>Unlike {@link AuthorizeKey} which only exposes the raw value of ANY request field, this class
+ * only annotates request fields that reference to resources and maps them to resource identifiers
+ * (UUIDs). The resource key is used to retrieve the resource identifier, which is then used to
+ * authorize the request. As an example, suppose you are making a request the retrieve a schema,
+ * the parameter that contains the schema name might be defined in the request as:
  *
- * <p>@Param("full_Name") String fullName
+ * <p>@AuthorizeResourceKey(SCHEMA) @Param("full_Name") String fullName
  *
  * <p>This annotation would take the value of the fullName parameter and use it to retrieve the
  * schema resource identifier looking up the identifier from the persistence layer (database).
--- a/server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
@@ -42,16 +42,20 @@ import static io.unitycatalog.server.auth.decorator.KeyLocator.Source.SYSTEM;
  * Armeria access control Decorator.
  *
  * <p>This decorator provides the ability to protect Armeria service methods with per method access
- * control rules. This decorator is used in conjunction with two annotations, @AuthorizeExpression
- * and @AuthorizeResourceKey to define authorization rules and identify requests parameters for
- * objects to authorize with.
+ * control rules. This decorator is used in conjunction with following 3 annotations to define
+ * authorization rules and identify requests parameters for objects to authorize with:
  *
- * <p>{@code @AuthorizeExpression} - This defines a Spring Expression Language expression to
- * evaluate to make an authorization decision. {@code @AuthorizeResourceKey} - This annotation is
- * used to define request and payload parameters for the authorization context. These are typically
- * things like catalog, schema and table names. This annotation may be used at both the method and
- * method parameter context. It may be specified more than once per method to map parameters to
- * object keys.
+ * <p>1. {@code @AuthorizeExpression} - This defines a Spring Expression Language expression to
+ * evaluate to make an authorization decision.
+ *
+ * <p>2. {@code @AuthorizeResourceKey} - This annotation is used to define request and payload
+ * parameters for the authorization context. These are typically things like catalog, schema and
+ * table names. This annotation may be used at both the method and method parameter context. It may
+ * be specified more than once per method to map parameters to object keys.
+ *
+ * <p>3. {@code @AuthorizeKey} - This annotation is used to define request and payload parameters
+ * for that are usually not resources but parameters of the operation. For example, table_type,
+ * operation, etc.
  */
 public class UnityAccessDecorator implements DecoratingHttpServiceFunction {
 
@@ -117,7 +121,7 @@ public class UnityAccessDecorator implements DecoratingHttpServiceFunction {
     // we want to authorize against.
 
     Map<SecurableType, Object> resourceKeys = new HashMap<>();
-    Map<String, Object> otherKeys = new HashMap<>();
+    Map<String, Object> nonResourceValues = new HashMap<>();
 
     // Split up the locators by type, because we have to extract the value from the request
     // different ways for different types
@@ -143,7 +147,7 @@ public class UnityAccessDecorator implements DecoratingHttpServiceFunction {
       if (l.getType().isPresent()) {
         resourceKeys.put(l.getType().get(), value);
       } else {
-        otherKeys.put(l.getVariableName(), value);
+        nonResourceValues.put(l.getVariableName(), value);
       }
     });
 
@@ -151,7 +155,7 @@ public class UnityAccessDecorator implements DecoratingHttpServiceFunction {
       // If we don't have any PAYLOAD locators, we're ready to evaluate the authorization and allow
       // or deny the request.
       LOGGER.debug("Checking authorization before method.");
-      checkAuthorization(principal, expression, resourceKeys, otherKeys);
+      checkAuthorization(principal, expression, resourceKeys, nonResourceValues);
 
       return delegate.serve(ctx, req);
     } else {
@@ -163,7 +167,7 @@ public class UnityAccessDecorator implements DecoratingHttpServiceFunction {
           req.contentType(),
           payloadLocators,
           resourceKeys,
-          otherKeys);
+          nonResourceValues);
 
       // Note that peekData only gets called for requests that actually have data (like PUT&POST)
 
@@ -171,7 +175,7 @@ public class UnityAccessDecorator implements DecoratingHttpServiceFunction {
         LOGGER.debug("Authorization peekData invoked.");
 
         if (peekDataHandler.processPeekData(data)) {
-          checkAuthorization(principal, expression, resourceKeys, otherKeys);
+          checkAuthorization(principal, expression, resourceKeys, nonResourceValues);
         }
       });
 
@@ -199,9 +203,9 @@ public class UnityAccessDecorator implements DecoratingHttpServiceFunction {
       UUID principal,
       String expression,
       Map<SecurableType, Object> resourceKeys,
-      Map<String, Object> otherKeys) {
+      Map<String, Object> nonResourceValues) {
     LOGGER.debug("resourceKeys = {}", resourceKeys);
-    LOGGER.debug("otherKeys = {}", otherKeys);
+    LOGGER.debug("nonResourceValues = {}", nonResourceValues);
 
     Map<SecurableType, Object> resourceIds = keyMapper.mapResourceKeys(resourceKeys);
 
@@ -211,7 +215,7 @@ public class UnityAccessDecorator implements DecoratingHttpServiceFunction {
 
     LOGGER.debug("resourceIds = {}", resourceIds);
 
-    if (!evaluator.evaluate(principal, expression, resourceIds, otherKeys)) {
+    if (!evaluator.evaluate(principal, expression, resourceIds, nonResourceValues)) {
       throw new BaseException(ErrorCode.PERMISSION_DENIED, "Access denied.");
     }
   }
@@ -250,8 +254,8 @@ public class UnityAccessDecorator implements DecoratingHttpServiceFunction {
         locators.add(KeyLocator.from(key, parameter));
       }
 
-      AuthorizeKey[] allOtherKeys = parameter.getAnnotationsByType(AuthorizeKey.class);
-      for (AuthorizeKey key : allOtherKeys) {
+      AuthorizeKey[] allNonResourceKeys = parameter.getAnnotationsByType(AuthorizeKey.class);
+      for (AuthorizeKey key : allNonResourceKeys) {
         locators.add(KeyLocator.from(key, parameter));
       }
     }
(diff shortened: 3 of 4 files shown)


server/src/main/java/io/unitycatalog/server/auth/annotation/AuthorizeParam.java
```diff
@@ -0,0 +1,66 @@
...
+ * <p><b>Null Handling:</b> If the specified field doesn't exist or is null, the variable will be
+ * set to null in the SpEL context. Expressions should handle this using null checks: {@code
+ * #operation == null || #operation == 'READ'}
+ *
+ * @see AuthorizeKey for mapping payload fields to resource identifiers
+ * @see AuthorizeExpression for defining authorization expressions
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.PARAMETER)
+public @interface AuthorizeParam {
```
yili-db: This is the annotation to use on service methods.

server/src/main/java/io/unitycatalog/server/auth/decorator/ParamLocator.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * <p>Example: For {@code @AuthorizeParam(key = "operation", name = "op")}, this locator would
+ * extract the "operation" field value from the request payload and make it available as "#op" in
+ * the SpEL authorization expression.
+ *
+ * @see io.unitycatalog.server.auth.annotation.AuthorizeParam
+ */
+@Builder
+@Getter
+public class ParamLocator {
```
yili-db: Similar to `KeyLocator`, this is the helper class used for finding out the actual values.
Using both the annotator class and a locator class is an established pattern for the resource annotation already. So here I am just following that pattern.

server/src/main/java/io/unitycatalog/server/auth/decorator/ParamLocator.java
```diff
@@ -0,0 +1,27 @@
...
+  /** The key path to extract from payload (e.g., "operation" or "config.mode") */
+  private String key;
```
yili-db: Corresponds to `AuthorizeParam.key`

server/src/main/java/io/unitycatalog/server/auth/decorator/ParamLocator.java
```diff
@@ -0,0 +1,27 @@
...
+
+  /** The SpEL variable name to expose this parameter as (e.g., "operation" or "op") */
```
yili-db: The **materialized** `AuthorizeParam.name`.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -204,7 +212,7 @@ private void checkAuthorization(
 
     LOGGER.debug("resourceIds = {}", resourceIds);
 
-    if (!evaluator.evaluate(principal, expression, resourceIds)) {
+    if (!evaluator.evaluate(principal, expression, resourceIds, parameters)) {
```
yili-db: This is where the evaluation is done.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -348,9 +395,21 @@ private boolean processPeekData(HttpData data) {
...
                 });
 
-            payloadLocators.forEach(l -> resourceKeys.put(
+            // Extract resource keys from KeyLocators
+            payloadKeyLocators.forEach(l -> resourceKeys.put(
                 l.getType(),
                 findPayloadValue(l.getKey(), payload)));
+
+            // Extract parameter values from ParamLocators
+            payloadParamLocators.forEach(l -> {
```
yili-db: This is where the values are pulled from the payload, using the info of the locators.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessEvaluator.java
```diff
@@ -91,6 +108,8 @@ public boolean evaluate(
 
     resourceIds.forEach((k, v) -> context.setVariable(k.name().toLowerCase(), v));
 
+    parameters.forEach(context::setVariable);
```
yili-db: This is how the extra parameters are supplied to the expression evaluation

server/src/main/java/io/unitycatalog/server/auth/annotation/AuthorizeParam.java
```diff
@@ -0,0 +1,66 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Expose a request payload field as a variable in authorization expressions.
+ *
+ * <p>Unlike {@code @AuthorizeKey}, which maps payload fields to resource identifiers (UUIDs), this
+ * annotation exposes the raw payload value directly to the SpEL expression context. This is useful
```
tdas: what is SpEL

yili-db: Spring Expression Language

tdas: > raw payload value


Both AuthorizeKey and this AuthorizeParam exposes the value from the request fields isnt it? the major difference between the type of the value - AuthorizeKey exposes resource id related fields, AuthorizeParam can expose ANY field whether or not related to the resource. 

is this it? 

if so, then please clarify this key distinction in the docs.

tdas: and make the docs consistent across the classes in this PR when you add this detail to this class.

server/src/main/java/io/unitycatalog/server/auth/decorator/ParamLocator.java
tdas: Lets make this AuthorizeParamLocator (to match with AuthorizeParam) 

and in a seprate refactor PR .. rename KeyLocator to be consistent.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -119,35 +123,36 @@ private HttpResponse authorizeByRequest(
...
-    // Split up the locators by type, because we have to extract the value from the request
+    // Split up the key locators by type, because we have to extract the value from the request
     // different ways for different types
-
-    List<KeyLocator> systemLocators = locators.stream()
+    List<KeyLocator> systemKeyLocators = keyLocators.stream()
         .filter(l -> l.getSource().equals(SYSTEM))
         .toList();
-    List<KeyLocator> paramLocators = locators.stream()
+    List<KeyLocator> paramKeyLocators = keyLocators.stream()
```
tdas: what does paramKeyLocators mean?? param here is ambiguous.

in fact, what is the different between AuthorizedParam and AuthorizedKey with source = PARAM ??

server/src/main/java/io/unitycatalog/server/service/TemporaryModelVersionCredentialsService.java
```diff
@@ -1,61 +1,60 @@
...
                                                  Repositories repositories) {
-    this.evaluator = new UnityAccessEvaluator(authorizer);
     this.cloudCredentialVendor = cloudCredentialVendor;
-    this.keyMapper = new KeyMapper(repositories);
     this.modelRepository = repositories.getModelRepository();
-    this.userRepository = repositories.getUserRepository();
   }
 
   @Post("")
+  @AuthorizeExpression("""
```
tdas, line 43: what is the change happening in this file? 

if this PR is not just introducing the annotation but its also changing production logic. please make sure the title and description matches this.

tdas (changes requested):
bunch of questions. 

but overall.. the difference between AuthorizeKey (which itself can be on a PARAM) and AuthorizeParam needs to be made clear. its not yet. 

In fact, its making me question on why we cant just use the AuthorizeKey for this purpose? is it because authorize key is tied to resource (that resource id is the "key") and authorizeparam can be arbitrary. if so, lets make this clear by naming things better - - AuthorizeResource (or AuthorizeResourceKey, instead of AuthorizeKey) 
   - can be instantiated by various source of info including param, system, etc.
   - must be used for resource only
- AuthorizeParam 
   - can be only be instantiated with request parameters as the source (is it?) 
   - not recommended to be used for resources.

server/src/main/java/io/unitycatalog/server/service/TemporaryModelVersionCredentialsService.java
yili-db: In all these `*Service.java` files, the production logic is unchanged.
These changes are using the new non-resource `@AuthorizeParam` to construct a new auth expression to replace the manual `authorizeForOperation` function. And that is the major purpose of introducing this new annotation.

server/src/main/java/io/unitycatalog/server/auth/annotation/AuthorizeParam.java
yili-db: Yes that's right.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
yili-db: Renamed.

yili-db:
@tdas To address your comments, I have made the following rename/adjustments:
- `@AuthorizeResourceKey`: This is renamed from existing `@AuthorizeKey`. Annotation for resource keys.
- `@AuthorizeKey`: This was the `@AuthorizeParam`. Renamed to avoid the "param" in name. Annotation for all keys (also resource if the purpose is just for checking for null). No resource id mapping.
- `KeyLocator`: This class is extended to cover BOTH annotations. No more `ParamLocator`.

The two annotations are now very similar except that `@AuthorizeKey` doesn't have a securable type. However they can still not use the same annotation class name because annotation must have a value for securable type if there's a field for it. They can't share any base class neither.
But `KeyLocator` is now extended to handle both of them.

Please take another look. Thank you!

server/src/main/java/io/unitycatalog/server/auth/annotation/AuthorizeKey.java
yili-db, line 1: This is actually a new file. The old `AuthorizeKey.java‎` was renamed to `AuthorizeResourceKey.java‎`

server/src/main/java/io/unitycatalog/server/auth/annotation/AuthorizeResourceKey.java
yili-db, line 1: This is the old AuthorizeKey.java‎ file renamed.

tdas:
That all makes sense to me. Thanks for the refactor...I think this is a more understandable (and also lesser) code.

yili-db:
@tdas Also for ease of review, I have two commits in this PR with one of them doing just the rename. You can check them individually to get a better view because otherwise git doesn't have the correct rename relationship.

tdas:
Would it be possible to split the rename into a separate pr?

yili-db:
> Would it be possible to split the rename into a separate pr?

@tdas 
Yes. https://github.com/unitycatalog/unitycatalog/pull/1307
This PR is also rebased.

server/src/main/java/io/unitycatalog/server/auth/annotation/AuthorizeResourceKey.java
```diff
@@ -0,0 +1,62 @@
...
+
+import io.unitycatalog.server.model.SecurableType;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Map a request parameter to a unity catalog resource key.
```
tdas, line 11: Add a note differentiating this from authorizekey

server/src/main/java/io/unitycatalog/server/auth/decorator/KeyLocator.java
```diff
@@ -1,19 +1,79 @@
...
+ * <p>Example: For {@code @AuthorizeKey(key = "operation")}, this locator would extract the
+ * "operation" field value from the request payload and make it available as "#operation" in the
+ * SpEL authorization expression.
+ *
+ * @see AuthorizeKey
+ * @see AuthorizeResourceKey
+ */
 @Builder
 @Getter
 public class KeyLocator {
```
tdas, line 29: I still think this should be renamed to authorizeKeyLocator

tdas: Follow up pr

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -41,18 +40,18 @@
...
- * and @AuthorizeKey to define authorization rules and identify requests parameters for objects
- * to authorize with.
- * <p>
- * {@code @AuthorizeExpression} - This defines a Spring Expression Language expression to evaluate
- * to make an authorization decision.
- * {@code @AuthorizeKey} - This annotation is used to define request and payload parameters for the
- * authorization context. These are typically things like catalog, schema and table names. This
- * annotation may be used at both the method and method parameter context. It may be specified
- * more than once per method to map parameters to object keys.
+ * and @AuthorizeResourceKey to define authorization rules and identify requests parameters for
```
tdas: This does not mention authorizekey

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -118,6 +117,7 @@ private HttpResponse authorizeByRequest(
     // we want to authorize against.
 
     Map<SecurableType, Object> resourceKeys = new HashMap<>();
+    Map<String, Object> otherKeys = new HashMap<>();
```
tdas: nonResourceKeys

By itself "other" is not very meaningful

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -228,50 +235,24 @@ private static List<KeyLocator> findAuthorizeKeys(Method method) {
...
-            locators.add(KeyLocator.builder()
-                .source(PARAM)
-                .type(key.value())
-                .key(param.value())
-                .build());
-          } else {
-            LOGGER.warn("Couldn't find param key for authorization key");
-          }
-        }
+      AuthorizeKey[] allOtherKeys = parameter.getAnnotationsByType(AuthorizeKey.class);
```
tdas: nonResourceKeys

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessEvaluator.java
```diff
@@ -75,8 +75,25 @@ protected boolean authorizeAll(Object... parameters) {
...
+   * @param principal The principal UUID
+   * @param expression The SpEL authorization expression
+   * @param resourceIds Map of resource types to their UUIDs
+   * @param otherKeys Map of parameter names to their values (from @AuthorizeKey)
+   */
+  public boolean evaluate(
+      UUID principal,
+      String expression,
+      Map<SecurableType, Object> resourceIds,
+      Map<String, Object> otherKeys) {
```
tdas: nonResourceValues

It's not keys any more but map of values from the request isnt it?

tdas (changes requested):
Please resolve the merged previous or in the stack and solve this.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessEvaluator.java
yili-db: Yes. Actually all the `otherKeys` are values from the beginning.

PR #1309, opened 2026-01-19
Code refactor: Use NormalizedURL in credential vending code paths

yili-db:
**PR Checklist**

**Description of changes**

Code refactor: Use NormalizedURL in credential vending code paths

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
@@ -188,10 +188,6 @@ public class FileOperations {
     }
   }
 
-  public static void assertValidLocation(String location) {
-    UriUtils.validateURI(URI.create(location));
-  }
-
   private String getManagedTablesStorageRoot() {
     // Use local tmp directory as default storage root
     return serverProperties.get(Property.TABLE_STORAGE_ROOT);
--- a/server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
+++ b/server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
@@ -6,7 +6,6 @@ import io.unitycatalog.server.model.AwsCredentials;
 import io.unitycatalog.server.model.AzureUserDelegationSAS;
 import io.unitycatalog.server.model.GcpOauthToken;
 import io.unitycatalog.server.model.TemporaryCredentials;
-import io.unitycatalog.server.persist.utils.FileOperations;
 import io.unitycatalog.server.service.credential.aws.AwsCredentialVendor;
 import io.unitycatalog.server.service.credential.azure.AzureCredential;
 import io.unitycatalog.server.service.credential.azure.AzureCredentialVendor;
@@ -45,9 +44,6 @@ public class CloudCredentialVendor {
   }
 
   public TemporaryCredentials vendCredential(CredentialContext context) {
-    String location = context.getLocations().get(0);
-    FileOperations.assertValidLocation(location);
-
     TemporaryCredentials temporaryCredentials = new TemporaryCredentials();
 
     switch (context.getStorageScheme()) {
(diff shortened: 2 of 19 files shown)


server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -188,10 +188,6 @@ private URI modifyS3Directory(URI parsedUri, boolean createOrDelete) {
     }
   }
 
-  public static void assertValidLocation(String location) {
```
yili-db, line 191: The only usage of this function is removed in this PR

server/src/main/java/io/unitycatalog/server/service/credential/azure/AzureCredentialVendor.java
```diff
@@ -1,31 +1,28 @@
 package io.unitycatalog.server.service.credential.azure;
 
 import io.unitycatalog.server.service.credential.CredentialContext;
+import io.unitycatalog.server.utils.NormalizedURL;
 import io.unitycatalog.server.utils.ServerProperties;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 public class AzureCredentialVendor {
   private final Map<String, ADLSStorageConfig> adlsConfigurations;
```
yili-db, line 10: This map is unchanged because the key is storage account name, not URL

server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
```diff
@@ -50,9 +49,6 @@ public TemporaryCredentials vendCredential(
   }
 
   public TemporaryCredentials vendCredential(CredentialContext context) {
-    String location = context.getLocations().get(0);
```
yili-db, line 48: NormalizedURL no longer needs validation

server/src/main/java/io/unitycatalog/server/service/credential/azure/AzureCredentialVendor.java
tdas: I don't get it. Doesn't the whole normalizedurl contain the full object path?

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -253,8 +253,8 @@ public ServerProperties(Properties inputProperties) {
   private void validateProperties() {
     for (Property property : Property.values()) {
       String value = get(property);
-      if (value == null || value.isBlank()) {
-        // null or empty values are OK.
+      if (value == null) {
+        // null means not set.
```
tdas, line 257: How is this change related to the refactor?

server/src/test/java/io/unitycatalog/server/utils/NormalizedURLTest.java
```diff
@@ -83,4 +83,18 @@ public void testToStandardizedURIString() {
...
+  @Test
+  public void testGetStorageBase() {
+    assertThat(NormalizedURL.from("s3://bucket/path").getStorageBase())
+        .isEqualTo(NormalizedURL.from("s3://bucket"));
+    assertThat(NormalizedURL.from("s3://bucket/path/to/file").getStorageBase())
+        .isEqualTo(NormalizedURL.from("s3://bucket"));
+    assertThat(NormalizedURL.from("gs://bucket/path").getStorageBase())
+        .isEqualTo(NormalizedURL.from("gs://bucket"));
+    assertThat(
+            NormalizedURL.from("abfs://container@account.dfs.core.windows.net/path")
```
tdas, line 96: No test for abfss?

server/src/main/java/io/unitycatalog/server/service/credential/azure/AzureCredentialVendor.java
yili-db: This map specifically only wants the **storage account name** portion of the URL authority. I think the purpose is so that URLs of multiple containers in the same account can share the same config. So I just leave it as is.
For example the `storageBase` (`NormalizedURL`) would look like this: "`abfss://<container>@<account>.dfs.core.windows.net/`". But this map wants to have one single config for "`<account>`"

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
yili-db: This is because empty values are already removed by line 279.

And line 279 is because the empty string is not accepted as a valid `NormalizedURL.from` even if the empty string value is never used in the code path. So removing them in line 279 makes sense.

tdas:
please resolve conflicts then i can merge

PR #1311, opened 2026-01-20
add view support

TheThow:
**PR Checklist**

**Description of changes**

Related issue: https://github.com/unitycatalog/unitycatalog/issues/1279

* Adds Views as a special Table entity to the unity catalog
* Integrates views into Iceberg and UC API
* Does not support view versioning or updates yet

View may be created via the CLI like
```
bin/uc view create --full_name unity.default.my_view --columns "id INT, name STRING, marks INT" --view_definition '[{"dialect":"spark","sql":"SELECT * FROM `default`.`marksheet_uniform`"}]'
```

TODOs:
* fix pagination for both tables and views - seems a bit tricky

Happy to split this into smaller chunks and make adjustments to make it work with the current state of spark

### DIFF
--- a/api/delta-docs/Apis/DeltaTablesApi.md
+++ b/api/delta-docs/Apis/DeltaTablesApi.md
@@ -107,7 +107,7 @@ No authorization required
 
 Load table metadata
 
-    Load table metadata including columns, properties, and optionally credentials. 
+    Load table metadata including columns, properties, unbackfilled commits, and the latest table version. Credentials are not included; clients obtain them from the dedicated credentials endpoints. 
 
 ### Parameters
 
--- a/api/delta-docs/Models/DeltaCommit.md
+++ b/api/delta-docs/Models/DeltaCommit.md
@@ -5,7 +5,7 @@
 |------------ | ------------- | ------------- | -------------|
 | **version** | **Long** | Commit version | [default to null] |
 | **timestamp** | **Long** | In-commit timestamp, in epoch milliseconds | [default to null] |
-| **file-name** | **String** | UUID-based commit file name | [default to null] |
+| **file-name** | **String** | UUID-based name of the staged commit file under _delta_log/_staged_commits/. Written by the committing client and returned verbatim to readers, which read the file at that path; when the same version is also published as _delta_log/&lt;v&gt;.json, this catalog-named file is the authoritative copy.  | [default to null] |
 | **file-size** | **Long** | Commit file size in bytes | [default to null] |
 | **file-modification-timestamp** | **Long** | File modification timestamp, in epoch milliseconds | [default to null] |
 
--- a/api/delta-docs/Models/DeltaLoadTableResponse.md
+++ b/api/delta-docs/Models/DeltaLoadTableResponse.md
@@ -4,7 +4,7 @@
 | Name | Type | Description | Notes |
 |------------ | ------------- | ------------- | -------------|
 | **metadata** | [**DeltaTableMetadata**](DeltaTableMetadata.md) | Complete table metadata including schema and properties | [default to null] |
-| **commits** | [**List**](DeltaCommit.md) | All unbackfilled CCv2 commits | [optional] [default to null] |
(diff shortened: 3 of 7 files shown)


yili-db:
Hi @TheThow,

Thanks for putting this together, and for offering to split it up. View support is genuinely wanted (#1279).

Before this grows further, I'd like to suggest a different modeling approach that I think will fit the codebase better and avoid a parallel object hierarchy. The repo already has a view-like object shipped today: `METRIC_VIEW`. Rather than introducing a separate `View` entity (`ViewInfo`, `view_id`, `/views`, `ViewService`, `ViewRepository`, `ViewRepresentationDAO`), I'd propose `VIEW` follow the exact same pattern that `METRIC_VIEW` already uses.

## How `METRIC_VIEW` is implemented today

A metric view is **not** a separate entity. It's a value of the existing `TableType` enum, created and read through the normal `/tables` API. There is no `/metric-views` endpoint and no `MetricViewService`.

- **Type**: `TableType` already contains `MANAGED`, `EXTERNAL`, `MATERIALIZED_VIEW`, `METRIC_VIEW`, `STREAMING_TABLE` (`api/all.yaml`). Of these, `METRIC_VIEW` is the one with a full create/read implementation today; `MATERIALIZED_VIEW` and `STREAMING_TABLE` exist in the enum but `TableRepository` currently rejects their creation as not-yet-supported. So the enum is the established extension point for new table-like kinds, and `METRIC_VIEW` is the worked example to copy.
- **Model**: the SQL and lineage live on the existing `CreateTable` / `TableInfo` schemas as `view_definition` (the SQL text) and `view_dependencies` (a `DependencyList`). No new top-level model.
- **Create**: `TableService.createTable` -> `TableRepository`, branching on `table_type`:
  ```java
  } else if (tableType == TableType.METRIC_VIEW) {
    storageLocation = null;
    validateMetricView(createTable);
    tableUUID = UUID.randomUUID();
  }
  ...
  .viewDefinition(createTable.getViewDefinition())
  ...
  if (tableType == TableType.METRIC_VIEW) {
    // persist view_dependencies via the shared DependencyRepository
  }
  ```
  `validateMetricView` just checks that `view_definition` and `view_dependencies` are present.
- **Read**: the normal `getTable` path. The view-specific bit is a small `table_type` branch in `RepositoryUtils` that loads `view_dependencies` back from the shared dependency store.
- **Persistence**: one `TableInfoDAO` row with `type = METRIC_VIEW` + `view_definition`, columns in the usual columns table, and dependency rows in the shared `DependencyRepository` / `DependencyDAO`. No view-specific table.

So a metric view reuses the table RPCs, the table DAO, the columns store, and the dependency machinery end to end.

## Proposal: model `VIEW` the same way

1. **Add `VIEW` to the `TableType` enum** instead of a separate `ViewInfo` / `view_id`. It's the missing sibling next to `MATERIALIZED_VIEW` / `METRIC_VIEW`.
2. **Use the existing `/tables` API** (`createTable` / `getTable` / `listTables` / `deleteTable`) with `table_type = VIEW`, rather than a parallel `/views` family + `ViewService` + `ViewRepository`. This drops a large amount of the new surface in the PR.
3. **Store the view SQL in the existing `view_definition` field**, lineage in `view_dependencies` via the shared `DependencyRepository`, and columns through the standard table columns path -- exactly as `METRIC_VIEW` does, so no view-specific DAOs are needed. Add a `validateView` modeled on `validateMetricView`.

Benefits: no second object hierarchy to keep in sync, views inherit table listing / privileges / dependency handling for free, and it stays consistent with the view-like object the project already ships.

Happy to help review once it's reframed around the `TableType` approach, and to point at the exact `METRIC_VIEW` call sites if useful.

TheThow:
Hi @yili-db thanks for the comment! I'll create a new PR for your proposal.

(I accidentally also deleted this PR prematurely by clicking the UI - those were the initially proposed changes for the future: https://github.com/unitycatalog/unitycatalog/compare/main...TheThow:unitycatalog:view-support)

(never using the git UI again lol)

TheThow:
Hi @yili-db - please have a look at https://github.com/unitycatalog/unitycatalog/pull/1637

PR #1312, opened 2026-01-20
Refactor schema&catalog repo: centralize DAO lookups and improve null-safety

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1312/files) to review incremental changes.
- [**stack/catalog_schema_repo**](https://github.com/unitycatalog/unitycatalog/pull/1312) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1312/files)]

---------
**PR Checklist**

**Description of changes**
Refactor schema&catalog repo: centralize DAO lookups and improve null-safety

Key Changes

- The lookups of schema by name is actually involving finding both catalog and schema. Often callers will need to find both. This PR moves the lookup to RepositoryUtils.
- Add `getCatalogDaoOpt()` and `getSchemaDaoOpt()` returning Optional<DAO> for safe lookups
- Add `getCatalogAndSchemaDaoOpt()` and `getCatalogAndSchemaDaoOrThrow()` for combined lookups
- Add `getXOrThrow()` to automatically fail with NOT_FOUND. Remove repetitive null-check-and-throw patterns throughout repositories
- Replace `Pair<String, String>` with `CatalogAndSchemaNames` record for better semantics
- Rename `processChild*` methods to `deleteChild*` for clarity (tables, volumes, functions, models)

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
@@ -23,7 +23,6 @@ import java.util.Optional;
 import java.util.UUID;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
-import org.hibernate.query.Query;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -58,7 +57,7 @@ public class CatalogRepository {
     return TransactionManager.executeWithTransaction(
         sessionFactory,
         session -> {
-          if (getCatalogDAO(session, createCatalog.getName()) != null) {
+          if (RepositoryUtils.getCatalogDaoOpt(session, createCatalog.getName()).isPresent()) {
             throw new BaseException(
                 ErrorCode.ALREADY_EXISTS, "Catalog already exists: " + createCatalog.getName());
           }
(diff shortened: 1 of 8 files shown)


server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
```diff
@@ -120,19 +116,13 @@ public CatalogInfo getCatalog(String name) {
         /* readOnly = */ true);
   }
 
-  public CatalogInfoDAO getCatalogDAO(Session session, String name) {
-    Query<CatalogInfoDAO> query =
-        session.createQuery("FROM CatalogInfoDAO WHERE name = :value", CatalogInfoDAO.class);
-    query.setParameter("value", name);
-    query.setMaxResults(1);
-    return query.uniqueResult();
+  public CatalogInfoDAO getCatalogDaoOrThrow(Session session, String name) {
```
tdas, line 119:
```suggestion
  public CatalogInfoDAO getCatalogDAOOrThrow(Session session, String name) {
```
more consistent with other places with DAO in the names?

tdas: if you cahnge, apply it to all changes in this PR

yili-db: @tdas 
`DAO` is indeed more common in the code base. But if there's any upper case letter right after `DAO`, it causes a "uppercase collision".
In this case `DAOOr` is less readable than `DaoOr`. I guess for that reason the existing function `getRegisteredModelDaoOrThrow` was named as such.
If you do think the consistency is more important in this case, I can rename it.

tdas: good point. let it be like this.

PR #1313, opened 2026-01-21
Add storage credential vending via external locations

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1313/files) to review incremental changes.
- [**stack/sc_vend**](https://github.com/unitycatalog/unitycatalog/pull/1313) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1313/files)]

---------
**PR Checklist**

**Description of changes**
Add storage credential vending via external locations

Introduce StorageCredentialVendor to vend temporary cloud credentials using storage credentials associated with external locations. The UC server assumes IAM roles specified in external location credentials with external IDs.

Key changes:
- Add StorageCredentialVendor as entry point for credential vending
- Refactor CredentialContext to include Optional<CredentialDAO>
- Add master role STS client in AwsCredentialVendor for assuming storage roles
- Add getExternalLocationCredentialForPath in ExternalLocationUtils

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
@@ -1,19 +1,10 @@
 package io.unitycatalog.server.persist.utils;
 
-import com.amazonaws.auth.AWSStaticCredentialsProvider;
-import com.amazonaws.auth.BasicSessionCredentials;
-import com.amazonaws.services.s3.AmazonS3;
-import com.amazonaws.services.s3.AmazonS3ClientBuilder;
-import com.amazonaws.services.s3.model.ListObjectsRequest;
-import com.amazonaws.services.s3.model.ObjectListing;
-import com.amazonaws.services.s3.model.ObjectMetadata;
-import com.amazonaws.services.s3.model.PutObjectRequest;
 import io.unitycatalog.server.exception.BaseException;
 import io.unitycatalog.server.exception.ErrorCode;
 import io.unitycatalog.server.utils.NormalizedURL;
 import io.unitycatalog.server.utils.ServerProperties;
 import io.unitycatalog.server.utils.ServerProperties.Property;
-import java.io.ByteArrayInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URI;
@@ -23,11 +14,9 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Comparator;
 import java.util.stream.Stream;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class FileOperations {
-  private static final Logger LOGGER = LoggerFactory.getLogger(FileOperations.class);
+
   private final ServerProperties serverProperties;
   private static String modelStorageRootCached;
   private static String modelStorageRootPropertyCached;
@@ -107,8 +96,6 @@ public class FileOperations {
       } catch (RuntimeException | IOException e) {
         throw new BaseException(ErrorCode.INTERNAL, "Failed to delete directory: " + path, e);
       }
-    } else if (directoryUri.getScheme().equals("s3")) {
-      modifyS3Directory(directoryUri, false);
     } else {
       throw new BaseException(
           ErrorCode.INVALID_ARGUMENT, "Unsupported URI scheme: " + directoryUri.getScheme());
@@ -130,61 +117,6 @@ public class FileOperations {
       }
     } else {
       throw new FileNotFoundException("Directory does not exist: " + dirPath);
-    }
-  }
-
-  private URI modifyS3Directory(URI parsedUri, boolean createOrDelete) {
-    String bucketName = parsedUri.getHost();
-    String path = parsedUri.getPath().substring(1); // Remove leading '/'
-    String accessKey = serverProperties.get(Property.AWS_ACCESS_KEY);
-    String secretKey = serverProperties.get(Property.AWS_SECRET_KEY);
-    String sessionToken = serverProperties.get(Property.AWS_SESSION_TOKEN);
-    String region = serverProperties.get(Property.AWS_REGION);
-
-    BasicSessionCredentials sessionCredentials =
-        new BasicSessionCredentials(accessKey, secretKey, sessionToken);
-    AmazonS3 s3Client =
-        AmazonS3ClientBuilder.standard()
-            .withCredentials(new AWSStaticCredentialsProvider(sessionCredentials))
-            .withRegion(region)
-            .build();
-
-    if (createOrDelete) {
-
-      if (!path.endsWith("/")) {
-        path += "/";
-      }
-      if (s3Client.doesObjectExist(bucketName, path)) {
-        throw new BaseException(ErrorCode.ALREADY_EXISTS, "Directory already exists: " + path);
-      }
-      try {
-        // Create empty content
-        byte[] emptyContent = new byte[0];
-        ByteArrayInputStream emptyContentStream = new ByteArrayInputStream(emptyContent);
-
-        // Set metadata for the empty content
-        ObjectMetadata metadata = new ObjectMetadata();
-        metadata.setContentLength(0);
-        s3Client.putObject(new PutObjectRequest(bucketName, path, emptyContentStream, metadata));
-        LOGGER.debug("Directory created successfully: {}", path);
-        return URI.create(String.format("s3://%s/%s", bucketName, path));
-      } catch (Exception e) {
-        throw new BaseException(ErrorCode.INTERNAL, "Failed to create directory: " + path, e);
-      }
-    } else {
-      ObjectListing listing;
-      ListObjectsRequest req = new ListObjectsRequest().withBucketName(bucketName).withPrefix(path);
-      do {
-        listing = s3Client.listObjects(req);
-        listing
-            .getObjectSummaries()
-            .forEach(
-                object -> {
-                  s3Client.deleteObject(bucketName, object.getKey());
-                });
-        req.setMarker(listing.getNextMarker());
-      } while (listing.isTruncated());
-      return URI.create(String.format("s3://%s/%s", bucketName, path));
     }
   }
 


server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
```diff
@@ -31,29 +84,32 @@ private CredentialsGenerator createCredentialsGenerator(S3StorageConfig config)
 
     if (config.getSessionToken() != null && !config.getSessionToken().isEmpty()) {
       // if a session token was supplied, then we will just return static session credentials
-      return new CredentialsGenerator.StaticCredentialsGenerator(
-          config.getAccessKey(), config.getSecretKey(), config.getSessionToken());
+      return new CredentialsGenerator.StaticCredentialsGenerator(config);
     }
 
-    if (config.getAccessKey() != null && !config.getAccessKey().isEmpty()) {
```
yili-db, line 38: This `if` is moved into constructor of `CredentialsGenerator.StsCredentialsGenerator`

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
```diff
@@ -31,29 +84,32 @@ private CredentialsGenerator createCredentialsGenerator(S3StorageConfig config)
...
   }
 
   public Credentials vendAwsCredentials(CredentialContext context) {
-    S3StorageConfig config = s3Configurations.get(context.getStorageBase());
-    if (config == null) {
-      throw new BaseException(ErrorCode.FAILED_PRECONDITION, "S3 bucket configuration not found.");
+    CredentialsGenerator generator;
+    if (context.getCredentialDAO().isPresent()) {
+      // Use the master role STS generator
+      generator = getAwsS3MasterRoleStsGenerator();
```
yili-db, line 130: This is the new path and the `else` is the old one.

server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
```diff
@@ -31,64 +33,46 @@ public CloudCredentialVendor(
     this.gcpCredentialVendor = gcpCredentialVendor;
   }
 
-  public TemporaryCredentials vendCredential(
-      NormalizedURL path,
```
yili-db, line 35: Moved to `StorageCredentialVendor`

server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
```diff
@@ -31,64 +33,46 @@ public CloudCredentialVendor(
...
-    URI storageLocationUri = path.toUri();
-    // TODO: At some point, we need to check if user/subject has privileges they are asking for
-    CredentialContext credentialContext = CredentialContext.create(storageLocationUri, privileges);
-    return vendCredential(credentialContext);
-  }
-
   public TemporaryCredentials vendCredential(CredentialContext context) {
-    TemporaryCredentials temporaryCredentials = new TemporaryCredentials();
-
-    switch (context.getStorageScheme()) {
```
yili-db, line 49: This `switch` is split up into different functions

server/src/main/java/io/unitycatalog/server/service/credential/CredentialContext.java
```diff
@@ -16,20 +23,39 @@ public enum Privilege {
...
+  private final UriScheme storageScheme;
+  // The storage base (e.g., s3://bucket) for looking up per-bucket configurations
+  private final NormalizedURL storageBase;
+  // The privileges (SELECT, UPDATE) that determine read/write permissions
+  private final Set<Privilege> privileges;
+  // The specific locations that credentials should grant access to. This is a list of locations to
+  // be a little future-proofing when a table could have more than 1 location where files belonging
+  // to table are located.
+  private final List<NormalizedURL> locations;
+  // If a storage credential of external location for this path is found, the credential will
```
yili-db, line 36: All fields above remain as is but `final` was added.
Only `credentialDAO` is the new field.

server/src/main/java/io/unitycatalog/server/service/iceberg/TableConfigService.java
```diff
@@ -14,68 +16,67 @@
...
+  private final StorageCredentialVendor storageCredentialVendor;
   private final Map<NormalizedURL, S3StorageConfig> s3Configurations;
+  // FIXME!! privileges are defaulted to READ only here for now as Iceberg REST impl doesn't
+  // support write
+  private final Set<CredentialContext.Privilege> privileges =
+      Set.of(CredentialContext.Privilege.SELECT);
 
-  public TableConfigService(CloudCredentialVendor cloudCredentialVendor,
-      ServerProperties serverProperties) {
+  public TableConfigService(
```
yili-db, line 32: Also this class can be rewritten in a much better way by utilizing `FileIOFactory` instead of duplicating lots of code.
For now I am just replacing `cloudCredentialVendor.vendAwsCredential` etc with the cloud agnostic version `storageCredentialVendor.vendCredential` to keep them working as is.
The improvement can be another PR.

server/src/main/java/io/unitycatalog/server/service/iceberg/FileIOFactory.java
```diff
@@ -23,19 +24,22 @@
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.s3.S3Client;
-import software.amazon.awssdk.services.sts.model.Credentials;
 
 import java.util.Map;
 import java.util.Set;
 
 public class FileIOFactory {
```
yili-db, line 31: This class can be rewritten in a better way.
For now I am just replacing cloudCredentialVendor.vendAwsCredential etc with the cloud agnostic version storageCredentialVendor.vendCredential to keep them working as is. This also caused some diffs here.

server/src/main/java/io/unitycatalog/server/utils/NormalizedURL.java
```diff
@@ -127,6 +118,9 @@ public URI toUri() {
    * </pre>
    */
   public static String normalize(String inputPath) {
+    if (inputPath == null || inputPath.isBlank()) {
```
yili-db, line 121: This check is not new logic, but simply prints out better error message.
If it's null or empty, previously it would fail at the line below with a general error message.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -102,18 +96,6 @@ private void setDefaults(UnityCatalogServer.Builder unityCatalogServerBuilder) {
...
     }
-    if (unityCatalogServerBuilder.cloudCredentialVendor == null) {
-      AwsCredentialVendor awsCredentialVendor =
-          new AwsCredentialVendor(unityCatalogServerBuilder.serverProperties);
-      AzureCredentialVendor azureCredentialVendor =
-          new AzureCredentialVendor(unityCatalogServerBuilder.serverProperties);
-      GcpCredentialVendor gcpCredentialVendor =
-          new GcpCredentialVendor(unityCatalogServerBuilder.serverProperties);
-      CloudCredentialVendor cloudCredentialVendor =
-          new CloudCredentialVendor(
```
yili-db, line 113: Moved to a later stage so that a `StorageCredentialVendor` can be constructed after `getExternalLocationUtils` is ready.

server/src/test/java/io/unitycatalog/server/base/credential/BaseCredentialCRUDTest.java
```diff
@@ -31,7 +32,6 @@ public abstract class BaseCredentialCRUDTest extends BaseCRUDTest {
   private static final String NEW_ROLE_ARN = "arn:aws:iam::987654321098:role/new-role-name";
   private static final String EXTERNAL_LOCATION_NAME = "uc_testexternallocation";
   private static final String URL = "s3://unitycatalog-test";
-  private static final String S3_MASTER_ROLE_ARN = "arn:aws:iam::1234567:role/UCMasterRole-EXAMPLE";
```
yili-db, line 34: Moved to `TestUtils`

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
```diff
@@ -64,20 +129,18 @@ protected void setUpCredentialOperations() {
...
-            .sessionToken("test-session-token")
-            .build();
-    serverProperties.entrySet().stream()
-            .filter(e -> e.getKey().toString().startsWith("s3.bucketPath"))
-            .map(e -> e.getValue().toString())
-            .forEach(path -> doReturn(awsCredential)
-                    .when(awsCredentialVendor)
-                    .vendAwsCredentials(
-                            argThat(isCredentialContextForCloudPath(UriScheme.S3, path))));
+  private void setupAwsCredentials(ServerProperties serverProperties) {
```
yili-db, line 150: Previously it doesn't exercise the STS assume with role ARN but simply return a preset credential before it reaches the STS code.
With the new setup it goes all the way down to:
```
return stsClient.assumeRole(roleRequestBuilder.build()).credentials();  // <-- this is the new mocking point
```
This setup works for all four of `SdkTemporary*CredentialTest.java`

server/src/test/java/io/unitycatalog/server/service/credential/CloudCredentialVendorTest.java
```diff
@@ -178,11 +201,90 @@ public void testMissingGcpBucketConfigurationFails() {
...
-                credentialsOperations.vendCredential(
-                    NormalizedURL.from("gs://missing/abc"),
-                    Set.of(CredentialContext.Privilege.UPDATE)))
+            () -> vendCredential("gs://missing/abc", Set.of(CredentialContext.Privilege.UPDATE)))
         .isInstanceOf(BaseException.class)
         .hasMessageContaining("Unknown GCS storage configuration");
   }
+
+  @Test
+  public void testVendCredentialWithExternalLocationCredential() {
```
yili-db, line 210: This is a mocked **unit** test.

server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
```diff
@@ -31,64 +33,46 @@ public CloudCredentialVendor(
...
-
-        // Explicitly set the expiration time for the temporary credentials if it's a non-static
-        // credential. For static credential, the expiration time can be nullable.
-        if (awsSessionCredentials.expiration() != null) {
-          temporaryCredentials.expirationTime(awsSessionCredentials.expiration().toEpochMilli());
-        }
-      }
+    return switch (context.getStorageScheme()) {
+      case ABFS, ABFSS -> vendAzureCredential(context);
+      case GS -> vendGcpToken(context);
```
tdas: nit:
why token and not credential. 
and its specifically GCS credential not GCP .. right? 

similarly S3Credential

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -221,6 +222,56 @@ public static List<Pair<SecurableType, IdentifiableDAO>> getAllEntityDAOsWithURL
...
+   * then returns the storage credential associated with that external location. The credential can
+   * be used to vend a temporary credential for data access.
+   *
+   * @param url the storage path to find credentials for (e.g., s3://bucket/path/to/data)
+   * @return the storage credential if an external location covers this path, empty otherwise
+   * @throws BaseException with FAILED_PRECONDITION if multiple external locations cover the path
+   *     (invalid state) or if the credential referenced by the external location does not exist
+   */
+  public Optional<CredentialDAO> getExternalLocationCredentialForPath(NormalizedURL url) {
+    return TransactionManager.executeWithTransaction(
```
tdas, line 238: nit: this is very nested inside a return.. with returns inside the return code block. confusing to read.  can this broken down to return the daos into a variable and then rest of the logic to use the daos is outside the executeWithTxn?

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -221,6 +222,56 @@ public static List<Pair<SecurableType, IdentifiableDAO>> getAllEntityDAOsWithURL
...
+        session -> {
+          List<ExternalLocationDAO> externalLocationDAOs =
+              ExternalLocationUtils.<ExternalLocationDAO>getEntityDAOsWithURLOverlap(
+                      session,
+                      url,
+                      SecurableType.EXTERNAL_LOCATION,
+                      /* limit= */ 2,
+                      /* includeParent= */ true,
+                      /* includeSelf= */ true,
+                      /* includeSubdir= */ false)
```
tdas: can you comment on why these set of params? the logic behind them?

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -221,6 +222,56 @@ public static List<Pair<SecurableType, IdentifiableDAO>> getAllEntityDAOsWithURL
...
+   * <p>This method looks up external locations whose URL is a parent of or equals the given path,
+   * then returns the storage credential associated with that external location. The credential can
+   * be used to vend a temporary credential for data access.
+   *
+   * @param url the storage path to find credentials for (e.g., s3://bucket/path/to/data)
+   * @return the storage credential if an external location covers this path, empty otherwise
+   * @throws BaseException with FAILED_PRECONDITION if multiple external locations cover the path
+   *     (invalid state) or if the credential referenced by the external location does not exist
+   */
+  public Optional<CredentialDAO> getExternalLocationCredentialForPath(NormalizedURL url) {
```
tdas: this is not returning credential!!! only the credential DAO. this was very confusing to read.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
```diff
@@ -7,18 +7,71 @@
...
 
+/**
+ * Vends AWS credentials for accessing S3 storage.
+ *
+ * <p>This class supports two modes of credential vending:
+ *
+ * <ol>
+ *   <li><b>External Location Credentials:</b> When a storage credential (CredentialDAO) is
+ *       associated with an external location, the master role STS client assumes the storage role
+ *       specified in the credential with the appropriate external ID.
```
tdas: can you explain here what is external id? 

actually... related to my previous comment in the previous PR... i think this is best place to put in the AWS specific details. so please put here the workflow of how all this master role and STS and assume role with external id works like i am five :D

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
```diff
@@ -7,18 +7,71 @@
...
+   */
+  public AwsCredentialVendor(
+      ServerProperties serverProperties, Supplier<StsClientBuilder> stsClientBuilderSupplier) {
+    this.stsClientBuilderSupplier = stsClientBuilderSupplier;
+    this.perBucketS3Configs = serverProperties.getS3Configurations();
+    // awsS3MasterRoleConfig.awsRoleArn is null. The ARN to assume comes from credential securable.
+    this.awsS3MasterRoleConfig = serverProperties.getS3MasterRoleConfiguration();
+  }
+
+  private CredentialsGenerator createPerBucketCredentialsGenerator(S3StorageConfig config) {
```
tdas, line 103: this is only used for the static credentials set in the server properties.. right? 
if so please add it to the method docs.

tdas: actually.. i am confused .. it returns `createStsCredentialsGenerator` at the end.


i think this and createStsCredentialsGenerator definitely needs docs

server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java
```diff
@@ -35,6 +35,15 @@ public GcpCredentialVendor(ServerProperties serverProperties) {
   }
 
   public AccessToken vendGcpToken(CredentialContext credentialContext) {
+    credentialContext
```
tdas, line 38: what is the expected behavior with this? if there is server prop setting for GCS... that should continue working .. right?

server/src/main/java/io/unitycatalog/server/service/credential/StorageCredentialVendor.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * <p>This class is responsible for:
+ *
+ * <ul>
+ *   <li>Looking up external location credentials based on the requested path
+ *   <li>Delegating to cloud-specific credential vendors (AWS, Azure, GCP)
+ *   <li>Returning scoped temporary credentials with appropriate privileges
+ * </ul>
+ */
+public class StorageCredentialVendor {
```
tdas, line 23: i am wondering why do we need this layer above. why not update CloudCredentialVendor to deal with credentialDAO with ExternalLocationUtils?

tdas: pros
- we wont have two similarly named classes.
- maybe, all usages of CloudCredentialVendor will not require refactoring? though i see that it will be nice to refactor a bit to make all places just call `vendCredential(path, priv)` instead of having to create credcontext, etc.. which is just impl details. that simple refactoring could be completely different follow up PR.

server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
yili-db: `vendGcpToken`: Will change it. It was this name because the result is `gcpOauthToken`.

GCS vs GCP:
1. In `TemporaryCredentials`, they are called `awsTempCredentials`, `azureUserDelegationSas` and `gcpOauthToken`. That's defined in the API.
2. The credentials are usually generated by a different service than cloud storage. And credentials may be used for more purposes than accessing storage, unless specifically restricted.
3. Even if these credentials are only used for storage access, it is possible that one of the three clouds may provide a newer type of storage in the future. For example one day AWS comes up with `S4` that we have to support. There are already `ABFS` and `ABFSS`.

server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java
yili-db: Yes it should. In normal cases this `getCredentialDAO` should be empty for GCP so that it falls back to the server prop.

But if external location & storage credential are created for GCP, which shouldn't happen as it's not implemented yet, I don't want it to ignore them and fallback to server prop silently.

server/src/main/java/io/unitycatalog/server/service/credential/StorageCredentialVendor.java
yili-db: The major reason is **testability**. `UnityCatalogServer.Builder` has a `CloudCredentialVendor` that can be mocked for integration test. It's better if `CloudCredentialVendor` is kept the same way as today to just do cloud credential vending. This way we can easily mock it for integration test while still exercising the real `ExternalLocationUtils` logic.

yili-db: For example this is how credential is mocked in Spark integration tests:
https://github.com/unitycatalog/unitycatalog/blob/a03655cbb8960171e4f8a305b6a0683a4bbdfaa7/connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java#L100
```
    serverProperties.put("s3.bucketPath.0", "s3://test-bucket0");
    serverProperties.put("s3.accessKey.0", "accessKey0");
    serverProperties.put("s3.secretKey.0", "secretKey0");
    serverProperties.put("s3.sessionToken.0", "sessionToken0");
...
```
After they are deprecated, we will move to the other way:
```
      unityCatalogServer =
          UnityCatalogServer.builder()
              .port(port)
              .serverProperties(initServerProperties)
              .credentialOperations(cloudCredentialVendor)  <--- mock this before server starts
              .build();
```

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
yili-db: Split into two functions to reduce nest levels

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
yili-db: added comments here

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java
yili-db: added comments here

server/src/main/java/io/unitycatalog/server/service/credential/StorageCredentialVendor.java
yili-db: Also, today we only have credential for storage. **Just in case**, if we will ever have [credential for service](https://docs.databricks.com/aws/en/connect/unity-catalog/cloud-services/service-credentials), this `StorageCredentialVendor` can be just for storage.

tdas: thats a fair argument. alright, its stays as is.

server/src/main/java/io/unitycatalog/server/service/credential/CloudCredentialVendor.java
tdas: sounds good. fair argument about gcs vs gcp. but yeah in our code lets be consistent in using Credential instead of Token.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -199,9 +199,9 @@ public enum Property {
     MODEL_STORAGE_ROOT("storage-root.models", "file:///tmp/ucroot", STORAGE_PATH_VALIDATOR),
     TABLE_STORAGE_ROOT("storage-root.tables", "file:///tmp/ucroot", STORAGE_PATH_VALIDATOR),
     AWS_MASTER_ROLE_ARN("aws.masterRoleArn"),
-    AWS_S3_ACCESS_KEY("aws.s3.accessKey"),
-    AWS_S3_SECRET_KEY("aws.s3.secretKey"),
-    AWS_S3_SESSION_TOKEN("aws.s3.sessionToken"),
+    AWS_ACCESS_KEY("aws.accessKey"),
+    AWS_SECRET_KEY("aws.secretKey"),
+    AWS_SESSION_TOKEN("aws.sessionToken"),
```
tdas, line 204: these existed before and were used for something else? or they were added in the earlier work for building the credential support

server/src/main/java/io/unitycatalog/server/service/credential/aws/CredentialsGenerator.java
tdas, line 34: follow up PR.. lets make this AwsCredentialsGenerator. 
not obvious that this is for AWS only others. 
and matches other class AwsCredentialVendor

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTestWithMockCredentials.java
tdas: nice!

tdas: this is good context. please add this as inline comments on what test coverage this level of mocking is achieving.

tdas: ~also what is EchoAwsStsClient~ nvm. i see it below. i thought this is something that aws provided.

tdas: still do add an line comment here pointing to the docs on EchoStsClient.

server/src/test/java/io/unitycatalog/server/sdk/tempcredential/SdkTemporaryModelVersionCredentialTest.java
```diff
@@ -298,4 +298,21 @@ public void testGenerateTemporaryCredentialsWhereConfIsProvided(
           .isInstanceOf(ApiException.class);
     }
   }
+
+  @Test
+  public void testGenerateTemporaryCredentialsFromMasterRole() throws ApiException {
```
tdas: this is only for Aws right? then we should put it in the name.

tdas: please apply the same rename to all such tests.

server/src/test/java/io/unitycatalog/server/sdk/tempcredential/SdkTemporaryVolumeCredentialTest.java
```diff
@@ -46,13 +48,15 @@ protected SchemaOperations createSchemaOperations(ServerConfig serverConfig) {
...
 
+  @SneakyThrows
   @BeforeEach
   @Override
   public void setUp() {
     super.setUp();
     temporaryCredentialsApi = new TemporaryCredentialsApi(TestUtils.createApiClient(serverConfig));
     volumeOperations = createVolumeOperations(serverConfig);
     schemaOperations = createSchemaOperations(serverConfig);
+    createCatalogAndSchema();
```
tdas: shouldnt this be in the parent setup?

server/src/test/java/io/unitycatalog/server/sdk/tempcredential/TemporaryPathCredentialAccessControlTest.java
```diff
@@ -69,23 +76,37 @@ public class TemporaryPathCredentialAccessControlTest extends SdkAccessControlBa
...
-    serverProperties.put("s3.secretKey.0", "secretKey0");
-    serverProperties.put("s3.sessionToken.0", "sessionToken0");
-    serverProperties.put("s3.credentialsGenerator.0", TestAwsCredentialsGenerator.class.getName());
+    serverProperties.put(
+        ServerProperties.Property.AWS_MASTER_ROLE_ARN.getKey(), TEST_AWS_MASTER_ROLE_ARN);
+  }
+
+  // This is access control test. So it doesn't care about the actual credential vending. Just
+  // mock with something that returns credential.
+  private void setupMockCloudCrednetialVendor() {
```
tdas:
```suggestion
  private void setupMockCloudCredentialVendor() {
```

tdas:
looks mostly good.. just a few comments .. and general high level question on testing strategy. i am trying feel confident about the test coverage. can you summarize how all the lines changed and improvement of actually using external location + credential being used  have been tested by the test changes you did?

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
yili-db: This was used by `modifyS3Directory` which is used for deleting S3 directory recursively when a table or volume is deleted: 
https://github.com/unitycatalog/unitycatalog/blob/a03655cbb8960171e4f8a305b6a0683a4bbdfaa7/server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java#L139
It is like this since day 1. But it doesn't work even before this PR. It's not even using the per-bucket config.

Maybe I should disable that code path until it's re-implemented correctly?

server/src/main/java/io/unitycatalog/server/service/credential/aws/CredentialsGenerator.java
yili-db: Will do renaming in a follow up PR

yili-db:
> looks mostly good.. just a few comments .. and general high level question on testing strategy. i am trying feel confident about the test coverage. can you summarize how all the lines changed and improvement of actually using external location + credential being used have been tested by the test changes you did?

@tdas This PR changes the credential vending into two approaches:
1. using storage credential assigned to external location
2. Otherwise, using the legacy per-bucket config as fallback.

For 2, it's already tested in existing tests based on `BaseCRUDTestWithMockCredentials` and `BaseSparkIntegrationTest` etc (search for "s3.bucketPath.0" in code base).

For 1, it's covered by the 4 new test cases in `SdkTemporary*CredentialTest` that does the following:
1. In setup of `BaseCRUDTestWithMockCredentials`, create external location and storage credential
2. In the same setup, mock a `EchoAwsStsClient` to replace the real AWS STS client.
3. In each of the 4 `SdkTemporary*CredentialTest`, generate credential for an entity covered by the external location. This exercised all the logic down to the point of calling `stsClient.assumeRole()`.
4. This verifies that the credential is indeed produced by `EchoAwsStsClient` by:
  a. using the correct master role configured in `server.properties`
  b. assuming the correct role ARN in credential securable
  c. using the correct external ID

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
yili-db: On another thought, I just deleted this `modifyS3Directory` since it doesn't work anyway.
(commit d7e086d16f24a49ebd07a87954c12873a45d7ea5 in this PR)
The callers to it always catch all exceptions:
```
      try {
        fileOperations.deleteDirectory(volumeInfoDAO.getStorageLocation());
      } catch (Exception e) {
        LOGGER.error("Error deleting volume directory", e);
      }
```

tdas:
Thank you for the explanation. Lgtm but tests are failing.

yili-db:
> Thank you for the explanation. Lgtm but tests are failing.

@tdas Tests passed. Thank you!

PR #1315, opened 2026-01-23
Change OpenAPI yaml file to allow specifying managed location for catalog and schema

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1315/files) to review incremental changes.
- [**stack/managed_location_api**](https://github.com/unitycatalog/unitycatalog/pull/1315) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1315/files)]
  - [stack/managed_location_server](https://github.com/unitycatalog/unitycatalog/pull/1317) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1317/files/13cd2c8d3511ac4418dc3d51be47e8dab0ecaf3e..7fdb17ec8f793760e354a4465837b2f86a057458)]
    - [stack/managed_location_cli](https://github.com/unitycatalog/unitycatalog/pull/1318) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1318/files/7fdb17ec8f793760e354a4465837b2f86a057458..28ffff91a4fdf8a6e0e8e51b45e72b0bd915d834)]
      - [stack/managed_location_access](https://github.com/unitycatalog/unitycatalog/pull/1321) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1321/files/28ffff91a4fdf8a6e0e8e51b45e72b0bd915d834..5411012f9adeba619c37105ebd2f4a843b7aae64)]

---------
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**
Change OpenAPI yaml file to allow specifying managed location for catalog and schema

PR #1317, opened 2026-01-26
Add managed storage location support for catalogs and schemas

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1317/files) to review incremental changes.
- [**stack/managed_location_server**](https://github.com/unitycatalog/unitycatalog/pull/1317) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1317/files)]
  - [stack/managed_location_cli](https://github.com/unitycatalog/unitycatalog/pull/1318) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1318/files/d26b5e821fa0cc76ba8a9a822c22770b86c2f4fc..8f21b26e612923bd68fbacf63f63c9ce7c496566)]
    - [stack/managed_location_access](https://github.com/unitycatalog/unitycatalog/pull/1321) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1321/files/8f21b26e612923bd68fbacf63f63c9ce7c496566..3131aecc688aa9ac726ad7bb6d4984ba0c51b459)]

---------
**PR Checklist**

**Description of changes**
Add managed storage location support for catalogs and schemas

This change enables catalogs and schemas to have their own managed storage locations, allowing child entities (tables, volumes, models) to automatically inherit storage paths from their parent catalog or schema.

Key changes:

1. Storage Location Model:
   - Add `storageRoot` and `storageLocation` columns to `CatalogInfoDAO` and `SchemaInfoDAO`
   - `storageRoot`: User-provided storage root URL
   - `storageLocation`: Computed location with `__unitystorage` prefix (e.g., `storageRoot/__unitystorage/catalogs/{catalogId}`)

2. Reserved Namespace:
   - Introduce `__unitystorage` as a reserved path prefix for managed storage

3. Path Construction (FileOperations):
   - `createManagedCatalogDirectory`: Creates catalog storage under `storageRoot/__unitystorage/catalogs/{catalogId}`
   - `createManagedSchemaDirectory`: Creates schema storage under `storageRoot/__unitystorage/schemas/{schemaId}`
   - Child entities (tables/volumes/models) are placed under parent's storageLocation (e.g., `schemaStorageLocation/tables/{tableId}`)

4. Storage Root Resolution:
   - `ExternalLocationUtils.getManagedStorageRoot()` implements this resolution
   - It picks schema's storageLocation first, if set
   - Otherwise it pick catalog's storageLocation if set
   - If neither is set, fallback to server property `TABLE_STORAGE_ROOT` for backward compatibility

5. Validation:
   - Prevent external entities from using paths containing `__unitystorage`
   - Prevent external entities from overlapping with managed storage locations
   - Check both catalog/schema storage and uncommitted staging tables

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/base/managedlocation/BaseManagedLocationTest.java
+++ b/server/src/test/java/io/unitycatalog/server/base/managedlocation/BaseManagedLocationTest.java
@@ -45,7 +45,6 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
-// Split this into a base class, then one for catalog one for schema
 public abstract class BaseManagedLocationTest extends BaseCRUDTest {
 
   protected static final String MANAGED_VOLUME_NAME1 = "uc_managedvolume1";


ai/dev/start_unitycatalog.sh
```diff
@@ -69,7 +69,8 @@ cleanup_on_failure() {
 }
 
 echo "Starting UC server..."
-bin/start-uc-server >"$LOG_FILE" 2>&1 &
+# TODO: deprecate storage-root.models with managed storage of catalog or schema
+env "storage-root.models=file:///tmp/ucroot" bin/start-uc-server >"$LOG_FILE" 2>&1 &
```
yili-db, line 73: This was in `server.properties` as default value. It's no longer set as default value.
Later we should migrate the pre-built catalog or schema to have a managed storage.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/BaseCredRenewITTest.java
```diff
@@ -143,8 +143,6 @@ public void afterEach() {
 
     // Close the session.
     callQuietly(() -> session.close());
-
-    super.cleanUp();
```
yili-db, line 147: `super.cleanUp` no longer exist.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
```diff
@@ -156,6 +155,5 @@ public void cleanUp() {
     } catch (Exception e) {
       // Ignore
     }
-    super.cleanUp();
```
yili-db, line 159: `super.cleanUp` no longer exist.

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -287,13 +313,25 @@ static <T extends IdentifiableDAO> Query<T> generateEntitiesDAOsWithURLOverlapQu
...
     if (daoClassInfo == null) {
       throw new IllegalArgumentException(
           "Unsupported securable type for URL overlap check: " + securableType);
     }
+    return generateEntitiesDAOsWithURLOverlapQuery(
+        session, url, daoClassInfo, limit, includeParent, includeSelf, includeSubdir);
+  }
 
+  @VisibleForTesting
+  static <T extends IdentifiableDAO> Query<T> generateEntitiesDAOsWithURLOverlapQuery(
```
yili-db, line 391: This new spin-off function is here so that it can accept the `UNCOMMITTED_STAGING_TABLE_DAO_INFO` as a parameter which doesn't have a corresponding securable type.

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -323,7 +361,15 @@ static <T extends IdentifiableDAO> Query<T> generateEntitiesDAOsWithURLOverlapQu
       condition = inCondition;
     } else if (hasLikeCondition) {
       condition = likeCondition;
+    } else {
+      // This is not supposed to happen unless there's a bug.
```
yili-db, line 430: This check is just redundant safeguard.

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -323,7 +361,15 @@ static <T extends IdentifiableDAO> Query<T> generateEntitiesDAOsWithURLOverlapQu
...
+      throw new RuntimeException("At least one condition must be provided");
     }
+
+    if (!daoClassInfo.filterCondition.isEmpty()) {
+      condition = String.format("%s AND (%s)", daoClassInfo.filterCondition, condition);
```
yili-db, line 435: Only staging table use this to say `uncommited only`

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -380,4 +426,150 @@ static String escapeLikePattern(String value) {
...
+    CatalogInfoDAO catalogInfoDAO = catalogAndSchemaDao.catalogInfoDAO();
+    SchemaInfoDAO schemaInfoDAO = catalogAndSchemaDao.schemaInfoDAO();
+    if (schemaInfoDAO.getStorageLocation() != null) {
+      return NormalizedURL.from(schemaInfoDAO.getStorageLocation());
+    } else if (catalogInfoDAO.getStorageLocation() != null) {
+      return NormalizedURL.from(catalogInfoDAO.getStorageLocation());
+    } else {
+      return fallbackStorageRoot
+          .get()
+          .map(fallback -> NormalizedURL.from(fallback + "/" + Constants.MANAGED_STORAGE_PREFIX))
```
yili-db, line 632: `MANAGED_STORAGE_PREFIX` is already part of `getStorageLocation()` but not `fallbackStorageRoot`

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -29,12 +29,12 @@ private enum Operation {
 
   private UriUtils() {}
 
-  public static String createStorageLocationPath(String uri) {
-    return updateDirectoryFromUri(uri, Operation.CREATE, Optional.empty()).toString();
+  public static void createStorageLocationPath(String uri) {
```
yili-db: return values are never used. also the return value was always the same as input `uri`

server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
```diff
@@ -53,6 +56,15 @@ public CatalogInfo addCatalog(CreateCatalog createCatalog) {
             .updatedAt(createTime)
             .updatedBy(callerId)
             .properties(createCatalog.getProperties());
+    NormalizedURL storageRoot = NormalizedURL.from(createCatalog.getStorageRoot());
+    if (storageRoot != null) {
+      // storageRoot, if set, is already authorized as a valid external location by CatalogService.
+      ExternalLocationUtils.validateNotSameOrUnderManagedStorage(storageRoot);
```
yili-db: Does not need to check `validateNotAboveManagedStorage` because `storageRoot` can cover other catalogs. It's the `storageLocation` that will be unique.

server/src/main/java/io/unitycatalog/server/persist/SchemaRepository.java
```diff
@@ -68,6 +71,18 @@ public SchemaInfo createSchema(CreateSchema createSchema) {
                   .updatedAt(createTime)
                   .updatedBy(callerId)
                   .properties(createSchema.getProperties());
+          NormalizedURL storageRoot = NormalizedURL.from(createSchema.getStorageRoot());
+          if (storageRoot != null) {
+            // storageRoot, if set, is already authorized as a valid external location by
+            // SchemaService.
+            ExternalLocationUtils.validateNotSameOrUnderManagedStorage(storageRoot);
```
yili-db: Does not need to check `validateNotAboveManagedStorage` because `storageRoot` can cover other catalogs. It's the `storageLocation` that will be unique.

server/src/main/java/io/unitycatalog/server/persist/ExternalLocationRepository.java
```diff
@@ -74,6 +74,7 @@ public ExternalLocationInfo addExternalLocation(CreateExternalLocation createExt
           }
 
           NormalizedURL url = NormalizedURL.from(createExternalLocation.getUrl());
+          ExternalLocationUtils.validateNotSameOrUnderManagedStorage(url);
```
yili-db: Does not need to check `validateNotAboveManagedStorage` because as long as an existing external location doesn't exist, it should be allowed to create even if it covers existing managed storage.
This is useful so that accidental or intentional deletion of external location where catalogs&schemas already live can be recovered by creating it again.

The same story for creating external locations where external tables volumes already live.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -562,8 +565,7 @@ public ModelVersionInfo createModelVersion(CreateModelVersion createModelVersion
         UUID modelId = existingRegisteredModel.getId();
         Long version = existingRegisteredModel.getMaxVersionNumber() + 1;
         storageLocation =
-            fileOperations.getModelVersionStorageLocation(
-                catalogId.toString(), schemaId.toString(), modelId.toString(), modelVersionId);
+            NormalizedURL.from(existingRegisteredModel.getUrl() + "/versions/" + modelVersionId);
```
yili-db: `catalogId` or `schemaId`, plus `modelId`, are already in `existingRegisteredModel`.

server/src/test/java/io/unitycatalog/server/base/BaseCRUDTest.java
```diff
@@ -1,31 +1,12 @@
...
-import java.util.Optional;
-import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 
 public abstract class BaseCRUDTest extends BaseServerTest {
 
   protected CatalogOperations catalogOperations;
 
-  @AfterEach
-  public void cleanUp() {
```
yili-db, line 16: This is redundant. `BaseServerTest` cleans the database.

server/src/test/java/io/unitycatalog/server/sdk/managedlocation/SdkManagedLocationTest.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @Override
+  protected ModelOperations createModelOperations(ServerConfig serverConfig) {
+    return new SdkModelOperations(TestUtils.createApiClient(serverConfig));
+  }
+
+  @SneakyThrows
+  @Test
+  public void testCreateVolumeOverOrphanStagingTable() {
```
yili-db, line 69: This is an **extra** test that can only be done with API.

etc/conf/server.properties
```diff
@@ -20,23 +20,6 @@ server.cookie-timeout=P5D
 # Default: false (disabled)
 server.managed-table.enabled=false
 
-# Define the model storage root.  Cloud storage or file based allowed.
```
tdas, line 23: why are you removing these? whats the replacement configuration that makes it all work?

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -69,6 +92,9 @@ SecurableType.REGISTERED_MODEL, new DaoClassInfo(RegisteredModelInfoDAO.class, "
       Stream.concat(Stream.of(SecurableType.EXTERNAL_LOCATION), DATA_SECURABLE_TYPES.stream())
           .toList();
 
+  private static final List<SecurableType> CATALOG_AND_SCHEMA_SECURABLE_TYPES =
```
tdas: nit: cant we call this managed securable type containers? thats what differentiates catalog and schema from other securable types.. isnt it?

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -380,4 +426,150 @@ static String escapeLikePattern(String value) {
...
+   *
+   * <p>This check ensures that user-specified paths (such as external location URLs or storage
+   * roots for catalogs/schemas) do not use or interfere with the reserved {@code __unitystorage}
+   * namespace used for managed entities.
+   *
+   * @param url the URL to validate
+   * @throws BaseException with ErrorCode.INVALID_ARGUMENT if the URL contains the managed storage
+   *     prefix
+   */
+  public static void validateNotSameOrUnderManagedStorage(NormalizedURL url) {
```
tdas: validateNotSameOrUnderManagedStoragePrefix

etc/conf/server.properties
yili-db: They are deprecated and should be replaced by `storage_root` of catalog or schema which is path under `external location`.

ai/dev/start_unitycatalog.sh
yili-db: @BenWilson2 @serena-ruan Please take a look at this PR.
This PR introduces the storage location support so that models will inherit storage location from parent schema or catalog. It still allows the server property `storage-root.models` as a **fallback** if no storage root is defined in schema and catalog. But `storage-root.models` is no longer set in the default `server.properties` file.
Here for these AI tests I just put this workaround to keep them going the **same way as before**. Eventually I will migrate them to also use managed storage location specified by schema or catalog.

server/src/main/java/io/unitycatalog/server/persist/CatalogRepository.java
```diff
@@ -53,6 +56,14 @@ public CatalogInfo addCatalog(CreateCatalog createCatalog) {
             .updatedAt(createTime)
             .updatedBy(callerId)
             .properties(createCatalog.getProperties());
+    NormalizedURL storageRoot = NormalizedURL.from(createCatalog.getStorageRoot());
+    if (storageRoot != null) {
+      ExternalLocationUtils.validateNotSameOrUnderManagedStoragePrefix(storageRoot);
+      NormalizedURL storageLocation =
+          repositories.getFileOperations().createManagedCatalogDirectory(storageRoot, catalogId);
```
tdas:
```suggestion
          repositories.getFileOperations().createManagedLocationForCatalog(storageRoot, catalogId);
```
similarly for `...ForSchema`. `ManagedCatalog` is confusing.. and `Directory` is a different term from `Location`

tdas: also update naming style for other stuff like models, etc.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -188,16 +137,33 @@ private URI modifyS3Directory(URI parsedUri, boolean createOrDelete) {
...
-  private String getManagedTablesStorageRoot() {
-    // Use local tmp directory as default storage root
-    return serverProperties.get(Property.TABLE_STORAGE_ROOT);
+  private NormalizedURL createManagedEntityDirectory(
+      NormalizedURL storageRoot, String prefix, UUID entityId) {
+    return NormalizedURL.from(
+        String.join("/", List.of(storageRoot.toString(), prefix, entityId.toString())));
+  }
+
+  public NormalizedURL createManagedSchemaDirectory(NormalizedURL storageRoot, UUID schemaId) {
```
tdas: update all names based on my comment https://github.com/unitycatalog/unitycatalog/pull/1317/changes#r2732575495

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
tdas: can you add short docs? totally not obvious from reading the code what is it creating/updating in?

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -188,16 +137,33 @@ private URI modifyS3Directory(URI parsedUri, boolean createOrDelete) {
...
+    return createManagedEntityDirectory(storageRoot, MANAGED_STORAGE_SCHEMA_PREFIX, schemaId);
+  }
+
+  public NormalizedURL createManagedCatalogDirectory(NormalizedURL storageRoot, UUID catalogId) {
+    return createManagedEntityDirectory(storageRoot, MANAGED_STORAGE_CATALOG_PREFIX, catalogId);
+  }
+
+  // The following methods do not add a __unitystorage prefix because the storageRoot
+  // is expected to be the storageLocation of a catalog or schema, which already includes
+  // the __unitystorage prefix.
```
tdas: do you mean that this is the storage_location defined in the spec 
https://github.com/unitycatalog/unitycatalog/pull/1315/changes#diff-8d03742f9785228675d14f784a7c0affcc9400e8571603bde6ede194c052d985R1867

if so, then lets name it clearly as storageRoot is confusing. how about "schemaStorageLocation` as these have to be under a schema .. isnt it? also.. lets make it unambiguous in the docs with examples.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -215,11 +219,12 @@ public RegisteredModelInfo createRegisteredModel(CreateRegisteredModel createReg
...
           RepositoryUtils.getCatalogAndSchemaDaoOrThrow(session, catalogName, schemaName);
-      CatalogInfoDAO catalogInfoDAO = catalogAndSchemaDao.catalogInfoDAO();
-      SchemaInfoDAO schemaInfoDAO = catalogAndSchemaDao.schemaInfoDAO();
+      NormalizedURL storageRoot =
+          ExternalLocationUtils.getManagedStorageRoot(
+              catalogAndSchemaDao, this::getDefaultModelsStorageRoot);
       NormalizedURL storageLocation =
-          fileOperations.getModelStorageLocation(
-              catalogInfoDAO.getId().toString(), schemaInfoDAO.getId().toString(), modelId);
+          fileOperations.createManagedModelDirectory(storageRoot, modelId);
```
tdas: why cant we collapse these 2 lines in a single call that take the `catalogAndSchemaDao` and returns managed securable location. then it avoids the storageRoot ambiguity that i highlighted here https://github.com/unitycatalog/unitycatalog/pull/1317/changes#r2732998289

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
```diff
@@ -158,4 +163,17 @@ public StagingTableDAO commitStagingTable(
...
+
+  /**
+   * Gets the default managed table storage root from server properties.
+   *
+   * <p>This is used as a fallback when neither the catalog nor schema has a managed storage
+   * location configured. It's being deprecated.
+   *
+   * @return optional containing the storage root URL if configured in server properties
+   */
+  private Optional<NormalizedURL> getDefaultManagedTablesStorageRoot() {
```
tdas, line 174: why is this only in stageTableRepo? other securables also use this property root .. isnt it?

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -196,8 +196,8 @@ public enum Property {
     REDIRECT_PORT("server.redirect-port", POSITIVE_INTEGER_VALIDATOR),
     COOKIE_TIMEOUT("server.cookie-timeout", "P5D", DURATION_VALIDATOR),
     MANAGED_TABLE_ENABLED("server.managed-table.enabled", "false", BOOLEAN_VALIDATOR),
-    MODEL_STORAGE_ROOT("storage-root.models", "file:///tmp/ucroot", STORAGE_PATH_VALIDATOR),
-    TABLE_STORAGE_ROOT("storage-root.tables", "file:///tmp/ucroot", STORAGE_PATH_VALIDATOR),
+    MODEL_STORAGE_ROOT("storage-root.models", STORAGE_PATH_VALIDATOR), // Deprecated
```
tdas, line 200: add that its replaced by external location

server/src/test/java/io/unitycatalog/server/persist/utils/ExternalLocationUtilsTest.java
```diff
@@ -238,7 +238,7 @@ public void testUnsupportedSecurableType() {
                 ExternalLocationUtils.getEntityDAOsWithURLOverlap(
                     null,
                     NormalizedURL.from("s3://bucket/path"),
-                    SecurableType.CATALOG,
+                    SecurableType.CREDENTIAL,
```
tdas, line 242: whats this change?  changing the securable type seems a pretty fundamental change in the test?

server/src/test/java/io/unitycatalog/server/persist/utils/ExternalLocationUtilsTest.java
```diff
@@ -251,7 +251,7 @@ public void testUnsupportedSecurableType() {
                 ExternalLocationUtils.getEntityDAOsWithURLOverlap(
                     null,
                     NormalizedURL.from("s3://bucket/path"),
-                    SecurableType.SCHEMA,
+                    SecurableType.FUNCTION,
```
tdas, line 255: same question as above.

server/src/test/java/io/unitycatalog/server/persist/utils/ExternalLocationUtilsTest.java
```diff
@@ -478,5 +478,24 @@ public void testGenerateEntitiesDAOsWithURLOverlapQueryWithRealSession() {
             + "ESCAPE '\\' ORDER BY LENGTH(url) DESC",
         List.of("file:///tmp", "file:///", "file:///tmp/path"),
         "file:///tmp/path/%");
+
+    // Test 11: look for uncommitted staging tables
```
tdas, line 483: what is this testing.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: Renamed to `parentStorageLocation`. It can be from schema or catalog.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
yili-db: Having the middle result of `catalogAndSchemaDao` here is for the purpose of saving one extra DB query because at the minimum these new tables/volumes/models needs a schema ID.
To retrieve the schema ID by `catalog_name.schema_name`, it's essentially the same process of `getCatalogAndSchemaDaoOrThrow` again.

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
yili-db: This is only for tables. Models have `storage-root.models`.
There's no equivalent for volume since volume did not have managed volume support until now.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
yili-db: Added: `// storage-root.* are replaced by managed storage locations of catalog and schema.`

server/src/test/java/io/unitycatalog/server/persist/utils/ExternalLocationUtilsTest.java
yili-db: Previously catalog and schema are not supported as a securable type to search for overlapping URL.
This PR extended implementation of `getEntityDAOsWithURLOverlap` so that now they are supported. The tests are adjusted to try something else that aren't supported.

server/src/test/java/io/unitycatalog/server/persist/utils/ExternalLocationUtilsTest.java
yili-db: ditto.

server/src/test/java/io/unitycatalog/server/persist/utils/ExternalLocationUtilsTest.java
yili-db: This is to cover the newly introduced ability to search for uncommitted staging tables with overlapping URL to prevent an external table or volume to be created.
Because staging tables, when not committed, are not deleted even after the catalog and schema are deleted. So to prevent leaking of abandoned staging table data, a new check is introduced.

In code it's commented as:
```
    // volumes, committed tables, and models are already checked by KeyMapper. Uncommitted staging
    // tables aren't yet.
    // In normal cases the staging tables are under catalogs or schemas so the above check already
    // makes sure the input url doesn't cover any staging tables or tables. However, because the
    // staging tables are not deleted synchronously when the catalog/schema is deleted, there can be
    // orphaned uncommited staging tables remaining after the catalog/schema is gone. This check
    // covers the corner cases.
```

ai/dev/start_unitycatalog.sh
serena-ruan: I see, that makes sense! Thanks for the update!

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
```diff
@@ -29,20 +28,14 @@ private enum Operation {
 
   private UriUtils() {}
```
tdas, line 29: i am wondering what is the purpose of this class. its actually two sets of things
1. path operations, creating and deleting dirs - thats way more related to FileOperations and should be just in FileOperations.. isnt it?
2. URI validation - this should be in NormalizedURL isnt it?

if you agree.. then 
1. move these new public functions to FileOperations so that all caller dont need to be updated.
2. a follow up PR can be to migrate rest of the details of this unnecessary class to relevant classes

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
tdas: but shouldnt this logic be consolidated in the same location as other path constructions? basically FileOperations?

server/src/test/java/io/unitycatalog/server/utils/FileOperationsTest.java
```diff
@@ -1,60 +1,55 @@
...
+        fileOperations.createManagedLocationForModel(parentStorageLocation, modelId);
+    assertThat(modelPathUri.toString()).isEqualTo("file:///tmp/storage/models/" + modelId);
+
+    UriUtils.createStorageLocationDir(modelPathUri);
+    UriUtils.deleteStorageLocationDir(modelPathUri);
+
+    // cleanup the created storage directory
+    UriUtils.deleteStorageLocationDir(NormalizedURL.from("file:/tmp/storage"));
+
+    // Test table directory creation
```
tdas: the above ones are testing uriutils as the test name says.. but the lower ones are not testing uriutils.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -188,16 +137,36 @@ private URI modifyS3Directory(URI parsedUri, boolean createOrDelete) {
     }
   }
 
-  private String getManagedTablesStorageRoot() {
-    // Use local tmp directory as default storage root
-    return serverProperties.get(Property.TABLE_STORAGE_ROOT);
+  private NormalizedURL createManagedEntityDirectory(
```
tdas: they are not actually creating the directory .. is it? if not .. then all of these methods should get or generate.

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
yili-db: Yes I agree that this class is redundant and should be removed or merged into `FileOperations`. That's why I **removed** 3 now unused functions from this class in this PR.

The diff of these two "new" functions are about:
1. Adding comment
2. Removing unused String return value

I'd like to eventually eliminate this class but in a follow up PR because merging this class with `FileOperations` would conflict with https://github.com/unitycatalog/unitycatalog/pull/1313

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: Renamed to `getManagedLocationFor*`

server/src/test/java/io/unitycatalog/server/utils/FileOperationsTest.java
yili-db: Split into two testing methods

server/src/main/java/io/unitycatalog/server/persist/utils/UriUtils.java
tdas: your plan sounds good.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -188,16 +137,43 @@ private URI modifyS3Directory(URI parsedUri, boolean createOrDelete) {
     }
   }
 
-  private String getManagedTablesStorageRoot() {
-    // Use local tmp directory as default storage root
-    return serverProperties.get(Property.TABLE_STORAGE_ROOT);
+  private static NormalizedURL getManagedLocationForEntity(
```
tdas: hey .. i am wondering .. `getManagedStorageLocation` is in `ExternalLocationUtils` but `getManagedLocationForEntity` is in `FileOperations`. that does not make sense to me. both these sets of `ManagedLocation` related function should be in the same class. 

Now fileOperations seems to be a class about actually do any file or file system operations. so these methods should not be in FileOperations. 

ExternalLocationUtils seems a better place (since managed locations can only be on external locations). Alternatively, if you think it make sense to create a ManagedLocationUtils... i am good with that too.

server/src/test/java/io/unitycatalog/server/utils/FileOperationsTest.java
```diff
@@ -1,60 +1,57 @@
...
-
-    assertThatThrownBy(() -> UriUtils.createStorageLocationPath(".."))
-        .isInstanceOf(BaseException.class);
+    // cleanup the created storage directory
+    UriUtils.deleteStorageLocationDir(parentStorageLocation);
+  }
 
-    assertThatThrownBy(() -> UriUtils.deleteStorageLocationPath(""))
-        .isInstanceOf(BaseException.class);
+  @Test
```
tdas: update this according to whatever you decide here - https://github.com/unitycatalog/unitycatalog/pull/1317/changes#r2742871684

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
```diff
@@ -380,4 +427,150 @@ static String escapeLikePattern(String value) {
...
+   * </ol>
+   *
+   * <p>This validation should be called when creating external entities (tables, volumes) to ensure
+   * they don't interfere with managed storage.
+   *
+   * @param session the Hibernate session for database queries
+   * @param url the URL to validate
+   * @throws BaseException with ErrorCode.INVALID_ARGUMENT if the URL overlaps with managed storage
+   */
+  public static void validateNotOverlapWithManagedStorage(Session session, NormalizedURL url) {
```
tdas, line 579: is ManagedStorage and ManagedLocation the same thing? if so, we should make it the same name.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: Yes moving to `ExternalLocationUtils` makes sense. I'll keep the name `ExternalLocationUtils` since it doesn't just cover managed location but it also works for external tables etc.

server/src/test/java/io/unitycatalog/server/utils/FileOperationsTest.java
yili-db: Moving to `ExternalLocationUtilsTest`

server/src/main/java/io/unitycatalog/server/persist/utils/ExternalLocationUtils.java
yili-db: @tdas 
For reference: https://docs.databricks.com/aws/en/connect/unity-catalog/cloud-storage/managed-storage
Yes they are about the same thing, but focusing on different aspects. Both terms are used in this doc.

- Managed storage is the concept that UC manages all storage under it.
- Managed location is short for managed storage location: the URL where the managed storage is.

So for some of these functions like `getManagedLocationForCatalog`, it's clear that managed location makes more sense since all it does is just generate an URL.
For other functions like `validateNotOverlapWithManagedStorage` , managed storage makes more sense since it's referring to all the files stored and managed under the managed storage.
And the permission has to be called `CREATE MANAGED STORAGE`.
But there are other functions that may be hard to tell if their naming is accurate or not.

At least, it seems that we can not do a blanket renaming of one to another?

tdas: i see. lets leave it as is then.

server/src/test/java/io/unitycatalog/server/base/managedlocation/BaseManagedLocationTest.java
```diff
@@ -0,0 +1,389 @@
...
+import java.util.stream.Stream;
+import lombok.SneakyThrows;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+// Split this into a base class, then one for catalog one for schema
```
tdas: what does this mean? 
is it a TODO? it definitely not a good class docs.

yili-db: It is an old TODO that I forgot to remove. It no longer makes sense. I just removed it.

server/src/test/java/io/unitycatalog/server/base/managedlocation/BaseManagedLocationTest.java
```diff
@@ -0,0 +1,388 @@
...
+import java.util.stream.Stream;
+import lombok.SneakyThrows;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public abstract class BaseManagedLocationTest extends BaseCRUDTest {
```
tdas, line 48: This test organization is making me wonder.. if there should be a single testsuite for managed locations across all current and future securables OR should each securable have a managed location test added to its own testsuite. I feel like it should be latter... because then each securable follows that common model and the whole thing scales.

yili-db: The reason I want to have all of managed tables, volumes and models in this same test is because they all share the same test setup: catalog with storage location, or schema with storage location, or both. This setup is distinct from normal `*CrudTest` setup of each securables.

I can split them into 3 different base classes based on this one: `BaseTableManagedLocationTest`, `BaseVolumeManagedLocationTest` and `BaseModelManagedLocationTest`. Under the hood they still exercise the same tests as this combined base class so the change is cosmetic. **However**, with these 3 base classes, they each will fanout into 2 actual tests (`Sdk*` and `Cli*`) making it a total of **6** boilerplate test classes.

tdas: yeah.. i see the challenge. this is good for now.

Issue #1101, opened 2025-10-07
Suggest to make unitycatalog to be a pure Java repository


openinx:
Currently,  the unitycatalog is mixed by Java and Scala code, while 90% of the code is Java code.  I can only see 3 files are Scala code. 

```bash
find . -type f -name '*.scala' 
./connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
./project/Tarball.scala
./project/PythonClientPostBuild.scala
./project/ReleaseSettings.scala
```

Personally,  I will suggest to turn the unitycatalog to be a pure java repository, which will make us easier to manage the whole dev loop.  For example, all of the following dev loop can focus only on Java

* Codestyle enforcement
* Github CI testing
* Test framework and abstraction. 

Also if we finally turn to pure java repo,  then we also don't need to be distract for different Scala compatible issues.

tdas:
I agree with this! because we started with Scala for unitycatalog-spark, we had to cross publish with multiple scala version for Spark 3.5. With Spark 4.0 its only scala 2.13, but who know what future spark versions will depend on. so its better to make it java. 

however, it will be a maven published name change. `unitycatalog-spark_2.13` --> `unitycatalog-spark`. so we have to update our docs and everything. lets target next 0.4 release.

PR #1320, opened 2026-01-26
[Spark] Remove the scala dependency from unitycatalog-spark connector.

openinx:
**PR Checklist**

**Description of changes**

I created an issue https://github.com/unitycatalog/unitycatalog/issues/1101 for making the unitycatalog project to be a pure java project, and currently the only scala part is unitycatalog-spark connector. 

Let's just translate the whole `UCSingleCatalog` class into a pure java one, and then we can confidently to remove the scale dependency completely from our unitycatalog project. 

And with this, we can publish it in unitycatalog 0.4.0 !

**NOTICE**

After this PR, it will rename the artifacts name from `unitycatalog-spark_2.13-0.3.0-SNAPSHOT.jar` to `unitycatalog-spark-0.3.0-SNAPSHOT.jar`

### DIFF
--- a/connectors/spark/src/main/java/io/unitycatalog/spark/UCProxy.java
+++ b/connectors/spark/src/main/java/io/unitycatalog/spark/UCProxy.java
@@ -70,12 +70,11 @@ public class UCProxy implements TableCatalog, SupportsNamespaces {
   private final URI uri;
   private final TokenProvider tokenProvider;
   private final boolean renewCredEnabled;
-  private final SchemasApi schemaApi;
+  private final SchemasApi schemasApi;
   private final TablesApi tablesApi;
   private final TemporaryCredentialsApi temporaryCredentialsApi;
 
   private String name = null;
-  private SchemasApi schemasApi = null;
 
   UCProxy(
       URI uri,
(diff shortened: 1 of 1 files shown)


connectors/spark/src/main/java/io/unitycatalog/spark/UCProxy.java
```diff
@@ -0,0 +1,487 @@
...
+import org.apache.spark.sql.types.StringType;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+import org.apache.spark.sql.types.TimestampNTZType;
+import org.apache.spark.sql.types.TimestampType;
+import org.apache.spark.sql.util.CaseInsensitiveStringMap;
+import scala.Option;
+import scala.jdk.javaapi.CollectionConverters;
+
+public class UCProxy implements TableCatalog, SupportsNamespaces {
```
openinx, line 66: This is the java class which was translated from [UCSingleCatalog#UCProxy](https://github.com/unitycatalog/unitycatalog/pull/1320/changes#diff-b496bfc650c9a1c585fb5709e010388ca74382670b419be5ebf4faa1bc499dd4L278),  making it a separate java class will be much clear.

PR #1321, opened 2026-01-26
Require CREATE_MANAGED_STORAGE permission on external location when creating catalog&schema with storage_root

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1321/files) to review incremental changes.
- [**stack/managed_location_access**](https://github.com/unitycatalog/unitycatalog/pull/1321) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1321/files)]

---------
**PR Checklist**

**Description of changes**
Require CREATE_MANAGED_STORAGE permission on external location when creating catalog&schema with storage_root

### DIFF
--- a/examples/cli/src/test/java/io/unitycatalog/cli/access/CliAccessControlBaseCrudTest.java
+++ b/examples/cli/src/test/java/io/unitycatalog/cli/access/CliAccessControlBaseCrudTest.java
@@ -155,22 +155,24 @@ public class CliAccessControlBaseCrudTest extends BaseAccessControlCRUDTest {
               "--credential_name",
               "admin_cred"));
 
-  protected final List<Step> grantCreateManagedStoragePermissionSteps =
-      List.of(
-          Step.TokenStep.of(SUCCEED, "admin"),
-          Step.CommandStep.of(
-              SUCCEED,
-              1,
-              "permission",
-              "create",
-              "--securable_type",
-              "external_location",
-              "--name",
-              "admin_el",
-              "--principal",
-              "principal-1@localhost",
-              "--privilege",
-              "CREATE MANAGED STORAGE"));
+  protected final List<Step> grantExternalLocationPermissionSteps(
+      String principal, String permission) {
+    return List.of(
+        Step.TokenStep.of(SUCCEED, "admin"),
+        Step.CommandStep.of(
+            SUCCEED,
+            1,
+            "permission",
+            "create",
+            "--securable_type",
+            "external_location",
+            "--name",
+            "admin_el",
+            "--principal",
+            principal,
+            "--privilege",
+            permission));
+  }
 
   public void testSteps(List<Step> steps)
       throws IOException, NoSuchAlgorithmException, InvalidKeySpecException {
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/unitycatalog/server/service/SchemaService.java
```diff
@@ -45,14 +47,36 @@ public SchemaService(UnityCatalogAuthorizer authorizer, Repositories repositorie
...
+   * location (rather than tables etc.) owns the path.
+   */
   @Post("")
   @AuthorizeExpression("""
-      #authorize(#principal, #catalog, OWNER) ||
-      #authorizeAll(#principal, #catalog, USE_CATALOG, CREATE_SCHEMA)
+      (#authorize(#principal, #catalog, OWNER) ||
+       #authorizeAll(#principal, #catalog, USE_CATALOG, CREATE_SCHEMA)) &&
+      (#storage_root == null ||
+       (#table == null && #volume == null && #registered_model == null &&
```
tdas: i think this needs the no_overlap thing.

tdas (changes requested):
left comment.

server/src/main/java/io/unitycatalog/server/service/CatalogService.java
```diff
@@ -41,10 +43,34 @@ public CatalogService(UnityCatalogAuthorizer authorizer, Repositories repositori
...
+   * location (rather than tables etc.) owns the path.
+   */
   @Post("")
-  @AuthorizeExpression("#authorizeAny(#principal, #metastore, OWNER, CREATE_CATALOG)")
+  @AuthorizeExpression("""
+      #authorizeAny(#principal, #metastore, OWNER, CREATE_CATALOG) &&
+      (#storage_root == null ||
+       (#no_overlap_with_data_securable &&
+        #external_location != null &&
+        #authorizeAny(#principal, #external_location, OWNER, CREATE_MANAGED_STORAGE)))
```
tdas, line 67: can you show which are the tests that give all the conditional branch coverage?

yili-db: - `#storage_root == null`: This is creating catalog/schema without location so it is already covered by existing tests
- `#no_overlap_with_data_securable`: This is covered by `// Create a table, then a schema under the table. It should fail.`
- `#external_location != null`: catalog/schema creation with location must have matching external location. This is covered by `// Try to create a schema at the location before External Location is created would fail`
- `#authorizeAny(#principal, #external_location`: This is covered by `// Try to create a schema at the location and still fail due to lack of permission`
The same tests repeat for catalog as well.

tdas: but these are only Cli tests.. why are these conditions covered by core server / sdk tests?

yili-db: Unlike ordinary CRUD tests which can run the same tests cases with both SDK and CLI, access control tests can either be SDK or CLI but not both. Unless all these tests are replicated in the other way.
Most access control tests are CLI only. A few of them are SDK only if CLI doesn't have the operations, for example DeltaCommit.
In this case, there is no existing SDK access control tests for catalog and schema. But they are only in CLI. So I just added the new test cases here.

tdas: okay. .not going to block this PR.. but why cant ACL tests be both SDK and CLI? 
if we have to pick one.. then its better to test SDK tests as those are in the server module. code in server module must be tested by test code in server module. 

in a follow up PR, could you address this? hopefully claude makes it fairly easy to make theh change.

PR #1323, opened 2026-01-27
[Spark] Separate the java files into java folder so that we can reformat all Java classes.

openinx:
**PR Checklist**

**Description of changes**

In this PR,  I moved all the java files under the `connectors/spark/src/scala/io/unitycatalog/spark` to the `connectors/spark/src/java/io/unitycatalog/spark`.  And also enabled the automatic `javafmtAll`, so that we can automatically reformat all the unitycatalog spark connector's java files. 

After this PR, it will be much more easier to get this PR in https://github.com/unitycatalog/unitycatalog/pull/1320/changes,  because we can focus the `UCSingleCatalog` changes in that PR, without distraction from other Java files changes.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -313,7 +313,7 @@ lazy val server = (project in file("server"))
     mainClass := Some(orgName + ".server.UnityCatalogServer"),
     commonSettings,
     javaOnlyReleaseSettings,
-    javafmtCheckSettings,
+    javafmtCheckSettings(),
     javaCheckstyleSettings("dev/checkstyle-config.xml"),
     Compile / compile / javacOptions ++= Seq(
       "-processor",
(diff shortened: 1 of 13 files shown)


build.sbt
```diff
@@ -313,7 +313,7 @@ lazy val server = (project in file("server"))
     mainClass := Some(orgName + ".server.UnityCatalogServer"),
     commonSettings,
     javaOnlyReleaseSettings,
-    javafmtCheckSettings,
+    javafmtCheckSettings(),
```
openinx, line 316: Nit: the intellij idea is more encouraged to use this `javafmtCheckSettings()`, rather than `javafmtCheckSettings`.  because it's more easier to detect as a method rather than a variable.

tdas: okay. no issues at all.

Issue #891, opened 2025-02-09
io.unitycatalog.client.ApiException: DataType not supported: date


leminkhoa:
**Describe the bug**

I recently tried to create an external table using Spark Connector:
```
# Create tables under this raw schema
create_customers = f"""
CREATE TABLE IF NOT EXISTS {CATALOG}.{RAW_SCHEMA}.customers (
    customer_id STRING,
    gender STRING,
    first_name STRING,
    last_name STRING,
    email STRING,
    yob INTEGER,
    phone_number STRING,
    job STRING,
    address STRING,
    first_transaction DATE,
    membership STRING,
    last_processed_ts TIMESTAMP
)
USING DELTA
LOCATION '{STORAGE_FOLDER}/{CATALOG}/{RAW_SCHEMA}/customers'
COMMENT 'This table stores customer personal information'
TBLPROPERTIES ('domain' = 'customer')
"""

# Execute the SQL query
spark.sql(create_customers)
```

This raised an issue:
```
Py4JJavaError: An error occurred while calling o37.sql.
: io.unitycatalog.client.ApiException: DataType not supported: date
	at io.unitycatalog.spark.UCProxy.convertDataTypeToTypeName(UCSingleCatalog.scala:365)
	at io.unitycatalog.spark.UCProxy.$anonfun$createTable$5(UCSingleCatalog.scala:327)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
	at scala.collection.TraversableLike.map(TraversableLike.scala:286)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:279)
	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
	at io.unitycatalog.spark.UCProxy.createTable(UCSingleCatalog.scala:319)
	at org.apache.spark.sql.connector.catalog.TableCatalog.createTable(TableCatalog.java:223)
	at io.unitycatalog.spark.UCProxy.createTable(UCSingleCatalog.scala:205)
	at org.apache.spark.sql.connector.catalog.DelegatingCatalogExtension.createTable(DelegatingCatalogExtension.java:111)
[...]

Issue #801, opened 2024-12-13
Spark-SQL Read data failed when the table field using ARRAY or STRUCT.


horizonzy:
Hello, I have a question about the unity catalog.
When I use the spark-sql to try read data from the unity catalog table, it shows the error. (The table is external)
[INCOMPLETE_TYPE_DEFINITION.STRUCT] Incomplete complex type: The definition of "STRUCT" type is incomplete. You must provide at least one field type. For example: "STRUCT<name STRING, phone DECIMAL(10, 0)>".(line 1, pos 0)
Or
[INCOMPLETE_TYPE_DEFINITION.ARRAY] Incomplete complex type: The definition of "ARRAY" type is incomplete. You must provide an element type. For example: "ARRAY<INT>".(line 1, pos 0)

And this is my code convert delta STRUCT type to unity catalog ColumnInfo
    public static List<ColumnInfo> convertAvroSchemaToColumns(StructType structType) throws ApiException {
        List<ColumnInfo> columns = new ArrayList<>();
        StructField[] fields = structType.getFields();
        for (int i = 0; i < fields.length; i++) {
            StructField field = fields[i];
            ColumnInfo column = new ColumnInfo();
            column.setName(field.getName());
            column.setNullable(field.isNullable());
            column.setTypeText(field.getDataType().getSimpleString());
            column.setTypeName(convertAvroTypeToUnityType(field.getDataType()));
            column.setTypeJson(field.getDataType().toJson());
            column.setPosition(i);
            columns.add(column);
        }
        return columns;
    }

    private static ColumnTypeName convertAvroTypeToUnityType(DataType dataType) throws ApiException {
        if (dataType instanceof StringType) {
            return ColumnTypeName.STRING;
        } else if (dataType instanceof BooleanType) {
            return ColumnTypeName.BOOLEAN;
        } else if (dataType instanceof ShortType) {
            return ColumnTypeName.SHORT;
        } else if (dataType instanceof IntegerType) {
            return ColumnTypeName.INT;
        } else if (dataType instanceof LongType) {
            return ColumnTypeName.LONG;
        } else if (dataType instanceof FloatType) {
            return ColumnTypeName.FLOAT;
        } else if (dataType instanceof DoubleType) {
            return ColumnTypeName.DOUBLE;
        } else if (dataType instanceof ByteType) {
            return ColumnTypeName.BYTE;
        } else if (dataType instanceof BinaryType) {
            return ColumnTypeName.BINARY;
[...]

PR #1325, opened 2026-01-27
[SPARK] Support more data types when creating tables.

arturowczarek:
Solves #1004, #891, #801

**PR Checklist**

- [X] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

The Spark plugin didn't have support for types such as Date, Struct or Array. This change adds the support.

The change also adds ExternalTablesManager to create external tables and clean up after use.

Caution for reviewers: there is a small, but important property that must be present in Spark session to propagate table schemas to Unity Catalog (see `DELTA_UPDATE_CATALOG_ENABLED_PROPERTY`). Without it Unity Catalog would record the table but not its schema. In the test we use it together with quite low-level method of loading and accessing the table, because we want to avoid other mechanisms that could infer the schema from the Delta table files (instead of the Unity Catalog).


arturowczarek:
@tdas Please have a look at this pull request. It adds support not only for `Data` type but also for `Array` and `Struct`. The tests are more low-level, avoiding the trap of using Delta schema inference (instead of schema in the Unity Catalog)

tdas:
my apologies. i missed this request. 
@yili-db @openinx can one of you review this?

arturowczarek:
@yili-db, @openinx Is there any chance for your attention before the change becomes stale?

openinx:
Hi @arturowczarek ,  thank you so much for your contribution and sorry for the delayed response.  As this change relates to the delta's schema sync to oss uc server, and we need the PRs in both oss uc and oss delta side to support the full functionality. 

Currently, @yili-db also have two PRs for those changes. 
1. https://github.com/unitycatalog/unitycatalog/pull/1382
2. https://github.com/delta-io/delta/pull/6204

arturowczarek:
@openinx @yili-db @tdas @dennyglee
I have to say I’m very disappointed. Looking at the pull requests, it feels like instead of helping contributors improve their work and build a stronger community, you often end up ghosting them and then taking over their work.

I completely understand that you might not have time to review everything or that you have other priorities. That’s totally fair. But situations like this are very discouraging. Instead of simply saying something like “the implementation is fine, but the tests should be written differently,” the PR was ignored for weeks and later rewritten under your own name.

Experiences like this make it hard for contributors to stay motivated.

openinx:
Hi @arturowczarek Thanks for sharing this, and I’m really sorry the experience felt discouraging. That definitely wasn’t our intention.

You’re right that contributors deserve timely feedback and clearer guidance. In this case we should have communicated earlier instead of leaving the PR without response for so long.

We really appreciate the effort you put into the contribution, and we’d be happy to continue collaborating if you’re open to it.

yili-db:
Implemented by #1382

PR #1331, opened 2026-01-29
Add UniForm support to DeltaCommit API

ChengJiX:
**PR Checklist**

**Description of changes**

make Delta commit API `commit` support optional UniForm related metadata `uniform` inside request body:
```
"uniform" : {
  "iceberg" : {
    "metadata_location": "s3://...", 
    "converted_delta_version: 123,
    "converted_delta_timestamp": "2025-01-04T03:13:11.423Z",
    "base_converted_delta_version": 111
}
```
This is the support on write path for UniForm related metadata. The support on read path will be handled in a follow-up PR

**Tests**

Add tests. Run with
```
./build/sbt "server/testOnly io.unitycatalog.server.sdk.deltacommits.*" 
```

### DIFF
--- a/api/Models/DeltaUniformIceberg.md
+++ b/api/Models/DeltaUniformIceberg.md
@@ -4,9 +4,9 @@
 | Name | Type | Description | Notes |
 |------------ | ------------- | ------------- | -------------|
 | **metadata\_location** | **URI** | The latest Iceberg metadata location. Example: s3://abc/def/metadata/v1.json  | [default to null] |
-| **converted\_delta\_version** | **Long** | The converted Delta version corresponding to the metadata location. It should match the Delta version in the commit info. Example: 1044  | [default to null] |
-| **converted\_delta\_timestamp** | **String** | The timestamp when the corresponding conversion happened. The string must represent a valid instant in UTC with ISO 8601 format Example: 2025-01-04T03:13:11.423Z  | [default to null] |
-| **base\_converted\_delta\_version** | **Long** | Optional version indicating the converted Delta version for the last conversion before the conversion corresponding to the metadata location. Example: 1042  | [optional] [default to null] |
+| **converted\_delta\_version** | **Long** | The Delta version that was converted to Iceberg to produce the Iceberg metadata location. It should match the Delta version in the commit info. Example: 1044  | [default to null] |
+| **converted\_delta\_timestamp** | **String** | The timestamp that Delta finished conversion to produce the Iceberg metadata location. The string must represent a valid instant in UTC with ISO 8601 format. Example: 2025-01-04T03:13:11.423Z  | [default to null] |
+| **base\_converted\_delta\_version** | **Long** | Optional Delta version used to incrementally convert Delta changes to Iceberg changes to produce the latest Iceberg metadata at the metadata_location. Example: 1042  | [optional] [default to null] |
 
 [[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
 


server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
```diff
@@ -693,4 +696,230 @@ public void testCommitWithMetadata() throws ApiException {
...
+              TableInfoDAO.class,
+              UUID.fromString(tableInfo.getTableId()));
+      assertNotNull(tableInfoDAO);
+      assertEquals("s3://my-bucket/metadata/v2.json",
+          tableInfoDAO.getUniformIcebergMetadataLocation());
+      assertEquals(101L, tableInfoDAO.getUniformIcebergConvertedDeltaVersion());
+      assertEquals("2025-01-29T11:00:00.000",
+          tableInfoDAO.getUniformIcebergConvertedDeltaTimestamp());
+    }
+  }
```
ChengJiX, line 787: Nits: should be able to consolidate some shared code paths as helper functions for making tests more extendable

api/all.yaml
```diff
@@ -2987,6 +2987,15 @@ components:
             needs to keep track of commits of versions <= this version.
         metadata:
           $ref: '#/components/schemas/DeltaMetadata'
+        uniform:
+          $ref: '#/components/schemas/Uniform'
+          description: |
+            Optional UniForm metadata to be committed alongside this Delta commit.
```
yili-db: So this is in parallel to `commit_info`. It may or may not come with the same `commit_info` it corresponds to, right?
Or is the intention to make `uniform` always pairing with `commit_info` of the same version?

server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
```diff
@@ -79,6 +79,12 @@ public class TableInfoDAO extends IdentifiableDAO {
   @Column(name = "uniform_iceberg_metadata_location", length = 65535)
   private String uniformIcebergMetadataLocation;
 
+  @Column(name = "uniform_iceberg_converted_delta_version")
+  private Long uniformIcebergConvertedDeltaVersion;
+
+  @Column(name = "uniform_iceberg_converted_delta_timestamp")
+  private String uniformIcebergConvertedDeltaTimestamp;
```
yili-db: Should be `Date`

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -83,6 +85,9 @@ public class DeltaCommitRepository {
    */
   private static final int NUM_COMMITS_PER_BATCH = 20;
 
+  /** The maximum size of the JSON that contains the Delta-to-Iceberg conversion information */
+  private static final int MAX_DELTA_UNIFORM_ICEBERG_SIZE = 1024 * 1024; // 1MB
```
yili-db: This is too big. See the limit in DAO.
```
@Column(name = "uniform_iceberg_metadata_location", length = 65535)
```

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -673,6 +676,36 @@ private List<DeltaCommitDAO> getFirstAndLastCommits(Session session, UUID tableI
...
 
+  /**
+   * Updates table with metadata and uniform information from a commit, then persists changes.
+   *
+   * @param session the Hibernate session for database operations
+   * @param tableId the unique identifier of the table
+   * @param tableInfoDAO the table information data access object to update
+   * @param commit the commit request containing optional metadata and uniform information
+   */
+  private static void updateTableFromCommit(
```
yili-db, line 687: You need to modify `updateTableMetadata` directly and to take commit.getUniform() as one extra parameter.
`updateTableMetadata` already does `session.merge(tableInfoDAO);`. Better not to do it twice.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -717,11 +750,29 @@ private static void updateTableMetadata(
...
+   *
+   * @param tableInfoDAO the table information data access object to update
+   * @param uniform the uniform metadata containing conversion information
+   */
+  private static void updateTableUniform(TableInfoDAO tableInfoDAO, Uniform uniform) {
+    if (uniform.getIceberg() != null) {
+      UniformIceberg icebergMetadata = uniform.getIceberg();
+      tableInfoDAO.setUniformIcebergMetadataLocation(
+          icebergMetadata.getMetadataLocation() != null
+              ? icebergMetadata.getMetadataLocation().toString()
```
yili-db: call `NormalizedURL.normalize` before saving it in DAO.

api/all.yaml
```diff
@@ -2987,6 +2987,15 @@ components:
             needs to keep track of commits of versions <= this version.
         metadata:
           $ref: '#/components/schemas/DeltaMetadata'
+        uniform:
+          $ref: '#/components/schemas/Uniform'
```
tdas: can add Delta
```suggestion
          $ref: '#/components/schemas/DeltaUniform'
```
lets add Delta to every Delta related component.

api/all.yaml
```diff
@@ -3048,6 +3057,45 @@ components:
     DeltaCommitResponse:
       type: object
       description: The response for the Delta commit action
+    Uniform:
```
tdas: DeltaUniform

api/all.yaml
```diff
@@ -3048,6 +3057,45 @@ components:
...
+      description: |
+        Uniform metadata that can be atomically committed alongside a Delta commit.
+        Uniform enables Delta tables to be read by other table formats (e.g. Iceberg)
+        without data conversion. When this message is provided in a Commit request,
+        UC will atomically update the table's format-specific metadata during the commit.
+      properties:
+        iceberg:
+          $ref: '#/components/schemas/UniformIceberg'
+          description: Metadata for Delta UniForm Iceberg conversion.
+    UniformIceberg:
```
tdas: DeltaUniformIceberg

api/all.yaml
```diff
@@ -3048,6 +3057,45 @@ components:
...
+      type: object
+      description: |
+        Iceberg conversion metadata tracking the Delta-to-Iceberg conversion state.
+        It contains the metadata location pointer, converted Delta version, and timestamp
+        of the corresponding conversion.
+      properties:
+        metadata_location:
```
tdas, line 3081: looking at this .. can we make this iceberg_metadata_location. since all the fields here has `delta` explicitly in it.. .. its more consistent to be explicit here as well,

api/all.yaml
ChengJiX: Hi @yili-db, it should be pairing with the `commit_info` of the same version (the `convertedDeltaVersion ` should match with version inside `DeltaCommitInfo`) - technically it can relax this restriction, but in implementation we will require this for DeltaCommit API. Let me add a validation on server side

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
ChengJiX: Hi @yili-db, I have refactored it so that `session.merge(tableInfoDAO)` is only done at most once - only at the bottom of `updateTableFromCommit`. I removed the original `session.merge(tableInfoDAO)` from `updateTableMetadata`: [Link](https://github.com/unitycatalog/unitycatalog/pull/1331/changes#diff-df349b58b0a893abbc64af5ae060645fdd0d5175308fefdc4ff7336a93d3dec1L721).

api/all.yaml
yili-db: This `metadata_location`, which is part of `DeltaUniformIceberg`, doesn't need to have the `iceberg` prefix again.
In JSON it appears as:
```
"uniform" : {
  "iceberg" : {
    "metadata_location": "s3://..."
```

tdas: okay. i guess if unspecified ... its iceberg by default. i am good with that. but please update the description above to make it obvious

ChengJiX: Yeah its already mentioning "iceberg metadata" inside description

api/all.yaml
yili-db: If we always require pairing for UniForm tables, the description should say that, instead of "Typically set..."

api/all.yaml
```diff
@@ -3048,6 +3057,54 @@ components:
...
       description: The response for the Delta commit action
+    DeltaUniform:
+      type: object
+      description: |
+        Uniform metadata that can be atomically committed alongside a Delta commit.
+        Uniform enables Delta tables to be read by other table formats (e.g. Iceberg)
+        without data conversion. When this message is provided in a Commit request,
+        UC will atomically update the table's format-specific metadata during the commit.
+      properties:
+        iceberg:
```
yili-db, line 3069: Make `iceberg` required.

api/all.yaml
```diff
@@ -3048,6 +3057,54 @@ components:
...
+            The timestamp when the corresponding conversion happened.
+            The string must represent a valid instant in UTC with ISO 8601 format
+            Example: 2025-01-04T03:13:11.423Z
+        base_converted_delta_version:
+          type: integer
+          format: int64
+          description: |
+            Optional version indicating the converted Delta version for the last conversion
+            before the conversion corresponding to the metadata location.
+            Example: 1042
```
yili-db, line 3106: This means that conversion isn't required for each single Delta commit?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -835,6 +889,62 @@ private static void validateCommit(DeltaCommit commit) {
...
+    // We check the size of the Delta-to-Iceberg conversion information here to fail early if
+    // it exceeds the maximum size of 1MB. The size is not accurate in terms of the size of
+    // the object in the database but serves as a sanity check to ensure that we're not storing
+    // excessively large objects.
+    int size =
+        iceberg.getMetadataLocation().toString().length()
+            + iceberg.getConvertedDeltaTimestamp().length();
+
+    ValidationUtils.checkArgument(
+        size <= MAX_DELTA_UNIFORM_ICEBERG_SIZE,
```
yili-db, line 938: `MAX_DELTA_UNIFORM_ICEBERG_SIZE` is only the limit for `getMetadataLocation`
`getConvertedDeltaTimestamp` will be converted into `Date`

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -717,11 +750,31 @@ private static void updateTableMetadata(
...
+   * Updates table uniform metadata based on the uniform information provided in a commit.
+   *
+   * <p>This method updates the table's UniForm conversion metadata, including Iceberg metadata
+   * location, converted Delta version, and conversion timestamp.
+   *
+   * @param tableInfoDAO the table information data access object to update
+   * @param uniform the uniform metadata containing conversion information
+   */
+  private static void updateTableUniform(TableInfoDAO tableInfoDAO, DeltaUniform uniform) {
+    if (uniform.getIceberg() != null) {
```
yili-db: this `if` is now redundant.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -835,6 +889,62 @@ private static void validateCommit(DeltaCommit commit) {
...
+    if (commit.getCommitInfo() != null) {
+      ValidationUtils.checkArgument(
+          iceberg.getConvertedDeltaVersion().equals(commit.getCommitInfo().getVersion()),
+          "uniform.iceberg.converted_delta_version (%d) must equal commit version (%d)",
+          iceberg.getConvertedDeltaVersion(),
+          commit.getCommitInfo().getVersion());
+    }
+
+    // We check the size of the Delta-to-Iceberg conversion information here to fail early if
+    // it exceeds the maximum size of 1MB. The size is not accurate in terms of the size of
```
yili-db: not `1MB`

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -835,6 +889,62 @@ private static void validateCommit(DeltaCommit commit) {
...
+   * Validates Delta UniForm Iceberg metadata to ensure all required fields are present, the size is
+   * within limits, and the converted delta version matches the commit version.
+   *
+   * @param commit the commit containing version information
+   * @param uniform the uniform metadata to validate
+   * @throws BaseException if validation fails
+   */
+  private static void validateUniformIceberg(DeltaCommit commit, DeltaUniform uniform) {
+    if (uniform == null || uniform.getIceberg() == null) {
+      return; // DeltaUniform is optional
```
yili-db, line 902: It should be mandatory? https://github.com/unitycatalog/unitycatalog/pull/1331/changes#r2765808574

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -717,11 +750,31 @@ private static void updateTableMetadata(
...
+   * location, converted Delta version, and conversion timestamp.
+   *
+   * @param tableInfoDAO the table information data access object to update
+   * @param uniform the uniform metadata containing conversion information
+   */
+  private static void updateTableUniform(TableInfoDAO tableInfoDAO, DeltaUniform uniform) {
+    if (uniform.getIceberg() != null) {
+      DeltaUniformIceberg icebergMetadata = uniform.getIceberg();
+      tableInfoDAO.setUniformIcebergMetadataLocation(
+          icebergMetadata.getMetadataLocation() != null
```
yili-db: Are the null checks redundant? They are already verified in `validateUniformIceberg`.

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
```diff
@@ -693,4 +699,264 @@ public void testCommitWithMetadata() throws ApiException {
...
   }
+
+  @Test
+  public void testCommitWithUniform() throws ApiException {
+    String timestamp1 = Instant.now().toString();
+    DeltaUniform uniform = new DeltaUniform();
+    DeltaUniformIceberg icebergMetadata =
+        new DeltaUniformIceberg()
+            .metadataLocation(java.net.URI.create("s3://my-bucket/metadata/v1.json"))
+            .convertedDeltaVersion(1L)  // Must match commit version
```
yili-db, line 752: What if it doesn't match? Or it's missing the conversion?
Would be good to cover these cases.

api/all.yaml
ChengJiX: Umm maybe we shall not make it required? There are some other UniForms like Hudi UniForm which can exist separately with Iceberg. Maybe we should leave a possibility for that? But it probably won't happen so we can also make it required now and later change it if needed

yili-db (changes requested):
left comments

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
ChengJiX: Ah yeah, we don't need the check on `ConvertedDeltaTimestamp` now as we validate its a Date compatible timestamp string now.

api/all.yaml
ChengJiX: For UniForm enabled tables, conversion is required for each commit. But for UniForm disabled tables, conversion is not required. And we do allow enabling/disabling UniForm on same tables for multiple times

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
ChengJiX: Yeah its redundant - currently just an extra safety. Let me remove it

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
ChengJiX: Hi @yili-db, we have the validation here: https://github.com/unitycatalog/unitycatalog/pull/1331/changes#diff-df349b58b0a893abbc64af5ae060645fdd0d5175308fefdc4ff7336a93d3dec1R927. We also tested the validation: https://github.com/unitycatalog/unitycatalog/pull/1331/changes#diff-52d156946b0141e68a091d61230e2264a6a1acac4860d8a9e1cef5d3350e0804R947

yili-db: I see. But what if it's missing for a UniForm table? Is it OK for now?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
ChengJiX: `uniform` is optional as UniForm-disabled tables can commit without it

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
ChengJiX: Hi @yili-db, what do you mean by "if it's missing for a UniForm table"? Do you mean the conversion is missing for a UniForm table? The client will makes sure UniForm metadata would be committed when its a UniForm table. On server side I don't think we shall check if the request is coming from UniForm enabled table?

yili-db: So for UniForm tables, clients may commit:
```
commit { commit_info: v1, uniform.iceberg: v1} // this is the ideal case
commit { commit_info: v2} // no iceberg, this is fine? 
commit { uniform.iceberg: v2} // supposedly client will commit later?, but what if client never commits iceberg v2 but then go forward with v3 already?
commit { commit_info: v3, uniform.iceberg: v3} like this ?
```

ChengJiX: Yeah the following is fine
```
commit { commit_info: v2} // no iceberg. It can happen if Uniform is disabled in v2
commit { commit_info: v3, uniform.iceberg: v3} // It can happen If Uniform is re-enabled in v3
```
When UniForm is disabled/enabled, it would be a new commit. So if a client doesn't add uniform metadata in a commit, it means uniform metadata is not needed for the table when that commit takes effect

tdas: i think we need to make this server implementation stronger to begin with - if uniform is enabled on a table we dont accept commits without uniform info. 

later we can relax that. .. but lets start with a stronger forcing function against which we can start testing.

tdas: just to be clear... i am okay merging this PR and making the blocking an independent follow up PR.

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
yili-db: Yes. The entire `try` block that verifies `TableInfoDAO` can be made a function

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
ChengJiX: Hi @tdas, thanks for the review! Sounds good. Let me do it in a follow up PR if decided

api/all.yaml
ChengJiX: Done

api/all.yaml
ChengJiX: Made it required. Can relax it when needed

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -835,6 +882,63 @@ private static void validateCommit(DeltaCommit commit) {
...
+   * @throws BaseException if validation fails
+   */
+  private static void validateUniformIceberg(DeltaCommit commit, DeltaUniform uniform) {
+    if (uniform == null) {
+      return; // DeltaUniform is optional
+    }
+
+    // If uniform is specified, iceberg must be set
+    ValidationUtils.checkArgument(
+        uniform.getIceberg() != null, "Field cannot be null in uniform: iceberg");
```
ChengJiX, line 907: Maybe still good to leave a validation in case the spec is changed?

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
ChengJiX: Done!

api/all.yaml
```diff
@@ -2987,6 +2987,15 @@ components:
...
           $ref: '#/components/schemas/DeltaMetadata'
+        uniform:
+          $ref: '#/components/schemas/DeltaUniform'
+          description: |
+            Optional UniForm metadata to be committed alongside this Delta commit.
+            Uniform enables Delta tables to be read by other table formats (e.g., Iceberg) without data conversion.
+            It is set by UniForm conversion processes after converting a Delta commit.
+            Independent of the `metadata` field, both can be provided together or separately. 
+            When set, UC will atomically update the table's format-specific metadata (e.g. DeltaUniformIceberg),
+            keeping the conversion state synchronized with the Delta version.
```
tdas, line 2997: add a statement saying whether this field is strictly required by a UC API compatible implementation is left up to the implementation.

api/all.yaml
```diff
@@ -3048,6 +3057,56 @@ components:
...
+          $ref: '#/components/schemas/DeltaUniformIceberg'
+          description: Metadata for Delta UniForm Iceberg conversion.
+      required:
+        - iceberg
+    DeltaUniformIceberg:
+      type: object
+      description: |
+        Iceberg conversion metadata tracking the Delta-to-Iceberg conversion state.
+        It contains the metadata location pointer, converted Delta version, timestamp
+        , and optional base converted Delta version of the corresponding conversion.
```
tdas: `,` in a odd position

api/all.yaml
```diff
@@ -3048,6 +3057,56 @@ components:
...
+          type: string
+          format: uri
+          description: |
+            The latest Iceberg metadata location.
+            Example: s3://abc/def/metadata/v1.json
+        converted_delta_version:
+          type: integer
+          format: int64
+          description: |
+            The converted Delta version corresponding to the metadata location.
```
```suggestion
            The Delta version that was converted to Iceberg to produce the latest Iceberg metadata at the metadata_location.
```
more direct, rather than "corresponding"

api/all.yaml
```diff
@@ -3048,6 +3057,56 @@ components:
...
+            It should match the Delta version in the commit info.
+            Example: 1044
+        converted_delta_timestamp:
+          type: string
+          description: |
+            The timestamp when the corresponding conversion happened.
```
tdas: update this to match the above.

api/all.yaml
```diff
@@ -3048,6 +3057,56 @@ components:
...
+            The string must represent a valid instant in UTC with ISO 8601 format
+            Example: 2025-01-04T03:13:11.423Z
+        base_converted_delta_version:
+          type: integer
+          format: int64
+          description: |
+            Optional version indicating the converted Delta version for the last conversion
```
```suggestion
            Optional Delta version used to incrementally convert Delta changes to Iceberg changes to produce the latest Iceberg metadata at the metadata_location.
```

api/all.yaml
tdas: Lets add a statement saying whether the conversion is required for every Delta version is left up to the implementations of this UC API.

ChengJiX:
Ah let me fix the rebase issue

tdas (approved):
LGTM

PR #1334, opened 2026-02-03
Handle credential generation failures gracefully in UCSingleCatalog

murali-db:
## Summary

This PR enables Unity Catalog Spark connector to handle tables where credentials cannot be generated gracefully. Currently, when credentials cannot be generated for such tables, the connector throws an exception. With this change, the connector returns the table metadata without credentials (when a catalog conf is set), allowing downstream code the opportunity to obtain credentials from elsewhere (e.g., a different endpoint).

## Changes

```mermaid
sequenceDiagram
    participant User
    participant UCSingleCatalog
    participant UCProxy
    participant SparkSession
    participant Delta

    User->>UCSingleCatalog: initialize(options with serverSidePlanning.enabled=true)
    UCSingleCatalog->>UCProxy: new UCProxy(sspEnabled=true)
    UCProxy->>UCProxy: Store sspEnabled
    
    User->>UCSingleCatalog: loadTable()
    UCSingleCatalog->>UCProxy: loadTable()
    UCProxy->>UCProxy: Try credentials (fail)
    UCProxy->>UCProxy: Check catalog sspEnabled
    alt SSP Enabled
        UCProxy->>SparkSession: "set(spark.databricks.delta.catalog.enableServerSidePlanning, true)"
        UCProxy->>Delta: Return table with empty credentials
        Delta->>UCProxy: Uses SSP for file access
    else SSP Disabled
        UCProxy-->>User: Throw ApiException
    end
```

- **Changed `temporaryCredentials`** to allow null when credentials cannot be generated
- **Wrapped credential generation** in nested try-catch blocks:
  - First attempts READ_WRITE credentials
  - Falls back to READ credentials if READ_WRITE fails
  - If SSP catalog conf is enabled -> Returns `null` if both fail (indicates user is not authorized for direct credentials) 
  - Else, same exception thrown as before
- **Updated `extraSerdeProps` handling** to return empty map when credentials are unavailable

### Before
- Tables where credentials cannot be generated would fail to load with an exception
- Query would fail completely

### After
- Tables where credentials cannot be generated load successfully without credentials
- Downstream code can handle these tables using alternative approaches
- Regular tables continue to receive credentials as before

## Testing

```mermaid
sequenceDiagram
    participant Test
    participant SparkSession
    participant UCSingleCatalog
    participant UCServer
    
    Note over Test,UCServer: Scenario 1: SSP Disabled
    Test->>SparkSession: create (SSP=false)
    Test->>SparkSession: CREATE EXTERNAL TABLE (unconfigured location)
    Test->>SparkSession: load table
    SparkSession->>UCSingleCatalog: loadTable()
    UCSingleCatalog->>UCServer: generateTemporaryTableCredentials()
    UCServer-->>UCSingleCatalog: 404 (no credentials configured)
    UCSingleCatalog->>UCSingleCatalog: Check serverSidePlanningEnabled=false
    UCSingleCatalog-->>Test: Throw ApiException
    
    Note over Test,UCServer: Scenario 2: SSP Enabled
    Test->>SparkSession: create (SSP=true)
    Test->>SparkSession: CREATE EXTERNAL TABLE (unconfigured location)
    Test->>SparkSession: load table
    SparkSession->>UCSingleCatalog: loadTable()
    UCSingleCatalog->>UCServer: generateTemporaryTableCredentials()
    UCServer-->>UCSingleCatalog: 404 (no credentials configured)
    UCSingleCatalog->>UCSingleCatalog: Check serverSidePlanningEnabled=true
    UCSingleCatalog->>SparkSession: "set(spark.databricks.delta.catalog.enableServerSidePlanning, true)"
    UCSingleCatalog-->>SparkSession: Return table with empty credentials
    SparkSession-->>Test: Table loaded successfully
```

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/S3CredentialTestFileSystem.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/S3CredentialTestFileSystem.java
@@ -49,8 +49,6 @@ public class S3CredentialTestFileSystem extends CredentialTestFileSystem {
           assertThat(credentials.secretAccessKey()).isEqualTo("secretKey1");
           assertThat(credentials.sessionToken()).isEqualTo("sessionToken1");
         }
-      } else if (BaseSparkIntegrationTest.NO_CREDS_BUCKET.equals(host)) {
-        // No credential validation - allows SSP fallback testing with empty credentials.
       } else {
         throw new RuntimeException("invalid path: " + f);
       }


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -340,23 +345,50 @@ private class UCProxy(
...
-            )
+          try {
+            tableOp = TableOperation.READ
+            temporaryCredentialsApi
+              .generateTemporaryTableCredentials(
+                new GenerateTemporaryTableCredential().tableId(tableId).operation(tableOp)
+              )
+          } catch {
+            case e: ApiException =>
+              if (serverSidePlanningEnabled) {
```
tdas: it will be more readable if this chunk of code is moved. to a separate funciton rather than this deep nesting.

tdas: in fact why does this have to inside ApiException. there is already a check below on whether temporaryCredentials is null or not. this whole check trigger for serverSidePlanningEnabled only if temporaryCredentials is null .. right? so why put this code there?

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaExternalTableReadWriteTest.java
```diff
@@ -39,4 +57,82 @@ public void testDeltaPathTable() throws IOException {
...
   }
+
+  /** Creates a SparkSession with SSP (server-side planning) optionally enabled for the catalog. */
+  private SparkSession createSparkSessionWithSsp(boolean sspEnabled) {
+    SparkSession.Builder builder =
+        SparkSession.builder()
+            .appName("test-ssp")
+            .master("local[*]")
+            .config("spark.sql.shuffle.partitions", "4")
+            .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
```
tdas: does the code changes need to be tested with a delta table? its unrelated right?

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaExternalTableReadWriteTest.java
```diff
@@ -39,4 +57,82 @@ public void testDeltaPathTable() throws IOException {
...
+    }
+
+    if (!sspEnabled) {
+      // SSP disabled (default): loadTable() throws ApiException because credential API fails
+      assertThat(caughtException).isInstanceOf(ApiException.class);
+    } else {
+      // SSP enabled: loadTable() succeeds with empty credentials (no ApiException)
+      // The SSP fallback logs: "Server-side planning enabled... Proceeding with empty credentials"
+      assertThat(caughtException).isNull();
+      assertThat(loadedTable).isNotNull();
```
tdas: you are not testing that the delta SSP conf is being set?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -340,23 +345,38 @@ private class UCProxy(
...
-      renewCredEnabled,
-      locationUri.getScheme,
-      uri.toString,
-      tokenProvider,
-      tableId,
-      tableOp,
-      temporaryCredentials,
-    )
+    // If credentials are null here, it means serverSidePlanningEnabled 
+    // is true, so enable server side planning spark config
```
murali-db: in fact, we wouldn't even reach here.. throw happens earlier.

yili-db: I don't understand this comment. It **should** reach here from line R356, right?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -340,23 +345,38 @@ private class UCProxy(
...
+    if (temporaryCredentials == null) {
```
yili-db: `if (serverSidePlanningEnabled && temporaryCredentials == null)` is more readable and future proof.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -465,6 +485,28 @@ private class UCProxy(
...
 
+  /**
+   * Enables server-side planning by setting the appropriate Spark config.
+   * Called when credential vending fails but SSP is enabled, allowing Delta
+   * to use server-side planning for data access instead of credentials from UC.
+   */
+  private def enableServerSidePlanningConfig(identifier: TableIdentifier): Unit = {
+    SparkSession.getActiveSession match {
+      case Some(spark) =>
+        spark.conf.set("spark.databricks.delta.catalog.enableServerSidePlanning", "true")
```
yili-db, line 496: I am not sure here. Is the `spark.conf` global as part of the spark session? Or is it reset after **this** table loading?

connectors/spark/src/test/java/io/unitycatalog/spark/S3CredentialTestFileSystem.java
```diff
@@ -49,6 +49,8 @@ protected void checkCredentials(Path f) {
           assertThat(credentials.secretAccessKey()).isEqualTo("secretKey1");
           assertThat(credentials.sessionToken()).isEqualTo("sessionToken1");
         }
+      } else if (BaseSparkIntegrationTest.NO_CREDS_BUCKET.equals(host)) {
+        // No credential validation - allows SSP fallback testing with empty credentials.
```
yili-db: Is this actually required? This code is about accessing files using the preset mock credential. In SSP case, it should not access the cloud files, right?
Or is it because the implementation isn't complete so that you can not test the entire process yet?

murali-db: this is actually to allow for tests. when setting up the tests, we don't apply any cred validation and allow writes. but when uc wants to actually to vend creds it would fail as there is no creds set to vend

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
murali-db: the intention is to set it for the whole spark session. ideally is that this conf is per-table load; this should be addressed in a future follow up (move the conf to be table scoped). but we already have some client side code which relies on the session-level conf being set. and this operation here allows us to set that session-level conf to influence the client behavior.

connectors/spark/src/test/java/io/unitycatalog/spark/S3CredentialTestFileSystem.java
yili-db: What's the failure here? I thought this is emulating cloud FS access (after cred vending) which your test shouldn't reach?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
yili-db: So it's OK to leave it on for the entire session, including consequent operations on other tables as well?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
murali-db: or the true path, yes. just wanted to note that if the `throw e` occurs (when serverSidePlanningEnabled is false) then exception propagates to caller of loadTable.

connectors/spark/src/test/java/io/unitycatalog/spark/S3CredentialTestFileSystem.java
murali-db: you're right, we don't need this now. only if there are any future tests that actually try to access the files but the ssp case, we don't

yili-db (approved):

PR #1335, opened 2026-02-03
Unity Catalog Managed Tables Specification

zhenlineo:
**PR Checklist**

**Description of changes**

Add the md specification for Unity Catalog Managed Tables

### DIFF
--- a/spec/protocols/ManagedTablesSpec.md
+++ b/spec/protocols/ManagedTablesSpec.md
@@ -7,7 +7,7 @@
 ## Table of Contents
 
 - [Introduction](#introduction)
-- [Terminologies](#terminologies)
+- [Terminology](#terminology)
 - [APIs](#apis)
   - [Create a Staging Table](#create-a-staging-table)
   - [Create a Table](#create-a-table)
@@ -18,7 +18,7 @@
 ---
 
 ## Introduction
-This specification defines how Unity Catalog will serve as the managing catalog for [Catalog-Managed Tables](https://github.com/delta-io/delta/blob/master/protocol_rfcs/catalog-managed.md).
+This specification defines how Unity Catalog will serve as the managing catalog for [Catalog-Managed Tables](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#catalog-managed-tables).
 Catalog-Managed Tables is a new Delta table feature which makes the catalog the source of truth for commits to a table.
 It defines a managing catalog to perform commits according to the Catalog-Managed Tables specification.
 
@@ -29,7 +29,7 @@ This specification defines the APIs to enable the following operations between t
 - Running maintenance operations.
 
 ### What is a Unity Catalog Managed Table?
-A Unity Catalog Managed table is a table whose data location and lifecycle (including deletion of files when dropped)
+A Unity Catalog managed table is a table whose data location and lifecycle (including deletion of files when dropped)
 are fully controlled by the Unity Catalog.
 It is different from external tables, where the catalog only stores metadata pointing to data in an externally managed
 location and does not delete the underlying files when the table is dropped.
@@ -44,7 +44,7 @@ Storage location | Assigned | User-specified
 
 
 ## Terminology
-This section recaps key terminologies used in the catalog-managed tables specification:
+This section recaps key terminologies used in the Catalog-Managed Tables specification:
 - **Commit**: A set of actions that moves a Delta table from version `v−1` to `v`, stored either as a Delta log file or inline in the catalog.
 - **Staged commit**: A commit written to `_delta_log/_staged_commits/<v>.<uuid>.json`, not yet guaranteed to be accepted; the catalog decides whether to ratify it.
 - **Ratified commit**: A proposed commit that the catalog has chosen as the winner for version `v`; it is part of the official history, whether or not it is already published.
@@ -53,14 +53,14 @@ This section recaps key terminologies used in the catalog-managed tables specifi
 For further details, refer to the [Catalog-Managed Tables](https://github.com/delta-io/delta/blob/master/protocol_rfcs/catalog-managed.md).
 
 ## APIs
-A Unity Catalog Managed table needs to be registered and used according to the following specification:
-- **Create a table**: A Unity Catalog Managed table needs to first be created.
-  - The client sends a `POST` request to the [`createStagingTable`](#create-a-staging-table) API to initiate the creation of a new Managed table in Unity Catalog.
+A Unity Catalog managed table needs to be registered and used according to the following specification:
+- **Create a table**: A Unity Catalog managed table needs to first be created.
+  - The client sends a `POST` request to the [`createStagingTable`](#create-a-staging-table) API to initiate the creation of a new managed table in Unity Catalog.
   - The server upon receiving the request to create a staging table, allocates a unique table ID and a path for the table.
   - The client writes the first commit `0.json` under the `_delta_log` in the assigned table path. The `0.json` file should include all the table features and properties required by this specification. See the full list in [Required Table Features and Properties on a Unity Catalog Managed Table](#required-table-features-and-properties-on-a-unity-catalog-managed-table). Then the client finishes the table creation by sending a `POST` request to the [`createTable`](#create-a-table) API.
     - For CTAS (Create Table As Select), the `0.json` file would also contain `AddFile` actions.
-    - For REPLACE, do not drop and recreate the table. Instead, add a new commit that updates the table metadata and removes all existing data files. This approach ensures atomicity, preserves the table ID and history. The replaced table remains to be a Unity Catalog Managed Table.
-  - When a create table request is received, the server first confirms the existence of a table with the specified table ID at the provided location. If the table has the Catalog Managed feature enabled, the server then validates that all necessary table features and properties are correctly configured. Finally, the server registers the table as created at commit version 0.
+    - For REPLACE, do not drop and recreate the table. Instead, add a new commit that updates the table metadata and removes all existing data files. This approach ensures atomicity, preserves the table ID and history. The replaced table remains to be a Unity Catalog managed table.
+  - When a create table request is received, the server first confirms the existence of a table with the specified table ID at the provided location. If the table has the Catalog-Managed feature enabled, the server then validates that all necessary table features and properties are correctly configured. Finally, the server registers the table as created at commit version `0`.
 - **Get the table**: After the table is created successfully, the client can use the three-part name (`catalog_name.schema_name.table_name`) to retrieve the table entry.
   - The client sends a `GET` call to the [`getTable`](#get-a-table) API by providing the fully qualified three-part name.
   - The server then responds with detailed information of the table entry, including the table ID, path etc.
@@ -72,7 +72,7 @@ A Unity Catalog Managed table needs to be registered and used according to the f
 - **Write to the table**:
   - The client should first write a staged commit file and then propose to the server to accept this commit by sending a `POST` to the [`commit`](#write-a-commit) API. If the commit contains table protocol or metadata changes, it must insert them as part of the `commit` request. This POST request could also contain an optional field `uniform/iceberg` for UniForm/Managed Iceberg tables.
   - The server verifies the version to commit equals to the last ratified commit version + 1 and accepts the commit. If the request contains the field `uniform/iceberg`, the field will be persisted as part of the table’s latest state in the catalog.
-  - After the commit is successful, the client must publish the ratified commit by copying the committed file to the `_delta_log` folder. After all the commits till this version has been published, the client should inform the server the commit has been published by sending another backfill commit, as a `POST` request to the `commit` API.
+  - After the commit is successful, the client must publish the ratified commit by copying the committed file to the `_delta_log` folder. After all commits up to this version have been published, the client should notify the server by sending a backfill commit via a `POST` request to the `commit` API.
   - The server verifies that the version is equal or less than the last ratified version and optionally removes the entries for the ratified commits.
 - **Metrics and running maintenance operations**: The client should send [metrics](#write-metrics) after each successful commit. These metrics may be used for Unity Catalog to schedule and manage maintenance operations. Unity Catalog enforces a maintenance policy for catalog‑managed tables. By default, clients are only allowed to run checkpoints, log compaction, and version checksum. All other maintenance operations, such as metadata cleanup, OPTIMIZE, VACUUM, and REORG, are disallowed.
 
@@ -86,7 +86,7 @@ POST .../staging-tables
 
 Proposes a staged table for Unity Catalog to allocate a table ID and path for the table.
 Once created, the table ID and path are not mutable.
-A staged table is not yet a real table, as a result, it cannot be used with table APIs such as the `GET tables` API.
+A staged table is not yet a real table, as a result, it cannot be used with table APIs such as the `getTable` API.
 
 #### Request body
 
@@ -140,7 +140,7 @@ PERMISSION_DENIED | 403 | User lacks necessary permissions (CREATE on schema, US
 POST .../tables
 ```
 
-Creates a new Unity Catalog Managed table in the specified catalog and schema. When creating the table, the `0.json` should contain all the following table features and properties.
+Creates a new Unity Catalog managed table in the specified catalog and schema. When creating the table, the `0.json` should contain all the following table features and properties.
 
 #### Required Table Features and Properties on a Unity Catalog Managed Table
 
@@ -173,20 +173,20 @@ Field Name | Data Type | Description | Optional/Required
 name | string | Name of the staging table, relative to parent schema. | required
 catalog_name | string | Name of the parent catalog where the staging table will be created. | required
 schema_name | string | Name of the parent schema relative to its parent catalog. | required
-table_type | string | The type of the table. For Unity Catalog Managed Tables. Must be set to “MANAGED”. | required
-data_source_format | string | Specify the format of the table.<br>For Unity Catalog Managed Tables, must be set to “DELTA” for delta tables. | required
-columns | array of ColumnInfos | The array of ColumnInfo definitions of the table's columns. | optional
+table_type | string | The type of the table. For Unity Catalog managed tables. Must be set to “MANAGED”. | required
+data_source_format | string | Specify the format of the table.<br>For Unity Catalog managed tables, must be set to “DELTA” for delta tables. | required
+columns | array of ColumnInfo | The array of `ColumnInfo` definitions of the table's columns. | optional
 &nbsp;&nbsp;columnInfo.name | string | Name of Column. | optional
 &nbsp;&nbsp;columnInfo.type_text | string | Full data type specification as SQL/catalogString text. | optional
 &nbsp;&nbsp;columnInfo.type_json | string | Full data type specification, JSON-serialized. | optional
 &nbsp;&nbsp;columnInfo.type_name | string | The type of the column. One of BOOLEAN, BYTE, SHORT, INT, LONG, FLOAT, DOUBLE, DATE, TIMESTAMP, TIMESTAMP_NTZ, CHAR, STRING, BINARY, INTERVAL, ARRAY, STRUCT, MAP, DECIMAL, etc. | optional
-&nbsp;&nbsp;columnInfo.type_precision | int32 | Digits of precision; required for DecimalTypes | optional
+&nbsp;&nbsp;columnInfo.type_precision | int32 | Digits of precision; required for DecimalTypes. | optional
 &nbsp;&nbsp;columnInfo.type_scale | int32 | Digits to the right of decimal; Required for DecimalTypes. | optional
 &nbsp;&nbsp;columnInfo.type_internal_type | string | Format of IntervalType. | optional
 &nbsp;&nbsp;columnInfo.position | int32 | Ordinal position of column (starting at position 0). | optional
 &nbsp;&nbsp;columnInfo.nullable | boolean | Whether the field may be Null. Default to true. | optional
 &nbsp;&nbsp;columnInfo.partition_index | int32 | Partition index for column. | optional
-storage_location | string | For Unity Catalog Managed Tables, this must be the staging_location of staging table. | required
+storage_location | string | For Unity Catalog managed tables, this must be the staging_location of staging table. | required
 properties | object | Table properties as key-value pairs. | optional
 
 #### Responses
@@ -339,20 +339,20 @@ Field Name | Data Type | Description | Optional/Required
 name | string | Name of the staging table, relative to parent schema. | required
 catalog_name | string | Name of the parent catalog where the staging table will be created. | required
 schema_name | string | Name of the parent schema relative to its parent catalog. | required
-table_type | string | The type of the table.<br>For Unity Catalog Managed Tables, set the value to be “MANAGED”. | required
-data_source_format | string | Specify the format of the table.<br>For Unity Catalog Managed Tables, set the value to be “DELTA” for delta tables. | required
-columns | array of ColumnInfos | The array of ColumnInfo definitions of the table's columns. | optional
+table_type | string | The type of the table.<br>For Unity Catalog managed tables, set the value to be “MANAGED”. | required
+data_source_format | string | Specify the format of the table.<br>For Unity Catalog managed tables, set the value to be “DELTA” for delta tables. | required
+columns | array of ColumnInfo | The array of `ColumnInfo` definitions of the table's columns. | optional
 &nbsp;&nbsp;columnInfo.name | string | Name of the column. | optional
 &nbsp;&nbsp;columnInfo.type_text | string | Full data type specification as SQL/catalogString text. | optional
 &nbsp;&nbsp;columnInfo.type_json | string | Full data type specification, JSON-serialized. | optional
 &nbsp;&nbsp;columnInfo.type_name | string | The type of the column. One of BOOLEAN, BYTE, SHORT, INT, LONG, FLOAT, DOUBLE, DATE, TIMESTAMP, TIMESTAMP_NTZ, CHAR, STRING, BINARY, INTERVAL, ARRAY, STRUCT, MAP, DECIMAL, etc. | optional
-&nbsp;&nbsp;columnInfo.type_precision | int32 | Digits of precision; required for DecimalTypes | optional
+&nbsp;&nbsp;columnInfo.type_precision | int32 | Digits of precision; required for DecimalTypes. | optional
 &nbsp;&nbsp;columnInfo.type_scale | int32 | Digits to the right of decimal; Required for DecimalTypes. | optional
 &nbsp;&nbsp;columnInfo.type_internal_type | string | Format of IntervalType. | optional
 &nbsp;&nbsp;columnInfo.position | int32 | Ordinal position of column (starting at position 0). | optional
 &nbsp;&nbsp;columnInfo.nullable | boolean | Whether the field may be Null. Default to true. | optional
 &nbsp;&nbsp;columnInfo.partition_index | int32 | Partition index for column. | optional
-storage_location | string | Storage root URL for table | required
+storage_location | string | Storage root URL for table. | required
 properties | object | Table properties as key-value pairs. | optional
 owner | string | Username of current owner of table. | required
 created_at | int64 | Time at which this table was created, in epoch milliseconds. | required
@@ -438,8 +438,8 @@ PERMISSION_DENIED | 403 | User lacks necessary permissions (SELECT on the table,
 ```
 GET .../delta/commits
 ```
-Retrieve all ratified commits from Unity Catalog.
-Note: Server implementations may choose to paginate responses. A paginated response is indicated by the highest returned version in the commits array being less than `latest_table_version`. In this case, the client should send another GET request with an adjusted `start_version` to retrieve any remaining commits.
+Retrieves all ratified commits from Unity Catalog.  
+Note: Server implementations may choose to paginate responses. A paginated response is indicated by the highest returned version in the commits array being less than `latest_table_version`. In this case, the client should send another request with an adjusted `start_version` to retrieve any remaining commits.
 
 #### Request body
 
@@ -462,7 +462,7 @@ commits | array of CommitInfo | List of ratified commit files tracked in Unity C
 &nbsp;&nbsp;commitInfo.file_name | string | Commit filename. | required
 &nbsp;&nbsp;commitInfo.file_size | int64 | Size of the commit file in bytes. | required
 &nbsp;&nbsp;commitInfo.file_modification_timestamp | int64 | Filesystem modification time in milliseconds since epoch. | required
-latest_table_version | int64 | The latest version tracked by UC for this table. This is the absolute latest version, regardless of `end_version` parameter. Clients can use this to know if they have all commits and cannot load a table version beyond this. 0 is returned for a newly created managed table before any commit is posted. -1 is returned if Unity Catalog does not manage this table. | required
+latest_table_version | int64 | The latest version tracked by Unity Catalog for this table. This is the absolute latest version, regardless of `end_version` parameter. Clients can use this to know if they have all commits and cannot load a table version beyond this. 0 is returned for a newly created managed table before any commit is posted. -1 is returned if Unity Catalog does not manage this table. | required
 
 #### Request-Response Samples
 
@@ -548,7 +548,7 @@ Error Code | HTTP Status | Description
 -|-|-
 TABLE_DOES_NOT_EXIST | 404 | The specified table doesn't exist.<br>The client should check the spelling of the table or create the table first.
 PERMISSION_DENIED | 403 | User lacks necessary permissions (SELECT on table).<br>The client should check with their Unity Catalog Admin to grant them the necessary rights to query commits.
-INVALID_PARAMETER_VALUE | 400 | Server side check failed to get commits for Unity Catalog Managed Tables.<br>The client has provided wrong inputs to the server. This could include but not limited to the following errors: (1) missing required fields, (2) provided fields are invalid such as the start version is negative, the start version is greater than the end version, etc.
+INVALID_PARAMETER_VALUE | 400 | Server side check failed to get commits for Unity Catalog managed tables.<br>The client has provided wrong inputs to the server. This could include but not limited to the following errors: (1) missing required fields, (2) provided fields are invalid such as the start version is negative, the start version is greater than the end version, etc.
 
 
 
@@ -572,11 +572,13 @@ commit_info | CommitInfo | Contains information about the new commit being regis
 &nbsp;&nbsp;commitInfo.file_name | string | UUID-based commit filename. | required
 &nbsp;&nbsp;commitInfo.file_size | int64 | Size of the commit file in bytes. | required
 &nbsp;&nbsp;commitInfo.file_modification_timestamp | int64 | Filesystem modification time in milliseconds since epoch. | required
-latest_published_version | int64 | Notifies UC that commits up to this version have been published. Constraint: At least one of `commit_info` or `latest_published_version` must be present. | optional
-metadata | Metadata | Delta metadata action if the commit changes table metadata. The posted metadata object is the desired final state, so all existing data that needs to be preserved must be included as well as the fields that changed. Any properties returned from Unity Catalog should be included. When to include: Schema changes, property updates, comment changes. Requires `commit_info`. | optional
+latest_published_version | int64 | Notifies Unity Catalog that commits up to this version have been published. Constraint: At least one of `commit_info` or `latest_published_version` must be present. | optional
+metadata | Metadata | Delta metadata action if the commit changes table metadata. The posted `Metadata` object is the desired final state, so all existing data that needs to be preserved must be included as well as the fields that changed. Any properties returned from Unity Catalog should be included. When to include: Schema changes, property updates, comment changes. Requires `commit_info`. | optional
 &nbsp;&nbsp;metadata.description | string | The table comment. | optional
-&nbsp;&nbsp;metadata.schema | array of ColumnInfos | The table schema defined by an array of `ColumnInfo` objects. | optional
-&nbsp;&nbsp;metadata.properties | object | Table properties as key-value pairs. When updating, always includes all the required [properties](#required-table-features-and-properties-on-a-unity-catalog-managed-table). | optional
+&nbsp;&nbsp;metadata.schema | Columns | The table schema defined by a `Columns` object. | optional
+&nbsp;&nbsp;&nbsp;&nbsp;metadata.schema.columns | array of ColumnInfo | The table schema defined by an array of `ColumnInfo` objects. | optional
+&nbsp;&nbsp;metadata.properties | Properties | Table properties defined by an `Properties` object. | optional
+&nbsp;&nbsp;&nbsp;&nbsp;metadata.properties.properties | object | Table properties as key-value pairs. When updating, always includes all the required [properties](#required-table-features-and-properties-on-a-unity-catalog-managed-table). | optional
 uniform | Uniform | UniForm conversion information related to this Delta commit. | optional
 &nbsp;&nbsp;uniform.iceberg | Iceberg | Uniform Iceberg info. | optional
 &nbsp;&nbsp;&nbsp;&nbsp;uniform.iceberg.metadata_location | string | Iceberg metadata location converted up to the Delta version. | required
@@ -612,45 +614,49 @@ uniform | Uniform | UniForm conversion information related to this Delta commit.
     {}
     ```
 
-2. Propose a new staged commit with a Metadata change
+   2. Propose a new staged commit with a metadata change
 
-    Request:
-    ```json
-    {
-      "table_id": "abcdef12-3456-7890-abcd-ef1234567890",
-      "table_uri": "s3://my-bucket/main/default/my_table",
-      "commit_info": {
-        "version": 43,
-        "timestamp": 1704067300000,
-        "file_name": "v.uuid1.json",
-        "file_size": 3072,
-        "file_modification_timestamp": 1704067300000
-      },
-      "metadata": {
-        "description": "Updated table description",
-        "schema": [
-          {"name": "id", "type": "long", "nullable": false},
-          {"name": "name", "type": "string", "nullable": true},
-          {"name": "created_at", "type": "timestamp", "nullable": false}
-        ],
-       "properties": {
-         "delta.minReaderVersion": "3",
-         "delta.minWriterVersion": "7",
-         "delta.enableInCommitTimestamps": "true",
-         "delta.feature.inCommitTimestamp": "supported",
-         "delta.feature.vacuumProtocolCheck": "supported",
-         "delta.feature.catalogManaged": "supported",
-         "io.unitycatalog.tableId": "abcdef12-3456-7890-abcd-ef1234567890",
-         "delta.lastUpdateVersion": "0",
-         "delta.lastCommitTimestamp": "1704067400000"
-       }
-     }
-    }
-    ```
-    Response: 200
-    ```json
-    {}
-    ```
+       Request:
+       ```json
+       {
+         "table_id": "abcdef12-3456-7890-abcd-ef1234567890",
+         "table_uri": "s3://my-bucket/main/default/my_table",
+         "commit_info": {
+           "version": 43,
+           "timestamp": 1704067300000,
+           "file_name": "v.uuid1.json",
+           "file_size": 3072,
+           "file_modification_timestamp": 1704067300000
+         },
+         "metadata": {
+           "description": "Updated table description",
+           "schema": {
+             "columns":[
+               {"name": "id", "type_text": "long", "nullable": false},
+               {"name": "name", "type_text": "string", "nullable": true},
+               {"name": "created_at", "type_text": "timestamp", "nullable": false}
+             ]
+           },
+           "properties": {
+             "properties": {
+               "delta.minReaderVersion": "3",
+               "delta.minWriterVersion": "7",
+               "delta.enableInCommitTimestamps": "true",
+               "delta.feature.inCommitTimestamp": "supported",
+               "delta.feature.vacuumProtocolCheck": "supported",
+               "delta.feature.catalogManaged": "supported",
+               "io.unitycatalog.tableId": "abcdef12-3456-7890-abcd-ef1234567890",
+               "delta.lastUpdateVersion": "0",
+               "delta.lastCommitTimestamp": "1704067400000"
+             }
+           }
+         }
+      }
+       ```
+       Response: 200
+       ```json
+       {}
+       ```
 
 3. Update the latest published version
 
@@ -704,7 +710,7 @@ Error Code | HTTP Status | Description
 -|-|-
 TABLE_DOES_NOT_EXIST | 404 | The specified table doesn't exist.<br>The client should check if the table ID is correct for this table.
 PERMISSION_DENIED | 403 | Insufficient permissions (MODIFY on table).<br>The client should check with their Unity Catalog Admin to grant them the necessary rights to write commits.
-INVALID_PARAMETER_VALUE | 400 | Server side check failed to write commits to Unity Catalog Managed Tables.<br>The client has provided wrong inputs to the server. This could include but not limited to the following errors: (1) missing required fields, (2) provided fields are invalid, such as version is not positive, timestamp is not positive, file name is missing, file size is not positive, file modification time is not positive, (3) commit version is not exactly `last_version + 1`, (4) last published version > current table version, etc.
+INVALID_PARAMETER_VALUE | 400 | Server side check failed to write commits to Unity Catalog managed tables.<br>The client has provided wrong inputs to the server. This could include but not limited to the following errors: (1) missing required fields, (2) provided fields are invalid, such as version is not positive, timestamp is not positive, file name is missing, file size is not positive, file modification time is not positive, (3) commit version is not exactly `last_version + 1`, (4) last published version > current table version, etc.
 ALREADY_EXISTS | 409 | Commit version already exists (concurrent write).<br>The client should follow the Delta protocol to rebuild the snapshot and retry the commit with a new version.
 
 
@@ -715,7 +721,7 @@ ALREADY_EXISTS | 409 | Commit version already exists (concurrent write).<br>The 
 POST .../delta/metrics
 ```
 
-Publish metrics information for maintenance operations provided by the Unity Catalog.
+Publishes metrics information for maintenance operations provided by the Unity Catalog.
 
 #### Request body
 
@@ -786,7 +792,7 @@ Error Code | HTTP Status | Description
 -|-|-
 TABLE_DOES_NOT_EXIST | 404 | The specified table doesn't exist.<br>The client should check if the table ID is correct for this table.
 PERMISSION_DENIED | 403 | Insufficient permissions (SELECT on table).<br>The client should check with their Unity Catalog Admin to grant them the necessary rights to publish metrics.
-INVALID_PARAMETER_VALUE | 400 | Server side check failed to post metrics for Unity Catalog Managed Tables.<br>The client has provided wrong inputs to the server. This could include but not limited to the following errors:(1) missing required fields, (2) provided fields are invalid such as non-positive numbers, etc.
+INVALID_PARAMETER_VALUE | 400 | Server side check failed to post metrics for Unity Catalog managed tables.<br>The client has provided wrong inputs to the server. This could include but not limited to the following errors:(1) missing required fields, (2) provided fields are invalid such as non-positive numbers, etc.
 
 ## API Reference
 


spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
+# Unity Catalog Managed Tables Specification
+
+> **Status**: Draft  
+> **Version**: 0.1.0  
+> **Last Updated**: 2026-02-02
+## Table of Contents
```
tdas, line 7: [AI Generated] Missing blank line between the metadata block and `## Table of Contents`. Also, the TOC has inconsistent capitalization: "Create a Staging Table" / "Get a Table" (title case) vs "Get commits" / "Write a commit" / "Write metrics" (lowercase). These should all use consistent casing to match their section headings.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+Comparison with External Tables:
+
+Feature | Managed Table | External Table
+-|-|-
+Storage ownership | Unity Catalog | User
+Auto-cleanup on drop | Yes | No
+Storage location | Assigned | User-specified
+
+
+## Terminologies
```
tdas: [AI Generated] nit: "Terminologies" → "Terminology". "Terminology" is the standard uncountable form in English for a section heading.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+This section recaps key terminologies used in the catalog-managed tables specification:
+- **Commit**: A set of actions that moves a Delta table from version ` v−1` to `v`, stored either as a Delta log file or inline in the catalog.
```
tdas: [AI Generated] nit: The backtick formatting `` ` v−1` `` has a leading space and uses a Unicode minus sign (U+2212) instead of a regular hyphen. Should be `` `v-1` ``.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+  - The server verifies the version to commit equals to the last ratified commit version + 1 and accepts the commit. If the request contains the field `uniform/iceberg`, the field will be persisted as part of the table’s latest state in the catalog.
+  - After the commit is successful, the client must publish the ratified commit by copying the committed file to the `_delta_log` folder, and inform the server that the file has been published by sending another backfill commit, as a `POST` request to the `commit` API.
+  - The server verifies that the version is equal or less than the last ratified version and optionally removes the entries for the ratified commits.
+- **Metrics and running maintenance operations**: The client should send [metrics](#write-metrics) after each successful commit. These metrics may be used for Unity Catalog to schedule and manage maintenance operations. The Unity Catalog follows the maintenance operation policy on Catalog-managed Tables. By default it disallows clients to run all maintenance operations, including but not limit to metadata cleanup, OPTIMIZE, VACUUM, and REORG.
```
tdas: [AI Generated] Grammar: "not limit to" should be "not limited to". This appears in at least 4 places in the document (here, and in error descriptions for Create Table, Get Commits, Write Commit, and Write Metrics).

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+&nbsp;&nbsp;columnInfo.name | string | Name of Column. | optional
+&nbsp;&nbsp;columnInfo.type_text | string | Full data type specification as SQL/catalogString text. | optional
+&nbsp;&nbsp;columnInfo.type_json | string | Full data type specification, JSON-serialized. | optional
+&nbsp;&nbsp;columnInfo.type_name | string | The type of the column. One of BOOLEAN, BYTE, SHORT, INT, LONG, FLOAT, DOUBLE, DATE, TIMESTAMP, TIMESTAMP_NTZ, CHAR, STRING, BINARY, INTERVAL, ARRAY, STRUCT, MAP, DECIMAL, etc. | optional
+&nbsp;&nbsp;columnInfo.type_precision | int32 | Digits of precision; required for DecimalTypes | optional
+&nbsp;&nbsp;columnInfo.type_scale | int32 | Digits to the right of decimal; Required for DecimalTypes. | optional
+&nbsp;&nbsp;columnInfo.type_internal_type | string | Format of IntervalType. | optional
+&nbsp;&nbsp;columnInfo.position | int32 | Ordinal position of column (starting at position 0). | optional
+&nbsp;&nbsp;columnInfo.nullable | boolean | Whether the field may be Null. Default to true. | optional
+&nbsp;&nbsp;columnInfo.position | int32 | Partition index for column. | optional
```
tdas: [AI Generated] Bug: `columnInfo.position` is defined twice. This second entry describes "Partition index for column" and should be `columnInfo.partition_index` — matching the response sample on line 246 which uses `"partition_index": 0`.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+owner | string | Username of current owner of table. | required
+created_at | int64 | Time at which this table was created, in epoch milliseconds. | required
+created_by | string | Username of table creator. | required
+updated_at | int64 | Time at which this table was last modified, in epoch milliseconds. | required
+updated_by | string | Username of user who last modified the table. | optional
+table_id | string | Unique identifier for the table.<br>Type: UUID string. | required
+
+Request-Response Samples
+
+Request: `POST .../tables/main.default.my_table`
```
tdas: [AI Generated] Bug: This says `POST .../tables/main.default.my_table` but the API is defined as `GET .../tables/{full_name}`. Should be `GET`.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+    }
+    ```
+
+#### Errors
+
+Possible Errors on this API are listed below:
+
+Error Code | HTTP Status | Description
+-|-|-
+CATALOG_DOES_NOT_EXIST | 404 | The specified catalog doesn't exist.<br>The client should check the spelling of the catalog name or create the catalog first.
```
tdas: [AI Generated] The Get Commits request only takes `table_id` and `table_uri` — not `catalog_name` or `schema_name`. Why would `CATALOG_DOES_NOT_EXIST` and `SCHEMA_DOES_NOT_EXIST` be returned here? If the server resolves these internally from the table ID, that should be explained. Otherwise these error codes seem inapplicable to this API.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+metadata | Metadata | Delta metadata action if the commit changes table metadata. The posted metadata object is the desired final state, so all existing data that needs to be preserved must be included as well as the fields that changed. Any properties returned from Unity Catalog should be included. When to include: Schema changes, property updates, comment changes. Requires `commit_info`. | optional
+&nbsp;&nbsp;metadata.id | string | The internal table ID created by Delta. | required
+&nbsp;&nbsp;metadata.name | string | The table name. | optional
+&nbsp;&nbsp;metadata.description | string | The table comment. | optional
+&nbsp;&nbsp;metadata.format | Format | Object that contains format information. | optional
+&nbsp;&nbsp;&nbsp;&nbsp;metadata.format.provider | string | The format of the files, e.g. "parquet". | optional
+&nbsp;&nbsp;&nbsp;&nbsp;metadata.format.options | object | Format-specific options as key-value pairs. | optional
+&nbsp;&nbsp;metadata.schema_string | array of ColumnInfos | The table schema as Unity Catalog defined by an array of `ColumnInfo` objects. | optional
```
tdas: [AI Generated] The field is named `schema_string` but its type is declared as "array of ColumnInfos". In the sample (lines 658-661), it's shown as an array of Delta-style schema objects (`{"name": "id", "type": "long", ...}`), which is neither a string nor a UC ColumnInfo format. In Delta protocol, `schemaString` is a JSON string. The name, declared type, and sample are all inconsistent with each other. What exactly should the server expect here?

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+&nbsp;&nbsp;metadata.partition_columns | array of strings | The partition columns. | optional
+&nbsp;&nbsp;metadata.properties | object | Table properties as key-value pairs. When updating, always includes all the required [properties](#required-table-features-and-properties-on-a-unity-catalog-managed-table). | optional
+&nbsp;&nbsp;metadata.created_time | long | Timestamp when metadata was created in milliseconds since epoch. | optional
+&nbsp;&nbsp;protocol.min_reader_version | int32 | Minimum reader version required. | required
```
tdas: [AI Generated] The `protocol.*` fields are listed under the `metadata` prefix, but in the Delta protocol, Protocol and Metadata are separate actions. Is `protocol` truly a sub-object of `metadata` in this API, or a sibling field in the request body? The hierarchy should be clarified, and a sample showing a protocol change would help.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+         "delta.enableInCommitTimestamps": "true",
+         "delta.feature.inCommitTimestamp": "supported",
+         "delta.feature.vacuumProtocolCheck": "supported",
+         "delta.feature.catalogManaged": "supported",
+         "io.unitycatalog.tableId": "abcdef12-3456-7890-abcd-ef1234567890",
+         "delta.lastUpdateVersion": "0",
+         "delta.lastCommitTimestamp": "1704067400000"
+       }
+     },
+     "created_time": 1804067300000
```
tdas: [AI Generated] Bug: `created_time` is placed outside the `metadata` object (at the same level as `table_id`, `table_uri`, etc.), but according to the field definitions it should be `metadata.created_time` — nested inside `metadata`.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+      "commit_info": {
+        "version": 42,
+        "timestamp": 1704067200000,
+        "file_name": "v.uuid1.json",
+        "file_size": 2048,
+        "file_modification_timestamp": 1704067200000
+      },
+      "uniform": {
+        "iceberg": {
+          "metadata_location": "s3://my-bucket/warehouse/main.db/my_table/metadata/00042-xxxx.metadata.json"
```
tdas: [AI Generated] Bug: Invalid JSON — missing comma after the `metadata_location` value.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+#### Request body
+
+Field Name | Data Type | Description | Optional/Required
+-|-|-|-
+table_id | string | Uniquely identifies the Delta table in Unity Catalog.<br>Format: UUID string. | required
+table_uri | string | The storage location of the Delta table.<br>Format: URI string (`s3://`, `abfss://`, `gs://`, etc.). | required
+report | Report | Contains information about which metrics are being sent. | optional
+&nbsp;&nbsp;report.commit_report | CommitReport | The commit report metrics. | optional
+&nbsp;&nbsp;&nbsp;&nbsp;report.commit_report.num_files_added | int64 | Number of files added by this commit. | optional
+&nbsp;&nbsp;&nbsp;&nbsp;report.commit_report.num_bytes_added | string | Number of bytes added by this commit. | optional
```
tdas: [AI Generated] Bug: `num_bytes_added` is typed as `string`, but all other byte-count fields (`num_bytes_removed`, `num_clustered_bytes_added`, etc.) are `int64`. Should this be `int64` as well?

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+  "report": {
+    "commit_report": {
+      "num_files_added": 10,
+      "num_bytes_added": 104857600,
+      "num_clustered_bytes_added": 52428800,
+      "num_files_removed": 2,
+      "num_bytes_removed": 8192,
+      "num_clustered_bytes_removed": 4096,
+      "file_size_histogram": {
+        "sorted_bin_boundaries": [1048576, 4194304, 16777216],
```
tdas: [AI Generated] The spec for `sorted_bin_boundaries` says "The first element must be 0", but this sample starts with `1048576`. The sample should be updated to match the spec (i.e., start with `0`).

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,829 @@
...
+
+#### Errors
+
+Possible Errors on this API are listed below:
+
+Error Code | HTTP Status | Description
+-|-|-
+TABLE_DOES_NOT_EXIST | 404 | The specified table doesn't exist.<br>The client should check if the table ID is correct for this table.
+PERMISSION_DENIED | 403 | Insufficient permissions (SELECT on table).<br>The client should check with their Unity Catalog Admin to grant them the necessary rights to write commits.
+INVALID_PARAMETER_VALUE | 400 | Server side check failed to get commits for Unity Catalog Managed Tables.<br>The client has provided wrong inputs to the server. This could include but not limit to the following errors:(1) missing required fields, (2) provided fields are invalid such as the start version is negative, the start version is greater than the end version , etc.
```
tdas: [AI Generated] Bug: This error description is copy-pasted from the Get Commits API — it references "get commits", "start version", and "end version", none of which are relevant to the Write Metrics API. This needs to be rewritten for the metrics context.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+
+Field Name | Data Type | Description | Optional/Required
+-|-|-|-
+commits | array of CommitInfo | List of ratified commit files tracked in Unity Catalog. Guaranteed to be contiguous and within the specified [`start_version`, `end_version`] range.<br/>Type: Array of CommitInfo objects.<br/>Note: Can be in arbitrary order. Empty array is valid if no commits in the requested range. | required
+&nbsp;&nbsp;commitInfo.version | int64 | The commit's table version. | required
+&nbsp;&nbsp;commitInfo.timestamp | int64 | Commit timestamp in milliseconds since epoch. | required
+&nbsp;&nbsp;commitInfo.file_name | string | Commit filename. | required
+&nbsp;&nbsp;commitInfo.file_size | int64 | Size of the commit file in bytes. | required
+&nbsp;&nbsp;commitInfo.file_modification_timestamp | int64 | Filesystem modification time in milliseconds since epoch. | required
+latest_table_version | int64 | The latest version tracked by UC for this table. This is the absolute latest version, regardless of `end_version` parameter. Clients can use this to know if they have all commits and cannot load a table version beyond this. -1 is returned if Unity Catalog does not manage this table. | required
```
yili-db: It needs to mention that 0 is returned for a newly created **managed** table before any commit is posted.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+
+#### Errors
+
+Possible Errors on this API are listed below:
+
+Error Code | HTTP Status | Description
+-|-|-
+TABLE_DOES_NOT_EXIST | 404 | The specified table doesn't exist.<br>The client should check the spelling of the table or create the table first.
+PERMISSION_DENIED | 403 | User lacks necessary permissions (SELECT on table).<br>The client should check with their Unity Catalog Admin to grant them the necessary rights to query commits.
+INVALID_PARAMETER_VALUE | 400 | Server side check failed to get commits for Unity Catalog Managed Tables.<br>The client has provided wrong inputs to the server. This could include but not limit to the following errors: (1) missing required fields, (2) provided fields are invalid such as the start version is negative, the start version is greater than the end version, etc.
```
tdas: [AI Generated] nit: "not limit to" → "not limited to". This was fixed in the other 4 locations but missed here.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+&nbsp;&nbsp;metadata.id | string | The internal table ID created by Delta. | required
+&nbsp;&nbsp;metadata.name | string | The table name. | optional
+&nbsp;&nbsp;metadata.description | string | The table comment. | optional
+&nbsp;&nbsp;metadata.format | Format | Object that contains format information. | optional
+&nbsp;&nbsp;&nbsp;&nbsp;metadata.format.provider | string | The format of the files, e.g. "parquet". | optional
+&nbsp;&nbsp;&nbsp;&nbsp;metadata.format.options | object | Format-specific options as key-value pairs. | optional
+&nbsp;&nbsp;metadata.schema | array of ColumnInfos | The table schema defined by an array of `ColumnInfo` objects. | optional
+&nbsp;&nbsp;metadata.partition_columns | array of strings | The partition columns. | optional
+&nbsp;&nbsp;metadata.properties | object | Table properties as key-value pairs. When updating, always includes all the required [properties](#required-table-features-and-properties-on-a-unity-catalog-managed-table). | optional
+&nbsp;&nbsp;metadata.created_time | long | Timestamp when metadata was created in milliseconds since epoch
```
tdas: [AI Generated] This row is missing the `| optional` (or `| required`) column at the end, which will break the markdown table rendering. Also, the type is `long` while all other timestamp fields in the doc use `int64` — should be consistent.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+- **Read from the table**: To read from the table, the client needs to obtain all necessary commits to reconstruct the snapshot:
+  - The client obtains all the published commits by listing all the files in `_delta_log`, as well as to fetch the ratified commits from the server by performing a `GET` call to the [`commits`](#get-commits) API.
+  - The server returns all ratified commits for the table. The server can optionally decide to paginate the response.
+  - The client combines the published commits, ratified commits and rebuilds the current snapshot. When building the snapshot, the client should not read any commits that go beyond max ratified version returned by the server.
+- **Write to the table**:
+  - The client should first write a staged commit file and then propose to the server to accept this commit by sending a `POST` to the [`commit`](#write-a-commit) API. If the commit contains table protocol or metadata changes, it must insert them as part of the `commit` request. This POST request could also contain an optional field `uniform/iceberg` for UniForm/Managed Iceberg tables.
```
tdas, line 73: [AI Generated] This still says "If the commit contains table protocol or metadata changes, it must insert them as part of the `commit` request." But the `protocol` fields were removed from the Write Commit request body in this revision. How should a client communicate protocol changes to the server? Either update this text to remove the protocol reference, or add the protocol fields back to the Write Commit API.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+      },
+      "metadata": {
+        "id": "12345678-1234-5678-1234-567812345678",
+        "name": "my_table",
+        "description": "Updated table description",
+        "format": {
+          "provider": "parquet",
+          "options": {}
+        },
+        "schema": [
```
tdas: [AI Generated] The field was renamed from `schema_string` to `schema` (good), but the sample still uses Delta-style schema format (`{"name": "id", "type": "long", ...}`) rather than UC ColumnInfo format (`type_text`, `type_json`, `type_name`, `position`, etc.) as used in the Create Table samples. The field definition says "array of ColumnInfo objects" — the sample should match whichever format the server actually expects.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+```
+
+#### Errors
+
+Possible Errors on this API are listed below:
+
+Error Code | HTTP Status | Description
+-|-|-
+TABLE_DOES_NOT_EXIST | 404 | The specified table doesn't exist.<br>The client should check if the table ID is correct for this table.
+PERMISSION_DENIED | 403 | Insufficient permissions (SELECT on table).<br>The client should check with their Unity Catalog Admin to grant them the necessary rights to write commits.
```
tdas: [AI Generated] The PERMISSION_DENIED description still says "rights to write commits" — should say "rights to post metrics" or "write metrics".

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+A staged table is not yet a real table, as a result, it cannot be used with table APIs such as the `GET tables` API.
+
+#### Request body
+
+Field Name | Data Type | Description | optional/required
+-|-|-|-
+name | string | Name of the staging table, relative to parent schema. | required
+catalog_name | string | Name of the parent catalog where the staging table will be created. | required
+schema_name | string | Name of the parent schema relative to its parent catalog. | required
+managed_storage_path_type | string | The type of managed storage path. Specifies what kind of storage location is being provisioned.<br/>For Unity Catalog Managed Tables, the value must to be `“MANAGED_TABLE_PATH”`. | required
```
tdas: [AI Generated] nit: "the value must to be" → "the value must be". Also, this is the only field table with lowercase `optional/required` in the header — all others use `Optional/Required`.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+  - The client sends a `POST` request to the [`staging-tables`](#create-a-staging-table) API to initiate the creation of a new Managed table in Unity Catalog.
+  - The server upon receiving the request to create a staging table, allocates a unique table ID and a path for the table.
+  - The client writes the first commit `0.json` under the `_delta_log` in the assigned table path. The `0.json` file should include all the table features and properties required by this specification. See the full list in [Required Table Features and Properties on a Unity Catalog Managed Table](#required-table-features-and-properties-on-a-unity-catalog-managed-table).
+    - For CTAS (Create Table As Select), the `0.json` file would also contain `AddFile` actions. Then the client finishes the table creation by sending a `POST` request to [`tables`](#create-a-table) API.
+    - For REPLACE, do not drop and recreate the table. Instead, add a new commit that updates the table metadata and removes all existing data files. This approach ensures atomicity, preserves the table ID and history. The replaced table remains to be a Unity Catalog Managed Table.
```
openinx: > Instead, add a new commit that updates the table metadata and removes all existing data files

nit: for RTAS, it actually replace (not remove) all existing data files. I'm fine if we are talking about `REPLACE` not RTAS here.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+  - When a create table request is received, the server first confirms the existence of a table with the specified table ID at the provided location. If the table has the Catalog Managed feature enabled, the server then validates that all necessary table features and properties are correctly configured. Finally, the server registers the table as created at commit version 0.
```
tdas: [AI Generated] The create flow has the client write `0.json` directly to storage and then call `POST .../tables` to register. Version 0 is never submitted through the `POST .../delta/commit` API. But the read flow says "the client should not read any commits that go beyond max ratified version returned by the server."

So after table creation, what does `GET .../delta/commits` return? Specifically:
- `latest_table_version` must be `0`, otherwise the read constraint blocks the client from reading version 0 at all (since `-1` means "UC does not manage this table").
- Does the `commits` array include a CommitInfo entry for version 0, or is it empty (with the client expected to read `0.json` from storage as a published commit)?

The spec should explicitly state that when the server processes the `POST .../tables` request and registers the table at version 0, it also records version 0 in its commit tracking so that `latest_table_version >= 0`.

tdas (commented):
[AI Generated] One more specification gap — see inline comment on the Create Table / Get Commits interaction.

spec/protocols/ManagedTablesSpec.md
openinx: > If the table has the Catalog Managed feature enabled, the server then validates that all necessary table features and properties are correctly configured

Hi @yili-db , do we oss uc server already have the check for the necessary table features and properties ?

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+
+```
+POST .../delta/commit
+```
+
+Proposes a staged commit to the catalog and informs Unity Catalog that a commit has been published.
+
+#### Request body
+
+Field Name | Data Type | Description | Optional/Required
```
tdas, line 565: [AI Generated] The Catalog-Managed Tables RFC requires: "The commitInfo action must also contain a field `txnId` that stores a unique transaction identifier string." This is a hard writer requirement for catalog-managed tables, but the Write Commit API has no `txnId` field and the Required Table Properties section doesn't mention it. Should the server validate its presence? Should it be part of the commit request body?

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+#### Errors
+
+Possible Errors on this API are listed below:
+
+Error Code | HTTP Status | Description
+-|-|-
+TABLE_DOES_NOT_EXIST | 404 | The specified table doesn't exist.<br>The client should check if the table ID is correct for this table.
+PERMISSION_DENIED | 403 | Insufficient permissions (MODIFY on table).<br>The client should check with their Unity Catalog Admin to grant them the necessary rights to write commits.
+INVALID_PARAMETER_VALUE | 400 | Server side check failed to write commits to Unity Catalog Managed Tables.<br>The client has provided wrong inputs to the server. This could include but not limited to the following errors: (1) missing required fields, (2) provided fields are invalid, such as version is not positive, timestamp is not positive, file name is missing, file size is not positive, file modification time is not positive, (3) commit version is not exactly `last_version + 1`, (4) last published version > current table version, etc.
+ALREADY_EXISTS | 409 | Commit version already exists (concurrent write).<br>The client should follow the Delta protocol to rebuild the snapshot and retry the commit with a new version.
```
tdas, line 714: [AI Generated] The RFC recommends: "For performance reasons, Delta clients are encouraged to establish an API contract where the catalog provides the latest ratified commit information whenever a commit fails due to version conflict." The ALREADY_EXISTS (409) error currently returns no body. Consider returning `latest_table_version` (and optionally the latest ratified commits) in the error response so the client can rebuild and retry without a separate Get Commits round-trip.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+- **Get the table**: After the table is created successfully, the client can use the three-part name (`catalog_name.schema_name.table_name`) to retrieve the table entry.
+  - The client sends a `GET` call to the [`tables`](#get-a-table) API by providing the fully qualified three-part name.
+  - The server then responds with detailed information of the table entry, including the table ID, path etc.
+- **Read from the table**: To read from the table, the client needs to obtain all necessary commits to reconstruct the snapshot:
+  - The client obtains all the published commits by listing all the files in `_delta_log`, as well as to fetch the ratified commits from the server by performing a `GET` call to the [`commits`](#get-commits) API.
+  - The server returns all ratified commits for the table. The server can optionally decide to paginate the response.
+  - The client combines the published commits, ratified commits and rebuilds the current snapshot. When building the snapshot, the client should not read any commits that go beyond max ratified version returned by the server.
```
tdas, line 71: [AI Generated] The read flow says "The client combines the published commits, ratified commits and rebuilds the current snapshot" but doesn't specify precedence. The Catalog-Managed Tables RFC explicitly requires: "When the catalog returns a ratified commit for version v, readers must use that catalog-supplied commit and ignore any published Delta file for version v that might also be present." This precedence rule should be stated here — without it, a client implementer won't know which source wins when a commit exists in both.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+- **Write to the table**:
+  - The client should first write a staged commit file and then propose to the server to accept this commit by sending a `POST` to the [`commit`](#write-a-commit) API. If the commit contains table protocol or metadata changes, it must insert them as part of the `commit` request. This POST request could also contain an optional field `uniform/iceberg` for UniForm/Managed Iceberg tables.
+  - The server verifies the version to commit equals to the last ratified commit version + 1 and accepts the commit. If the request contains the field `uniform/iceberg`, the field will be persisted as part of the table’s latest state in the catalog.
+  - After the commit is successful, the client must publish the ratified commit by copying the committed file to the `_delta_log` folder, and inform the server that the file has been published by sending another backfill commit, as a `POST` request to the `commit` API.
```
tdas: [AI Generated] The RFC requires: "Commits must be published in order. That is, version v - 1 must be published before version v." This is not stated anywhere in this spec. Without this, a client could incorrectly publish version 5 before version 4, corrupting the table for filesystem-based readers. This should be stated as a requirement, likely in the write flow or as a note on `latest_published_version`.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+  - The server verifies that the version is equal or less than the last ratified version and optionally removes the entries for the ratified commits.
+- **Metrics and running maintenance operations**: The client should send [metrics](#write-metrics) after each successful commit. These metrics may be used for Unity Catalog to schedule and manage maintenance operations. The Unity Catalog follows the maintenance operation policy on Catalog-managed Tables. By default it disallows clients to run all maintenance operations, including but not limited to metadata cleanup, OPTIMIZE, VACUUM, and REORG.
```
tdas: [AI Generated] The RFC says any component — client-side, server-side, or both — can publish ratified commits: "it does not matter whether the client-side, server-side, or both catalog components initiate publishing." But this spec says "the client must publish the ratified commit." This is more restrictive than the protocol allows and doesn't address the case where a client crashes after ratification but before publishing. Consider clarifying that the server (or another client) may also publish ratified commits as a recovery mechanism.

tdas: [AI Generated] The RFC carves out specific maintenance exceptions: "The only exceptions are checkpoints, log compaction, and version checksum, as they are essential for all basic table operations." This spec says maintenance is disallowed "including but not limited to metadata cleanup, OPTIMIZE, VACUUM, and REORG" without mentioning these exceptions. A client reading only this spec might think checkpoints and log compaction are also prohibited.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
+# Unity Catalog Managed Tables Specification
```
tdas, line 1: [AI Generated] The RFC defines three types of commit storage: published, staged, and inline (stored directly in the catalog). This spec only supports staged commits (file-based). If inline commits are intentionally out of scope for v0.1, consider adding a brief note stating that — otherwise a reader comparing this spec against the RFC will wonder if it's an omission.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+  - When a create table request is received, the server first confirms the existence of a table with the specified table ID at the provided location. If the table has the Catalog Managed feature enabled, the server then validates that all necessary table features and properties are correctly configured. Finally, the server registers the table as created at commit version 0.
+- **Get the table**: After the table is created successfully, the client can use the three-part name (`catalog_name.schema_name.table_name`) to retrieve the table entry.
+  - The client sends a `GET` call to the [`tables`](#get-a-table) API by providing the fully qualified three-part name.
+  - The server then responds with detailed information of the table entry, including the table ID, path etc.
+- **Read from the table**: To read from the table, the client needs to obtain all necessary commits to reconstruct the snapshot:
+  - The client obtains all the published commits by listing all the files in `_delta_log`, as well as to fetch the ratified commits from the server by performing a `GET` call to the [`commits`](#get-commits) API.
```
openinx: > as well as to fetch the ratified commits from the server by performing a `GET` call to the [`commits`](#get-commits) API

ni: Should we make it more clearly ?   `GET` call to the `[get-commits](#get-commits)` API.

spec/protocols/ManagedTablesSpec.md
openinx, line 73: > sending a `POST` to the [`commit`](#write-a-commit) API.

we need to be more clear,  saying it's a `writeCommit` or `getCommit` api.

spec/protocols/ManagedTablesSpec.md
openinx: > inform the server that the file has been published by sending another backfill commit, as a `POST` request to the `commit` API.

Pls add a link for the `commit` API.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+- **Write to the table**:
+  - The client should first write a staged commit file and then propose to the server to accept this commit by sending a `POST` to the [`commit`](#write-a-commit) API. If the commit contains table protocol or metadata changes, it must insert them as part of the `commit` request. This POST request could also contain an optional field `uniform/iceberg` for UniForm/Managed Iceberg tables.
+  - The server verifies the version to commit equals to the last ratified commit version + 1 and accepts the commit. If the request contains the field `uniform/iceberg`, the field will be persisted as part of the table’s latest state in the catalog.
+  - After the commit is successful, the client must publish the ratified commit by copying the committed file to the `_delta_log` folder, and inform the server that the file has been published by sending another backfill commit, as a `POST` request to the `commit` API.
+  - The server verifies that the version is equal or less than the last ratified version and optionally removes the entries for the ratified commits.
```
openinx, line 76: > equal or less than the last ratified version

Why will it be **less than the last ratified version** ? because of the concurrent committers ? and other commiter just advanced the version number since the lastest commit from this comitter ?

spec/protocols/ManagedTablesSpec.md
openinx: @yili-db  The client should send [metrics](#write-metrics) after each successful commit

Do we uc server already support the metric push at the oss-uc server side ?

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+- **Metrics and running maintenance operations**: The client should send [metrics](#write-metrics) after each successful commit. These metrics may be used for Unity Catalog to schedule and manage maintenance operations. The Unity Catalog follows the maintenance operation policy on Catalog-managed Tables. By default it disallows clients to run all maintenance operations, including but not limited to metadata cleanup, OPTIMIZE, VACUUM, and REORG.
+
+The rest of the specification outlines the APIs and detailed requirements.
+
+### Create a Staging Table
```
openinx, line 81: What's the difference betweent this part and the existing `createStagingTable` ?  https://github.com/unitycatalog/unitycatalog/blob/main/api/Apis/TablesApi.md#createstagingtable

And how could we align them if there is any one-side change ?  I will suggest to unify them into one place, then we don't need to maintain two copies of the same protocol, thoughts ?

spec/protocols/ManagedTablesSpec.md
openinx: > The type of managed storage path. Specifies what kind of storage location is being provisioned.<br/>For Unity Catalog Managed Tables, the value must to be `“MANAGED_TABLE_PATH”`

What are the other values besides the `MANAGED_TABLE_PATH` ?   If it's the only one value, then maybe we don't need this, as we can always set this at the server side.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+managed_storage_path_type | string | The type of managed storage path. Specifies what kind of storage location is being provisioned.<br/>For Unity Catalog Managed Tables, the value must to be `“MANAGED_TABLE_PATH”`. | required
+
+#### Responses
+
+**200: Request completed successfully.** It echoes all the fields in the request body and adds the additional following fields:
+
+Field Name | Data Type | Description | Optional/Required
+-|-|-|-
+id | string | Unique identifier generated for the staging table. This can be used to reference the staging table in subsequent operations.<br/>Type: UUID string. | required
+staging_location | string | URI generated for the staging table. This is the actual cloud storage location where data will be staged.<br/>Format: URI string (`s3://`, `abfss://`, `gs://`, etc.). | required
```
openinx, line 106: For oss-uc server, `file:///` is one of the format.  @yili-db , do we need to mention this ?

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+access_point | string | Access point associated with the staging location URI. For AWS, this would be an S3 access point ARN. | optional
```
openinx: > access_point

Is this only meaningful for aws ? what's the meaning for azure/gcp ?

openinx: Seems not a such intuitive field, which use the access_point to represent the aws ARN.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+
+#### Required Table Features and Properties on a Unity Catalog Managed Table
+
+*The list of features in this table are subject to change.*
+
+Feature Name | Table Feature | Table Properties | Description
+-|-|-|-
+[Catalog-Managed Tables](https://github.com/delta-io/delta/blob/master/protocol_rfcs/catalog-managed.md) | catalogManaged | `io.unitycatalog.tableId = <tableID>` | Required.
+[In-Commit Timestamps](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#in-commit-timestamps) | inCommitTimestamp | `delta.enableInCommitTimestamps = true` | Required for the correctness of time travel.
+[VACUUM Protocol Check](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#vacuum-protocol-check) | vacuumProtocolCheck | | Required for proper vacuum operations.
```
openinx, line 153: what's the example value for this vacuum field ?

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+  "catalog_name": "main",
+  "schema_name": "default",
+  "table_type": "MANAGED",
+  "data_source_format": "DELTA",
+  "storage_location": "s3://my-bucket/main/default/my_table/",
+  "columns": [
+    {
+      "name": "id",
+      "type_text": "bigint",
+      "type_json": "{\"name\":\"id\",\"type\":\"long\",\"nullable\":true,\"metadata\":{}}",
```
openinx, line 218: So this is a string, not a nested json object, rigtht ?

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+    "io.unitycatalog.tableId": "abcdef12-3456-7890-abcd-ef1234567890",
+    "delta.lastUpdateVersion": "0",
+    "delta.lastCommitTimestamp": "1704067400000"
+  },
+  "owner": "user@databricks.com",
+  "created_at": 1705600000000,
+  "created_by": "user@databricks.com",
+  "updated_at": 1705600000000,
+  "updated_by": "user@databricks.com",
+  "table_id": "abcdef12-3456-7890-abcd-ef1234567890"
```
openinx, line 307: No databricks mentioned, pls.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+created_by | string | Username of table creator. | required
+updated_at | int64 | Time at which this table was last modified, in epoch milliseconds. | required
+updated_by | string | Username of user who last modified the table. | optional
+table_id | string | Unique identifier for the table.<br>Type: UUID string. | required
+
+Request-Response Samples
+
+Request: `GET .../tables/main.default.my_table`
+```json
+{}
```
openinx: Need an empty json, or no need it at all, for getTable API.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+    "io.unitycatalog.tableId": "abcdef12-3456-7890-abcd-ef1234567890",
+    "delta.lastUpdateVersion": "0",
+    "delta.lastCommitTimestamp": "1704067400000"
+  },
+  "owner": "user@databricks.com",
+  "created_at": 1705600000000,
+  "created_by": "user@databricks.com",
+  "updated_at": 1705600000000,
+  "updated_by": "user@databricks.com",
+  "table_id": "abcdef12-3456-7890-abcd-ef1234567890"
```
openinx, line 421: same as above.

spec/protocols/ManagedTablesSpec.md
yili-db: I think that's OK. It says `etc` anyway. `file:///` is mostly used for testing.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+Auto-cleanup on drop | Yes | No
+Storage location | Assigned | User-specified
+
+
+## Terminology
+This section recaps key terminologies used in the catalog-managed tables specification:
+- **Commit**: A set of actions that moves a Delta table from version `v−1` to `v`, stored either as a Delta log file or inline in the catalog.
+- **Staged commit**: A commit written to `_delta_log/_staged_commits/<v>.<uuid>.json`, not yet guaranteed to be accepted; the catalog decides whether to ratify it.
+- **Ratified commit**: A proposed commit that the catalog has chosen as the winner for version `v`; it is part of the official history, whether or not it is already published.
+- **Published commit**: A ratified commit that has been written as `_delta_log/<v>.json`; the presence of this file proves that version `v` is ratified.
```
yili-db: "the presence of this file ... is ratified"? Or "is published"

spec/protocols/ManagedTablesSpec.md
yili-db: @openinx 
Should REPLACE also call `staging-table`? Or is it just another commit?
Here because it's an item below the `staging-table`, it sounds like CTAS RTAS REPLACE they all need to call `staging-table` anyway.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+For further details, refer to the [Catalog-Managed Tables](https://github.com/delta-io/delta/blob/master/protocol_rfcs/catalog-managed.md).
+
+## APIs
+A Unity Catalog Managed table needs to be registered and used according to the following specification:
+- **Create a table**: A Unity Catalog Managed table needs to first be created.
+  - The client sends a `POST` request to the [`staging-tables`](#create-a-staging-table) API to initiate the creation of a new Managed table in Unity Catalog.
+  - The server upon receiving the request to create a staging table, allocates a unique table ID and a path for the table.
+  - The client writes the first commit `0.json` under the `_delta_log` in the assigned table path. The `0.json` file should include all the table features and properties required by this specification. See the full list in [Required Table Features and Properties on a Unity Catalog Managed Table](#required-table-features-and-properties-on-a-unity-catalog-managed-table).
+    - For CTAS (Create Table As Select), the `0.json` file would also contain `AddFile` actions. Then the client finishes the table creation by sending a `POST` request to [`tables`](#create-a-table) API.
```
yili-db: "Then the client finishes the table creation by sending" This should be an item by itself and should not be associated with CTAS.

spec/protocols/ManagedTablesSpec.md
yili-db: Not yet

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+    - For REPLACE, do not drop and recreate the table. Instead, add a new commit that updates the table metadata and removes all existing data files. This approach ensures atomicity, preserves the table ID and history. The replaced table remains to be a Unity Catalog Managed Table.
+  - When a create table request is received, the server first confirms the existence of a table with the specified table ID at the provided location. If the table has the Catalog Managed feature enabled, the server then validates that all necessary table features and properties are correctly configured. Finally, the server registers the table as created at commit version 0.
+- **Get the table**: After the table is created successfully, the client can use the three-part name (`catalog_name.schema_name.table_name`) to retrieve the table entry.
+  - The client sends a `GET` call to the [`tables`](#get-a-table) API by providing the fully qualified three-part name.
```
yili-db: `tables` is the list table API.
Here it should be `tables/{full_name}`

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+  - After the commit is successful, the client must publish the ratified commit by copying the committed file to the `_delta_log` folder, and inform the server that the file has been published by sending another backfill commit, as a `POST` request to the `commit` API.
+  - The server verifies that the version is equal or less than the last ratified version and optionally removes the entries for the ratified commits.
+- **Metrics and running maintenance operations**: The client should send [metrics](#write-metrics) after each successful commit. These metrics may be used for Unity Catalog to schedule and manage maintenance operations. The Unity Catalog follows the maintenance operation policy on Catalog-managed Tables. By default it disallows clients to run all maintenance operations, including but not limited to metadata cleanup, OPTIMIZE, VACUUM, and REORG.
+
+The rest of the specification outlines the APIs and detailed requirements.
+
+### Create a Staging Table
+
+```
+POST .../staging-tables
```
yili-db, line 84: Is it intentional to omit the base URL?

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+&nbsp;&nbsp;columnInfo.type_text | string | Full data type specification as SQL/catalogString text. | optional
+&nbsp;&nbsp;columnInfo.type_json | string | Full data type specification, JSON-serialized. | optional
+&nbsp;&nbsp;columnInfo.type_name | string | The type of the column. One of BOOLEAN, BYTE, SHORT, INT, LONG, FLOAT, DOUBLE, DATE, TIMESTAMP, TIMESTAMP_NTZ, CHAR, STRING, BINARY, INTERVAL, ARRAY, STRUCT, MAP, DECIMAL, etc. | optional
+&nbsp;&nbsp;columnInfo.type_precision | int32 | Digits of precision; required for DecimalTypes | optional
+&nbsp;&nbsp;columnInfo.type_scale | int32 | Digits to the right of decimal; Required for DecimalTypes. | optional
+&nbsp;&nbsp;columnInfo.type_internal_type | string | Format of IntervalType. | optional
+&nbsp;&nbsp;columnInfo.position | int32 | Ordinal position of column (starting at position 0). | optional
+&nbsp;&nbsp;columnInfo.nullable | boolean | Whether the field may be Null. Default to true. | optional
+&nbsp;&nbsp;columnInfo.partition_index | int32 | Partition index for column. | optional
+storage_location | string | Storage root URL for table. | required
```
yili-db: "For Unity Catalog Managed Tables, this must be the `staging_location` of staging table."

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+
+Request: `GET .../tables/main.default.my_table`
+```json
+{}
+```
+
+Response: 200
+
+```json
+{
```
yili-db, line 371: This is essentially duplicating the previous response of create table.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+&nbsp;&nbsp;commitInfo.version | int64 | The table version to commit. | required
+&nbsp;&nbsp;commitInfo.timestamp | int64 | In-commit timestamp in milliseconds since epoch. | required
+&nbsp;&nbsp;commitInfo.file_name | string | UUID-based commit filename. | required
+&nbsp;&nbsp;commitInfo.file_size | int64 | Size of the commit file in bytes. | required
+&nbsp;&nbsp;commitInfo.file_modification_timestamp | int64 | Filesystem modification time in milliseconds since epoch. | required
+latest_published_version | int64 | Notifies UC that commits up to this version have been published. Constraint: At least one of `commit_info` or `latest_published_version` must be present. | optional
+metadata | Metadata | Delta metadata action if the commit changes table metadata. The posted metadata object is the desired final state, so all existing data that needs to be preserved must be included as well as the fields that changed. Any properties returned from Unity Catalog should be included. When to include: Schema changes, property updates, comment changes. Requires `commit_info`. | optional
+&nbsp;&nbsp;metadata.id | string | The internal table ID created by Delta. | required
```
yili-db: What does it mean to have id in metadata?
If it means to change table id, it isn't possible to do so.
If it means the current table id, it's already specified in top level as `table_id`

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+&nbsp;&nbsp;metadata.name | string | The table name. | optional
```
yili-db: Is this for table renaming? But renaming table shouldn't involve a commit, right? Table name isn't part of table metadata anyway.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+&nbsp;&nbsp;metadata.description | string | The table comment. | optional
+&nbsp;&nbsp;metadata.format | Format | Object that contains format information. | optional
```
yili-db: What is this used for? `REPLACE` a table using another format? I don't know if it's possible to do so.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+&nbsp;&nbsp;&nbsp;&nbsp;metadata.format.provider | string | The format of the files, e.g. "parquet". | optional
+&nbsp;&nbsp;&nbsp;&nbsp;metadata.format.options | object | Format-specific options as key-value pairs. | optional
+&nbsp;&nbsp;metadata.schema | array of ColumnInfos | The table schema defined by an array of `ColumnInfo` objects. | optional
+&nbsp;&nbsp;metadata.partition_columns | array of strings | The partition columns. | optional
```
yili-db: What's the relationship between `columnInfo.partition_index` and `partition_columns`?

spec/protocols/ManagedTablesSpec.md
yili-db: What is server supposed to do with it? It doesn't seem to matter and server doesn't return it back to client in get-commits.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+&nbsp;&nbsp;metadata.properties | object | Table properties as key-value pairs. When updating, always includes all the required [properties](#required-table-features-and-properties-on-a-unity-catalog-managed-table). | optional
+&nbsp;&nbsp;metadata.created_time | long | Timestamp when metadata was created in milliseconds since epoch
+uniform | Uniform | UniForm conversion information related to this Delta commit. | optional
+&nbsp;&nbsp;uniform.iceberg | Iceberg | Uniform Iceberg info. | optional
+&nbsp;&nbsp;&nbsp;&nbsp;uniform.iceberg.metadata_location | string | Iceberg metadata location converted up to the Delta version. | required
+&nbsp;&nbsp;&nbsp;&nbsp;uniform.iceberg.converted_delta_version | int64 | The Delta version under conversion. | required
+&nbsp;&nbsp;&nbsp;&nbsp;uniform.iceberg.converted_delta_timestamp | string | Timestamp of the conversion in ISO 8601 timestamp which is 27 characters. | required
+
```
yili-db, line 588: Also base_converted_delta_version
https://github.com/unitycatalog/unitycatalog/commit/1e4f7fdf290b2c3f792035fe7fb1c8e54a9fd952#diff-8d03742f9785228675d14f784a7c0affcc9400e8571603bde6ede194c052d985R3100

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+        "format": {
+          "provider": "parquet",
+          "options": {}
+        },
+        "schema": [
+          {"name": "id", "type": "long", "nullable": false},
+          {"name": "name", "type": "string", "nullable": true},
+          {"name": "created_at", "type": "timestamp", "nullable": false}
+        ],
+       "properties": {
```
yili-db:
```
"properties": {
  "properties": {
     ...
  }
}
```

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,826 @@
...
+-|-|-
+TABLE_DOES_NOT_EXIST | 404 | The specified table doesn't exist.<br>The client should check if the table ID is correct for this table.
+PERMISSION_DENIED | 403 | Insufficient permissions (SELECT on table).<br>The client should check with their Unity Catalog Admin to grant them the necessary rights to write commits.
+INVALID_PARAMETER_VALUE | 400 | Server side check failed to post metrics for Unity Catalog Managed Tables.<br>The client has provided wrong inputs to the server. This could include but not limited to the following errors:(1) missing required fields, (2) provided fields are invalid such as non-positive numbers, etc.
+
+## API Reference
+
+Method | HTTP request | Description
+-|-|-
+[**createStagingTable**](../../apis/Apis/TablesApi.md#createstagingtable) | **POST** /staging-tables | Create a Staging Table
```
yili-db: apis/Apis...  --> api/Apis... (This is the correct path)

spec/protocols/ManagedTablesSpec.md
zhenlineo: On purpose left.

spec/protocols/ManagedTablesSpec.md
zhenlineo: This is for REPLACE only. We've decided to keep the log.

spec/protocols/ManagedTablesSpec.md
zhenlineo: I added the first time it showed. Then I skipped for all following references.

spec/protocols/ManagedTablesSpec.md
zhenlineo: Backfill can be old. let say we write 5, we can only backfilled to 4. It is not ensured to always to be 5. as 5 might failed.

spec/protocols/ManagedTablesSpec.md
zhenlineo: That's the API doc, this is a spec. They should be the same.

spec/protocols/ManagedTablesSpec.md
zhenlineo: This is already part of the API.

spec/protocols/ManagedTablesSpec.md
zhenlineo: It just has no more properties, it only has a feature name.

spec/protocols/ManagedTablesSpec.md
zhenlineo: It is a different id used by delta.

spec/protocols/ManagedTablesSpec.md
zhenlineo: It is part of delta logs.

spec/protocols/ManagedTablesSpec.md
zhenlineo: To my best understanding, `partition_index` is the index of the col in `partition_columns`.

spec/protocols/ManagedTablesSpec.md
zhenlineo: Why we want to have two layers?

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,813 @@
...
+`delta.lastUpdateVersion = 0` | Indicates the last version at which the metadata was updated.
+`delta.lastCommitTimestamp = t` | `t` is the timestamp of commit 0, in milliseconds since epoch.
+`clusteringColumns` | If the [Clustered Table Feature](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#clustered-table) is supported. Column paths are serialized as a JSON array of string arrays. Each string array represents a path from the root schema to a column, with each element being a field name. For example, top-level column `id` and nested column `address.city` serialize to `[["id"], ["address", "city"]]`.
+
+#### Request body
+
+Field Name | Data Type | Description | Optional/Required
+-|-|-|-
+name | string | Name of the staging table, relative to parent schema. | required
+catalog_name | string | Name of the parent catalog where the staging table will be created. | required
```
tdas, line 174: [AI Generated] The field descriptions here say "Name of the staging table" and "parent catalog where the staging table will be created." This is the Create Table API — by this point the table is being promoted from staging to a real table. These descriptions are copy-pasted from the Create Staging Table API and should say "table" or "managed table", not "staging table." Same issue in the Get Table response fields (lines 345-346).

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,813 @@
...
+
+1. Propose a simple staged commit
+
+    Request:
+    ```json
+    {
+      "table_id": "abcdef12-3456-7890-abcd-ef1234567890",
+      "table_uri": "s3://my-bucket/main/default/my_table",
+      "commit_info": {
+        "version": 42,
```
tdas, line 603: [AI Generated] The `file_name` values in the samples are inconsistent with each other and with the Terminology section:
- Write Commit samples use `"v.uuid1.json"` — looks like a placeholder, not a realistic filename.
- Get Commits samples use `"00000000-0000-0000-0000-00000000002a.json"` — a pure UUID with no version prefix.
- The Terminology section defines the format as `<version>.<uuid>.json` (e.g. `00000000000000000042.3a0d65cd-4056-49b8-937b-95f9e3ee90e5.json`).

The samples should use consistent, realistic filenames matching the defined format so implementers know exactly what to expect.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,813 @@
...
+&nbsp;&nbsp;commitInfo.file_name | string | UUID-based commit filename. | required
+&nbsp;&nbsp;commitInfo.file_size | int64 | Size of the commit file in bytes. | required
+&nbsp;&nbsp;commitInfo.file_modification_timestamp | int64 | Filesystem modification time in milliseconds since epoch. | required
+latest_published_version | int64 | Notifies Unity Catalog that commits up to this version have been published. Constraint: At least one of `commit_info` or `latest_published_version` must be present. | optional
+metadata | Metadata | Delta metadata action if the commit changes table metadata. The posted `Metadata` object is the desired final state, so all existing data that needs to be preserved must be included as well as the fields that changed. Any properties returned from Unity Catalog should be included. When to include: Schema changes, property updates, comment changes. Requires `commit_info`. | optional
+&nbsp;&nbsp;metadata.description | string | The table comment. | optional
+&nbsp;&nbsp;metadata.schema | Columns | The table schema defined by a `Columns` object. | optional
```
tdas, line 578: [AI Generated] `metadata.schema` (line 584) and `columns` in Create Table (line 184) are both defined as "array of ColumnInfo objects", but they use completely different formats:
- Create Table sample uses UC ColumnInfo: `{"name": "id", "type_text": "bigint", "type_json": "...", "type_name": "LONG", "position": 0, ...}`
- Write Commit sample uses Delta-style: `{"name": "id", "type": "long", "nullable": false}`

Within the same spec, "ColumnInfo" means two different schemas depending on the API. If the intent is different formats, that should be explicit. If it's the same format, the Write Commit sample needs to be updated.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,813 @@
...
+GET .../delta/commits
+```
+Retrieves all ratified commits from Unity Catalog.  
+Note: Server implementations may choose to paginate responses. A paginated response is indicated by the highest returned version in the commits array being less than `latest_table_version`. In this case, the client should send another request with an adjusted `start_version` to retrieve any remaining commits.
+
+#### Request body
+
+Field Name | Data Type | Description | Optional/Required
+-|-|-|-
+table_id | string | Uniquely identifies the Delta table in Unity Catalog.<br>Type: UUID string. | required
```
tdas, line 448: [AI Generated] Both `table_id` and `table_uri` are required here and in Write Commit. But the server already stores `table_uri` when the table is created — it's derivable from `table_id` alone. Having both required means a client could provide inconsistent values. Is the server expected to validate they match? If so, there should be an error code for mismatch. If `table_uri` is for optimization (avoiding a server-side lookup), consider making it optional.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -0,0 +1,813 @@
...
+- **Get the table**: After the table is created successfully, the client can use the three-part name (`catalog_name.schema_name.table_name`) to retrieve the table entry.
+  - The client sends a `GET` call to the [`getTable`](#get-a-table) API by providing the fully qualified three-part name.
+  - The server then responds with detailed information of the table entry, including the table ID, path etc.
+- **Read from the table**: To read from the table, the client needs to obtain all necessary commits to reconstruct the snapshot:
+  - The client obtains all the published commits by listing all the files in `_delta_log`, as well as to fetch the ratified commits from the server by performing a `GET` call to the [`getCommits`](#get-commits) API.
+  - The server returns all ratified commits for the table. The server can optionally decide to paginate the response.
+    - For a newly created table, the server should return `0` when no commits has been made to this table.
```
tdas, line 70: [AI Generated] nit: Two grammar fixes:
- Line 76: "when no commits **has** been made" → "when no commits **have** been made"
- Line 81: "After all the commits till this version **has** been published" → "After all the commits till this version **have** been published"

tdas (commented):
[AI Generated] Latest revision looks much better — good cleanup on the metadata fields, staging table simplification, and the version 0 / maintenance exceptions clarifications. A few new issues below.

tdas:
The last round of comments can be addressed in the follow up PR. They are relatively minor. Let me merge this specification. Thank you for doing this @zhenlineo !!!!

tdas:
@openinx @yili-db please feel free to leave more comments. and @zhenlineo please address these comments in a follow up.

PR #1292, opened 2026-01-07
Add comprehensive authorization for storage credential CRUD operations

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1292/files) to review incremental changes.
- [**stack/sc_crud_perm**](https://github.com/unitycatalog/unitycatalog/pull/1292) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1292/files)]
  - [stack/sc_delete_force](https://github.com/unitycatalog/unitycatalog/pull/1293) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1293/files/53cb8a1eeb17e2531415215fefc31629678f2ec1..ed53e49b1dd722d4eadd4226644e7c840776c460)]

---------
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**
Add comprehensive authorization for storage credential CRUD operations
This commit implements fine-grained access control for storage credential operations, moving from a simple metastore owner check to a role-based permission system.

- Added CREATE_STORAGE_CREDENTIAL privilege to the privilege enum
- This allows granting users permission to create storage credentials without requiring full metastore ownership

Implemented authorization rules for all CRUD operations:

**CREATE**:
- Metastore owners can create credentials (existing behavior)
- Users with CREATE_STORAGE_CREDENTIAL on metastore can create credentials
- Note: Service credentials and CREATE_SERVICE_CREDENTIAL are not supported

**GET/LIST**:
- Metastore owners can view all credentials (existing behavior)
- Credential owners can view their own credentials
- Users with any permission on a credential can view it
- List operation uses post-filtering to return only authorized credentials

**UPDATE/DELETE**:
- Metastore owners can update/delete any credential
- Credential owners can update/delete their own credentials

PR #1344, opened 2026-02-06
Implement authorization filtering for LIST operations

yili-db:
**PR Checklist**

**Description of changes**
Implement authorization filtering for LIST operations

This commit introduces a new authorization filtering mechanism for LIST operations that filters response items based on user permissions.
Previously, LIST operations can not have the auth expression annotated in the method. Instead, they had to resolve the user principal and resource IDs then call `evaluator.filter` from each function body.

Key Changes:

1. New `ResultFilter` class
   - Filters list responses by evaluating authorization expressions for each item
   - Tracks whether filter was called to enable security enforcement

2. New `@ResponseAuthorizeFilter` annotation
   - Marks service methods that return filtered lists
   - Required for any LIST operation that needs item-level filtering

3. Enhanced `UnityAccessDecorator`
   - Creates ResultFilter instance from authorization parameters
   - Now enforces that all methods must have authorization annotated

4. New `applyResponseFilter()` method in `AuthorizedService`
   - Convenience method for services to apply filtering
   - Services can extend `AuthorizedService` and call this method

Addresses https://github.com/unitycatalog/unitycatalog/pull/1292/changes/BASE..788af87ef37e40581437cc4d7904be6197fde3ac#r2677860675

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/auth/decorator/ResultFilter.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/decorator/ResultFilter.java
@@ -1,0 +1,195 @@
+package io.unitycatalog.server.auth.decorator;
+
+import io.unitycatalog.server.auth.annotation.ResponseAuthorizeFilter;
+import io.unitycatalog.server.model.CatalogInfo;
+import io.unitycatalog.server.model.CredentialInfo;
+import io.unitycatalog.server.model.ExternalLocationInfo;
+import io.unitycatalog.server.model.FunctionInfo;
+import io.unitycatalog.server.model.RegisteredModelInfo;
+import io.unitycatalog.server.model.SchemaInfo;
+import io.unitycatalog.server.model.SecurableType;
+import io.unitycatalog.server.model.TableInfo;
+import io.unitycatalog.server.model.VolumeInfo;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import io.unitycatalog.server.service.AuthorizedService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static io.unitycatalog.server.model.SecurableType.CATALOG;
+import static io.unitycatalog.server.model.SecurableType.REGISTERED_MODEL;
+import static io.unitycatalog.server.model.SecurableType.SCHEMA;
+
+/**
+ * Filters list responses based on user authorization permissions.
+ *
+ * <p>This class is used in the authorization decorator layer to filter LIST operation results. When
+ * a service method is annotated with {@code @ResponseAuthorizeFilter}, the {@link
+ * UnityAccessDecorator} creates a ResultFilter instance and stores it in the request context. The
+ * service method then retrieves this filter and calls it to remove items that the user is not
+ * authorized to access.
+ *
+ * <p><b>How It Works:</b>
+ *
+ * <ol>
+ *   <li>UnityAccessDecorator intercepts HTTP requests and extracts authorization parameters
+ *       (principal ID, expression, resource IDs) from method annotations
+ *   <li>It creates a ResultFilter with these parameters and stores it in the request context
+ *   <li>The service method retrieves the filter via {@code applyResponseFilter()} and applies it to
+ *       the response list
+ *   <li>The filter evaluates authorization for each item and removes unauthorized items
+ * </ol>
+ *
+ * <p><b>Resource ID Resolution:</b><br>
+ * For each item in the list, the filter resolves resource IDs of items but expect the catalog and
+ * schema IDs to be provided in the request already. Special handling exists for REGISTERED_MODEL
+ * because models can be listed without specifying catalog/schema filters. In this case, the filter
+ * extracts the catalog and schema names from each model and resolves them to IDs.
+ *
+ * <p><b>Security:</b><br>
+ * The {@code wasCalled()} method allows UnityAccessDecorator to verify that the filter was actually
+ * used. If a method is annotated with {@code @ResponseAuthorizeFilter} but doesn't call the filter,
+ * a security exception is thrown to prevent data leakage.
+ *
+ * @see ResponseAuthorizeFilter
+ * @see UnityAccessDecorator
+ * @see AuthorizedService#applyResponseFilter
+ */
+public class ResultFilter {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ResultFilter.class);
+
+  private final UUID principalId;
+  private final String expression;
+  private final UnityAccessEvaluator evaluator;
+  private final Map<SecurableType, UUID> resourceIds;
+  private final Map<String, Object> nonResourceValues;
+  private final KeyMapper keyMapper;
+  private boolean called = false;
+
+  public ResultFilter(
+      UnityAccessEvaluator evaluator,
+      UUID principalId,
+      String expression,
+      Map<SecurableType, UUID> resourceIds,
+      Map<String, Object> nonResourceValues,
+      KeyMapper keyMapper) {
+    this.principalId = principalId;
+    this.expression = expression;
+    this.evaluator = evaluator;
+    this.resourceIds = resourceIds;
+    this.nonResourceValues = nonResourceValues;
+    this.keyMapper = keyMapper;
+  }
+
+  /**
+   * Filters a list of resources based on the user's authorization permissions.
+   *
+   * <p>This method evaluates the authorization expression for each item in the list and removes
+   * items that the user is not authorized to access. For each item, it:
+   *
+   * <ol>
+   *   <li>Resolves the item's resource ID (e.g., table ID, volume ID)
+   *   <li>Resolves parent (catalog and schema) resource IDs if it's registered model
+   *   <li>Evaluates the authorization expression with all resolved and provided IDs
+   *   <li>Removes the item if the expression evaluates to false
+   * </ol>
+   *
+   * <p>If an error occurs during authorization evaluation for an item, that item is filtered out
+   * (removed from the list) as a security precaution.
+   *
+   * @param securableType The type of resources being filtered (TABLE, VOLUME, FUNCTION, etc.)
+   * @param items The list of items to filter in-place. This list is modified by removing
+   *     unauthorized items.
+   * @param <T> The type of items in the list (e.g., TableInfo, VolumeInfo)
+   * @throws RuntimeException if the securableType is already resolved in the pre-resolved IDs,
+   *     which indicates a programming error
+   */
+  public <T> void filter(SecurableType securableType, List<T> items) {
+    called = true;
+
+    if (items == null || items.isEmpty()) {
+      return;
+    }
+
+    LOGGER.debug("Filtering {} items with authorization expression", items.size());
+    if (resourceIds.containsKey(securableType)) {
+      // This simply indicates a bug in code.
+      throw new RuntimeException("Securable type " + securableType + " is already resolved");
+    }
+
+    items.removeIf(
+        item -> {
+          try {
+            // Resolve the item's resource ID and any parent resource IDs to make a complete
+            // collection of all needed resource IDs for this item.
+            Map<SecurableType, UUID> resourceIdsForItem =
+                resolveResourceIdsForItem(securableType, item, resourceIds);
+            boolean authorized =
+                evaluator.evaluate(principalId, expression, resourceIdsForItem, nonResourceValues);
+            if (!authorized) {
+              LOGGER.debug("Item filtered out: {}", item.getClass().getSimpleName());
+            }
+            return !authorized;
+          } catch (Exception e) {
+            LOGGER.warn(
+                "Error evaluating authorization for item, filtering out: {}", e.getMessage());
+            return true;
+          }
+        });
+
+    LOGGER.debug("After filtering: {} items remain", items.size());
+  }
+
+  private Map<SecurableType, UUID> resolveResourceIdsForItem(
+      SecurableType securableType, Object item, Map<SecurableType, UUID> preResolvedIds) {
+    // First, resolve the item's own resource ID
+    UUID itemId = resolveResourceId(securableType, item);
+    Map<SecurableType, UUID> combined = new HashMap<>(preResolvedIds);
+    combined.put(securableType, itemId);
+    // For REGISTERED_MODEL, resolve catalog and schema if both are not present.
+    if (securableType == REGISTERED_MODEL
+        && !preResolvedIds.containsKey(CATALOG)
+        && !preResolvedIds.containsKey(SCHEMA)) {
+      RegisteredModelInfo model = (RegisteredModelInfo) item;
+      combined.putAll(
+          keyMapper.mapResourceKeys(
+              Map.of(CATALOG, model.getCatalogName(), SCHEMA, model.getSchemaName())));
+    }
+    // For everything else, the catalog and schema IDs are already included in preResolvedIds
+    // if needed.
+    return combined;
+  }
+
+  private UUID resolveResourceId(SecurableType securableType, Object item) {
+    String id = switch (securableType) {
+      case TABLE -> ((TableInfo)item).getTableId();
+      case VOLUME -> ((VolumeInfo)item).getVolumeId();
+      case FUNCTION -> ((FunctionInfo)item).getFunctionId();
+      case REGISTERED_MODEL -> ((RegisteredModelInfo)item).getId();
+      case CATALOG -> ((CatalogInfo)item).getId();
+      case SCHEMA -> ((SchemaInfo)item).getSchemaId();
+      case CREDENTIAL -> ((CredentialInfo)item).getId();
+      case EXTERNAL_LOCATION -> ((ExternalLocationInfo)item).getId();
+      default -> throw new RuntimeException("Unsupported securable type: " + securableType);
+    };
+    return UUID.fromString(id);
+  }
+
+  /**
+   * Returns whether the filter has been called.
+   *
+   * <p>This method is used by {@link UnityAccessDecorator} to verify that methods annotated with
+   * {@code @ResponseAuthorizeFilter} actually call the filter on successful responses. This is a
+   * security enforcement mechanism to prevent data leakage when developers forget to filter
+   * results.
+   *
+   * @return true if {@link #filter} has been called at least once, false otherwise
+   */
+  public boolean wasCalled() {
+    return called;
+  }
+}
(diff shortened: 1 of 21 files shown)


server/src/main/java/io/unitycatalog/server/service/MetastoreService.java
```diff
@@ -16,6 +17,7 @@ public MetastoreService(Repositories repositories) {
   }
 
   @Get("/metastore_summary")
+  @AuthorizeExpression("#principal != null")
```
tdas, line 20: [AI Generated] This `@AuthorizeExpression` is never evaluated. There are no `@AuthorizeResourceKey` annotations on this method, so `locators` is empty in `UnityAccessDecorator.serve()`. When locators is empty, the decorator logs a warning and falls through to `delegate.serve(ctx, req)` without evaluating the expression. This endpoint is effectively unprotected.

Either add `@AuthorizeResourceKey(METASTORE)` here, or handle the empty-locators case in the decorator to still evaluate the expression.

server/src/main/java/io/unitycatalog/server/auth/decorator/ResultFilter.java
```diff
@@ -0,0 +1,195 @@
...
+public class ResultFilter {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ResultFilter.class);
+
+  private final UUID principalId;
+  private final String expression;
+  private final UnityAccessEvaluator evaluator;
+  private final Map<SecurableType, UUID> resourceIds;
+  private final Map<String, Object> nonResourceValues;
+  private final KeyMapper keyMapper;
+  private boolean called = false;
```
tdas: [AI Generated] This field needs to be `volatile` (or `AtomicBoolean`). `filter()` writes `called = true` during service execution, and `wasCalled()` reads it in `afterRequest()` via a `thenApply` callback that may execute on a different event loop thread. Without a memory visibility guarantee, `wasCalled()` could read `false` even after `filter()` was called, causing false security violation errors.

```suggestion
  private volatile boolean called = false;
```

server/src/main/java/io/unitycatalog/server/service/AuthorizedService.java
```diff
@@ -69,4 +77,29 @@ protected void removeHierarchicalAuthorizations(String resourceId, String parent
...
+   *
+   * @param securableType The type of resources being filtered (TABLE, VOLUME, FUNCTION, etc.)
+   * @param items The list of items to filter. The list is modified in-place by removing
+   *     unauthorized items.
+   * @param <T> The type of items in the list
+   */
+  protected <T> void applyResponseFilter(SecurableType securableType, List<T> items) {
+    if (serverProperties.isAuthorizationEnabled()) {
+      ResultFilter resultFilter =
+          ServiceRequestContext.current().attr(UnityAccessDecorator.RESULT_FILTER_ATTR);
```
tdas, line 103: [AI Generated] `resultFilter` will be `null` if `RESULT_FILTER_ATTR` was not set in the request context (e.g., if the code path doesn't go through the decorator). This will NPE on the next line. Add a null check with a clear error message:

```java
ResultFilter resultFilter =
    ServiceRequestContext.current().attr(UnityAccessDecorator.RESULT_FILTER_ATTR);
if (resultFilter == null) {
    throw new BaseException(ErrorCode.INTERNAL,
        "Authorization filter not initialized — ensure the request goes through UnityAccessDecorator.");
}
resultFilter.filter(securableType, items);
```

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -200,14 +213,107 @@ private static Object findPayloadValue(String key, Map<String, Object> payload)
...
+    }
+
+    abstract void evaluateBeforeRequest(
+        UUID principal,
+        String expression,
+        Map<SecurableType, UUID> resourceIds,
+        Map<String, Object> nonResourceValues);
+
+    HttpResponse afterRequest(HttpResponse response) {
+      return HttpResponse.of(response.aggregate().thenApply(
```
tdas: [AI Generated] `response.aggregate()` buffers the entire response in memory for **every** request, including `RequestEvaluationAction` where `evaluateAfterRequest()` is a no-op. This adds unnecessary latency and memory pressure for all non-LIST operations (single-item GET, DELETE, POST, etc.).

`RequestEvaluationAction` should override `afterRequest` to pass the response through directly:

```java
// In RequestEvaluationAction:
@Override
HttpResponse afterRequest(HttpResponse response) {
    return response;
}
```

The `beforeRequestWasCalled` safety check can be moved into `beforeRequest` or the `serve` method itself for the non-filter path.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -196,35 +205,27 @@ public LoadViewResponse loadView(
...
-                catalog,
-                namespace,
-                Optional.of(Integer.MAX_VALUE),
-                Optional.empty(),
-                Optional.empty(),
-                Optional.empty())
-            .aggregate()
-            .join();
+    ListTablesResponse tables =
+        tableRepository.listTables(
```
tdas, line 223: [AI Generated] The old code called `tableService.listTables()` which applied authorization filtering. This now calls `tableRepository.listTables()` directly, bypassing the filter entirely. While the current `METASTORE OWNER` gate restricts access to admins, if this placeholder authorization is relaxed later (as noted in the PR description), this becomes an unfiltered data leak. At minimum, add a TODO documenting this risk.

tdas (commented):
[AI Generated]

## Summary

This PR centralizes authorization filtering for LIST operations into the decorator layer, replacing per-service `filterXxx` methods with a shared `ResultFilter` mechanism. The removal of duplicated filtering logic and the `#defer` concept are welcome improvements.

However, there are correctness issues (MetastoreService auth bypass, thread-safety), a performance concern (unnecessary response buffering for all requests), and critically, **no tests** for this security-sensitive refactoring.

See inline comments for specifics. This is an **auth/security domain change** — needs domain expert sign-off.

### Needs Tests (Blocking)

There are zero test changes in this PR. This is a security-critical refactoring of the authorization layer. At minimum:
- `ResultFilter`: authorized items kept, unauthorized removed, errored items removed, empty/null lists, `wasCalled()` enforcement.
- `UnityAccessDecorator`: `RequestEvaluationAction` vs `ResultFilterAction` paths, `beforeRequestWasCalled` safety net.
- Integration: verify LIST endpoints filter correctly end-to-end with the new mechanism.
- Branch coverage: `resolveResourceIdsForItem` has special `REGISTERED_MODEL` logic — test listing models with and without catalog/schema filters.

### PR Scope

The title says "Implement authorization filtering for LIST operations" but the PR also adds `@AuthorizeExpression` to all Iceberg endpoints (production behavior change), removes `#defer` support, changes `KeyMapper` return type, and removes the `evaluate()` overload. The Iceberg auth changes in particular feel like a separate PR — the description calls them "placeholder" and "work-in-progress" but they change production behavior.

server/src/main/java/io/unitycatalog/server/auth/decorator/ResultFilter.java
yili-db: Good catch. Fixed.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
yili-db: Good point. Refactored the class hierarchy — `afterRequest` is now abstract. `RequestEvaluationAction` passes the response through directly since auth is fully evaluated before the request. `ResultFilterAction` keeps the buffering since it needs to verify the filter was actually called. Also moved `beforeRequestWasCalled`, `method`, and the safety checks out of the base class and into `ResultFilterAction` where they belong. Then replaced `beforeRequestWasCalled` with a `resultFilter == null` check since it's equivalent and eliminates redundant state.

server/src/main/java/io/unitycatalog/server/service/AuthorizedService.java
yili-db: Fixed. Added a null check that throws `BaseException(ErrorCode.INTERNAL, ...)` with a clear message.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
yili-db: The old code called `tableService.listTables()` as a direct Java method call, not an HTTP request. `UnityAccessDecorator` is an HTTP decorator — it only intercepts external HTTP requests. So `@ResponseAuthorizeFilter` on `TableService.listTables()` was never triggered and `RESULT_FILTER_ATTR` was never set for this context. The old code didn't actually exercise table-level authorization filtering either. No regression here.

server/src/main/java/io/unitycatalog/server/service/MetastoreService.java
yili-db: Good catch. Removed the `if (!locators.isEmpty())` guard in the decorator and let empty locators flow through `authorizeByRequest` normally — all three locator sublists end up empty, `resourceKeys` stays `{}`, and the expression gets evaluated via the synchronous path. No special case needed. Also found and fixed a related bug: `AuthService.logout()` had no `@AuthorizeExpression`, which caused a 500 when the decorator hit the `expression == null` throw. Added `@AuthorizeExpression("#principal != null")` there.

yili-db:
This PR is a refactor of the filtering mechanism and does not change authorization behavior. Existing integration tests cover the full auth flow end-to-end for all resource types: `SdkCatalogAccessControlCRUDTest`, `SdkSchemaAccessControlCRUDTest`, `SdkTableAccessControlCRUDTest`, `SdkFunctionAccessControlCRUDTest`, `SdkVolumeAccessControlCRUDTest`, `SdkModelAccessControlCRUDTest`, `SdkCredentialAccessControlTest`, and `SdkExternalLocationAccessControlTest`. These exercise the decorator → ResultFilter → filtered response path and validate the refactored code. All 144 tests pass.

yili-db:
@tdas Also I have split out some unrelated changes to https://github.com/unitycatalog/unitycatalog/pull/1429. Please take a look at that one first. Thank you!

yili-db:
This PR has been rebased. It no longer contains Iceberg auth changes.

yili-db:
@tdas Please take another look. Thank you!

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -200,14 +206,87 @@ private static Object findPayloadValue(String key, Map<String, Object> payload)
     }
   }
 
-  private void checkAuthorization(
-      UUID principal,
-      String expression,
-      Map<SecurableType, Object> resourceKeys,
-      Map<String, Object> nonResourceValues) {
-    LOGGER.debug("resourceKeys = {}", resourceKeys);
+  private abstract static class AbstractEvaluationAction {
```
openinx: nit: seems like this can be an interface ?

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -200,14 +206,87 @@ private static Object findPayloadValue(String key, Map<String, Object> payload)
...
+      ctx.setAttr(RESULT_FILTER_ATTR, resultFilter);
+    }
+
+    @Override
+    HttpResponse afterRequest(HttpResponse response) {
+      return HttpResponse.of(response.aggregate().thenApply(
+        aggregated -> {
+          if (aggregated.status().isSuccess()) {
+            if (resultFilter == null) {
+              LOGGER.error(
```
openinx, line 276: if we already throw the exception , then seems like we don't need to log here.

openinx: since the server will always print the stacktrace in the entrypoint , right ?

yili-db: The critical info `method.getName()` is redacted from the exception. It's meant for admins/developers to debug a software bug.
By the time it reaches here, the execution of the method is already done and returned so stack trace shows nothing.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -200,14 +206,87 @@ private static Object findPayloadValue(String key, Map<String, Object> payload)
...
+        Map<String, Object> nonResourceValues) {
+      resultFilter =
+          new ResultFilter(
+              evaluator, principal, expression, resourceIds, nonResourceValues, keyMapper);
+      ctx.setAttr(RESULT_FILTER_ATTR, resultFilter);
+    }
+
+    @Override
+    HttpResponse afterRequest(HttpResponse response) {
+      return HttpResponse.of(response.aggregate().thenApply(
```
openinx: **Performance: all LIST responses are fully buffered in memory.**

`response.aggregate()` buffers the entire HTTP response body just to check `wasCalled()`. But the filtering (`applyResponseFilter`) already runs synchronously in the handler before `HttpResponse.ofJson()` is called — by the time `afterRequest` runs, the list is already filtered. The buffer only serves the enforcement check. For large `listTables` results this allocates O(response size) per concurrent request.

Consider setting a `CompletableFuture<Boolean>` flag when `applyResponseFilter` is called and checking that in `afterRequest` without aggregating the response body.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -200,14 +206,87 @@ private static Object findPayloadValue(String key, Map<String, Object> payload)
...
+        aggregated -> {
+          if (aggregated.status().isSuccess()) {
+            if (resultFilter == null) {
```
openinx, line 274: **This null check is unreachable in practice.**

`resultFilter` is set in `beforeRequest`. For GET/LIST endpoints (no PAYLOAD locators), `beforeRequest` runs synchronously before `delegate.serve()`, so `resultFilter` is always non-null when `afterRequest` executes.

Also the error message `"did not execute evaluateAction"` is a framework-internal term that won't help a developer understand what they forgot to do. If this check is kept as a safety net, the message should say something like `"did not call applyResponseFilter()"`.

server/src/main/java/io/unitycatalog/server/auth/decorator/ResultFilter.java
```diff
@@ -0,0 +1,195 @@
...
+public class ResultFilter {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ResultFilter.class);
+
+  private final UUID principalId;
+  private final String expression;
+  private final UnityAccessEvaluator evaluator;
+  private final Map<SecurableType, UUID> resourceIds;
+  private final Map<String, Object> nonResourceValues;
+  private final KeyMapper keyMapper;
+  private volatile boolean called = false;
```
openinx: `volatile` provides visibility but `AtomicBoolean` would make the intent clearer. The happens-before is implicitly provided by the `CompletableFuture` chain in `afterRequest`, but that's subtle. `AtomicBoolean` is self-documenting for a flag that is set once and checked once across threads.

server/src/main/java/io/unitycatalog/server/auth/decorator/ResultFilter.java
```diff
@@ -0,0 +1,195 @@
...
+    UUID itemId = resolveResourceId(securableType, item);
+    Map<SecurableType, UUID> combined = new HashMap<>(preResolvedIds);
+    combined.put(securableType, itemId);
+    // For REGISTERED_MODEL, resolve catalog and schema if both are not present.
+    if (securableType == REGISTERED_MODEL
+        && !preResolvedIds.containsKey(CATALOG)
+        && !preResolvedIds.containsKey(SCHEMA)) {
+      RegisteredModelInfo model = (RegisteredModelInfo) item;
+      combined.putAll(
+          keyMapper.mapResourceKeys(
```
openinx, line 161: **N+1 DB lookups when listing all models without a catalog/schema filter.**

When `listRegisteredModels` is called without `catalog_name`/`schema_name`, this does 2 DB lookups (catalog + schema) per model in the result. The old `filterModels` had the same pattern so this is unchanged behavior — but it is an existing N+1 that is now more visible here.

If `catalog_name`/`schema_name` are provided, wiring them as `@AuthorizeResourceKey` on `listRegisteredModels` would pre-resolve the IDs once and let `preResolvedIds.containsKey(CATALOG)` skip this block entirely for the common filtered case.

openinx (commented):
This PR eliminates ~200 lines of per-service `filterXxx()` boilerplate and centralizes item-level authorization filtering into the decorator layer. The direction is correct — replacing `#defer` with an active enforcement mechanism (`wasCalled()`) is a meaningful security improvement.

A few issues worth addressing before merging, detailed in the inline comments below.

server/src/main/java/io/unitycatalog/server/auth/decorator/ResultFilter.java
yili-db: Yes that's right. This PR does not aim to optimize for the N+1 query when listing models with no catalog and schema specified.
`preResolvedIds` is a simple optimization that is easy to do here. But the "list models from all catalogs and all schemas" is just complex to optimize and it's a very special use case only for models.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
yili-db: Yes it's a safety net check. If it happens, it's not because `resultFilter` was not called. But because of `evaluateAction.beforeRequest` was not called.

`evaluateAction.beforeRequest` is not always called synchronously before delegate.serve(). It can be called inside the lamda of `peekData`. There's no simple guarantee that it must be called. I even have another PR https://github.com/unitycatalog/unitycatalog/pull/1494 to extra guard this.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
yili-db: Changed to only wait for the headers to finish to avoid buffering the entire body.
`CompletableFuture<Boolean>` can't work here because we still need to consider the case where it fails before calling filter function. `response.split().headers().thenApply(` is the best alternative.

tdas (approved):

PR #1354, opened 2026-02-11
[Server] getCommits returns latestTableVersion=0 instead of -1 for newly created managed Delta tables

TimothyW553:
**PR Checklist**

**Description of changes**
  - Fix `DeltaCommitRepository.getCommits()` to return `latestTableVersion=0` instead of `-1` for newly created managed Delta tables with no commits
  - All tables reaching `getCommits()` are already validated as managed Delta tables (via `validateTable()`), so UC is the source of truth and version 0 is the correct initial state

**Test plan**
  - Updated existing assertion in testBasicCoordinatedCommitsCRUD from expectedLatestTableVersion = -1 to expectedLatestTableVersion = 0

### DIFF
--- a/spec/protocols/ManagedTablesSpec.md
+++ b/spec/protocols/ManagedTablesSpec.md
@@ -1,0 +1,813 @@
+# Unity Catalog Managed Tables Specification
+
+> **Status**: Proposed  
+> **Version**: 0.1.0  
+> **Last Updated**: 2026-02-17
+
+## Table of Contents
+
+- [Introduction](#introduction)
+- [Terminology](#terminology)
+- [APIs](#apis)
+  - [Create a Staging Table](#create-a-staging-table)
+  - [Create a Table](#create-a-table)
+  - [Get a Table](#get-a-table)
+  - [Get Commits](#get-commits)
+  - [Write a Commit](#write-a-commit)
+  - [Write Metrics](#write-metrics)
+---
+
+## Introduction
+This specification defines how Unity Catalog will serve as the managing catalog for [Catalog-Managed Tables](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#catalog-managed-tables).
+Catalog-Managed Tables is a new Delta table feature which makes the catalog the source of truth for commits to a table.
+It defines a managing catalog to perform commits according to the Catalog-Managed Tables specification.
+
+This specification defines the APIs to enable the following operations between the server (Unity Catalog) and clients:
+- Creating a Catalog-Managed table with necessary table features and properties.
+- Retrieving and reading from the table.
+- Writing to the table and publishing commits.
+- Running maintenance operations.
+
+### What is a Unity Catalog Managed Table?
+A Unity Catalog managed table is a table whose data location and lifecycle (including deletion of files when dropped)
+are fully controlled by the Unity Catalog.
+It is different from external tables, where the catalog only stores metadata pointing to data in an externally managed
+location and does not delete the underlying files when the table is dropped.
+
+Comparison with External Tables:
+
+Feature | Managed Table | External Table
+-|-|-
+Storage ownership | Unity Catalog | User
+Auto-cleanup on drop | Yes | No
+Storage location | Assigned | User-specified
+
+
+## Terminology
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -166,9 +166,9 @@ public DeltaGetCommitsResponse getCommits(DeltaGetCommits rpc) {
           MAX_NUM_COMMITS_PER_TABLE);
     }
     if (commitCount == 0) {
-      // No commit exist yet. Not even any backfilled one. That means the table has never sent any
-      // commit coordinated by UC.
-      return new DeltaGetCommitsResponse().latestTableVersion(-1L);
+      // Table is validated as a managed Delta table. UC is the source of truth for
+      // managed tables, so a newly created table with no commits is at version 0.
+      return new DeltaGetCommitsResponse().latestTableVersion(0L);
```
murali-db, line 175: what are the downstream implications of this? are there any clients currently relying on the -1 being returned for new tables?

timothyw553: Currently Delta Kernel's UCCatalogManagedClient breaks on -1 

If you take a look here: https://github.com/delta-io/delta/blob/master/kernel/unitycatalog/src/main/java/io/delta/kernel/unitycatalog/UCCatalogManagedClient.java#L129
```java
  final long maxUcTableVersion = response.getLatestTableVersion(); // -1                                                                                                  
  snapshotBuilder.withMaxCatalogVersion(maxUcTableVersion);        // throws!
```
`withMaxCatalogVersion(-1)` throws `IllegalArgumentException: A valid version must be >= 0`. So returning -1 for a newly created table actively breaks kernel-based snapshot loading in OSS UC. Returning 0 fixes this.

api/Models/DeltaGetCommitsResponse.md
```diff
@@ -4,7 +4,7 @@
 | Name | Type | Description | Notes |
 |------------ | ------------- | ------------- | -------------|
 | **commits** | [**List**](DeltaCommitInfo.md) | The list of unbackfilled Delta table commits. Can be in arbitrary order. | [default to null] |
-| **latest\_table\_version** | **Long** | Represents the latest version of the table tracked by UC. If no commits have occurred via UC yet,  then UC cannot determine the latest version and returns -1. Use this field to manage pagination —  if the returned commits don&#39;t cover the range up to latest_table_version or end_version (whichever is smaller),  it indicates that more unbackfilled commits may be available.  | [default to null] |
+| **latest\_table\_version** | **Long** | Represents the latest version of the table tracked by UC. For a newly created managed table with no commits, this returns 0. Use this field to manage pagination — if the returned commits don&#39;t cover the range up to latest_table_version or end_version (whichever is smaller), it indicates that more unbackfilled commits may be available.  | [default to null] |
```
openinx, line 7: So this sentence was actually wrong, right ?  I mean the actual newly created table version is 0, rather than -1 , even before this change ?

openinx: So seems like you changed the bavavior from -1 to 0,  okay. ( just looked the following code change).

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
openinx: Yes,same question, as @murali-db  . 

Do we need to create a testing PR in the delta repo, so that it could pull this PR and deploy it locally, and trigger all the tests, and see if any break ?  I did the same before for CTAS test. 
https://github.com/delta-io/delta/pull/6002/changes/BASE..beaa5cb3cf277cb80c1b168652467834da2b5455

tdas: we should consider having master build set up in each repo using the master build of the other repo. 
that way we can catch issues earlier than later

tdas (approved):

murali-db (approved):

PR #1355, opened 2026-02-13
Allow setting transaction isolation level in getCommits to avoid getting incomplete commit list.

yili-db:
**PR Checklist**

**Description of changes**

Allow setting transaction isolation level in getCommits to avoid getting incomplete commit list.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
@@ -118,7 +118,11 @@ public class DeltaCommitRepository {
           return query.list();
         },
         "Failed to get latest commits",
-        /* readOnly= */ true);
+        /* readOnly= */ true,
+        // Use REPEATABLE_READ isolation to ensure consistent snapshot and prevent version numbers
+        // from appearing to go backwards during concurrent writes. This is critical for get
+        // commits that expect monotonic version progression after posting a commit.
+        Optional.of(java.sql.Connection.TRANSACTION_REPEATABLE_READ));
   }
 
   /**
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -118,7 +118,11 @@ private List<DeltaCommitDAO> getAllCommitDAOsDesc(UUID tableId) {
           return query.list();
         },
         "Failed to get latest commits",
-        /* readOnly= */ true);
+        /* readOnly= */ true,
+        // Use REPEATABLE_READ isolation to ensure consistent snapshot and prevent version numbers
+        // from appearing to go backwards during concurrent writes. This is critical for get
+        // commits that expect monotonic version progression after posting a commit.
+        Optional.of(java.sql.Connection.TRANSACTION_REPEATABLE_READ));
   }
```
tdas, line 127: any test for this?

tdas: @yili-db ??

tdas: also.. this description does not explain the problem. i dont understand whats the problem this is solving.

yili-db: Sorry I missed this comment. This bug was found by @TimothyW553 during manual testing using Delta Spark by repeatedly committing to UC and get the commit version back.
Getting commits can get transiently decreasing version numbers after committing a newer version and getting that newer version number, due to the weak transaction consistency level causing H2 to return stale result.
This PR fixes that.
However it's a time consuming and non-deterministic test to repro this bug. So it's not included in this PR.

Issue #1364, opened 2026-02-22
Postgres schema init fails for use of BINARY


tcarland:
**Describe the bug**

Use of Postgres as the database causes an issue with Hibernate's DDL for the `uc_delta_commits` table, due to the column definition annotation setting the column type as BINARY, which is not supported in Postgresdb.

```
org.hibernate.tool.schema.spi.CommandAcceptanceException: Error executing DDL "create table uc_delta_commits (id BINARY(16) not null, commit_file_modification_timestamp timestamp(6) not null, commit_filename varchar(255) not null, commit_filesize bigint not null, commit_timestamp timestamp(6) not null, commit_version bigint not null, is_backfilled_latest_commit boolean not null, table_id BINARY(16) not null, primary key (id))" via JDBC [ERROR: type "binary" does not exist
```

**To Reproduce**

Deploy unitycatalog server (tag v0.4.0) using postgresdb.

**Expected behavior**

Column type should use `bytea` for UUID columns under PG

wahahawasabi:
This was fixed in #1446 (commit 94bcba67) by @tiennguyen-onehouse — `columnDefinition="BINARY(16)"` removed from `DeltaCommitDAO.java`, letting Hibernate 6.5 auto-map `UUID` per dialect. Fix is on `main`. Closing.

PR #1365, opened 2026-02-22
drop mysql specific column identifier to allow hibernate to choose.

tcarland:
**PR Checklist**

**Description of changes**

Drop Column identifier defining the column type as BINARY as this is only supported by MySQL. Instead, by not defining, Hibernate will choose either `binary` for MySQL or `bytea` for PG.

Issue #1364 

This threw an exception in the log as described in the Issue. Tested that the uc server initializes properly and without error.


yili-db:
This was fixed in https://github.com/unitycatalog/unitycatalog/pull/1446 (commit https://github.com/unitycatalog/unitycatalog/commit/94bcba6735e3463f5d5312301ac36ff404047714) by @tiennguyen-onehouse

PR #1379, opened 2026-03-05
Change ModelRepository to use TransactionManager.executeWithTransaction

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1379/files) to review incremental changes.
- [**stack/model_repo_tx**](https://github.com/unitycatalog/unitycatalog/pull/1379) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1379/files)]
  - [stack/concurrent_tx](https://github.com/unitycatalog/unitycatalog/pull/1380) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1380/files/c00cc26683498b2efa08c5b9bda322b00e4f3611..f4609279fd2bb6415f2e66382dc308f2a3994cab)]

---------
**PR Checklist**

**Description of changes**

Change ModelRepository to use TransactionManager.executeWithTransaction
Also fixes bugs:
- Wrong query parameter type in getModelVersionDao
- Wrong log message in createModelVersion
- TransactionManager.executeWithTransaction to not lose stack trace of cause

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
@@ -19,6 +19,7 @@ import io.unitycatalog.server.persist.utils.ExternalLocationUtils;
 import io.unitycatalog.server.persist.utils.FileOperations;
 import io.unitycatalog.server.persist.utils.PagedListingHelper;
 import io.unitycatalog.server.persist.utils.RepositoryUtils;
+import io.unitycatalog.server.persist.utils.TransactionManager;
 import io.unitycatalog.server.utils.IdentityUtils;
 import io.unitycatalog.server.utils.NormalizedURL;
 import io.unitycatalog.server.utils.ServerProperties;
@@ -30,7 +31,6 @@ import java.util.Optional;
 import java.util.UUID;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
-import org.hibernate.Transaction;
 import org.hibernate.query.Query;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -89,7 +89,7 @@ public class ModelRepository {
         "FROM ModelVersionInfoDAO t WHERE t.registeredModelId = :registeredModelId AND t.version = :version";
     Query<ModelVersionInfoDAO> query = session.createQuery(hql, ModelVersionInfoDAO.class);
     query.setParameter("registeredModelId", modelId);
-    query.setParameter("version", version.toString());
+    query.setParameter("version", version);
     query.setMaxResults(1);
     LOGGER.info("Finding model version by registeredModelId: {} and version: {}", modelId, version);
     return query.uniqueResult(); // Returns null if no result is found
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -89,7 +89,7 @@ public ModelVersionInfoDAO getModelVersionDao(Session session, UUID modelId, Lon
         "FROM ModelVersionInfoDAO t WHERE t.registeredModelId = :registeredModelId AND t.version = :version";
     Query<ModelVersionInfoDAO> query = session.createQuery(hql, ModelVersionInfoDAO.class);
     query.setParameter("registeredModelId", modelId);
-    query.setParameter("version", version.toString());
+    query.setParameter("version", version);
```
tdas, line 92: [AI Generated]

This changes behavior: `version` is a `Long`, and the old code was passing `version.toString()` (a `String`) to the Hibernate query parameter. The DAO field `ModelVersionInfoDAO.version` is `Long`, so this fix is correct — but it is a **bug fix**, not a refactor.

Per repo convention, refactor PRs should not contain logic changes. Can you either call this out explicitly in the PR description as an intentional fix, or separate it?

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -540,65 +483,41 @@ public ModelVersionInfo createModelVersion(CreateModelVersion createModelVersion
...
-        if (tx != null && tx.getStatus().canRollback()) {
-          try {
-            // For now, never delete.  We will implement a soft delete later.
-            // UriUtils.deleteStorageLocationPath(storageLocation);
-          } catch (Exception deleteErr) {
-            LOGGER.error(
-                "Unable to delete storage location {} during rollback: {}",
-                storageLocation,
-                deleteErr.getMessage());
+    LOGGER.info("Creating Model Version for: {}", registeredModelFullName);
```
tdas, line 486: [AI Generated]

Good catch — the old code was logging "Creating Registered Model" inside `createModelVersion`, which was clearly wrong. But same note as the `version.toString()` fix: this is a bug fix inside a refactor PR. Worth calling out in the description.

tdas (commented):
[AI Generated]

Overall this is a clean refactor that brings `ModelRepository` in line with how `TableRepository`, `CatalogRepository`, and `SchemaRepository` already use `TransactionManager`. The `readOnly` flags are correct and the lambda captures are handled properly. Left a few comments.

**Exception cause chain is lost in `TransactionManager`**

This is the most impactful issue across all repositories using `TransactionManager`, not just this PR. The `TransactionManager` catch block uses the 2-arg `BaseException` constructor:

```java
throw new BaseException(ErrorCode.INTERNAL, errorMessage + ": " + e.getMessage());
```

The old `ModelRepository` code used the 3-arg constructor which preserves the original exception as a cause:

```java
throw new BaseException(ErrorCode.INTERNAL, "Error creating registered model: " + fullName, e);
```

Without the chained cause, production stack traces for non-`BaseException` failures (Hibernate constraint violations, network errors, etc.) lose the original stack trace, making debugging harder. Suggest fixing `TransactionManager` to use `new BaseException(ErrorCode.INTERNAL, errorMessage + ": " + e.getMessage(), e)`. Could be a preceding PR in the stack.

**Tests**

No test files were added or modified. There are existing tests (`SdkModelCRUDTest`, `BaseModelCRUDTest`). Can you confirm they pass with these changes? In particular, the `version.toString()` to `version` change (see inline comment) modifies Hibernate query parameter binding — if no existing test exercises `getModelVersionDao`, this needs coverage.

**PR description**

The description says *what* was changed but not *why*. From the stacked PR #1380 it is clear this is prerequisite work for concurrent transaction support. Would be helpful to state that in the description — what happens without this change?

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
yili-db: Updated description.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
yili-db: Updated description

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
yili-db: It was tested by BaseModelCRUDTest.testModelCRUD but H2 seems to work fine with mismatching types.

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -209,125 +202,98 @@ public RegisteredModelInfo createRegisteredModel(CreateRegisteredModel createReg
...
+          registeredModelInfo.setStorageLocation(storageLocation.toString());
+          RegisteredModelInfoDAO registeredModelInfoDAO =
+              RegisteredModelInfoDAO.from(registeredModelInfo);
+          registeredModelInfoDAO.setSchemaId(schemaInfoDAO.getId());
+          registeredModelInfoDAO.setMaxVersionNumber(0L);
+          session.persist(registeredModelInfoDAO);
+          FileOperations.createStorageLocationDir(storageLocation);
+          return registeredModelInfo;
+        },
+        "Failed to create registered model",
```
tdas, line 232: The old code included the entity name in the INTERNAL error message:
```java
"Error creating registered model: " + fullName
```
The new message loses that context. Same issue in `updateRegisteredModel` (line 377), `updateModelVersion` (line 622), and `finalizeModelVersion` (line 706).

Meanwhile `createModelVersion` (line 519) *does* include context: `"Failed to create model version for model: " + registeredModelFullName`.

Since the `TransactionManager` wraps non-`BaseException` errors with this message, it's the primary string that shows up in logs and API responses for unexpected failures. Consider adding the entity identifier back for consistency, e.g.:
```java
"Failed to create registered model: " + fullName
```

server/src/main/java/io/unitycatalog/server/persist/ModelRepository.java
```diff
@@ -540,65 +483,41 @@ public ModelVersionInfo createModelVersion(CreateModelVersion createModelVersion
...
-          "Error creating model version for model: " + registeredModelFullName,
-          e);
-    }
-    return modelVersionInfo;
+          UUID modelId = existingRegisteredModel.getId();
+          Long version = existingRegisteredModel.getMaxVersionNumber() + 1;
+          NormalizedURL storageLocation =
+              ExternalLocationUtils.getManagedLocationForModelVersion(
+                  NormalizedURL.from(existingRegisteredModel.getUrl()), modelVersionId);
+          modelVersionInfo.setVersion(version);
```
tdas, line 509: nit: `modelVersionInfo` is created before the lambda (line 471) and mutated here inside it. This is correct today since `TransactionManager` doesn't retry, but if the stacked PR #1380 introduces retry semantics for concurrent transactions, these mutations (`setVersion`, `setStorageLocation`) would be applied multiple times on the same object. Worth keeping in mind — or alternatively, construct the `ModelVersionInfo` return value fresh inside the lambda after the persist succeeds, similar to how `updateModelVersion` and `finalizeModelVersion` build from the DAO.

tdas (commented):
Follow-up review — the TransactionManager cause-preservation fix and description update look good. Two remaining observations below.

PR #1380, opened 2026-03-05
Prevent concurrent duplicate writes with serializable isolation and unique constraints

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1380/files/c00cc26683498b2efa08c5b9bda322b00e4f3611..f4609279fd2bb6415f2e66382dc308f2a3994cab) to review incremental changes.
- [stack/model_repo_tx](https://github.com/unitycatalog/unitycatalog/pull/1379) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1379/files)]
  - [**stack/concurrent_tx**](https://github.com/unitycatalog/unitycatalog/pull/1380) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1380/files/c00cc26683498b2efa08c5b9bda322b00e4f3611..f4609279fd2bb6415f2e66382dc308f2a3994cab)]

---------
**PR Checklist**

**Description of changes**
Prevent concurrent duplicate writes with serializable isolation and unique constraints

TransactionManager now uses `REPEATABLE_READ` for reads and `SERIALIZABLE` for writes, retrying up to 3 times on serialization conflicts or unique-constraint violations. Added `@UniqueConstraint` to all entity DAOs as a safety net for databases without SSI (e.g. H2). Added a concurrent create-table test to verify correctness.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/dao/CatalogInfoDAO.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/dao/CatalogInfoDAO.java
@@ -4,6 +4,7 @@ import io.unitycatalog.server.model.CatalogInfo;
 import jakarta.persistence.Column;
 import jakarta.persistence.Entity;
 import jakarta.persistence.Table;
+import jakarta.persistence.UniqueConstraint;
 import java.time.Instant;
 import java.util.Date;
 import java.util.UUID;
@@ -15,7 +16,13 @@ import lombok.Setter;
 import lombok.experimental.SuperBuilder;
 
 @Entity
-@Table(name = "uc_catalogs")
+@Table(
+    name = "uc_catalogs",
+    uniqueConstraints = {
+      @UniqueConstraint(
+          name = "uk_catalog_name",
+          columnNames = {"name"})
+    })
 // Lombok
 @Getter
 @Setter
--- a/server/src/main/java/io/unitycatalog/server/persist/dao/CredentialDAO.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/dao/CredentialDAO.java
@@ -17,6 +17,7 @@ import jakarta.persistence.EnumType;
 import jakarta.persistence.Enumerated;
 import jakarta.persistence.Lob;
 import jakarta.persistence.Table;
+import jakarta.persistence.UniqueConstraint;
 import java.util.Date;
 import java.util.Optional;
 import java.util.UUID;
@@ -37,7 +38,13 @@ import lombok.experimental.SuperBuilder;
  * @see AwsCredentialVendor
  */
 @Entity
-@Table(name = "uc_credentials")
+@Table(
+    name = "uc_credentials",
+    uniqueConstraints = {
+      @UniqueConstraint(
+          name = "uk_credential_name",
+          columnNames = {"name"}),
+    })
 // Lombok
 @Getter
 @Setter
--- a/server/src/main/java/io/unitycatalog/server/persist/utils/TransactionManager.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/utils/TransactionManager.java
@@ -2,16 +2,21 @@ package io.unitycatalog.server.persist.utils;
 
 import io.unitycatalog.server.exception.BaseException;
 import io.unitycatalog.server.exception.ErrorCode;
-import java.util.Optional;
+import java.sql.Connection;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.Transaction;
+import org.hibernate.exception.ConstraintViolationException;
+import org.hibernate.exception.LockAcquisitionException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /** Utility class for managing database transactions. */
 public class TransactionManager {
   private static final Logger LOGGER = LoggerFactory.getLogger(TransactionManager.class);
+
+  /** Maximum number of times a transaction is tried (including the first attempt). */
+  private static final int MAX_TRANSACTION_TRY_COUNT = 3;
 
   /**
    * Functional interface for database operations that return a result.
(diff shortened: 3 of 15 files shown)


server/src/main/java/io/unitycatalog/server/persist/dao/CredentialDAO.java
```diff
@@ -37,7 +38,13 @@
  * @see AwsCredentialVendor
  */
 @Entity
-@Table(name = "uc_credentials")
+@Table(
```
tdas, line 41: what will be the impact of this change when user migrate from their backend databases to use this new DAO

server/src/main/java/io/unitycatalog/server/persist/utils/TransactionManager.java
```diff
@@ -26,77 +31,91 @@ public interface DatabaseOperation<R> {
...
-      boolean readOnly,
-      Optional<Integer> isolationLevel) {
-    try (Session session = sessionFactory.openSession()) {
-      if (readOnly) {
-        session.setDefaultReadOnly(true);
-      }
+      boolean readOnly) {
+    int isolationLevel =
+        readOnly ? Connection.TRANSACTION_REPEATABLE_READ : Connection.TRANSACTION_SERIALIZABLE;
+    for (int attempt = 0; attempt < MAX_TRANSACTION_TRY_COUNT; attempt++) {
```
tdas, line 69: [AI Generated] Why blanket `SERIALIZABLE` for ALL write transactions? The previous code only used custom isolation for the specific case that needed it (`getAllCommitDAOsDesc`). Now every write across the entire server — catalogs, schemas, tables, volumes, functions, credentials, external locations, models, users (13+ repository callers of `executeWithTransaction`) — will run at `SERIALIZABLE`.

On production databases like PostgreSQL, `SERIALIZABLE` incurs additional overhead: SSI tracking, higher abort rates under concurrency, and increased retry loops. Has this been measured? Was `REPEATABLE_READ` + unique constraints considered as a less aggressive alternative for writes?

server/src/main/java/io/unitycatalog/server/persist/utils/TransactionManager.java
```diff
@@ -26,77 +31,91 @@ public interface DatabaseOperation<R> {
...
-      } finally {
-        // Restore original isolation level to avoid polluting connection pool
-        if (isolationLevel.isPresent()) {
+        Transaction tx = session.beginTransaction();
+        try {
+          R result = operation.execute(session);
+          tx.commit();
+          return result;
+        } catch (LockAcquisitionException | ConstraintViolationException e) {
+          tx.rollback();
```
tdas, line 89: [AI Generated] Retrying on _any_ `ConstraintViolationException` seems too broad. The intent is to catch concurrent-duplicate-insert races so the retry re-runs the business logic and produces a proper `ALREADY_EXISTS` error. But `ConstraintViolationException` also fires for other constraint violations (e.g., NOT NULL, foreign key, or a hypothetical primary key UUID collision). In those cases, retrying will just fail again 3 times with the same error, wasting time and then returning a misleading "serialization conflict" message instead of the real cause.

Should this be more selective — e.g., only retry on unique-constraint violations, or at least log the constraint name so operators can distinguish race-condition retries from genuine bugs?

server/src/main/java/io/unitycatalog/server/persist/utils/TransactionManager.java
```diff
@@ -26,77 +31,91 @@ public interface DatabaseOperation<R> {
...
+        } finally {
+          // Restore original isolation level to avoid polluting connection pool
           session.doWork(
               connection -> {
                 connection.setTransactionIsolation(originalIsolation[0]);
               });
         }
       }
     }
+    throw new BaseException(
```
tdas, line 118: [AI Generated] The original exception is lost here. When all retries are exhausted, the last `LockAcquisitionException` or `ConstraintViolationException` is discarded and replaced with a generic message. This makes debugging production issues very difficult — there's no stack trace showing what the actual database conflict was.

Please track the last exception through the retry loop and pass it as the cause:
```java
throw new BaseException(
    ErrorCode.INTERNAL,
    errorMessage + ": serialization conflict, max retries exceeded",
    lastException);
```

server/src/main/java/io/unitycatalog/server/persist/dao/CatalogInfoDAO.java
```diff
@@ -15,7 +16,13 @@
 import lombok.experimental.SuperBuilder;
 
 @Entity
-@Table(name = "uc_catalogs")
+@Table(
+    name = "uc_catalogs",
+    uniqueConstraints = {
```
tdas, line 21: [AI Generated] Adding `@UniqueConstraint` across all 10 entity DAOs (catalogs, schemas, tables, volumes, functions, credentials, external locations, registered models, model versions, users) is a DDL change. If Hibernate's `hbm2ddl.auto` is set to `update`, it will attempt to add these constraints on server startup.

If a production database already has duplicate rows (which is exactly the race condition this PR is trying to prevent), `ALTER TABLE ADD CONSTRAINT` will **fail** and the server won't start. What is the migration path? Is there a migration script to deduplicate existing rows first? This needs to be documented or addressed before merge.

Also: for `ModelVersionInfoDAO`, the existing `@Index(name = "idx_model_version")` is being _replaced_ with a `@UniqueConstraint`. Does Hibernate drop the old index automatically, or will there be a stale index left behind?

server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableCRUDTest.java
```diff
@@ -369,6 +372,99 @@ public void testStagingTableCreationWithNonExistentSchemaShouldFail() throws Exc
...
+    // what the HTTP layer reported.
+    List<TableInfoDAO> dbRows = findAllTablesBySchemaAndName(tableName);
+    // Primary assertion: the DB must always have exactly one row for this table name.
+    assertThat(dbRows).hasSize(1);
+
+    // Secondary assertion: exactly one HTTP request should have succeeded.
+    long success = workers.stream().filter(x -> x.result != null).count();
+    assertThat(success).isEqualTo(1);
+
+    // The failing thread must have received ALREADY_EXISTS (409).
```
tdas: [AI Generated] This assertion can vacuously pass. If both workers succeed (the exact bug this PR prevents), `filter(x -> x.result == null)` produces an empty stream, and `forEach` on an empty stream is a no-op — no assertion is ever executed, so the test passes silently.

Add an explicit assertion on the failure count before this block:
```java
long failures = workers.stream().filter(x -> x.error != null).count();
assertThat(failures).isEqualTo(1);
```

tdas (commented):
[AI Generated] See inline comments. Summary of concerns:

**Blocking:**
- Adding `@UniqueConstraint` to all entity DAOs is a DDL change that can break server startup on upgrade if duplicate rows exist. Migration path needs to be addressed.
- Lost root cause exception on retry exhaustion in `TransactionManager`.

**Questions:**
- Why blanket `SERIALIZABLE` for all writes rather than targeted use? Performance impact on all 13+ repository callers?
- Retrying on all `ConstraintViolationException` types seems too broad.

**Test gaps:**
- Unique constraints added to 10 entity DAOs but only tables are tested for concurrent creation. At minimum, verify constraints work for a couple other entity types.
- The SSI retry path (`LockAcquisitionException`) is untested since H2 doesn't implement SSI.
- The failure assertion in the concurrent test can vacuously pass (see inline comment).

**Positive:** The `TransactionManager` Javadoc explaining the isolation level rationale is excellent. The `ModelRepository` refactor to remove manual try/catch/rollback boilerplate is a good cleanup.

PR #1382, opened 2026-03-05
Fix type_text and type_name in UCSingleCatalog

yili-db:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

Fix type_text and type_name in UCSingleCatalog

PR #1383, opened 2026-03-06
Set partition index of columns in UCProxy

yili-db:
**PR Checklist**

**Description of changes**

Set partition index of columns in UCProxy. This is required by Delta to send table schema to UC.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -499,15 +499,19 @@ private class UCProxy(
     }
     createTable.setStorageLocation(storageLocation)
 
-    val partitionColNames: Seq[String] = partitions
-      .filter(_.name() == "identity")
-      .map { t =>
-        val fieldNames = t.references().flatMap(_.fieldNames())
-        assert(fieldNames.length == 1,
-          s"Expected single-field partition reference but got: ${fieldNames.mkString(".")}")
-        fieldNames.head
+    val partitionColNames: Seq[String] = partitions.flatMap { t =>
+      t.name() match {
+        case "identity" =>
+          val fieldNames = t.references().flatMap(_.fieldNames())
+          require(fieldNames.length == 1,
+            s"Expected single-field partition reference but got: ${fieldNames.mkString(".")}")
+          Some(fieldNames.head)
+        case "cluster_by" =>
+          None
+        case other =>
+          throw new ApiException(s"Unsupported partition transform: $other")
       }
-      .toSeq
+    }.toSeq
     val columns: Seq[ColumnInfo] = schema.fields.toSeq.zipWithIndex.map { case (field, i) =>
       val column = new ColumnInfo()
       column.setName(field.name)
@@ -519,7 +523,7 @@ private class UCProxy(
       column.setTypeName(convertDataTypeToTypeName(field.dataType))
       column.setTypeJson(field.dataType.json)
       column.setPosition(i)
-      val partitionIdx = partitionColNames.indexOf(field.name)
+      val partitionIdx = partitionColNames.indexWhere(_.equalsIgnoreCase(field.name))
       if (partitionIdx >= 0) column.setPartitionIndex(partitionIdx)
       column
     }
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
@@ -441,9 +441,9 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
   }
 
   /**
-   * Specification for one column in testTableWithComplexTypes: its SQL DDL type, the SQL literal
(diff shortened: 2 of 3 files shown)


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -510,6 +519,8 @@ private class UCProxy(
       column.setTypeName(convertDataTypeToTypeName(field.dataType))
       column.setTypeJson(field.dataType.json)
       column.setPosition(i)
+      val partitionIdx = partitionColNames.indexOf(field.name)
+      if (partitionIdx >= 0) column.setPartitionIndex(partitionIdx)
```
tdas, line 527: same concern on testing as the previous PR #1382 ... is this delta-only code? and if not, can we not test this using parquet?

yili-db: Yes. Added a simple check in ExternalTableReadWriteTest.java

tdas (commented):
Thanks for adding the test — that addresses my earlier concern.

A few more things to consider:

1. **Replace `assert` with `require` or an explicit throw**: The `assert(fieldNames.length == 1, ...)` can be elided at runtime with `-Xdisable-assertions`. Since this is a production code path, use `require(...)` or throw an explicit exception instead.

2. **Silent drop of non-identity partitions**: The code filters for `_.name() == "identity"` and quietly ignores other transform types (`bucket`, `truncate`, `years`, etc.). If a user creates a table with e.g. `PARTITIONED BY (years(ts))`, the partition info is silently lost. Consider logging a warning or throwing an error to make the limitation explicit.

3. **Case sensitivity on column name matching**: `partitionColNames.indexOf(field.name)` uses case-sensitive comparison. Spark is typically case-insensitive with column names. Likely fine if Spark normalizes names before passing them here, but worth verifying.

4. **Test coverage**: The test only covers a single partition column. A test with multiple partition columns would validate that the ordering (`partitionIndex` 0, 1, 2...) is correct.

Item 1 is the main actionable one. The rest are nice-to-haves.

yili-db:
> Thanks for adding the test — that addresses my earlier concern.
> 
> A few more things to consider:
> 
> 1. **Replace `assert` with `require` or an explicit throw**: The `assert(fieldNames.length == 1, ...)` can be elided at runtime with `-Xdisable-assertions`. Since this is a production code path, use `require(...)` or throw an explicit exception instead.
> 2. **Silent drop of non-identity partitions**: The code filters for `_.name() == "identity"` and quietly ignores other transform types (`bucket`, `truncate`, `years`, etc.). If a user creates a table with e.g. `PARTITIONED BY (years(ts))`, the partition info is silently lost. Consider logging a warning or throwing an error to make the limitation explicit.
> 3. **Case sensitivity on column name matching**: `partitionColNames.indexOf(field.name)` uses case-sensitive comparison. Spark is typically case-insensitive with column names. Likely fine if Spark normalizes names before passing them here, but worth verifying.
> 4. **Test coverage**: The test only covers a single partition column. A test with multiple partition columns would validate that the ordering (`partitionIndex` 0, 1, 2...) is correct.
> 
> Item 1 is the main actionable one. The rest are nice-to-haves.

1. Fixed
2. Only identity partitions can reach here as a valid partition. cluster by is intentionally skpped. Others are rejected by Delta: https://github.com/delta-io/delta/blob/60d4535bb9ee8d14400ca20986eef5261598b26f/spark/src/main/scala/org/apache/spark/sql/delta/catalog/AbstractDeltaCatalog.scala#L511. Updated the code to throw error for unexpected transform.
3. Fixed.
4. Added extra test

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -606,6 +622,7 @@ public void testTableWithSupportedDataTypes() throws ApiException {
 
     session = createSparkSessionWithCatalogs(SPARK_CATALOG, CATALOG_NAME);
     String tableName = TEST_TABLE + "_complex_type";
+    List<String> partitionColumns = List.of("col_daytime_interval", "col_string", "col_bigint");
```
yili-db, line 625: Multiple out-of-order partition columns added

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -615,13 +632,16 @@ public void testTableWithSupportedDataTypes() throws ApiException {
                 .setColumns(
                     cols.stream()
                         .map(c -> Pair.of(c.getName(), c.getSqlType()))
-                        .collect(Collectors.toList())));
+                        .collect(Collectors.toList()))
+                .setPartitionColumns(partitionColumns));
+    String colNames = cols.stream().map(ColSpec::getName).collect(Collectors.joining(", "));
     sql(
-        "INSERT INTO %s VALUES (%s)",
+        "INSERT INTO %s (%s) VALUES (%s)",
```
yili-db, line 639: Query must specify full list of columns after adding the partitions.

PR #1384, opened 2026-03-07
Update ManagedTablesSpec to clarify that storage ownership is always …

michelleon:
…user-owned

**PR Checklist**

**Description of changes**

Update ManagedTablesSpec docs to clarify that storage ownership is always user-owned

### DIFF
--- a/spec/protocols/ManagedTablesSpec.md
+++ b/spec/protocols/ManagedTablesSpec.md
@@ -38,7 +38,7 @@ Comparison with External Tables:
 
 Feature | Managed Table | External Table
 -|-|-
-Storage ownership | Unity Catalog | User
+Storage ownership | User | User
 Auto-cleanup on drop | Yes | No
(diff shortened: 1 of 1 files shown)


spec/protocols/ManagedTablesSpec.md
```diff
@@ -38,7 +38,7 @@ Comparison with External Tables:
 
 Feature | Managed Table | External Table
 -|-|-
-Storage ownership | Unity Catalog | User
+Storage ownership | User | User
```
tdas, line 41: If it's the same then there is no need for this row. But there is a difference - UC generates and provides the path for managed tables. So maybe the right text here is who is "managing" the storage, UC or the user. In both cases, storage is owned and provided by user. Wdyt?

michelleon: We should remove this row then to reduce confusion, the row two below for "Storage location" specifies that

PR #1387, opened 2026-03-09
CLI support creation of catalog managed tables

yili-db:
**PR Checklist**

**Description of changes**

CLI support creation of catalog managed tables.
Also migrates external table creation from the deprecated Table.createTransactionBuilder API to TableManager.buildCreateTableTransaction.

### DIFF
--- a/examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
+++ b/examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
  */
 public class DeltaKernelUtils {
   private static final Logger LOGGER = LoggerFactory.getLogger(DeltaKernelUtils.class);
+  private static final String ENGINE_INFO = "UnityCatalogCli";
 
   public static Map<String, String> createDeltaTable(
       String tablePath,
@@ -64,20 +65,22 @@ public class DeltaKernelUtils {
         // Catalog-managed table: use UCCatalogManagedClient for coordinated commits.
         // "type"="static" means a pre-configured bearer token; see TokenProvider.create() javadoc.
         TokenProvider tokenProvider =
-            TokenProvider.create(Map.of("type", "static", "token", authToken));
+            TokenProvider.create(
+                Map.of("type", "static", "token", authToken != null ? authToken : ""));
         UCTokenBasedRestClient ucClient =
             new UCTokenBasedRestClient(serverUrl, tokenProvider, Map.of());
         UCCatalogManagedClient ucCatalogManagedClient = new UCCatalogManagedClient(ucClient);
         commitResult =
             ucCatalogManagedClient
-                .buildCreateTableTransaction(tableId, tablePath, tableSchema, "uc-cli")
+                .buildCreateTableTransaction(
+                    tableId, substituteSchemeForS3(tablePath), tableSchema, ENGINE_INFO)
                 // TODO: temporary workaround — Delta Kernel should auto-enable all features
                 // required by catalogManaged. Remove once that is fixed in the kernel.
                 // User-specified properties are also passed here so the kernel writes them into
                 // the Delta log; the returned deltaTableProperties will reflect the actual state.
                 .withTableProperties(
                     mergeProperties(
-                        Map.of("delta.feature.vacuumProtocolCheck", "supported"), userProperties))
+                        userProperties, Map.of("delta.feature.vacuumProtocolCheck", "supported")))
                 .build(engine)
                 .commit(engine, CloseableIterable.emptyIterable() /* dataActions */);
         LOGGER.info(
@@ -87,7 +90,8 @@ public class DeltaKernelUtils {
         // UnityCatalogUtils.getPropertiesForCreate requires SnapshotImpl (an internal Delta
         // Kernel class) to access protocol/table-property internals not exposed on the public
         // Snapshot interface. The cast is safe here because the post-commit snapshot returned
-        // by Delta Kernel is always a SnapshotImpl at runtime.
+        // by Delta Kernel is always a SnapshotImpl at runtime. This should be fixed in the
+        // kernel by exposing the needed accessors on the public Snapshot interface.
         SnapshotImpl postCreateSnapshot =
             (SnapshotImpl)
                 commitResult
@@ -98,10 +102,11 @@ public class DeltaKernelUtils {
                                 "Post-commit snapshot unavailable after table creation"));
         return UnityCatalogUtils.getPropertiesForCreate(engine, postCreateSnapshot);
       } else {
-        // External table: write Delta log directly to storage, no UC commit coordination needed
+        // External table: write Delta log directly to storage, no UC commit coordination needed.
+        // commit() throws on failure, so no explicit version check is needed.
         commitResult =
             TableManager.buildCreateTableTransaction(
-                    substituteSchemeForS3(tablePath), tableSchema, "UnityCatalogCli")
+                    substituteSchemeForS3(tablePath), tableSchema, ENGINE_INFO)
                 .build(engine)
                 .commit(engine, CloseableIterable.emptyIterable() /* dataActions */);
         LOGGER.info(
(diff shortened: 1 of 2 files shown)


examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
```diff
@@ -40,33 +45,90 @@
...
+                .commit(engine, CloseableIterable.emptyIterable() /* dataActions */);
+        LOGGER.info(
+            "Table created successfully at version {}: {}",
+            commitResult.getVersion(),
+            tablePathUri);
+        // UnityCatalogUtils.getPropertiesForCreate requires SnapshotImpl (an internal Delta
+        // Kernel class) to access protocol/table-property internals not exposed on the public
+        // Snapshot interface. The cast is safe here because the post-commit snapshot returned
+        // by Delta Kernel is always a SnapshotImpl at runtime.
+        SnapshotImpl postCreateSnapshot =
```
tdas, line 95: Casting to the internal `SnapshotImpl` is fragile — if the Delta Kernel team refactors this class, this will break at runtime with a `ClassCastException` (no compile-time safety net). The comment explains the rationale well, but is there a way to get `UnityCatalogUtils.getPropertiesForCreate` to accept the public `Snapshot` interface instead? If not, would be good to track this as tech debt with a follow-up issue.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
```diff
@@ -40,33 +45,90 @@
...
+        // Catalog-managed table: use UCCatalogManagedClient for coordinated commits.
+        // "type"="static" means a pre-configured bearer token; see TokenProvider.create() javadoc.
+        TokenProvider tokenProvider =
+            TokenProvider.create(Map.of("type", "static", "token", authToken));
+        UCTokenBasedRestClient ucClient =
+            new UCTokenBasedRestClient(serverUrl, tokenProvider, Map.of());
+        UCCatalogManagedClient ucCatalogManagedClient = new UCCatalogManagedClient(ucClient);
+        commitResult =
+            ucCatalogManagedClient
+                .buildCreateTableTransaction(tableId, tablePath, tableSchema, "uc-cli")
```
tdas: For external tables, the path goes through `substituteSchemeForS3()` (line 104), but here for managed tables the raw `tablePath` is passed as-is. If managed tables can be backed by S3 storage, this could be a bug — Hadoop's S3 connector requires `s3a://` not `s3://`. Is the substitution intentionally omitted here (e.g., handled internally by `UCCatalogManagedClient`)?

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
```diff
@@ -40,33 +45,90 @@
...
+                // TODO: temporary workaround — Delta Kernel should auto-enable all features
+                // required by catalogManaged. Remove once that is fixed in the kernel.
+                // User-specified properties are also passed here so the kernel writes them into
+                // the Delta log; the returned deltaTableProperties will reflect the actual state.
+                .withTableProperties(
+                    mergeProperties(
+                        Map.of("delta.feature.vacuumProtocolCheck", "supported"), userProperties))
```
tdas: Since `userProperties` takes precedence in `mergeProperties` (overrides win), a user could accidentally override `vacuumProtocolCheck` to something invalid. Should the argument order be swapped so the required property always wins?

Also, is there a tracking issue for the TODO to remove this workaround once Delta Kernel handles it natively?

examples/cli/src/test/java/io/unitycatalog/cli/table/CliTableCreationTest.java
```diff
@@ -172,10 +183,5 @@ private void createTableAndAssertReadTableSucceeds(
                 DeltaKernelUtils.readDeltaTable(
                     tableOperations.getTable(TABLE_FULL_NAME).getStorageLocation(), null, 100))
         .doesNotThrowAnyException();
```
tdas, line 185: The explicit table deletion and directory cleanup were removed from this helper. Is this now handled by `cleanUp()` / `@AfterEach`? If `cleanUp()` only drops the schema/catalog (and not individual tables), tables could leak between test runs.

tdas (commented):
Thanks for the PR! Overall the approach is sound and the code is well-commented. A few items to consider:

**Removed version check:** The original code verified `commitResult.getVersion() >= 0` and threw on failure. Both the managed and external table paths now skip this check. If the newer Delta Kernel API guarantees that `commit()` either throws or returns a valid version, this is fine — but that assumption should be documented or asserted. Otherwise a silent failure could go undetected.

**API migration:** The external table path was also refactored from `Table.createTransactionBuilder` to `TableManager.buildCreateTableTransaction`. This is a separate behavioral change from managed-table support — worth noting in the PR description so it's easier to bisect if external table creation regresses.

**Growing parameter lists:** `createDeltaTable` now takes 7 parameters. Consider introducing a context/config object to improve readability and make future extensions easier. (Not a blocker.)

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
yili-db: Good catch! Tracked as a follow-up in #1388.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
yili-db: Fixed

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
yili-db: Good points! The property order is fixed.
For the `vacuumProtocolCheck` workaround, tracked as a follow-up in #1389.

examples/cli/src/test/java/io/unitycatalog/cli/table/CliTableCreationTest.java
yili-db: Added comments. The clean up is done by base class clean up and `@TempDir`.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
```diff
@@ -40,33 +45,93 @@
...
-      // Set the schema of the new table on the transaction builder
-      txnBuilder = txnBuilder.withSchema(engine, tableSchema);
-      // Build the transaction
-      Transaction txn = txnBuilder.build(engine);
-      // create an empty table
-      TransactionCommitResult commitResult = txn.commit(engine, CloseableIterable.emptyIterable());
-      if (commitResult.getVersion() >= 0) {
-        LOGGER.info("Table created successfully at: {}", tablePathUri);
+
+      TransactionCommitResult commitResult;
```
tdas, line 63: **Blocking:** If `authToken` is null (e.g. user forgot to provide it), `Map.of("type", "static", "token", authToken)` will throw a `NullPointerException` — `Map.of()` does not accept null values (Java 9+). This would produce an unhelpful stack trace instead of a clear error.

Suggest adding a guard before this line:
```java
if (authToken == null || authToken.isBlank()) {
    throw new CliException("Auth token is required for managed table creation");
}
```

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
```diff
@@ -40,33 +45,93 @@
...
+      if (tableId != null) {
+        // Catalog-managed table: use UCCatalogManagedClient for coordinated commits.
+        // "type"="static" means a pre-configured bearer token; see TokenProvider.create() javadoc.
+        TokenProvider tokenProvider =
+            TokenProvider.create(Map.of("type", "static", "token", authToken));
+        UCTokenBasedRestClient ucClient =
+            new UCTokenBasedRestClient(serverUrl, tokenProvider, Map.of());
```
tdas, line 71: **Blocking (minor):** The engine identifier here is `"uc-cli"` but the external table path on line 107 uses `"UnityCatalogCli"`. This string gets recorded in the Delta log's commit info, so the inconsistency could make auditing/debugging harder. Suggest using a single constant for both paths.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
yili-db: Fixed. Null token is allowed because most tests don't have token.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
yili-db: Fixed.

PR #1396, opened 2026-03-11
Add Spark catalog aliasing to UC catalog support

BenMagyar:
**PR Checklist**

**Description of changes**

The Spark UC connector currently does not correctly alias 3-part identifiers. The catalog name that would be used would be the Spark catalog name and not the actual warehouse catalog name. 

When:

```
"spark.sql.catalog.target.warehouse": "actual"
```

is set `target.schema.table` identifiers should resolve to `actual.schema.table` but currently resolves instead to the Spark catalog name of `target`. This change keeps track of the resolved catalog so we can use the warehouse option.


yili-db:
@BenMagyar Thank you for sending out a PR.
What is the usage here? Is it because you need to load two different catalogs with the same name from two different metastore/warehouse/URL? Then one of them must be referred as an different alias?

PR #1397, opened 2026-03-12
Cross version test with Delta master

yili-db:
**PR Checklist**

**Description of changes**
Cross version test with Delta master SNAPSHOT in CI

Test matrix:
- Spark 4.0.0 + Delta 4.1.0: all tests (unit + integration + license check)
- Spark 4.1.0 + Delta 4.1.0: integration tests only (spark)
- Spark 4.1.0 + Delta master-SNAPSHOT: integration tests only (spark)

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -18,28 +18,32 @@ jobs:
   test:
     runs-on: ubuntu-latest
 
-    name: test (Spark ${{ matrix.spark-version }}, Delta ${{ matrix.delta-version }})
-    continue-on-error: ${{ matrix.allow-failure || false }}
+    name: >-
+      ${{ matrix.non-blocking && '[Non Blocking]' || '' }}
+      ${{ matrix.run-only-spark-tests && 'Spark tests' || 'All tests' }}
+      (Spark ${{ matrix.spark-version }}, Delta ${{ matrix.delta-version }})
+    continue-on-error: ${{ matrix.non-blocking || false }}
     strategy:
       fail-fast: false
       matrix:
         include:
           - spark-version: "4.0.0"
             delta-version: "4.1.0"
-            # Only one combo needs to run unit tests and license check since they
-            # don't depend on Spark/Delta versions.
-            run-unit-tests: true
-            allow-failure: false
+            # This is the primary combo: runs ALL tests (unit, spark, license).
+            # Other combos run only spark connector tests since unit tests and
+            # license checks don't depend on Spark/Delta versions.
+            run-only-spark-tests: false
+            non-blocking: false
           - spark-version: "4.1.0"
             delta-version: "4.1.0"
-            run-unit-tests: false
-            allow-failure: false
+            run-only-spark-tests: true
+            non-blocking: false
           - spark-version: "4.1.0"
             delta-version: "master-SNAPSHOT"
-            run-unit-tests: false
+            run-only-spark-tests: true
             # Cross-version testing with Delta master is informational; failures
             # indicate upcoming incompatibilities but should not block PRs.
-            allow-failure: true
+            non-blocking: true
 
     steps:
     - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
@@ -91,14 +95,16 @@ jobs:
     - name: Build and verify the local deployment.
       run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION -J-Xmx2G clean package publishM2
 
-    - name: Run all tests with coverage
-      if: matrix.run-unit-tests
-      run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION -J-Xmx2G +jacoco
-
-    - name: Run integration tests with coverage
-      if: ${{ !matrix.run-unit-tests }}
-      run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION -J-Xmx2G spark/jacoco
+    - name: Run tests with coverage
+      run: |
+        if [[ "${{ matrix.run-only-spark-tests }}" == "true" ]]; then
+          echo "Running spark connector tests only"
+          build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION -J-Xmx2G spark/jacoco
+        else
+          echo "Running all tests"
+          build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION -J-Xmx2G +jacoco
+        fi
 
     - name: Run license check
-      if: matrix.run-unit-tests
+      if: ${{ !matrix.run-only-spark-tests }}
       run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION licenseCheck


.github/workflows/unit-tests.yml
```diff
@@ -17,10 +17,16 @@ jobs:
   test:
     runs-on: ubuntu-latest
 
-    name: test (Spark ${{ matrix.spark-version }})
+    name: test (Spark ${{ matrix.spark-version }}, Delta ${{ matrix.delta-version }})
     strategy:
       matrix:
-        spark-version: ["4.0.0", "4.1.0"]
+        include:
+          - spark-version: "4.0.0"
```
yili-db, line 30: only default combination should run non-integrationt ests

.github/workflows/unit-tests.yml
```diff
@@ -18,10 +18,27 @@ jobs:
   test:
     runs-on: ubuntu-latest
 
-    name: test (Spark ${{ matrix.spark-version }})
+    name: test (Spark ${{ matrix.spark-version }}, Delta ${{ matrix.delta-version }})
+    continue-on-error: ${{ matrix.allow-failure || false }}
     strategy:
```
murali-db, line 26:
```suggestion
    strategy:
    fail-fast: false
```

default `fail-fast` is `true`, meaning if ANY matrix combo fails, ALL other running combos are cancelled.

.github/workflows/unit-tests.yml
```diff
@@ -18,10 +18,27 @@ jobs:
...
+            # don't depend on Spark/Delta versions.
+            run-unit-tests: true
+            allow-failure: false
+          - spark-version: "4.1.0"
+            delta-version: "4.1.0"
+            run-unit-tests: false
+            allow-failure: false
+          - spark-version: "4.1.0"
+            delta-version: "master-SNAPSHOT"
+            run-unit-tests: false
```
murali-db:
```suggestion
            run-unit-tests: false
            continue-on-error: true  # SNAPSHOT may break; treat as canary
```

the Delta master-SNAPSHOT job may be marked as `continue-on-error: true` so it doesn't block the overall CI status?

yili-db: It is already:
```
continue-on-error: ${{ matrix.allow-failure || false }}
```

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db, line 1: Tests start to fail because some tests only run with Delta 4.2 which is what this PR enables. And that's working as intended.

Fixing the failing tests here.

.github/workflows/unit-tests.yml
```diff
@@ -18,10 +18,28 @@ jobs:
...
+    name: test (Spark ${{ matrix.spark-version }}, Delta ${{ matrix.delta-version }})
+    continue-on-error: ${{ matrix.allow-failure || false }}
     strategy:
+      fail-fast: false
       matrix:
-        spark-version: ["4.0.0", "4.1.0"]
+        include:
+          - spark-version: "4.0.0"
+            delta-version: "4.1.0"
+            # Only one combo needs to run unit tests and license check since they
```
tdas: i dont get this. only one combo needs to run ALL unit tests.. other combos do need to run unit tests but only uc-spark. isnt it?

.github/workflows/unit-tests.yml
```diff
@@ -31,16 +49,56 @@ jobs:
...
     - name: Compile integration tests to verify compilation and checkstyle issues.
-      run: build/sbt -DsparkVersion=${{ matrix.spark-version }} integrationTests/test:compile
+      run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION -J-Xmx2G integrationTests/test:compile
 
     - name: Build and verify the local deployment.
-      run: build/sbt -DsparkVersion=${{ matrix.spark-version }} clean package publishM2
+      run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION -J-Xmx2G clean package publishM2
+
+    - name: Run all tests with coverage
+      if: matrix.run-unit-tests
```
tdas: this whole complexity is making me wonder whether we should split the actions into server-client tests and spark tests

.github/workflows/unit-tests.yml
yili-db: yes,
one combo: run all tests including unit tests, and uc-spark tests
other combos: run only uc-spark tests

.github/workflows/unit-tests.yml
yili-db: This is done this way for two reasons:
1. Less resource usage. a single command runs all tests so that it is only compiled once. Also minimize the overhead of having extra runner.
2. A simple `sbt` command that runs all tests covers new tests introduced later. If we have `sbt test server client etc` here and later we have a new project, it's not automatically included.

.github/workflows/unit-tests.yml
tdas: okay the comment does not say this. please improve the comment.

tdas: even the variable name can be improved ... like run-only-spark-tests...thats completely self explanatory

.github/workflows/unit-tests.yml
```diff
@@ -18,10 +18,28 @@ jobs:
   test:
     runs-on: ubuntu-latest
```
tdas, line 19: are the delta master builds to be considered non-blocking for merging? i think so... so that failures dont block merges. But if so, lets name those cases as non-block please. see how it has been marked like that in delta for UC master tests

.github/workflows/unit-tests.yml
```diff
@@ -18,10 +18,28 @@ jobs:
...
 
-    name: test (Spark ${{ matrix.spark-version }})
+    name: test (Spark ${{ matrix.spark-version }}, Delta ${{ matrix.delta-version }})
```
tdas: shouldnt this be backported to 0.4 branch as well? to make sure the 0.4.x will work with next version of Delta as well? @openinx how will you make sure that 0.4.2 will work with Delta 4.2?

.github/workflows/unit-tests.yml
```diff
@@ -31,16 +49,56 @@ jobs:
...
-      run: build/sbt -DsparkVersion=${{ matrix.spark-version }} clean package publishM2
+      run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION -J-Xmx2G clean package publishM2
+
+    - name: Run all tests with coverage
+      if: matrix.run-unit-tests
+      run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION -J-Xmx2G +jacoco
 
-    - name: Run tests with coverage
-      run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -J-Xmx2G +jacoco
+    - name: Run integration tests with coverage
```
tdas: this is not integration tests.. this is spark tests. isnt there a separate set of integration tests?

.github/workflows/unit-tests.yml
tdas: okay. 
but at least we can simplify this into a single if statement like `run: | if ...` used above. thats is cleaner than have this condition twice in two different steps. you can always echo which tests you are running.

.github/workflows/unit-tests.yml
yili-db: Yes I'll backport to 0.4.

PR #1400, opened 2026-03-13
[Spark] Fix pagination in listTables and listNamespaces

vladanvasi-db:
**PR Checklist**

**Description of changes**

Both `listTables` and `listNamespaces` in `UCProxy` only fetched the first page of results, ignoring `next_page_token`. This silently truncated results when there were more items than fit on a single page.

Fix by looping through all pages until `getNextPageToken()` returns null.

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
@@ -503,6 +503,9 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
           .containsExactly("pagination_table_1", "pagination_table_2", "pagination_table_3");
     } finally {
       PagedListingHelper.DEFAULT_PAGE_SIZE = originalPageSize;
+      sql("DROP TABLE IF EXISTS %s.%s.pagination_table_1", SPARK_CATALOG, SCHEMA_NAME);
+      sql("DROP TABLE IF EXISTS %s.%s.pagination_table_2", SPARK_CATALOG, SCHEMA_NAME);
+      sql("DROP TABLE IF EXISTS %s.%s.pagination_table_3", SPARK_CATALOG, SCHEMA_NAME);
     }
   }
 
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/SchemaOperationsTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/SchemaOperationsTest.java
@@ -85,6 +85,8 @@ public class SchemaOperationsTest extends BaseSparkIntegrationTest {
           .containsExactly("pagination_schema_1", "pagination_schema_2", SCHEMA_NAME);
     } finally {
       PagedListingHelper.DEFAULT_PAGE_SIZE = originalPageSize;
+      sql("DROP DATABASE IF EXISTS %s.pagination_schema_1", SPARK_CATALOG);
+      sql("DROP DATABASE IF EXISTS %s.pagination_schema_2", SPARK_CATALOG);
     }
   }
 }


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -374,10 +374,16 @@ private class UCProxy(
```
tdas, line 505: no tests? please add some tests!

vladanvasi-db: Added tests with mocked responses. Please take a look

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -746,9 +752,16 @@ private class UCProxy(
...
-    schemasApi.listSchemas(name, 0, null).getSchemas.asScala.map { schema =>
-      Array(schema.getName)
-    }.toArray
+    val schemas = scala.collection.mutable.ArrayBuffer.empty[Array[String]]
+    var pageToken: String = null
+    do {
+      val response = schemasApi.listSchemas(name, 0, pageToken)
+      response.getSchemas.asScala.foreach { schema =>
+        schemas += Array(schema.getName)
+      }
```
urosstan-db: Can we do batch add?
```suggestion
      schemas ++= response.getSchemas.asScala
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -746,9 +752,16 @@ private class UCProxy(
   }
 
   override def listNamespaces(): Array[Array[String]] = {
-    schemasApi.listSchemas(name, 0, null).getSchemas.asScala.map { schema =>
-      Array(schema.getName)
-    }.toArray
+    val schemas = scala.collection.mutable.ArrayBuffer.empty[Array[String]]
+    var pageToken: String = null
+    do {
+      val response = schemasApi.listSchemas(name, 0, pageToken)
```
urosstan-db:
```suggestion
      val response = schemasApi.listSchemas(name, /* limit */ 0, pageToken)
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
vladanvasi-db: Done, thanks

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
vladanvasi-db: Done, thanks

connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogProxyTest.java
```diff
@@ -0,0 +1,186 @@
...
+    assertThat(result[0]).containsExactly("schema1");
+    assertThat(result[1]).containsExactly("schema2");
+    assertThat(result[2]).containsExactly("schema3");
+
+    verify(mockSchemasApi).listSchemas(eq(CATALOG_NAME), eq(0), isNull());
+    verify(mockSchemasApi).listSchemas(eq(CATALOG_NAME), eq(0), eq("token1"));
+  }
+
+  @Test
+  public void testListNamespacesEmptyResult() throws Exception {
```
urosstan-db: nit: this test does not bring huge value

connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogProxyTest.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.List;
+import org.apache.spark.sql.connector.catalog.Identifier;
+import org.apache.spark.sql.connector.catalog.SupportsNamespaces;
+import org.apache.spark.sql.connector.catalog.TableCatalog;
+import org.apache.spark.sql.util.CaseInsensitiveStringMap;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/** Tests that UCProxy correctly paginates through listTables and listNamespaces responses. */
+public class UCSingleCatalogProxyTest {
```
urosstan-db: nit: I would probably call this UCProxySuite and use scala since main class is scala as well

vladanvasi-db: I will check since UCProxy is a private class and we must load it via reflection

connectors/spark/src/test/scala/io/unitycatalog/spark/UCProxySuite.scala
```diff
@@ -0,0 +1,152 @@
+package io.unitycatalog.spark
+
+import io.unitycatalog.client.ApiClient
```
tdas: why cant we test end-to-end directly with UCSingleCatalog instead of testing just UCProxy?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -505,10 +505,14 @@ private class UCProxy(
 
     val catalogName = this.name
     val schemaName = namespace.head
-    val maxResults = 0
-    val pageToken = null
-    val response: ListTablesResponse = tablesApi.listTables(catalogName, schemaName, maxResults, pageToken)
-    response.getTables.toSeq.map(table => Identifier.of(namespace, table.getName)).toArray
+    val tables = scala.collection.mutable.ArrayBuffer.empty[Identifier]
```
openinx, line 508: nit:  instead of using the full qualified class name here, could you please add the package import in the header and then use the simple class name here ? that's more clear.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -505,10 +505,14 @@ private class UCProxy(
...
+    var pageToken: String = null
+    do {
+      val response = tablesApi.listTables(catalogName, schemaName, /* limit */ 0, pageToken)
+      tables ++= response.getTables.asScala.map(table => Identifier.of(namespace, table.getName))
+      pageToken = response.getNextPageToken
+    } while (pageToken != null)
```
openinx, line 514: If there is no more page, will the `pageToken` be empty string ?  if so, then it will be possible to get stuck on a dead loop ?  Please help to check this.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -482,6 +483,32 @@ public void testTableOperations(String scheme, boolean renewCredEnabled) {
...
+  public void testListTablesPagination() {
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG);
+    Integer originalPageSize = PagedListingHelper.DEFAULT_PAGE_SIZE;
+    try {
+      PagedListingHelper.DEFAULT_PAGE_SIZE = 2;
+      setupTable(SPARK_CATALOG, "pagination_table_1");
+      setupTable(SPARK_CATALOG, "pagination_table_2");
+      setupTable(SPARK_CATALOG, "pagination_table_3");
+      List<Row> tables = sql("SHOW TABLES in %s.%s", SPARK_CATALOG, SCHEMA_NAME);
+      assertThat(tables).hasSize(3);
```
openinx, line 496: Could we ensure that the listTable APi was indeed called 2 times , rather than 1 times, although we already set set the servers side pagination to 2.  but it's still possible for the client to call the API once, which actually worries me is: it may still get the accurate call times.

connectors/spark/src/test/java/io/unitycatalog/spark/SchemaOperationsTest.java
```diff
@@ -63,4 +64,29 @@ public void testLoadSchema() {
...
+      // Default schema already exists from setUp, create 2 more for 3 total
+      sql("CREATE DATABASE %s.pagination_schema_1", SPARK_CATALOG);
+      sql("CREATE DATABASE %s.pagination_schema_2", SPARK_CATALOG);
+      List<Row> schemas = sql("SHOW SCHEMAS IN %s", SPARK_CATALOG);
+      assertThat(schemas).hasSize(3);
+      List<String> schemaNames =
+          schemas.stream()
+              .map(row -> row.getString(0))
+              .sorted()
+              .collect(java.util.stream.Collectors.toList());
```
openinx, line 83: nit: could you please use the simple class name here, rather than the full qualified class name, that's better for readability.

connectors/spark/src/test/java/io/unitycatalog/spark/UCProxySuite.java
```diff
@@ -0,0 +1,190 @@
...
+      if (argsByType.containsKey(type)) {
+        args[i] = argsByType.get(type);
+      } else if (type == boolean.class) {
+        args[i] = false;
+      } else {
+        args[i] = null;
+      }
+    }
+
+    Object proxyObj = ctor.newInstance(args);
```
openinx, line 79: Question:  here i'm wondering why we need to use the dynamic class loading ? will it be simple to use the `new UCProxy(...)` directly , as we already mocked those arguments for the constructor ?

connectors/spark/src/test/java/io/unitycatalog/spark/UCProxySuite.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  // -- listTables tests --
+
+  @Test
+  public void testListTablesSinglePage() throws Exception {
+    ListTablesResponse response =
+        new ListTablesResponse()
+            .tables(List.of(new TableInfo().name("table1"), new TableInfo().name("table2")))
+            .nextPageToken(null);
```
openinx, line 99: what's the behavior for `""` for the nextPageToken ?

connectors/spark/src/test/java/io/unitycatalog/spark/UCProxySuite.java
```diff
@@ -0,0 +1,190 @@
...
+    when(mockTablesApi.listTables(eq(CATALOG_NAME), eq(SCHEMA_NAME), eq(0), isNull()))
+        .thenReturn(response);
+
+    Identifier[] result = proxy.listTables(NAMESPACE);
+
+    assertThat(result).hasSize(2);
+    assertThat(result[0].name()).isEqualTo("table1");
+    assertThat(result[1].name()).isEqualTo("table2");
```
openinx, line 107: nit: Could you please assert the `Identifier` directly, which I think also includes the `catalog` equals check.

connectors/spark/src/test/java/io/unitycatalog/spark/UCProxySuite.java
```diff
@@ -0,0 +1,190 @@
...
+        .thenReturn(page1);
+    when(mockTablesApi.listTables(eq(CATALOG_NAME), eq(SCHEMA_NAME), eq(0), eq("token1")))
+        .thenReturn(page2);
+
+    Identifier[] result = proxy.listTables(NAMESPACE);
+
+    assertThat(result).hasSize(3);
+    assertThat(result[0].name()).isEqualTo("table1");
+    assertThat(result[1].name()).isEqualTo("table2");
+    assertThat(result[2].name()).isEqualTo("table3");
```
openinx, line 131: ditto,  it will be better to use the `assertThat(results[0]).isEqualTo(Identifer.of(...))`.

Issue #1378, opened 2026-03-03
Spawn too many threads when running the uc-spark-connector with the remote UC server.


openinx:
I made a simple PR to reproduce this issue here. 

https://github.com/delta-io/delta/pull/6142/changes

When I run this against the remote UC server via the following command line:

```bash
export UC_REMOTE=true
export UC_URI=$UC_URI
export UC_TOKEN=$UC_TOKEN
export UC_CATALOG_NAME=main
export UC_SCHEMA_NAME=demo_zh
export UC_BASE_TABLE_LOCATION=$S3_BASE_LOCATION
build/sbt -mem 4096 "sparkUnityCatalog/testOnly io.sparkuctest.UCThreadLeakReproTest"
```

And when I `jstack` the thread with multiple times,  actually we can clearly see the increasing threads number. 

```bash
jstack -l 24490 | grep 'sdk-ScheduledExecutor' 
"sdk-ScheduledExecutor-1-0" #103 [67843] daemon prio=5 os_prio=31 cpu=0.30ms elapsed=361.47s tid=0x00000009d3411500 nid=67843 waiting on condition  [0x0000000179aaa000]
"sdk-ScheduledExecutor-1-1" #104 [85251] daemon prio=5 os_prio=31 cpu=0.46ms elapsed=361.47s tid=0x00000009d3411c00 nid=85251 waiting on condition  [0x0000000179cb6000]
"sdk-ScheduledExecutor-1-2" #105 [84739] daemon prio=5 os_prio=31 cpu=0.15ms elapsed=361.01s tid=0x00000009d3413800 nid=84739 waiting on condition  [0x0000000179ec2000]
"sdk-ScheduledExecutor-1-3" #106 [84483] daemon prio=5 os_prio=31 cpu=0.13ms elapsed=361.01s tid=0x00000009d1268000 nid=84483 waiting on condition  [0x000000017a0ce000]
"sdk-ScheduledExecutor-2-0" #107 [68867] daemon prio=5 os_prio=31 cpu=0.38ms elapsed=360.52s tid=0x00000009d1268700 nid=68867 waiting on condition  [0x000000017a2da000]
"sdk-ScheduledExecutor-2-1" #108 [83971] daemon prio=5 os_prio=31 cpu=0.15ms elapsed=360.52s tid=0x00000009d1268e00 nid=83971 waiting on condition  [0x000000017a4e6000]
[...]

PR #1402, opened 2026-03-15
[Spark] Use the CredScopedFileSystem for the unitycatalog spark connector.

openinx:
**PR Checklist**

**Description of changes**

This PR fix this https://github.com/unitycatalog/unitycatalog/issues/1378. 

Let's explain how is this problem happened, and how is this solution solved the problem. 

### Why the problem ?  

Before this patch, we intentionally disabled all `s3`, `gcs`, `azure` filesystem implementation cache. 
1. [s3 disable cache](https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/java/io/unitycatalog/spark/auth/CredPropsUtil.java#L95-L96).  
2. [abfs disable cache](https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/java/io/unitycatalog/spark/auth/CredPropsUtil.java#L124-L125)
3. [gcs disable cache](https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/java/io/unitycatalog/spark/auth/CredPropsUtil.java#L110)

And if we don't disable the filesystem cache, then in the same spark job,  the same executor will always use the same  filesystem instance to access all the objects under the  `s3://bucket/*`.  Essentially,  the hadoop `FileSystem` can only support to access the same bucket with only credential. 

In this [FileSystem.get(uri, conf)](https://github.com/apache/hadoop/blob/a178eb780f5c983ab7e5c5a4add8a3027060c68d/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L537-L559), we can clearly see it will use the `scheme` and `authority`  as the cache key  ( if the uri is `s3://bucket/path/to/key`, the scheme is `s3`, the authority is `bucket` ).  then for every filesystem access to the `s3://bucket/`, it will always use the same filesytem instance. 

But the problem is,  it's possible to access multiple unity catalog managed table in the same process, under the same `s3://bucket/`. while different unity catalog managed table will require different unique credentials.  if we use the same S3AFileSystem instance to access all the table data files,  then we will mess up the multiple table credentials in the same table. 

That's why we intentionally set the `fs.<scheme>.disable.cache` to true, which disable all the filesystem cache.  And because the cache disabling, then for every object request, it will always create a new  FileSystem instance to access the object in the storage.  that's why it finally caused the problem that we describe here : https://github.com/unitycatalog/unitycatalog/issues/1378

### How to solve the problem ? 

First of all, we definitely cannot set the `fs.<scheme>.disable.cache` to be `false`,  otherwise the multiple table credentials problem will mix in the same filesystem instance. 

Secondly,  if we review the `FileSystem#get` code seriously,  its process is here ( for the `fs.<scheme>.disable.cache=false` case ). 

1.  [FileSystem.get(URI uri, Configuration conf)](https://github.com/apache/hadoop/blob/a178eb780f5c983ab7e5c5a4add8a3027060c68d/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L555)
2. [FileSystem.createFileSystem(uri, conf)](https://github.com/apache/hadoop/blob/a178eb780f5c983ab7e5c5a4add8a3027060c68d/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L3609-L3634)

The key point is here:  [fs.initialize(uri, conf)](https://github.com/apache/hadoop/blob/a178eb780f5c983ab7e5c5a4add8a3027060c68d/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L3620). 

For every newly created `FileSystem` instance, it will always call the `initialize` method.  And actually our solution can fully leverage this `initialize` method.  which means, if we have a `CredentialScopeFileSystem` to be specified by all the cloud vendor implementation,  and the `CredentialScopeFileSystem` have a global cache, to maintain the `<credKey,  S3AFileSystem/GCSFileSystem/AbfsFileSystem>` . 

then because we already set the `fs.<scheme>.disable.cache` to be false, then for every data file access, we will still create a new `CredentialScopeFileSystem`.  But the core design for the `CredentialScopeFileSystem` is:   the `CredentialScopeFileSystem` is actually every lightweight, basically it's just a  filesystem wrapper,  then actually it have no cost to new a `CredentialScopeFileSystem`. 

And for every `CredentialScopeFileSystem#initialize`,  it will always use the hadoop configuration to get the real `S3AFileSystem/GCSFileSystem/AbfsFileSystem` from the global cache,  then basically we are always reusing the same `S3AFileSystem/..` instance, if the `credKey` are the same. 

Essentially, this solution disabled the default  hadoop `FileSystem` cache,  but introduced another credentials level cache.  and with this approach,  we successfully avoid to always new the filesystem for every data file access. 

The CredScopedFileSystem [javadoc](https://github.com/unitycatalog/unitycatalog/pull/1402/changes#diff-7220a08f3f7a13be3ed103d85231996d9a5362ec19cfc6808e8f3727381523e1R32-R71) is also a good material to describe the design.

### DIFF
--- a/connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedKey.java
+++ b/connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedKey.java
@@ -6,6 +6,20 @@ import java.util.Objects;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 
+/**
+ * Cache key that identifies a credential scope for {@link CredScopedFileSystem}.
+ *
+ * <p>There are three implementations:
+ *
+ * <ul>
+ *   <li>{@link TableCredScopedKey} — keyed by table ID and operation; used for table-level
+ *       temporary credentials.
+ *   <li>{@link PathCredScopedKey} — keyed by path and operation; used for path-level temporary
+ *       credentials.
+ *   <li>{@link DefaultCredScopedKey} — keyed by URI scheme and authority; used as a fallback when
+ *       no Unity Catalog credential type is present in the configuration.
+ * </ul>
+ */
 public interface CredScopedKey {
 
   static CredScopedKey create(URI uri, Configuration conf) {


tdas:
Can you provide a detailed description of the solution? With architecture diagrams if possible?

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
```diff
@@ -0,0 +1,414 @@
...
+ *       {@code CredScopedFileSystem} for each file access. Because {@code CredScopedFileSystem} is
+ *       a thin, stateless wrapper, this is cheap.
+ *   <li><b>Global credential-scoped cache for the real delegate.</b> {@code CredScopedFileSystem}
+ *       maintains a static {@link #CACHE} keyed by {@link CredScopedKey}, which encodes the
+ *       credential scope (table ID + operation, or path + operation). On each {@link
+ *       #initialize(URI, Configuration)} call the key is derived from the Hadoop {@link
+ *       Configuration} injected by {@link CredPropsUtil}, and the corresponding real {@link
+ *       FileSystem} (e.g. {@code S3AFileSystem}) is looked up or created. Requests that share the
+ *       same credential scope therefore reuse the same underlying connection pool, while requests
+ *       with different credentials transparently receive their own isolated instance.
```
tdas: Can you explain how the scheme-to-fs mappings are set to ensure this new fs is loaded for table paths?

tdas (changes requested):

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
openinx: Hi @tdas .  I already added the description in this PR,  please take a look. https://github.com/unitycatalog/unitycatalog/pull/1402#issue-4077750683

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
```diff
@@ -0,0 +1,414 @@
...
+ * </ol>
+ *
+ * <p>All public {@link FileSystem} operations are delegated to the credential-scoped instance,
+ * including methods that Hadoop's base class leaves unimplemented (ACLs, XAttrs, snapshots, storage
+ * policies, etc.), so callers see a fully functional filesystem regardless of which underlying
+ * implementation backs it.
+ */
+public class CredScopedFileSystem extends FileSystem {
+
+  private static final Map<CredScopedKey, FileSystem> CACHE = Maps.newConcurrentMap();
```
tdas: [AI Generated]

**Cache eviction is missing.** This static `CACHE` only grows — there is no eviction, TTL, size bound, or cleanup on `close()`. Over a long-running Spark session that accesses many distinct tables (each with its own credential scope), this will accumulate `FileSystem` instances and their associated connection pools indefinitely.

Consider adding either:
- A bounded cache with LRU eviction (e.g. Guava `CacheBuilder` with `maximumSize` and a `removalListener` that calls `fs.close()`), or
- A `WeakValueMap` so entries are eligible for GC when no wrapper references them, or
- At minimum, a `static void clearCache()` method that tests and shutdown hooks can call.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
```diff
@@ -0,0 +1,414 @@
...
+  }
+
+  private static FileSystem newFileSystem(URI uri, Configuration conf) {
+    try {
+      Configuration fsConf = new Configuration(conf);
+
+      // S3: restore impl to the real S3AFileSystem, overriding the CredScopedFileSystem set by
+      // CredPropsUtil. S3 requires explicit configuration because Hadoop does not discover
+      // S3AFileSystem via the service loader by default.
+      fsConf.set("fs.s3.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem");
```
tdas: [AI Generated]

**Original filesystem impl configs are silently overwritten.** `CredPropsUtil` overwrites `fs.<scheme>.impl` with `CredScopedFileSystem` (discarding whatever was there before), and then `newFileSystem()` attempts to restore the original by **hardcoding** `S3AFileSystem` for S3 and relying on service-loader discovery for GCS/Azure. If a user had configured a custom filesystem implementation (e.g. a MinIO-specific FS, or a non-default GCS impl not registered via service loader), that configuration is permanently lost.

A more robust approach: have `CredPropsUtil` save the original impl value into a side-channel config key before overwriting it, so `newFileSystem()` can restore it precisely:

```java
// In CredPropsUtil, before overwriting:
String original = conf.get("fs.s3a.impl");
if (original != null) {
    set("fs.s3a.impl.original", original);
}
set("fs.s3a.impl", CredScopedFileSystem.class.getName());

// In newFileSystem(), restore from the side channel:
String originalImpl = fsConf.get("fs.s3a.impl.original");
if (originalImpl != null) {
    fsConf.set("fs.s3a.impl", originalImpl);
} else {
    fsConf.set("fs.s3a.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem");
}
```

This way the real delegate implementation is always the one the user originally configured.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
```diff
@@ -0,0 +1,414 @@
...
+    return delegate.getStoragePolicy(src);
+  }
+
+  @Override
+  public Collection<? extends BlockStoragePolicySpi> getAllStoragePolicies() throws IOException {
+    return delegate.getAllStoragePolicies();
+  }
+
+  @Override
+  public void close() throws IOException {
```
tdas: [AI Generated]

**`close()` intentionally skips the delegate, but this means delegates are never closed.** Since the delegates live in the static `CACHE` and are not registered in Hadoop's own cache (because `impl.disable.cache=true`), Hadoop's shutdown hooks won't reach them either. This means cloud SDK clients (and their thread pools, connection pools, etc.) are only cleaned up on JVM exit.

This is acceptable for short-lived Spark jobs, but in long-running environments (notebooks, Spark Thrift Server, etc.) it could be problematic — especially combined with the unbounded cache growth noted above.

build.sbt
```diff
@@ -24,7 +24,7 @@ lazy val scala213 = "2.13.17"
 lazy val deltaVersion = "4.1.0"
 lazy val sparkVersion = sys.props.getOrElse("sparkVersion", "4.0.0")
 lazy val sparkMajorMinorVersion = sparkVersion.split("\\.").take(2).mkString(".")
-lazy val hadoopVersion = "3.4.2"
+lazy val hadoopVersion = "3.4.0"
```
tdas: [AI Generated]

nit: The Hadoop version downgrade from 3.4.2 to 3.4.0 is not explained in the PR description. Is this intentional? If so, it would be good to document the reason (compatibility issue? test failure with 3.4.2?). If not, this should probably be reverted to avoid losing bugfixes/security patches from 3.4.1 and 3.4.2.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFs.java
```diff
@@ -0,0 +1,13 @@
+package io.unitycatalog.spark.fs;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.DelegateToFileSystem;
+
+public class CredScopedFs extends DelegateToFileSystem {
```
tdas, line 21: need class docs.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -246,13 +250,20 @@ public static Stream<Arguments> credentialRenewalArguments() {
...
-                        OptionsUtil.WAREHOUSE, CATALOG_NAME,
-                        OptionsUtil.RENEW_CREDENTIAL_ENABLED, String.valueOf(RENEW_CRED_ENABLED))));
+                        OptionsUtil.URI,
+                        SERVER_URL,
+                        UC_PROPS_AUTH_TYPE,
+                        "oauth",
+                        UC_PROPS_OAUTH_URI,
+                        OAUTH_URI,
+                        UC_PROPS_OAUTH_CLIENT_ID,
+                        OAUTH_CLIENT_ID,
```
tdas: [AI Generated]

These formatting-only changes (reformatting `Map.of(...)` arguments) are unrelated to the `CredScopedFileSystem` feature. Can you pull these out into a separate PR? They clutter the diff and make it harder to review the actual logic changes.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
openinx: Yes,  I love this idea.  Actually,  we can use this approach to pass the test which leveraged the `S3CredFileSystem` to mock the credential before.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
openinx: TODO in the next update.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
openinx: we should close the filesystem during the eviction .

connectors/spark/src/main/java/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -91,6 +93,10 @@ private static class S3PropsBuilder extends PropsBuilder<S3PropsBuilder> {
 
     S3PropsBuilder() {
       // Common properties for S3.
+      set("fs.s3.impl", CredScopedFileSystem.class.getName());
```
openinx: Let's add a flag to disable by default, but will enable in the next release after oss uc 0.4.1

connectors/spark/src/main/java/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -89,11 +91,22 @@ public Map<String, String> build() {
...
 
-    S3PropsBuilder() {
+    S3PropsBuilder(boolean credScopedFsEnabled) {
       // Common properties for S3.
       set("fs.s3a.path.style.access", "true");
       set("fs.s3.impl.disable.cache", "true");
       set("fs.s3a.impl.disable.cache", "true");
+
+      if (credScopedFsEnabled) {
+        // Override the impl so Hadoop routes through CredScopedFileSystem. The original impl is
```
openinx: I think those comments are redundant,  as we have the javadoc for the credscopedFileSystem there.

connectors/spark/src/main/java/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -335,6 +382,7 @@ public static Map<String, String> createTableCredProps(
 
   public static Map<String, String> createPathCredProps(
```
openinx, line 472: Add the javadoc for the `credScopedFsEnabled`  please.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
```diff
@@ -0,0 +1,448 @@
...
+  /**
+   * LRU cache of real {@link FileSystem} instances keyed by credential scope. Evicted entries are
+   * closed to release connection pools and SDK thread pools (e.g. AWS sdk-ScheduledExecutor
+   * threads). The cache is bounded to prevent unbounded growth when many distinct credential scopes
+   * are accessed in a long-running session.
+   */
+  /** Visible for testing. */
+  static final Cache<CredScopedKey, FileSystem> CACHE =
+      CacheBuilder.newBuilder()
+          .maximumSize(100)
```
openinx: Could also define a system property to set the max size for cache, like 

```java
  // The credential cache, for saving QPS to unity catalog server.
  static final Cache<String, GenericCredential> globalCache;
  private static final String UC_CREDENTIAL_CACHE_MAX_SIZE =
  private static final long UC_CREDENTIAL_CACHE_MAX_SIZE_DEFAULT = 1024;

  static {
    long maxSize = Long.getLong(UC_CREDENTIAL_CACHE_MAX_SIZE, UC_CREDENTIAL_CACHE_MAX_SIZE_DEFAULT);
    globalCache = CacheBuilder.newBuilder().maximumSize(maxSize).build();
  }
```

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFs.java
openinx: Done.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedKey.java
```diff
@@ -0,0 +1,102 @@
+package io.unitycatalog.spark.fs;
+
+import io.unitycatalog.spark.UCHadoopConf;
+import java.net.URI;
+import java.util.Objects;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+
+public interface CredScopedKey {
```
openinx, line 23: All the `CredScopedKey`  implementation needs to implement the `toString` method, as the `CredScopedFileSystem#initialize(..)` will use the key to concat the error message.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
```diff
@@ -0,0 +1,448 @@
...
+ *       same credential scope therefore reuse the same underlying connection pool, while requests
+ *       with different credentials transparently receive their own isolated instance.
+ * </ol>
+ *
+ * <p>All public {@link FileSystem} operations are delegated to the credential-scoped instance,
+ * including methods that Hadoop's base class leaves unimplemented (ACLs, XAttrs, snapshots, storage
+ * policies, etc.), so callers see a fully functional filesystem regardless of which underlying
+ * implementation backs it.
+ */
+public class CredScopedFileSystem extends FileSystem {
```
openinx: Please see if it's possible to use the `FilterFileSystem` to avoid implement all the methods in the `FileSystem`. 

https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FilterFileSystem.java

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
```diff
@@ -0,0 +1,448 @@
...
+      // S3: restore impl using the side-channel key saved by CredPropsUtil before it overrode
+      // fs.<scheme>.impl with CredScopedFileSystem. The side-channel lets callers (including
+      // tests) substitute a custom implementation by setting fs.s3a.impl.original; the fallback
+      // is S3AFileSystem because Hadoop does not discover it via the service loader by default.
+      fsConf.set(
+          "fs.s3.impl",
+          fsConf.get("fs.s3.impl.original", "org.apache.hadoop.fs.s3a.S3AFileSystem"));
+      fsConf.set(
+          "fs.s3a.impl",
+          fsConf.get("fs.s3a.impl.original", "org.apache.hadoop.fs.s3a.S3AFileSystem"));
```
openinx: We did not set the `fs.s3.impl.original` when we override the default `fs.s3.impl` here https://github.com/unitycatalog/unitycatalog/pull/1402/changes#diff-0b41f7485377b6e5c30391a7f6e75018947f105e8dc58e0def7b30bdf8fb48c3R105-R108, then we loss the original implementation. 

So we should set those `fs.<scheme>.impl.original`  in the CredPropsUtil before we do the override. 

Please also add the test coverage for those, including `s3`, `s3a`, `gcs`, and `abfs`, `abfss`.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
openinx: The one risk that we listed all the methods in the `FileSystem` here is: if there is any change in the public methods in the `FileSystem`, cross different hadoop versions.  then we may need to implement different versions of hadoop `FileSystem` in this class, which is hard to maintain. 

But if we use the `FilterFileSystem`, then we can easily use the same class to support different hadoop versions.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
```diff
@@ -0,0 +1,448 @@
...
+      fsConf.set(
+          "fs.AbstractFileSystem.s3a.impl",
+          fsConf.get("fs.AbstractFileSystem.s3a.impl.original", "org.apache.hadoop.fs.s3a.S3A"));
+      fsConf.set("fs.s3.impl.disable.cache", "true");
+      fsConf.set("fs.s3a.impl.disable.cache", "true");
+
+      // GCS: impl does not need to be explicitly restored — GoogleHadoopFileSystem is registered
+      // via the Java service loader and Hadoop discovers it automatically from the URI scheme.
+      fsConf.unset("fs.gs.impl");
+      fsConf.unset("fs.AbstractFileSystem.gs.impl");
```
openinx: Please still use the restore code, otherwise if someone explicitly specify some gcs hadoop filesystem, then it won't work,  that's not we want.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
```diff
@@ -0,0 +1,448 @@
...
+      fsConf.set("fs.gs.impl.disable.cache", "true");
+
+      // Azure: same as GCS — AzureBlobFileSystem and SecureAzureBlobFileSystem are registered
+      // via the Java service loader, so only the CredScopedFileSystem override needs clearing.
+      fsConf.unset("fs.abfs.impl");
+      fsConf.unset("fs.abfss.impl");
+      fsConf.unset("fs.AbstractFileSystem.abfs.impl");
+      fsConf.unset("fs.AbstractFileSystem.abfss.impl");
```
openinx: Ditto for abfs/abfss.

connectors/spark/src/main/java/io/unitycatalog/spark/utils/OptionsUtil.java
```diff
@@ -12,6 +12,16 @@ private OptionsUtil() {}
   public static final String RENEW_CREDENTIAL_ENABLED = "renewCredential.enabled";
   public static final boolean DEFAULT_RENEW_CREDENTIAL_ENABLED = true;
 
+  /**
+   * When {@code true}, wraps the cloud filesystem with {@link
+   * io.unitycatalog.spark.fs.CredScopedFileSystem} to reuse underlying connection pools across
```
openinx: > reuse underlying connection pools across requests that share the same credential scope, preventing AWS SDK v2 thread leaks (issue #1378).

Not aws s3 specifically,   I think we should say it will enable the filesystem cache per-credential, not per bucket. with this we can fully reuse the existing credential request specified filesystem, so that we don't need to new filesystem for every file access. ( refine it).

connectors/spark/src/main/java/io/unitycatalog/spark/utils/OptionsUtil.java
```diff
@@ -12,6 +12,16 @@ private OptionsUtil() {}
...
+   * requests that share the same credential scope, preventing AWS SDK v2 thread leaks (issue
+   * #1378). Disabled by default; will be enabled by default in a future release.
+   */
+  public static final String CRED_SCOPED_FS_ENABLED = "credScopedFs.enabled";
+
```
openinx: The blank line can be removed.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -412,7 +419,7 @@ private class UCProxy(
             //       for read or write, we can request the proper credential after fixing Spark.
             new GenerateTemporaryTableCredential().tableId(tableId).operation(tableOp)
           )
-      }       catch {
+      } catch {
```
openinx: Don't change the unrelated code.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AwsCredScopedFsRenewITTest.java
```diff
@@ -0,0 +1,31 @@
...
+  @Override
+  protected Map<String, String> catalogExtraProps() {
+    String catalogKey = "spark.sql.catalog." + CATALOG_NAME;
+    return Map.of(
+        // Enable CredScopedFileSystem wrapping for this catalog.
+        catalogKey + "." + OptionsUtil.CRED_SCOPED_FS_ENABLED,
+        "true",
+        // Tell CredScopedFileSystem to restore S3CredFileSystem as the real delegate so the
+        // credential-checking logic in the parent test still runs.
+        "spark.hadoop.fs.s3.impl.original",
```
openinx: No, from the testing side, we should always set the `fs.s3.impl` and `fs.s3a.impl`, the `fs.s3.impl.original` and `fs.s3a.impl.original` shouldn't always be exposed to end users.

connectors/spark/src/test/java/io/unitycatalog/spark/fs/CredScopedFileSystemCacheTest.java
```diff
@@ -0,0 +1,74 @@
...
+  private static Configuration tableConf(String tableId, String op) {
+    Configuration conf = new Configuration();
+    conf.set(UCHadoopConf.UC_CREDENTIALS_TYPE_KEY, UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE);
+    conf.set(UCHadoopConf.UC_TABLE_ID_KEY, tableId);
+    conf.set(UCHadoopConf.UC_TABLE_OPERATION_KEY, op);
+    return conf;
+  }
+
+  @Test
+  void sameScope_reusesSameDelegate() throws Exception {
```
openinx: Don't use the underscore for the method name.

connectors/spark/src/test/java/io/unitycatalog/spark/fs/CredScopedKeyTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.net.URI;
+import org.apache.hadoop.conf.Configuration;
+import org.junit.jupiter.api.Test;
+
+class CredScopedKeyTest {
+
+  // ── PathCredScopedKey ────────────────────────────────────────────────────
+
+  @Test
+  void pathKey_equalWhenSamePathAndOp() {
```
openinx: Don't use the underscore for the method name.  Also please remove the comments in this class.

integration-tests/src/test/java/io/unitycatalog/integrationtests/SparkCredentialRenewalTest.java
```diff
@@ -233,10 +233,14 @@ public static Stream<Arguments> credentialRenewalArguments() {
                     tableType,
                     baseLocation,
                     Map.of(
-                        OptionsUtil.URI, SERVER_URL,
-                        UC_PROPS_TOKEN, AUTH_TOKEN,
-                        OptionsUtil.WAREHOUSE, CATALOG_NAME,
-                        OptionsUtil.RENEW_CREDENTIAL_ENABLED, String.valueOf(RENEW_CRED_ENABLED))));
+                        OptionsUtil.URI,
```
openinx: Don't change those unrelated part in the SparkCredentialRenewalTest

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -419,6 +429,28 @@ object UCSingleCatalog {
...
+   * instantiate the wrong delegate.
+   */
+  def sessionHadoopFsImplProps(): util.Map[String, String] = {
+    val fsImplKeys = Set(
+      "fs.s3.impl", "fs.s3a.impl", "fs.gs.impl", "fs.abfs.impl", "fs.abfss.impl",
+      "fs.AbstractFileSystem.s3.impl", "fs.AbstractFileSystem.s3a.impl",
+      "fs.AbstractFileSystem.gs.impl", "fs.AbstractFileSystem.abfs.impl",
+      "fs.AbstractFileSystem.abfss.impl")
+    val result = new util.HashMap[String, String]()
+    for ((k, v) <- SparkSession.active.conf.getAll if fsImplKeys.contains(k)) result.put(k, v)
```
tdas: [AI Generated]

**Bug: this method always returns an empty map.** `SparkSession.active.conf.getAll` returns keys with the `spark.hadoop.` prefix (e.g. `spark.hadoop.fs.s3a.impl`), but `fsImplKeys` contains the unprefixed form (`fs.s3a.impl`). The `contains` check never matches, so `existingProps` is always empty, and `saveAndOverride` always falls back to the hardcoded default.

This means the entire purpose of `saveAndOverride` — preserving a user's custom filesystem implementation — is silently broken. A user who configured `spark.hadoop.fs.s3a.impl = com.example.CustomS3` would still get `S3AFileSystem` as the delegate.

Fix: read from the Hadoop Configuration (where the prefix is already stripped) instead of the Spark RuntimeConfig:

```scala
def sessionHadoopFsImplProps(): util.Map[String, String] = {
  val hadoopConf = SparkSession.active.sparkContext.hadoopConfiguration
  val fsImplKeys = Seq(
    "fs.s3.impl", "fs.s3a.impl", "fs.gs.impl", "fs.abfs.impl", "fs.abfss.impl",
    "fs.AbstractFileSystem.s3.impl", "fs.AbstractFileSystem.s3a.impl",
    "fs.AbstractFileSystem.gs.impl", "fs.AbstractFileSystem.abfs.impl",
    "fs.AbstractFileSystem.abfss.impl")
  val result = new util.HashMap[String, String]()
  for (key <- fsImplKeys) {
    val value = hadoopConf.get(key)
    if (value != null) result.put(key, value)
  }
  result
}
```

Note: if `credScopedFsEnabled` was already applied in a previous credential-vending call, the Hadoop conf may already contain `CredScopedFileSystem`. You may need to guard against this (e.g. skip keys whose value is already `CredScopedFileSystem`).

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * LRU cache of real {@link FileSystem} instances keyed by credential scope. Evicted entries are
+   * closed to release connection pools and SDK thread pools (e.g. AWS sdk-ScheduledExecutor
+   * threads). The cache is bounded to prevent unbounded growth when many distinct credential scopes
+   * are accessed in a long-running session. The maximum size can be tuned via the system property
+   * {@code unitycatalog.credScopedFs.cache.maxSize}.
+   */
+  /** Visible for testing. */
+  static final Cache<CredScopedKey, FileSystem> CACHE;
```
tdas, line 66: [AI Generated]

nit: There are two consecutive `/** */` javadoc blocks here — the descriptive one (lines 57-64) and `/** Visible for testing. */` (line 65). These should be merged into a single javadoc block. Currently the first block is orphaned (not attached to any declaration) and only the second one is associated with the `CACHE` field.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedKey.java
```diff
@@ -0,0 +1,117 @@
...
```
tdas, line 23: [AI Generated]

Needs interface-level javadoc. This is the key abstraction for the credential-scoped caching strategy — worth documenting what a "credential scope" means, the three key types (table, path, noop), and how it's used by `CredScopedFileSystem`.

tdas (commented):
[AI Generated]

Great improvements since the last round — FilterFileSystem, LRU cache with close-on-evict, the feature flag, and the saveAndOverride pattern all look good. A few issues remain, one of which looks like a bug.

connectors/spark/src/main/java/io/unitycatalog/spark/auth/CredPropsUtil.java
```diff
@@ -80,6 +82,19 @@ public T pathOperation(PathOperation pathOp) {
...
 
+    /**
+     * Saves the current value of {@code key} from {@code existingProps} (falling back to {@code
+     * defaultOriginal}) under {@code key + ".original"}, then overrides {@code key} with {@code
+     * newValue}. This lets {@link CredScopedFileSystem#newFileSystem} restore the real delegate
+     * implementation after the wrapper has been installed.
+     */
+    public T saveAndOverride(
+        Map<String, String> existingProps, String key, String defaultOriginal, String newValue) {
+      builder.put(key + ".original", existingProps.getOrDefault(key, defaultOriginal));
```
tdas: this is wrong. you have to get the properties from hadoopconf and save that.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: Great finding,  the test also failed because of this. I updated the code.

connectors/spark/src/main/java/io/unitycatalog/spark/auth/CredPropsUtil.java
openinx: Done, just construct the `Map<String, String>` from the hadoop conf.

openinx: Please see here: https://github.com/unitycatalog/unitycatalog/pull/1402/changes#diff-b496bfc650c9a1c585fb5709e010388ca74382670b419be5ebf4faa1bc499dd4R442

build.sbt
openinx: Done.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedFileSystem.java
openinx: Done.

connectors/spark/src/main/java/io/unitycatalog/spark/fs/CredScopedKey.java
openinx: Done.

PR #1405, opened 2026-03-17
[Spark] Preserve Delta provider on managed replace

TimothyW553:
**PR Checklist**

**Description of changes**
- preserve the existing Delta provider for replace on existing UC-managed Delta tables when the caller omits `USING DELTA`
- reject explicit provider changes on existing UC-managed Delta replace paths
- add staging-table coverage for omitted-provider normalization and provider-mismatch rejection

essentially what can happen is:
```sql
-- create a UC-managed delta table
spark-sql ()> CREATE TABLE main.demo_zh.t_rtas USING DELTA
            >   TBLPROPERTIES ('delta.feature.catalogManaged'='supported')
            >   AS SELECT 1 AS i, 'old' AS s;

-- replace without specifying USING DELTA
spark-sql ()> REPLACE TABLE main.demo_zh.t_rtas
            >   TBLPROPERTIES ('delta.feature.catalogManaged'='supported')
            >   AS SELECT 2 AS i, 'new' AS s;
[FAILURE FOR REPLACE]

-- look at table 
spark-sql ()> SELECT * FROM main.demo_zh.t_rtas;
--> TABLE IS NOT FOUND AND IT HAS BEEN DROPPED
```

this is because it was default to regular drop, create table replace path instead of using Delta staging atomic replace path.
```java
override def stageReplace(
      ident: Identifier,
      schema: StructType,
      partitions: Array[Transform],
      properties: util.Map[String, String]): StagedTable =
    recordFrameProfile("DeltaCatalog", "stageReplace") {
      if (DeltaSourceUtils.isDeltaDataSourceName(getProvider(properties))) {
        new StagedDeltaTableV2(
          ident,
          schema,
          partitions,
          properties,
          TableCreationModes.Replace
        )
      } else {
        super.dropTable(ident)
        val table = createCatalogTable(ident, schema, partitions, properties
        )
        BestEffortStagedTable(ident, table, this)
      }
    }
```

This is now working because we pass in the provider `DELTA`. If the user specifies any other format, it will fail on the UC side.

**Testing**
- build/sbt -J-Xmx4g -J-XX:+UseG1GC -DsparkVersion=4.0.0 'spark/testOnly io.unitycatalog.spark.UCSingleCatalogStagingTableTest'

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -231,9 +231,10 @@ class UCSingleCatalog
     Option(properties.get(TableCatalog.PROP_PROVIDER))
       .filterNot(_.equalsIgnoreCase(existingProvider))
       .foreach(provider => throw new ApiException(
-        s"$operation on existing UC-managed Delta table $fullTableName cannot change provider " +
-          s"from ${existingProvider.toUpperCase(Locale.ROOT)} to " +
-          s"${provider.toUpperCase(Locale.ROOT)}."))
+        s"$operation is only supported for Unity Catalog managed Delta tables and requires " +
+          s"USING DELTA. Cannot change table format from " +
+          s"${existingProvider.toUpperCase(Locale.ROOT)} to " +
+          s"${provider.toUpperCase(Locale.ROOT)} for $fullTableName."))
     val newProps = new util.HashMap[String, String]
     newProps.putAll(properties)
     newProps.put(TableCatalog.PROP_PROVIDER, existingProvider)
@@ -351,7 +352,7 @@ class UCSingleCatalog
       existingTable.get,
       properties,
       "REPLACE TABLE")
-    assertProviderSpecified("REPLACE TABLE", newProps)
+    UCSingleCatalog.requireProviderSpecified("REPLACE TABLE", newProps)
     stagingCatalog.stageReplace(ident, schema, partitions, newProps)
   }
 
@@ -375,7 +376,7 @@ class UCSingleCatalog
       validateManagedDeltaCreateProperties(properties)
       stageManagedDeltaTableAndGetProps(ident, properties)
     }
-    assertProviderSpecified("CREATE OR REPLACE TABLE", newProps)
+    UCSingleCatalog.requireProviderSpecified("CREATE OR REPLACE TABLE", newProps)
     stagingCatalog.stageCreateOrReplace(ident, schema, partitions, newProps)
   }
 
@@ -420,16 +421,6 @@ class UCSingleCatalog
       stagingCatalog.stageCreate(ident, schema, partitions, properties)
     }
   }
-
-  private def assertProviderSpecified(
-      operation: String,
-      properties: util.Map[String, String]): Unit = {
-    Preconditions.checkArgument(
-      properties.get(TableCatalog.PROP_PROVIDER) != null,
-      "%s requires '%s' to be set",
-      operation,
-      TableCatalog.PROP_PROVIDER)
-  }
 }
 
 object UCSingleCatalog {
@@ -445,6 +436,15 @@ object UCSingleCatalog {
     props.putAll(credentialProps.map {
       case (k, v) => (prefix + k, v)
     }.asJava)
+  }
+
+  def requireProviderSpecified(
+      operation: String,
+      properties: util.Map[String, String]): Unit = {
+    Preconditions.checkArgument(
+      properties.get(TableCatalog.PROP_PROVIDER) != null,
+      "%s requires USING <format> (for example, USING DELTA)",
+      operation)
   }
 
   /**
@@ -623,11 +623,7 @@ private class UCProxy(
 
   override def createTable(ident: Identifier, schema: StructType, partitions: Array[Transform], properties: util.Map[String, String]): Table = {
     UCSingleCatalog.checkUnsupportedNestedNamespace(ident.namespace())
-    Preconditions.checkArgument(
-      properties.get(TableCatalog.PROP_PROVIDER) != null,
-      "%s requires '%s' to be set",
-      "CREATE TABLE",
-      TableCatalog.PROP_PROVIDER)
+    UCSingleCatalog.requireProviderSpecified("CREATE TABLE", properties)
 
     val createTable = new CreateTable()
     createTable.setName(ident.name())
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/DeltaManagedTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/DeltaManagedTableReadWriteTest.java
@@ -204,6 +204,25 @@ public abstract class DeltaManagedTableReadWriteTest extends BaseTableReadWriteT
                 "CREATE OR REPLACE TABLE %s (i INT, extra_col INT) USING DELTA", fullTableName)));
   }
 
+  @Test
+  public void testManagedDeltaReplaceRejectsProviderChangeWithoutDroppingTable()
+      throws ApiException {
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG, CATALOG_NAME);
+    ensureSparkCatalogSchemaExists();
+    String fullTableName =
+        setupTable(new TableSetupOptions().setCatalogName(CATALOG_NAME).setTableName(TEST_TABLE));
+    sql("INSERT INTO %s SELECT 1, 'old'", fullTableName);
+    String originalTableId = loadTableInfo(fullTableName).getTableId();
+
+    assertThatThrownBy(
+            () -> sql("REPLACE TABLE %s USING PARQUET AS SELECT 2 AS i, 'new' AS s", fullTableName))
+        .hasMessageContaining("requires USING DELTA")
+        .hasMessageContaining("Cannot change table format from DELTA to PARQUET");
+
+    validateRows(sql("SELECT * FROM %s", fullTableName), Pair.of(1, "old"));
+    assertManagedTableHasUcProperties(loadTableInfo(fullTableName), originalTableId);
+  }
+
   @Override
   protected void prepareExistingTableForDdl(TableSetupOptions options) {
     TableSetupOptions existingOptions =
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogStagingTableTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogStagingTableTest.java
@@ -256,9 +256,57 @@ public class UCSingleCatalogStagingTableTest {
                         UCTableProperties.DELTA_CATALOG_MANAGED_KEY_NEW,
                         UCTableProperties.DELTA_CATALOG_MANAGED_VALUE)))
         .isInstanceOf(ApiException.class)
-        .hasMessageContaining("cannot change provider from DELTA to PARQUET");
+        .hasMessageContaining("requires USING DELTA")
+        .hasMessageContaining("Cannot change table format from DELTA to PARQUET");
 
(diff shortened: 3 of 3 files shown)


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -226,8 +227,16 @@ class UCSingleCatalog
         s"Invalid table metadata for $fullTableName: tableId must be set")
     }
     // Third, build the properties Delta needs in order to write back to the current managed table.
+    val existingProvider = tableInfo.getDataSourceFormat.getValue.toLowerCase(Locale.ROOT)
+    Option(properties.get(TableCatalog.PROP_PROVIDER))
```
yili-db, line 231: In `createTable` there is:
```
assert(properties.get(TableCatalog.PROP_PROVIDER) != null)
```
Should the stage* functions also check this?

timothyw553: I've added the check for stageReplace and stageCreateOrReplace

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -409,6 +420,13 @@ class UCSingleCatalog
       stagingCatalog.stageCreate(ident, schema, partitions, properties)
     }
   }
+
+  private def assertProviderSpecified(
+      operation: String,
+      properties: util.Map[String, String]): Unit = {
+    assert(properties.get(TableCatalog.PROP_PROVIDER) != null,
```
yili-db: Please also call this function from `createTable`.
Also the `assert` isn't good here. It's usually meant for impossible situations.
Should use `Preconditions.checkArgument` here.

timothyw553: done. and noted for the future, thank you.

yili-db (approved):

connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogStagingTableTest.java
```diff
@@ -215,6 +215,52 @@ public void testStageReplaceExistingManagedTablePreservesUserProperties() throws
     verify(mocks.tempCredsApi).generateTemporaryTableCredentials(any());
   }
 
+  @Test
```
openinx, line 218: [AI Generated] UCSingleCatalogStagingTableTest.java — Missing stageCreateOrReplace counterpart tests.
                                                                                                                                                                                                                 
  The two new tests only cover the stageReplace path. The stageCreateOrReplace method with an existing table goes through the same loadExistingManagedTablePropsForReplace, so the same omitted-provider
  injection and provider-change rejection apply — but there are no tests for it. Please add:                                                                                                                     
  - testStageCreateOrReplaceExistingManagedTableInjectsDeltaProviderWhenOmitted
  - testStageCreateOrReplaceExistingManagedTableRejectsProviderChange

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -226,8 +227,16 @@ class UCSingleCatalog
         s"Invalid table metadata for $fullTableName: tableId must be set")
     }
     // Third, build the properties Delta needs in order to write back to the current managed table.
+    val existingProvider = tableInfo.getDataSourceFormat.getValue.toLowerCase(Locale.ROOT)
+    Option(properties.get(TableCatalog.PROP_PROVIDER))
+      .filterNot(_.equalsIgnoreCase(existingProvider))
+      .foreach(provider => throw new ApiException(
+        s"$operation on existing UC-managed Delta table $fullTableName cannot change provider " +
```
openinx: As this error message is user-facing, could you please make this more explicit and saying it requires the `USING DELTA` clause ?

openinx: Otherwise, the users may confuse the error message and have to read the code.

connectors/spark/src/test/java/io/unitycatalog/spark/UCSingleCatalogStagingTableTest.java
```diff
@@ -215,6 +215,52 @@ public void testStageReplaceExistingManagedTablePreservesUserProperties() throws
     verify(mocks.tempCredsApi).generateTemporaryTableCredentials(any());
```
openinx, line 215: I think we need a e2e test test case to simulate the whole process that we encountered in the bug. I mean previously if we replace the delta table with non-delta table, it will fail and drop table. we should add a test to clearly say it won't drop the table again.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -602,7 +623,11 @@ private class UCProxy(
 
   override def createTable(ident: Identifier, schema: StructType, partitions: Array[Transform], properties: util.Map[String, String]): Table = {
     UCSingleCatalog.checkUnsupportedNestedNamespace(ident.namespace())
-    assert(properties.get(TableCatalog.PROP_PROVIDER) != null)
+    Preconditions.checkArgument(
+      properties.get(TableCatalog.PROP_PROVIDER) != null,
+      "%s requires '%s' to be set",
```
openinx: I think from the end-users,  they actually don't understand what's the PROVIDER.  we'd better clearly say `USING delta` is required. or just give a example, if the catalog support to other providers ( for delta, parquet etc). w

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -409,6 +420,16 @@ class UCSingleCatalog
       stagingCatalog.stageCreate(ident, schema, partitions, properties)
     }
   }
+
+  private def requireProviderSpecified(
+      operation: String,
+      properties: util.Map[String, String]): Unit = {
+    Preconditions.checkArgument(
+      properties.get(TableCatalog.PROP_PROVIDER) != null,
+      "%s requires '%s' to be set",
```
openinx: This is the same.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -602,7 +623,11 @@ private class UCProxy(
 
   override def createTable(ident: Identifier, schema: StructType, partitions: Array[Transform], properties: util.Map[String, String]): Table = {
     UCSingleCatalog.checkUnsupportedNestedNamespace(ident.namespace())
-    assert(properties.get(TableCatalog.PROP_PROVIDER) != null)
+    Preconditions.checkArgument(
```
openinx: Could we share the same `requireProviderSpecified` here ?  I see the code is almost the same.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: question is should we only gate "DELTA" or just whatever `existingTable.provider` is. I opted for `existingTable.provider`

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: ok explicitly added `"$operation requires USING DELTA"` if it is null.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: done

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: yes, done. just called it for createTable

timothyw553:
FYI: the `null` check for the REPLACE path does not fail if the underlying/existing table is still DELTA.

This is because for REPLACE path we:
1. Get properties from existing table, and get the existing provider (e.g., DELTA)
2. If we explicitly change it (e.g., USING PARQUET) it will fail
3. Inject the provider into the properties.

THEN we do a null check. The invariant here is that the provider will be DELTA.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -602,7 +623,7 @@ private class UCProxy(
 
   override def createTable(ident: Identifier, schema: StructType, partitions: Array[Transform], properties: util.Map[String, String]): Table = {
     UCSingleCatalog.checkUnsupportedNestedNamespace(ident.namespace())
-    assert(properties.get(TableCatalog.PROP_PROVIDER) != null)
+    UCSingleCatalog.requireProviderSpecified("CREATE TABLE", properties)
```
openinx, line 626: For create table,  we should support all the table types, including delta tables, and paurquet/csv/... external tables.  So the error message should be wrong if we are creating a non-delta table. 

```java
  def requireProviderSpecified(
      operation: String,
      properties: util.Map[String, String]): Unit = {
    Preconditions.checkArgument(
      properties.get(TableCatalog.PROP_PROVIDER) != null,
      "%s requires USING DELTA",
      operation)
  }
```

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaManagedTableReadWriteTest.java
```diff
@@ -204,6 +204,24 @@ public void testManagedDeltaReplaceRejectsMetadataChanges() {
...
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG, CATALOG_NAME);
+    ensureSparkCatalogSchemaExists();
+    String fullTableName =
+        setupTable(new TableSetupOptions().setCatalogName(CATALOG_NAME).setTableName(TEST_TABLE));
+    sql("INSERT INTO %s SELECT 1, 'old'", fullTableName);
+    String originalTableId = loadTableInfo(fullTableName).getTableId();
+
+    assertThatThrownBy(
+            () -> sql("REPLACE TABLE %s USING PARQUET AS SELECT 2 AS i, 'new' AS s", fullTableName))
+        .hasMessageContaining("requires USING DELTA");
```
openinx: This error message `requires USING DELTA` for `REPLACE TABLE %s USING PARQUET` is weird.  I think it's good for use clearly say:   REPLACE TABLE is only supported for unitycatalog managed delta table, if you want to replace the existing table with a new delta table, it always requires the `USING DELTA` clause.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -226,8 +227,16 @@ class UCSingleCatalog
         s"Invalid table metadata for $fullTableName: tableId must be set")
     }
     // Third, build the properties Delta needs in order to write back to the current managed table.
+    val existingProvider = tableInfo.getDataSourceFormat.getValue.toLowerCase(Locale.ROOT)
+    Option(properties.get(TableCatalog.PROP_PROVIDER))
+      .filterNot(_.equalsIgnoreCase(existingProvider))
+      .foreach(provider => throw new ApiException(
+        s"$operation on existing UC-managed Delta table $fullTableName requires USING DELTA " +
+          s"and cannot change provider from ${existingProvider.toUpperCase(Locale.ROOT)} to " +
```
openinx: `cannot change provider from` -> `Cannot change table format from `.  The users actually don't have concept about the provider ?  but they have the concept of table format.

PR #1406, opened 2026-03-19
Replace hardcoded logging config with classpath-based Log4j2 auto-discovery

arturowczarek:
**PR Checklist**

**Description of changes**

Replace hardcoded logging config with classpath-based Log4j2 auto-discovery

Solves #1407

---

The server and CLI used `static {}` initializer blocks to hardcode Log4j2 configuration to
`etc/conf/server.log4j2.properties` via `Configurator.initialize()`. This caused several
problems:

- **Tests had no control over logging** — the static block fired on class load, locking logging to a production file-based config with no console output
- **Relative `etc/conf/` path was fragile** — broke when CWD wasn't the project root (IDE runs, Docker, embedding the server as a library)
- **No runtime configurability** — changing log level or directory required editing a file; no system property overrides
- **Errors hidden in tests** — the file-only appender wrote to `etc/logs/server.log` where nobody checked during test runs

## Changes

**Core logging changes:**
- Remove `static {}` blocks and `Configurator` import from `UnityCatalogServer` and `UnityCatalogCli`
- Add `server/src/main/resources/log4j2.properties` — classpath-based config with console + rolling file appenders, system property overrides (`uc.log.level`, `uc.log.dir`), and per-package level tuning for noisy libraries (Hibernate, H2, Armeria)
- Add `server/src/test/resources/log4j2-test.properties` — console-only, quieter config auto-discovered by Log4j2's naming convention
- Add `examples/cli/src/main/resources/log4j2.properties` — console-only config for CLI

**Dependency fixes in `build.sbt`:**
- Remove `slf4j-log4j12` (Log4j 1.x bridge that conflicted with the Log4j2 setup)
- Add `log4j-core` (required runtime dependency, was only transitively available)

**Helm chart updates:**
- Remove `log4j2ConfigTemplate` from `_config.tpl` and its ConfigMap entry. How Helm uses the default 
- Remove `envsubst` step for `server.log4j2.properties` in init container
- Add `JAVA_OPTS` env var with `-Duc.log.level` to the server container, driven by existing `server.logLevel` Helm value.

**Removed files:**
- `etc/conf/server.log4j2.properties`
- `etc/conf/cli.log4j2.properties`
They are no longer needed. Either the auto-discovered defaults are used or the ones configured as system properties.

## Test plan

- [x] `sbt compile` passes
- [x] `sbt "server/testOnly io.unitycatalog.server.sdk.catalog.SdkCatalogCRUDTest"` passes
  with correct console output (UC logs at INFO, Hibernate/H2 quiet)
- [x] Verify `JAVA_OPTS=-Duc.log.level=debug bin/start-uc-server` produces debug output
- [x] Verify `bin/start-uc-server` without JAVA_OPTS uses info level defaults
- [x] When used in other project the configuration file with `.test` is used
- [x] Helm template renders correctly (`helm template` dry-run)


arturowczarek:
@openinx Do you mind doing a review and merging it? Logging has been set up in the second commit in this repository and deserves an update.

server/src/main/resources/log4j2.properties
```diff
@@ -0,0 +1,41 @@
...
+appender.console.type = Console
+appender.console.name = Console
+appender.console.layout.type = PatternLayout
+appender.console.layout.pattern = %d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n
+
+# Rolling file appender
+appender.rolling.type = RollingFile
+appender.rolling.name = RollingFile
+appender.rolling.fileName = ${sys:uc.log.dir:-logs}/server.log
+appender.rolling.filePattern = ${sys:uc.log.dir:-logs}/server-%d{yyyy-MM-dd}-%i.log.gz
```
timothyw553, line 14: is `$HOME/logs` writable by the runtime user? I'm not totally sure, could you verify or perhaps write to `$HOME/etc`? and could you add a packaged-image test that starts the server and verifies the log file is created.

timothyw553:
```suggestion
appender.rolling.fileName = ${sys:uc.log.dir:-etc/logs}/server.log
appender.rolling.filePattern = ${sys:uc.log.dir:-etc/logs}/server-%d{yyyy-MM-dd}-%i.log.gz
```

examples/cli/src/main/resources/log4j2.properties
timothyw553, line 1: perhaps can we rename this to `examples/cli/src/main/resources/log4j2-cli.properties`

server/src/main/resources/log4j2.properties
timothyw553, line 1: can we rename this to `examples/cli/src/main/resources/log4j2-server.properties`

docs/server/logging.md
```diff
@@ -0,0 +1,90 @@
...
+bin/start-uc-server -Duc.log.dir=/var/log/unitycatalog
+```
+
+## Using a Custom Configuration File
+
+For full control over logging (appenders, filters, per-package levels), provide your own
+Log4j2 configuration file using the standard `log4j.configurationFile` system property:
+
+```sh
+bin/start-uc-server -Dlog4j.configurationFile=/path/to/my-log4j2.properties
```
timothyw553, line 44: nit: should it be `JAVA_OPTS="-Duc.log.level=debug" bin/start-uc-server` for example? and for others
- `JAVA_OPTS="-Duc.log.dir=/var/log/unitycatalog" bin/start-uc-server`
- `JAVA_OPTS="-Duc.log.level=debug" bin/uc <command>`
- `CLI_JAVA_COMMAND="java $JAVA_OPTS -cp $(cat "$1") io.unitycatalog.cli.UnityCatalogCli"`

server/src/test/resources/log4j2-test.properties
```diff
@@ -0,0 +1,23 @@
...
+appender.console.layout.type = PatternLayout
+appender.console.layout.pattern = %d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n
+
+rootLogger.level = warn
+rootLogger.appenderRefs = console
+rootLogger.appenderRef.console.ref = Console
+
+# UC server logs at info level during tests
+logger.uc.name = io.unitycatalog
+logger.uc.level = info
```
timothyw553, line 16: should it be like this?
```suggestion
logger.uc.level = ${sys:uc.log.level:-info}
```

and perhaps change build.sbt correspondingly 
```sbt
  Test / javaOptions ++=
    sys.props.get("uc.log.level").map(v => s"-Duc.log.level=$v").toSeq
```

timothyw553: and add a test asserting the effective logger level with and without the override

timothyw553 (commented):
I left some review comments, please take a look :) @arturowczarek

Issue #1385, opened 2026-03-07
PostgreSQL: DELETE ... LIMIT causes syntax error in DeltaCommitRepository


MarcRevo:
## Summary
UPDATE, DELETE, and DROP TABLE on managed Delta tables fail when Unity Catalog uses PostgreSQL as metadata database. The server executes `DELETE ... LIMIT` which PostgreSQL does not support.

## Environment
- **Unity Catalog server**: 0.4.0 
- **Spark**: 4.0.1
- **Delta Lake**: 4.0.1
- **PostgreSQL**: 17.7
- **Database**: PostgreSQL (configured via hibernate.properties)

### 1. Create managed table and insert data
CREATE TABLE IF NOT EXISTS development.sales.orders_xx (
  order_id STRING,
  amount DOUBLE,
  order_ts TIMESTAMP
) USING DELTA TBLPROPERTIES ('delta.feature.catalogManaged'='supported');

INSERT INTO development.sales.orders_xx
VALUES 
  ('o-001', 29.99, current_timestamp()),
  ('o-002', 15.50, current_timestamp()),
  ('o-003', 99.00, current_timestamp());

### 2. Try an update
UPDATE development.sales.orders_xx
SET amount = 39.99
WHERE order_id = 'o-001';

### Client (Spark / PySpark)
Py4JJavaError: An error occurred while calling o66.sql.: io.delta.storage.commit.CommitFailedException: Unexpected commit failure (HTTP 500): {"error_code":"INTERNAL","details":[{"reason":"INTERNAL","metadata":{},\"@type\":\"google.rpc.ErrorInfo\"}],"stack_trace":null,"message":"Error committing to table: d4749687-e73e-4fec-b501-82ed26f21889: JDBC exception executing SQL [/* dynamic native SQL query */ DELETE FROM uc_delta_commits WHERE table_id = ? AND commit_version <= ? LIMIT ?] [ERROR: syntax error at or near \"LIMIT\"\n  Position: 106] [n/a]"}

### Unity Catalog server pod
armeria-common-worker-epoll-3-21 ERROR ERROR: syntax error at or near "LIMIT"  Position: 106

### PostgreSQL pod
ERROR:  syntax error at or near "LIMIT" at character 106STATEMENT:  /* dynamic native SQL query */ DELETE FROM uc_delta_commits WHERE table_id = $1 AND commit_version <= $2 LIMIT $3
For DROP TABLE, PostgreSQL shows:
[...]

PR #1446, opened 2026-03-30
fix(server): use subquery for DELETE batch to support PostgreSQL

tiennguyen-onehouse:
## Summary

Fixes PostgreSQL compatibility for Delta commit operations. Addresses #1385.

### `DeltaCommitRepository.java` — Fix `DELETE...LIMIT` syntax (core bug fix)

PostgreSQL does not support `LIMIT` in `DELETE` statements, causing `syntax error at or near "LIMIT"` for any UPDATE, DELETE, or DROP TABLE on managed Delta tables. Replaced `DELETE FROM uc_delta_commits WHERE ... LIMIT` with a cross-database subquery pattern: `DELETE FROM ... WHERE id IN (SELECT id FROM (SELECT id FROM ... LIMIT) AS batch_to_delete)`. Two methods fixed: `deleteCommitsUpTo()` and `deleteCommits()`.

### `DeltaCommitDAO.java` — Remove `columnDefinition = "BINARY(16)"` (PostgreSQL blocker)

PostgreSQL does not have a `BINARY` type — the old code prevented the server from even starting on PostgreSQL because Hibernate fails at schema creation. `columnDefinition` is a literal SQL string with no per-dialect fallback, so there is no way to keep it and support PostgreSQL. Removing it lets Hibernate 6.5 auto-map `java.util.UUID` to the correct type per dialect (`UUID` on H2, `uuid` on PostgreSQL, `binary(16)` on MySQL). This aligns `DeltaCommitDAO` with every other DAO in the codebase — it was the only one using `columnDefinition`. See compatibility analysis below.

### `UnityCatalogServer.java` / `HibernateConfigurator.java` — Dependency-inject the test database (per review feedback)

`UnityCatalogServer.Builder` now accepts an optional `HibernateConfigurator` (default `null`, in which case the server builds its own from server properties — production behavior unchanged). `HibernateConfigurator` gained a constructor taking explicit hibernate `Properties`, and its test-env block is back to plain hardcoded H2.

`BaseServerTest` builds one `HibernateConfigurator`, passes it to the server builder (so the test harness and the server share a single `SessionFactory` instead of each creating their own), and exposes a `setUpHibernateProperties(Properties)` hook that database-specific test subclasses override. This resolves the long-standing `TODO: use dependency injection for test hibernate properties` and keeps `server.properties` / `hibernate.properties` as independent config namespaces (earlier revisions of this PR routed the overrides through system properties, then `ServerProperties` — both reverted).

### `build.sbt` — Add test dependencies

Added Testcontainers (postgresql, mysql, junit-jupiter) + PostgreSQL and MySQL JDBC drivers, all test-scope only.

### `PostgresDeltaCommitsCRUDTest.java` — New PostgreSQL integration test

Extends `SdkDeltaCommitsCRUDTest` to inherit all 10 test methods and runs them against a real PostgreSQL 16 database via Testcontainers. Uses `@Testcontainers(disabledWithoutDocker = true)` so the test is automatically skipped when Docker is not available — no build configuration changes needed. Exercises both fixed DELETE methods through backfill and table deletion flows.

### `MySQLDeltaCommitsCRUDTest.java` — New MySQL integration test

Same pattern against MySQL 8.4. MySQL is the dialect that actually constrains the batch-DELETE shape (it rejects `LIMIT` inside an `IN(...)` subquery — error 1235 — and a subquery reading the table being deleted from — error 1093), so this suite is the only thing that guards the derived-table wrapper against a refactor that H2 and PostgreSQL would happily accept. It also converts the "MySQL: zero behavioral change" claim in the compatibility analysis below from reasoning into a tested result. Happy to drop this test if you feel it stretches the PR's scope — it's an independent commit.

## `columnDefinition = "BINARY(16)"` removal — compatibility analysis

`DeltaCommitDAO` was the **only DAO** in the codebase using `columnDefinition = "BINARY(16)"` for UUID fields. All other DAOs (e.g., `PropertyDAO`, `CatalogInfoDAO`) already rely on Hibernate's automatic UUID type mapping. This change aligns `DeltaCommitDAO` with the rest of the codebase.

### Why this is needed

`columnDefinition` is a literal SQL string with no dialect fallback. There is no JPA/Hibernate mechanism to specify different column definitions per database (e.g., `BINARY(16)` for H2, `uuid` for PostgreSQL). Without removing it, PostgreSQL users cannot start the server at all — Hibernate fails at schema creation because `BINARY(16)` is not valid PostgreSQL syntax.

### What Hibernate 6.5 generates without `columnDefinition`

| Database   | Column type generated | Old code (`BINARY(16)`) | Change? |
|------------|----------------------|------------------------|---------|
| H2         | `UUID`               | `BINARY(16)`           | Yes     |
| PostgreSQL | `uuid`               | `BINARY(16)` (invalid) | Fix     |
| MySQL      | `binary(16)`         | `BINARY(16)`           | **No**  |

- **MySQL**: Zero behavioral change. Hibernate 6.x maps `java.util.UUID` to `binary(16)` on MySQL regardless of `columnDefinition`.
- **PostgreSQL**: `BINARY(16)` does not exist in PostgreSQL. The old code was **already broken** — the server could never start. This change is the fix.
- **H2**: Column type changes from `BINARY(16)` to `UUID`. See upgrade/downgrade analysis below.

### H2 upgrade/downgrade compatibility (verified empirically with H2 2.2.224)

**Upgrade** (existing `BINARY(16)` columns → new code without `columnDefinition`):
- `hbm2ddl.auto=update` (used in production) never alters existing column types — existing `BINARY(16)` columns remain as-is
- JDBC read/write of `java.util.UUID` works correctly through `BINARY(16)` columns
- All 4 operations tested: INSERT, READ, QUERY, DELETE with subquery — all pass

**Downgrade** (new `UUID` columns → old code with `columnDefinition = "BINARY(16)"`):
- `hbm2ddl.auto=update` does not alter existing `UUID` columns back to `BINARY(16)`
- `columnDefinition` only affects DDL generation (CREATE TABLE), not runtime data access
- JDBC read/write of `java.util.UUID` works correctly through `UUID` columns regardless of `columnDefinition`
- All 4 operations tested: INSERT, READ, QUERY, DELETE with subquery — all pass

### Cross-database compatibility test results (17/17 passed)

```
================================================================
H2: Old code=BINARY(16), New code=UUID (type CHANGES)
================================================================
  PASS: H2 downgrade: insert into UUID col
  PASS: H2 downgrade: read UUID from UUID col
  PASS: H2 downgrade: query by UUID
  PASS: H2 downgrade: delete with subquery
  PASS: H2 upgrade: insert UUID into BINARY(16) col
  PASS: H2 upgrade: read UUID from BINARY(16) col
  PASS: H2 upgrade: query by UUID on BINARY(16) col
  PASS: H2 upgrade: delete with subquery on BINARY(16)

================================================================
PostgreSQL: Old code=BINARY(16) which is INVALID in PostgreSQL.
  New code=uuid. No downgrade scenario possible.
================================================================
  PASS: PostgreSQL: BINARY(16) correctly rejected (type "binary" does not exist)
  PASS: PostgreSQL: insert UUID
  PASS: PostgreSQL: read UUID
  PASS: PostgreSQL: query by UUID
  PASS: PostgreSQL: delete with subquery

================================================================
MySQL: Old code=BINARY(16), New code=binary(16) (NO CHANGE)
================================================================
  PASS: MySQL: insert UUID as bytes into binary(16)
  PASS: MySQL: read UUID from binary(16)
  PASS: MySQL: query by UUID on binary(16)
  PASS: MySQL: delete with subquery on binary(16)

TOTAL: 17 passed, 0 failed
```

## How we tested

**1. Existing H2 unit tests (10/10 passed):**

`SdkDeltaCommitsCRUDTest` — exercises backfill (triggers `deleteCommitsUpTo`) and table deletion (triggers `deleteCommits`).

**2. New PostgreSQL + MySQL integration tests (10/10 each):**

`PostgresDeltaCommitsCRUDTest` and `MySQLDeltaCommitsCRUDTest` — use [Testcontainers](https://testcontainers.com/) to spin up real PostgreSQL 16 / MySQL 8.4 databases and run all 10 DeltaCommitsCRUD tests against each. Automatically skipped when Docker is unavailable (`disabledWithoutDocker = true`).

```bash
# Run the database integration tests (requires Docker)
build/sbt "server/testOnly *PostgresDeltaCommitsCRUDTest *MySQLDeltaCommitsCRUDTest"
```

```
SdkDeltaCommitsCRUDTest (H2):      10/10 passed
PostgresDeltaCommitsCRUDTest:      10/10 passed
MySQLDeltaCommitsCRUDTest:         10/10 passed
Passed: Total 30, Failed 0, Errors 0, Passed 30
```

Note: this repo currently runs no test CI on PRs (only DCO), so the results above are from local runs at the current head.

**3. Static analysis test script (22/22 passed):**

[Test script gist](https://gist.github.com/tiennguyen-onehouse/760d0f685ff686fc28eed53c3616d648) validates the SQL patterns are structurally correct.

**4. Cross-database compatibility test (17/17 passed):**

Standalone JDBC test verifying UUID column type interoperability across H2, PostgreSQL, and MySQL — covering both upgrade and downgrade scenarios. Results shown above.

## Test plan

- [x] H2 unit tests pass (10/10)
- [x] PostgreSQL integration tests pass (10/10)
- [x] MySQL integration tests pass (10/10)
- [x] Static analysis checks pass (22/22)
- [x] Cross-database UUID compatibility verified (17/17)
- [x] No other `DELETE...LIMIT` patterns exist in the codebase (verified via grep)


tiennguyen-onehouse:
Hi maintainers — this PR adds a Testcontainers-based PostgreSQL integration test alongside the SQL fix. It uses `@Testcontainers(disabledWithoutDocker = true)` so it auto-skips when Docker is not available and doesn't affect existing test runs.

If integration tests are not the practice of this repo, happy to remove the test and keep the PR to just the production code changes (DeltaCommitRepository, DeltaCommitDAO, and HibernateConfigurator).

tiennguyen-onehouse:
One note on the SQL change — the subquery pattern adds a small overhead compared to the original `DELETE...LIMIT`:

- **Before**: single pass scan → filter → delete (up to batch size)
- **After**: inner SELECT finds matching IDs (same scan + filter + LIMIT), then DELETE by primary key lookup

The extra cost is an `IN (...)` lookup on the primary key, which is fast. This also only runs during background batch cleanup (`NUM_COMMITS_PER_BATCH`), not on a hot path. We don't expect a noticeable difference, but let us know if performance here is a concern.

MarcRevo:
@tiennguyen-onehouse Thanks a lot for your work related to my issue

MarcRevo:
Just to confirm, this would also resolve the issue when performing an UPDATE on a managed Delta table we are seeing the same syntax error at or near "LIMIT" in DELETE FROM uc_delta_commits during the commit phase. Good to know this patch covers both cases.

build.sbt
```diff
@@ -395,6 +395,12 @@ lazy val server = (project in file("server"))
         exclude("org.apache.logging.log4j", "log4j-to-slf4j"),
       "javax.xml.bind" % "jaxb-api" % "2.3.1" % Test,
 
+      // Integration testing
+      "org.testcontainers" % "testcontainers" % "1.19.8" % Test,
+      "org.testcontainers" % "postgresql" % "1.19.8" % Test,
+      "org.testcontainers" % "junit-jupiter" % "1.19.8" % Test,
+      "org.postgresql" % "postgresql" % "42.7.3" % Test,
```
yili-db: Can you please bump it up to `42.7.12`?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -614,14 +614,25 @@ private static int deleteCommitsUpTo(Session session, UUID tableId, long upToCom
...
+    NativeQuery<?> query = session.createNativeQuery(buildBatchDeleteQuery("table_id = :tableId"));
     query.setParameter("tableId", tableId);
     query.setParameter("numCommitsPerBatch", NUM_COMMITS_PER_BATCH);
     return query.executeUpdate();
   }
 
+  /**
+   * Builds a batch DELETE query using a subquery instead of DELETE...LIMIT for cross-database
+   * compatibility. PostgreSQL does not support LIMIT in DELETE statements directly.
+   */
```
yili-db, line 733: Please also comment on why there are two `SELECT` layers.

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/PostgresDeltaCommitsCRUDTest.java
```diff
@@ -0,0 +1,39 @@
...
+  @Container
+  private static final PostgreSQLContainer<?> POSTGRES =
+      new PostgreSQLContainer<>("postgres:16-alpine")
+          .withDatabaseName("unitycatalog_test")
+          .withUsername("test")
+          .withPassword("test");
+
+  @BeforeAll
+  static void setUpPostgres() {
+    POSTGRES.start();
```
yili-db: Is `start` redundant? @Testcontainers + @Container already manage lifecycle supposedly.

yili-db (commented):
Overall it's a good fix. Left some minor comments. Thank you!

build.sbt
tiennguyen-onehouse: Done — bumped to 42.7.12 in 9c62d26.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
tiennguyen-onehouse: Good catch — documented in the Javadoc (9335d3f, wording simplified in e8a28c5). Short version: the outer derived-table layer exists purely for MySQL, which rejects both LIMIT directly inside an IN(...) subquery (error 1235) and a subquery that reads the table being deleted from (error 1093). Wrapping the inner SELECT in `... FROM (...) AS batch_to_delete` materializes the ids into a temp table first, which sidesteps both. PostgreSQL and H2 would accept the flat single-subquery form, so the Javadoc now warns against flattening it — the H2/Postgres tests would still pass but MySQL would fail at runtime.

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/PostgresDeltaCommitsCRUDTest.java
tiennguyen-onehouse: You're right — it was redundant. The @Testcontainers extension's BeforeAllCallback starts static @Container fields before any user @BeforeAll runs, so the manual start() was a no-op. Removed in f798ad4, which also replaces the @BeforeAll/@AfterAll System.setProperty lifecycle entirely: the test now overrides BaseServerTest.setUpProperties() and passes the connection settings through ServerProperties, so the Postgres config is scoped to this fixture instead of JVM-global system properties (and HibernateConfigurator's pre-existing TODO about injecting test hibernate properties is resolved as a side effect).

tiennguyen-onehouse (commented):
Thank you @yili-db for the review! I have addressed the comments. I also found a resource leak in UnityCatalogServer. I opened this PR. https://github.com/unitycatalog/unitycatalog/pull/1660/changes. Let me know if you can review that as well.

server/src/main/java/io/unitycatalog/server/persist/utils/HibernateConfigurator.java
```diff
@@ -93,18 +91,32 @@ public static Properties setupHibernateProperties(ServerProperties serverPropert
...
       hibernateProperties.setProperty(
-          "hibernate.connection.url", "jdbc:h2:mem:testdb;DB_CLOSE_DELAY=-1");
+          "hibernate.connection.driver_class",
+          serverProperties.getProperty("hibernate.connection.driver_class", "org.h2.Driver"));
+      hibernateProperties.setProperty(
+          "hibernate.connection.url",
+          serverProperties.getProperty(
+              "hibernate.connection.url", "jdbc:h2:mem:testdb;DB_CLOSE_DELAY=-1"));
+      for (String key :
+          new String[] {"hibernate.connection.username", "hibernate.connection.password"}) {
```
yili-db: `List.of()`

server/src/main/java/io/unitycatalog/server/persist/utils/HibernateConfigurator.java
```diff
@@ -93,18 +91,32 @@ public static Properties setupHibernateProperties(ServerProperties serverPropert
...
 
-    // TODO: use dependency injection for test hibernate properties
     if ("test".equals(serverProperties.get(Property.SERVER_ENV))) {
-      hibernateProperties.setProperty("hibernate.connection.driver_class", "org.h2.Driver");
+      // Connection overrides let a test point Hibernate at an external database
+      // (e.g., PostgreSQL via Testcontainers). Defaults to H2 in-memory when not overridden.
       hibernateProperties.setProperty(
-          "hibernate.connection.url", "jdbc:h2:mem:testdb;DB_CLOSE_DELAY=-1");
+          "hibernate.connection.driver_class",
+          serverProperties.getProperty("hibernate.connection.driver_class", "org.h2.Driver"));
```
yili-db: I don't think it should be done this way.
`serverProperties` corresponds to `./etc/conf/server.properties`.
`hibernateProperties` corresponds to `./etc/conf/hibernate.properties`.
It's awkward to override values in one file with values from another.

Here the purpose is to override `hibernateProperties` for `PostgresDeltaCommitsCRUDTest`. This should be done in a different way: `BaseServerTest` have a function that overrides the `hibernateProperties` and `PostgresDeltaCommitsCRUDTest` can override it. However this won't work neither since it only changes the `HibernateConfigurator` of `BaseServerTest` but `UnityCatalogServer` uses a different one.
In order to get it working, `UnityCatalogServer.Builder` needs to store this customized `HibernateConfigurator` (default=null) and feed it from `BaseServerTest` to `UnityCatalogServer` (they will share the same `HibernateConfigurator`).

Unfortunately this is becoming more complicated to properly setup postgres testing.

tiennguyen-onehouse: Done in 1692a460 — implemented as you suggested: `UnityCatalogServer.Builder` now takes an optional `HibernateConfigurator` (default null keeps the current server behavior), `BaseServerTest` builds one and feeds it to the builder so the test harness and the server share the same `HibernateConfigurator`/session factory, and exposes a `setUpHibernateProperties(Properties)` hook that `PostgresDeltaCommitsCRUDTest` overrides. `ServerProperties` is fully reverted to its original form, so `server.properties` and `hibernate.properties` stay independent. All suites pass locally (30/30 across H2, PostgreSQL, and MySQL).

server/src/main/java/io/unitycatalog/server/persist/utils/HibernateConfigurator.java
tiennguyen-onehouse: This loop is gone entirely in 1692a460 — the test-env block is back to plain hardcoded H2, with the database override now happening through the injected `HibernateConfigurator` instead.

tiennguyen-onehouse:
Heads-up on the latest push (e08e5baa): I added `MySQLDeltaCommitsCRUDTest` alongside the PostgreSQL one (~35 lines: a `MySQLContainer` subclass reusing the same `setUpHibernateProperties` hook, plus two test-scope deps).

Rationale: MySQL is the dialect that actually constrains the batch-DELETE shape in `DeltaCommitRepository` — it rejects `LIMIT` inside an `IN(...)` subquery (error 1235) and a subquery reading the table being deleted from (error 1093). H2 and PostgreSQL both accept the flattened form, so before this suite the "do not flatten" warning in the Javadoc was enforced by nothing — a refactor could pass every existing test and still break MySQL at runtime. It also turns the "MySQL: zero behavioral change" claim from the `BINARY(16)` compatibility analysis into a tested result.

All three suites pass locally: 30/30 (H2, PostgreSQL 16, MySQL 8.4). Like the Postgres suite it auto-skips when Docker is unavailable. If you feel it stretches the scope of this PR, happy to drop it — it's an isolated commit that reverts cleanly.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -410,6 +383,18 @@ public UnityCatalogServer.Builder serverProperties(ServerProperties serverProper
       return this;
     }
 
+    /**
+     * Uses the given {@link HibernateConfigurator} instead of creating one from the server
+     * properties. Lets tests share the server's session factory and customize the hibernate
+     * properties (e.g. run against PostgreSQL via Testcontainers). The caller keeps ownership:
+     * stopping the server does not close the injected configurator's session factory.
```
yili-db: FYI this ownership doc will be violated by the other PR https://github.com/unitycatalog/unitycatalog/pull/1660 once it closes the session factory when server closes.

yili-db (approved):

server/src/main/java/io/unitycatalog/server/persist/dao/DeltaCommitDAO.java
```diff
@@ -30,10 +30,10 @@ public class DeltaCommitDAO {
   // optionally commit_version. But this is a required field as a unique ID in the database.
   @Id
   @UuidGenerator
-  @Column(name = "id", columnDefinition = "BINARY(16)")
+  @Column(name = "id")
```
tdas, line 33: what will be the impact of this change on existing database + deployment? how do we know that this is backward compatible?

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/MySQLDeltaCommitsCRUDTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+/**
+ * Runs all Delta Commits CRUD tests against a real MySQL database via Testcontainers. Automatically
+ * skipped when Docker is not available. MySQL is the dialect that constrains the batch DELETE in
+ * DeltaCommitRepository: it rejects LIMIT directly inside an IN(...) subquery (error 1235) and a
+ * subquery selecting from the table being deleted from (error 1093), so this suite is what guards
+ * the derived-table shape of that query against regressions that H2 and PostgreSQL would not catch.
+ */
+@Testcontainers(disabledWithoutDocker = true)
```
tdas, line 15: [AI-assisted]

these skip without docker and you note the repo runs no test CI on PRs. can we make sure to first fix the repo to run dockerized tests and make sure that tests marked with this tag actually run? only then it makes sense to merge this PR with the confidence that things are working. 

alternative: why do we need to run these in docker? cant we always run these tests with different databases injected?

tdas (changes requested):
`server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java:386`
[AI-assisted]

stop() closes no session factory at all — not the injected one, not the self-built one. so "the caller keeps ownership: stopping the server does not close the injected configurator's session factory" draws a contrast that doesn't exist. someone reads this and assumes the non-injected factory gets closed on stop. either drop the framing or actually close the one we build.

a maintainer trusting this comment either 'fixes' a nonexistent leak by closing the injected factory (breaking the shared-ownership tests) or assumes cleanup on stop that never happens.

server/src/main/java/io/unitycatalog/server/persist/dao/DeltaCommitDAO.java
yili-db: [AI Generated]

Net effect of upgrading an existing server (removing `columnDefinition = "BINARY(16)"`):

Production uses `hbm2ddl.auto=update`, which is additive-only — it never drops or retypes an existing column. So on any existing DB the physical `uc_delta_commits.id` / `table_id` columns keep their current type; the new mapping only affects DDL for a *fresh* schema.

- **MySQL:** no change. Hibernate 6.x maps `java.util.UUID` → `binary(16)` regardless of `columnDefinition`, so old and new DDL are identical.
- **H2:** existing columns stay `BINARY(16)` (not migrated). The new code reads/writes `java.util.UUID` through them fine — H2 stores UUID as a 16-byte value and converts implicitly between `UUID` and `BINARY(16)`, and these columns are only used with equality predicates (`WHERE table_id = ?`, `id IN (...)`), no ordering/range. Only brand-new H2 databases get `UUID` as the declared type.
- **PostgreSQL:** this is a fix, not a compat risk. `BINARY(16)` is invalid in PostgreSQL (`type "binary" does not exist`), so the server could never start — there is no prior working table to preserve. The new code generates native `uuid`.

Summary: MySQL unchanged, H2 unchanged and still works, PostgreSQL newly fixed.

One gap worth closing: no committed test exercises the H2 upgrade path — all tests run with `create-drop` (fresh schema). Worth adding a test that seeds a table with the legacy `BINARY(16)` column and runs the new mapping against it, so the H2 upgrade case is enforced in CI rather than argued.
However in this repo we *never* had any prod deployment upgrade test. This will eventually be needed to make it truly production ready.
I don't think this should be done in just this PR. But it's a much larger effort on our roadmap.

tdas: I agree that this is a larger effort. non-trivial to begin with. let's not block this PR on that.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -410,6 +413,18 @@ public UnityCatalogServer.Builder serverProperties(ServerProperties serverProper
...
+     */
```
yili-db, line 421: [AI Generated]

Good catch — in this PR `stop()` closes no session factory at all (neither injected nor self-built), so the "stopping the server does not close the injected configurator's session factory" contrast is misleading. Suggesting wording that drops the false contrast and just states the true ownership fact:

```suggestion
    /**
     * Uses the given {@link HibernateConfigurator} instead of creating one from the server
     * properties. Lets tests share the server's session factory and customize the hibernate
     * properties (e.g. run against PostgreSQL via Testcontainers). The server never closes this
     * factory, so the caller owns its lifecycle.
     */
```

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/MySQLDeltaCommitsCRUDTest.java
yili-db: We have agreed that having these tests merged as-is is good enough for this PR.
I will investigate how to enable docker in this repo as a followup task.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
yili-db: This is to address comment  https://github.com/unitycatalog/unitycatalog/pull/1446#pullrequestreview-4687409493 to make the comment accurate.
The actual fix belongs to https://github.com/unitycatalog/unitycatalog/pull/1660

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/MySQLDeltaCommitsCRUDTest.java
tiennguyen-onehouse: Thank you so much @yili-db for addressing the comments! Really sorry on being non-responsive. I was overwhelmed with work. I'm happy to open another PR for the prod deployment upgrade test. I have actually added the docker test run below and tested them in my fork. I added a tag ("database") for them. If you still want to move them into a separate PR, I would happy to do so.

yili-db: Yes please remove the new github workflow for now.
We have a bit of supply chain hardening to do before we add any more github workflows. Thank you!

tiennguyen-onehouse: Done in `8f5b7e47` — removed the `Database Integration Tests` workflow. Makes sense to hold off until the supply-chain hardening is in place.

I also dropped the `@Tag("database")` markers, since their only purpose was the workflow's tag selector; with the workflow gone they'd just be orphaned. When you set up docker CI you can pick whatever tagging scheme fits. The Postgres and MySQL suites themselves are untouched — still `@Testcontainers(disabledWithoutDocker = true)`, so they run where Docker is available and skip otherwise.

The PR diff is now back to the DB fix + tests. Thanks for driving this while I was heads-down.

tdas (approved):
thanks for addressing the comments. regarding the stop issue, i see that there is follow up work planned. so i am okay with skipping that in this PR. 

thank you @tiennguyen-onehouse for your contribution and @yili-db for reviewing. LGTM.

PR #1456, opened 2026-03-31
Remove duplicate credential properties without option. prefix

### LINKED ISSUE #1356: Remove the duplicate UC properties setting for both `key` and `option.key`.
In this code : https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala#L294-L299

We set the same properities twice, one for `key` and another for `option.key`.  which cause the problem that we discussed in here: https://github.com/delta-io/delta/pull/6046/changes#r2796674010

We should remove the `fs.*`  keys and only keep the `option.fs.*`, just keep an issue here.

niruta25:
**PR Checklist**

**Description of changes**

## Summary
- Remove bare `fs.*` credential properties from `UCSingleCatalog.setCredentialProps`, keeping only `option.`-prefixed keys
- Delta previously required both forms, but delta-io/delta#6046 has been merged and Delta now only reads `option.`-prefixed keys
- This resolves the duplicate properties issue discussed in https://github.com/delta-io/delta/pull/6046/changes#r2796674010

## What changed
`setCredentialProps` in `UCSingleCatalog.scala` was setting every credential property twice: once as bare `fs.*` keys and once with the `option.` prefix. The bare keys are now removed, along with the TODO comment that flagged this for cleanup.

## User impact
- **No user-facing behavior change.** Credential properties are set internally by the Spark connector — users do not configure these directly.
- With the Delta-side fix (delta-io/delta#6046) already merged, only the `option.`-prefixed keys are read. The bare `fs.*` keys were redundant and their removal eliminates a source of confusion when debugging table properties.
- Users on **Delta 4.x+ with UC Spark connector** will see cleaner table properties with no duplicate entries.

## Test plan
- [x] Existing `CredPropsUtilTest` (7 tests) — all pass
- [x] Spark connector unit tests (115 passed, 27 skipped due to missing cloud credentials — pre-existing, unrelated to this change)
- [x] `build/sbt javafmtAll` — no formatting issues

Fixes #1356


niruta25:
@tdas @dennyglee  @openinx . Any thoughts on this? Quick EZ PR.

yili-db:
This seems to be implemented by #1789: at least they aren't sent to server any more. 
@openinx Should this `option.*` properties be removed at all? Not even return to Delta?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -473,9 +473,6 @@ object UCSingleCatalog {
 
   def setCredentialProps(props: util.HashMap[String, String],
                          credentialProps: util.Map[String, String]): Unit = {
-    props.putAll(credentialProps)
```
timothyw553, line 476: I don't think we should remove the `fs.*` properties yet. Delta 4.1.0 uses `option.fs.*` to identify option names, but reads each value from the matching `fs.*` entry.

With this change, Delta produces an empty `writeOptions` map and cloud operations can lose their credentials. [Delta #6046](https://github.com/delta-io/delta/pull/6046) forwards extracted options but does not change this parsing. Please keep both forms. If the goal is cleaner stored properties, filter these credentials before sending table metadata to UC instead.

timothyw553 (changes requested):
i dont recommend this yet -- this could regressions in dependents.

PR #1479, opened 2026-04-07
[Spark] Enable credScopedFs by default in Spark connector

openinx:
**PR Checklist**

**Description of changes**

Enable `credScopedFs.enabled` by default in the Spark connector (`DEFAULT_CRED_SCOPED_FS_ENABLED = true`). Previously users had to explicitly set `credScopedFs.enabled=true` in their catalog options to opt in. Since the credential-scoped filesystem provides better credential isolation and security, it should be the default behavior going forward. Users who need the old behavior can still opt out by setting `credScopedFs.enabled=false`.

Also updates `BaseSparkIntegrationTest` helper overloads to default to `credScopedFsEnabled=true`, so non-parameterized tests exercise the new default. The parameterized tests already cover both `false` and `true` combinations explicitly.


connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
```diff
@@ -46,11 +46,7 @@ private void createCommonResources() throws ApiException {
   }
 
   protected SparkSession createSparkSessionWithCatalogs(String... catalogs) {
-    return createSparkSessionWithCatalogs(false, false, catalogs);
-  }
-
-  protected SparkSession createSparkSessionWithCatalogs(boolean renewCred, String... catalogs) {
```
openinx, line 52: This method actually is unused any more, so we should be safe to remove it.

tdas (approved):

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -478,15 +478,19 @@ object UCSingleCatalog {
...
+   * any existing {@code fs.<scheme>.impl} values before overriding them with the credential-scoped
+   * filesystem wrapper. Using {@code sessionState.newHadoopConf()} (instead of
+   * {@code sparkContext.hadoopConfiguration}) ensures the lookup also sees Hadoop properties set
+   * directly via {@code SparkSession.Builder.config("fs.<scheme>.impl", ...)} or {@code SET}
+   * commands, not only those prefixed with {@code spark.hadoop.}.
    */
   def sessionHadoopConf(): Configuration = {
     SparkSession.getActiveSession
-      .map(_.sparkContext.hadoopConfiguration)
+      .map(_.sessionState.newHadoopConf())
```
openinx, line 493: `UCSingleCatalog.sessionHadoopConf()` currently returns `sparkContext.hadoopConfiguration`, which **only contains `spark.hadoop.*`-prefixed entries**. Switching it to `sessionState.newHadoopConf()` makes it also include Hadoop settings set via `spark.conf.set(...)`, `SparkSession.Builder.config(...)`, or SQL `SET` — which is what Spark SQL and Delta themselves use (per [SPARK-23514](https://issues.apache.org/jira/browse/SPARK-23514)).

This matters now because the PR enables `CredScopedFileSystem` by default. The wrapper reads the user's existing `fs.<scheme>.impl` from this conf to remember the "original" delegate. With the old API, any non-prefixed override was invisible, so the wrapper silently delegated to the wrong (real cloud) FS — failing the credential-renewal tests and creating the same footgun for real users.

PR #1482, opened 2026-04-09
Add uv.lock and exclude-newer to all ai/ modules

TomeHirata:
## Summary

- Adds `[tool.uv]` with `exclude-newer = "2026-03-10T00:00:00Z"` to all 10 `pyproject.toml` files under `ai/`
- Generates `uv.lock` for each module (core + 9 integrations)
- For packages with `requires-python = ">=3.9"` that have databricks optional deps requiring Python>=3.10, adds `environments = ["python_version >= '3.10'"]` to restrict lock resolution scope (core, anthropic, gemini, litellm, openai, autogen)


ai/core/pyproject.toml
```diff
@@ -127,3 +127,7 @@ convention = "google"
 
 [tool.pytest.ini_options]
 pythonpath = ["src"]
+
+[tool.uv]
+exclude-newer = "2026-03-10T00:00:00Z"
```
openinx, line 132: Hi @TomeHirata ,  after this uv.lock work,  next plan I think it's to enable the workflow ? right ?

ai/core/uv.lock
```diff
@@ -2,17 +2,28 @@ version = 1
 revision = 3
 requires-python = ">=3.9"
```
seewishnew, line 3: Nit: the pyproject seems to require python >= 3.10 -- should this uv.lock also have 3.10 as the sane lower limit?

ai/integrations/anthropic/uv.lock
```diff
@@ -0,0 +1,2919 @@
+version = 1
+revision = 3
+requires-python = ">=3.9"
```
seewishnew, line 3: Same nit about potential python version mismatch as in the core/uv.lock above

ai/integrations/llama_index/pyproject.toml
```diff
@@ -70,4 +70,7 @@ docstring-code-format = true
 docstring-code-line-length = 88
 
 [tool.ruff.lint.pydocstyle]
-convention = "google"
\ No newline at end of file
+convention = "google"
+
+[tool.uv]
+exclude-newer = "2026-03-10T00:00:00Z"
```
seewishnew, line 76: Some of these pyprojects don't seem to provide a lowerbound for the python version, should we take an opinionated stance here and make it 3.10 minimum for all these projects?

ai/core/pyproject.toml
seewishnew: * As long as the workflows are updated to install frozen dependencies from the lock files.

TomeHirata:
This change is made in https://github.com/unitycatalog/unitycatalog/pull/1467

PR #1488, opened 2026-04-14
Implement Delta REST Catalog loadTable endpoint

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1488/files) to review incremental changes.
- [**stack/drc_loadtable**](https://github.com/unitycatalog/unitycatalog/pull/1488) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1488/files)]
  - [stack/drc_get_param](https://github.com/unitycatalog/unitycatalog/pull/1499) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1499/files/8b968878117e49caf205eae1584bbb964908677d..fe561fea278407099e68aa1778d05e2bb25c518c)]
    - [stack/drc_cred](https://github.com/unitycatalog/unitycatalog/pull/1500) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1500/files/fe561fea278407099e68aa1778d05e2bb25c518c..82449580bc53760649fe486a2b48437b577f2567)]

---------
**PR Checklist**

**Description of changes**
Implement Delta REST Catalog loadTable endpoint

Adds GET /delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table} returning LoadTableResponse with metadata, typed columns, properties, partition columns, commits, and uniform metadata in a single REPEATABLE_READ transaction.

- TableRepository.loadTableForDelta assembles the full response from DAO. Columns parsed from typeJson via ColumnUtils with Jackson mixins for Spark camelCase support. Commits via refactored DeltaCommitRepository.getUnbackfilledCommits (reused by existing getCommits). last-commit-version from table properties, not commit list.
- Fix ApiClientBuilder DeltaTypeModule registration (getObjectMapper returns a copy). TestUtils.createApiClient now uses ApiClientBuilder.
- Tests: SdkLoadTableTest (external + managed commit/backfill flow), ColumnUtilsTest, access control in SdkTableAccessControlCRUDTest.

### DIFF
--- a/api/delta-docs/Models/DeltaCommit.md
+++ b/api/delta-docs/Models/DeltaCommit.md
@@ -4,10 +4,10 @@
 | Name | Type | Description | Notes |
 |------------ | ------------- | ------------- | -------------|
 | **version** | **Long** | Commit version | [default to null] |
-| **timestamp** | **Long** | In-commit timestamp | [default to null] |
+| **timestamp** | **Long** | In-commit timestamp, in epoch milliseconds | [default to null] |
 | **file-name** | **String** | UUID-based commit file name | [default to null] |
 | **file-size** | **Long** | Commit file size in bytes | [default to null] |
-| **file-modification-timestamp** | **Long** | File modification timestamp | [default to null] |
+| **file-modification-timestamp** | **Long** | File modification timestamp, in epoch milliseconds | [default to null] |
 
 [[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
 
--- a/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
@@ -101,32 +101,57 @@ public class DeltaCommitRepository {
   }
 
   /**
-   * Retrieves all commits for a table in descending order by version up to NUM_COMMITS_PER_BATCH.
+   * Result of querying unbackfilled commits for a table.
+   *
+   * @param commits unbackfilled commits (descending version order, newest first)
+   * @param latestTableVersion the latest commit version (0 if no commits)
+   * @param oldestVersion the oldest commit version in the DB (used for pagination base)
    */
-  private List<DeltaCommitDAO> getAllCommitDAOsDesc(UUID tableId) {
-    return TransactionManager.executeWithTransaction(
-        sessionFactory,
-        session -> {
-          TableInfoDAO tableInfoDAO = session.get(TableInfoDAO.class, tableId);
-          if (tableInfoDAO == null) {
-            throw new BaseException(ErrorCode.TABLE_NOT_FOUND, "Table not found: " + tableId);
-          }
-          validateTable(tableInfoDAO);
+  record CommitQueryResult(
+      List<DeltaCommitDAO> commits, long latestTableVersion, long oldestVersion) {}
 
-          Query<DeltaCommitDAO> query =
-              session.createQuery(
-                  "FROM DeltaCommitDAO WHERE tableId = :tableId ORDER BY commitVersion DESC",
-                  DeltaCommitDAO.class);
-          query.setParameter("tableId", tableId);
-          query.setMaxResults(NUM_COMMITS_PER_BATCH);
-          return query.list();
-        },
-        "Failed to get latest commits",
-        /* readOnly= */ true,
-        // Use REPEATABLE_READ isolation to ensure consistent snapshot and prevent version numbers
-        // from appearing to go backwards during concurrent writes. This is critical for get
-        // commits that expect monotonic version progression after posting a commit.
-        Optional.of(java.sql.Connection.TRANSACTION_REPEATABLE_READ));
+  /**
+   * Query unbackfilled commits for a table within an existing session. Returns commits in
+   * descending version order (newest first) and the latest table version.
+   *
+   * <p>Handles empty tables (returns version 0) and fully backfilled tables (returns empty list
+   * with correct version).
+   */
+  CommitQueryResult getUnbackfilledCommits(Session session, UUID tableId) {
+    Query<DeltaCommitDAO> query =
+        session.createQuery(
+            "FROM DeltaCommitDAO WHERE tableId = :tableId ORDER BY commitVersion DESC",
+            DeltaCommitDAO.class);
+    query.setParameter("tableId", tableId);
+    query.setMaxResults(NUM_COMMITS_PER_BATCH);
+    List<DeltaCommitDAO> allDesc = query.list();
+
+    if (allDesc.isEmpty()) {
+      return new CommitQueryResult(List.of(), 0L, 0L);
+    }
+
+    int commitCount = allDesc.size();
+    if (commitCount > MAX_NUM_COMMITS_PER_TABLE) {
+      LOGGER.error(
+          "Table {} has {} commits, exceeds limit {}.",
+          tableId,
+          commitCount,
+          MAX_NUM_COMMITS_PER_TABLE);
+    }
+
+    DeltaCommitDAO newestCommit = allDesc.get(0);
+    long latestVersion = newestCommit.getCommitVersion();
+    long oldestVersion = allDesc.get(allDesc.size() - 1).getCommitVersion();
+    if (newestCommit.isBackfilledLatestCommit()) {
+      return new CommitQueryResult(List.of(), latestVersion, oldestVersion);
+    }
+
+    // Only the latest backfilled version is kept in the DB as a
+    // marker (isBackfilledLatestCommit=true). All other backfilled
+    // commits are deleted. Filter out this single marker.
+    List<DeltaCommitDAO> unbackfilled =
+        allDesc.stream().filter(c -> !c.isBackfilledLatestCommit()).toList();
+    return new CommitQueryResult(unbackfilled, latestVersion, oldestVersion);
   }
 
   /**
@@ -163,56 +188,38 @@ public class DeltaCommitRepository {
         endVersion.filter(x -> x < startVersion).isEmpty(),
         "end_version must be >=start_version if set");
 
-    List<DeltaCommitDAO> allCommitDAOsDesc = getAllCommitDAOsDesc(tableId);
-    int commitCount = allCommitDAOsDesc.size();
-    if (commitCount > MAX_NUM_COMMITS_PER_TABLE) {
-      // This should never occur. But this is recoverable and not fatal.
-      LOGGER.error(
-          "Table {} has {} commits, which exceeds the limit of {}.",
-          tableId,
-          commitCount,
-          MAX_NUM_COMMITS_PER_TABLE);
-    }
-    if (commitCount == 0) {
-      // Table is validated as a managed Delta table. UC is the source of truth for
-      // managed tables, so a newly created table with no commits is at version 0.
-      return new DeltaGetCommitsResponse().latestTableVersion(0L);
-    }
+    return TransactionManager.executeWithTransaction(
+        sessionFactory,
+        session -> {
+          // Validate table exists and is managed Delta
+          TableInfoDAO tableInfoDAO = session.get(TableInfoDAO.class, tableId);
+          if (tableInfoDAO == null) {
+            throw new BaseException(ErrorCode.TABLE_NOT_FOUND, "Table not found: " + tableId);
+          }
+          validateTable(tableInfoDAO);
 
-    // In case there's only one commit in database, firstCommitDAO and lastCommitDAO will be the
-    // same object.
-    DeltaCommitDAO firstCommitDAO = allCommitDAOsDesc.get(allCommitDAOsDesc.size() - 1);
-    DeltaCommitDAO lastCommitDAO = allCommitDAOsDesc.get(0);
-    assert firstCommitDAO.getCommitVersion() <= lastCommitDAO.getCommitVersion();
-    if (lastCommitDAO.isBackfilledLatestCommit()) {
-      // The last commit is already backfilled. Just return an empty list. No need to return any
-      // actual commits.
-      return new DeltaGetCommitsResponse().latestTableVersion(lastCommitDAO.getCommitVersion());
-    }
+          CommitQueryResult result = getUnbackfilledCommits(session, tableId);
 
-    // The last version to return if endVersion is not set. It's limited by pagination limit.
-    // In normal cases the pagination limitation should not happen at all since the limit is the
-    // same limit that a table can have as many unbackfilled commits as possible. But it is
-    // implemented this way just in case.
-    long paginatedEndVersionInclusive =
-        Math.max(startVersion, firstCommitDAO.getCommitVersion()) + MAX_NUM_COMMITS_PER_TABLE - 1;
-    // The actual last version to return
-    long effectiveEndVersionInclusive =
-        Math.min(endVersion.orElse(Long.MAX_VALUE), paginatedEndVersionInclusive);
+          // Apply version range filter + pagination
+          long paginatedEnd =
+              Math.max(startVersion, result.oldestVersion()) + MAX_NUM_COMMITS_PER_TABLE - 1;
+          long effectiveEnd = Math.min(endVersion.orElse(Long.MAX_VALUE), paginatedEnd);
 
-    // Filter result and return
-    List<DeltaCommitInfo> commits =
-        allCommitDAOsDesc.stream()
-            .filter(
-                c ->
-                    !c.isBackfilledLatestCommit()
-                        && c.getCommitVersion() >= startVersion
-                        && c.getCommitVersion() <= effectiveEndVersionInclusive)
-            .map(DeltaCommitDAO::toCommitInfo)
-            .collect(Collectors.toList());
-    return new DeltaGetCommitsResponse()
-        .commits(commits)
-        .latestTableVersion(lastCommitDAO.getCommitVersion());
+          List<DeltaCommitInfo> commits =
+              result.commits().stream()
+                  .filter(
+                      c ->
+                          c.getCommitVersion() >= startVersion
+                              && c.getCommitVersion() <= effectiveEnd)
+                  .map(DeltaCommitDAO::toCommitInfo)
+                  .collect(Collectors.toList());
+          return new DeltaGetCommitsResponse()
+              .commits(commits)
+              .latestTableVersion(result.latestTableVersion());
+        },
+        "Failed to get commits",
+        /* readOnly= */ true,
+        Optional.of(java.sql.Connection.TRANSACTION_REPEATABLE_READ));
   }
 
   /**
--- a/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
@@ -1,5 +1,7 @@
 package io.unitycatalog.server.persist;
 
+import static java.sql.Connection.TRANSACTION_REPEATABLE_READ;
+
 import io.unitycatalog.server.exception.BaseException;
 import io.unitycatalog.server.exception.ErrorCode;
 import io.unitycatalog.server.model.ColumnInfos;
@@ -219,7 +221,7 @@ public class DeltaCommitRepository {
         },
         "Failed to get commits",
         /* readOnly= */ true,
-        Optional.of(java.sql.Connection.TRANSACTION_REPEATABLE_READ));
+        Optional.of(TRANSACTION_REPEATABLE_READ));
   }
 
   /**
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -1,5 +1,7 @@
 package io.unitycatalog.server.persist;
 
+import static java.sql.Connection.TRANSACTION_REPEATABLE_READ;
+
 import io.unitycatalog.server.delta.model.DeltaCommit;
 import io.unitycatalog.server.delta.model.LoadTableResponse;
 import io.unitycatalog.server.delta.model.StructType;
@@ -214,7 +216,7 @@ public class TableRepository {
         },
         "Failed to load table",
         /* readOnly = */ true,
-        Optional.of(java.sql.Connection.TRANSACTION_REPEATABLE_READ));
+        Optional.of(TRANSACTION_REPEATABLE_READ));
   }
 
   private TableMetadata buildTableMetadata(
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -221,9 +221,11 @@ public class TableRepository {
       Session session, TableInfoDAO dao, String catalog, String schema, String table) {
     TableMetadata metadata = new TableMetadata();
     Long updatedAt = dao.getUpdatedAt() != null ? dao.getUpdatedAt().getTime() : null;
-    // When updatedAt is null (shouldn't happen for a persisted table but defensively handled),
-    // the etag becomes "etag-null" which is a stable distinct value for that edge case.
-    metadata.setEtag("etag-" + updatedAt);
+    // Normal case: etag keyed on the table's last update time. If updatedAt is missing
+    // (shouldn't happen for a persisted table, but defensively handled), fall back to the
+    // table UUID so the etag stays unique per table rather than collapsing to "etag-null"
+    // across rows.
+    metadata.setEtag(updatedAt != null ? "etag-" + updatedAt : "etag-" + dao.getId());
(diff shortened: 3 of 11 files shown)


server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -161,6 +171,189 @@ public TableInfo getTable(String fullName) {
...
+            "Table {}.{}.{} has invalid partition indices, expected {} but got {}",
+            catalog,
+            schema,
+            table,
+            i,
+            partitionInfos.get(i).getPartitionIndex());
+        break;
+      }
+    }
+    metadata.setPartitionColumns(partitionInfos.stream().map(ColumnInfo::getName).toList());
```
tdas, line 294: [AI Assisted] If `partitionIndex` values are not a contiguous `0..n-1` sequence (e.g. missing an index), this loop logs and `break`s but execution still sets `partitionColumns` from the full sorted `partitionInfos` list. Clients can then see partition column names that do not correspond to a valid partition spec. Should we omit or clear `partitionColumns` when validation fails, or document that this path is intentionally best-effort with a possibly inconsistent list?

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -161,6 +171,189 @@ public TableInfo getTable(String fullName) {
...
+        Optional.of(java.sql.Connection.TRANSACTION_REPEATABLE_READ));
+  }
+
+  private TableMetadata buildTableMetadata(
+      Session session, TableInfoDAO dao, String catalog, String schema, String table) {
+    TableMetadata metadata = new TableMetadata();
+    Long updatedAt = dao.getUpdatedAt() != null ? dao.getUpdatedAt().getTime() : null;
+    // When updatedAt is null (shouldn't happen for a persisted table but defensively handled),
+    // the etag becomes "etag-null" which is a stable distinct value for that edge case.
+    metadata.setEtag("etag-" + updatedAt);
```
tdas, line 226: [AI Assisted] When `updatedAt` is null, the etag becomes the literal string `etag-null`. That is stable but not unique across tables if more than one row ever lacked `updatedAt`. The comment says it is a "stable distinct value" for that edge case, which is misleading unless we know only one such table can exist. Consider incorporating something unique (e.g. table UUID) or tightening the invariant so `updatedAt` is never null for persisted tables.

tdas:
[AI Assisted] **Test gap:** The new `populatePartitionColumns` behavior (non-contiguous `partitionIndex` → empty `partitionColumns` + warn) is not covered by an integration test. Consider adding a `SdkLoadTableTest` case that creates a table with invalid partition indices (e.g. `0` and `2` without `1`, or duplicate indices) and asserts `metadata.getPartitionColumns()` is empty. Optional: if you can surface `updatedAt == null` in a test row, assert etag falls back to `"etag-" + tableId` without colliding across tables.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -101,32 +101,57 @@ public DeltaCommitRepository(SessionFactory sessionFactory, ServerProperties ser
...
-  private List<DeltaCommitDAO> getAllCommitDAOsDesc(UUID tableId) {
-    return TransactionManager.executeWithTransaction(
-        sessionFactory,
-        session -> {
-          TableInfoDAO tableInfoDAO = session.get(TableInfoDAO.class, tableId);
-          if (tableInfoDAO == null) {
-            throw new BaseException(ErrorCode.TABLE_NOT_FOUND, "Table not found: " + tableId);
-          }
-          validateTable(tableInfoDAO);
+  record CommitQueryResult(
```
openinx, line 112: will we be allowed to use the new keyword `record` ?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -101,32 +101,57 @@ public DeltaCommitRepository(SessionFactory sessionFactory, ServerProperties ser
...
+    query.setMaxResults(NUM_COMMITS_PER_BATCH);
+    List<DeltaCommitDAO> allDesc = query.list();
+
+    if (allDesc.isEmpty()) {
+      return new CommitQueryResult(List.of(), 0L, 0L);
+    }
+
+    int commitCount = allDesc.size();
+    if (commitCount > MAX_NUM_COMMITS_PER_TABLE) {
+      LOGGER.error(
```
openinx, line 137: In this case,  should we throw the exception or just logging the error message is good enough ? seems like  we'd better to throw the exception?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -101,32 +101,57 @@ public DeltaCommitRepository(SessionFactory sessionFactory, ServerProperties ser
...
+          "Table {} has {} commits, exceeds limit {}.",
+          tableId,
+          commitCount,
+          MAX_NUM_COMMITS_PER_TABLE);
+    }
+
+    DeltaCommitDAO newestCommit = allDesc.get(0);
+    long latestVersion = newestCommit.getCommitVersion();
+    long oldestVersion = allDesc.get(allDesc.size() - 1).getCommitVersion();
```
openinx, line 146: qq:  has the `allDesc` array already been sorted from latest to oldest  ?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
openinx, line 179: Seems like there is no usage for the `getCommits`  ?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -163,56 +188,38 @@ public DeltaGetCommitsResponse getCommits(DeltaGetCommits rpc) {
...
-    // In normal cases the pagination limitation should not happen at all since the limit is the
-    // same limit that a table can have as many unbackfilled commits as possible. But it is
-    // implemented this way just in case.
-    long paginatedEndVersionInclusive =
-        Math.max(startVersion, firstCommitDAO.getCommitVersion()) + MAX_NUM_COMMITS_PER_TABLE - 1;
-    // The actual last version to return
-    long effectiveEndVersionInclusive =
-        Math.min(endVersion.orElse(Long.MAX_VALUE), paginatedEndVersionInclusive);
+          // Apply version range filter + pagination
+          long paginatedEnd =
```
openinx, line 206: Seems like every table only has max 10 version, then do we still need to support the pagination based on version ? because the version list should always be very small.  correct me if i understood incorrectly.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -161,6 +171,195 @@ public TableInfo getTable(String fullName) {
...
+                        .timestamp(c.getCommitTimestamp().getTime())
+                        .fileName(c.getCommitFilename())
+                        .fileSize(c.getCommitFilesize())
+                        .fileModificationTimestamp(
+                            c.getCommitFileModificationTimestamp().getTime()))
+            .toList();
+    response.setCommits(commits);
+  }
+
+  private static void populateUniformMetadata(LoadTableResponse response, TableInfoDAO dao) {
```
openinx, line 340: @ChengJiX ,  Could you pls also double check the uniform protocol, for this new catalog api ?  I think your work will depend on this .

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -161,6 +171,195 @@ public TableInfo getTable(String fullName) {
...
+        commitRepo.getUnbackfilledCommits(session, tableId);
+    response.setLatestTableVersion(result.latestTableVersion());
+
+    List<DeltaCommit> commits =
+        result.commits().stream()
+            .map(
+                c ->
+                    new DeltaCommit()
+                        .version(c.getCommitVersion())
+                        .timestamp(c.getCommitTimestamp().getTime())
```
openinx, line 331: nit: what's is the time unit for the timestamp ?

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -161,6 +171,195 @@ public TableInfo getTable(String fullName) {
...
+      return;
+    }
+    UniformMetadataIceberg iceberg = new UniformMetadataIceberg().metadataLocation(uniformLocation);
+    if (dao.getUniformIcebergConvertedDeltaVersion() != null) {
+      iceberg.convertedDeltaVersion(dao.getUniformIcebergConvertedDeltaVersion());
+    }
+    if (dao.getUniformIcebergConvertedDeltaTimestamp() != null) {
+      iceberg.convertedDeltaTimestamp(dao.getUniformIcebergConvertedDeltaTimestamp().getTime());
+    }
+    response.setUniform(new UniformMetadata().iceberg(iceberg));
```
openinx, line 352: Do we also need to fill the `UniformMetadataIceberg`'s `convertedDeltaTimestamp` and `baseConvertedDeltaVersion` ?

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
```diff
@@ -78,4 +84,25 @@ public CatalogConfig getConfig(
     // For now, we only have 1.0 as the first protocol version. Input protocolVersions is ignored.
     return new CatalogConfig().endpoints(ENDPOINTS).protocolVersion("1.0");
   }
+
+  // ==================== Load Table API ====================
+
+  @Get("/delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}")
```
openinx, line 90: what's is the unitycatalog API definition that I can compare to ?  ( I remember we have a page to list all the api definition right )

server/src/test/java/io/unitycatalog/server/sdk/access/SdkTableAccessControlCRUDTest.java
```diff
@@ -50,6 +50,14 @@ public void testTableAccess() {
     TablesApi regular2TablesApi = new TablesApi(TestUtils.createApiClient(regular2Config));
     SchemasApi regular2SchemasApi = new SchemasApi(TestUtils.createApiClient(regular2Config));
 
+    // Delta REST API clients for loadTable tests
+    io.unitycatalog.client.delta.api.TablesApi adminDeltaApi =
```
openinx, line 54: any reason we use the full qualified class name ?

server/src/test/java/io/unitycatalog/server/utils/TestUtils.java
```diff
@@ -63,17 +66,13 @@ public class TestUtils {
...
-    apiClient.setHost(uri.getHost());
-    apiClient.setPort(port);
-    apiClient.setScheme(uri.getScheme());
-    if (serverConfig.getAuthToken() != null && !serverConfig.getAuthToken().isEmpty()) {
-      apiClient.setRequestInterceptor(
-          request -> request.header("Authorization", "Bearer " + serverConfig.getAuthToken()));
-    }
-    return apiClient;
+    String token = serverConfig.getAuthToken() != null ? serverConfig.getAuthToken() : "";
+    return ApiClientBuilder.create()
```
openinx, line 71: Thanks for adopting the new api , looks good.

openinx (commented):
Thanks @yili-db for the work,  left several comments.

server/src/test/java/io/unitycatalog/server/sdk/access/SdkTableAccessControlCRUDTest.java
yili-db: Yes. `io.unitycatalog.client.api.TablesApi` (from `all.yaml`, no `delta`) already exists and is also tested in the same file.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
yili-db: It's put in `delta.yaml` file in prior PR.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: We shouldn't. However it's already part of the existing commit API contract.
This new API doesn't need pagination. This code is shared between both new and old API.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: It's still used by `DeltaCommitsService.java` for the old API.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: Yes. It has `ORDER BY commitVersion DESC` in the query above.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: When it exceeds commit limit (which should never happen):
1. committing a new commit is blocked.
2. getting commits is still allowed because that's the way client can backfill and recover. Otherwise this table is permanently broken.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: We already do. In the server side it's already Java 17.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: `convertedDeltaTimestamp` is filled above at line 348.
`baseConvertedDeltaVersion` is not tracked and should not be returned. It's meant for the client to tell server "which previous version I am doing the conversion based on".

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: Added in yaml file: `in epoch milliseconds`

api/delta-docs/Models/DeltaCommit.md
```diff
@@ -4,10 +4,10 @@
 | Name | Type | Description | Notes |
 |------------ | ------------- | ------------- | -------------|
 | **version** | **Long** | Commit version | [default to null] |
-| **timestamp** | **Long** | In-commit timestamp | [default to null] |
+| **timestamp** | **Long** | In-commit timestamp, in epoch milliseconds | [default to null] |
```
openinx, line 7: thanks for the update,  it's much clear !

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
openinx: yea, got it.  just saw the sql. resolved.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
openinx: oh, sorry.  My intellij references was broken !  yes, you are right.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -163,56 +188,38 @@ public DeltaGetCommitsResponse getCommits(DeltaGetCommits rpc) {
...
-    }
+    return TransactionManager.executeWithTransaction(
+        sessionFactory,
+        session -> {
+          // Validate table exists and is managed Delta
+          TableInfoDAO tableInfoDAO = session.get(TableInfoDAO.class, tableId);
+          if (tableInfoDAO == null) {
+            throw new BaseException(ErrorCode.TABLE_NOT_FOUND, "Table not found: " + tableId);
+          }
+          validateTable(tableInfoDAO);
```
openinx, line 201: Inside the `validateTable`,  do you think we should use the `equalsIgnoreCase` to see if the data source fromat is `DELTA` or `delta` .  seems like there are some cases where we actually have the lower case ?

yili-db:
> [AI Assisted] **Test gap:** The new `populatePartitionColumns` behavior (non-contiguous `partitionIndex` → empty `partitionColumns` + warn) is not covered by an integration test. Consider adding a `SdkLoadTableTest` case that creates a table with invalid partition indices (e.g. `0` and `2` without `1`, or duplicate indices) and asserts `metadata.getPartitionColumns()` is empty. Optional: if you can surface `updatedAt == null` in a test row, assert etag falls back to `"etag-" + tableId` without colliding across tables.

@tdas 
Added a test section `Non-contiguous partition indices: empty partitionColumns` in `SdkLoadTableTest`: creates a managed Delta table with `partitionIndex = 0` and `partitionIndex = 2` (index 1 missing) and asserts `metadata.getPartitionColumns()` is empty. Passes against the current `populatePartitionColumns` fix.

Skipped the `updatedAt == null` test — the write path always sets `updatedAt`, so surfacing that edge would require direct DAO mutation and the assertion would just mirror the conditional in `buildTableMetadata`. Happy to add it if you'd rather see it pinned.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -163,56 +188,38 @@ public DeltaGetCommitsResponse getCommits(DeltaGetCommits rpc) {
...
+                              && c.getCommitVersion() <= effectiveEnd)
+                  .map(DeltaCommitDAO::toCommitInfo)
+                  .collect(Collectors.toList());
+          return new DeltaGetCommitsResponse()
+              .commits(commits)
+              .latestTableVersion(result.latestTableVersion());
+        },
+        "Failed to get commits",
+        /* readOnly= */ true,
+        Optional.of(java.sql.Connection.TRANSACTION_REPEATABLE_READ));
```
openinx: nit:   if possible pls don't use the full qualified class name here.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: It's OK to have user inputs as lower case. But `tableInfoDAO.getDataSourceFormat()` is persisted in database and it's only initialized from string value of a enum like `.dataSourceFormat(tableInfo.getDataSourceFormat().toString())` so it's always upper case.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -161,6 +171,195 @@ public TableInfo getTable(String fullName) {
...
+          }
+
+          TableMetadata metadata = buildTableMetadata(session, dao, catalog, schema, table);
+
+          LoadTableResponse response = new LoadTableResponse();
+          response.setMetadata(metadata);
+
+          // Commits (managed Delta tables only)
+          if (TableType.MANAGED.toString().equals(dao.getType())
+              && DataSourceFormat.DELTA.toString().equals(dao.getDataSourceFormat())) {
```
openinx, line 208: Seems comment as the above https://github.com/unitycatalog/unitycatalog/pull/1488/changes#r3112622267

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -161,6 +171,195 @@ public TableInfo getTable(String fullName) {
...
+        metadata.setColumns(emptySchema);
+      }
+    } catch (Exception e) {
+      LOGGER.warn(
+          "Failed to parse columns for table {}.{}.{}, returning empty schema",
+          catalog,
+          schema,
+          table,
+          e);
+      metadata.setColumns(emptySchema);
```
openinx, line 255: Is this correct if we are failing to parse the schema, and silently set a `emptySchema` without breaking the path ?  I'm not sure,  just for confirming.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -161,6 +171,195 @@ public TableInfo getTable(String fullName) {
...
+        // partition list rather than a possibly-partial one the client can't reconcile.
+        LOGGER.warn(
+            "Table {}.{}.{} has invalid partition indices, expected {} but got {}; "
+                + "emitting empty partition columns",
+            catalog,
+            schema,
+            table,
+            i,
+            partitionInfos.get(i).getPartitionIndex());
+        metadata.setPartitionColumns(List.of());
```
openinx, line 296: same questions here.  will it be okay to silently ignore the error ?

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
openinx: looks good,thanks.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: Returning empty schema is still better than returning nothing which blocks access to the entire table.
Delta tables can still be accessed by reading table schema from commit logs. And a update call can repair the table.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: Updated in the other comment. It's better than blocking access entirely.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
openinx: `baseConvertedDeltaVersion` is used for incremental conversion I think, according to the API doc. 
I thin @ChengJiX   will continues to implement the incremental conversion for uniform table.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -161,6 +171,195 @@ public TableInfo getTable(String fullName) {
...
+   * failing the entire response.
+   */
+  public LoadTableResponse loadTableForDelta(String catalog, String schema, String table) {
+    return TransactionManager.executeWithTransaction(
+        sessionFactory,
+        session -> {
+          TableInfoDAO dao = findTable(session, catalog, schema, table);
+          if (dao == null) {
+            throw new BaseException(
+                ErrorCode.TABLE_NOT_FOUND,
```
openinx, line 195: QQ: will this error respect the protocol that we defined in the `delta.yaml` ? 

```
  /delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}:
    parameters:
      - $ref: '#/components/parameters/catalog'
      - $ref: '#/components/parameters/schema'
      - $ref: '#/components/parameters/table'
    get:
      tags:
        - Tables
      operationId: loadTable
      summary: Load table metadata
      description: |
        Load table metadata including columns, properties, and optionally credentials.
      responses:
        '200':
          description: Table loaded successfully
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/LoadTableResponse'
        '401':
          $ref: '#/components/responses/UnauthorizedResponse'
        '403':
          $ref: '#/components/responses/ForbiddenResponse'
        '404':
          $ref: '#/components/responses/NotFoundResponse'
```

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
openinx: okay,  got it, thanks for the clarification.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
openinx: okay...

openinx (approved):
Looks good to me overall.  for the `uniform` protocol,  @ChengJiX  pls take a look.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: This `BaseException` itself is just an exception. `DeltaRestExceptionHandler` converts it to the correct form in response.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
ChengJiX: Hi @openinx @yili-db, thanks for the work! Yeah there's no need to return `baseConvertedDeltaVersion` for `loadTable`. It is not stored inside db and its sent via updateTable request for validation purpose only

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
ChengJiX: Yeah the change looks good. It returns UniForm metadata with loadTable request

ChengJiX (approved):
Uniform change LGTM. Thanks for the work!

tdas (approved):

PR #1491, opened 2026-04-17
Add cross-channel authz probes to guard parameter-swap attacks

yili-db:
**PR Checklist**

**Description of changes**

Add cross-channel authz probes to guard parameter-swap attacks

`UnityAccessDecorator` picks a locator source (URL path, URL query, or request body) per `@Param`. Moving a parameter to the "wrong" channel must not skip authz; the latent danger is a PAYLOAD-source key on a body-less method, safe today only because Armeria blocks body-less GETs that need POJO binding.

These probes extend existing denial tests with raw `HttpClient` calls that retry the denied operation through the wrong channel, asserting no data leak and no unauthorized mutation, across read (GET) and write (PATCH/POST) paths and all three sources:

- `SdkCatalogAccessControlCRUDTest` (path): body-override GET still 403; no-path GET routes to `listCatalogs` and must not leak `admincatalog1`; `updateCatalog` PATCH on a non-owned catalog stays 403 and leaves it unmodified.
- `DeltaConfigAccessControlTest` (query): `getConfig` with `catalog` moved to the body is rejected without the config (404 today, before authz runs).
- `SdkDeltaCommitsAccessControlCRUDTest` (body): body-less `getCommits` with fields on the query is rejected without commit data (400); `postCommit` with `table_id` moved to the query never applies a commit (500 today -- fail-closed but ungraceful).

Shared `TestUtils.sendRaw(config, method, path, Optional<String> jsonBody)` handles the raw HTTP (`sendRawGet` delegates to it), consolidating the former POST-only helper. No production code changes, no new test files.

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/sdk/access/SdkCatalogAccessControlCRUDTest.java
+++ b/server/src/test/java/io/unitycatalog/server/sdk/access/SdkCatalogAccessControlCRUDTest.java
@@ -14,7 +14,9 @@ import io.unitycatalog.client.model.UpdateCatalog;
 import io.unitycatalog.server.base.ServerConfig;
 import io.unitycatalog.server.persist.model.Privileges;
 import io.unitycatalog.server.utils.TestUtils;
+import java.net.http.HttpResponse;
 import java.util.List;
+import java.util.Optional;
 import lombok.SneakyThrows;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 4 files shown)


yili-db:
To be clear: there is no security issue found in current code base. These tests are only here to guard against future changes.

openinx (approved):
The newly added test cases are hardening the behavior to prevent future breaking,  overall looks good to me.

PR #1494, opened 2026-04-17
Fail-closed guard for PAYLOAD-source authorization

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1494/files) to review incremental changes.
- [**auth-block-empty-body**](https://github.com/unitycatalog/unitycatalog/pull/1494) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1494/files)] ← _this PR_
  - stack/armeria_builder
  - stack/auth-hardening-wip

---------
**PR Checklist**

**Description of changes**
Fail-closed guard for PAYLOAD-source authorization

UnityAccessDecorator authorizes body-field (PAYLOAD) locators inside Armeria's `peekData` callback, which only fires for a request that delivers a complete JSON body. A body-less, non-JSON, or malformed request silently skips that authorization while the handler still runs. This is defense in depth: every current PAYLOAD endpoint also binds a required POJO body, so such requests already fail Jackson binding with a 400.

- Add a per-request `AUTH_PENDING` attribute, set on entry to the PAYLOAD path and cleared only after the authorization evaluation runs.
- Add `AuthorizationGateConverter`, which denies with `403 PERMISSION_DENIED` at body-binding time if `AUTH_PENDING` is still set.
- Wire the gate via `gatedJackson()` in `UnityCatalogServer`.

Follow-up: widening `peekData`'s payload-recognition heuristic (charset-qualified content types, trailing whitespace) is deferred so the gate cannot deny valid non-SDK requests; current SDK clients are unaffected.

Tested: `SdkCatalogAccessControlCRUDTest` asserts a body-less `POST /catalogs` returns 403 before the handler runs.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizationGateConverter.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizationGateConverter.java
@@ -1,0 +1,49 @@
+package io.unitycatalog.server.auth.decorator;
+
+import com.linecorp.armeria.common.AggregatedHttpRequest;
+import com.linecorp.armeria.server.ServiceRequestContext;
+import com.linecorp.armeria.server.annotation.RequestConverterFunction;
+import io.unitycatalog.server.exception.BaseException;
+import io.unitycatalog.server.exception.ErrorCode;
+import java.lang.reflect.ParameterizedType;
+import java.util.Objects;
+
+/**
+ * Fail-closed wrapper that enforces {@link UnityAccessDecorator}'s PAYLOAD-source authorization
+ * before delegating to an inner body converter (e.g. Jackson). The decorator sets {@link
+ * UnityAccessDecorator#AUTH_PENDING} on entry to the PAYLOAD path and clears it only after {@code
+ * checkAuthorization} succeeds; if the flag is still {@code true} when a parameter is being bound,
+ * {@code peekData}'s callback never fired (no body, non-JSON content-type, malformed JSON, trailing
+ * whitespace) and the request must be denied.
+ *
+ * <p>Wrapping Jackson (rather than sitting in front of it as a separate chain element) means
+ * services register a single converter and the gate is impossible to forget when a new annotated
+ * service is added.
+ *
+ * <p>Scope: gates only parameters bound via the converter chain. Endpoints with no body-bound
+ * parameter (pure {@code @Param}/path) don't trigger this converter, but those endpoints also don't
+ * use PAYLOAD locators, so the decorator never sets the flag and no gate is needed.
+ */
+public final class AuthorizationGateConverter implements RequestConverterFunction {
+
+  private final RequestConverterFunction delegate;
+
+  public AuthorizationGateConverter(RequestConverterFunction delegate) {
+    this.delegate = Objects.requireNonNull(delegate, "delegate");
+  }
+
+  @Override
+  public Object convertRequest(
+      ServiceRequestContext ctx,
+      AggregatedHttpRequest request,
+      Class<?> expectedResultType,
+      ParameterizedType expectedParameterizedResultType)
+      throws Exception {
+    if (Boolean.TRUE.equals(ctx.attr(UnityAccessDecorator.AUTH_PENDING))) {
+      throw new BaseException(
+          ErrorCode.PERMISSION_DENIED, "Authorization could not be verified for this request");
+    }
+    return delegate.convertRequest(
+        ctx, request, expectedResultType, expectedParameterizedResultType);
+  }
+}
--- a/server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizationGateConverter.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizationGateConverter.java
@@ -11,10 +11,10 @@ import java.util.Objects;
 /**
  * Fail-closed wrapper that enforces {@link UnityAccessDecorator}'s PAYLOAD-source authorization
  * before delegating to an inner body converter (e.g. Jackson). The decorator sets {@link
- * UnityAccessDecorator#AUTH_PENDING} on entry to the PAYLOAD path and clears it only after {@code
- * checkAuthorization} succeeds; if the flag is still {@code true} when a parameter is being bound,
- * {@code peekData}'s callback never fired (no body, non-JSON content-type, malformed JSON, trailing
- * whitespace) and the request must be denied.
+ * UnityAccessDecorator#AUTH_PENDING} on entry to the PAYLOAD path and clears it only after the
+ * authorization evaluation succeeds; if the flag is still {@code true} when a parameter is being
+ * bound, {@code peekData}'s callback never fired (no body, non-JSON content-type, malformed JSON,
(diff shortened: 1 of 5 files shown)


server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -170,13 +188,19 @@ private HttpResponse authorizeByRequest(
           resourceKeys,
           nonResourceValues);
 
-      // Note that peekData only gets called for requests that actually have data (like PUT&POST)
+      // peekData's lambda only fires when a body chunk arrives AND processPeekData returns true
+      // (JSON content-type + chunk ending in '}'). Any other path (no body, non-JSON, malformed
+      // JSON, trailing whitespace) silently skips authorization. Mark the request auth-pending
+      // here and clear it only after checkAuthorization runs; AuthorizationGateConverter denies
+      // any body-binding attempt where this flag is still set.
+      ctx.setAttr(AUTH_PENDING, true);
```
yili-db, line 196: This is the extra defense added, because `checkAuthorization` can be skipped by sending no a malformed request. It will be caught later by other code, but this extra check makes it even safer.

timothyw553 (approved):
LGTM. solution makes sense -- a skipped check turns into a request denied, right?

one thing that AI flagged was the "is this JSON?" check uses an exact match (`equals(MediaType.JSON)` see https://github.com/yili-db/unitycatalog/blob/78f138912cafdbd974869a092665dc1cac2ea80c/server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java#L442), so a content-type like `application/json; charset=utf-8` wouldn't match. in practice, if everyone goes through this SDK, which sends plain `application/json`, then it is fine.

but if they don't go through the sdk, instead, we should use Armeria's contentType.isJson()

yili-db:
> LGTM. solution makes sense -- a skipped check turns into a request denied, right?
> 
> one thing that AI flagged was the "is this JSON?" check uses an exact match (`equals(MediaType.JSON)` see https://github.com/yili-db/unitycatalog/blob/78f138912cafdbd974869a092665dc1cac2ea80c/server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java#L442), so a content-type like `application/json; charset=utf-8` wouldn't match. in practice, if everyone goes through this SDK, which sends plain `application/json`, then it is fine.
> 
> but if they don't go through the sdk, instead, we should use Armeria's contentType.isJson()

Yes it's a defense in depth to turn **any** skip of auth into an error. This is one baby step forward. I will have more PRs to handle other issue including the JSON type matching, JSON completeness guessing, etc.

server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizationGateConverter.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Override
+  public Object convertRequest(
+      ServiceRequestContext ctx,
+      AggregatedHttpRequest request,
+      Class<?> expectedResultType,
+      ParameterizedType expectedParameterizedResultType)
+      throws Exception {
+    if (Boolean.TRUE.equals(ctx.attr(UnityAccessDecorator.AUTH_PENDING))) {
```
openinx, line 42: when will we set this `UnityAccessDecorator.AUTH_PENDING` attr in the context ?

server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizationGateConverter.java
```diff
@@ -0,0 +1,49 @@
...
+ * <p>Scope: gates only parameters bound via the converter chain. Endpoints with no body-bound
+ * parameter (pure {@code @Param}/path) don't trigger this converter, but those endpoints also don't
+ * use PAYLOAD locators, so the decorator never sets the flag and no gate is needed.
+ */
+public final class AuthorizationGateConverter implements RequestConverterFunction {
+
+  private final RequestConverterFunction delegate;
+
+  public AuthorizationGateConverter(RequestConverterFunction delegate) {
+    this.delegate = Objects.requireNonNull(delegate, "delegate");
```
openinx: Can you give a more clear error message if it's null ?

server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizationGateConverter.java
```diff
@@ -0,0 +1,49 @@
...
+import java.lang.reflect.ParameterizedType;
+import java.util.Objects;
+
+/**
+ * Fail-closed wrapper that enforces {@link UnityAccessDecorator}'s PAYLOAD-source authorization
+ * before delegating to an inner body converter (e.g. Jackson). The decorator sets {@link
+ * UnityAccessDecorator#AUTH_PENDING} on entry to the PAYLOAD path and clears it only after the
+ * authorization evaluation succeeds; if the flag is still {@code true} when a parameter is being
+ * bound, {@code peekData}'s callback never fired (no body, non-JSON content-type, malformed JSON,
+ * trailing whitespace) and the request must be denied.
```
openinx, line 17: >if the flag is still {@code true} when a parameter is being
> * bound, {@code peekData}'s callback never fired (no body, non-JSON content-type, malformed JSON,
 > * trailing whitespace) and the request must be denied.


https://github.com/unitycatalog/unitycatalog/pull/1494/changes#diff-b9504c8a386aa1943559c357fa7ce1fcc952f4e0f45df99b89b17807213699faR42-R45 
In that case,  we actually are throwing the `PERMISSION_DENIED` exception, which actually swallow the real malformed detailed.  will that make it even worst to the developer to understand the response ?

server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizationGateConverter.java
yili-db: It is set here: https://github.com/unitycatalog/unitycatalog/pull/1494/changes#diff-9e4896b86c010ff1d4cf496f17095ee6bf98561fc8da1bceade0cb542af3a599R202 when it is sure that some auth parameters reside in the payload.

server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizationGateConverter.java
yili-db: This is not the perfect error message throwing out.
The priority here is security: making sure it fails with `PERMISSION_DENIED` instead of going through.

In a follow up PR I will deal with this malformed/incomplete body problem and get rid of the `processPeekData` entirely.

server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizationGateConverter.java
yili-db: This is never meant to be null at all. It's just the production version of `assert`.
Removing this `"delegate"` string. The exception `NullPointerException` should be self explanatory.

openinx (approved):
LGTM, thanks @yili-db .

PR #1498, opened 2026-04-19
fix: increase property_value column length to support Azure SAS tokens

sambhav8695:
**PR Checklist**

**Description of changes**
## Problem
When creating a Parquet table backed by Azure Data Lake Storage (ADLS),
Unity Catalog stores Azure SAS tokens as table properties. These tokens
routinely exceed 255 characters (observed: 306+ chars), causing a SQL
`VALUE_TOO_LONG` error on insert:

This prevents any Azure-backed table creation when SAS token
authentication is used via client ID / tenant ID / app registration.

## Root Cause
`PropertyDAO` maps `property_value` with `length = 255`. Azure SAS tokens
generated by the credential vending path are longer than this limit.

## Fix
Increase the column length from 255 → 2000 in the `PropertyDAO` entity
annotation, and add a corresponding DB migration (if applicable).

## Testing
- [x] Created a Parquet table on ADLS Gen2 with SAS token auth
- [x] Verified `uc_properties` row is persisted successfully
- [x] Existing unit/integration tests pass

## Notes
- No sensitive credentials are included in this PR
- This is a backward-compatible schema change (widening a column)

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/dao/PropertyDAO.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/dao/PropertyDAO.java
@@ -49,7 +49,7 @@ public class PropertyDAO {
   @Column(name = "property_key", nullable = false)
   private String key;
 
-  @Column(name = "property_value", nullable = false)
+  @Column(name = "property_value", nullable = false, length = 2000)
   private String value;
 
   public static List<PropertyDAO> from(


yili-db:
@sambhav8695 Do you mean storing Azure SAS tokens in table properties? That doesn't look like a correct use case of table property.

yili-db:
Ah now I have a better idea what is the real problem. The token was sent to server for parquet tables (but not Delta).
This has been fixed in https://github.com/unitycatalog/unitycatalog/pull/1789

sambhav8695:
Thanks @yili-db  for looking into this and identifying the actual root cause.

My initial investigation focused on the `VALUE_TOO_LONG` failure caused by the SAS token exceeding the `property_value` limit, but I agree that increasing the column size would only address the symptom.

The important issue is that the credential was being sent to the UC server for Parquet tables in the first place. It's good to see that this has now been addressed in #1789 .

Thanks for the clarification and the fix!

PR #1500, opened 2026-04-20
Delta REST Catalog: vend temporary table credentials

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1500/files) to review incremental changes.
- [**stack/drc_cred**](https://github.com/unitycatalog/unitycatalog/pull/1500) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1500/files)]
  - [stack/staging_tables](https://github.com/unitycatalog/unitycatalog/pull/1511) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1511/files/2afe5b9fe0b9c0c80409bfbf2d4a13023b4f265a..1f366b09cb1233fda4164c59d82fb563ba041f12)]
    - [stack/create_table](https://github.com/unitycatalog/unitycatalog/pull/1516) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1516/files/1f366b09cb1233fda4164c59d82fb563ba041f12..bf3490620d18fae22fc342636e9ca848fa1fdd70)]
      - stack/drc_update_table

---------
**PR Checklist**

**Description of changes**
Delta REST Catalog: vend temporary table credentials

Adds GET endpoints for cloud storage credentials: one scoped by operation (READ / READ_WRITE) for regular tables, one always-READ_WRITE for staging tables. UC's nested TemporaryCredentials are flattened to the spec's s3.* / azure.* / gcs.* config fields via a new DeltaCredentialsMapper.

Also drops `additionalProperties` from the StorageCredentialConfig schema so typed setters actually populate the JSON response (previously dead — the generated class extended HashMap and Jackson bypassed bean introspection). A regression test pins the sparse-JSON wire contract so the NON_NULL mapper dependency stays explicit.

Staging-table credential lookup uses a dedicated getStagingTableStorageLocation() so the endpoint rejects regular-table UUIDs instead of silently vending credentials against them.

### DIFF
--- a/api/delta.yaml
+++ b/api/delta.yaml
@@ -1307,7 +1307,7 @@ components:
         timestamp:
           type: integer
           format: int64
-          description: In-commit timestamp
+          description: In-commit timestamp, in epoch milliseconds
         file-name:
           type: string
           description: UUID-based commit file name
@@ -1318,7 +1318,7 @@ components:
         file-modification-timestamp:
           type: integer
           format: int64
-          description: File modification timestamp
+          description: File modification timestamp, in epoch milliseconds
       required:
         - version
         - timestamp
--- a/api/delta.yaml
+++ b/api/delta.yaml
@@ -618,8 +618,8 @@ components:
         config:
           type: object
           description: |
-            Cloud provider-specific credential configuration as string key-value pairs.
-            Not all keys are present -- only keys for the relevant cloud provider are included.
+            Cloud provider-specific credential configuration.
+            Only keys for the relevant cloud provider are populated; the others are null.
           properties:
             s3.access-key-id:
               type: string
@@ -636,8 +636,6 @@ components:
             gcs.oauth-token:
               type: string
               description: GCP OAuth token
-          additionalProperties:
-            type: string
         expiration-time-ms:
           type: integer
           format: int64
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
@@ -9,6 +9,7 @@ import com.linecorp.armeria.server.annotation.ExceptionHandler;
 import com.linecorp.armeria.server.annotation.Get;
 import com.linecorp.armeria.server.annotation.Param;
 import com.linecorp.armeria.server.annotation.ProducesJson;
+import io.unitycatalog.server.auth.AuthorizeExpressions;
 import io.unitycatalog.server.auth.UnityCatalogAuthorizer;
 import io.unitycatalog.server.auth.annotation.AuthorizeExpression;
 import io.unitycatalog.server.auth.annotation.AuthorizeKey;
@@ -20,7 +21,6 @@ import io.unitycatalog.server.delta.model.LoadTableResponse;
 import io.unitycatalog.server.exception.BaseException;
 import io.unitycatalog.server.exception.DeltaRestExceptionHandler;
 import io.unitycatalog.server.exception.ErrorCode;
-import io.unitycatalog.server.model.TableInfo;
 import io.unitycatalog.server.model.TemporaryCredentials;
 import io.unitycatalog.server.persist.CatalogRepository;
 import io.unitycatalog.server.persist.Repositories;
@@ -102,15 +102,7 @@ public class DeltaRestCatalogService extends AuthorizedService {
 
   @Get("/delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}")
   @ProducesJson
-  @AuthorizeExpression(
-      """
-      #authorize(#principal, #metastore, OWNER) ||
-      #authorize(#principal, #catalog, OWNER) ||
-      (#authorize(#principal, #schema, OWNER) && #authorize(#principal, #catalog, USE_CATALOG)) ||
-      (#authorize(#principal, #schema, USE_SCHEMA) &&
-          #authorize(#principal, #catalog, USE_CATALOG) &&
-          #authorizeAny(#principal, #table, OWNER, SELECT, MODIFY))
-      """)
+  @AuthorizeExpression(AuthorizeExpressions.GET_TABLE)
   @AuthorizeResourceKey(METASTORE)
   public LoadTableResponse loadTable(
       @Param("catalog") @AuthorizeResourceKey(CATALOG) String catalog,
@@ -129,25 +121,13 @@ public class DeltaRestCatalogService extends AuthorizedService {
    */
   @Get("/delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}/credentials")
   @ProducesJson
-  // Mirror TemporaryTableCredentialsService's rule. Credential vending is stricter than metadata
-  // reads: admins/owners above the table still need an explicit table-level privilege. Keep this
-  // in sync with TemporaryTableCredentialsService so both entry points grant the same access.
-  @AuthorizeExpression(
-      """
-      #authorizeAny(#principal, #schema, OWNER, USE_SCHEMA) &&
-      #authorizeAny(#principal, #catalog, OWNER, USE_CATALOG) &&
-      (#operation == 'READ'
-          ? #authorizeAny(#principal, #table, OWNER, SELECT)
-          : (#authorize(#principal, #table, OWNER) ||
-              #authorizeAll(#principal, #table, SELECT, MODIFY)))
-      """)
+  @AuthorizeExpression(AuthorizeExpressions.VEND_TABLE_CREDENTIAL)
   public CredentialsResponse getTableCredentials(
       @Param("catalog") @AuthorizeResourceKey(CATALOG) String catalog,
       @Param("schema") @AuthorizeResourceKey(SCHEMA) String schema,
       @Param("table") @AuthorizeResourceKey(TABLE) String table,
       @Param("operation") @AuthorizeKey CredentialOperation operation) {
-    TableInfo tableInfo = tableRepository.getTable(catalog + "." + schema + "." + table);
-    NormalizedURL storageLocation = NormalizedURL.from(tableInfo.getStorageLocation());
+    NormalizedURL storageLocation = tableRepository.getTableStorageLocation(catalog, schema, table);
     TemporaryCredentials credentials =
         storageCredentialVendor.vendCredential(storageLocation, toPrivileges(operation));
     return DeltaCredentialsMapper.toCredentialsResponse(
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
@@ -11,20 +11,28 @@ import com.linecorp.armeria.server.annotation.Param;
 import com.linecorp.armeria.server.annotation.ProducesJson;
 import io.unitycatalog.server.auth.UnityCatalogAuthorizer;
 import io.unitycatalog.server.auth.annotation.AuthorizeExpression;
+import io.unitycatalog.server.auth.annotation.AuthorizeKey;
 import io.unitycatalog.server.auth.annotation.AuthorizeResourceKey;
 import io.unitycatalog.server.delta.model.CatalogConfig;
+import io.unitycatalog.server.delta.model.CredentialOperation;
+import io.unitycatalog.server.delta.model.CredentialsResponse;
 import io.unitycatalog.server.delta.model.LoadTableResponse;
 import io.unitycatalog.server.exception.BaseException;
 import io.unitycatalog.server.exception.DeltaRestExceptionHandler;
 import io.unitycatalog.server.exception.ErrorCode;
+import io.unitycatalog.server.model.TableInfo;
+import io.unitycatalog.server.model.TemporaryCredentials;
 import io.unitycatalog.server.persist.CatalogRepository;
 import io.unitycatalog.server.persist.Repositories;
 import io.unitycatalog.server.persist.TableRepository;
 import io.unitycatalog.server.service.AuthorizedService;
+import io.unitycatalog.server.service.credential.CredentialContext;
 import io.unitycatalog.server.service.credential.StorageCredentialVendor;
+import io.unitycatalog.server.utils.NormalizedURL;
 import io.unitycatalog.server.utils.ServerProperties;
-
 import java.util.List;
+import java.util.Set;
+import java.util.UUID;
 
 /**
  * Delta REST Catalog Service -- REST API for Delta tables.
@@ -52,6 +60,7 @@ public class DeltaRestCatalogService extends AuthorizedService {
 
   private final CatalogRepository catalogRepository;
   private final TableRepository tableRepository;
+  private final StorageCredentialVendor storageCredentialVendor;
 
   public DeltaRestCatalogService(
       UnityCatalogAuthorizer authorizer,
@@ -61,6 +70,7 @@ public class DeltaRestCatalogService extends AuthorizedService {
     super(authorizer, repositories, serverProperties);
     this.catalogRepository = repositories.getCatalogRepository();
     this.tableRepository = repositories.getTableRepository();
+    this.storageCredentialVendor = storageCredentialVendor;
   }
 
   // ==================== Configuration API ====================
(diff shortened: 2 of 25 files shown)


api/delta.yaml
```diff
@@ -618,8 +618,8 @@ components:
         config:
           type: object
           description: |
-            Cloud provider-specific credential configuration as string key-value pairs.
-            Not all keys are present -- only keys for the relevant cloud provider are included.
+            Cloud provider-specific credential configuration.
+            Only keys for the relevant cloud provider are populated; the others are null.
```
tdas, line 622: shouldnt this change MD files?

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
```diff
@@ -105,4 +116,90 @@ public LoadTableResponse loadTable(
...
+
+  /**
+   * Vend temporary cloud storage credentials for a table identified by three-part name. The {@code
+   * operation} query param scopes the returned credentials: {@code READ} requires {@code SELECT}
+   * (or OWNER); {@code READ_WRITE} requires {@code OWNER} or both {@code SELECT} and {@code
+   * MODIFY}.
+   */
+  @Get("/delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}/credentials")
+  @ProducesJson
+  // Allow vending credentials to (in priority order):
```
tdas, line 130: is the auth logic same as the UC credential endpoint?

yili-db: Yes it is

api/delta.yaml
yili-db: These descriptions never made it to the MD files due to limitation of the OpenAPI doc generator.
They are made part of the generated `StorageCredentialConfig.java` file.

api/delta.yaml
```diff
@@ -636,8 +636,6 @@ components:
             gcs.oauth-token:
               type: string
               description: GCP OAuth token
-          additionalProperties:
```
yili-db, line 639: This is removed because the OpenAPI generated code can't actually support both named properties and additional unnamed properties.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
```diff
@@ -105,4 +116,82 @@ public LoadTableResponse loadTable(
...
+  // in sync with TemporaryTableCredentialsService so both entry points grant the same access.
+  @AuthorizeExpression(
+      """
+      #authorizeAny(#principal, #schema, OWNER, USE_SCHEMA) &&
+      #authorizeAny(#principal, #catalog, OWNER, USE_CATALOG) &&
+      (#operation == 'READ'
+          ? #authorizeAny(#principal, #table, OWNER, SELECT)
+          : (#authorize(#principal, #table, OWNER) ||
+              #authorizeAll(#principal, #table, SELECT, MODIFY)))
+      """)
```
tdas: [AI Assisted] This `@AuthorizeExpression` string is character-for-character identical to `TemporaryTableCredentialsService.java:38-45`. The inline comment above says 'Keep this in sync with TemporaryTableCredentialsService' — but comments don't get enforced and drift here is near-certain over time, which is exactly the 'multiplicative maintenance burden' pattern we try to avoid.

Suggest extracting to a shared constant so there's literally one source of truth:

```java
public final class TableCredentialAuthz {
  public static final String VEND_TABLE_CREDENTIAL_EXPRESSION = """
      #authorizeAny(#principal, #schema, OWNER, USE_SCHEMA) &&
      #authorizeAny(#principal, #catalog, OWNER, USE_CATALOG) &&
      (#operation == 'READ'
          ? #authorizeAny(#principal, #table, OWNER, SELECT)
          : (#authorize(#principal, #table, OWNER) ||
              #authorizeAll(#principal, #table, SELECT, MODIFY)))
      """;
}
```

and then both sites use `@AuthorizeExpression(TableCredentialAuthz.VEND_TABLE_CREDENTIAL_EXPRESSION)`. `@AuthorizeExpression`'s value is a constant-expression attribute so this works. An alternative is a composed meta-annotation `@AuthorizeTableCredentialVending`.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
```diff
@@ -105,4 +116,82 @@ public LoadTableResponse loadTable(
...
+  public CredentialsResponse getTableCredentials(
+      @Param("catalog") @AuthorizeResourceKey(CATALOG) String catalog,
+      @Param("schema") @AuthorizeResourceKey(SCHEMA) String schema,
+      @Param("table") @AuthorizeResourceKey(TABLE) String table,
+      @Param("operation") @AuthorizeKey String operation) {
+    CredentialOperation op = parseOperation(operation);
+    TableInfo tableInfo = tableRepository.getTable(catalog + "." + schema + "." + table);
```
tdas: [AI Assisted] Two small issues with `getTable(catalog + "." + schema + "." + table)`:

1. **Round-trip.** `TableRepository.getTable(String)` immediately does `fullName.split("\\.")` and rejects anything whose parts count isn't 3. We're reconstructing a dotted name only for `getTable` to split it back. Not correctness-critical but wasted work and obscures intent.
2. **Dot-in-name.** If any of `catalog`/`schema`/`table` contains a `.`, the reconstructed string can't be unpacked back correctly. Not sure whether dots are reachable here, but the safer form is right there — `loadTableForDelta(catalog, schema, table)` (right next door in the same file) already takes three separate strings and calls `findTable(session, catalog, schema, table)` directly.

I see `IcebergRestCatalogService.java:155,176` uses the same pattern, so this PR isn't introducing it — but worth adding a sibling `public TableInfo getTable(String catalog, String schema, String table)` in `TableRepository` (≤10 lines, wrapping `findTable`) and using it here, so future DRC endpoints have the right API to reach for. Even cleaner: add a lightweight `getStorageLocation(catalog, schema, table)` returning just `NormalizedURL` (mirroring the staging helper added in this PR), since the endpoint only uses `storageLocation` off the returned `TableInfo` anyway — fetching the full `TableInfo` is more work than needed.

server/src/test/java/io/unitycatalog/server/sdk/access/SdkTableAccessControlCRUDTest.java
```diff
@@ -152,6 +162,36 @@ public void testTableAccess() {
...
+    // authz test cares only that it's not PERMISSION_DENIED.
+    try {
+      regular2DeltaCredsApi.getTableCredentials(
+          CredentialOperation.READ_WRITE, "cat_pr1", "sch_pr1", "tbl_pr1");
+    } catch (ApiException ex) {
+      int forbiddenCode = ErrorCode.PERMISSION_DENIED.getHttpStatus().code();
+      assertThat(ex.getCode())
+          .as("expected non-403 after authz passes")
+          .isNotEqualTo(forbiddenCode);
+    }
```
tdas: [AI Assisted] This test passes silently if `getTableCredentials` doesn't throw at all — the `isNotEqualTo(forbiddenCode)` assertion is only reached on the exception path. So the test technically can't tell the difference between 'authz passed, call succeeded' and 'authz passed, call threw a 500' versus 'authz passed, call threw a 403 that was never checked'. (The last one can't happen with current control flow, but the shape of the test is weaker than the intent.)

Tighter form, makes the 'no throw is a pass' state explicit:

```java
Throwable thrown = catchThrowable(() ->
    regular2DeltaCredsApi.getTableCredentials(
        CredentialOperation.READ_WRITE, "cat_pr1", "sch_pr1", "tbl_pr1"));
if (thrown != null) {
  assertThat(((ApiException) thrown).getCode())
      .as("authz passes; any failure must not be PERMISSION_DENIED")
      .isNotEqualTo(ErrorCode.PERMISSION_DENIED.getHttpStatus().code());
}
```

Or — preferred — if vending reliably succeeds for this path in the test env, just `assertThat(result).isNotNull()`.

server/src/test/java/io/unitycatalog/server/sdk/delta/SdkDeltaCredentialsTest.java
```diff
@@ -0,0 +1,163 @@
...
+    assertThatThrownBy(() -> deltaCredentialsApi.getStagingTableCredentials(randomId))
+        .isInstanceOf(ApiException.class)
+        .hasMessageContaining("Staging table not found");
+
+    // -------- getStagingTableCredentials: rejects regular-table UUID --------
+    // Endpoint is scoped to staging tables only; passing a real table's UUID must 404.
+    UUID regularTableId = UUID.fromString(readTable.getTableId());
+    assertThatThrownBy(() -> deltaCredentialsApi.getStagingTableCredentials(regularTableId))
+        .isInstanceOf(ApiException.class)
+        .hasMessageContaining("Staging table not found");
```
tdas: [AI Assisted] Testing gap (minor): this case covers 'caller passes a regular-table UUID they own' → 404, which exercises the new `getStagingTableStorageLocation` reject logic. But the 'caller passes a regular-table UUID they don't own' case isn't exercised anywhere — it would hit authz first (since `@AuthorizeResourceKey(TABLE)` resolves the regular-table UUID in the authz graph and the caller isn't OWNER) and return 403, not 404.

Not a correctness concern (fail-closed is fine) but worth one assertion + comment so the expected-status story is documented. Fits better in `SdkStagingTableAccessControlTest` since it needs authz-enabled base.

server/src/test/java/io/unitycatalog/server/service/delta/DeltaCredentialsMapperTest.java
```diff
@@ -0,0 +1,157 @@
...
+    StorageCredentialConfig config =
+        DeltaCredentialsMapper.toCredentialsResponse(
+                "s3://bucket/path", uc, CredentialOperation.READ)
+            .getStorageCredentials()
+            .get(0)
+            .getConfig();
+
+    // Use the same mapper config as the Delta REST Catalog response converter.
+    ObjectMapper deltaMapper =
+        JsonMapper.builder().serializationInclusion(JsonInclude.Include.NON_NULL).build();
```
tdas: [AI Assisted] This mapper config (`JsonMapper.builder().serializationInclusion(NON_NULL).build()`) is hand-duplicated from `UnityCatalogServer.java:298-299`. The test's own docstring says 'Use the same mapper config as the Delta REST Catalog response converter' — but they can drift independently and this sort of hand-maintained contract is exactly what the 'regression test pin' is *trying* to prevent.

I see `DeltaModelSerializationTest.java:53` already has the same rebuild, so this PR is following an existing pattern — fine, but a small follow-up opportunity: expose a `DeltaRestCatalogMappers.RESPONSE_MAPPER` (or similar) constant in the server module and have both tests + `UnityCatalogServer` import it. Alternatively, a comment at `UnityCatalogServer.java:298` pointing at these tests so a dev touching the server mapper knows to update them in lockstep.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
```diff
@@ -105,4 +116,82 @@ public LoadTableResponse loadTable(
...
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT, "Missing required query parameter: operation");
+    }
+    try {
+      return CredentialOperation.fromValue(operation);
+    } catch (IllegalArgumentException e) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT,
+          "Invalid operation: " + operation + ". Must be one of: READ, READ_WRITE");
+    }
```
tdas: [AI Assisted] Testing gap: `parseOperation`'s second throw branch (the `catch (IllegalArgumentException)` → 'Invalid operation: ...') is currently unexercised. `SdkDeltaCredentialsTest` only covers the null-operation case, and that is caught by the SDK's client-side required-parameter check before it reaches the server — so this branch isn't hit on any test path.

A server-side test that bypasses the SDK validator and sends a raw HTTP request with e.g. `?operation=nonsense` would cover it. Could be a small `WebClient`-based test alongside the others, or a curl-style test.

server/src/test/java/io/unitycatalog/server/sdk/access/SdkTableAccessControlCRUDTest.java
```diff
@@ -152,6 +162,36 @@ public void testTableAccess() {
...
+        () ->
+            regular2DeltaCredsApi.getTableCredentials(
+                CredentialOperation.READ_WRITE, "cat_pr1", "sch_pr1", "tbl_pr1"));
+
+    // Grant MODIFY to regular-2 so the READ_WRITE path passes authz.
+    grantPermissions(
+        "regular-2@localhost", SecurableType.TABLE, "cat_pr1.sch_pr1.tbl_pr1", Privileges.MODIFY);
+
+    // getTableCredentials READ_WRITE (regular-2) -> SELECT + MODIFY -> authz passes.
+    // Credential vending against /tmp/tbl_pr1 will fail with a non-authz error (no cloud config);
```
tdas: [AI Assisted] Testing gap: the READ_WRITE authz expression has an OR branch — `#authorize(#principal, #table, OWNER) || #authorizeAll(#principal, #table, SELECT, MODIFY)`. The new tests exercise the `SELECT && MODIFY` arm but not the `OWNER` arm (i.e., OWNER without explicit MODIFY). Adding a single case where the caller is TABLE owner but has no MODIFY grant would cover the OR completely and give full conditional-branch coverage on the READ_WRITE path.

tdas (commented):
[AI Assisted] Overall this is a well-scoped, well-tested change — clean split between mapper (unit tested per-cloud), endpoints (integration tested), and authz (access-control tested separately). The `additionalProperties`/Jackson-HashMap fix is a real latent bug well-pinned by a regression test, and the staging-only `getStagingTableStorageLocation` is a nice targeted tightening.

A few suggestions below; none are blocking. Most consequential is the verbatim-duplicated `@AuthorizeExpression` string (also called out only by a 'keep in sync' comment) — worth extracting to a shared constant so the two credential-vending paths literally share one source of truth.

Also flagging that this touches authz/security, so would be good to get explicit sign-off from an authz module owner in addition to the DRC reviewers.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
yili-db: Done.
Extracted to AuthorizeExpressions.VEND_TABLE_CREDENTIAL (chose AuthorizeExpressions over TableCredentialAuthz so the class can hold future shared expressions — e.g., GET_TABLE is already in there too, shared between TableService.getTable and DeltaRestCatalogService.loadTable).
Both sites now use `@AuthorizeExpression(AuthorizeExpressions.VEND_TABLE_CREDENTIAL)` — single source of truth.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
yili-db: Done. Added TableRepository.getTableStorageLocation(catalog, schema, table) that takes the three parts directly and returns just NormalizedURL (mirroring the staging helper). getTableCredentials uses it — no dotted-name round-trip, and no hydrating the full TableInfo when we only need the storage URL.

server/src/test/java/io/unitycatalog/server/sdk/access/SdkTableAccessControlCRUDTest.java
yili-db: Done. Went with the preferred form — vending does succeed reliably in this path, so the test now asserts the positive shape directly.
Extracted assertReadWriteCredentialsVended(api) that pins .hasSize(1) and operation == READ_WRITE on the returned credential. Reused for both the SELECT && MODIFY arm and the new OWNER-arm case (see r3124871533), so both OR branches assert identically.

server/src/test/java/io/unitycatalog/server/service/delta/DeltaCredentialsMapperTest.java
yili-db: Done. Added `DeltaRestCatalogMappers` holding a single `MAPPER` constant

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
yili-db: `parseOperation` has been removed. The method has switched to use a enum typed `CredentialOperation operation`.
Now the question is whether the enum parameter handling invalid input from request and what error response is server replying in this case. This is handled by another WIP PR https://github.com/unitycatalog/unitycatalog/pull/1512 to make sure enum or integer or json payload is correctly mapped to `400 INVALID_ARGUMENT`

server/src/test/java/io/unitycatalog/server/sdk/access/SdkTableAccessControlCRUDTest.java
yili-db: Done. Added a case at [SdkTableAccessControlCRUDTest.java:190-193](https://github.com/unitycatalog/unitycatalog/pull/1500/changes#diff-38daaa1980e0eddbe602227ed095744f4a93b62bf35bee258925b4fd17c9718fR190)

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
```diff
@@ -89,20 +100,63 @@ public CatalogConfig getConfig(
 
   @Get("/delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}")
   @ProducesJson
-  @AuthorizeExpression(
```
openinx, line 95: qq:  I actually don't understand what're those parts for ?  what's the diff between keeping them and removing them ?

openinx: oh,  I see.  you moved those to a constant string.

openinx (approved):
LGTM, thanks @yili-db .

tdas (approved):

PR #1503, opened 2026-04-20
Add CI job running Spark tests with deltaRestApi.enabled=true

timothyw553:
**PR Checklist**

**Description of changes**

Adds `.github/workflows/spark-tests-drc-enabled.yml` that runs the Spark connector test suite with the DRC flag on. Surfaces the flag as both the `DELTA_REST_API_ENABLED` env var and the sbt system property `-DdeltaRestApiEnabled=true`, so later PRs can plumb either into the test harness.

Today the flag toggles no behavior (the flag-scaffolding PR only reads it into a field), so this job's current value is regression protection: flag-on must not break any existing test.

**User impact:** none. New CI job runs alongside existing ones; does not gate merges today.

**Testing:** workflow runs green on this PR's own CI (to verify).

### DIFF
--- a/.github/workflows/spark-tests-drc-enabled.yml
+++ b/.github/workflows/spark-tests-drc-enabled.yml
@@ -1,0 +1,38 @@
+name: Spark Tests (DRC enabled)
+
+# Runs the Spark connector test suite with the Delta REST Catalog flag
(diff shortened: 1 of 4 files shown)


TimothyW553:
Superseded — the CI job belongs on the delta-io/delta side, not UC. Will be reopened against delta-io/delta separately.

PR #1504, opened 2026-04-20
Add DeltaRestClientProvider; wire UCSingleCatalog and UCProxy

timothyw553:
**PR Checklist**

**Description of changes**

Introduces the `DeltaRestClientProvider` interface at `clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java` and wires it into both `UCSingleCatalog` and `UCProxy`.

- **`UCProxy`** is the single flag gate: takes `deltaRestApiEnabled` in its ctor, holds a lazy DRC `TablesApi`, and returns `null` from `getDeltaTablesApi()` when the flag is off.
- **`UCSingleCatalog`** implements the same interface and forwards via its new `ucProxy` field (not through `delegate`), so the call resolves whether `delegate` is `UCProxy` directly or a `DeltaCatalog` wrapper.

The interface lives in the UC client jar so both UC connector and Delta can depend on it without a circular dep. Delta's `UCDeltaClient` (delta-io/delta#6575) will wrap the `ApiClient` returned by `getApiClient()`.

**Why two implementers:**
- Delta-side callers (e.g. `AbstractDeltaCatalog.loadTable`) reach `UCProxy` via `this.delegate` (set by `setDelegateCatalog` at `UCSingleCatalog.scala:76`) and can cast directly.
- External callers (e.g. a future commit coordinator) reach `UCSingleCatalog` by catalog name.

A single implementer would force one of those callers through a brittle lookup.

**User impact:** none. The interface is wired but no consumer calls it yet; runtime behavior is unchanged regardless of the flag.

**Testing:** added `DeltaRestClientProviderTest` — 4 cases: flag on returns shared `TablesApi`, flag off returns `null` from both entry points, forwarding works when `delegate` is a non-provider mock (simulates Delta-loaded prod), `TablesApi` is cached across calls. `sbt spark/testOnly *DeltaRestClientProviderTest *UCSingleCatalogDeltaRestApiFlagTest` — 7/7 pass.

### DIFF
--- a/.github/workflows/spark-tests-drc-enabled.yml
+++ b/.github/workflows/spark-tests-drc-enabled.yml
@@ -1,0 +1,38 @@
+name: Spark Tests (DRC enabled)
+
+# Runs the Spark connector test suite with the Delta REST Catalog flag
+# (spark.sql.catalog.<cat>.deltaRestApi.enabled) turned on. Future PRs will
+# plumb this env var through the test harness to toggle DRC routing in
+# integration tests. For now this job proves the flag-on path compiles and
+# existing tests stay green with the system property set.
+
+on:
+  push:
+    branches: [ "main", "branch-*" ]
+  pull_request:
+    branches: [ "main", "branch-*" ]
+
+permissions:
+  contents: read
+
+jobs:
+  spark-tests-drc-enabled:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


TimothyW553:
Superseded by a redo: closing and rebuilding the provider wiring PR with tests that drive the public TableCatalog API instead of reflecting on private fields.

PR #1505, opened 2026-04-21
When MAVEN_PROXY_URL is set, also use local as a repo

yili-db:
**PR Checklist**

**Description of changes**
When MAVEN_PROXY_URL is set, also use local as a repo. This is needed by `sbt publishLocal` when using maven proxy. Otherwise an error pops up:
```
[error] (server / publishLocal) Undefined resolver 'local'
```

### DIFF
--- a/build/sbt
+++ b/build/sbt
@@ -67,10 +67,11 @@ realpath () {
 }
 
 if [[ -n "$MAVEN_PROXY_URL" ]]; then
-  # If MAVEN_PROXY_URL is set, use it as the sole repository for all dependencies.
+  # If MAVEN_PROXY_URL is set, use it (and local) as the sole repository for all dependencies.
   SBT_REPOSITORIES_CONFIG=$(mktemp)
   cat > "$SBT_REPOSITORIES_CONFIG" <<EOF
 [repositories]
+  local
   maven-proxy: $MAVEN_PROXY_URL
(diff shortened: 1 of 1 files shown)


openinx (approved):
Looks good to me.

build/sbt
```diff
@@ -67,10 +67,11 @@ realpath () {
 }
 
 if [[ -n "$MAVEN_PROXY_URL" ]]; then
-  # If MAVEN_PROXY_URL is set, use it as the sole repository for all dependencies.
+  # If MAVEN_PROXY_URL is set, use it (and local) as the sole repository for all dependencies.
   SBT_REPOSITORIES_CONFIG=$(mktemp)
   cat > "$SBT_REPOSITORIES_CONFIG" <<EOF
 [repositories]
+  local
```
seewishnew, line 74: Would it might make sense for this to come after `maven-proxy` so that we're not running the risk of cross-polluting builds? 

This way `sbt publishLocal` will find what it's looking for outputting the jar, but during compile/update it will also always consult fetching from the maven-proxy remote repository first before consulting local ivy cache repository for this unity-catalog build.

seewishnew: Discussed offline, more of a nit, not a blocker.

tdas (approved):

PR #1509, opened 2026-04-21
Add DeltaRestClientProvider and deltaRestApi.enabled runtime flag

timothyw553:
**PR Checklist**

**Description of changes**

Adds:

1. A per-catalog Spark option `deltaRestApi.enabled` (default `false`).
```
spark.sql.catalog.unity=io.unitycatalog.spark.UCSingleCatalog
spark.sql.catalog.unity.uri=http://localhost:8080
spark.sql.catalog.unity.token=<token>
spark.sql.catalog.unity.deltaRestApi.enabled=true
```
2. A Java interface `DeltaRestClientProvider` that Delta Spark uses to reach UC's HTTP client from a UC catalog plugin.

```java
public interface DeltaRestClientProvider {
  Optional<DeltaTablesApi> getDeltaTablesApi();   // present iff deltaRestApi.enabled=true
  ApiClient getApiClient();                       // always non-null once the catalog is initialized
}
```

`DeltaTablesApi` = `io.unitycatalog.client.delta.api.TablesApi` — the codegen client for DRC endpoints. Not the legacy `io.unitycatalog.client.api.TablesApi`. `ApiClient` is UC's base HTTP client.

Both `UCProxy` and `UCSingleCatalog` implement the interface — Delta reaches `UCProxy` via its own delegate chain, external callers (e.g. the future commit coordinator) reach `UCSingleCatalog` by catalog name.

**Architecture**

```
UCSingleCatalog : DeltaRestClientProvider              (the class Spark registers)
   │
   └── delegate = DeltaCatalog                         (when Delta is on the classpath,
         │                                              the production default;
         │                                              if not, delegate = UCProxy directly)
         │
         └── delegateCatalog = UCProxy : DeltaRestClientProvider
                                                        (set by UCSingleCatalog via
                                                         setDelegateCatalog at init)
               ├── deltaTablesApi   ← null when flag off, built at ctor when on;
               │                      wrapped in Optional by getDeltaTablesApi
               └── apiClient        ← returned by getApiClient unconditionally

Delta code pattern-matches its own delegate (= UCProxy) as DeltaRestClientProvider
and calls getApiClient() to build its own wrapper.
```

### DIFF
--- a/clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
+++ b/clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
@@ -1,0 +1,47 @@
+package io.unitycatalog.client.delta;
+
+import io.unitycatalog.client.ApiClient;
+import io.unitycatalog.client.delta.api.TablesApi;
+import java.util.Optional;
+
+/**
+ * Interface for catalog implementations that can provide a Delta REST Catalog {@link TablesApi}
+ * client. This enables a single shared client instance owned by the catalog (UCProxy) to be reused
+ * by all consumers -- AbstractDeltaCatalog, UCCommitCoordinatorClient, and the Kernel v2 connector
+ * -- instead of each creating their own HTTP client to the same UC server.
+ *
+ * <p>Implementations should return the same cached {@link TablesApi} instance on every call. The
+ * caller must not close or modify the returned client.
+ *
+ * <p>This interface lives in {@code unitycatalog-client} so that both the UC Spark connector (which
+ * implements it) and Delta Spark (which consumes it) can depend on it without a circular
+ * dependency.
+ */
+public interface DeltaRestClientProvider {
+
+  /**
+   * Returns the shared Delta REST Catalog {@link TablesApi} client when the Delta REST Catalog
+   * integration is enabled for this catalog (see the {@code deltaRestApi.enabled} option), or
+   * {@link Optional#empty()} when it is disabled. The returned instance is owned by the catalog and
+   * must not be closed by the caller.
+   *
+   * <p>The current Delta Spark consumer builds its own {@link TablesApi} from {@link
+   * #getApiClient()} rather than reusing this cached instance; this method is intended for future
+   * consumers (commit coordinator, Kernel v2 connector) that want to share UC's cached client
+   * instead of constructing their own.
+   *
+   * @return the cached DRC {@link TablesApi} when DRC is enabled, otherwise {@link
+   *     Optional#empty()} — indicating the caller should fall back to the legacy UC API via {@link
+   *     #getApiClient()}.
+   */
+  Optional<TablesApi> getDeltaTablesApi();
+
+  /**
+   * Returns the base {@link ApiClient} for UC API access. Consumers use it to construct typed
+   * wrappers (legacy {@code TablesApi}, {@code DeltaCommitsApi}, DRC {@code TablesApi}, etc.)
+   * against UC's shared HTTP pool. This exists so consumers don't build a duplicate {@link
+   * ApiClient} from URL + token (as Delta did previously) and end up with two HTTP connection pools
+   * pointing at the same UC server. Implementations must return a non-null client.
+   */
+  ApiClient getApiClient();
+}
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -2,6 +2,8 @@ package io.unitycatalog.spark
 
 import io.unitycatalog.client.api.{SchemasApi, TablesApi, TemporaryCredentialsApi}
 import io.unitycatalog.client.auth.TokenProvider
+import io.unitycatalog.client.delta.DeltaRestClientProvider
+import io.unitycatalog.client.delta.api.{TablesApi => DeltaTablesApi}
 import io.unitycatalog.client.model.{TableInfo, _}
 import io.unitycatalog.client.retry.JitterDelayRetryPolicy
 import io.unitycatalog.client.{ApiClient, ApiException}
@@ -22,7 +24,7 @@ import org.sparkproject.guava.base.Preconditions
 
 import java.net.URI
 import java.util
-import java.util.Locale
+import java.util.{Locale, Optional}
 import scala.collection.JavaConverters._
 import scala.collection.convert.ImplicitConversions._
 import scala.collection.mutable.ArrayBuffer
@@ -34,15 +36,18 @@ import scala.language.existentials
 class UCSingleCatalog
   extends StagingTableCatalog
   with SupportsNamespaces
+  with DeltaRestClientProvider
   with Logging {
 
   private[this] var uri: URI = null
   private[this] var tokenProvider: TokenProvider = null
   private[this] var renewCredEnabled: Boolean = false
   private[this] var credScopedFsEnabled: Boolean = false
-  private[this] var apiClient: ApiClient = null;
+  private[this] var deltaRestApiEnabled: Boolean = false
+  private[this] var apiClient: ApiClient = null
   private[this] var temporaryCredentialsApi: TemporaryCredentialsApi = null
   private[this] var tablesApi: TablesApi = null
+  private[this] var ucProxy: UCProxy = null
 
   @volatile private var delegate: TableCatalog = null
 
@@ -61,32 +66,43 @@ class UCSingleCatalog
     val serverSidePlanningEnabled = OptionsUtil.getBoolean(options,
       OptionsUtil.SERVER_SIDE_PLANNING_ENABLED,
       OptionsUtil.DEFAULT_SERVER_SIDE_PLANNING_ENABLED)
+    deltaRestApiEnabled = OptionsUtil.getBoolean(options,
+      OptionsUtil.DELTA_REST_API_ENABLED,
+      OptionsUtil.DEFAULT_DELTA_REST_API_ENABLED)
 
     apiClient = ApiClientFactory.createApiClient(
       JitterDelayRetryPolicy.builder().build(),uri, tokenProvider)
     temporaryCredentialsApi = new TemporaryCredentialsApi(apiClient)
     tablesApi = new TablesApi(apiClient)
-    val proxy = new UCProxy(uri, tokenProvider, renewCredEnabled, credScopedFsEnabled,
-      serverSidePlanningEnabled, apiClient, tablesApi, temporaryCredentialsApi)
-    proxy.initialize(name, options)
+    ucProxy = new UCProxy(uri, tokenProvider, renewCredEnabled, credScopedFsEnabled,
+      serverSidePlanningEnabled, deltaRestApiEnabled,
+      apiClient, tablesApi, temporaryCredentialsApi)
+    ucProxy.initialize(name, options)
     if (UCSingleCatalog.LOAD_DELTA_CATALOG.get()) {
       try {
         delegate = Class.forName("org.apache.spark.sql.delta.catalog.DeltaCatalog")
           .getDeclaredConstructor().newInstance().asInstanceOf[TableCatalog]
-        delegate.asInstanceOf[DelegatingCatalogExtension].setDelegateCatalog(proxy)
+        delegate.asInstanceOf[DelegatingCatalogExtension].setDelegateCatalog(ucProxy)
         delegate.initialize(name, options)
         UCSingleCatalog.DELTA_CATALOG_LOADED.set(true)
       } catch {
         case e: ClassNotFoundException =>
           logWarning("DeltaCatalog is not available in the classpath", e)
-          delegate = proxy
+          delegate = ucProxy
       }
     } else {
-      delegate = proxy
+      delegate = ucProxy
     }
   }
 
   override def name(): String = delegate.name()
+
+  // Forward directly to UCProxy, not through `delegate`. In production `delegate` is
+  // DeltaCatalog, which does not implement DeltaRestClientProvider — routing through
+  // it would always return empty even when the flag is on.
+  override def getDeltaTablesApi(): Optional[DeltaTablesApi] = ucProxy.getDeltaTablesApi()
+
+  override def getApiClient(): ApiClient = ucProxy.getApiClient()
 
   override def listTables(namespace: Array[String]): Array[Identifier] = delegate.listTables(namespace)
 
@@ -546,11 +562,27 @@ private class UCProxy(
     renewCredEnabled: Boolean,
     credScopedFsEnabled: Boolean,
     serverSidePlanningEnabled: Boolean,
+    deltaRestApiEnabled: Boolean,
     apiClient: ApiClient,
     tablesApi: TablesApi,
-    temporaryCredentialsApi: TemporaryCredentialsApi) extends TableCatalog with SupportsNamespaces with Logging {
+    temporaryCredentialsApi: TemporaryCredentialsApi)
+  extends TableCatalog
+  with SupportsNamespaces
+  with DeltaRestClientProvider
+  with Logging {
   private[this] var name: String = null
   private[this] var schemasApi: SchemasApi = null
+  // Null when DRC is disabled; constructed once at UCProxy ctor when enabled.
+  // The flag decides construction directly; callers see this through the Optional
+  // wrapper in getDeltaTablesApi().
+  private[this] val deltaTablesApi: DeltaTablesApi =
+    if (deltaRestApiEnabled) new DeltaTablesApi(apiClient) else null
+
+  override def getDeltaTablesApi(): Optional[DeltaTablesApi] =
+    Optional.ofNullable(deltaTablesApi)
+
+  // Always exposes the underlying ApiClient so callers can build legacy API wrappers.
+  override def getApiClient(): ApiClient = apiClient
 
   override def initialize(name: String, options: CaseInsensitiveStringMap): Unit = {
     this.name = name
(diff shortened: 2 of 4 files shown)


clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
```diff
@@ -0,0 +1,41 @@
...
+   */
+  TablesApi getDeltaTablesApi();
+
+  /**
+   * Returns the base {@link io.unitycatalog.client.ApiClient} for legacy UC API access,
+   * or {@code null} when the implementation does not expose its client. Returned as
+   * {@link Object} so that callers in modules that do not depend on the generated
+   * client package can receive it opaquely and cast when they need it.
+   */
+  default Object getApiClient() { return null; }
```
openinx: Why this return value is an `Object` rather than the concret interface or class ?  that looks like vague for the API

openinx: nit: does it has the checkstyle validation also ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -88,6 +96,17 @@ class UCSingleCatalog
 
   override def name(): String = delegate.name()
 
+  // External callers (commit coordinator, tests, anyone using Spark's catalog manager)
+  // only see UCSingleCatalog. Without this override they couldn't reach the DRC client
+  // that UCProxy holds, because UCProxy is package-private.
+  override def getDeltaTablesApi(): DeltaTablesApi = delegate match {
```
openinx: why will we use this `getDeltaTablesApi` ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -88,6 +96,17 @@ class UCSingleCatalog
...
+    case provider: DeltaRestClientProvider => provider.getDeltaTablesApi()
+    case _ => null
+  }
+
+  // Always exposes the ApiClient this catalog built in initialize().
+  override def getApiClient(): Object = apiClient
```
openinx: I suggest not to use a abiratry Object as the return value in the common API,  which is pretty easy to be buggy.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -546,11 +565,25 @@ private class UCProxy(
...
-    temporaryCredentialsApi: TemporaryCredentialsApi) extends TableCatalog with SupportsNamespaces with Logging {
+    temporaryCredentialsApi: TemporaryCredentialsApi)
+  extends TableCatalog
+  with SupportsNamespaces
+  with DeltaRestClientProvider
+  with Logging {
   private[this] var name: String = null
   private[this] var schemasApi: SchemasApi = null
+  // Cached DRC TablesApi. `lazy val` means it's never constructed when the flag is off.
+  private[this] lazy val deltaTablesApi: DeltaTablesApi = new DeltaTablesApi(apiClient)
```
openinx: why not explicitly use the flag to determine if want to initialize the `DeltaTableApi` or not ?  this seems to be easily broken.

openinx: you already have a `getDeltaTablesApi`, right ?

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
timothyw553: OK, we can change it to ApiClient

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: updated to use ApiClient instead.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: updated. lazy field is now null when flag off, constructed once at ctor when on. 

getter is a trivial `def getDeltaTablesApi() = deltaTablesApi`

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: the method exists so that callers reaching this catalog through Spark's public catalog manager can share the HTTP client this catalog builds at init. without it, every caller has to construct its own duplicate client against the same server.

future consumers will will get to re-use the pre-built shared client.

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
```diff
@@ -0,0 +1,41 @@
...
+   * REST Catalog integration is disabled for this catalog (see the {@code deltaRestApi.enabled}
+   * option). Callers must null-check. The returned instance is owned by the catalog and must not be
+   * closed by the caller.
+   *
+   * <p>The current Delta Spark consumer builds its own {@link TablesApi} from {@link
+   * #getApiClient()} rather than reusing this cached instance; this method is intended for future
+   * consumers (commit coordinator, Kernel v2 connector) that want to share UC's cached client
+   * instead of constructing their own.
+   */
+  TablesApi getDeltaTablesApi();
```
openinx: If the `deltaRestApi.enabled = false`,   will it still use the same `DeltaRestClientProvider` ?  and seems like the api will just return a null value.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: Will the delta repo need to cast this `StagingTableCatalog` abstracted class into a concrete `DeltaRestClientProvider` , so that they can access those methods defined in `DeltaRestClientProvider` API ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -546,11 +565,25 @@ private class UCProxy(
...
+  extends TableCatalog
+  with SupportsNamespaces
+  with DeltaRestClientProvider
+  with Logging {
   private[this] var name: String = null
   private[this] var schemasApi: SchemasApi = null
+  // Null when DRC is disabled; constructed once at UCProxy ctor when enabled.
+  // The flag decides construction directly — callers must null-check.
+  private[this] val deltaTablesApi: DeltaTablesApi =
+    if (deltaRestApiEnabled) new DeltaTablesApi(apiClient) else null
```
openinx, line 579: If the `deltaRestApiEnabled=false` but someone called the `deltaTableApi`,  I think we'd better to explicitly throw the exception, and saying this is unsupported.

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
timothyw553: yes, the interface is always implemented (even when `deltaRestApi.enabled = false`) the flag only gates what `getDeltaTablesApi()` returns.

Flag on (`deltaRestApi.enabled = true`):
 * provider.getDeltaTablesApi() → non-null DRC TablesApi → use DRC endpoint.
 * provider.getApiClient()      → ApiClient (if you want to build other wrappers, this ApiClient will be shared).

Flag off (`deltaRestApi.enabled = false`):
 * provider.getDeltaTablesApi() → null → skip DRC.
 * provider.getApiClient()      → ApiClient → build legacy TablesApi from it → use legacy endpoint.

In both cases the provider is implemented and `getApiClient()` works — only the DRC client's presence flips.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: Zoom out,  if we have two variants of `UCProxy` instances ( with and without deltaRestApi support ). 

Then,  I will suggest to have two subclasses for the UCProxy. 
1.  UCProxy.  ( without `deltaRestApi` support ) 
2. DeltaRestUCProxy extends UCProxy ( with `deltaRestApi` support ). 

And we can use the `deltaRestApiEnabled` switch to determine which uc proxy we should initialize.  And we can also use this chance to move the scala `UCProxy` to Java classes.   (Let's gradually move the scala to Java ).

What do you think ?

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
timothyw553: alternative: use `Optional<TaplesApi>`?

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaRestClientProviderTest.java
```diff
@@ -0,0 +1,125 @@
...
+  public void flagEnabled_catalogOwnsCachedDeltaTablesApi() {
+    // Flag on: the catalog returns a single cached TablesApi. The assignment to
+    // DeltaRestClientProvider is a compile-time check that UCSingleCatalog
+    // implements the interface. The isSameAs assertion locks in the lazy-val
+    // caching invariant; if someone changes it to build per-call, this test fails.
+    UCSingleCatalog catalog = catalogWithFlag(true);
+    DeltaRestClientProvider provider = catalog;
+    TablesApi first = provider.getDeltaTablesApi();
+    assertThat(first).isNotNull();
+    assertThat(provider.getDeltaTablesApi()).isSameAs(first);
```
openinx: This seems meaningless,  the `provider.getDeltaTablesApi` should always be `first`, why do we still. need do the assertion ?

timothyw553: Dropped the `isSameAs` — you were right, tautological on a `val`. Renamed the test to `flagEnabled_returnsDeltaTablesApi` since it now just asserts the Optional is present.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: Yes — Delta-side code pattern-matches the `CatalogPlugin` Spark returns to `DeltaRestClientProvider`, then calls the interface methods:

```scala
// In Delta Spark code (delta-io/delta)
val catalog: CatalogPlugin = session.sessionState().catalogManager().catalog("unity")

catalog match {
  case provider: DeltaRestClientProvider =>
    val apiClient = provider.getApiClient()            // always non-null once initialized
    val drcOpt   = provider.getDeltaTablesApi()        // present iff DRC enabled

    // Build whatever wrappers Delta needs from the shared ApiClient,
    // e.g. new UCTokenBasedDeltaRestClient(apiClient, drcOpt.isPresent())

  case _ =>
    // Not a UC catalog — fall through
}
```

The cast is the bridge between Spark's `CatalogPlugin` type (what Spark hands out) and the DRC interface (what we need to pull the shared client).

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
timothyw553: Shipped `Optional<TablesApi>` in the latest push — interface, both impls, and tests updated.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: i'm actually pushing back on this (just for this PR)

only 1 of UCProxy's ~30 methods (getDeltaTablesApi) actually changes based on the DRC flag. the other 29 (staging, loadTable, listTables, credentials, etc.) behave the same way regardless -- seems like a lot create DeltaRestUCProxy right now for 1 method

for scala -> java: yes, i agree, but UCProxy is ~700 lines and we can migrate that later.

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
```diff
@@ -0,0 +1,46 @@
...
+ * -- instead of each creating their own HTTP client to the same UC server.
+ *
+ * <p>Implementations should return the same cached {@link TablesApi} instance on every call. The
+ * caller must not close or modify the returned client.
+ *
+ * <p>This interface lives in {@code unitycatalog-client} so that both the UC Spark connector (which
+ * implements it) and Delta Spark (which consumes it) can depend on it without a circular
+ * dependency.
+ */
+public interface DeltaRestClientProvider {
```
openinx, line 20: notice:  this API is a public user-facing API,  we will need to carefully design the API, since we will need to maintain the compability cross different untycatalog-client versions.

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
```diff
@@ -0,0 +1,46 @@
...
+   *     Optional#empty()} — indicating the caller should fall back to the legacy UC API via {@link
+   *     #getApiClient()}.
+   */
+  Optional<TablesApi> getDeltaTablesApi();
+
+  /**
+   * Returns the base {@link ApiClient} for legacy UC API access, or {@code null} when the
+   * implementation does not expose its client.
+   */
+  default ApiClient getApiClient() {
```
openinx: If this return value is also nullable,  then would it be better for us to return it as a `Optional<ApiClient>` ?

timothyw553: Even better — since every implementer of this interface needs an `ApiClient` to satisfy the contract (the DRC `TablesApi` is constructed from it), I made the method abstract and dropped the nullable default. Non-nullable now without needing an `Optional` wrapper.

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
```diff
@@ -0,0 +1,46 @@
...
+   */
+  Optional<TablesApi> getDeltaTablesApi();
+
+  /**
+   * Returns the base {@link ApiClient} for UC API access. Consumers use it to construct typed
+   * wrappers against the shared HTTP pool — e.g. legacy {@code TablesApi} and {@code
+   * DeltaCommitsApi} for the non-DRC commit path, or DRC {@code TablesApi} for Delta REST Catalog
+   * endpoints. Implementations must return a non-null client.
+   */
+  ApiClient getApiClient();
```
openinx, line 46: Discussed with @TimothyW553 offline.  for the coming versions, we will need to support both legacy API and new API. 

The new `io.unitycatalog.client.delta.api.TablesApi` actually includes all the legacy API  (DeltaCommits API, tableAPI etc) set that required to build the table user story. 


that's why we have to expose the api client here, to make sure all the legacy and new api can reuse the same api client ( with the same connection pools ... )

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: yes,  okay. I agreed this work should be blocked by the scala -> java migration.

openinx (approved):
give my approval, but @yili-db  and @tdas pls help to take a look since this is a user-facing API change.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -88,6 +96,17 @@ class UCSingleCatalog
 
   override def name(): String = delegate.name()
 
+  // External callers (commit coordinator, tests, anyone using Spark's catalog manager)
+  // only see UCSingleCatalog. Without this override they couldn't reach the DRC client
+  // that UCProxy holds, because UCProxy is package-private.
+  override def getDeltaTablesApi(): Optional[DeltaTablesApi] = delegate match {
```
yili-db: delegate, when UCSingleCatalog works with Delta, is the DeltaCatalog not the UCProxy. In this case, the UCProxy is the `delegate.delegate`

You need to make `ucproxy` a private member in `UCSingleCatalog` then directly refer to that.

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
timothyw553: +1 just updated the comment to explain this

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: updated routing the delegate path properly now

so now when delegate matches properly and `getDeltaTablesApi()` forwards to `ucProxy.getDeltaTablesApi()`. same idea for `getApiClient()`

yili-db (approved):
@tdas please take a look. Thanks!

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaRestClientProvider.java
```diff
@@ -0,0 +1,47 @@
+package io.unitycatalog.client.delta;
+
+import io.unitycatalog.client.ApiClient;
+import io.unitycatalog.client.delta.api.TablesApi;
+import java.util.Optional;
+
+/**
+ * Interface for catalog implementations that can provide a Delta REST Catalog {@link TablesApi}
+ * client. This enables a single shared client instance owned by the catalog (UCProxy) to be reused
+ * by all consumers -- AbstractDeltaCatalog, UCCommitCoordinatorClient, and the Kernel v2 connector
```
tdas, line 10: I dont we should put classes in other repos as references here. these can easily get stale.

PR #1493, opened 2026-04-17
[API] Add METRIC_VIEW table type and view_definition/view_dependencies

chenwang-databricks:
## Summary

Adds metric view support to the Unity Catalog OpenAPI specification. This is the first PR in a 3-PR series; the follow-ups add server-side persistence and the Spark connector implementation.

## Motivation

Metric views provide a powerful semantic layer on top of tables. Adding support to OSS UC unlocks the feature for the open-source Spark community and keeps wire compatibility with the Databricks UC API.

## Changes

`api/all.yaml`:

- Add `METRIC_VIEW` to the `TableType` enum.
- Add optional `view_definition` (string) and `view_dependencies` (`DependencyList`) fields to `TableInfo` and `CreateTable`.
- Drop `data_source_format` and `storage_location` from `CreateTable.required` so metric views (which have neither) can be created.

Generated artifacts regenerated to match:

- `api/Models/*.md` (OpenAPI markdown docs)
- `api/README.md`
- `ui/src/types/api/catalog.gen.ts` (UI typed client)

## Compatibility

The new fields on `TableInfo` and `CreateTable` are optional. Relaxing `CreateTable.required` is wire-backwards-compatible: existing clients that send `data_source_format` and `storage_location` continue to work unchanged.

The schema shape matches the corresponding Databricks UC API so the same client can target either server.

## Stack

This PR is the bottom of a 3-PR stack:

1. (this PR) [API] Add METRIC_VIEW table type and view_definition/view_dependencies
2. [Server] Add METRIC_VIEW persistence and CRUD support
3. [Spark Connector] Metric view support in UC Spark connector

## Test Plan

- `./build/sbt client/compile serverModels/compile server/compile` succeeds with the new schema.
- `./build/sbt apiDocs/generate` was run to regenerate `api/Models/*.md` and `api/README.md`.
- `npx openapi-typescript --enum api/all.yaml -o ui/src/types/api/catalog.gen.ts` would normally regenerate the TS file; the equivalent edits were applied manually here (offline environment).

**PR Checklist**

- [ ] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

PR #1510, opened 2026-04-21
[Server] Add METRIC_VIEW persistence and CRUD support

chenwang-databricks:
Implements server-side support for the METRIC_VIEW table type added in the previous API PR: https://github.com/unitycatalog/unitycatalog/pull/1493

Persistence layer:
- TableInfoDAO: add view_definition column (length 16777215); make data_source_format and storage_location null-safe to support metric views which have neither of them
- DependencyDAO (new): Hibernate entity for the uc_dependencies table with an idx_dependent (dependent_type, dependent_id) index to keep get_table latency bounded when joining dependencies
- DependencyRepository (new): create / get / delete on uc_dependencies, scoped to the caller's session
- Repositories + HibernateConfigurator: wire up the new repository and register DependencyDAO

Business logic:
- TableRepository.createTable: add a METRIC_VIEW branch that validates view_definition and view_dependencies are present, then persists the dependency records via DependencyRepository.
- TableRepository.getTable / listTables: attach view_dependencies on read for METRIC_VIEW rows via a new attachDependencies helper.
- TableRepository.deleteTable: cascade-delete dependency rows for METRIC_VIEW.

Tests:
- BaseMetricViewCRUDTest covers create/get/list/delete with both an asset source and a SQL source, plus negative cases for missing view_definition and missing view_dependencies.
- SdkMetricViewCRUDTest wires the base test to the SDK driver.

This is the second PR in a 3-PR series adding metric view support to Unity Catalog OSS. The Spark connector PR comes next.

**PR Checklist**

**Description of changes**

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/DependencyRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/DependencyRepository.java
@@ -1,0 +1,63 @@
+package io.unitycatalog.server.persist;
+
+import io.unitycatalog.server.persist.dao.DependencyDAO;
+import io.unitycatalog.server.persist.dao.DependencyDAO.DependentType;
+import java.util.List;
+import java.util.UUID;
+import org.hibernate.Session;
+import org.hibernate.query.Query;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Repository for managing view/metric-view dependency records in uc_dependencies. Methods accept a
+ * Session so they can participate in the caller's transaction.
+ */
+public class DependencyRepository {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DependencyRepository.class);
+
+  public void createDependencies(
+      Session session,
+      UUID dependentId,
+      DependentType dependentType,
+      List<DependencyDAO> dependencies) {
+    for (DependencyDAO dep : dependencies) {
+      dep.setDependentId(dependentId);
+      dep.setDependentType(dependentType);
+      session.persist(dep);
+    }
+    LOGGER.debug(
+        "Created {} dependencies for {}:{}", dependencies.size(), dependentType, dependentId);
+  }
+
+  public List<DependencyDAO> getDependencies(
+      Session session, UUID dependentId, DependentType dependentType) {
+    String hql =
+        "FROM DependencyDAO d WHERE d.dependentId = :dependentId"
+            + " AND d.dependentType = :dependentType";
+    Query<DependencyDAO> query = session.createQuery(hql, DependencyDAO.class);
+    query.setParameter("dependentId", dependentId);
+    query.setParameter("dependentType", dependentType);
+    return query.list();
+  }
+
+  public void deleteDependencies(Session session, UUID dependentId, DependentType dependentType) {
+    String hql =
+        "DELETE FROM DependencyDAO d WHERE d.dependentId = :dependentId"
+            + " AND d.dependentType = :dependentType";
+    Query<?> query = session.createQuery(hql);
+    query.setParameter("dependentId", dependentId);
+    query.setParameter("dependentType", dependentType);
+    int deleted = query.executeUpdate();
+    LOGGER.debug("Deleted {} dependencies for {}:{}", deleted, dependentType, dependentId);
+  }
+
+  public void updateDependencies(
+      Session session,
+      UUID dependentId,
+      DependentType dependentType,
+      List<DependencyDAO> dependencies) {
+    deleteDependencies(session, dependentId, dependentType);
+    createDependencies(session, dependentId, dependentType, dependencies);
+  }
+}
--- a/server/src/main/java/io/unitycatalog/server/persist/DependencyRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/DependencyRepository.java
@@ -51,13 +51,4 @@ public class DependencyRepository {
     int deleted = query.executeUpdate();
     LOGGER.debug("Deleted {} dependencies for {}:{}", deleted, dependentType, dependentId);
   }
-
-  public void updateDependencies(
-      Session session,
-      UUID dependentId,
-      DependentType dependentType,
-      List<DependencyDAO> dependencies) {
-    deleteDependencies(session, dependentId, dependentType);
-    createDependencies(session, dependentId, dependentType, dependencies);
-  }
 }
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -170,7 +170,8 @@ public class TableRepository {
           TableInfo tableInfo = tableInfoDAO.toTableInfo(true, catalogName, schemaName);
           RepositoryUtils.attachProperties(
               tableInfo, tableInfo.getTableId(), Constants.TABLE, session);
-          attachDependencies(tableInfo, tableInfoDAO, session);
+          RepositoryUtils.attachDependencies(
+              tableInfo, tableInfoDAO, session, repositories.getDependencyRepository());
           return tableInfo;
         },
         "Failed to get table",
@@ -409,6 +410,7 @@ public class TableRepository {
           }
           TableType tableType = Objects.requireNonNull(createTable.getTableType());
           String tableID;
+          UUID metricViewTableUUID = null;
           NormalizedURL storageLocation;
           if (tableType == TableType.EXTERNAL) {
             storageLocation = NormalizedURL.from(createTable.getStorageLocation());
@@ -429,29 +431,9 @@ public class TableRepository {
                     .commitStagingTable(session, callerId, storageLocation);
             tableID = stagingTableDAO.getId().toString();
           } else if (tableType == TableType.METRIC_VIEW) {
-            if (createTable.getViewDefinition() == null
-                || createTable.getViewDefinition().isEmpty()) {
-              throw new BaseException(
-                  ErrorCode.INVALID_ARGUMENT, "view_definition is required for metric view");
-            }
-            DependencyList viewDeps = createTable.getViewDependencies();
-            if (viewDeps == null
-                || viewDeps.getDependencies() == null
-                || viewDeps.getDependencies().isEmpty()) {
-              throw new BaseException(
-                  ErrorCode.INVALID_ARGUMENT, "view_dependencies is required for metric view");
-            }
             storageLocation = null;
-            UUID tableUUID = UUID.randomUUID();
-            tableID = tableUUID.toString();
-            DependencyDAO.DependentType dependentType = DependencyDAO.DependentType.TABLE;
-            List<DependencyDAO> depDAOs =
-                viewDeps.getDependencies().stream()
-                    .map(dep -> DependencyDAO.from(dep, tableUUID, dependentType))
-                    .collect(Collectors.toList());
-            repositories
-                .getDependencyRepository()
-                .createDependencies(session, tableUUID, dependentType, depDAOs);
+            metricViewTableUUID = validateMetricView(createTable);
+            tableID = metricViewTableUUID.toString();
           } else if (tableType == TableType.STREAMING_TABLE) {
             throw new BaseException(
                 ErrorCode.INVALID_ARGUMENT, "STREAMING TABLE creation is not supported yet.");
@@ -495,10 +477,39 @@ public class TableRepository {
           PropertyDAO.from(tableInfo.getProperties(), tableInfoDAO.getId(), Constants.TABLE)
               .forEach(session::persist);
           session.persist(tableInfoDAO);
+          if (tableType == TableType.METRIC_VIEW) {
+            DependencyDAO.DependentType dependentType = DependencyDAO.DependentType.TABLE;
+            UUID tableUUID = Objects.requireNonNull(metricViewTableUUID);
+            List<DependencyDAO> depDAOs =
+                createTable.getViewDependencies().getDependencies().stream()
+                    .map(dep -> DependencyDAO.from(dep, tableUUID, dependentType))
+                    .collect(Collectors.toList());
+            repositories
+                .getDependencyRepository()
+                .createDependencies(session, tableUUID, dependentType, depDAOs);
+          }
           return tableInfo;
         },
         "Error creating table: " + fullName,
         /* readOnly = */ false);
+  }
+
+  private static UUID validateMetricView(CreateTable createTable) {
+    if (createTable.getViewDefinition() == null || createTable.getViewDefinition().isEmpty()) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT, "view_definition is required for metric view");
+    }
+    DependencyList viewDeps = createTable.getViewDependencies();
+    if (viewDeps == null || viewDeps.getDependencies() == null) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT, "view_dependencies is required for metric view");
+    }
+    if (viewDeps.getDependencies().isEmpty()) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT,
+          "view_dependencies must contain at least one entry for metric view");
+    }
+    return UUID.randomUUID();
   }
 
   public TableInfoDAO findBySchemaIdAndName(Session session, UUID schemaId, String name) {
@@ -576,7 +587,8 @@ public class TableRepository {
         RepositoryUtils.attachProperties(
             tableInfo, tableInfo.getTableId(), Constants.TABLE, session);
       }
-      attachDependencies(tableInfo, tableInfoDAO, session);
+      RepositoryUtils.attachDependencies(
+          tableInfo, tableInfoDAO, session, repositories.getDependencyRepository());
       result.add(tableInfo);
     }
     return new ListTablesResponse().tables(result).nextPageToken(nextPageToken);
@@ -627,19 +639,5 @@ public class TableRepository {
     PropertyRepository.findProperties(session, tableInfoDAO.getId(), Constants.TABLE)
         .forEach(session::remove);
     session.remove(tableInfoDAO);
-  }
-
-  private void attachDependencies(TableInfo tableInfo, TableInfoDAO tableInfoDAO, Session session) {
-    if (TableType.METRIC_VIEW.getValue().equals(tableInfoDAO.getType())) {
-      List<DependencyDAO> deps =
-          repositories
-              .getDependencyRepository()
-              .getDependencies(session, tableInfoDAO.getId(), DependencyDAO.DependentType.TABLE);
-      // Always attach a dependency list (even when empty) for syntactic consistency: callers
-      // can tell "metric view with no recorded deps" apart from "field not set", and a
-      // hypothetical metric view that aggregates only constants would still round-trip cleanly.
-      tableInfo.setViewDependencies(
-          new DependencyList().dependencies(DependencyDAO.toDependencyList(deps)));
-    }
   }
 }
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -203,20 +203,6 @@ public class TableRepository {
                 "Table not found: " + catalog + "." + schema + "." + table);
           }
 
-          // Guard non-Delta entries (metric views, parquet/csv/etc. tables) before they reach
-          // `buildTableMetadata`. The downstream code calls `NormalizedURL.normalize(dao.getUrl())`
-          // (throws "Path cannot be null or empty" when the row has no storage location, e.g.
-          // metric views) and `DataSourceFormat.fromValue(dao.getDataSourceFormat())` (throws
-          // IllegalArgumentException on null), both of which surface as misleading errors to a
-          // Delta REST client that simply asked for a name that happens to live in the same
-          // schema. Sibling guard to `DeltaCommitRepository.validateTable`.
-          if (dao.getDataSourceFormat() == null
-              || !DataSourceFormat.DELTA.toString().equals(dao.getDataSourceFormat())) {
-            throw new BaseException(
-                ErrorCode.INVALID_ARGUMENT,
-                "Table is not a Delta table: " + catalog + "." + schema + "." + table);
-          }
-
           TableMetadata metadata = buildTableMetadata(session, dao, catalog, schema, table);
 
           LoadTableResponse response = new LoadTableResponse();
@@ -423,13 +409,16 @@ public class TableRepository {
                 ErrorCode.TABLE_ALREADY_EXISTS, "Table already exists: " + fullName);
           }
           TableType tableType = Objects.requireNonNull(createTable.getTableType());
-          String tableID;
-          UUID metricViewTableUUID = null;
+          // `tableUUID` is the table's primary key. The shape is uniform across the three
+          // creatable branches (external, managed, metric view); the only divergence is the
+          // source of the UUID (random for external/metric-view, staging-table id for managed).
+          // The string form is generated exactly once below at `tableInfo.tableId(...)`.
+          UUID tableUUID;
           NormalizedURL storageLocation;
           if (tableType == TableType.EXTERNAL) {
             storageLocation = NormalizedURL.from(createTable.getStorageLocation());
             ExternalLocationUtils.validateNotOverlapWithManagedStorage(session, storageLocation);
-            tableID = UUID.randomUUID().toString();
+            tableUUID = UUID.randomUUID();
           } else if (tableType == TableType.MANAGED) {
             storageLocation = NormalizedURL.from(createTable.getStorageLocation());
             serverProperties.checkManagedTableEnabled();
@@ -443,11 +432,11 @@ public class TableRepository {
                 repositories
                     .getStagingTableRepository()
                     .commitStagingTable(session, callerId, storageLocation);
-            tableID = stagingTableDAO.getId().toString();
+            tableUUID = stagingTableDAO.getId();
           } else if (tableType == TableType.METRIC_VIEW) {
             storageLocation = null;
-            metricViewTableUUID = validateMetricView(createTable);
-            tableID = metricViewTableUUID.toString();
+            validateMetricView(createTable);
+            tableUUID = UUID.randomUUID();
           } else if (tableType == TableType.STREAMING_TABLE) {
             throw new BaseException(
                 ErrorCode.INVALID_ARGUMENT, "STREAMING TABLE creation is not supported yet.");
@@ -476,7 +465,7 @@ public class TableRepository {
                   .updatedBy(callerId)
                   .storageLocation(storageLocation != null ? storageLocation.toString() : null)
                   .viewDefinition(createTable.getViewDefinition())
-                  .tableId(tableID);
+                  .tableId(tableUUID.toString());
 
           TableInfoDAO tableInfoDAO = TableInfoDAO.from(tableInfo, schemaId);
           // create columns
@@ -493,7 +482,6 @@ public class TableRepository {
           session.persist(tableInfoDAO);
           if (tableType == TableType.METRIC_VIEW) {
             DependencyDAO.DependentType dependentType = DependencyDAO.DependentType.TABLE;
-            UUID tableUUID = Objects.requireNonNull(metricViewTableUUID);
             List<DependencyDAO> depDAOs =
                 createTable.getViewDependencies().getDependencies().stream()
                     .map(dep -> DependencyDAO.from(dep, tableUUID, dependentType))
@@ -508,7 +496,7 @@ public class TableRepository {
         /* readOnly = */ false);
   }
 
-  private static UUID validateMetricView(CreateTable createTable) {
+  private static void validateMetricView(CreateTable createTable) {
     if (createTable.getViewDefinition() == null || createTable.getViewDefinition().isEmpty()) {
       throw new BaseException(
           ErrorCode.INVALID_ARGUMENT, "view_definition is required for metric view");
@@ -523,7 +511,6 @@ public class TableRepository {
           ErrorCode.INVALID_ARGUMENT,
           "view_dependencies must contain at least one entry for metric view");
     }
-    return UUID.randomUUID();
   }
 
   public TableInfoDAO findBySchemaIdAndName(Session session, UUID schemaId, String name) {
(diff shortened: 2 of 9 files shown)


server/src/main/java/io/unitycatalog/server/persist/dao/DependencyDAO.java
```diff
@@ -0,0 +1,107 @@
...
+  private String dependencyCatalog;
+
+  @Column(name = "dependency_schema")
+  private String dependencySchema;
+
+  @Column(name = "dependency_name")
+  private String dependencyName;
+
+  @Column(name = "dependency_target_id")
+  private UUID dependencyTargetId;
```
chenwang-databricks: This is not really needed anywhere, right? If so, I would rather we don't add it for now.

server/src/main/java/io/unitycatalog/server/persist/dao/DependencyDAO.java
```diff
@@ -0,0 +1,107 @@
...
+@ToString
+@Builder
+public class DependencyDAO {
+  @Id
+  @UuidGenerator
+  @Column(name = "id", updatable = false, nullable = false)
+  private UUID id;
+
+  @Column(name = "dependent_type", nullable = false)
+  private String dependentType;
```
chenwang-databricks: Is it possible to make the dependent type an enum? Currently, we only support "TABLE".

server/src/main/java/io/unitycatalog/server/persist/dao/DependencyDAO.java
```diff
@@ -0,0 +1,107 @@
...
+
+  @Column(name = "dependent_id", nullable = false)
+  private UUID dependentId;
+
+  @Column(name = "dependency_type", nullable = false)
+  private String dependencyType;
```
chenwang-databricks: Similarly, for dependency type, can we make it an enum and only support "TABLE/FUNCTION" for now based on the API model defined in api.yaml?

server/src/main/java/io/unitycatalog/server/persist/dao/DependencyDAO.java
```diff
@@ -0,0 +1,107 @@
...
+
+    if (dependency.getTable() != null) {
+      builder.dependencyType("TABLE");
+      String fullName = dependency.getTable().getTableFullName();
+      String[] parts = fullName.split("\\.");
+      if (parts.length == 3) {
+        builder.dependencyCatalog(parts[0]);
+        builder.dependencySchema(parts[1]);
+        builder.dependencyName(parts[2]);
+      } else {
```
chenwang-databricks: You should expect the dependency always to pass in the table full name, which is the three-part name. And if it's not, we should throw an exception here.

server/src/main/java/io/unitycatalog/server/persist/dao/DependencyDAO.java
```diff
@@ -0,0 +1,107 @@
...
+
+  @Column(name = "dependency_target_id")
+  private UUID dependencyTargetId;
+
+  /** Converts a Dependency API model to a DependencyDAO for a given dependent. */
+  public static DependencyDAO from(Dependency dependency, UUID dependentId, String dependentType) {
+    DependencyDAOBuilder builder =
+        DependencyDAO.builder().dependentId(dependentId).dependentType(dependentType);
+
+    if (dependency.getTable() != null) {
```
chenwang-databricks: The dependency can also contain a function in it, so we should handle that as well.

server/src/main/java/io/unitycatalog/server/persist/dao/DependencyDAO.java
```diff
@@ -0,0 +1,107 @@
...
+      }
+    }
+
+    return builder.build();
+  }
+
+  /** Converts this DAO to a Dependency API model. */
+  public Dependency toDependency() {
+    Dependency dependency = new Dependency();
+    if ("TABLE".equals(dependencyType)) {
```
chenwang-databricks: Similarly, you should handle function as well.

server/src/main/java/io/unitycatalog/server/persist/dao/DependencyDAO.java
```diff
@@ -0,0 +1,107 @@
...
+      String fullName = dependencyCatalog + "." + dependencySchema + "." + dependencyName;
+      dependency.setTable(new TableDependency().tableFullName(fullName));
+    }
+    return dependency;
+  }
+
+  public String getDependencyFullName() {
```
chenwang-databricks: Where is this used? If it is not used anywhere, we should remove it.

server/src/test/java/io/unitycatalog/server/base/table/BaseMetricViewCRUDTest.java
```diff
@@ -0,0 +1,172 @@
...
+                    && TableType.METRIC_VIEW.equals(t.getTableType()));
+
+    // --- Delete ---
+    tableOperations.deleteTable(METRIC_VIEW_FULL_NAME);
+    assertThatThrownBy(() -> tableOperations.getTable(METRIC_VIEW_FULL_NAME))
+        .isInstanceOf(Exception.class);
+  }
+
+  @Test
+  public void testMetricViewWithSqlSource() throws Exception {
```
chenwang-databricks: I don't think we need this. You can remove this.

server/src/main/java/io/unitycatalog/server/persist/dao/DependencyDAO.java
```diff
@@ -0,0 +1,150 @@
...
+
+  /** Converts a Dependency API model to a DependencyDAO for a given dependent. */
+  public static DependencyDAO from(
+      Dependency dependency, UUID dependentId, DependentType dependentType) {
+    DependencyDAOBuilder builder =
+        DependencyDAO.builder().dependentId(dependentId).dependentType(dependentType);
+
+    if (dependency.getTable() != null) {
+      builder.dependencyType(DependencyType.TABLE);
+      populateThreePartName(builder, dependency.getTable().getTableFullName(), "table");
```
chenwang-databricks: Please use the enum instead of literal string

server/src/main/java/io/unitycatalog/server/persist/dao/DependencyDAO.java
```diff
@@ -0,0 +1,150 @@
...
+    } else if (dependency.getFunction() != null) {
+      builder.dependencyType(DependencyType.FUNCTION);
+      populateThreePartName(builder, dependency.getFunction().getFunctionFullName(), "function");
+    } else {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT, "Dependency must have either a table or a function set");
```
chenwang-databricks: Just say unsupported dependency type in the error message

server/src/test/java/io/unitycatalog/server/base/table/BaseMetricViewCRUDTest.java
```diff
@@ -0,0 +1,135 @@
...
+import io.unitycatalog.client.model.TableDependency;
+import io.unitycatalog.client.model.TableInfo;
+import io.unitycatalog.client.model.TableType;
+import io.unitycatalog.server.utils.TestUtils;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import org.junit.jupiter.api.Test;
+
+public abstract class BaseMetricViewCRUDTest extends BaseTableCRUDTestEnv {
```
chenwang-databricks, line 27: Can you add a few more failing test cases, like dependency doesn't contain a full name, and so on and so forth based on the latest logic

server/src/main/java/io/unitycatalog/server/persist/dao/DependencyDAO.java
```diff
@@ -0,0 +1,153 @@
...
+        break;
+      case FUNCTION:
+        dependency.setFunction(new FunctionDependency().functionFullName(fullName));
+        break;
+    }
+    return dependency;
+  }
+
+  public static List<Dependency> toDependencyList(List<DependencyDAO> daos) {
+    if (daos == null) {
```
yili-db: This is redundant. It's never null

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -424,6 +428,32 @@ public TableInfo createTable(CreateTable createTable) {
...
+            if (viewDeps == null
+                || viewDeps.getDependencies() == null
+                || viewDeps.getDependencies().isEmpty()) {
+              throw new BaseException(
+                  ErrorCode.INVALID_ARGUMENT, "view_dependencies is required for metric view");
+            }
+            storageLocation = null;
+            tableID = UUID.randomUUID().toString();
+            if (viewDeps.getDependencies() != null) {
+              UUID tableUUID = UUID.fromString(tableID);
```
yili-db: To avoid a roundtrip, you can do this:
```
            UUID tableUUID = UUID.randomUUID();
            tableID = tableUUID.toString();
```

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -589,8 +621,26 @@ public void deleteTable(Session session, UUID schemaId, String tableName) {
           .getDeltaCommitRepository()
           .permanentlyDeleteTableCommits(session, tableInfoDAO.getId());
     }
+    if ("METRIC_VIEW".equals(tableInfoDAO.getType())) {
```
yili-db: Use enum `.getValue()`

server/src/test/java/io/unitycatalog/server/base/table/BaseMetricViewCRUDTest.java
```diff
@@ -0,0 +1,217 @@
...
+  @Test
+  public void testCreateMetricViewWithoutDefinitionFails() throws Exception {
+    CreateTable badRequest =
+        new CreateTable()
+            .name(METRIC_VIEW_NAME)
+            .catalogName(TestUtils.CATALOG_NAME)
+            .schemaName(TestUtils.SCHEMA_NAME)
+            .tableType(TableType.METRIC_VIEW)
+            .viewDependencies(makeDependencyList(SOURCE_TABLE_FULL_NAME));
+    assertThatThrownBy(() -> tableOperations.createTable(badRequest)).isInstanceOf(Exception.class);
```
yili-db: `isInstanceOf(Exception.class);` is not useful.
Please validate the actual errors. Also consider reusing helper functions `assertApiException`

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -424,6 +428,32 @@ public TableInfo createTable(CreateTable createTable) {
...
+            DependencyList viewDeps = createTable.getViewDependencies();
+            if (viewDeps == null
+                || viewDeps.getDependencies() == null
+                || viewDeps.getDependencies().isEmpty()) {
+              throw new BaseException(
+                  ErrorCode.INVALID_ARGUMENT, "view_dependencies is required for metric view");
+            }
+            storageLocation = null;
+            tableID = UUID.randomUUID().toString();
+            if (viewDeps.getDependencies() != null) {
```
yili-db: already checked a few lines above.

server/src/main/java/io/unitycatalog/server/persist/DependencyRepository.java
```diff
@@ -0,0 +1,63 @@
...
+        "DELETE FROM DependencyDAO d WHERE d.dependentId = :dependentId"
+            + " AND d.dependentType = :dependentType";
+    Query<?> query = session.createQuery(hql);
+    query.setParameter("dependentId", dependentId);
+    query.setParameter("dependentType", dependentType);
+    int deleted = query.executeUpdate();
+    LOGGER.debug("Deleted {} dependencies for {}:{}", deleted, dependentType, dependentId);
+  }
+
+  public void updateDependencies(
```
yili-db: Will this be useful? Do we support updating metric view?

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -589,8 +621,26 @@ public void deleteTable(Session session, UUID schemaId, String tableName) {
...
   }
+
+  private void attachDependencies(TableInfo tableInfo, TableInfoDAO tableInfoDAO, Session session) {
+    if ("METRIC_VIEW".equals(tableInfoDAO.getType())) {
+      List<DependencyDAO> deps =
+          repositories
+              .getDependencyRepository()
+              .getDependencies(session, tableInfoDAO.getId(), DependencyDAO.DependentType.TABLE);
+      if (!deps.isEmpty()) {
+        tableInfo.setViewDependencies(
```
yili-db: For a metric view, should there always be some deps? Even for an empty view, should tableInfo have an empty dependency list instead of null?

server/src/test/java/io/unitycatalog/server/base/table/BaseMetricViewCRUDTest.java
```diff
@@ -0,0 +1,217 @@
...
+          + "  - name: event_count\n"
+          + "    expr: count(*)";
+
+  protected static final Map<String, String> PROPERTIES =
+      Map.of("team", "analytics", "refresh", "daily");
+
+  private static DependencyList makeDependencyList(String... tableFullNames) {
+    DependencyList depList = new DependencyList();
+    depList.setDependencies(
+        java.util.Arrays.stream(tableFullNames)
```
yili-db: avoid fully qualified name

server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
```diff
@@ -76,6 +76,9 @@ public class TableInfoDAO extends IdentifiableDAO {
       fetch = FetchType.LAZY)
   private List<ColumnInfoDAO> columns;
 
+  @Column(name = "view_definition", length = 65535)
```
yili-db: Is 64KB enough?

yili-db (commented):
DependencyDAO.from and toDependency both have a FUNCTION branch, but only the non-three-part-name negative case exercises it. Add one happy-path test that creates a metric view with a function dependency and asserts round-trip through get/list, so the function branch has branch coverage.

server/src/main/java/io/unitycatalog/server/persist/DependencyRepository.java
chenwang-databricks: Updating metric view is not in scope of the initial release. I think we can remove this operation for now

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
chenwang-databricks: For a valid metric view, it should always have at least one table dependency

server/src/test/java/io/unitycatalog/server/base/table/BaseMetricViewCRUDTest.java
chenwang-databricks: OK will move to the import

server/src/test/java/io/unitycatalog/server/base/table/BaseMetricViewCRUDTest.java
```diff
@@ -0,0 +1,217 @@
...
+            .map(name -> new Dependency().table(new TableDependency().tableFullName(name)))
+            .collect(java.util.stream.Collectors.toList()));
```
chenwang-databricks: Move this to import as well

server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
chenwang-databricks: is 1MB fine? I don't want to blow up the storage while having a reasonable limit

yili-db: It is more of a product requirement question. If 64KB works for metric view, that's best as it is guaranteed to work anyway. Otherwise we'll do some research to see.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: Then you can skip this isEmpty. Even if it's empty, like a dummy metric view that only returns a constant number for example, this dependency list can be empty instead of null which I think is more syntax correct.

yili-db (approved):
@tdas please take a look

server/src/main/java/io/unitycatalog/server/persist/dao/DependencyDAO.java
```diff
@@ -0,0 +1,155 @@
...
+    final String fullName;
+    if (dependency.getTable() != null) {
+      dependencyType = DependencyType.TABLE;
+      fullName = dependency.getTable().getTableFullName();
+    } else if (dependency.getFunction() != null) {
+      dependencyType = DependencyType.FUNCTION;
+      fullName = dependency.getFunction().getFunctionFullName();
+    } else {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Unsupported dependency type");
+    }
```
tdas, line 103: [AI Assisted] The OpenAPI spec for `Dependency` says "Either the table field or the function field must be defined" (api/all.yaml line 2111-2112). Today this code silently picks `table` when both are populated. Suggest rejecting the both-set case explicitly:

```java
boolean hasTable = dependency.getTable() != null;
boolean hasFunction = dependency.getFunction() != null;
if (hasTable == hasFunction) {
  throw new BaseException(ErrorCode.INVALID_ARGUMENT,
      "Dependency must have exactly one of table or function set");
}
```

The negative test set covers "neither" but not "both" — worth adding a test for the both-set case while you're here.

tdas (commented):
[AI Assisted] Reviewed against the cloned repo at PR head `105098b`. Skipped duplicates of existing threads from @chenwang-databricks and @yili-db (enums for dependent/dependency type, three-part-name validation, function-dep handling, removing `updateDependencies`, `assertApiException` helper, `tableUUID` reuse, `length=1MB` debate, FQ-name imports). Two extra items I couldn't anchor to a diff line, plus inline comments below:

**A. PR description out-of-sync with code.** Description says "TableInfoDAO: add view_definition column (length 65535)" but after the latest commit the column is `length = 16777215` (MEDIUMTEXT). Please update the description to match.

**B. `loadTableForDelta` will throw a misleading error on METRIC_VIEW.** This file isn't in the diff so commenting here. The Delta REST endpoint at `DeltaRestCatalogService.loadTable` is not gated by table type; it dispatches straight to `tableRepository.loadTableForDelta`. After this PR `dao.getDataSourceFormat()` and `dao.getUrl()` can be null for METRIC_VIEW. Inside `buildTableMetadata`, `NormalizedURL.normalize(dao.getUrl())` then throws `BaseException(INVALID_ARGUMENT, "Path cannot be null or empty")`. A Delta REST client that asks by name for a metric view will get that confusing message instead of a clear "this is not a Delta table". Suggest adding an explicit format/type guard at the top of `loadTableForDelta` (similar to `DeltaCommitRepository.validateTable`).

Verified items withdrawn: (1) UC uses Hibernate `hbm2ddl.auto=update` so no manual migration needed; (2) `attachDependencies` per-row queries are consistent with the existing `RepositoryUtils.attachProperties` pattern, no worse than the status quo.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -424,6 +428,30 @@ public TableInfo createTable(CreateTable createTable) {
                     .getStagingTableRepository()
                     .commitStagingTable(session, callerId, storageLocation);
             tableID = stagingTableDAO.getId().toString();
+          } else if (tableType == TableType.METRIC_VIEW) {
```
tdas, line 450: [AI Assisted] This `else if (tableType == TableType.METRIC_VIEW)` branch is now ~22 lines compared with 2–10 in the other branches. Per the readability guideline of preferring named helpers over deep inline blocks, please extract this into a `validateAndPrepareMetricView(...)` private method that does the validation + dep-list building and returns the `(tableID, depDAOs)` pair. That keeps `createTable` symmetric across table types and makes future additions (MATERIALIZED_VIEW, STREAMING_TABLE) cheap.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -424,6 +428,30 @@ public TableInfo createTable(CreateTable createTable) {
...
+                || createTable.getViewDefinition().isEmpty()) {
+              throw new BaseException(
+                  ErrorCode.INVALID_ARGUMENT, "view_definition is required for metric view");
+            }
+            DependencyList viewDeps = createTable.getViewDependencies();
+            if (viewDeps == null
+                || viewDeps.getDependencies() == null
+                || viewDeps.getDependencies().isEmpty()) {
+              throw new BaseException(
+                  ErrorCode.INVALID_ARGUMENT, "view_dependencies is required for metric view");
```
tdas: [AI Assisted] This same error message is used for both the missing case (`viewDeps == null` / `getDependencies() == null`) and the empty case (`getDependencies().isEmpty()`). A user who sent `view_dependencies: { dependencies: [] }` will be told "view_dependencies is required for metric view" even though they did send the field. If the empty case is intentionally rejected (still in discussion in @yili-db's thread), please give it a distinct message like "view_dependencies must contain at least one entry for metric view".

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -424,6 +428,30 @@ public TableInfo createTable(CreateTable createTable) {
...
+            UUID tableUUID = UUID.randomUUID();
+            tableID = tableUUID.toString();
+            DependencyDAO.DependentType dependentType = DependencyDAO.DependentType.TABLE;
+            List<DependencyDAO> depDAOs =
+                viewDeps.getDependencies().stream()
+                    .map(dep -> DependencyDAO.from(dep, tableUUID, dependentType))
+                    .collect(Collectors.toList());
+            repositories
+                .getDependencyRepository()
+                .createDependencies(session, tableUUID, dependentType, depDAOs);
```
tdas: [AI Assisted] Order of operations: `createDependencies` is called here, but the parent `session.persist(tableInfoDAO)` happens ~40 lines later. Both run in the same transaction so rollback is correct today, but the code reads "insert children, then insert parent" which is fragile to future refactors (e.g. an early `flush()` or splitting transactions would leave FK-orphan rows). Suggest moving this `createDependencies` call down next to the column/property persistence after `session.persist(tableInfoDAO)`. As a side effect the METRIC_VIEW `if`-branch shrinks to just validation + tableID assignment, which also helps the extract-helper suggestion above.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -589,8 +619,27 @@ public void deleteTable(Session session, UUID schemaId, String tableName) {
...
+      repositories
+          .getDependencyRepository()
+          .deleteDependencies(session, tableInfoDAO.getId(), DependencyDAO.DependentType.TABLE);
+    }
     PropertyRepository.findProperties(session, tableInfoDAO.getId(), Constants.TABLE)
         .forEach(session::remove);
     session.remove(tableInfoDAO);
   }
+
+  private void attachDependencies(TableInfo tableInfo, TableInfoDAO tableInfoDAO, Session session) {
```
tdas: [AI Assisted] Naming/structure consistency nit: the codebase's convention for this kind of thing is the static `RepositoryUtils.attachProperties(...)` helper used by `Schema/Function/Catalog/TableRepository`. `attachDependencies` is instead a private instance method on `TableRepository` because it needs `repositories.getDependencyRepository()`. Either:
- (a) move it to `RepositoryUtils.attachDependencies` and pass the `DependencyRepository` as a parameter, or
- (b) make it `DependencyRepository.attachToTableInfo(TableInfo, UUID, Session)`.

Either preserves the existing `attachX` convention. Minor nit, follow-up is fine.

server/src/test/java/io/unitycatalog/server/base/table/BaseMetricViewCRUDTest.java
```diff
@@ -0,0 +1,278 @@
...
+        .schemaName(TestUtils.SCHEMA_NAME)
+        .tableType(TableType.METRIC_VIEW)
+        .viewDefinition(VIEW_DEFINITION_ASSET_SOURCE)
+        .viewDependencies(makeDependencyList(SOURCE_TABLE_FULL_NAME));
+  }
+
+  @Test
+  public void testMetricViewCRUD() throws Exception {
+    assertThatThrownBy(() -> tableOperations.getTable(METRIC_VIEW_FULL_NAME))
+        .isInstanceOf(Exception.class);
```
tdas: [AI Assisted] @yili-db already flagged `isInstanceOf(Exception.class)` on the negative tests; the same anti-pattern is here on the pre-create check (and at line 117–118 after delete). A test that asserts "some Exception" passes even on an unrelated NPE from a future refactor. Please switch to `assertApiException(..., ErrorCode.TABLE_NOT_FOUND, ...)` so we know the *right* failure happened.

server/src/test/java/io/unitycatalog/server/base/table/BaseMetricViewCRUDTest.java
```diff
@@ -0,0 +1,278 @@
...
+        .as("Metric view should appear in listTables")
+        .anyMatch(
+            t ->
+                METRIC_VIEW_NAME.equals(t.getName())
+                    && TableType.METRIC_VIEW.equals(t.getTableType()));
+
+    // --- Delete ---
+    tableOperations.deleteTable(METRIC_VIEW_FULL_NAME);
+    assertThatThrownBy(() -> tableOperations.getTable(METRIC_VIEW_FULL_NAME))
+        .isInstanceOf(Exception.class);
```
tdas: [AI Assisted] Same as the pre-create assertion at line 70 — please assert `ErrorCode.TABLE_NOT_FOUND` via `assertApiException` here too, so this assertion still has teeth if a regression makes `getTable` throw something else (e.g. NPE from null url/format) instead of a clean NOT_FOUND.

server/src/test/java/io/unitycatalog/server/base/table/BaseMetricViewCRUDTest.java
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  @Test
+  public void testCreateMetricViewWithoutDefinitionFails() {
```
tdas: [AI Assisted] These seven `testCreateMetricView*Fails` tests are all the same shape: build a `CreateTable`, mutate one field, call `assertApiException(..., expectedCode, expectedMessage)`. Per the test-consolidation guideline ("avoid the cost of generating the client/info/engine objects" per case), consider one `@ParameterizedTest` driven by a `(label, mutator, expectedCode, expectedMessage)` list. Drops 7 fixture spin-ups to 1 and gives a single obvious place to add new validation cases.

tdas (commented):
[AI Assisted] Reviewed against the cloned repo at PR head `418501e`. Thanks for the thorough pass on the round-1 feedback — everything I flagged inline plus the PR-description sync is addressed cleanly:

- `DependencyDAO.from` now rejects the both-set case explicitly and a corresponding parameterized test case was added.
- `validateMetricView` extracted as a static helper; the METRIC_VIEW branch in `createTable` is now ~4 lines.
- Dependencies are persisted *after* `session.persist(tableInfoDAO)`, with `metricViewTableUUID` carrying state across the two phases.
- `attachDependencies` lives in `RepositoryUtils` now, matching `attachProperties`.
- Both `isInstanceOf(Exception.class)` calls in `BaseMetricViewCRUDTest` switched to `assertApiException(..., TABLE_NOT_FOUND, ...)`.
- Three distinct messages for null / empty-list / unsupported-entry view_dependencies.
- The seven `*Fails` tests are folded into one `@ParameterizedTest` with a `negativeCreateCases()` source — nicely structured.

One item from my round-1 body comment is still open:

**`loadTableForDelta` regression for METRIC_VIEW** — the Delta REST endpoint at `DeltaRestCatalogService.loadTable` still dispatches unconditionally into `tableRepository.loadTableForDelta`, which calls `NormalizedURL.normalize(dao.getUrl())` (throws `INVALID_ARGUMENT, "Path cannot be null or empty"` for metric views) and `toDeltaFormat(dao.getDataSourceFormat())` (`DataSourceFormat.fromValue(null)` throws `IllegalArgumentException`). Net effect: a Delta REST client that asks by name for a metric view in the same schema will get a misleading error instead of a clear "not a Delta table". I'd suggest a guard at the top of `loadTableForDelta` (sibling to the existing one in `DeltaCommitRepository.validateTable`):

```java
if (dao.getDataSourceFormat() == null
    || !DataSourceFormat.DELTA.toString().equals(dao.getDataSourceFormat())) {
  throw new BaseException(
      ErrorCode.INVALID_ARGUMENT,
      "Table is not a Delta table: " + catalog + "." + schema + "." + table);
}
```

If you'd rather defer this to a follow-up PR (since `DeltaRestCatalogService` is technically not in this diff), totally fine — just please file an issue tracking it so it doesn't get lost when a future user hits it.

**Tiny description nit (non-blocking):** the description still says `DependencyRepository (new): create / get / delete / update`, but the `update` operation was removed per @yili-db's earlier feedback. Drop the `/ update`.

Leaving this as `COMMENTED` rather than approving since I'm not a UC server-persist owner. From my lens the only outstanding item is the `loadTableForDelta` guard above; everything else looks great.

tdas (approved):
almost LGTM

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -245,6 +203,20 @@ public LoadTableResponse loadTableForDelta(String catalog, String schema, String
                 "Table not found: " + catalog + "." + schema + "." + table);
           }
 
+          // Guard non-Delta entries (metric views, parquet/csv/etc. tables) before they reach
```
yili-db: This could be a problem. We are considering limited support to allow non-Delta tables to be loaded.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -471,6 +444,10 @@ public TableInfo createTable(CreateTable createTable) {
                     .getStagingTableRepository()
                     .commitStagingTable(session, callerId, storageLocation);
             tableID = stagingTableDAO.getId().toString();
+          } else if (tableType == TableType.METRIC_VIEW) {
+            storageLocation = null;
+            metricViewTableUUID = validateMetricView(createTable);
```
yili-db: Why have another variable?
Also validateMetricView returning a random UUID is confusing. By reading this I thought the ID is some existing ID.

yili-db (approved):

PR #1513, opened 2026-04-22
[Connectors][Spark] Add metric-view support to the UC Spark connector

chenwang-databricks:
**PR Checklist**

**Description of changes**

This PR adds metric-view support to the UC Spark connector, on top of the server-side
metric-view persistence shipped in #1510. With these changes:

- Spark users can `CREATE VIEW <ident> WITH METRICS ... LANGUAGE YAML AS $$ ... $$`
  against a Unity Catalog catalog, and the connector routes the create through UC's
  `tablesApi.createTable` with `tableType = METRIC_VIEW` plus the structured
  `view_dependencies` payload.
- `SELECT * FROM <metric_view>` resolves the row through `UCProxy.loadTable` (the
  single-RPC primitive), returns a `MetadataTable(ViewInfo)`, and Spark's analyzer
  expands the YAML body into the underlying aggregation over the source table.
- `DROP VIEW`, `SHOW VIEWS`, `DESCRIBE TABLE` / `EXTENDED` all work end-to-end against
  metric views. `DROP TABLE` on a metric view raises `WRONG_COMMAND_FOR_OBJECT_TYPE`
  ("Use DROP VIEW instead").

### 1. `UCSingleCatalog` mixes in `TableViewCatalog`

Spark's [SPARK-52729](https://github.com/apache/spark/pull/51419) introduced
`ViewCatalog` for v2 catalogs that expose views, and a sibling interface for catalogs
that expose **both** tables and views in a shared identifier namespace; the latter was
renamed to `TableViewCatalog` (with companion renames `MetadataOnlyTable` ->
`MetadataTable`, `loadRelation` -> `loadTableOrView`) and grew an abstract
`renameView(oldIdent, newIdent)` in
[SPARK-56655](https://github.com/apache/spark/commit/384cd0d3e83). UC exposes both
tables and views, so `UCSingleCatalog` and `UCProxy` mix in `TableViewCatalog` and
implement the view-side methods (`createView`, `dropView`, `renameView`, `listViews`,
`loadTableOrView`).

The orthogonality contract (`TableCatalog` methods behave as if views don't exist, and
vice versa) is honored:

- `listTables` filters out view-like UC `TableType`s via the shared `viewLikeUcTypes` map (see below).
- `listViews` filters in the same set.
- `loadTable` rejects view-like rows by throwing `NoSuchTableException`; `loadView`
  throws `NoSuchViewException` for table rows. (See the section "Single-RPC primitive"
  below for *which* layer enforces this -- it shifted between `UCProxy` and
  `UCSingleCatalog` in the latest revision.)
- `dropTable` returns `false` for a view (does not delete); `dropView` returns `false`
  for a table. SPARK-56655's `DropTableExec` / `DropViewExec` decode the false-return
  into the user-facing `WRONG_COMMAND_FOR_OBJECT_TYPE` error.
- `createTable` over a view ident maps UC 409 to `TableAlreadyExistsException`;
  `createView` over a table ident maps to `ViewAlreadyExistsException`.
- `renameView` is wired through both `UCSingleCatalog` and `UCProxy` but currently
  throws `UnsupportedOperationException` -- UC OSS has no rename-view endpoint,
  mirroring the existing `renameTable` stub.

### 2. View-like UC table type -> Spark table type mapping

`UCSingleCatalog.viewLikeUcTypes: Map[TableType, Option[String]]` is the single
source of truth for which UC `TableType` values are view-like, and what Spark
`TableSummary` table-type string they map to. Today: `METRIC_VIEW` -> `Some("METRIC_VIEW")`,
`MATERIALIZED_VIEW` -> `None` (no Spark mapping yet). Predicates derived from the map
(`isViewLikeTableType`, `isViewCommandsSupportedTableType`, `ucTableTypeToSparkViewType`,
`sparkViewTypeToUcTableType`) all read from this map, so adding a new view-like UC type
in the future is a one-line change. `createView` uses the reverse lookup
`sparkViewTypeToUcTableType` to derive `CreateTable.tableType` from the supplied
`ViewInfo.properties().get(PROP_TABLE_TYPE)`, so it works for any view kind in the map
without a touch on the method.

### 3. `UCSingleCatalog` keeps the Delta delegate; `UCProxy` is the v2 entry point

`UCSingleCatalog` continues to wrap `DeltaCatalog` (when on the classpath) as its
`delegate` for table operations -- DeltaCatalog needs to intercept managed/external Delta
table create/load paths and add Delta-specific behavior. View operations bypass
`delegate` and call `ucProxy` directly because Delta has no role in view handling. The
two references are documented inline at the field declarations.

### 4. Single-RPC primitive: `UCProxy.loadTable` (review feedback from 2026-05-08)

This is a structural change vs. the first revision of this PR, addressing the main
concern raised in the review meeting on 2026-05-08:

> Previously, every `loadTableOrView` call to a non-view UC table issues **two**
> `tablesApi.getTable` RPCs -- one in `UCSingleCatalog.loadTableOrView` to classify
> view vs. table, and a second through `delegate.loadTable -> Delta -> super.loadTable
> -> UCProxy.loadTableOrView`. With the metric-view PR landed, every read in the
> catalog flows through `loadTableOrView`, so this doubles RPC traffic to the UC server
> for every table load.

Fix ("flip the logic"):

- **`UCProxy.loadTable` is now the single primitive.** It does the UC `getTable` once
  and dispatches by `TableType`: a view-like row is wrapped as
  `MetadataTable + ViewInfo`; everything else goes through `loadV1Table` (which
  generates temp credentials).
- **`UCProxy.loadTableOrView` is now a one-line passthrough to `loadTable`.**
- **`UCSingleCatalog.loadTableOrView` is now a one-line `delegate.loadTable(ident)`
  call.** When `delegate` is Delta wrapping `UCProxy`, the call chain is
  `Delta.loadTable -> super.loadTable (DelegatingCatalogExtension) -> UCProxy.loadTable`,
  and Delta's `case o => o` fallthrough returns the `MetadataTable` unchanged for views.
  **One RPC for everything** -- table loads, view loads, view-vs-table classification.
- **`UCSingleCatalog.loadTable` re-checks the result** and throws `NoSuchTableException`
  if `delegate.loadTable` came back with a `MetadataTable + ViewInfo`. This preserves
  the `TableViewCatalog.loadTable` orthogonality contract -- callers that explicitly
  want only tables (e.g. write paths) shouldn't accidentally get a view back from the
  now-permissive `UCProxy.loadTable`.

Updated data flow:

```
CREATE VIEW ... WITH METRICS:
  Spark CreateMetricViewCommand
    -> DataSourceV2Strategy intercept (non-session catalog)
    -> CreateV2MetricViewExec (extends V2CreateViewPreparation)
    -> UCSingleCatalog.createView -> ucProxy.createView
    -> UCProxy.createView builds CreateTable(METRIC_VIEW) with:
       - tableType derived from ViewInfo.properties().get(PROP_TABLE_TYPE)
       - viewDefinition = original YAML body
       - viewDependencies = converted from Spark's structural DependencyList
       - columns: each column's `typeJson` is built by UCColumnJson (preserves
         metric_view.type / metric_view.expr column metadata)
    -> tablesApi.createTable

SELECT * FROM <table_or_view_ident> (single-RPC path):
  Spark resolver
    -> UCSingleCatalog.loadTableOrView(ident)
    -> delegate.loadTable(ident)            (delegate = DeltaCatalog)
    -> Delta.loadTable -> super.loadTable   (DelegatingCatalogExtension)
    -> UCProxy.loadTable(ident)             (the single primitive)
       -> tablesApi.getTable                (ONE UC RPC)
       -> if view-like (per viewLikeUcTypes):
            new MetadataTable(toViewInfo(t), ident.toString)
          else:
            loadV1Table(ident, t)
    -> Delta.loadTable `case o => o` returns MetadataTable unchanged
    -> Spark analyzer expands MetadataTable+ViewInfo via V1Table.toCatalogTable
       -> CatalogTableType.METRIC_VIEW -> ResolveMetricView (existing V1 rule)

```

`loadView`, `tableExists`, and `viewExists` on `UCProxy` are intentionally **not**
overridden -- they're inherited from `TableViewCatalog`'s defaults, which discriminate
on the `MetadataTable + ViewInfo` shape against the new `loadTable` automatically. This
keeps the single-RPC perf entry point as the only place we need to maintain.

### 5. Structural dependency conversion

UC's wire format keeps the legacy dot-joined `tableFullName: String`. Spark's
`TableDependency` is structural (`String[] nameParts`, with explicit value-semantics
overrides on the record so two TableDependencies with the same parts compare equal --
auto-generated record `equals`/`hashCode` on an array field would otherwise fall through
to reference equality). The structural form preserves arity for multi-level-namespace
catalogs and is unambiguous against quoted identifiers containing literal `.`. v1 sources
resolved through the session catalog are normalized by Spark's
`MetricViewHelper.qualifyV1` to a stable 3-part `[spark_catalog, db, table]` shape before
they reach the connector, so the dot-joined wire payload is consistently fully qualified.
The connector translates at the boundary:

- `UCProxy.toUcDependencyList`: dot-joins Spark's `nameParts` to UC's `tableFullName` via
  Scala's native `Array.mkString` (Java arrays implicit-convert to `ArrayOps`).
- `UCProxy.fromUcDependencyList`: splits UC's `tableFullName` back into structural parts
  via the `Dependency.table(String[])` factory (non-vararg).

`Dependency` is a sealed interface that also permits `FunctionDependency` plus a
`Dependency.function(String[])` factory; per Spark's class-level Javadoc those are
groundwork for future producers (e.g. SQL UDF dep tracking). Today the connector
unconditionally drops anything that is not a `TableDependency` because UC OSS does not
persist function dependencies.

The dot-flatten boundary is lossy for identifiers containing literal `.`, but UC's
server-side dependency tracking has the same limitation today, so no fidelity is lost vs.
the pre-PR producer that emitted dot-joined strings directly.

### 6. Column metadata round-trip via `UCColumnJson`

`UCProxy.buildColumnInfos` calls `UCColumnJson.buildStructFieldJson`
helper on each `Column`, producing the canonical `StructField`-shape JSON
(`{name, type, nullable, metadata}`) that Unity Catalog stores in `ColumnInfo.typeJson`.
The shape is byte-compatible with what `StructField.toJson` emits and with what the
Databricks Runtime writes today, so the column-level metadata keys the analyzer attaches
to metric-view dimensions and measures (`metric_view.type`, `metric_view.expr`) survive
the round trip through Unity Catalog and are visible to every UC client.

On the read side, `UCProxy.toViewInfo` calls
`UCColumnJson.parseStructFieldJson(col.getTypeJson)` to reconstruct each Spark `Column`
-- the `comment` is lifted out into `Column.comment()`, non-comment metadata stays in
`metadataInJSON`, and the `dataType` (including complex nested types) is parsed via
`DataType.fromJson` internally.

`UCColumnJson` uses **only public Spark APIs** (`Column.{name, dataType, nullable,
comment, metadataInJSON}`, `DataType.{json, fromJson}`, `Column.create` 5-arg factory)
plus json4s for the StructField-shape envelope. 

### 7. `UCTableProperties` reserved-keys list

`V2_TABLE_PROPERTIES` (the set of Spark V2 properties stripped before forwarding to UC,
because UC promotes them to first-class fields on `CreateTable`) gained two entries:
`PROP_COLLATION` (default-collation passthrough) and `PROP_TABLE_TYPE` (the
`METRIC_VIEW` discriminator that Spark sets on the `ViewInfo` properties bag and that
UC sees via the dedicated `tableType` field on `CreateTable`).

## Tests

`UCProxySuite` (Java) exercises the connector end-to-end with a mocked `TablesApi`:

- **Listing semantics:** pagination, METRIC_VIEW filtered out of `listTables`,
  METRIC_VIEW filtered into `listViews`.
- **CREATE VIEW WITH METRICS payload:** asserts `CreateTable.tableType = METRIC_VIEW`,
  `viewDefinition`, columns include the `UCColumnJson`-built `typeJson` with
  `metric_view.type` / `metric_view.expr` keys, and `view_dependencies` round-trips
  structurally (Spark structural -> UC wire dot-joined -> Spark structural).
- **`loadTableOrView` -> `MetadataTable(ViewInfo)`:** for a UC METRIC_VIEW row,
  asserts the wrapper's `name()` is the v2 `ident.toString()` (per Spark spec).
- **`UCProxy.loadTable` is permissive for view-like rows:**
  `testLoadTableReturnsMetadataTableForMetricView` asserts that a UC METRIC_VIEW row
  comes back as `MetadataTable + ViewInfo` (the new contract; the
  `NoSuchTableException` rejection moved up to `UCSingleCatalog.loadTable`).
- **Single-RPC regression guards:**
  `testLoadTableTriggersExactlyOneGetTableRpc` and
  `testLoadTableOrViewTriggersExactlyOneGetTableRpc` use Mockito
  `verify(mockTablesApi, times(1)).getTable(...)` to pin RPC count at exactly one per
  identifier resolution, both for `loadTable` (table path) and `loadTableOrView` (view
  path). Locks in the fix for the double-RPC bug raised in the 2026-05-08 review.
- **Cross-type filtering (TableViewCatalog contract):** `loadView` throws
  `NoSuchViewException` for a regular table; `dropTable` returns `false` and does not
  call `deleteTable` for a metric view; `dropView` returns `false` for a regular table.
- **Cross-type collision (active rejection):** `createView` maps UC 409 to
  `ViewAlreadyExistsException`; `createTable` maps UC 409 to
  `TableAlreadyExistsException`.

### Manual end-to-end verification (single-RPC fix)

In addition to the unit tests, the single-RPC change was exercised end-to-end against
a local Spark 4.2 + Delta master + UC OSS server (with this branch's UC connector +
`unitycatalog-client-0.5.0-SNAPSHOT` + `unitycatalog-hadoop-0.5.0-SNAPSHOT` on the
classpath):

- `SELECT * FROM unity.default.<delta_table>` (Delta path) -- succeeds, returns rows.
- `CREATE VIEW unity.default.mv1 WITH METRICS LANGUAGE YAML AS $$ ... $$` -- succeeds,
  UC server has the row with `tableType = METRIC_VIEW` and the YAML body.
- `SELECT * FROM unity.default.mv1` -- the connector resolves the view
  (`UCSingleCatalog.loadTableOrView` -> `delegate.loadTable` -> Delta ->
  `UCProxy.loadTable` -> `MetadataTable + ViewInfo`) without raising
  `NoSuchTableException`, proving the flipped path works on the connector side. The
  query then trips an upstream Spark planner-rule bug (`AssertionError: No plan for
  MetricViewPlaceholder`) that is **not** related to this PR -- it's a separate
  apache/spark issue in the metric-view planner expansion for V2 catalogs and is being
  tracked upstream.

## Open follow-ups (tracked separately)

- Spark-version branching in the UC OSS test infrastructure so the metric-view tests
  only run against Spark 4.2+.


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -100,6 +107,39 @@ class UCSingleCatalog
...
+  override def createView(ident: Identifier, info: ViewInfo): ViewInfo =
+    ucProxy.createView(ident, info)
+
+  override def replaceView(ident: Identifier, info: ViewInfo): ViewInfo =
+    throw new UnsupportedOperationException("Replacing a view is not supported yet")
+
+  override def dropView(ident: Identifier): Boolean =
+    ucProxy.dropView(ident)
+
+  override def loadRelation(ident: Identifier): Table = {
```
chenwang-databricks: In this new architecture, what are the differences between Load View, Load Table, and Load Relation? Is it possible that we are branching here to check the type of the identifier? If it is a table, then we load table; if it is a view, then we load view. 

I would also use it to help me to check the implementation on the Spark side. When a metric view is loaded, does the Spark code call load view directly or does it go through the load relation function here?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -564,21 +607,37 @@ private class UCProxy(
 
   override def listTables(namespace: Array[String]): Array[Identifier] = {
     UCSingleCatalog.checkUnsupportedNestedNamespace(namespace)
+    val tables = listUcTables(namespace)
+    tables
+      .filterNot(_.getTableType == TableType.METRIC_VIEW)
```
chenwang-databricks: Are we only filtering out metric view here? How about views? Materialized views?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -564,21 +607,37 @@ private class UCProxy(
...
+      .map(table => Identifier.of(namespace, table.getName))
+      .toArray
+  }
+
+  override def listViews(namespace: Array[String]): Array[Identifier] = {
+    UCSingleCatalog.checkUnsupportedNestedNamespace(namespace)
+    val tables = listUcTables(namespace)
+    tables
+      .filter(_.getTableType == TableType.METRIC_VIEW)
```
chenwang-databricks: Similarly here, should we list all the view-like types here?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -564,21 +607,37 @@ private class UCProxy(
...
+      .map(table => Identifier.of(namespace, table.getName))
+      .toArray
+  }
 
+  private def listUcTables(namespace: Array[String]): Seq[io.unitycatalog.client.model.TableInfo] = {
```
chenwang-databricks: Move the import to the header of the file and only use the short name.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -564,21 +607,37 @@ private class UCProxy(
...
     val catalogName = this.name
     val schemaName = namespace.head
-    val tables = ArrayBuffer.empty[Identifier]
+    val tables = ArrayBuffer.empty[io.unitycatalog.client.model.TableInfo]
```
chenwang-databricks: Move the import to the header of the file and only use the short name.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -564,21 +607,37 @@ private class UCProxy(
...
+      tables ++= response.getTables.asScala
       pageToken = response.getNextPageToken
     } while (pageToken != null && pageToken.nonEmpty)
-    tables.toArray
+    tables.toSeq
   }
 
-  override def loadTable(ident: Identifier): Table = {
-    val t = try {
+  private def getUcTable(ident: Identifier): io.unitycatalog.client.model.TableInfo = {
```
chenwang-databricks: Move the import to the header of the file and only use the short name.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -672,72 +765,180 @@ private class UCProxy(
       .asInstanceOf[Table]
   }
 
-  override def createTable(ident: Identifier, schema: StructType, partitions: Array[Transform], properties: util.Map[String, String]): Table = {
+  override def createView(ident: Identifier, info: ViewInfo): ViewInfo = {
     UCSingleCatalog.checkUnsupportedNestedNamespace(ident.namespace())
-    UCSingleCatalog.requireProviderSpecified("CREATE TABLE", properties)
+    val properties: util.Map[String, String] = info.properties()
+    val tableType = properties.get(TableCatalog.PROP_TABLE_TYPE)
+    if (TableSummary.METRIC_VIEW_TABLE_TYPE != tableType) {
```
chenwang-databricks: Please double check what are all the supported view types in UC.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -672,72 +765,180 @@ private class UCProxy(
...
+    }
+    if (t.getTableType != TableType.METRIC_VIEW) {
+      return false
+    }
+    tablesApi.deleteTable(UCSingleCatalog.fullTableNameForApi(this.name, ident)) == 200
+  }
+
+  override def createTable(
+      ident: Identifier,
+      tableInfo: org.apache.spark.sql.connector.catalog.TableInfo): Table = {
```
chenwang-databricks: use short name import

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -672,72 +765,180 @@ private class UCProxy(
...
+    Option(t.getViewDependencies).foreach { ucDeps =>
+      builder.withViewDependencies(UCProxy.fromUcDependencyList(ucDeps))
+    }
+    builder.build()
+  }
+
+  private def metadataJsonFromTypeJson(typeJson: String): String = {
+    if (typeJson == null || typeJson.isEmpty) return null
+    try {
+      val parsed = org.json4s.jackson.JsonMethods.parse(typeJson)
```
chenwang-databricks: Use short imports

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -672,72 +765,180 @@ private class UCProxy(
...
+      (parsed \ "metadata") match {
+        case obj: org.json4s.JObject =>
+          org.json4s.jackson.JsonMethods.compact(org.json4s.jackson.JsonMethods.render(obj))
```
chenwang-databricks: Use short imports

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -870,3 +1071,118 @@ private class UCProxy(
...
+    val list = Option(ucDeps.getDependencies)
+      .map(_.asScala.toSeq)
+      .getOrElse(Seq.empty)
+    val sparkDeps: Array[org.apache.spark.sql.connector.catalog.Dependency] = list.flatMap { d =>
+      Option(d.getTable).map { td =>
+        org.apache.spark.sql.connector.catalog.Dependency.table(td.getTableFullName)
+          .asInstanceOf[org.apache.spark.sql.connector.catalog.Dependency]
+      }
+    }.toArray
+    org.apache.spark.sql.connector.catalog.DependencyList.of(sparkDeps: _*)
```
chenwang-databricks: Use short imports

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -100,6 +125,65 @@ class UCSingleCatalog
...
+  override def renameView(oldIdent: Identifier, newIdent: Identifier): Unit =
+    ucProxy.renameView(oldIdent, newIdent)
+
+  /**
+   * `TableViewCatalog.loadTableOrView` is Spark's single-RPC perf entry point: the resolver
+   * calls it once per identifier and uses its return type to discriminate tables from views,
+   * instead of falling back from `loadTable` to `loadView`. We honor that contract by
+   * performing one UC `getTable` and dispatching on the UC `TableType`:
+   *
+   *   - For view-like UC table types (metric views today, materialized views in the future)
```
chenwang-databricks: Why are materialized views in the future? This is confusing. We already have these two view types defined, so we can just say the view is like UC tables, defined by its view table type function, something like that.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -672,72 +813,201 @@ private class UCProxy(
...
+    // UC's `TableType` enum has two view-like values: METRIC_VIEW (created via
+    // `CREATE METRIC VIEW`) and MATERIALIZED_VIEW (managed by Lakeflow / DLT pipelines, not
+    // by Spark DDL). MATERIALIZED_VIEW is not user-creatable through `ViewCatalog.createView`
+    // -- the pipeline runtime owns its lifecycle -- so the only view kind we accept here is
+    // METRIC_VIEW. Plain `CREATE VIEW` (`tableType == "VIEW"`) and any future view kinds are
+    // rejected explicitly so callers see a clear error rather than a silent type coercion.
+    if (TableSummary.METRIC_VIEW_TABLE_TYPE != tableType) {
+      throw new ApiException(
+        s"Unity Catalog only supports METRIC_VIEW via ViewCatalog.createView; got: $tableType")
+    }
```
chenwang-databricks: I think this part exposes too much about the specific implementation details about the Spark or pipeline stuff. I think we should make this part more generalized.

Basically, you have something similar to this: this is the accepted view-like, just like it's a table view-like function that you previously defined. We put a metric view or materialized view there to check, and if the table type is not like that, then we reject it here.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -672,72 +813,201 @@ private class UCProxy(
...
+    tablesApi.deleteTable(UCSingleCatalog.fullTableNameForApi(this.name, ident)) == 200
+  }
+
+  // UC OSS does not currently expose a rename endpoint for tables or views; mirror the
+  // existing `renameTable` stub below until UC adds a rename API.
+  override def renameView(oldIdent: Identifier, newIdent: Identifier): Unit = {
+    throw new UnsupportedOperationException("Renaming a view is not supported yet")
+  }
+
+  override def createTable(
```
chenwang-databricks: Should you keep the Create Table function not changed?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -672,72 +813,201 @@ private class UCProxy(
...
+    }.toArray
+
+    val props = new util.HashMap[String, String]()
+    Option(t.getProperties).foreach(props.putAll)
+    val sqlConfigs = extractSqlConfigs(props)
+
+    val builder = new ViewInfo.Builder()
+      .withColumns(columns)
+      .withProperties(props)
+      .withTableType(TableSummary.METRIC_VIEW_TABLE_TYPE)
```
chenwang-databricks: Should this be passed in? This should be a very generic function and shouldn't be specific to metric view only.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -672,72 +813,201 @@ private class UCProxy(
...
+      .withSchemaMode("UNSUPPORTED")
+      .withQueryColumnNames(columns.map(_.name()))
+    Option(t.getComment).foreach(builder.withComment)
+    Option(t.getViewDependencies).foreach { ucDeps =>
+      builder.withViewDependencies(UCProxy.fromUcDependencyList(ucDeps))
+    }
+    builder.build()
+  }
+
+  private def metadataJsonFromTypeJson(typeJson: String): String = {
```
chenwang-databricks: Why is this needed?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -806,6 +1076,17 @@ private class UCProxy(
   }
 
   override def dropTable(ident: Identifier): Boolean = {
```
chenwang-databricks: Do you need to make changes to this? I thought on the Spark side, it will call the corresponding view or table method on the UC catalog directly, right? Drop table on a metric view should not reach here?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -870,3 +1151,129 @@ private class UCProxy(
...
+      if (partitionIdx >= 0) column.setPartitionIndex(partitionIdx)
+      column
+    }
+  }
+
+  /**
+   * Builds the JSON string Unity Catalog stores in `ColumnInfo.type_json`. This is the full
+   * Spark `StructField` JSON (`name` / `type` / `nullable` / `metadata`), not the bare
+   * `DataType` JSON. We construct the `StructField` here rather than calling
+   * `CatalogV2Util.v2ColumnsToStructType` because that helper is `private[sql]` in Spark and
```
chenwang-databricks: Can we change that private[sql] to be able to be used here?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -537,6 +619,43 @@ object UCSingleCatalog {
...
+   * `TableSummary` table-type. Throws for view-like UC types we list but cannot yet load
+   * (e.g. `MATERIALIZED_VIEW` until Spark adds a `TableSummary.MATERIALIZED_VIEW_TABLE_TYPE`).
+   */
+  def ucTableTypeToSparkViewType(tableType: TableType): String =
+    viewLikeUcTypes.get(tableType).flatten.getOrElse(
+      throw new UnsupportedOperationException(
+        s"No Spark TableSummary view-type string for UC table type: $tableType"))
+
+  /** Used by `createView` to reject Spark `TableSummary` view-type strings UC OSS does not support. */
+  def isCreateViewSupportedTableType(tableType: String): Boolean =
```
chenwang-databricks: isViewCommandsSupportedTableType

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -537,6 +619,43 @@ object UCSingleCatalog {
...
+   * Used by `toViewInfo` to label the wrapped `MetadataTable` with the right Spark
+   * `TableSummary` table-type. Throws for view-like UC types we list but cannot yet load
+   * (e.g. `MATERIALIZED_VIEW` until Spark adds a `TableSummary.MATERIALIZED_VIEW_TABLE_TYPE`).
+   */
+  def ucTableTypeToSparkViewType(tableType: TableType): String =
+    viewLikeUcTypes.get(tableType).flatten.getOrElse(
+      throw new UnsupportedOperationException(
+        s"No Spark TableSummary view-type string for UC table type: $tableType"))
+
+  /** Used by `createView` to reject Spark `TableSummary` view-type strings UC OSS does not support. */
```
chenwang-databricks: Should be used by dropView as well

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -672,6 +834,78 @@ private class UCProxy(
...
+    throw new UnsupportedOperationException("Replacing a view is not supported yet")
+  }
+
+  override def dropView(ident: Identifier): Boolean = {
+    val t = try {
+      getUcTable(ident)
+    } catch {
+      case _: NoSuchTableException => return false
+    }
+    if (t.getTableType != TableType.METRIC_VIEW) {
```
chenwang-databricks: Use isCreateViewSupportedTableType

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -672,6 +834,78 @@ private class UCProxy(
...
+    if (!UCSingleCatalog.isCreateViewSupportedTableType(tableType)) {
+      throw new ApiException(
+        s"Unity Catalog does not support creating $tableType via ViewCatalog.createView")
+    }
+
+    val ct = new CreateTable()
+      .name(ident.name())
+      .schemaName(ident.namespace().head)
+      .catalogName(this.name)
+      .tableType(TableType.METRIC_VIEW)
```
chenwang-databricks: This should be passed in from the view info

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -870,3 +1166,105 @@ private class UCProxy(
...
   }
 }
+
+private object UCProxy {
+
+  /**
+   * Converts Spark's typed [[DependencyList]] into the wire-format UC [[UCDependencyList]].
+   * Only `TableDependency` is currently translated; UC OSS does not persist function
+   * dependencies (metric views disallow UDFs in their source) and other `Dependency`
+   * subtypes are dropped on the floor.
```
chenwang-databricks: Just say UC OSS does not persist function dependencies for now

PR #1516, opened 2026-04-23
New endpoint for Delta: createTable

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1516/files) to review incremental changes.
- [**stack/create_table**](https://github.com/unitycatalog/unitycatalog/pull/1516) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1516/files)]
  - [stack/create_table_last_ts](https://github.com/unitycatalog/unitycatalog/pull/1528) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1528/files/1b4f23bf314badedb5494ff53c1d38b5e23eb15d..f3c906d85402b43558503405f4aaf2d931c40c1b)]
  - [stack/create_table_tests](https://github.com/unitycatalog/unitycatalog/pull/1534) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1534/files/1b4f23bf314badedb5494ff53c1d38b5e23eb15d..65274357915c2ed117a7ace0adc7e9ceb9ade868)]
  - [stack/create_table_uniform](https://github.com/unitycatalog/unitycatalog/pull/1525) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1525/files/1b4f23bf314badedb5494ff53c1d38b5e23eb15d..cdd4b807aed94156712cd088f4dc710e59111322)]
    - stack/drc_update_table
  - [stack/not_supported_format](https://github.com/unitycatalog/unitycatalog/pull/1535) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1535/files/1b4f23bf314badedb5494ff53c1d38b5e23eb15d..74879e2f8a035c658f5b6ed957a552f855f4caff)]
  - stack/temp_path_cred
  - [stack/validate_columns](https://github.com/unitycatalog/unitycatalog/pull/1526) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1526/files/1b4f23bf314badedb5494ff53c1d38b5e23eb15d..4e6e9345b74524dbc1de0e7cdf14b01443eccd64)]
    - [stack/column_case_ins](https://github.com/unitycatalog/unitycatalog/pull/1537) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1537/files/4e6e9345b74524dbc1de0e7cdf14b01443eccd64..97c82cf53a398bd313deb60174de9d668208d03a)]

---------
**PR Checklist**

**Description of changes**
New endpoint for Delta: createTable

Adds POST /v1/catalogs/{catalog}/schemas/{schema}/tables at parity with UC REST TableService.createTable. MANAGED (finalize a staging table by location) and EXTERNAL flows are supported; DELTA is the only accepted data-source-format.

For MANAGED, UcManagedDeltaContract.validate enforces protocol versions, required features, reader-subset-of-writer, domain-metadata vs feature consistency, and required properties. validateTableIdProperty additionally pins properties[UC_TABLE_ID] to the staging-allocated UUID so a client cannot persist an internally-inconsistent table.

DeltaPropertyMapper projects protocol features (delta.feature.*) and domain-metadata (delta.clusteringColumns, delta.rowTracking.rowIdHighWaterMark) onto stored UC properties and may override client-supplied properties on conflict.

AuthorizeExpressions.CREATE_TABLE replaces the inlined SpEL on TableService.createTable so UC REST and DRC share one source of truth -- drift becomes a compile-time impossibility. AuthorizeKeyLocator.getVariableName maps hyphens to underscores so DRC's kebab-case table-type surfaces as the SpEL variable #table_type; snake_case keys are unaffected.

TableRepository.createTableForDelta returns LoadTableResponse in a single transaction, sharing the persisted DAO with buildLoadTableResponse (extracted from loadTableForDelta). Both create paths go through a generic createTableImpl helper. ColumnUtils gains toColumnInfo / toCatalogString / applyPartitionColumns for DRC StructField -> UC ColumnInfo.

It also tightens the existing UC REST POST /tables MANAGED path to require properties[delta.UCTableId]

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/auth/AuthorizeExpressions.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/AuthorizeExpressions.java
@@ -52,6 +52,28 @@ public final class AuthorizeExpressions {
       """;
 
   /**
+   * Authorization policy for creating a table (UC REST {@code POST /tables} and Delta REST Catalog
+   * {@code createTable}). Catalog {@code USE_CATALOG}/{@code OWNER} plus either schema
+   * {@code OWNER} or schema {@code USE_SCHEMA}+{@code CREATE_TABLE}. For EXTERNAL tables, the
+   * caller additionally needs {@code OWNER}/{@code CREATE_EXTERNAL_TABLE} on the external location
+   * (if one resolves) and the storage path must not overlap a data securable.
+   *
+   * <p>The {@code #table_type} SpEL variable comes from {@code @AuthorizeKey(key = "table-type")};
+   * kebab-case payload keys surface with hyphens mapped to underscores (see {@link
+   * io.unitycatalog.server.auth.decorator.AuthorizeKeyLocator#getVariableName}).
+   */
+  public static final String CREATE_TABLE =
+      """
+      #authorizeAny(#principal, #catalog, OWNER, USE_CATALOG) &&
+      (#authorize(#principal, #schema, OWNER) ||
+        #authorizeAll(#principal, #schema, USE_SCHEMA, CREATE_TABLE)) &&
+      (#table_type != 'EXTERNAL' ||
+        (#no_overlap_with_data_securable &&
+          (#external_location == null ||
+           #authorizeAny(#principal, #external_location, OWNER, CREATE_EXTERNAL_TABLE))))
+      """;
+
+  /**
    * Authorization policy for vending table credentials. Admin-above-the-table privileges on
    * their own are not sufficient; the caller must have an explicit table-level privilege
    * matching the requested operation. {@code READ} needs OWNER or SELECT; {@code READ_WRITE}
--- a/server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizeKeyLocator.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizeKeyLocator.java
@@ -43,14 +43,20 @@ public class AuthorizeKeyLocator {
   /**
    * Extracts the variable name from a key. For resources, returns their securable type as variable
    * name. e.g. "external_location" For other keys, returns the key name in annotated parameters or
-   * the last segment of the payload parameter. e.g. "config.operation" returns "operation"
+   * the last segment of the payload parameter. e.g. "config.operation" returns "operation".
+   *
+   * <p>Hyphens in the key are mapped to underscores so kebab-case payload fields (such as Delta
+   * REST Catalog's {@code table-type}) form valid SpEL identifiers like {@code #table_type}. The
+   * payload lookup still uses the original key verbatim; this transformation affects only the SpEL
+   * variable name.
    */
   public String getVariableName() {
     if (type.isPresent()) {
       return type.get().getValue();
     }
     int lastDot = key.lastIndexOf('.');
-    return lastDot >= 0 ? key.substring(lastDot + 1) : key;
+    String name = lastDot >= 0 ? key.substring(lastDot + 1) : key;
+    return name.replace('-', '_');
   }
 
   public static AuthorizeKeyLocator from(AuthorizeResourceKey key, Parameter parameter) {
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
@@ -1,0 +1,100 @@
+package io.unitycatalog.server.service.delta;
+
+import io.unitycatalog.server.delta.model.CreateTableRequest;
+import io.unitycatalog.server.delta.model.StructField;
+import io.unitycatalog.server.exception.BaseException;
+import io.unitycatalog.server.exception.ErrorCode;
+import io.unitycatalog.server.model.ColumnInfo;
+import io.unitycatalog.server.model.CreateTable;
+import io.unitycatalog.server.model.DataSourceFormat;
+import io.unitycatalog.server.model.TableType;
+import io.unitycatalog.server.utils.ColumnUtils;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Converts a Delta REST Catalog {@link CreateTableRequest} (with typed Delta columns and kebab-case
+ * field names) into the UC {@link CreateTable} (with UC {@link ColumnInfo}s and
+ * partition-index-per-column). The server holds path params for catalog and schema; the rest comes
+ * from the request body.
+ *
+ * <p>Required-field checks (name, location, columns, protocol, table-type, data-source-format) and
+ * the DELTA-only format rule apply to all tables. Domain-metadata vs feature consistency and the
+ * full UC catalog-managed contract ({@link UcManagedDeltaContract}) apply only to MANAGED tables;
+ * for EXTERNAL tables UC mirrors what the client wrote with the Delta log as the source of truth.
+ */
+public final class DeltaCreateTableMapper {
+
+  private DeltaCreateTableMapper() {}
+
+  public static CreateTable toCreateTable(String catalog, String schema, CreateTableRequest req) {
+    if (req == null) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Request body is required.");
+    }
+    if (req.getName() == null || req.getName().isBlank()) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table name is required.");
+    }
+    if (req.getLocation() == null || req.getLocation().isBlank()) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table location is required.");
+    }
+    if (req.getColumns() == null || req.getColumns().getFields() == null) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table columns are required.");
+    }
+
+    TableType tableType = toUCTableType(req.getTableType());
+    if (req.getProtocol() == null) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "protocol is required.");
+    }
+
+    // MANAGED-only: full UC catalog-managed contract (protocol versions + features + reader-subset
+    // + domain-metadata consistency + properties). EXTERNAL tables get a pass: UC mirrors what the
+    // client wrote; the Delta log is the source of truth.
+    if (tableType == TableType.MANAGED) {
+      UcManagedDeltaContract.validate(
+          req.getProtocol(), req.getDomainMetadata(), req.getProperties());
+    }
+
+    List<ColumnInfo> columns = new ArrayList<>();
+    List<StructField> fields = req.getColumns().getFields();
+    for (int i = 0; i < fields.size(); i++) {
+      columns.add(ColumnUtils.toColumnInfo(fields.get(i), i));
+    }
+    ColumnUtils.applyPartitionColumns(columns, req.getPartitionColumns());
+
+    return new CreateTable()
+        .name(req.getName())
+        .catalogName(catalog)
+        .schemaName(schema)
+        .tableType(tableType)
+        .dataSourceFormat(toUCDataSourceFormat(req.getDataSourceFormat()))
+        .columns(columns)
+        .comment(req.getComment())
+        .storageLocation(req.getLocation())
+        .properties(
+            DeltaPropertyMapper.mergeDerivedWithClient(
+                req.getProtocol(), req.getDomainMetadata(), req.getProperties()));
+  }
+
+  private static TableType toUCTableType(io.unitycatalog.server.delta.model.TableType type) {
+    if (type == null) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "table-type is required.");
+    }
+    return switch (type) {
+      case MANAGED -> TableType.MANAGED;
+      case EXTERNAL -> TableType.EXTERNAL;
+    };
+  }
+
+  private static DataSourceFormat toUCDataSourceFormat(
+      io.unitycatalog.server.delta.model.DataSourceFormat format) {
+    if (format == null) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "data-source-format is required.");
+    }
+    // Only DELTA is accepted.
+    if (format == io.unitycatalog.server.delta.model.DataSourceFormat.DELTA) {
+      return DataSourceFormat.DELTA;
+    }
+    throw new BaseException(
+        ErrorCode.INVALID_ARGUMENT, "Unsupported data-source-format: " + format.getValue());
+  }
+}
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
@@ -19,9 +19,11 @@ import java.util.List;
  * from the request body.
  *
  * <p>Required-field checks (name, location, columns, protocol, table-type, data-source-format) and
- * the DELTA-only format rule apply to all tables. Domain-metadata vs feature consistency and the
- * full UC catalog-managed contract ({@link UcManagedDeltaContract}) apply only to MANAGED tables;
- * for EXTERNAL tables UC mirrors what the client wrote with the Delta log as the source of truth.
+ * the DELTA-only format rule apply to all tables. The full UC catalog-managed contract ({@link
+ * UcManagedDeltaContract}) applies only to MANAGED tables; EXTERNAL tables skip contract validation
+ * but still go through the same {@link DeltaPropertyMapper} projection, so derived
+ * {@code delta.feature.*} and {@code delta.clusteringColumns} entries override any client-supplied
+ * values under those keys.
  */
 public final class DeltaCreateTableMapper {
 
@@ -73,7 +75,7 @@ public final class DeltaCreateTableMapper {
         .comment(req.getComment())
         .storageLocation(req.getLocation())
         .properties(
-            DeltaPropertyMapper.mergeDerivedWithClient(
+            DeltaPropertyMapper.buildStoredProperties(
                 req.getProtocol(), req.getDomainMetadata(), req.getProperties()));
   }
 
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
@@ -37,7 +37,9 @@ public final class DeltaCreateTableMapper {
     if (req.getLocation() == null || req.getLocation().isBlank()) {
       throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table location is required.");
     }
-    if (req.getColumns() == null || req.getColumns().getFields() == null) {
+    if (req.getColumns() == null
+        || req.getColumns().getFields() == null
+        || req.getColumns().getFields().isEmpty()) {
       throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table columns are required.");
     }
 
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaPropertyMapper.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaPropertyMapper.java
@@ -1,0 +1,115 @@
+package io.unitycatalog.server.service.delta;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.unitycatalog.server.delta.model.ClusteringDomainMetadata;
+import io.unitycatalog.server.delta.model.DeltaProtocol;
+import io.unitycatalog.server.delta.model.DomainMetadataUpdates;
+import io.unitycatalog.server.delta.model.RowTrackingDomainMetadata;
+import io.unitycatalog.server.exception.BaseException;
+import io.unitycatalog.server.exception.ErrorCode;
+import io.unitycatalog.server.service.delta.DeltaConsts.TableProperties;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Maps Delta REST Catalog {@code protocol} and {@code domain-metadata} blocks onto UC table
+ * properties. Designed to be shared by create, update, and commit endpoints: the mapping rules are
+ * specified by the Delta spec and don't vary by endpoint (only create is wired up today; update and
+ * commit are on follow-ups).
+ *
+ * <p>Every feature in {@code protocol.reader-features} and {@code protocol.writer-features} yields
+ * {@code delta.feature.<name> = supported}. Reader-writer features (which appear in both lists)
+ * collapse to a single entry via map dedup.
+ *
+ * <p>Domain-metadata entries yield their Delta-documented table-property projections:
+ *
+ * <ul>
+ *   <li>The {@code delta.clustering} domain-metadata entry's {@code clusteringColumns} field
+ *       projects to the {@code delta.clusteringColumns} table property as a JSON-encoded list of
+ *       column paths. Paths are arrays of segment names so nested columns survive as arrays rather
+ *       than collapsing into dotted strings.
+ *   <li>The {@code delta.rowTracking} domain-metadata entry's {@code rowIdHighWaterMark} field
+ *       projects to the {@code delta.rowTracking.rowIdHighWaterMark} table property.
+ * </ul>
+ *
+ * <p>Consistency rules (e.g. domain-metadata entry implies its writer feature) live in {@link
+ * UcManagedDeltaContract}, not here.
+ */
+public final class DeltaPropertyMapper {
+
+  private DeltaPropertyMapper() {}
+
+  private static final ObjectMapper JSON = new ObjectMapper();
+
+  /** Value emitted under {@code delta.feature.<name>} for each declared feature. */
+  private static final String FEATURE_SUPPORTED = "supported";
+
+  /**
+   * Returns the stored UC property map formed by merging protocol-derived, domain-metadata-derived,
+   * and client-supplied entries. Client-supplied entries take precedence so callers can pin
+   * engine-managed values (e.g. row-tracking materialized column names) or override a derived value
+   * if needed.
+   */
+  public static Map<String, String> mergeDerivedWithClient(
+      DeltaProtocol protocol,
+      DomainMetadataUpdates domainMetadata,
+      Map<String, String> clientProperties) {
+    Map<String, String> merged = new HashMap<>();
+    deriveFromProtocol(merged, protocol);
+    deriveFromDomainMetadata(merged, domainMetadata);
+    if (clientProperties != null) {
+      merged.putAll(clientProperties);
+    }
+    return merged;
+  }
+
+  /**
+   * Adds the {@code delta.feature.*} properties implied by a protocol block into {@code props}
+   * in-place. Iterates both reader- and writer-features; reader-writer features (which appear in
+   * both lists) collapse to a single entry via map dedup. The mapper does not assume the protocol
+   * has been validated -- if a reader-feature is missing from writer-features (a Delta-spec
+   * violation that {@link UcManagedDeltaContract#validate} catches), this method still emits an
+   * entry for it.
+   */
+  public static void deriveFromProtocol(Map<String, String> props, DeltaProtocol protocol) {
+    if (protocol == null) return;
+    addFeatureProperties(props, protocol.getReaderFeatures());
+    addFeatureProperties(props, protocol.getWriterFeatures());
+  }
+
+  /** Adds the table properties implied by a domain-metadata block into {@code props} in-place. */
+  public static void deriveFromDomainMetadata(
+      Map<String, String> props, DomainMetadataUpdates domainMetadata) {
+    if (domainMetadata == null) return;
+    ClusteringDomainMetadata clustering = domainMetadata.getDeltaClustering();
+    if (clustering != null && clustering.getClusteringColumns() != null) {
+      // Use a proper JSON encoder rather than string-joining so nested column paths don't
+      // collapse into dotted strings.
+      props.put(TableProperties.CLUSTERING_COLUMNS, toJson(clustering.getClusteringColumns()));
+    }
+    RowTrackingDomainMetadata rowTracking = domainMetadata.getDeltaRowTracking();
+    if (rowTracking != null && rowTracking.getRowIdHighWaterMark() != null) {
+      props.put(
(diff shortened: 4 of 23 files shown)


timothyw553:
will this include `GET /delta/v1/temporary-path-credentials`?

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -149,4 +156,132 @@ public static String toTypeJson(StructField field) {
...
+      case "long" -> ColumnTypeName.LONG;
+      case "float" -> ColumnTypeName.FLOAT;
+      case "double" -> ColumnTypeName.DOUBLE;
+      case "date" -> ColumnTypeName.DATE;
+      case "timestamp" -> ColumnTypeName.TIMESTAMP;
+      case "timestamp_ntz" -> ColumnTypeName.TIMESTAMP_NTZ;
+      case "string" -> ColumnTypeName.STRING;
+      case "binary" -> ColumnTypeName.BINARY;
+      case "variant" -> ColumnTypeName.VARIANT;
+      case "null" -> ColumnTypeName.NULL;
```
timothyw553: "void", "null" can both map to ColumnTypeName.NULL

yili-db:
> will this include `GET /delta/v1/temporary-path-credentials`?

It's already merged https://github.com/unitycatalog/unitycatalog/pull/1500

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -149,4 +156,132 @@ public static String toTypeJson(StructField field) {
...
+   * Convert a Delta REST Catalog {@link StructField} into a UC {@link ColumnInfo}. The reverse of
+   * {@link #toStructField(ColumnInfo)}: {@code typeJson} is the authoritative Spark-format column
+   * spec, {@code typeName} is the top-level ColumnTypeName, and {@code typeText} is the recursive
+   * Spark-style catalog string ({@code array<string>}, {@code struct<name:type>}). Column
+   * comments in {@code metadata.comment} (per Delta spec) are lifted into the dedicated
+   * {@code ColumnInfo.comment} field. Mirrors {@code UCSingleCatalog.createTable}'s per-column
+   * projection so DRC-created tables render identically to Spark-created ones.
+   */
+  public static ColumnInfo toColumnInfo(StructField field, int position) {
+    DeltaType type = field.getType();
```
timothyw553, line 181: can we make it defensive ensuring all fields are sent?

yili-db: You mean typePrecision, typeScale and typeIntervalType?
They aren't set by `UCSingleCatalog` neither.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
yili-db: Removed NULL. Actually none of them are in Delta spec.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    List<ColumnInfo> columns = new ArrayList<>();
+    List<StructField> fields = req.getColumns().getFields();
+    for (int i = 0; i < fields.size(); i++) {
+      columns.add(ColumnUtils.toColumnInfo(fields.get(i), i));
+    }
+    ColumnUtils.applyPartitionColumns(columns, req.getPartitionColumns());
+
+    return new CreateTable()
```
ChengJiX, line 68: Could we also add support for UniForm here? Seems that it needs a API change for `createTable` first, though

yili-db: Will be a follow up PR

openinx:
Hi @yili-db , could you pls help to rebase this to the latest main branch ?  seems like it's conflicting now.

server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizeKeyLocator.java
```diff
@@ -43,14 +43,20 @@ public enum Source {
   /**
    * Extracts the variable name from a key. For resources, returns their securable type as variable
    * name. e.g. "external_location" For other keys, returns the key name in annotated parameters or
-   * the last segment of the payload parameter. e.g. "config.operation" returns "operation"
+   * the last segment of the payload parameter. e.g. "config.operation" returns "operation".
+   *
+   * <p>Hyphens in the key are mapped to underscores so kebab-case payload fields (such as Delta
+   * REST Catalog's {@code table-type}) form valid SpEL identifiers like {@code #table_type}. The
```
openinx, line 49: > SpEL identifiers

Could you pls help me understand, what's the SpEL identifiers ?

openinx: okay, got it from google: https://www.google.com/search?q=SpEL+identifiers+in+java+

server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizeKeyLocator.java
```diff
@@ -43,14 +43,20 @@ public enum Source {
...
+   * variable name.
    */
   public String getVariableName() {
     if (type.isPresent()) {
       return type.get().getValue();
     }
     int lastDot = key.lastIndexOf('.');
-    return lastDot >= 0 ? key.substring(lastDot + 1) : key;
+    String name = lastDot >= 0 ? key.substring(lastDot + 1) : key;
+    return name.replace('-', '_');
```
openinx, line 59: my question is:   will there be any variable name defined with hyphen rather than the underscore ?  seems like it's weird.

server/src/main/java/io/unitycatalog/server/auth/AuthorizeExpressions.java
```diff
@@ -37,6 +37,42 @@ private AuthorizeExpressions() {}
...
+  /**
+   * Authorization policy for creating a staging table (UC REST {@code POST /staging-tables} and
+   * Delta REST Catalog {@code createStagingTable}). Catalog {@code USE_CATALOG}/{@code OWNER}
+   * plus either schema {@code OWNER} or schema {@code USE_SCHEMA}+{@code CREATE_TABLE}. Catalog
+   * OWNER alone is not sufficient.
+   */
+  public static final String CREATE_STAGING_TABLE =
+      """
+      (#authorizeAny(#principal, #catalog, OWNER, USE_CATALOG)
+        && #authorize(#principal, #schema, OWNER)) ||
```
openinx, line 49: Since here we use `||` , will it be a problem if someone use the DeltaRestApi authorization to request the old rest API, and pass the auth where it should be ?

openinx: (not sure if this is a problem)

server/src/main/java/io/unitycatalog/server/auth/AuthorizeExpressions.java
```diff
@@ -37,6 +37,42 @@ private AuthorizeExpressions() {}
...
+      """;
+
+  /**
+   * Authorization policy for creating a table (UC REST {@code POST /tables} and Delta REST Catalog
+   * {@code createTable}). Catalog {@code USE_CATALOG}/{@code OWNER} plus either schema
+   * {@code OWNER} or schema {@code USE_SCHEMA}+{@code CREATE_TABLE}. For EXTERNAL tables, the
+   * caller additionally needs {@code OWNER}/{@code CREATE_EXTERNAL_TABLE} on the external location
+   * (if one resolves) and the storage path must not overlap a data securable.
+   *
+   * <p>The {@code #table_type} SpEL variable comes from {@code @AuthorizeKey(key = "table-type")};
```
openinx, line 61: okay, because we already define the `table-type` in the DeltaRestApi protocol, that's why we have to convert it to be a SpEL variable which can be recongized by Java ?

server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
```diff
@@ -0,0 +1,100 @@
...
+    // client wrote; the Delta log is the source of truth.
+    if (tableType == TableType.MANAGED) {
+      UcManagedDeltaContract.validate(
+          req.getProtocol(), req.getDomainMetadata(), req.getProperties());
+    }
+
+    List<ColumnInfo> columns = new ArrayList<>();
+    List<StructField> fields = req.getColumns().getFields();
+    for (int i = 0; i < fields.size(); i++) {
+      columns.add(ColumnUtils.toColumnInfo(fields.get(i), i));
```
openinx, line 64: Could you pls add the validation to reject the invalid type json ? this is exactly the bug that @TimothyW553  is trying to fix in this PR: https://github.com/unitycatalog/unitycatalog/pull/1524

server/src/main/java/io/unitycatalog/server/auth/decorator/AuthorizeKeyLocator.java
yili-db: The new API choses to use kebab casing over the wire. This is for `table-type`.

server/src/main/java/io/unitycatalog/server/auth/AuthorizeExpressions.java
yili-db: This is just an auth expression shared between the old and new createTable endpoints.

server/src/main/java/io/unitycatalog/server/auth/AuthorizeExpressions.java
yili-db: Yes. But this is just SpEL's naming rule and not Java variable naming rule.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
yili-db: This new API is naturally immune to the invalid type-json problem. Because over the wire it's already required to be a valid json structure defined in `delta.yaml` file. It's already validated and deserialized as part of the request payload. There is no typeJson string in RPC:
```
        columns:
          $ref: '#/components/schemas/StructType'
```

openinx: Hi @yili-db ,  my point is:  if the client does not strickly follow the spec ( I know the unitycatalog-client definitely will follow the spec, because it's naturally generated from the spec. but there are other clients which are not generated from the spec,  such as rust-client, kernel-client ),  in that cases,  the server should explicitly report the failure and require the client to get a correctly formatted structure, right ? 

That's why I suggested to add the strictly check at the server side, so that the self-customized client have no chance to break the spec.

server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
```diff
@@ -86,6 +92,184 @@ private static Map<String, String> buildSuggestedProperties() {
...
+    }
+  }
+
+  private static void validateRequiredFeatures(DeltaProtocol protocol) {
+    Set<String> readerFeatures =
+        Set.copyOf(protocol.getReaderFeatures() != null ? protocol.getReaderFeatures() : List.of());
+    Set<String> writerFeatures =
+        Set.copyOf(protocol.getWriterFeatures() != null ? protocol.getWriterFeatures() : List.of());
+    for (String required : REQUIRED_READER_FEATURES) {
+      if (!readerFeatures.contains(required)) {
```
openinx, line 198: Are we still allowed to create the table like `CREATE TABLE xx (...) TBLPROPERTIES ('delta.feature.catalogManaged'='supported')`,  because I see here we are required to set all the table features explicitly, such as `delta.feature.deletionVectors` etc. 

So I think the current behavior is: we should be able to do the `CREATE TABLE xx (...) TBLPROPERTIES ('delta.feature.catalogManaged'='supported')`.  since we didn't explicitly add the `delta.feature.deletionVectors` in the `TBLPROPERTIES`, then is the delta-spark currently explicitly added it ?

openinx (approved):
I have two comments,  @yili-db  pls confirm and see if it makes sense. no other comments.

server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
yili-db: This question has two layers:
1. From SQL command to Delta Spark: Setting 'delta.feature.catalogManaged'='supported' is still required unless Delta Spark changes. But all other features shall be automatically turned on.
2. From Delta Spark to UC: It's the Delta Spark's responsibility to enable all required features and set all required table properties. UC will validate this.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
yili-db: @openinx Sounds good. I am making a follow up PR https://github.com/unitycatalog/unitycatalog/pull/1526 for this.

server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
openinx: okay, thanks for the clarification.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
openinx: thanks for the following up, @yili-db

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
```diff
@@ -160,6 +163,28 @@ public StagingTableResponse createStagingTable(
...
+  @ProducesJson
+  @AuthorizeExpression(AuthorizeExpressions.CREATE_TABLE)
+  public LoadTableResponse createTable(
+      @Param("catalog") @AuthorizeResourceKey(CATALOG) String catalog,
+      @Param("schema") @AuthorizeResourceKey(SCHEMA) String schema,
+      @AuthorizeResourceKey(value = EXTERNAL_LOCATION, key = "location")
+          @AuthorizeKey(key = "table-type")
+          CreateTableRequest request) {
+    CreateTable createTable = DeltaCreateTableMapper.toCreateTable(catalog, schema, request);
+    return tableRepository.createTableForDelta(createTable);
```
timothyw553: are we missing `initializeHierarchicalAuthorization` that TableService.createTable does for EXTERNAL tables?

https://github.com/unitycatalog/unitycatalog/blob/a4b491a2fcdc48a27771c6a3c9ed3a06669cfeac/server/src/main/java/io/unitycatalog/server/service/TableService.java#L89-L105

and then please add tests.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaPropertyMapper.java
```diff
@@ -0,0 +1,115 @@
...
+  public static Map<String, String> mergeDerivedWithClient(
+      DeltaProtocol protocol,
+      DomainMetadataUpdates domainMetadata,
+      Map<String, String> clientProperties) {
+    Map<String, String> merged = new HashMap<>();
+    deriveFromProtocol(merged, protocol);
+    deriveFromDomainMetadata(merged, domainMetadata);
+    if (clientProperties != null) {
+      merged.putAll(clientProperties);
+    }
```
timothyw553, line 65: it seems like we're letting client properties win here as we derive first then use clientproperties. should we do the existing check first, merge.putall, then derive properties?

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
timothyw553: also, should we add logging to these?

server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
```diff
@@ -0,0 +1,100 @@
...
+    if (req == null) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Request body is required.");
+    }
+    if (req.getName() == null || req.getName().isBlank()) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table name is required.");
+    }
+    if (req.getLocation() == null || req.getLocation().isBlank()) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table location is required.");
+    }
+    if (req.getColumns() == null || req.getColumns().getFields() == null) {
```
timothyw553: `getFields().isEmpty()` falls through. a request with `{ "columns": { "type": "struct", "fields": [] } }` creates a UC table with zero columns; subsequent Delta commits write an empty StructType.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -149,4 +156,157 @@ public static String toTypeJson(StructField field) {
...
+   * <p>Runs in {@code O(|columns| + |partitionColumns|)}: builds a name → ColumnInfo lookup once,
+   * then walks {@code partitionColumns} stamping indices and rejecting unknown names inline.
+   */
+  public static void applyPartitionColumns(
+      List<ColumnInfo> columns, List<String> partitionColumns) {
+    if (partitionColumns == null || partitionColumns.isEmpty()) {
+      return;
+    }
+    Map<String, ColumnInfo> columnsByName =
+        columns.stream().collect(Collectors.toMap(ColumnInfo::getName, Function.identity()));
```
timothyw553, line 300: [AI-generated] `Collector.toMap(...)` might get tripped up by duplicate column names. duplicate partition entries `(["a","a"])` silently produce a UC partition spec that the read path then warn-empties — Delta engines reading `_delta_log` are unaffected, but UC's own view of the table is wrong.

timothyw553 (commented):
hi @yili-db i have left some comments.

yili-db:
> will this include `GET /delta/v1/temporary-path-credentials`?

@TimothyW553 It belongs to another PR.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaRestCatalogService.java
yili-db: Fixed. Currently most successful operations don't emit a log.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
yili-db: Fixing in https://github.com/unitycatalog/unitycatalog/pull/1526

timothyw553 (approved):
LGTM. Thanks @yili-db for this.

but please note in the description that this also requires `properties[delta.UCTableId]` on the existing `POST /tables` endpoint for MANAGED Delta tables.

tdas (commented):
[AI Assisted] Posting a few follow-ups along with one review-level note on testing gaps. None are blockers; they would tighten the PR before merge.

**Testing gaps (review-level, since they span multiple files):**

1. **End-to-end precedence rule.** `DeltaPropertyMapperTest.mergeDerivedPropertiesWinOnConflict` pins the new "derived wins over client" precedence at the unit level. There is no integration assertion that a request with `properties: {"delta.feature.catalogManaged": "client-override"}` still surfaces as `"supported"` in the stored properties through the wire boundary. One extra assertion in `SdkCreateTableTest.testCreateTableEndpoint` would pin this end-to-end, which matters because the precedence flipped between commits (`ae74351` → `e5693da`) and could regress silently.

2. **`CREATE_TABLE` SpEL branch coverage from DRC.** The shared expression has the EXTERNAL arm `(no_overlap_with_data_securable && (external_location == null || authorize(... OWNER, CREATE_EXTERNAL_TABLE)))`. The new DRC integration tests cover `external_location == null` and the implicit no-overlap path; they do not cover (a) EXTERNAL with a path that overlaps a data securable, (b) EXTERNAL with a resolved external_location and OWNER, (c) EXTERNAL with a resolved external_location and `CREATE_EXTERNAL_TABLE`. Since the SpEL constant is now shared with `TableService.createTable`, those branches may already be covered by existing UC REST tests — could you confirm which test pins each branch (or add a small DRC-side test) so we know §2g full conditional branch coverage is met?

3. **MANAGED-via-DRC OWNER semantics.** With the May 1 change, MANAGED now skips `initializeHierarchicalAuthorization` on the DRC path while UC REST `TableService.createTable` still calls it unconditionally. If the staging caller and the createTable caller can be different identities, the two surfaces produce different OWNER rows for the same operation. There is no test that pins the intended behavior on the DRC side. A test where principal-A calls `createStagingTable` and principal-B calls `createTable` (with the appropriate grants) and we then assert who is OWNER on the resulting table would make the divergence with UC REST deliberate and regression-proof.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Converts a Delta REST Catalog {@link CreateTableRequest} (with typed Delta columns and kebab-case
+ * field names) into the UC {@link CreateTable} (with UC {@link ColumnInfo}s and
+ * partition-index-per-column). The server holds path params for catalog and schema; the rest comes
+ * from the request body.
+ *
+ * <p>Required-field checks (name, location, columns, protocol, table-type, data-source-format) and
+ * the DELTA-only format rule apply to all tables. Domain-metadata vs feature consistency and the
+ * full UC catalog-managed contract ({@link UcManagedDeltaContract}) apply only to MANAGED tables;
+ * for EXTERNAL tables UC mirrors what the client wrote with the Delta log as the source of truth.
```
tdas: [AI Assisted] After the May 1 precedence flip in `DeltaPropertyMapper.mergeDerivedWithClient` (server-derived now wins over client), this class doc no longer reflects the EXTERNAL behavior. EXTERNAL still goes through `mergeDerivedWithClient` at line 76, so derived `delta.feature.*` and `delta.clusteringColumns` projections **override** client-supplied values under those keys — UC isn't "mirroring what the client wrote" anymore. Either tighten the doc to say `for EXTERNAL tables UC skips contract validation but still applies the structured-block (protocol / domain-metadata) projections`, or skip `mergeDerivedWithClient` for EXTERNAL and pass `req.getProperties()` through literally. The integration test `externalTableRequest` happens to set only `delta.enableDeletionVectors` (not a derived key) so it doesn't catch this contradiction.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaPropertyMapper.java
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * Returns the stored UC property map formed by layering client-supplied entries first, then
+   * overlaying protocol- and domain-metadata-derived entries on top. The structured {@code
+   * protocol} and {@code domain-metadata} blocks are the canonical source for the keys this method
+   * derives ({@code delta.feature.*}, {@code delta.clusteringColumns}, {@code
+   * delta.rowTracking.rowIdHighWaterMark}); a stray client property under those names cannot
+   * silently override the projection. Client-only properties (e.g. {@code
+   * delta.rowTracking.materializedRowIdColumnName}) flow through untouched.
+   */
+  public static Map<String, String> mergeDerivedWithClient(
```
tdas: [AI Assisted] After the precedence flip the name is misleading: `mergeDerivedWithClient` reads as "client wins" (apply derived, then merge client on top) but the implementation is now the opposite, and the test was correctly renamed to `mergeDerivedPropertiesWinOnConflict`. Suggest renaming to something like `buildStoredProperties(protocol, domainMetadata, clientProperties)` or `mergeWithDerivedOverride` so the call site at `DeltaCreateTableMapper.java:76` makes the precedence obvious. Per §4a (names match behavior).

server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
```diff
@@ -86,6 +92,184 @@ private static Map<String, String> buildSuggestedProperties() {
...
+   * through staging and don't carry this contract.
+   *
+   * <p>Designed to be called from {@link DeltaCreateTableMapper}, future update / commit mappers,
+   * and anywhere else that needs to assert "this state is a valid UC-managed Delta table state."
+   */
+  public static void validate(
+      DeltaProtocol protocol,
+      DomainMetadataUpdates domainMetadata,
+      Map<String, String> properties) {
+    Objects.requireNonNull(protocol, "protocol");
```
tdas: [AI Assisted] Raw `Objects.requireNonNull(protocol, "protocol")` throws `NullPointerException` (HTTP 500) rather than `BaseException(INVALID_ARGUMENT)` (HTTP 400). The mapper guards this today, but the class doc explicitly says `validate` is intended to be called from "future update / commit mappers, and anywhere else that needs to assert this state…". A future caller that forgets the upstream null-check would surface a 500 NPE to the user instead of a clean 400. Suggest:

```suggestion
    if (protocol == null) {
      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "protocol is required.");
    }
```

Keeps the failure mode consistent with the rest of this class.

yili-db: Introduced `ValidationUtils.checkNotNull` in the existing `ValidationUtils.java` file for this purpose and replace these `Objects.requireNonNull` calls.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaPropertyMapper.java
yili-db: Renamed to `buildStoredProperties`.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
yili-db: Updated this comment

yili-db:
> [AI Assisted] Posting a few follow-ups along with one review-level note on testing gaps. None are blockers; they would tighten the PR before merge.
> 
> **Testing gaps (review-level, since they span multiple files):**
> 
> 1. **End-to-end precedence rule.** `DeltaPropertyMapperTest.mergeDerivedPropertiesWinOnConflict` pins the new "derived wins over client" precedence at the unit level. There is no integration assertion that a request with `properties: {"delta.feature.catalogManaged": "client-override"}` still surfaces as `"supported"` in the stored properties through the wire boundary. One extra assertion in `SdkCreateTableTest.testCreateTableEndpoint` would pin this end-to-end, which matters because the precedence flipped between commits (`ae74351` → `e5693da`) and could regress silently.
> 2. **`CREATE_TABLE` SpEL branch coverage from DRC.** The shared expression has the EXTERNAL arm `(no_overlap_with_data_securable && (external_location == null || authorize(... OWNER, CREATE_EXTERNAL_TABLE)))`. The new DRC integration tests cover `external_location == null` and the implicit no-overlap path; they do not cover (a) EXTERNAL with a path that overlaps a data securable, (b) EXTERNAL with a resolved external_location and OWNER, (c) EXTERNAL with a resolved external_location and `CREATE_EXTERNAL_TABLE`. Since the SpEL constant is now shared with `TableService.createTable`, those branches may already be covered by existing UC REST tests — could you confirm which test pins each branch (or add a small DRC-side test) so we know §2g full conditional branch coverage is met?
> 3. **MANAGED-via-DRC OWNER semantics.** With the May 1 change, MANAGED now skips `initializeHierarchicalAuthorization` on the DRC path while UC REST `TableService.createTable` still calls it unconditionally. If the staging caller and the createTable caller can be different identities, the two surfaces produce different OWNER rows for the same operation. There is no test that pins the intended behavior on the DRC side. A test where principal-A calls `createStagingTable` and principal-B calls `createTable` (with the appropriate grants) and we then assert who is OWNER on the resulting table would make the divergence with UC REST deliberate and regression-proof.

Thanks @tdas. Inline comments (class doc, rename to `buildStoredProperties`, `Objects.requireNonNull` -> `ValidationUtils.checkNotNull`) addressed in latest commits.

**E2E precedence**: done end-to-end for both derivation sources. The canonical fixture now injects two conflicting client properties: `delta.feature.catalogManaged = "client-override"` (protocol-derived) and `delta.clusteringColumns = "[[\"wrong\"]]"` (domain-metadata-derived). The canonical request carries a matching `deltaClustering` block with `clusteringColumns=[["id"]]`, and `managedProtocol` now includes CLUSTERING in writerFeatures. Assertions in `SdkCreateTableTest:testCreateTableEndpoint` pin that the stored values are `"supported"` and `"[[\"id\"]]"` respectively.

**SpEL branches**: shared SpEL (`AuthorizeExpressions.CREATE_TABLE` used by both surfaces). All four EXTERNAL branches pinned via **UC REST** (which shares the same SpEL) in `SdkExternalLocationAccessControlTest:332-389`; DRC side covers the no-overlap-allowed branch in `SdkTableAccessControlCRUDTest:234-267`. Explicit DRC tests for the other three branches: happy to file as a follow-up if you'd like.

**OWNER semantics**: the divergence doesn't actually exist. `StagingTableRepository.commitStagingTable:147-152` rejects `callerId != staging.createdBy` with `PERMISSION_DENIED`, so staging-creator = createTable-caller on both surfaces. OWNER is the createTable-caller in both. Updated `DeltaRestCatalogService.createTable` Javadoc to describe this accurately. The shared rejection logic is pinned via UC REST in `SdkStagingTableAccessControlTest.testManagedTableCreationByDifferentUserShouldFail:189-211`; DRC `createTable` routes through the same `commitStagingTable` call, so the rejection applies identically there. No DRC-entry test for this rejection today; happy to add one in the follow-up PR alongside the explicit DRC SpEL branch tests.

tdas (approved):
[AI Assisted] LGTM. Thanks for addressing the doc/naming/null-handling and pinning the derived-wins precedence end-to-end with the `client-override` and `[["wrong"]]` assertions in `testCreateTableEndpoint`. Verified the §9c MANAGED-OWNER answer against the local clone: `StagingTableRepository.commitStagingTable` rejects mismatched `callerId` so staging-creator and createTable-caller are always the same identity, no divergence with UC REST possible.

Non-blocking follow-up: the PR description still has two stale claims after the May 1 changes — "EXTERNAL skips the contract -- UC mirrors the wire payload" (EXTERNAL still goes through the projection now) and "client-supplied properties win on conflict" (precedence was flipped to derived-wins). Please update before merge so the description matches the code, and per @TimothyW553's earlier ask, mention that this also tightens the existing UC REST `POST /tables` MANAGED path to require `properties[delta.UCTableId]`.

PR #1518, opened 2026-04-24
Add Spark 4.2 compatibility for UCProxy

murali-db:
## Summary
- handle the Spark 4.2 `CatalogStorageFormat` ABI change in the UC Spark connector without breaking Spark 4.0 or 4.1
- add a focused compatibility test for the `CatalogStorageFormat` construction path used by `UCProxy`
- add Spark 4.2 snapshot resolver and non-blocking CI coverage for the new connector path

## Test plan
- [x] `sbt -DsparkVersion=4.0.0 -J-Xmx2G integrationTests/test:compile \"spark/testOnly io.unitycatalog.spark.CatalogStorageFormatCompatSuite\"`
- [x] `sbt -DsparkVersion=4.1.0 -J-Xmx2G integrationTests/test:compile \"spark/testOnly io.unitycatalog.spark.CatalogStorageFormatCompatSuite\"`
- [x] `sbt -DsparkVersion=4.2.0-SNAPSHOT \"show resolvers\"`
- [ ] let the non-blocking Spark 4.2 snapshot CI job validate the full snapshot dependency path end-to-end


.github/workflows/unit-tests.yml
```diff
@@ -32,18 +32,28 @@ jobs:
             # This is the primary combo: runs ALL tests (unit, spark, license).
             # Other combos run only spark connector tests since unit tests and
             # license checks don't depend on Spark/Delta versions.
+            run-compile-only: false
```
openinx: with this flag, will it automatically skip the `spark 4.0` and `spark 4.1` compilation ?  Actually,  I don't know what's the `run-compile-ony` meaning ... Lol

.github/workflows/unit-tests.yml
```diff
@@ -32,18 +32,28 @@ jobs:
...
             non-blocking: false
           - spark-version: "4.1.0"
             delta-version: "master-SNAPSHOT"
+            run-compile-only: false
             run-only-spark-tests: true
             # Cross-version testing with Delta master is informational; failures
             # indicate upcoming incompatibilities but should not block PRs.
             non-blocking: true
+          - spark-version: "4.2.0-SNAPSHOT"
+            delta-version: "n/a"
```
openinx: what will be the expected delta-version ?

.github/workflows/unit-tests.yml
```diff
@@ -32,18 +32,28 @@ jobs:
...
+            run-compile-only: true
+            run-only-spark-tests: false
```
openinx: what does this flag mean ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -439,6 +439,33 @@ object UCSingleCatalog {
   val LOAD_DELTA_CATALOG = ThreadLocal.withInitial[Boolean](() => true)
   val DELTA_CATALOG_LOADED = ThreadLocal.withInitial[Boolean](() => false)
 
+  // Spark 4.2 adds a `serdeName` field to CatalogStorageFormat.copy. Dispatch through
+  // reflection so one connector jar can keep working across Spark 4.0/4.1/4.2 runtimes.
+  private[spark] def catalogStorageFormatWithLocation(
```
openinx: okay,  seems like it's a compability issue between spark 4.2 and older spark versions.  will it be better for us to introduce the shim layer to solve the general compability issue ? I assume as time goes, we will support more and more spark versions, and I won't expect to have the hack code everywhere if more and more incompabilitiy issue happen.

.github/workflows/unit-tests.yml
murali-db: yeah, the name was confusing. i renamed it to `validation-mode` so the matrix reads as `all-tests`, `spark-tests`, or `spark-compile-only`.

this does not skip spark 4.0 / 4.1 validation: spark 4.0 still runs the full compile/test/license path, and spark 4.1 still runs spark connector tests. only the spark 4.2 row uses `spark-compile-only`.

.github/workflows/unit-tests.yml
murali-db: there is intentionally no delta version for this row right now. delta does not publish a spark 4.2 artifact yet, so i changed this from `n/a` to an empty `delta-version` and the job name now renders as just `spark 4.2.0-snapshot`.

that row compiles the uc spark connector against spark 4.2 without pulling in the delta test dependency.

.github/workflows/unit-tests.yml
murali-db: yep, agreed this needed a clearer name. i replaced the boolean with `validation-mode`.

`all-tests` is the existing full validation, `spark-tests` is the existing spark-only test path, and `spark-compile-only` is only for the spark 4.2 row until delta has a spark 4.2 artifact we can test against.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
murali-db: good call. i moved this into a small `SparkCatalogCompatibility` shim so the reflection is isolated there instead of living inline in `UCSingleCatalog`.

`UCSingleCatalog` now just calls the shim, and the focused test moved with it to `SparkCatalogCompatibilitySuite`. that should give us a cleaner place for future spark-version compatibility handling too.

.github/workflows/unit-tests.yml
openinx: okay,  so I think once we merged the spark 4.2 integration with delta,  then we can add the delta-version I think.

.github/workflows/unit-tests.yml
```diff
@@ -89,24 +95,36 @@ jobs:
...
         else
           echo "DELTA_VERSION=${{ matrix.delta-version }}" >> $GITHUB_ENV
         fi
 
+    - name: Compile spark connector against Spark 4.2
+      if: matrix.validation-mode == 'spark-compile-only'
+      run: |
+        build/sbt -DsparkVersion=${{ matrix.spark-version }} -J-Xmx2G \
+          "project spark" \
+          "set libraryDependencies := libraryDependencies.value.filterNot(_.organization == \"io.delta\")" \
```
openinx, line 114: will the compile still work ?  if we particularly filter out the `io.delta.*` from the classpath ?

seewishnew: yes -- this is also validated empirically: the [Non Blocking] Spark compile (Spark 4.2.0-SNAPSHOT) CI lane runs exactly this path and passes consistently.

This is because the connector's main sources (connectors/spark/src/main/) have no import io.delta.* at all, so removing the io.delta dependencies only drops test-scoped jars (delta-spark) and runtime jars (delta-kernel-api, delta-kernel-defaults, delta-storage, delta-kernel-unitycatalog) that the connector doesn't compile against directly.

.github/workflows/unit-tests.yml
```diff
@@ -20,8 +20,8 @@ jobs:
 
     name: >-
       ${{ matrix.non-blocking && '[Non Blocking]' || '' }}
-      ${{ matrix.run-only-spark-tests && 'Spark tests' || 'All tests' }}
-      (Spark ${{ matrix.spark-version }}, Delta ${{ matrix.delta-version }})
+      ${{ matrix.validation-mode == 'spark-compile-only' && 'Spark compile' || matrix.validation-mode == 'spark-tests' && 'Spark tests' || 'All tests' }}
```
tdas, line 26: can you add comments what the goal here is. this is hard to understand.

tdas (approved):
small comment otherwise lgtm. but get approval from others.

yili-db:
@murali-db why is the 4.2 only running as "Spark compile (Spark 4.2.0-SNAPSHOT)" in CI? Why not "Spark tests (Spark 4.2.0-SNAPSHOT, Delta master-SNAPSHOT)"

seewishnew:
@yili-db The UC integration tests depend on delta-spark (test-scoped) and delta-storage / delta-kernel-* (runtime-scoped). Delta does not yet publish artifacts built against Spark 4.2 — the cross-build is still stabilizing (see [delta-io/delta#6657](https://github.com/delta-io/delta/pull/6657)). Without a delta-spark jar compiled against Spark 4.2, the integration tests can't run.

The workaround is to compile the UC Spark connector against Spark 4.2 with io.delta.* dependencies filtered out (filterNot(_.organization == "io.delta")). This works because the connector's main sources (connectors/spark/src/main/) have zero import io.delta.* — the delta jars are only needed at test time and runtime, not compile time. This is validated empirically by the [Non Blocking] Spark compile (Spark 4.2.0-SNAPSHOT) CI lane which passes consistently.

.github/workflows/unit-tests.yml
openinx: Hi, @tdas , essentially,  it's compositing the job name, based on the various matrix variable.

.github/workflows/unit-tests.yml
```diff
@@ -32,18 +35,24 @@ jobs:
...
-            run-only-spark-tests: true
+            validation-mode: "spark-tests"
             # Cross-version testing with Delta master is informational; failures
             # indicate upcoming incompatibilities but should not block PRs.
             non-blocking: true
+          - spark-version: "4.2.0-SNAPSHOT"
+            delta-version: ""
+            validation-mode: "spark-compile-only"
+            # Delta does not yet publish a Spark 4.2 artifact. This mode applies
+            # only to Spark 4.2; Spark 4.0/4.1 continue to run compile + tests.
```
openinx, line 54: Pls file a separate issue to track, so that once we have an available delta-spark version, then we should add it to there , to enable the uc spark unit tests.

.github/workflows/unit-tests.yml
openinx: okay,  that's a bit hack, but I'm okay to use it as temporary approach to break the circular dependency.  that's why I ask to file a separate issue to track this in my last comment.

.github/workflows/unit-tests.yml
```diff
@@ -89,24 +98,36 @@ jobs:
...
+      run: |
+        build/sbt -DsparkVersion=${{ matrix.spark-version }} -J-Xmx2G \
+          "project spark" \
+          "set libraryDependencies := libraryDependencies.value.filterNot(_.organization == \"io.delta\")" \
+          compile
+
       # 'integrationTests' is not in root project, since we avoid running it via `./build/sbt test`,
       # Here, we explicitly compile `integrationTests` to catch compilation issues early.
     - name: Compile integration tests to verify compilation and checkstyle issues.
+      if: ${{ matrix.validation-mode != 'spark-compile-only' }}
```
openinx, line 120: this step is purely  compilation,   do we still need to exclude for spark 4.2 combos ?

.github/workflows/unit-tests.yml
```diff
@@ -89,24 +98,36 @@ jobs:
...
       run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION -J-Xmx2G integrationTests/test:compile
 
     - name: Build and verify the local deployment.
+      if: ${{ matrix.validation-mode != 'spark-compile-only' }}
```
openinx, line 124: pure compilation , same question as the above comment, do we need to explicitly exclude for spark 4.2 combos.

connectors/spark/src/main/scala/io/unitycatalog/spark/compat/SparkCatalogCompatibility.scala
```diff
@@ -0,0 +1,43 @@
...
+object SparkCatalogCompatibility {
+
+  /**
+   * Builds a CatalogStorageFormat across Spark versions.
+   *
+   * Spark 4.2 adds a `serdeName` field to CatalogStorageFormat.copy, while Spark 4.0 and
+   * Spark 4.1 do not have that parameter. Keep the ABI-specific reflection in this shim so
+   * connector code can use one compatibility entry point.
+   */
+  def catalogStorageFormatWithLocation(
```
openinx, line 19: Actually , I had a comment in @chenwang-databricks 's PR before,https://github.com/unitycatalog/unitycatalog/pull/1553/changes#r3251053300

my point is:  once we introduce the shim directories, then we should retire this dynamic method loading approach, instead we should use the correct shim approach. 

I'm fine to use this as a intermedia approach so that we can break the unitycatalog <-> delta circular dependency issue.

connectors/spark/src/test/scala/io/unitycatalog/spark/compat/SparkCatalogCompatibilitySuite.scala
```diff
@@ -0,0 +1,30 @@
+package io.unitycatalog.spark.compat
+
+import java.net.URI
+import java.util
+
+import org.junit.jupiter.api.Assertions.{assertEquals, assertNotNull, assertTrue}
+import org.junit.jupiter.api.Test
+
+class SparkCatalogCompatibilitySuite {
```
openinx, line 9: I think this can be removed now.  because if we can successfully run the compilation, then it should already call the correct `copy` method in spark, which the correctness should already been guaranteed by apache spark.  there is no much meaning for us to test it here.

build.sbt
```diff
@@ -24,6 +24,12 @@ lazy val scala213 = "2.13.17"
 lazy val deltaVersion = sys.props.getOrElse("deltaVersion", "4.1.0")
 lazy val sparkVersion = sys.props.getOrElse("sparkVersion", "4.0.0")
 lazy val sparkMajorMinorVersion = sparkVersion.split("\\.").take(2).mkString(".")
+lazy val sparkResolvers =
+  if (sparkVersion.endsWith("-SNAPSHOT")) {
+    Seq("Apache Snapshots" at "https://repository.apache.org/content/repositories/snapshots/")
```
openinx: I think we already add the maven repo in the global sbt repositories,https://github.com/unitycatalog/unitycatalog/pull/1518/changes#diff-c51168c2af3370181952a0a9b48c1e669e5fe67c13ccefe3261c32d434c71428R11  do we still need the spark specified `sparkResolvers` for snapshot here ?

build.sbt
```diff
@@ -60,6 +66,7 @@ lazy val commonSettings = Seq(
     ExclusionRule("org.slf4j", "slf4j-reload4j")
   ),
   resolvers += Resolver.mavenLocal,
+  resolvers ++= sparkResolvers,
```
openinx: I think we can just make all the modules use the same global resolvers, there seems like no necessary to separate the resolvers for different modules.

openinx (approved):
Generally looks good to,  No major blocker. Pls file the unitycatalog issue to track the remaining work once we break the circular dependencies if the delta + spark 4.2 is ready.

.github/workflows/unit-tests.yml
seewishnew: Good idea — I'll file a tracking issue for enabling the `delta-spark_4.2` test dependency once a released or snapshot-resolvable artifact is available. The issue will also cover removing the `uc.skipTestDeps` workaround that currently breaks the circular dependency for publish-only paths.

.github/workflows/unit-tests.yml
seewishnew: Agreed — the circular dep workaround is intentionally temporary. The follow-up Delta PR (after both #1518 and #1556 merge) will publish `delta-spark_4.2_2.13` to Maven Central with each Delta release, at which point `uc.skipTestDeps` can be removed and the 4.2 cell can use the released artifact directly. Will track in the issue above.

.github/workflows/unit-tests.yml
seewishnew: No exclusion needed for 4.2 here. The earlier "Build Delta SNAPSHOT from source" step publishes `delta-spark_4.2_2.13` to local Maven, and `$DELTA_VERSION` is set to that SNAPSHOT version. So `integrationTests/test:compile` resolves delta-spark from `~/.m2` and compiles successfully. This is already working in the shadow PR's CI.

.github/workflows/unit-tests.yml
seewishnew: Same answer as above — no exclusion needed. `$DELTA_VERSION` resolves to the locally-published SNAPSHOT, so `clean package publishM2` works for Spark 4.2 combos.

connectors/spark/src/main/scala/io/unitycatalog/spark/compat/SparkCatalogCompatibility.scala
seewishnew: Agree with the long-term direction. The reflection-based `SparkCatalogCompatibility` shim was introduced as a pragmatic bridge for this PR since the per-version shim source directories aren't in scope here. #1556 sets up `scala-shims/spark-{4.0,4.1,4.2}/` — once that merges, the reflection can be replaced with compile-time dispatch in a follow-up. That avoids scope overlap between #1518 and #1556.

connectors/spark/src/test/scala/io/unitycatalog/spark/compat/SparkCatalogCompatibilitySuite.scala
seewishnew: Fair point for compile-time shims — the compiler would enforce correctness. The test was added as a safety net specifically because the current approach uses reflection, which the compiler can't validate. I'd keep it while the reflection path is in use, and remove it when we switch to compile-time shims (per the discussion above). That way the cleanup is atomic.

build.sbt
seewishnew: The `sbt-config/repositories` file lists the GCS Maven Central mirror + Maven Central, but not Apache Snapshots. Adding Apache Snapshots globally would slow down resolution for non-snapshot builds (every module would check it and get 404s). `sparkResolvers` adds it conditionally — only when `sparkVersion` ends with `-SNAPSHOT` and only on spark-dependent modules. That said, the latency difference is small, so happy to consolidate if you prefer the simplicity.

build.sbt
seewishnew: The per-module `sparkResolvers` keeps Apache Snapshots scoped to modules that actually depend on Spark, and only when building for a SNAPSHOT version. Adding it globally would make server, client, hadoop, etc. all check Apache Snapshots on every resolution — minor latency hit but unnecessary. Happy to consolidate if you feel the simplicity outweighs the minor overhead.

.github/workflows/unit-tests.yml
seewishnew: https://github.com/unitycatalog/unitycatalog/issues/1564

seewishnew:
FYI — two additional Spark 4.2 compatibility issues were discovered during cross-build validation and fixed on #1556's branch (which is stacked on this PR). Documenting them here since they're in the same Spark 4.2 compat scope:

1. **`createNamespace` exception mapping** ([SPARK-55250](https://issues.apache.org/jira/browse/SPARK-55250)): Spark 4.2 removed the `namespaceExists` pre-check from `CreateNamespaceExec`, exposing that `UCProxy.createNamespace` throws raw `ApiException` instead of `NamespaceAlreadyExistsException`. Fixed in `UCSingleCatalog.scala`. No shim needed — safe on all Spark versions.

2. **`DESC EXTENDED` output change** ([SPARK-56678](https://issues.apache.org/jira/browse/SPARK-56678)): Spark 4.2 replaced the `Name` row with structured `Catalog`/`Namespace`/`Database`/`Table` rows. Fixed with a version-aware assertion in `DeltaManagedTableReadWriteTest.java`. Test-only change.

Both fixes are on `cross-build-infra` (#1556) and will flow through when that PR merges. See [#1556 comment](https://github.com/unitycatalog/unitycatalog/pull/1556#issuecomment-4510063292) for details.

build.sbt
seewishnew: Updated to simplify build.sbt and use blanket snapshots reference in the sbt resositories.

PR #1524, opened 2026-04-28
Serialize Spark connector type_json as StructField JSON

timothyw553:
## What changed

This PR fixes Spark table creation so UC stores full Spark `StructField` JSON in each column's `type_json`.

The expected `type_json` shape is:

```json
{"name":"<column_name>","type":<spark_data_type_json>,"nullable":<nullable>,"metadata":<metadata>}
```

The OSS UC server now validates this value during table creation and rejects invalid or incomplete `type_json` before saving the table.

## Why

Delta REST APIs use `type_json` to rebuild the full column schema.

A datatype-only value like `"integer"` is not enough because it loses the column name, nullability, and column metadata.

## Expected metadata

| SQL type | UC `type_name` | UC `type_text` | `type_json.type` |
| --- | --- | --- | --- |
| `TINYINT` | `BYTE` | `tinyint` | `"byte"` |
| `SMALLINT` | `SHORT` | `smallint` | `"short"` |
| `INT` | `INT` | `int` | `"integer"` |
| `BIGINT` | `LONG` | `bigint` | `"long"` |
| `FLOAT` | `FLOAT` | `float` | `"float"` |
| `DOUBLE` | `DOUBLE` | `double` | `"double"` |
| `DECIMAL(10,2)` | `DECIMAL` | `decimal(10,2)` | `"decimal(10,2)"` |
| `STRING` | `STRING` | `string` | `"string"` |
| `CHAR(10)` | `CHAR` | `char(10)` | `"char(10)"` |
| `VARCHAR(20)` | `STRING` | `varchar(20)` | `"varchar(20)"` |
| `BINARY` | `BINARY` | `binary` | `"binary"` |
| `BOOLEAN` | `BOOLEAN` | `boolean` | `"boolean"` |
| `DATE` | `DATE` | `date` | `"date"` |
| `TIMESTAMP` | `TIMESTAMP` | `timestamp` | `"timestamp"` |
| `TIMESTAMP_NTZ` | `TIMESTAMP_NTZ` | `timestamp_ntz` | `"timestamp_ntz"` |
| `ARRAY<INT>` | `ARRAY` | `array<int>` | `{"type":"array","elementType":"integer","containsNull":true}` |
| `MAP<STRING, INT>` | `MAP` | `map<string,int>` | `{"type":"map","keyType":"string","valueType":"integer","valueContainsNull":true}` |
| `STRUCT<a: INT, b: STRING>` | `STRUCT` | `struct<a:int,b:string>` | `{"type":"struct","fields":[{"name":"a","type":"integer","nullable":true,"metadata":{}},{"name":"b","type":"string","nullable":true,"metadata":{}}]}` |
| `VARIANT` | `VARIANT` | `variant` | `"variant"` |

The Spark integration test also checks `nullable`, `metadata`, and `partition_index`.

## Tests

Targeted tests passed:

```bash
build/sbt -J-Xmx2G "server/testOnly io.unitycatalog.server.sdk.tables.SdkTableCRUDTest" "server/testOnly io.unitycatalog.server.sdk.delta.SdkLoadTableTest" "server/testOnly io.unitycatalog.server.utils.ColumnUtilsTest" "spark/testOnly io.unitycatalog.spark.ParquetExternalTableReadWriteTest"
```


openinx:
Hi @TimothyW553 ,  the PR description is vague.  Could you pls give a matrix to help me understand, for every data type, what are the expected metadata that we required ? And how could we correctly validate all the metadata are correctly send ? do we also need to add the integration tests, so that even we run the tests against both local oss uc server and remote uc server, both should be failure if we don't have the fix.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -720,7 +722,7 @@ private class UCProxy(
       column.setNullable(field.nullable)
       column.setTypeText(field.dataType.catalogString)
       column.setTypeName(convertDataTypeToTypeName(field.dataType))
-      column.setTypeJson(field.dataType.json)
+      column.setTypeJson(toStructFieldJson(field))
```
openinx, line 686: Could you pls help to add the oss uc server-side validation, so that it will automatically reject the invalid json, and help us to do the e2e integration tests, where prevent us to do the regression in future ? 

Also cc @yili-db

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -564,6 +564,12 @@ private static final class ColSpec {
     private final String typeJson;
   }
 
+  private static String expectedStructFieldTypeJson(ColSpec spec) {
```
openinx: I think the test coverage is not enough,  I'd love to see for every datatype, what's the expected column type json.

yili-db: The `ColSpec.getTypeJson` is the expected type_json predefined by test, except that it needs to be wrapped by the following structure.

However @TimothyW553 this is now lacking test `nullable` and `metadata`. We need to have some columns to have nullable=false (is it possible?) and have some column level metadata. That also means the existing ColSpec is not enough to cover it. You need to slightly expand it in a smart way so that a partial ColSpec constructor sets both as null and empty, but a full scale ColSpec constructor can override both.

yili-db:
The fix looks good! Just some comment on testing.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
timothyw553: Expanded `testTableWithSupportedDataTypes` so each supported datatype has an explicit expected full `StructField` `type_json`.

timothyw553: Updated `ColSpec` with default and full constructors, then added `BOOLEAN NOT NULL` and a commented `STRING` column to verify `nullable=false` and non-empty metadata.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: Added OSS UC create-table validation to reject invalid or incomplete `type_json`, and added Spark end-to-end coverage to catch this regression.

timothyw553:
@yili-db Addressed the testing comment by adding coverage for `nullable=false` and non-empty column metadata in the existing integration test.

timothyw553:
@openinx Updated the PR description with the datatype matrix, added OSS server-side `type_json` validation, and expanded the Spark integration test to verify the expected metadata.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: Thanks.

openinx (approved):
LGTM, thanks @TimothyW553 .

yili-db (approved):

tdas (approved):
overall no concerns.

PR #1526, opened 2026-04-30
Strict column-shape validation for Delta createTable

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1526/files) to review incremental changes.
- [**stack/validate_columns**](https://github.com/unitycatalog/unitycatalog/pull/1526) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1526/files)]
  - [stack/column_case_ins](https://github.com/unitycatalog/unitycatalog/pull/1537) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1537/files/9cd3fd2bbd8fcbdeb23a20fcecadbe94b1306817..296a93b3f3307250a2d2bbfdb463ddcd682fe479)]

---------
**PR Checklist**

**Description of changes**
Strict column-shape validation for Delta createTable

Validate every required field on StructField, ArrayType, MapType, DecimalType, and PrimitiveType at the API boundary so non-SDK clients can't drive the server to 500s or persist malformed schemas. Adds a typed StructFieldMetadata wrapper in delta.yaml, drops `default: true` on contains-null / value-contains-null to match Kernel's parser, and pins the rules with raw-HTTP tests.

### DIFF
--- a/api/delta-docs/Models/StructField.md
+++ b/api/delta-docs/Models/StructField.md
@@ -6,7 +6,7 @@
 | **name** | **String** | Column name | [default to null] |
 | **type** | [**DeltaType**](DeltaType.md) |  | [default to null] |
 | **nullable** | **Boolean** | Whether this column can be null | [default to null] |
-| **metadata** | [**Map**](AnyType.md) | Additional column metadata (arbitrary key-value pairs, e.g., comment, delta.columnMapping.id) | [default to null] |
+| **metadata** | [**StructFieldMetadata**](StructFieldMetadata.md) |  | [default to null] |
 
 [[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
 
--- a/api/delta.yaml
+++ b/api/delta.yaml
@@ -1342,15 +1342,21 @@ components:
           type: boolean
           description: Whether this column can be null
         metadata:
-          type: object
-          description: Additional column metadata (arbitrary key-value pairs, e.g., comment, delta.columnMapping.id)
-          additionalProperties:
-            description: Metadata values can be strings, numbers, booleans, or nested objects
+          $ref: '#/components/schemas/StructFieldMetadata'
       required:
         - name
         - type
         - nullable
         - metadata
+
+    StructFieldMetadata:
+      type: object
+      description: Additional column metadata (arbitrary key-value pairs, e.g., comment, delta.columnMapping.id)
+      additionalProperties:
+        description: Metadata values can be strings, numbers, booleans, or nested objects
+      properties:
+        comment:
+          type: string
 
     DeltaType:
       description: |
@@ -1408,7 +1414,6 @@ components:
         contains-null:
           type: boolean
           description: Whether array elements can be null
-          default: true
       required:
         - element-type
         - contains-null
@@ -1425,7 +1430,6 @@ components:
         value-contains-null:
           type: boolean
           description: Whether map values can be null
-          default: true
       required:
         - key-type
         - value-type
--- a/api/delta.yaml
+++ b/api/delta.yaml
@@ -1355,8 +1355,14 @@ components:
       additionalProperties:
         description: Metadata values can be strings, numbers, booleans, or nested objects
       properties:
-        comment:
-          type: string
+        _placeholder:
+          type: object
+          description: |
+            Schema-only marker; never sent on the wire. Declared so the OpenAPI generator emits a
+            wrapper class for this schema instead of inlining `metadata` on the parent StructField
+            `new HashMap<>()` -- that would erase the "metadata absent" signal. Use the Map
+            accessors (e.g., `metadata.put("comment", ...)` and `metadata.get("comment")`) for 
+            actual metadata keys.
 
     DeltaType:
       description: |
--- a/clients/python/src/unitycatalog/delta/serde/delta_type_module.py
+++ b/clients/python/src/unitycatalog/delta/serde/delta_type_module.py
@@ -87,7 +87,7 @@ def _field_to_dict(sf):
         "name": sf.name,
         "type": _type_to_dict(sf.type),
         "nullable": sf.nullable,
-        "metadata": sf.metadata,
+        "metadata": sf.metadata.to_dict() if sf.metadata is not None else {},
     }
 
 
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
@@ -49,11 +49,6 @@ public final class DeltaCreateTableMapper {
     if (req.getLocation() == null || req.getLocation().isBlank()) {
       throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table location is required.");
     }
-    if (req.getColumns() == null
-        || req.getColumns().getFields() == null
-        || req.getColumns().getFields().isEmpty()) {
-      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table columns are required.");
-    }
     ColumnUtils.validateStructType(req.getColumns(), "columns");
 
     TableType tableType = toUCTableType(req.getTableType());
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
@@ -54,6 +54,11 @@ public final class DeltaCreateTableMapper {
         || req.getColumns().getFields().isEmpty()) {
       throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table columns are required.");
     }
+    if (req.getColumns().getFields().isEmpty()) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT, "Table must have at least one column.");
+    }
+    ColumnUtils.validateStructType(req.getColumns(), "columns");
 
     TableType tableType = toUCTableType(req.getTableType());
     if (req.getProtocol() == null) {
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
@@ -54,10 +54,6 @@ public final class DeltaCreateTableMapper {
         || req.getColumns().getFields().isEmpty()) {
       throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table columns are required.");
     }
-    if (req.getColumns().getFields().isEmpty()) {
-      throw new BaseException(
-          ErrorCode.INVALID_ARGUMENT, "Table must have at least one column.");
-    }
     ColumnUtils.validateStructType(req.getColumns(), "columns");
 
     TableType tableType = toUCTableType(req.getTableType());
--- a/server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
+++ b/server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
@@ -21,7 +21,9 @@ import io.unitycatalog.server.model.ColumnTypeName;
 import io.unitycatalog.server.persist.dao.ColumnInfoDAO;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.Locale;
 import java.util.Map;
 import java.util.function.Function;
@@ -240,8 +242,15 @@ public class ColumnUtils {
   public static void validateStructType(StructType structType, String path) {
     requireNonNull(structType.getFields(), path + ".fields");
     List<StructField> fields = structType.getFields();
+    Set<String> seen = new HashSet<>();
     for (int i = 0; i < fields.size(); i++) {
       validateStructField(fields.get(i), path + ".fields[" + i + "]");
+      String name = fields.get(i).getName();
+      if (!seen.add(name)) {
+        throw new BaseException(
+            ErrorCode.INVALID_ARGUMENT,
+            "Duplicate field name in " + path + ".fields: " + name);
+      }
     }
   }
 
@@ -463,6 +472,14 @@ public class ColumnUtils {
       List<ColumnInfo> columns, List<String> partitionColumns) {
     if (partitionColumns == null || partitionColumns.isEmpty()) {
       return;
+    }
+    Set<String> seen = new HashSet<>();
+    for (String partName : partitionColumns) {
+      if (!seen.add(partName)) {
+        throw new BaseException(
+            ErrorCode.INVALID_ARGUMENT,
+            "partition-columns contains duplicate entry: " + partName);
+      }
     }
     Map<String, ColumnInfo> columnsByName =
         columns.stream().collect(Collectors.toMap(ColumnInfo::getName, Function.identity()));
--- a/server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
+++ b/server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
@@ -3,7 +3,6 @@ package io.unitycatalog.server.utils;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonSetter;
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.unitycatalog.server.delta.model.ArrayType;
 import io.unitycatalog.server.delta.model.DecimalType;
@@ -154,27 +153,22 @@ public class ColumnUtils {
   }
 
   /**
-   * Validate that a UC column stores typeJson as Spark StructField JSON. The Spark connector and
-   * Delta REST APIs depend on name, type, nullable, and metadata being present in this payload.
+   * Validate that a UC column stores {@code typeJson} as Spark {@link StructField} JSON.
    */
   public static void validateTypeJson(ColumnInfo column) {
     if (column == null) {
       throw invalidTypeJson(null, "column cannot be null");
     }
-    JsonNode typeJson = parseTypeJson(column);
-    requireTypeJsonField(column, typeJson, StructField.JSON_PROPERTY_NAME);
-    requireTypeJsonField(column, typeJson, StructField.JSON_PROPERTY_TYPE);
-    requireTypeJsonField(column, typeJson, StructField.JSON_PROPERTY_NULLABLE);
-    JsonNode metadata = requireTypeJsonField(column, typeJson, StructField.JSON_PROPERTY_METADATA);
-    if (!metadata.isObject()) {
-      throw invalidTypeJson(column, "type_json.metadata must be a JSON object");
-    }
-
     StructField field;
     try {
       field = toStructField(column);
     } catch (IllegalStateException e) {
       throw invalidTypeJson(column, e.getMessage(), e);
+    }
+    try {
+      validateStructField(field, "type_json");
+    } catch (BaseException e) {
+      throw invalidTypeJson(column, e.getErrorMessage(), e);
     }
     if (!field.getName().equals(column.getName())) {
       throw invalidTypeJson(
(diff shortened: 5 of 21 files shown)


api/delta-docs/Models/ArrayType.md
```diff
@@ -4,7 +4,7 @@
 | Name | Type | Description | Notes |
 |------------ | ------------- | ------------- | -------------|
 | **element-type** | [**DeltaType**](DeltaType.md) |  | [default to null] |
-| **contains-null** | **Boolean** | Whether array elements can be null | [default to true] |
+| **contains-null** | **Boolean** | Whether array elements can be null | [default to null] |
```
openinx, line 7: Are the boolean allowed to be null ?  or it's just optional field ?  actually, it makes some kind of confusion for me to understand. 

If it's a required field, then it could be `true` or `false`. 
and if it's a optional field, then  it can be `null`, `true`, `false` ?

api/delta-docs/Models/StructField.md
```diff
@@ -6,7 +6,7 @@
 | **name** | **String** | Column name | [default to null] |
 | **type** | [**DeltaType**](DeltaType.md) |  | [default to null] |
 | **nullable** | **Boolean** | Whether this column can be null | [default to null] |
-| **metadata** | [**Map**](AnyType.md) | Additional column metadata (arbitrary key-value pairs, e.g., comment, delta.columnMapping.id) | [default to null] |
+| **metadata** | [**StructFieldMetadata**](StructFieldMetadata.md) |  | [default to null] |
```
openinx, line 9: Could you give some context about why we change this ?

yili-db: This is to workaround an OpenAPI code generator bug that if it remains a simple `map` then it's never null even if not set. It's initialized as `{}` even if the json string doesn't have `metadata={}`.
We want it to be null when not set so we can validate if it's set or not to enforce the real `required` syntax.
Only promoting it to a real type can do this.

api/delta-docs/Models/ArrayType.md
yili-db: Generated md files are not the source of truth. Only the yaml file is.
These fields are **required**. Setting a default value doesn't make much sense so I removed the default `true`.
Even for a required field, it can still be null because current OpenAPI deser doesn't enforce that. We need to validate. The default `true` would prevent us from checking if it's set or not.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -231,6 +206,87 @@ public static String toTypeJson(StructField field) {
...
+    validateType(map.getKeyType(), path + ".key-type");
+    validateType(map.getValueType(), path + ".value-type");
+  }
+
+  private static void validateDecimalType(DecimalType decimal, String path) {
+    requireNonNull(decimal.getPrecision(), path + ".precision");
+    requireNonNull(decimal.getScale(), path + ".scale");
+    int precision = decimal.getPrecision();
+    int scale = decimal.getScale();
+    if (precision < 0 || precision > 38) {
```
timothyw553, line 295: this passes if precision = 0, should this be < 1?

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -231,6 +206,87 @@ public static String toTypeJson(StructField field) {
...
 
+  /**
+   * Validate that a {@link StructType} is well-formed against the Delta spec's wire shape.
+   */
+  public static void validateStructType(StructType structType, String path) {
+    requireNonNull(structType.getFields(), path + ".fields");
+    List<StructField> fields = structType.getFields();
+    Set<String> seen = new HashSet<>();
+    for (int i = 0; i < fields.size(); i++) {
+      validateStructField(fields.get(i), path + ".fields[" + i + "]");
```
timothyw553: should we add `requireNonNull(fields.get(i), ...)`?

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
yili-db: It is copied from https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/types/DecimalType.java#L39

api/delta-docs/Models/ArrayType.md
tdas: no way to fix the generated doc? the doc is confusing. 
The doc does not say required or not either.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -232,6 +207,89 @@ public static String toTypeJson(StructField field) {
...
+    if (scale < 0 || scale > precision) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT,
+          String.format("%s.scale must be in [0, precision=%d], got %d.", path, precision, scale));
+    }
+  }
+
+  private static void validatePrimitiveType(PrimitiveType primitive, String path) {
+    requireNonNull(primitive.getType(), path + ".type");
+  }
```
tdas, line 319: [AI Assisted]

`validatePrimitiveType` only null-checks the discriminator; it does not reject unsupported primitive *values*. The `RawCreateTableTest` "void" case passes only because `toColumnInfos` → `resolveColumnTypeName` throws downstream. That couples the validator's completeness to the converter being run right after -- if any future call site does pre-flight validation alone, bad primitives silently pass.

Also, the downstream error ("Unsupported Delta primitive type: void") carries no path context (no `columns.fields[0].type`), unlike every other validation error in this file.

Consider folding the resolver into the validator and wrapping with the path:

```java
private static void validatePrimitiveType(PrimitiveType primitive, String path) {
  requireNonNull(primitive.getType(), path);
  try {
    resolveColumnTypeName(primitive);
  } catch (BaseException e) {
    throw new BaseException(
        ErrorCode.INVALID_ARGUMENT,
        path + ": " + e.getErrorMessage(), e);
  }
}
```

Separately: `requireNonNull(primitive.getType(), path + ".type")` produces a doubled `…type.type is required.` message because the caller (`validateType` at line 248) already passes a path ending in `.type`. The `.type` on a `PrimitiveType` is the discriminator, not a sub-field a user wrote -- dropping the suffix here gives cleaner errors.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -232,6 +207,89 @@ public static String toTypeJson(StructField field) {
...
+      if (!seen.add(name)) {
+        throw new BaseException(
+            ErrorCode.INVALID_ARGUMENT,
+            "Duplicate field name in " + path + ".fields: " + name);
+      }
+    }
+  }
+
+  private static void validateStructField(StructField field, String path) {
+    requireNonNull(field.getName(), path + ".name");
```
tdas: [AI Assisted]

`requireNonNull(field.getName(), …)` rejects `null` but not `""` or whitespace. There's no `ValidationUtils.validateSqlObjectName(...)` for column names (that runs only on the table name), so an empty/blank column name reaches `toColumnInfo` and can land in the DB.

```java
String name = field.getName();
if (name == null || name.isBlank()) {
  throw new BaseException(ErrorCode.INVALID_ARGUMENT, path + ".name is required.");
}
```

Worth a raw-HTTP test case alongside this fix.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -232,6 +207,89 @@ public static String toTypeJson(StructField field) {
...
+    Set<String> seen = new HashSet<>();
+    for (int i = 0; i < fields.size(); i++) {
+      String fieldPath = path + ".fields[" + i + "]";
+      requireNonNull(fields.get(i), fieldPath);
+      validateStructField(fields.get(i), fieldPath);
+      String name = fields.get(i).getName();
+      if (!seen.add(name)) {
+        throw new BaseException(
+            ErrorCode.INVALID_ARGUMENT,
+            "Duplicate field name in " + path + ".fields: " + name);
```
tdas: [AI Assisted]

Nit: the duplicate-field error message omits the offending index (`Duplicate field name in columns.fields: id`). Every other path in this file is consistent about indices (`columns.fields[1].type is required.`). For wide schemas the index helps:

```java
throw new BaseException(
    ErrorCode.INVALID_ARGUMENT,
    "Duplicate field name in " + path + ".fields[" + i + "]: " + name);
```

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -232,6 +207,89 @@ public static String toTypeJson(StructField field) {
...
+    validateType(map.getKeyType(), path + ".key-type");
+    validateType(map.getValueType(), path + ".value-type");
+  }
+
+  private static void validateDecimalType(DecimalType decimal, String path) {
+    requireNonNull(decimal.getPrecision(), path + ".precision");
+    requireNonNull(decimal.getScale(), path + ".scale");
+    int precision = decimal.getPrecision();
+    int scale = decimal.getScale();
+    if (precision < 0 || precision > 38) {
```
tdas, line 295: [AI Assisted]

On TimothyW553's `precision = 0` question: you cited Kernel's `DecimalType` as the source. Worth a one-liner in the code so the next reviewer doesn't ask again:

```java
if (precision < 0 || precision > 38) {
  // Matches io.delta.kernel.types.DecimalType bounds (precision=0 is allowed there
  // even though it's degenerate).
```

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -19,7 +20,9 @@
 import io.unitycatalog.server.persist.dao.ColumnInfoDAO;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.Locale;
```
tdas, line 26: [AI Assisted]

Nit: `import java.util.Locale;` is out of alphabetical order with the surrounding `HashSet`, `List`, `Set`, `Map` imports.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
```diff
@@ -54,6 +54,7 @@ public static Result toCreateTable(String catalog, String schema, CreateTableReq
         || req.getColumns().getFields().isEmpty()) {
       throw new BaseException(ErrorCode.INVALID_ARGUMENT, "Table columns are required.");
     }
+    ColumnUtils.validateStructType(req.getColumns(), "columns");
```
tdas, line 52: [AI Assisted]

The mapper does its own null/empty check on `req.getColumns().getFields()` (lines 52-56) *and then* calls `ColumnUtils.validateStructType(req.getColumns(), "columns")`, which would also catch the null-fields case. Minor duplication.

If `validateStructType` rejected empty `fields` with `path + ".fields cannot be empty."`, the mapper could drop its own check and you'd centralize the rules in one place.

tdas (commented):
[AI Assisted]

Reviewed end-to-end against a local checkout of `stack/validate_columns`. The core idea (drop `default: true`, recursively validate the wire shape, pin with raw-HTTP tests) is solid and well-targeted at the non-SDK-client gap. The `RawCreateTableTest` consolidation is a good pattern. Comments below are correctness sharpening and one design smell around `StructFieldMetadata` -- nothing blocking.

**Testing gaps worth filling in this PR or a follow-up:**

1. No raw-HTTP test for `DecimalType` validation: `precision < 0`, `precision > 38`, `scale > precision`, missing `precision`/`scale`. The bounds branches in `validateDecimalType` are currently uncovered end-to-end.
2. No raw-HTTP test for `MapType` missing `key-type` / `value-type` (only `value-contains-null` is covered).
3. No raw-HTTP test for a duplicate field name *inside a nested struct*. Since `validateStructType` recurses, a nested-dup pin would prove the recursion handles uniqueness correctly per-level.
4. The `req.getColumns() == null` branch in `DeltaCreateTableMapper.toCreateTable` (no `columns` key at all) is not pinned by a raw-HTTP test -- the empty-list case is, but they hit different branches.
5. No explicit Python test asserting `StructField.metadata` is coerced to `StructFieldMetadata` after `from_json` (the change in `_field_to_dict` relies on this coercion).

**What looks good:**

- `RawCreateTableTest` is exactly the right shape -- the SDK can't construct these payloads, and the per-case comments tie each assertion back to a specific bug class (e.g. PR #1524).
- Validation error paths (`columns.fields[0].type.element-type.fields[0].name is required.`) give clients a precise locator straight from the message.
- The `testDeserOmittedContainsNullLeavesNull` rename + intent comment is the right way to pin a behavior that flips with a generator-config change.

api/delta.yaml
```diff
@@ -1353,16 +1353,22 @@ components:
...
         - metadata
 
+    StructFieldMetadata:
+      type: object
+      description: Additional column metadata (arbitrary key-value pairs, e.g., comment, delta.columnMapping.id)
+      additionalProperties:
+        description: Metadata values can be strings, numbers, booleans, or nested objects
+      properties:
+        comment:
+          type: string
```
tdas: [AI Assisted]

`StructFieldMetadata` declares both `additionalProperties` *and* a typed `comment` property. On the server side this becomes a class that extends `HashMap<String, Object>` *and* has a typed `comment` getter/setter. Jackson then ignores the typed accessor for serialization (the Map view wins), which forces tests to do `metadata.put("comment", …)` rather than `metadata.setComment(…)` -- the test fixture comments call this out explicitly (`ColumnUtilsTest.java:325-326`, `DeltaModelSerializationTest.java:578-586`).

The typed `comment` field is effectively a footgun: the generated SDK exposes a setter that the wire silently ignores. If the only motivation for promoting `StructFieldMetadata` to a wrapper is to defeat the generator's `new HashMap<>()` default (per your reply to openinx), consider dropping the `comment` property and keeping it as a bare additionalProperties wrapper -- one source of truth, no dual-write hazard, and the `meta(...)` helper + lengthy explanation comments in the tests can go away.

server/src/test/java/io/unitycatalog/server/sdk/delta/RawCreateTableTest.java
```diff
@@ -0,0 +1,297 @@
...
+            },
+            "contains-null": true
+          },
+          "nullable": true,
+          "metadata": {}
+        }""",
+        "columns.fields[0].type.element-type.fields[0].name is required.");
+
+    // -------- columns.fields is an empty list --------
+    assertRejected(staging, "", "Table columns are required.");
```
tdas: [AI Assisted]

Nit: this case actually exercises an *empty* `fields` array (the empty string substitutes into `"fields": [%s]` producing `"fields": []`), not the no-`columns`-key-at-all case. The error comes from `DeltaCreateTableMapper` (lines 52-56), not the new validator. The `req.getColumns() == null` branch is a separate code path -- worth pinning too.

clients/python/src/unitycatalog/delta/serde/delta_type_module.py
```diff
@@ -87,7 +87,7 @@ def _field_to_dict(sf):
         "name": sf.name,
         "type": _type_to_dict(sf.type),
         "nullable": sf.nullable,
-        "metadata": sf.metadata,
+        "metadata": sf.metadata.to_dict() if sf.metadata is not None else {},
```
tdas, line 90: [AI Assisted]

Nit: `sf.metadata.to_dict() if sf.metadata is not None else {}` relies on Pydantic having coerced `metadata={}` (passed from `_parse_field` as a plain dict) into a `StructFieldMetadata` instance. The existing `test_complex_schema` round-trip exercises it implicitly, but an explicit assertion on the metadata type after `from_json` would make the contract obvious to future maintainers and would catch a Pydantic-config regression.

api/delta.yaml
yili-db: Removed this useless `comment` but introduce a `_placeholder` object field here for the same effect without confusion.

api/delta-docs/Models/ArrayType.md
yili-db: New custom mustache templates are needed to introduce the explicit required message.
A post-process step can remove the useless `[default to null]`.
Created a stacked PR https://github.com/unitycatalog/unitycatalog/pull/1578

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -150,28 +154,29 @@ public static StructField toStructField(ColumnInfo column) {
...
-
     StructField field;
     try {
       field = toStructField(column);
     } catch (IllegalStateException e) {
       throw invalidTypeJson(column, e.getMessage(), e);
     }
+    if (format == DataSourceFormat.DELTA) {
+      try {
+        validateStructField(field, "type_json");
```
timothyw553, line 175: to my understanding `validateTypeJson` is shared with other non-Delta formats. so this nullable check, etc is not being check for non-Delta formats. so i think that makes non-Delta format requests go from 400 "bad request" to 500 "internal server error" on NullPointerException

api/delta.yaml
```diff
@@ -1342,16 +1342,28 @@ components:
...
         - nullable
         - metadata
 
+    StructFieldMetadata:
+      type: object
+      description: Additional column metadata (arbitrary key-value pairs, e.g., comment, delta.columnMapping.id)
+      additionalProperties:
+        description: Metadata values can be strings, numbers, booleans, or nested objects
+      properties:
+        _placeholder:
```
timothyw553, line 1358: right now a column that doesn't have `metadata` is rejected while `{}` is accepted, should we allow not sending it and just treat missing as `{}`?

yili-db: No. `metadata` is required, even if it's empty.
`_placeholder` is not meant to keep it non-empty, but instead trying to keep the entire class in code generation.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
yili-db: Added a shallow check for non-Delta.
Ideally non-Delta tables should have the same validation as Delta and share the same code. However this code is already stepping into Delta-only territory as the `StructField` was designed for Delta tables.
It would be nice if the non-Delta tables happen to share the same structure and be compatible, however they will have different primary types so this code can't be easily shared.

api/delta-docs/Models/ArrayType.md
openinx: Thanks @yili-db .

api/delta.yaml
openinx: okay, so this is mainly used for `StructFieldMetadata` to define the `null` case.

yili-db: Yes. We need this to be a `StructFieldMetadata`, instead of a simple map, so that it's initialized as `null` when not set.
Otherwise an empty map and not set are both present as empty map.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -232,11 +220,121 @@ public static String toTypeJson(StructField field) {
...
+
+  private static void validatePrimitiveType(PrimitiveType primitive, String path) {
+    // The PrimitiveType discriminator (the "type" string) IS the primitive name; the caller's
+    // path already ends in `.type`, so no further suffix here. Resolve to confirm the name is in
+    // the closed set we support; rethrow with the field path attached so the error mirrors the
+    // shape of every other validation message in this file (downstream toColumnInfos would
+    // otherwise throw without path context).
+    requireNonNull(primitive.getType(), path);
+    try {
+      resolveColumnTypeName(primitive);
```
openinx, line 315: QQ:  I read the `resolveColumnTypeName` impl, and there seems like has some types that we did not fully handled but defined in the `ColumnTypeName`.  What's the plan to handle those types ? 

* INTERVAL
* CHAR
* NULL
* USER_DEFINED_TYPE
* TABLE_TYPE

openinx (approved):
Generally looks good to me, left one question.  Thanks @yili-db

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
yili-db: It's documented here: 
https://github.com/unitycatalog/unitycatalog/blob/c8d4cd17a48ddffac1fdec556409208376815330/server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java#L337
In short, they are not supported in Delta. `ColumnTypeName` is defined by `all.yaml` which covers more than just Delta.

tdas (approved):

PR #1527, opened 2026-04-30
[Hadoop] Add UC Delta credential support to UCCredentialHadoopConfs

timothyw553:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1527/files) to review incremental changes.
- [**stack/uc-hadoop-drc-creds**](https://github.com/unitycatalog/unitycatalog/pull/1527) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1527/files)]

---------

**PR Checklist**

- [x] A description of the changes is added.
- [ ] Tests are added or updated for behavior changes.
- [x] This change does not fix a GitHub issue.
- [ ] Documentation is updated, if needed.

**Description of changes**

Adds UC Delta credential support to the Hadoop credential configuration flow.

This PR:
- lets `UCCredentialHadoopConfs.Builder` accept an initial Delta `StorageCredential` and a `DeltaTableRef`
- routes renewal through the UC Delta `TemporaryCredentialsApi` when `fs.unitycatalog.credentials.api=uc_delta`
- converts Delta storage credentials into the existing S3, ABFS, and GCS temporary credential shapes
- selects the matching storage credential by table location, using longest prefix match when the API returns multiple credentials

Tests cover builder validation, Delta config keys, credential conversion, prefix selection, and Delta API renewal.

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
@@ -1,12 +1,14 @@
 package io.unitycatalog.hadoop;
 
 import io.unitycatalog.client.auth.TokenProvider;
+import io.unitycatalog.client.delta.model.StorageCredential;
 import io.unitycatalog.client.internal.Preconditions;
 import io.unitycatalog.client.model.PathOperation;
 import io.unitycatalog.client.model.TableOperation;
 import io.unitycatalog.client.model.TemporaryCredentials;
 import io.unitycatalog.hadoop.internal.CredPropsUtil;
-import io.unitycatalog.hadoop.internal.UCHadoopConf;
+import io.unitycatalog.hadoop.internal.DeltaStorageCredentialUtil;
+import io.unitycatalog.hadoop.internal.UCHadoopConfConstants;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -55,6 +57,7 @@ public final class UCCredentialHadoopConfs {
 
     private TokenProvider tokenProvider;
     private TemporaryCredentials initialCredentials;
+    private StorageCredential initialStorageCredential;
     private boolean credentialRenewalEnabled = true;
     private boolean credentialScopedFsEnabled = true;
     private Configuration hadoopConf = new Configuration(false);
@@ -80,9 +83,30 @@ public final class UCCredentialHadoopConfs {
      * (Required) The initial temporary credentials vended by UC (AWS session credentials, GCP OAuth
      * token, or Azure SAS). Typically, allocated once by the job driver and propagated to all
      * worker nodes so that each worker reuses the same credential rather than vending a new one.
+     *
+     * <p>Mutually exclusive with {@link #initialStorageCredential(StorageCredential)}.
      */
     public Builder initialCredentials(TemporaryCredentials initialCredentials) {
+      Preconditions.checkState(
+          initialStorageCredential == null,
+          "initialStorageCredential is already set; cannot also set initialCredentials. The two"
+              + " select different credentials APIs (UC vs UC Delta).");
       this.initialCredentials = initialCredentials;
+      return this;
+    }
+
+    /**
+     * (Required for UC Delta) The initial UC Delta storage credential returned by the UC Delta
+     * table API.
+     *
+     * <p>Mutually exclusive with {@link #initialCredentials(TemporaryCredentials)}.
+     */
+    public Builder initialStorageCredential(StorageCredential initialStorageCredential) {
+      Preconditions.checkState(
+          initialCredentials == null,
+          "initialCredentials is already set; cannot also set initialStorageCredential. The two"
+              + " select different credentials APIs (UC vs UC Delta).");
+      this.initialStorageCredential = initialStorageCredential;
       return this;
     }
 
@@ -132,7 +156,7 @@ public final class UCCredentialHadoopConfs {
       Preconditions.checkArgument(name != null && !name.isEmpty(), "engine version name required");
       Preconditions.checkArgument(
           version != null && !version.isEmpty(), "engine version value for '%s' required", name);
-      engineVersionProps.put(UCHadoopConf.UC_ENGINE_VERSION_PREFIX + name, version);
+      engineVersionProps.put(UCHadoopConfConstants.UC_ENGINE_VERSION_PREFIX + name, version);
     }
 
     /**
@@ -142,7 +166,7 @@ public final class UCCredentialHadoopConfs {
      * @throws IllegalStateException if a required field is missing
      */
     public Map<String, String> buildForTable(String tableId, TableOperation tableOperation) {
-      validate();
+      validateUcApi();
       return withEngineVersionProps(
           CredPropsUtil.createTableCredProps(
               credentialRenewalEnabled,
@@ -157,13 +181,59 @@ public final class UCCredentialHadoopConfs {
     }
 
     /**
+     * Builds Hadoop properties for a UC Delta table's storage location.
+     *
+     * @param tableId the UC table UUID, used to scope the credential cache and filesystem cache
+     * @param catalog the UC catalog name
+     * @param schema the UC schema name
+     * @param table the UC table name
+     * @param location the table storage location used to select a returned storage credential
+     * @return unmodifiable map; empty if the scheme is unrecognized
+     * @throws IllegalArgumentException if a table identity field is missing
+     * @throws IllegalStateException if a required builder field is missing
+     */
+    public Map<String, String> buildForTable(
+        String tableId, String catalog, String schema, String table, String location) {
+      Preconditions.checkArgument(tableId != null && !tableId.isEmpty(), "tableId is required");
+      Preconditions.checkArgument(catalog != null && !catalog.isEmpty(), "catalog is required");
+      Preconditions.checkArgument(schema != null && !schema.isEmpty(), "schema is required");
+      Preconditions.checkArgument(table != null && !table.isEmpty(), "table is required");
+      Preconditions.checkArgument(location != null && !location.isEmpty(), "location is required");
+      validateUcDeltaApi();
+      TemporaryCredentials tempCreds =
+          DeltaStorageCredentialUtil.toTemporaryCredentials(initialStorageCredential);
+      Map<String, String> props =
+          CredPropsUtil.createTableCredProps(
+              credentialRenewalEnabled,
+              credentialScopedFsEnabled,
+              hadoopConf,
+              scheme,
+              catalogUri,
+              tokenProvider,
+              tableId,
+              DeltaStorageCredentialUtil.toTableOperation(initialStorageCredential.getOperation()),
+              tempCreds);
+      if (props.isEmpty()) {
+        return props;
+      }
+      Map<String, String> deltaProps = new HashMap<>(props);
+      deltaProps.put(
+          UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_KEY, Boolean.TRUE.toString());
+      deltaProps.put(UCHadoopConfConstants.UC_DELTA_CATALOG_KEY, catalog);
+      deltaProps.put(UCHadoopConfConstants.UC_DELTA_SCHEMA_KEY, schema);
+      deltaProps.put(UCHadoopConfConstants.UC_DELTA_TABLE_NAME_KEY, table);
+      deltaProps.put(UCHadoopConfConstants.UC_DELTA_LOCATION_KEY, location);
+      return withEngineVersionProps(Collections.unmodifiableMap(deltaProps));
+    }
+
+    /**
      * Builds Hadoop properties for an <em>external path</em>.
      *
      * @return unmodifiable map; empty if the scheme is unrecognized
      * @throws IllegalStateException if a required field is missing
      */
     public Map<String, String> buildForPath(String path, PathOperation pathOperation) {
-      validate();
+      validateUcApi();
       return withEngineVersionProps(
           CredPropsUtil.createPathCredProps(
               credentialRenewalEnabled,
@@ -186,11 +256,23 @@ public final class UCCredentialHadoopConfs {
       return Collections.unmodifiableMap(merged);
     }
 
-    private void validate() {
+    private void validateCommon() {
       Preconditions.checkState(
           !credentialRenewalEnabled || tokenProvider != null,
           "tokenProvider is required when credential renewal is enabled");
-      Preconditions.checkState(initialCredentials != null, "initialCredentials is required");
+    }
+
+    private void validateUcApi() {
+      validateCommon();
+      Preconditions.checkState(
+          initialCredentials != null, "initialCredentials is required for the UC credentials API");
+    }
+
+    private void validateUcDeltaApi() {
+      validateCommon();
+      Preconditions.checkState(
+          initialStorageCredential != null,
+          "initialStorageCredential is required for the UC Delta credentials API");
     }
   }
 }
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/AbfsVendedTokenProvider.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/AbfsVendedTokenProvider.java
@@ -1,11 +1,12 @@
 package io.unitycatalog.hadoop.internal.auth;
 
 import io.unitycatalog.client.model.AzureUserDelegationSAS;
-import io.unitycatalog.hadoop.internal.UCHadoopConf;
+import io.unitycatalog.hadoop.internal.UCHadoopConfConstants;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.azurebfs.extensions.SASTokenProvider;
 import org.apache.hadoop.util.Preconditions;
 
+/** Hadoop ABFS SAS token provider backed by Unity Catalog temporary credentials. */
 public class AbfsVendedTokenProvider extends GenericCredentialProvider implements SASTokenProvider {
   public static final String ACCESS_TOKEN_KEY = "fs.azure.sas.fixed.token";
 
@@ -18,21 +19,22 @@ public class AbfsVendedTokenProvider extends GenericCredentialProvider implement
 
   @Override
   public GenericCredential initGenericCredential(Configuration conf) {
-    if (conf.get(UCHadoopConf.AZURE_INIT_SAS_TOKEN) != null
-        && conf.get(UCHadoopConf.AZURE_INIT_SAS_TOKEN_EXPIRED_TIME) != null) {
+    if (conf.get(UCHadoopConfConstants.AZURE_INIT_SAS_TOKEN) != null
+        && conf.get(UCHadoopConfConstants.AZURE_INIT_SAS_TOKEN_EXPIRED_TIME) != null) {
 
-      String sasToken = conf.get(UCHadoopConf.AZURE_INIT_SAS_TOKEN);
+      String sasToken = conf.get(UCHadoopConfConstants.AZURE_INIT_SAS_TOKEN);
       Preconditions.checkNotNull(
           sasToken,
           "Azure SAS token not set, please check " + "'%s' in hadoop configuration",
-          UCHadoopConf.AZURE_INIT_SAS_TOKEN);
+          UCHadoopConfConstants.AZURE_INIT_SAS_TOKEN);
 
-      long expiredTimeMillis = conf.getLong(UCHadoopConf.AZURE_INIT_SAS_TOKEN_EXPIRED_TIME, 0L);
+      long expiredTimeMillis =
+          conf.getLong(UCHadoopConfConstants.AZURE_INIT_SAS_TOKEN_EXPIRED_TIME, 0L);
       Preconditions.checkState(
           expiredTimeMillis > 0,
           "Azure SAS token expired time must be greater than 0, please check '%s' in hadoop "
               + "configuration",
-          UCHadoopConf.AZURE_INIT_SAS_TOKEN_EXPIRED_TIME);
+          UCHadoopConfConstants.AZURE_INIT_SAS_TOKEN_EXPIRED_TIME);
 
       return GenericCredential.forAzure(sasToken, expiredTimeMillis);
     } else {
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
@@ -1,21 +1,18 @@
 package io.unitycatalog.hadoop.internal.auth;
 
 import io.unitycatalog.client.ApiException;
-import io.unitycatalog.client.api.TemporaryCredentialsApi;
-import io.unitycatalog.client.auth.TokenProvider;
-import io.unitycatalog.client.internal.ApiClientUtils;
 import io.unitycatalog.client.internal.Clock;
-import io.unitycatalog.client.model.GenerateTemporaryPathCredential;
-import io.unitycatalog.client.model.GenerateTemporaryTableCredential;
-import io.unitycatalog.client.model.PathOperation;
-import io.unitycatalog.client.model.TableOperation;
-import io.unitycatalog.client.model.TemporaryCredentials;
-import io.unitycatalog.hadoop.internal.UCHadoopConf;
+import io.unitycatalog.hadoop.internal.UCHadoopConfConstants;
 import io.unitycatalog.hadoop.internal.util.BoundedKeyedCache;
-import java.net.URI;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.util.Preconditions;
 
+/**
+ * Base class for Hadoop credential providers backed by Unity Catalog temporary credentials.
+ *
+ * <p>Subclasses expose cloud-specific provider interfaces while this class handles renewal and
+ * cache lookup.
+ */
 public abstract class GenericCredentialProvider {
   // The credential cache, for saving QPS to unity catalog server.
   static final BoundedKeyedCache<String, GenericCredential> globalCache;
@@ -32,43 +29,34 @@ public abstract class GenericCredentialProvider {
   private Configuration conf;
   private Clock clock;
   private long renewalLeadTimeMillis;
-  private URI ucUri;
-  private TokenProvider tokenProvider;
   private String credUid;
   private boolean credCacheEnabled;
 
   private volatile GenericCredential credential;
-  private volatile TemporaryCredentialsApi tempCredApi;
+  private volatile TempCredentialApi credentialApi;
 
   protected void initialize(Configuration conf) {
     this.conf = conf;
 
     // Use the test clock if one is intentionally configured for testing.
-    String clockName = conf.get(UCHadoopConf.UC_TEST_CLOCK_NAME);
+    String clockName = conf.get(UCHadoopConfConstants.UC_TEST_CLOCK_NAME);
     this.clock = clockName != null ? Clock.getManualClock(clockName) : Clock.systemClock();
 
     this.renewalLeadTimeMillis =
         conf.getLong(
-            UCHadoopConf.UC_RENEWAL_LEAD_TIME_KEY, UCHadoopConf.UC_RENEWAL_LEAD_TIME_DEFAULT_VALUE);
+            UCHadoopConfConstants.UC_RENEWAL_LEAD_TIME_KEY,
+            UCHadoopConfConstants.UC_RENEWAL_LEAD_TIME_DEFAULT_VALUE);
 
-    String ucUriStr = conf.get(UCHadoopConf.UC_URI_KEY);
-    Preconditions.checkNotNull(
-        ucUriStr, "'%s' is not set in hadoop configuration", UCHadoopConf.UC_URI_KEY);
-    this.ucUri = URI.create(ucUriStr);
-
-    // Initialize the UCTokenProvider.
-    this.tokenProvider = TokenProvider.create(conf.getPropsWithPrefix(UCHadoopConf.UC_AUTH_PREFIX));
-
-    this.credUid = conf.get(UCHadoopConf.UC_CREDENTIALS_UID_KEY);
+    this.credUid = conf.get(UCHadoopConfConstants.UC_CREDENTIALS_UID_KEY);
     Preconditions.checkState(
         credUid != null && !credUid.isEmpty(),
         "Credential UID cannot be null or empty, '%s' is not set in hadoop configuration",
-        UCHadoopConf.UC_CREDENTIALS_UID_KEY);
+        UCHadoopConfConstants.UC_CREDENTIALS_UID_KEY);
 
     this.credCacheEnabled =
         conf.getBoolean(
-            UCHadoopConf.UC_CREDENTIAL_CACHE_ENABLED_KEY,
-            UCHadoopConf.UC_CREDENTIAL_CACHE_ENABLED_DEFAULT_VALUE);
+            UCHadoopConfConstants.UC_CREDENTIAL_CACHE_ENABLED_KEY,
+            UCHadoopConfConstants.UC_CREDENTIAL_CACHE_ENABLED_DEFAULT_VALUE);
 
     // The initialized credentials passing-through the hadoop configuration.
     this.credential = initGenericCredential(conf);
@@ -92,28 +80,15 @@ public abstract class GenericCredentialProvider {
     return credential;
   }
 
-  protected TemporaryCredentialsApi temporaryCredentialsApi() {
-    // Retry is automatically handled by RetryingHttpClient when configured
-    // via fs.unitycatalog.request.retry*
-    if (tempCredApi == null) {
+  TempCredentialApi tempCredentialApi() {
+    if (credentialApi == null) {
       synchronized (this) {
-        if (tempCredApi == null) {
-          // Propagate engine version metadata into the User-Agent header so the UC server can
-          // trace which engine versions are calling. Engines register these via
-          // UCCredentialHadoopConfs.Builder#addEngineVersions, which writes them into the Hadoop
-          // configuration under UC_ENGINE_VERSION_PREFIX.
-          tempCredApi =
-              new TemporaryCredentialsApi(
-                  ApiClientUtils.create(
-                      ucUri,
-                      tokenProvider,
-                      UCHadoopConf.createRequestRetryPolicy(conf),
-                      conf.getPropsWithPrefix(UCHadoopConf.UC_ENGINE_VERSION_PREFIX)));
+        if (credentialApi == null) {
+          credentialApi = TempCredentialApi.create(conf);
         }
       }
     }
-
-    return tempCredApi;
+    return credentialApi;
   }
 
   private GenericCredential renewCredential() throws ApiException {
@@ -124,47 +99,12 @@ public abstract class GenericCredentialProvider {
         if (cached != null && !cached.readyToRenew(clock, renewalLeadTimeMillis)) {
           return cached;
         }
-        // Renew the credential and update the cache.
-        GenericCredential renewed = createGenericCredentials();
-        globalCache.put(credUid, renewed);
-        return renewed;
+        GenericCredential created = tempCredentialApi().createCredential();
+        globalCache.put(credUid, created);
+        return created;
       }
     } else {
-      return createGenericCredentials();
+      return tempCredentialApi().createCredential();
     }
-  }
-
-  private GenericCredential createGenericCredentials() throws ApiException {
-    TemporaryCredentialsApi tempCredApi = temporaryCredentialsApi();
-
-    // Generate the temporary credential via requesting UnityCatalog.
-    TemporaryCredentials tempCred;
-    String type = conf.get(UCHadoopConf.UC_CREDENTIALS_TYPE_KEY);
-    if (UCHadoopConf.UC_CREDENTIALS_TYPE_PATH_VALUE.equals(type)) {
-      String path = conf.get(UCHadoopConf.UC_PATH_KEY);
-      String pathOperation = conf.get(UCHadoopConf.UC_PATH_OPERATION_KEY);
-
-      tempCred =
-          tempCredApi.generateTemporaryPathCredentials(
-              new GenerateTemporaryPathCredential()
-                  .url(path)
-                  .operation(PathOperation.fromValue(pathOperation)));
-    } else if (UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE.equals(type)) {
-      String tableId = conf.get(UCHadoopConf.UC_TABLE_ID_KEY);
-      String tableOperation = conf.get(UCHadoopConf.UC_TABLE_OPERATION_KEY);
-
-      tempCred =
-          tempCredApi.generateTemporaryTableCredentials(
-              new GenerateTemporaryTableCredential()
-                  .tableId(tableId)
-                  .operation(TableOperation.fromValue(tableOperation)));
-    } else {
-      throw new IllegalArgumentException(
-          String.format(
-              "Unsupported unity catalog temporary credentials type '%s', please check '%s'",
-              type, UCHadoopConf.UC_CREDENTIALS_TYPE_KEY));
-    }
-
-    return new GenericCredential(tempCred);
   }
 }
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
@@ -1,0 +1,55 @@
+package io.unitycatalog.hadoop.internal.auth;
+
+import io.unitycatalog.client.ApiClient;
+import io.unitycatalog.client.ApiException;
+import io.unitycatalog.client.auth.TokenProvider;
+import io.unitycatalog.client.internal.ApiClientUtils;
+import io.unitycatalog.client.internal.Preconditions;
+import io.unitycatalog.hadoop.internal.UCHadoopConfConstants;
+import java.net.URI;
+import org.apache.hadoop.conf.Configuration;
+
+/**
+ * Creates internal Hadoop credential wrappers from Unity Catalog temporary credential APIs.
+ *
+ * <p>This is an adapter over SDK-generated API clients, not a generated API client itself.
+ */
+interface TempCredentialApi {
+  GenericCredential createCredential() throws ApiException;
+
+  static TempCredentialApi create(Configuration conf) {
+    String useDeltaCredentialsApiValue =
+        conf.get(UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_KEY);
+    if (useDeltaCredentialsApiValue != null) {
+      useDeltaCredentialsApiValue = useDeltaCredentialsApiValue.trim();
+    }
+    Preconditions.checkArgument(
+        useDeltaCredentialsApiValue == null
+            || "true".equalsIgnoreCase(useDeltaCredentialsApiValue)
+            || "false".equalsIgnoreCase(useDeltaCredentialsApiValue),
+        "Unsupported value '%s' for '%s', expected true or false",
+        useDeltaCredentialsApiValue,
+        UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_KEY);
+    boolean useDeltaCredentialsApi =
+        useDeltaCredentialsApiValue == null
+            ? UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_DEFAULT_VALUE
+            : Boolean.parseBoolean(useDeltaCredentialsApiValue);
+
+    String ucUriStr = conf.get(UCHadoopConfConstants.UC_URI_KEY);
+    Preconditions.checkNotNull(
+        ucUriStr, "'%s' is not set in hadoop configuration", UCHadoopConfConstants.UC_URI_KEY);
+
+    ApiClient apiClient =
+        ApiClientUtils.create(
+            URI.create(ucUriStr),
+            TokenProvider.create(conf.getPropsWithPrefix(UCHadoopConfConstants.UC_AUTH_PREFIX)),
+            UCHadoopConfConstants.createRequestRetryPolicy(conf),
+            conf.getPropsWithPrefix(UCHadoopConfConstants.UC_ENGINE_VERSION_PREFIX));
+    if (useDeltaCredentialsApi) {
+      return new UCDeltaTempCredentialApi(
+          conf, new io.unitycatalog.client.delta.api.TemporaryCredentialsApi(apiClient));
+    }
+    return new UCTempCredentialApi(
+        conf, new io.unitycatalog.client.api.TemporaryCredentialsApi(apiClient));
+  }
+}
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -1,9 +1,5 @@
 package io.unitycatalog.hadoop.internal;
 
-import static io.unitycatalog.hadoop.internal.UCHadoopConf.FS_AZURE_ACCOUNT_AUTH_TYPE_PROPERTY_NAME;
-import static io.unitycatalog.hadoop.internal.UCHadoopConf.FS_AZURE_ACCOUNT_IS_HNS_ENABLED;
-import static io.unitycatalog.hadoop.internal.UCHadoopConf.FS_AZURE_SAS_TOKEN_PROVIDER_TYPE;
-
 import io.unitycatalog.client.auth.TokenProvider;
 import io.unitycatalog.client.internal.Preconditions;
 import io.unitycatalog.client.model.AwsCredentials;
@@ -53,7 +49,7 @@ public class CredPropsUtil {
     }
 
     public T uri(String uri) {
-      builder.put(UCHadoopConf.UC_URI_KEY, uri);
+      builder.put(UCHadoopConfConstants.UC_URI_KEY, uri);
       return self();
     }
 
@@ -62,42 +58,42 @@ public class CredPropsUtil {
       // implementation. So let's add the prefix here.
       tokenProvider
           .configs()
-          .forEach((key, value) -> builder.put(UCHadoopConf.UC_AUTH_PREFIX + key, value));
+          .forEach((key, value) -> builder.put(UCHadoopConfConstants.UC_AUTH_PREFIX + key, value));
       return self();
     }
 
     public T uid(String uid) {
-      builder.put(UCHadoopConf.UC_CREDENTIALS_UID_KEY, uid);
+      builder.put(UCHadoopConfConstants.UC_CREDENTIALS_UID_KEY, uid);
       return self();
     }
 
     public T credentialType(String credType) {
       Preconditions.checkArgument(
-          UCHadoopConf.UC_CREDENTIALS_TYPE_PATH_VALUE.equals(credType)
-              || UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE.equals(credType),
+          UCHadoopConfConstants.UC_CREDENTIALS_TYPE_PATH_VALUE.equals(credType)
+              || UCHadoopConfConstants.UC_CREDENTIALS_TYPE_TABLE_VALUE.equals(credType),
           "Invalid credential type '%s', must be either 'path' or 'table'.",
           credType);
-      builder.put(UCHadoopConf.UC_CREDENTIALS_TYPE_KEY, credType);
+      builder.put(UCHadoopConfConstants.UC_CREDENTIALS_TYPE_KEY, credType);
       return self();
     }
 
     public T tableId(String tableId) {
-      builder.put(UCHadoopConf.UC_TABLE_ID_KEY, tableId);
+      builder.put(UCHadoopConfConstants.UC_TABLE_ID_KEY, tableId);
       return self();
     }
 
     public T tableOperation(TableOperation tableOp) {
-      builder.put(UCHadoopConf.UC_TABLE_OPERATION_KEY, tableOp.getValue());
+      builder.put(UCHadoopConfConstants.UC_TABLE_OPERATION_KEY, tableOp.getValue());
       return self();
     }
 
     public T path(String path) {
-      builder.put(UCHadoopConf.UC_PATH_KEY, path);
+      builder.put(UCHadoopConfConstants.UC_PATH_KEY, path);
       return self();
     }
 
     public T pathOperation(PathOperation pathOp) {
-      builder.put(UCHadoopConf.UC_PATH_OPERATION_KEY, pathOp.getValue());
+      builder.put(UCHadoopConfConstants.UC_PATH_OPERATION_KEY, pathOp.getValue());
       return self();
     }
 
@@ -189,8 +185,8 @@ public class CredPropsUtil {
   private static class AbfsPropsBuilder extends PropsBuilder<AbfsPropsBuilder> {
 
     AbfsPropsBuilder(boolean credScopedFsEnabled, Configuration hadoopConf) {
-      set(FS_AZURE_ACCOUNT_AUTH_TYPE_PROPERTY_NAME, "SAS");
-      set(FS_AZURE_ACCOUNT_IS_HNS_ENABLED, "true");
+      set(UCHadoopConfConstants.FS_AZURE_ACCOUNT_AUTH_TYPE_PROPERTY_NAME, "SAS");
+      set(UCHadoopConfConstants.FS_AZURE_ACCOUNT_IS_HNS_ENABLED, "true");
       set("fs.abfs.impl.disable.cache", "true");
       set("fs.abfss.impl.disable.cache", "true");
 
@@ -243,18 +239,19 @@ public class CredPropsUtil {
     AwsCredentials awsCred = tempCreds.getAwsTempCredentials();
     S3PropsBuilder builder =
         new S3PropsBuilder(credScopedFsEnabled, hadoopConf)
-            .set(UCHadoopConf.S3A_CREDENTIALS_PROVIDER, AWS_VENDED_TOKEN_PROVIDER_CLASS)
+            .set(UCHadoopConfConstants.S3A_CREDENTIALS_PROVIDER, AWS_VENDED_TOKEN_PROVIDER_CLASS)
             .uri(uri)
             .tokenProvider(tokenProvider)
             .uid(UUID.randomUUID().toString())
-            .set(UCHadoopConf.S3A_INIT_ACCESS_KEY, awsCred.getAccessKeyId())
-            .set(UCHadoopConf.S3A_INIT_SECRET_KEY, awsCred.getSecretAccessKey())
-            .set(UCHadoopConf.S3A_INIT_SESSION_TOKEN, awsCred.getSessionToken());
+            .set(UCHadoopConfConstants.S3A_INIT_ACCESS_KEY, awsCred.getAccessKeyId())
+            .set(UCHadoopConfConstants.S3A_INIT_SECRET_KEY, awsCred.getSecretAccessKey())
+            .set(UCHadoopConfConstants.S3A_INIT_SESSION_TOKEN, awsCred.getSessionToken());
 
     // For the static credential case, nullable expiration time is possible.
     if (tempCreds.getExpirationTime() != null) {
       builder.set(
-          UCHadoopConf.S3A_INIT_CRED_EXPIRED_TIME, String.valueOf(tempCreds.getExpirationTime()));
+          UCHadoopConfConstants.S3A_INIT_CRED_EXPIRED_TIME,
+          String.valueOf(tempCreds.getExpirationTime()));
     }
 
     return builder;
@@ -269,7 +266,7 @@ public class CredPropsUtil {
       TableOperation tableOp,
       TemporaryCredentials tempCreds) {
     return s3TempCredPropsBuilder(credScopedFsEnabled, hadoopConf, uri, tokenProvider, tempCreds)
-        .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE)
+        .credentialType(UCHadoopConfConstants.UC_CREDENTIALS_TYPE_TABLE_VALUE)
         .tableId(tableId)
         .tableOperation(tableOp)
         .build();
@@ -284,7 +281,7 @@ public class CredPropsUtil {
       PathOperation pathOp,
       TemporaryCredentials tempCreds) {
     return s3TempCredPropsBuilder(credScopedFsEnabled, hadoopConf, uri, tokenProvider, tempCreds)
-        .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_PATH_VALUE)
+        .credentialType(UCHadoopConfConstants.UC_CREDENTIALS_TYPE_PATH_VALUE)
         .path(path)
         .pathOperation(pathOp)
         .build();
@@ -315,12 +312,12 @@ public class CredPropsUtil {
             .uri(uri)
             .tokenProvider(tokenProvider)
             .uid(UUID.randomUUID().toString())
-            .set(UCHadoopConf.GCS_INIT_OAUTH_TOKEN, gcpToken.getOauthToken());
+            .set(UCHadoopConfConstants.GCS_INIT_OAUTH_TOKEN, gcpToken.getOauthToken());
 
     // For the static credential case, nullable expiration time is possible.
     if (tempCreds.getExpirationTime() != null) {
       builder.set(
-          UCHadoopConf.GCS_INIT_OAUTH_TOKEN_EXPIRATION_TIME,
+          UCHadoopConfConstants.GCS_INIT_OAUTH_TOKEN_EXPIRATION_TIME,
           String.valueOf(tempCreds.getExpirationTime()));
     }
 
@@ -336,7 +333,7 @@ public class CredPropsUtil {
       TableOperation tableOp,
       TemporaryCredentials tempCreds) {
     return gcsTempCredPropsBuilder(credScopedFsEnabled, hadoopConf, uri, tokenProvider, tempCreds)
-        .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE)
+        .credentialType(UCHadoopConfConstants.UC_CREDENTIALS_TYPE_TABLE_VALUE)
         .tableId(tableId)
         .tableOperation(tableOp)
         .build();
@@ -351,7 +348,7 @@ public class CredPropsUtil {
       PathOperation pathOp,
       TemporaryCredentials tempCreds) {
     return gcsTempCredPropsBuilder(credScopedFsEnabled, hadoopConf, uri, tokenProvider, tempCreds)
-        .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_PATH_VALUE)
+        .credentialType(UCHadoopConfConstants.UC_CREDENTIALS_TYPE_PATH_VALUE)
         .path(path)
         .pathOperation(pathOp)
         .build();
@@ -374,16 +371,18 @@ public class CredPropsUtil {
     AzureUserDelegationSAS azureSas = tempCreds.getAzureUserDelegationSas();
     AbfsPropsBuilder builder =
         new AbfsPropsBuilder(credScopedFsEnabled, hadoopConf)
-            .set(FS_AZURE_SAS_TOKEN_PROVIDER_TYPE, ABFS_VENDED_TOKEN_PROVIDER_CLASS)
+            .set(
+                UCHadoopConfConstants.FS_AZURE_SAS_TOKEN_PROVIDER_TYPE,
+                ABFS_VENDED_TOKEN_PROVIDER_CLASS)
             .uri(uri)
             .tokenProvider(tokenProvider)
             .uid(UUID.randomUUID().toString())
-            .set(UCHadoopConf.AZURE_INIT_SAS_TOKEN, azureSas.getSasToken());
+            .set(UCHadoopConfConstants.AZURE_INIT_SAS_TOKEN, azureSas.getSasToken());
 
     // For the static credential case, nullable expiration time is possible.
     if (tempCreds.getExpirationTime() != null) {
       builder.set(
-          UCHadoopConf.AZURE_INIT_SAS_TOKEN_EXPIRED_TIME,
+          UCHadoopConfConstants.AZURE_INIT_SAS_TOKEN_EXPIRED_TIME,
           String.valueOf(tempCreds.getExpirationTime()));
     }
 
@@ -399,7 +398,7 @@ public class CredPropsUtil {
       TableOperation tableOp,
       TemporaryCredentials tempCreds) {
     return abfsTempCredPropsBuilder(credScopedFsEnabled, hadoopConf, uri, tokenProvider, tempCreds)
-        .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE)
+        .credentialType(UCHadoopConfConstants.UC_CREDENTIALS_TYPE_TABLE_VALUE)
         .tableId(tableId)
         .tableOperation(tableOp)
         .build();
@@ -414,7 +413,7 @@ public class CredPropsUtil {
       PathOperation pathOp,
       TemporaryCredentials tempCreds) {
     return abfsTempCredPropsBuilder(credScopedFsEnabled, hadoopConf, uri, tokenProvider, tempCreds)
-        .credentialType(UCHadoopConf.UC_CREDENTIALS_TYPE_PATH_VALUE)
+        .credentialType(UCHadoopConfConstants.UC_CREDENTIALS_TYPE_PATH_VALUE)
         .path(path)
         .pathOperation(pathOp)
         .build();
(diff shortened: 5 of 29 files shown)


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/AbfsVendedTokenProvider.java
```diff
@@ -1,10 +1,10 @@
...
 
 import io.unitycatalog.client.model.AzureUserDelegationSAS;
-import io.unitycatalog.spark.UCHadoopConf;
+import io.unitycatalog.hadoop.UCHadoopConf;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.azurebfs.extensions.SASTokenProvider;
-import org.apache.hadoop.shaded.com.google.common.base.Preconditions;
+import org.apache.hadoop.util.Preconditions;
 
 public class AbfsVendedTokenProvider extends GenericCredentialProvider implements SASTokenProvider {
```
openinx, line 10: since we cannot remove the `public` modifier from the class, because we want to use the default dynamic class load mechanism to load this class. 

but I think we can add a comment say, this class is for internal usage, don't guarantee the API compability.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+  private static final String UC_CREDENTIAL_CACHE_MAX_SIZE =
+      "unitycatalog.credential.cache.maxSize";
+  private static final long UC_CREDENTIAL_CACHE_MAX_SIZE_DEFAULT = 1024;
+
+  static {
+    final long maxSize =
+        Long.getLong(UC_CREDENTIAL_CACHE_MAX_SIZE, UC_CREDENTIAL_CACHE_MAX_SIZE_DEFAULT);
+    globalCache =
+        Collections.synchronizedMap(
+            new LinkedHashMap<String, GenericCredential>(16, 0.75f, true) {
```
openinx: nit:  we can use the `<>` directly here.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+    this.clock = clockName != null ? Clock.getManualClock(clockName) : Clock.systemClock();
+
+    this.renewalLeadTimeMillis =
+        conf.getLong(
+            UCHadoopConf.UC_RENEWAL_LEAD_TIME_KEY, UCHadoopConf.UC_RENEWAL_LEAD_TIME_DEFAULT_VALUE);
+
+    String ucUriStr = conf.get(UCHadoopConf.UC_URI_KEY);
+    Objects.requireNonNull(
+        ucUriStr,
+        String.format("'%s' is not set in hadoop configuration", UCHadoopConf.UC_URI_KEY));
```
openinx: You can use the `io.unitycatalog.client.internal.Preconditions` directly, then you don't need to manually format the error message any more.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+    this.ucUri = URI.create(ucUriStr);
+
+    this.tokenProvider = TokenProvider.create(conf.getPropsWithPrefix(UCHadoopConf.UC_AUTH_PREFIX));
+
+    this.credUid = conf.get(UCHadoopConf.UC_CREDENTIALS_UID_KEY);
+    checkState(
```
openinx: Same as above.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+import io.unitycatalog.hadoop.UCHadoopConf;
+import java.net.URI;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import org.apache.hadoop.conf.Configuration;
+
+public abstract class GenericCredentialProvider {
```
openinx, line 16: Pls remove this `public` modifier, since we don't want any other external accessor to access it.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+      return createDeltaGenericCredentials();
+    }
+    if (UCHadoopConf.UC_CREDENTIALS_API_LEGACY_VALUE.equals(credentialApi)) {
+      return createLegacyGenericCredentials();
+    }
+    throw new IllegalArgumentException(
+        String.format(
+            "Unsupported unity catalog credential API '%s', please check '%s'",
+            credentialApi, UCHadoopConf.UC_CREDENTIALS_API_KEY));
+  }
```
openinx: Could you pls define an interface to fully separate the process for legacy and delta API to generate the concrete credentials ? 

I have a simple demo to show the simple interface here. 

```diff
diff --git a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
index b2ba26cd..72fce8eb 100644
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
@@ -1,4 +1,43 @@
 package io.unitycatalog.hadoop.credentials;
 
-public class GenericCredentialApi {
+import io.unitycatalog.client.ApiException;
+import io.unitycatalog.hadoop.UCHadoopConf;
+import org.apache.hadoop.conf.Configuration;
+
+public interface GenericCredentialApi {
+
+  GenericCredential createGenericCredentials() throws ApiException;
+
+  static GenericCredentialApi create(Configuration conf) {
+    String type = conf.get(UCHadoopConf.UC_CREDENTIALS_TYPE_KEY);
+    if (UCHadoopConf.UC_CREDENTIALS_API_DELTA_VALUE.equals(type)) {
+      return new DeltaGenericCredentialApi();
+    } else if (UCHadoopConf.UC_CREDENTIALS_API_LEGACY_VALUE.equals(type)) {
+      return new LegacyGenericCredentialApi();
+    } else {
+      throw new IllegalStateException("Unsupported credentials type: " + type);
+    }
+  }
+
+  /**
+   * The LegacyGenericCredentialApi implementation.
+   */
+  class LegacyGenericCredentialApi implements GenericCredentialApi {
+
+    @Override
+    public GenericCredential createGenericCredentials() throws ApiException {
+      return null;
+    }
+  }
+
+  /**
+   * The DeltaGenericCredentialApi implementation.
+   */
+  class DeltaGenericCredentialApi implements GenericCredentialApi {
+
+    @Override
+    public GenericCredential createGenericCredentials() throws ApiException {
+      return null;
+    }
+  }
 }
```

openinx: With this,  I think it will much clear for us to handle both the `legacy` and `delta` api to generate the temp credentials. essentially, we separate the code from the start place.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+
+      CredentialsResponse response =
+          deltaTemporaryCredentialsApi()
+              .getTemporaryPathCredentials(path, toUCDeltaCredentialOperation(pathOperation));
+      return GenericCredential.fromStorageCredential(
+          selectStorageCredential(path, response.getStorageCredentials()));
+    } else if (UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE.equals(type)) {
+      String catalog = requireConf(UCHadoopConf.UC_TABLE_CATALOG_KEY);
+      String schema = requireConf(UCHadoopConf.UC_TABLE_SCHEMA_KEY);
+      String table = requireConf(UCHadoopConf.UC_TABLE_NAME_KEY);
```
openinx: wait, for the new api, we will need the `catalog.schema.table` to generate the temp credential ? for the old api, i think we just use the table id. does it work from the e2e now ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+      String location = requireConf(UCHadoopConf.UC_TABLE_LOCATION_KEY);
```
openinx: for table credential, why do we need the `location` ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+
+  private StorageCredential selectStorageCredential(
+      String location, List<StorageCredential> storageCredentials) {
+    StorageCredential bestMatch = null;
+    for (StorageCredential credential : storageCredentials) {
+      if (credential != null
+          && credential.getPrefix() != null
+          && matchesCredentialPrefix(location, credential.getPrefix())
+          && (bestMatch == null
+              || credential.getPrefix().length() > bestMatch.getPrefix().length())) {
```
openinx: Hi, @yili-db ,  could you give more context, why for a single credential request, it will produce a list of multiple temp credentials ?

openinx: And those `if` conditions need to be very carefully tested in our test coverage,  otherwise it will be very easy for us get the following `checkState`  exception.  @TimothyW553

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+    String normalizedLocation = stripTrailingSlash(location);
+    String normalizedPrefix = stripTrailingSlash(prefix);
+    return !normalizedPrefix.isEmpty()
+        && (normalizedLocation.equals(normalizedPrefix)
+            || (normalizedLocation.startsWith(normalizedPrefix)
+                && normalizedLocation.charAt(normalizedPrefix.length()) == '/'));
+  }
+
+  private String stripTrailingSlash(String value) {
+    return value.endsWith("/") ? value.substring(0, value.length() - 1) : value;
```
openinx: what if there are multiple trailing slash ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
yili-db: This design of RPC is for future proofing so that IF in the future a table can have more than 1 storage location this would generate credentials for each one of them.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  private static void checkState(boolean expression, String message, Object... args) {
+    if (!expression) {
+      throw new IllegalStateException(String.format(message, args));
+    }
+  }
```
openinx: this can be removed now, since we already have the internal `Precondition`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/fs/CredScopedFileSystem.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Map;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.FilterFileSystem;
+
+/**
+ * Hadoop {@link FileSystem} wrapper that lets several UC credential scopes coexist in one Spark
+ * session without forcing one global filesystem cache entry per cloud scheme.
+ */
+public class CredScopedFileSystem extends FilterFileSystem {
```
openinx, line 52: any change for this file ?

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/credentials/BaseTokenProviderTest.java
```diff
@@ -38,15 +38,15 @@ public abstract class BaseTokenProviderTest<T extends GenericCredentialProvider>
   public void before() {
     clockName = UUID.randomUUID().toString();
     clock = Clock.getManualClock(clockName);
-    GenericCredentialProvider.globalCache.invalidateAll();
+    GenericCredentialProvider.globalCache.clear();
```
openinx, line 41: why change this ?

openinx (commented):
Hi @TimothyW553 ,  left comments.  One key ask is:  we don't have any test coverage for the delta api credentials generation ?

I think it will be good for us to parameterize one key for the config `fs.unitycatalog.credentials.api`,  so that we can make all the existing test coverage to apply both `legacy` and `delta` apis.  we definitely need the full coverage.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+import java.net.URI;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import org.apache.hadoop.conf.Configuration;
+
+public abstract class GenericCredentialProvider {
+  static final Map<String, GenericCredential> globalCache;
```
yili-db: Please comment on what is the key

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -0,0 +1,297 @@
...
+      case "READ":
+      case "PATH_READ":
+        return CredentialOperation.READ;
+      case "READ_WRITE":
+      case "PATH_READ_WRITE":
+        return CredentialOperation.READ_WRITE;
+      default:
+        throw new IllegalArgumentException(
+            String.format(
+                "Unsupported UC Delta Rest Catalog API credential operation '%s'. "
```
yili-db: Just UC Delta API

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/fs/CredScopedFileSystem.java
timothyw553: Kept only the JDK cache change because this module should not depend on Spark shaded Guava.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/credentials/BaseTokenProviderTest.java
timothyw553: Moved this test to package-private helpers so it does not depend on the cache implementation.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
yili-db: @TimothyW553 Let's do it in the simpler and safer way for now:
If there is only one credential, pick it anyway.
Otherwise, still do the matching.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredPropsUtil.java
```diff
@@ -0,0 +1,894 @@
...
+  // Keep these as strings so callers can build credential options without loading cloud SDKs.
+  private static final String ABFS_VENDED_TOKEN_PROVIDER_CLASS =
+      "io.unitycatalog.hadoop.credentials.AbfsVendedTokenProvider";
+  private static final String AWS_VENDED_TOKEN_PROVIDER_CLASS =
+      "io.unitycatalog.hadoop.credentials.AwsVendedTokenProvider";
+  private static final String GCS_VENDED_TOKEN_PROVIDER_CLASS =
+      "io.unitycatalog.hadoop.credentials.GcsVendedTokenProvider";
+  private static final String CRED_SCOPED_FILE_SYSTEM_CLASS =
+      "io.unitycatalog.hadoop.fs.CredScopedFileSystem";
+  private static final String CRED_SCOPED_FS_CLASS = "io.unitycatalog.hadoop.fs.CredScopedFs";
```
yili-db: You can no longer use the class.name?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCHadoopConf.java
```diff
@@ -60,8 +60,17 @@ private UCHadoopConf() {}
   // credentials are indexed by this key and are not reused across different jobs.
   public static final String UC_CREDENTIALS_UID_KEY = "fs.unitycatalog.credentials.uid";
 
+  // Key indicating which Unity Catalog credentials API should be used for renewal.
+  public static final String UC_CREDENTIALS_API_KEY = "fs.unitycatalog.credentials.api";
+  public static final String UC_CREDENTIALS_API_LEGACY_VALUE = "legacy";
```
yili-db: Let's not call it legacy yet. It remains the API for all other table formats.
Let's say "UC" and "UC_DELTA"

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/AbfsVendedTokenProvider.java
timothyw553: Added the internal-use compatibility comment on the public provider classes.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: Changed this to use the diamond operator.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: Changed this to use UC internal Preconditions.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: Changed this one to use UC internal Preconditions too.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: Removed the public modifier from GenericCredentialProvider.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: Split credential creation into separate API implementations.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: The location is used to choose the matching credential prefix.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: Implemented Yi's simpler rule: use one credential directly, otherwise match by location.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: Fixed this to remove multiple trailing slashes.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: Removed the local checkState helper.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFs.java
```diff
@@ -0,0 +1,13 @@
+package io.unitycatalog.hadoop.fs;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.DelegateToFileSystem;
+
+public class CredScopedFs extends DelegateToFileSystem {
```
yili-db, line 21: The class comment was removed. Why?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: Not full E2E in this PR; Hadoop tests verify catalog/schema/table is sent to UC Delta API.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: Added a comment that the cache key is fs.unitycatalog.credentials.uid.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: Renamed the messages to UC Delta API.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredPropsUtil.java
timothyw553: Changed these constants to use class.getName().

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCHadoopConf.java
timothyw553: Renamed the credential API values to UC and UC_DELTA.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFs.java
timothyw553: Restored the CredScopedFs class comment.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCHadoopConf.java
```diff
@@ -60,8 +60,17 @@ private UCHadoopConf() {}
   // credentials are indexed by this key and are not reused across different jobs.
   public static final String UC_CREDENTIALS_UID_KEY = "fs.unitycatalog.credentials.uid";
 
+  // Key indicating which Unity Catalog credentials API should be used for renewal.
+  public static final String UC_CREDENTIALS_API_KEY = "fs.unitycatalog.credentials.api";
+  public static final String UC_CREDENTIALS_API_UC_VALUE = "UC";
+  public static final String UC_CREDENTIALS_API_UC_DELTA_VALUE = "UC_DELTA";
```
openinx: seems like we don't have the upper case values ? how about use the consistent lower cases.  and the `uc` vs `uc_delta` names, are not explicit to show the difference i think. any thought ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
```diff
@@ -1,33 +1,51 @@
...
       "unitycatalog.credential.cache.maxSize";
   private static final long UC_CREDENTIAL_CACHE_MAX_SIZE_DEFAULT = 1024;
 
   static {
-    long maxSize = Long.getLong(UC_CREDENTIAL_CACHE_MAX_SIZE, UC_CREDENTIAL_CACHE_MAX_SIZE_DEFAULT);
-    globalCache = CacheBuilder.newBuilder().maximumSize(maxSize).build();
+    final long maxSize =
+        Long.getLong(UC_CREDENTIAL_CACHE_MAX_SIZE, UC_CREDENTIAL_CACHE_MAX_SIZE_DEFAULT);
+    globalCache =
+        Collections.synchronizedMap(
```
openinx: oh,  this is not good.  why changed this ?   the `synchronizedMap` concurrent and throughput is much lower than the `CacheBuilder`,  and all the requests from the same table will request the global lock from the `synchronizedMap`, which will easily be the bottleneck of the whole job. 

I assume you want to remove the spark dependency from the uc-hadoop module, yes, that's the correct direction.

openinx: I'm okay with this PR,  but we need to follow PR to introduce the relocated guava to address this explicit bottleneck,  could you pls file a separate issue for it ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
```diff
@@ -0,0 +1,199 @@
...
+import org.apache.hadoop.conf.Configuration;
+
+interface GenericCredentialApi {
+  GenericCredential createGenericCredentials() throws ApiException;
+
+  static GenericCredentialApi create(
+      Configuration conf,
+      Supplier<TemporaryCredentialsApi> temporaryCredentialsApi,
+      Supplier<io.unitycatalog.client.delta.api.TemporaryCredentialsApi>
+          deltaTemporaryCredentialsApi) {
```
openinx: I think you can directly design the `create` as `create(Configuration conf,   ApiClient client)`.  with this,  we don't even need to expose the `TemporaryCredentialsApi` to outside. 

And the purpose of abstracting this `GenericCredentialApi` to try to hide the `TemporaryCredentialsApi` and delta's`TemporaryCredentialsApi` details and differences inside the interface,  isn't it ? 

Then let's also never expose the the two concrete class to upper layer.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
```diff
@@ -0,0 +1,199 @@
...
+    String value = conf.get(key);
+    Preconditions.checkArgument(value != null && !value.isEmpty(), "'%s' is not set", key);
+    return value;
+  }
+
+  private CredentialOperation toUCDeltaCredentialOperation(String operation) {
+    switch (operation) {
+      case "READ":
+      case "PATH_READ":
+        return CredentialOperation.READ;
```
openinx: Why for the new api,  we will have `READ` and `PATH_READ` two read kinds, while for the old new we only have one ?  what is the difference ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
```diff
@@ -0,0 +1,199 @@
...
+                "Unsupported UC Delta API credential operation '%s'. "
+                    + "Supported operations are READ, READ_WRITE, PATH_READ, and PATH_READ_WRITE.",
+                operation));
+    }
+  }
+
+  private StorageCredential selectStorageCredential(
+      String location, List<StorageCredential> storageCredentials) {
+    Preconditions.checkNotNull(
+        storageCredentials,
```
openinx: I think we will need to check `storageCredentials != null  && !storageCredentials.isEmpty()`, if not meet, then we'd better to throw the missing storage credentials exception.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
```diff
@@ -0,0 +1,199 @@
...
+
+  private String stripTrailingSlash(String value) {
+    int end = value.length();
+    int minLength = value.indexOf("://");
+    minLength = minLength >= 0 ? minLength + "://".length() : 1;
+    while (end > minLength && value.charAt(end - 1) == '/') {
+      end--;
+    }
+    return value.substring(0, end);
+  }
```
openinx: I suggest to make this as separate static method in the utility class, with that we can provide very detailed test coverage for all the corner cases.  Could you pls file a separate issue to track this and do this before the 0.5.0 release ?

openinx: for line 166 ~ line 198

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  private StorageCredential selectStorageCredential(
+      String location, List<StorageCredential> storageCredentials) {
+    Preconditions.checkNotNull(
+        storageCredentials,
+        "UC Delta API credential response for location '%s' is missing storage credentials.",
+        location);
+    if (storageCredentials.size() == 1) {
```
openinx: TODO:  do we need to check if the `storageCredentials.prefix` is exactly matching the `location` that you required ?

openinx: otherwise it still easy to be buggy.

yili-db: Unfortunately `NormalizedURL` isn't usable here. This is what server uses to normalize URLs from inputs.
So if prefix needs to match what is returned by a previous `loadTable`, that's fine because both are normalized.
If prefix needs to match a URL input by user, which I don't think it should happen, then it will be problematic unless it also calls `NormalizedURL`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/fs/CredScopedFileSystem.java
```diff
@@ -95,56 +100,64 @@ FileSystem getDelegate() {
   @Override
   public void initialize(URI uri, Configuration conf) throws IOException {
     CredScopedKey key = CredScopedKey.create(uri, conf);
-    try {
-      this.fs = CACHE.get(key, () -> newFileSystem(uri, conf));
-    } catch (ExecutionException e) {
-      throw new IOException("Failed to initialize filesystem for key " + key, e.getCause());
+    synchronized (CACHE) {
```
openinx: this will greatly decrease the throughput of the workload, since every file request will go through this `initialize` method,  and you are making it to be a global lock per process,  which means even we are accessing the filesystem for different `CredScopedKey`,  we are compete for the the same process level CACHE lock. 

While before our PR, we are compete for the `key` level lock, which will perform significantly better than this.  ( actually, we have users currently are trying to provide better performance. ) .

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/fs/CredScopedFileSystem.java
```diff
@@ -95,56 +100,64 @@ FileSystem getDelegate() {
...
+  private static void closeQuietly(FileSystem fs) {
+    if (fs == null) {
+      return;
+    }
+    try {
+      fs.close();
     } catch (IOException e) {
-      throw new RuntimeException(e);
+      // ignore close failures on eviction
     }
```
openinx: nit: will be good for us to have separate util to quitely close the `Closeable` resource, not for `FileSystem` only.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredPropsUtil.java
```diff
@@ -0,0 +1,897 @@
...
+import io.unitycatalog.hadoop.credentials.GcsVendedTokenProvider;
+import io.unitycatalog.hadoop.fs.CredScopedFileSystem;
+import io.unitycatalog.hadoop.fs.CredScopedFs;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.UUID;
+
+public class CredPropsUtil {
```
openinx: what's the difference for this file ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredPropsUtil.java
```diff
@@ -0,0 +1,897 @@
...
+              UCHadoopConf.UC_CREDENTIALS_API_UC_DELTA_VALUE);
+        } else {
+          return abfsFixedCredProps(credScopedFsEnabled, fsImplProps, credential);
+        }
+      default:
+        return Collections.emptyMap();
+    }
+  }
+
+  public static Map<String, String> createPathCredProps(
```
openinx: This is the core API that we need to seriously consider, because it will be cross-repo depended. and we have to consider the cross-repo compability issue.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredPropsUtil.java
```diff
@@ -0,0 +1,897 @@
...
+                    credScopedFsEnabled, fsImplProps, uri, tokenProvider, tempCreds),
+                tableId,
+                tableOp)
+            : abfsFixedCredProps(credScopedFsEnabled, fsImplProps, tempCreds);
+      default:
+        return Collections.emptyMap();
+    }
+  }
+
+  public static Map<String, String> createPathCredProps(
```
openinx: This seems have no usage ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredPropsUtil.java
```diff
@@ -0,0 +1,897 @@
...
+        }
+      default:
+        return Collections.emptyMap();
+    }
+  }
+
+  public static Map<String, String> createPathCredProps(
+      boolean renewCredEnabled,
+      boolean credScopedFsEnabled,
+      Map<String, String> fsImplProps,
```
openinx: If we consider this as a public API for cross-repo dependency, this field is pretty unclear.   Let's discuss how to design a carefully maintained public API for this.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCHadoopConf.java
timothyw553: Keeping `UC` and `UC_DELTA` here to match Yili’s earlier naming comment. Yili asked not to call the existing API `legacy` yet because it is still the UC credentials API for non-Delta table formats; `UC_DELTA` is the Delta-specific credentials API value.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
timothyw553: Preserving Yili’s rule here: if the response has exactly one storage credential, use it even when its prefix does not match. When the response has multiple credentials, the code now selects by normalized exact-or-parent prefix boundary and picks the longest match. I also extracted this into `StorageCredentialSelector` with tests for null/empty responses, null or prefixless multi-credential entries, exact/parent/sibling matching, trailing slash normalization, longest prefix wins, and the single-credential bypass.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredPropsUtil.java
timothyw553: Tightened this surface instead of expanding it: removed the unused UC Delta path `StorageCredential` overload, kept the existing standard UC table/path methods that Spark calls, and added Javadocs for the remaining public methods. I also renamed `fsImplProps` to `existingFsImplProps` in source so IDE hints explain that these are the pre-existing Hadoop filesystem implementation properties preserved under `<property>.original` when credential-scoped filesystems are enabled. If we want a stronger source-level cross-repo API, I agree the next shape should be a request object rather than more positional parameters; I kept that out of this PR to avoid adding another public type before we need it.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredPropsUtil.java
timothyw553: Renamed this source parameter to `existingFsImplProps` and documented it on the remaining public methods: these are the current Hadoop filesystem implementation properties, and when credential-scoped filesystems are enabled the overwritten values are copied to `<property>.original`. I also removed the unused UC Delta path overload so this comment now applies only to the existing standard UC path helper that Spark already calls.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredPropsUtil.java
timothyw553: This is the old Spark auth credential-property utility moved into the Hadoop module so the vended credential providers and `CredScopedFileSystem` can be used without a Spark dependency. The functional additions in this PR are the UC Delta `StorageCredential` table handling and the credential API config that lets renewal choose between the standard UC API and the UC Delta API.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: implemented this split as `UcGenericCredentialApi` and `UcDeltaGenericCredentialApi`, with the factory choosing by `fs.unitycatalog.credentials.api`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: added tests for null and empty lists, single credential fallback, prefix matching, trailing slashes, and longest match.

timothyw553: kept support for multiple credentials because Delta may return one credential per storage location in the future.

timothyw553: implemented this rule: use one credential directly, and match by prefix only when there are multiple credentials.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialProvider.java
timothyw553: replaced the synchronized-map cache with keyed bounded cache loading.

timothyw553: filed follow-up issue #1531 to track adding a relocated cache dependency.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
timothyw553: changed the factory to `GenericCredentialApi.create(Configuration, ApiClient)` and kept generated clients inside package-private implementations.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
timothyw553: the Delta API has separate table and path operation enums, so the code now maps each one explicitly.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
timothyw553: added validation so null or empty storage credential responses fail with a missing credentials error.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
timothyw553: moved prefix matching into `StorageCredentialSelector` static helpers and added focused tests for the corner cases.

timothyw553: moved the logic from that range into `StorageCredentialSelector` so it can be tested separately.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredentialApi.java
timothyw553: added prefix-boundary tests so sibling paths like `table-other` do not match `table`.

timothyw553: for table renewals we use the UC table location, and for path renewals the selector only normalizes trailing slashes.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/fs/CredScopedFileSystem.java
timothyw553: replaced the process-level filesystem cache lock with keyed loading and added a test that one scope does not block another.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/fs/CredScopedFileSystem.java
timothyw553: added `CloseableUtils.closeQuietly(Closeable)` and use it for cache eviction and clear.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredPropsUtil.java
timothyw553: removed the unused UC Delta path `StorageCredential` overload and kept only the path helper used by Spark.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredPropsUtil.java
```diff
@@ -0,0 +1,848 @@
...
+   * @param uri Unity Catalog server URI used by renewal providers
+   * @param tokenProvider Unity Catalog token provider used by renewal providers
+   * @param tableId Unity Catalog table ID retained for compatibility with the standard UC API
+   * @param catalog table catalog name used by the UC Delta API renewal request
+   * @param schema table schema name used by the UC Delta API renewal request
+   * @param table table name used by the UC Delta API renewal request
+   * @param location table storage location used to scope renewed credentials
+   * @param credential UC Delta storage credential selected for the table location
+   */
+  public static Map<String, String> createTableCredProps(
```
tdas: not in this relocation PR, but this function signature is very wide.

tdas: please put some structure to this. the order of the params are random. please think of some structure.
- scheme 
- table details (catalog, schema, tablename, tableid)
- token provider
- catalog-specific configs (renewCredEnabled, credScopedFsEnabled) ---> this should be a map
- existingFsImplProps

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/credentials/GenericCredential.java
```diff
@@ -0,0 +1,157 @@
+package io.unitycatalog.hadoop.credentials;
```
tdas: move all classes except the util to internal

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
```diff
@@ -92,30 +95,52 @@ public GenericCredential accessCredentials() {
...
+    if (deltaTempCredApi == null) {
+      synchronized (this) {
+        if (deltaTempCredApi == null) {
+          deltaTempCredApi =
+              new io.unitycatalog.client.delta.api.TemporaryCredentialsApi(apiClient());
+        }
+      }
+    }
+    return deltaTempCredApi;
+  }
```
openinx: I think we agreed that we should move this out of the `GenericCredentialProvider`, and make it as a separate `TempCredentialApi` interface, right ?

openinx: Pls don't mix those old and new API calls in the same `GenericCredentialProvider`,  use a small separate interface will cleanly move those noisy out of the main logic, which make the maintenance much better.  and can help us design the well-tested code.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
```diff
@@ -92,30 +95,52 @@ public GenericCredential accessCredentials() {
...
+              ApiClientUtils.create(
+                  ucUri,
+                  tokenProvider,
+                  UCHadoopConf.createRequestRetryPolicy(conf),
+                  conf.getPropsWithPrefix(UCHadoopConf.UC_ENGINE_VERSION_PREFIX));
         }
       }
     }
+    return apiClient;
+  }
```
openinx: If we have the separate and abstracted `TempCredentialsApi` , then we can move the entire ApiClient to that API implementation,  and this `GenericCredentialProvider` won't need to perceive the api client intialization.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
```diff
@@ -156,14 +185,59 @@ public Map<String, String> buildForTable(String tableId, TableOperation tableOpe
...
     }
 
+    /**
+     * Builds Hadoop properties for a <em>UC Delta table's</em> storage location. Renewal uses the
+     * UC Delta credentials API (catalog/schema/table + location).
+     *
+     * @return unmodifiable map; empty if the scheme is unrecognized
+     * @throws IllegalStateException if a required field is missing
+     */
+    public Map<String, String> buildForTable(DeltaTableRef table) {
```
openinx: the `tableId`,`catalog`, `schema`, `name`, `location` , are all the required parameters for the `buildforTable` right ?  As the `DeltaTableRef.Builder` can be optionally set those fields.  then it will be good for us to define those parameters clearly in this method signature , as required parameters.

```
public Map<String, String> buildForTable(String catalog, String schema, String table, String location)
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
```diff
@@ -92,28 +85,15 @@ public GenericCredential accessCredentials() {
...
-          // configuration under UC_ENGINE_VERSION_PREFIX.
-          tempCredApi =
-              new TemporaryCredentialsApi(
-                  ApiClientUtils.create(
-                      ucUri,
-                      tokenProvider,
-                      UCHadoopConf.createRequestRetryPolicy(conf),
-                      conf.getPropsWithPrefix(UCHadoopConf.UC_ENGINE_VERSION_PREFIX)));
+        if (credentialApi == null) {
+          credentialApi = TempCredentialApi.create(conf, ucUri, tokenProvider);
```
openinx: Hi @TimothyW553 ,  I just read the `ucUri` and `tokenProvider` initialization here: https://github.com/unitycatalog/unitycatalog/pull/1527/changes#diff-a4c901f71862e1cc1be5de6535b7b64de800ef0475794fc589068b80a848156bL54-L60

I think the two member initialization should be moved into the `TempCredentialApi.create`,  since those are the internal details of the concret `TempCredentialAPI` initialization.  

And the main logic of `GenericCredentialProvider` shoudn't perceive it.  And that also make more sense for use to use only the hadoop conf to create and initialize the `TempCredentialApi`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
```diff
@@ -125,46 +105,12 @@ private GenericCredential renewCredential() throws ApiException {
           return cached;
         }
         // Renew the credential and update the cache.
-        GenericCredential renewed = createGenericCredentials();
+        GenericCredential renewed = tempCredentialApi().renew();
```
openinx: I think the `tempCredentialApi().renew()` will be better to rename as `tempCredentialApi().createCredentials`, as the renew is the process that check the if the existing credential is valid or not, and if not then create a new credential to replace the existing one.  and we already have the `readyToRenew` check in the above few lines of code. 

That's why it makes more sense for us to call it as `createCredential` rather than the `renew`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
```diff
@@ -0,0 +1,76 @@
...
+      } else {
+        String tableId = conf.get(UCHadoopConf.UC_TABLE_ID_KEY);
+        String tableOperation = conf.get(UCHadoopConf.UC_TABLE_OPERATION_KEY);
+        tempCred =
+            api.generateTemporaryTableCredentials(
+                new GenerateTemporaryTableCredential()
+                    .tableId(tableId)
+                    .operation(TableOperation.fromValue(tableOperation)));
+      }
+      return new GenericCredential(tempCred);
```
openinx: I think we actually don't need to check those cred type properties from the hadoop conf , for every credential request. which actually it's not worth to do that. 

I will suggest to parse those `path/pathOperation` and `table/tableOperation` from the hadoop conf firstly, and for every create credential request, we always reuse those parsed path-arguments and table-arguments. 

```java
static GenericCredential createUcTableCred(
      TemporaryCredentialsApi api, String tableId, TableOperation tableOp)
      throws ApiException {
    TemporaryCredentials cred = api.generateTemporaryTableCredentials(
            .tableId(tableId)
            .operation(tableOp));
    return new GenericCredential(cred);
  }

  static GenericCredential createUcPathCred(
      TemporaryCredentialsApi api, String path, PathOperation pathOp) throws ApiException {
    TemporaryCredentials cred = api.generateTemporaryPathCredentials(
        new GenerateTemporaryPathCredential()
            .url(path)
            .operation(pathOp));

    return new GenericCredential(cred);
  }

  static TempCredentialApi forUc(Configuration conf, TemporaryCredentialsApi api) {
    String type = conf.get(UCHadoopConf.UC_CREDENTIALS_TYPE_KEY);
    if (UCHadoopConf.UC_CREDENTIALS_TYPE_PATH_VALUE.equals(type)) {
      String path = conf.get(UCHadoopConf.UC_PATH_KEY);
      String pathOperation = conf.get(UCHadoopConf.UC_PATH_OPERATION_KEY);

      return () -> createUcPathCred(api, path, PathOperation.valueOf(pathOperation));
    } else if (UCHadoopConf.UC_CREDENTIALS_TYPE_TABLE_VALUE.equals(type)) {
      String tableId = conf.get(UCHadoopConf.UC_TABLE_ID_KEY);
      String tableOperation = conf.get(UCHadoopConf.UC_TABLE_OPERATION_KEY);

      return () -> createUcTableCred(api, tableId, TableOperation.valueOf(tableOperation));
    } else {
      throw new IllegalArgumentException(
          String.format(
              "Unsupported unity catalog temporary credentials type '%s', please check '%s'",
              type,
              UCHadoopConf.UC_CREDENTIALS_TYPE_KEY));
    }
  }
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
```diff
@@ -0,0 +1,76 @@
...
+import io.unitycatalog.client.model.TemporaryCredentials;
+import io.unitycatalog.hadoop.internal.UCHadoopConf;
+import java.net.URI;
+import org.apache.hadoop.conf.Configuration;
+
+interface TempCredentialApi {
+  GenericCredential renew() throws ApiException;
+
+  static TempCredentialApi create(Configuration conf, URI ucUri, TokenProvider tokenProvider) {
+    String credentialApi = conf.get(UCHadoopConf.UC_CREDENTIALS_API_KEY);
```
openinx: I understood the `fs.unitycatalog.credentials.api` is the config key that decided to use the old cred api or the new cred api.  

will we introduce more config key for us to decide to call those apis that are not the cred related api ?  If so,  then we have to configure at least two hadoop config keys, for switch between the new api and old api ? 

I think we should unify those into one config key, so that people & users won't miss, and also lower the threshold for new one to use.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
```diff
@@ -0,0 +1,76 @@
...
+                    .operation(TableOperation.fromValue(tableOperation)));
+      }
+      return new GenericCredential(tempCred);
+    };
+  }
+
+  static TempCredentialApi forUcDelta(
+      Configuration conf, io.unitycatalog.client.delta.api.TemporaryCredentialsApi api) {
+    return () -> UCDeltaCredentials.renew(conf, api);
+  }
```
openinx: Maybe we just introduce the separate 3 class/interface into 3 java files. 
1.  define the `TempCredentialApi` interface and the `create` to factory the concret instance/implementation, according to the hadoop conf. 

2.  `UcTempCredentialAPi` class, impl the `TempCredentialApi`. separate java file. 

3. `UcDetlaTempCredentialApi` class, impl the `TempCredentialApi`, separate java file.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaCredentials.java
```diff
@@ -0,0 +1,186 @@
...
+import io.unitycatalog.client.model.TableOperation;
+import io.unitycatalog.client.model.TemporaryCredentials;
+import io.unitycatalog.hadoop.internal.UCHadoopConf;
+import java.util.List;
+import java.util.Map;
+import org.apache.hadoop.conf.Configuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class UCDeltaCredentials {
```
openinx: Do we need to make it `public` ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaCredentials.java
```diff
@@ -0,0 +1,186 @@
...
+  public static void writeKeys(
+      Map<String, String> props, String catalog, String schema, String name, String location) {
+    props.put(UCHadoopConf.UC_CREDENTIALS_API_KEY, API_VALUE);
+    props.put(CATALOG_KEY, catalog);
+    props.put(SCHEMA_KEY, schema);
+    props.put(NAME_KEY, name);
+    props.put(LOCATION_KEY, location);
+  }
+
+  public static GenericCredential renew(Configuration conf, TemporaryCredentialsApi api)
```
openinx: It's not `renew` credentials, it's particularly creating the table credential. so it'd better to rename it as `createDeltaTableCredential`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaCredentials.java
```diff
@@ -0,0 +1,186 @@
...
+        response != null,
+        "UC Delta API returned no credentials response for '%s.%s.%s'.",
+        catalog,
+        schema,
+        name);
+    StorageCredential picked = selectForLocation(location, response.getStorageCredentials());
+    if (picked.getPrefix() != null && !prefixCovers(location, picked.getPrefix())) {
+      LOG.warn(
+          "UC Delta credential for {}.{}.{} (location={}) has prefix '{}' that does not cover the"
+              + " location. Using anyway; expect IAM denial if the server is misconfigured.",
```
openinx: that's very dangerous.   If I just create a credentials whose prefix is not exactly matching the location, but I used it to access those location.

If the access is denied, then will it be better for us to throw the exception here ? 
If the access is working,  then I'm using a credential that access the location which is not matching the prefix requirement. must be some credentials leak or wrong issue. 

both are not make sense to me. 

So if the prefix mismatch,  then we should throw the exception directly, rather than continue the following process.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
timothyw553: done. `GenericCredentialProvider` now creates `TempCredentialApi` only from the hadoop conf

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
timothyw553: done. `renew()` is renamed to `createCredential()`

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
timothyw553: done. uc table/path request fields are parsed once and reused

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
timothyw553: done. we keep one api switch key: `fs.unitycatalog.credentials.api`

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
timothyw553: done. i split this into three files: `TempCredentialApi`, `UcTempCredentialApi`, and `UcDeltaTempCredentialApi`

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaCredentials.java
timothyw553: done. `UCDeltaCredentials` is package-private now

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaCredentials.java
timothyw553: done. the delta table credential method is now `createDeltaTableCredential`

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaCredentials.java
timothyw553: done. delta credential prefix mismatch now throws an exception

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
timothyw553: done. the delta table builder now requires `tableId`, `catalog`, `schema`, `table`, and `location`

timothyw553:
architecture

```text
+-------------------------+
| public builder          |
| UCCredentialHadoopConfs |
+-----------+-------------+
            |
            v
+-------------------------+
| hadoop config           |
| stores api choice       |
+-----------+-------------+
            |
            v
+-------------------------+
| credential provider     |
| cache + refresh         |
+-----------+-------------+
            |
            v
+-------------------------+
| TempCredentialApi       |
| chooses one path        |
+-----------+-------------+
            |
      +-----+------+
      |            |
      v            v
+-----------+  +----------------+
| UC API    |  | UC Delta API   |
| table/path|  | table location |
+-----+-----+  +--------+-------+
      |                 |
      |                 v
      |        +----------------+
      |        | delta util     |
      |        | pick + convert |
      |        +--------+-------+
      |                 |
      +--------+--------+
               |
               v
+-------------------------+
| GenericCredential       |
| used by cloud providers |
+-------------------------+
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConf.java
```diff
@@ -65,10 +65,19 @@ private UCHadoopConf() {}
   // credentials are indexed by this key and are not reused across different jobs.
   public static final String UC_CREDENTIALS_UID_KEY = "fs.unitycatalog.credentials.uid";
 
+  public static final String UC_CREDENTIALS_API_KEY = "fs.unitycatalog.credentials.api";
+  public static final String UC_CREDENTIALS_API_DELTA_VALUE = "uc_delta";
```
tdas: what are the possible values for this? why cant this just be a boolean?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConf.java
```diff
@@ -65,10 +65,19 @@ private UCHadoopConf() {}
...
+
   // Keys for table based temporary credential requests
   public static final String UC_TABLE_ID_KEY = "fs.unitycatalog.table.id";
   public static final String UC_TABLE_OPERATION_KEY = "fs.unitycatalog.table.operation";
 
+  // Keys for UC Delta table credential requests.
+  public static final String UC_DELTA_CATALOG_KEY = "fs.unitycatalog.delta.catalog";
+  public static final String UC_DELTA_SCHEMA_KEY = "fs.unitycatalog.delta.schema";
+  public static final String UC_DELTA_NAME_KEY = "fs.unitycatalog.delta.name";
```
tdas: what is name?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UcDeltaTempCredentialApi.java
```diff
@@ -0,0 +1,75 @@
...
+
+final class UcDeltaTempCredentialApi implements TempCredentialApi {
+  private final TemporaryCredentialsApi api;
+  private final CredentialOperation operation;
+  private final String catalog;
+  private final String schema;
+  private final String name;
+  private final String location;
+
+  UcDeltaTempCredentialApi(Configuration conf, TemporaryCredentialsApi api) {
```
tdas: Uc --> UC everywhere. there is already UCHadoopConf... keep it consistent.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
```diff
@@ -0,0 +1,43 @@
...
+            conf.getPropsWithPrefix(UCHadoopConf.UC_ENGINE_VERSION_PREFIX));
+    if (UCHadoopConf.UC_CREDENTIALS_API_DELTA_VALUE.equals(credentialApi)) {
+      return new UcDeltaTempCredentialApi(
+          conf, new io.unitycatalog.client.delta.api.TemporaryCredentialsApi(apiClient));
+    }
+    return new UcTempCredentialApi(
+        conf, new io.unitycatalog.client.api.TemporaryCredentialsApi(apiClient));
+  }
+
+  private static URI ucUri(Configuration conf) {
```
tdas: nit: why 1 single use, 2 line  private static method?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
```diff
@@ -0,0 +1,43 @@
...
+import io.unitycatalog.client.ApiClient;
+import io.unitycatalog.client.ApiException;
+import io.unitycatalog.client.auth.TokenProvider;
+import io.unitycatalog.client.internal.ApiClientUtils;
+import io.unitycatalog.client.internal.Preconditions;
+import io.unitycatalog.hadoop.internal.UCHadoopConf;
+import java.net.URI;
+import org.apache.hadoop.conf.Configuration;
+
+interface TempCredentialApi {
```
tdas, line 17: can you please add class docs to all these classes. these all classes and interface sound very similar to the SDK generated classes and it confusing. so (1) can you name them more differently, like GenericCredentialProvider (because its job is to provide/create GenericCredential)?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UcDeltaTempCredentialApi.java
```diff
@@ -0,0 +1,75 @@
...
+import io.unitycatalog.client.delta.api.TemporaryCredentialsApi;
+import io.unitycatalog.client.delta.model.CredentialOperation;
+import io.unitycatalog.client.delta.model.CredentialsResponse;
+import io.unitycatalog.client.internal.Preconditions;
+import io.unitycatalog.client.model.TableOperation;
+import io.unitycatalog.hadoop.internal.DeltaStorageCredentialUtil;
+import io.unitycatalog.hadoop.internal.UCHadoopConf;
+import org.apache.hadoop.conf.Configuration;
+
+final class UcDeltaTempCredentialApi implements TempCredentialApi {
```
tdas: class docs

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConf.java
tdas: shouldnt this be table name? its too ambiguous otherwise.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConf.java
tdas: can you rename this to UCHadoopConf --> UCHadoopConfConstants

to differentiate from UCCredentialHadoopConfs which provides public utils for hadoop confs.

tdas (approved):
high-level approval but please address my comments before merging

openinx:
Let me have a final check for this PR, thanks @tdas for the reviewing, and thanks @TimothyW553 for the work.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConf.java
openinx: let's just align all the three config keys as: 

* fs.unitycatalog.delta.catalog -> fs.unitycatalog.delta.catalogName
* fs.unitycatalog.delta.schema -> fs.unitycatalog.delta.schemaName
* fs.unitycatalog.delta.name -> fs.unitycatalog.delta.tableName

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
```diff
@@ -80,12 +83,33 @@ public Builder tokenProvider(TokenProvider tokenProvider) {
...
+     * <p>Mutually exclusive with {@link #initialCredentials(TemporaryCredentials)}.
+     */
+    public Builder initialStorageCredential(StorageCredential initialStorageCredential) {
+      Preconditions.checkState(
+          initialCredentials == null,
+          "initialCredentials is already set; cannot also set initialStorageCredential. The two"
+              + " select different credentials APIs (UC vs UC Delta).");
+      this.initialStorageCredential = initialStorageCredential;
+      return this;
+    }
```
openinx, line 111: Wait, currently, all the initial credentials are created from the old client api which are from a totally new difference code path in `UCSingleCatalog` , it's separated code from this `UCCredentialsHadoopConfs`. 

And currently our `fs.unitycatalog.credentials.api` only control the code inside the `UCCredentialsHadoopConfs`,  no control for the `UCSingleCatalog`.  Then how could we use the same setting to control the `UCSingleCatalog` 's credentials initialization ?

I think this is a key question.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
```diff
@@ -156,14 +180,60 @@ public Map<String, String> buildForTable(String tableId, TableOperation tableOpe
...
+     * @param catalog the UC catalog name
+     * @param schema the UC schema name
+     * @param table the UC table name
+     * @param location the table storage location used to select a returned storage credential
+     * @return unmodifiable map; empty if the scheme is unrecognized
+     * @throws IllegalArgumentException if a table identity field is missing
+     * @throws IllegalStateException if a required builder field is missing
+     */
+    public Map<String, String> buildForTable(
+        String tableId, String catalog, String schema, String table, String location) {
```
openinx, line 196: This is looks a bit weird.  because this API actually needs both `tableId` and `tableName`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
```diff
@@ -156,14 +180,60 @@ public Map<String, String> buildForTable(String tableId, TableOperation tableOpe
...
+          CredPropsUtil.createTableCredProps(
+              credentialRenewalEnabled,
+              credentialScopedFsEnabled,
+              hadoopConf,
+              scheme,
+              catalogUri,
+              tokenProvider,
+              tableId,
+              DeltaStorageCredentialUtil.toTableOperation(initialStorageCredential.getOperation()),
+              tempCreds);
```
openinx, line 215: Oh,  that's not good....... because `CredPropsUtil.createTableCredProps` was created for the old client api to generate the temp credential. you can see the tableId was designed for the old client temp cred generation. 

while for our new client api generation,  the `tableId` is totally no required. 

Also it's highly related to the `CredScopedFileSystem`,  the CredScopedKey is purely designed for old client api,  and if we want to support the new client,  then we will need to seriously consider this....

this comment is critical.

openinx: I mean the current problems are: 
1.  `tableId`  should be removed for the new `buildForTable` API.
2.  CredScopedFileSystem are mixed for the new client api and old client api, that's the most serious problem.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConf.java
timothyw553: done. renamed `UCHadoopConf` to `UCHadoopConfConstants` and updated references.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConf.java
timothyw553: done. renamed the delta table-name config to make `tableName` explicit.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UcDeltaTempCredentialApi.java
timothyw553: done. renamed `Uc*` classes to `UC*` for consistency.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
timothyw553: done. added class docs clarifying these are hadoop credential adapters, not sdk-generated APIs.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
timothyw553: done. inlined the single-use helper.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConf.java
timothyw553: done. changed this to a boolean selector for the delta credentials api, but we may revisit it as part of the larger drc switch design with zheng/yi.

openinx (approved):
LGTM for this PR, but we will have some fundamental problems need to solve for the following PRs. 
1.  moving the `initialCredentials/storageCredentials` inside the Builder itself, and not exposing any new/old client apis. 
2. Separeate the `CredPropsUtils` for both new & old client api, since two clients are setting different hadoop confs,  and if we don't fix this, it will mix the credentials in the CredScopedFileSystem, which causes serious bug. 


But I don't want to block the following work, so let me merge this one firstly.  @tdas @TimothyW553 . 

Huge thanks @tdas  for the reviewing work, and huge thanks @TimothyW553 for hard work.

PR #1528, opened 2026-05-01
Delta: last-commit-timestamp-ms on createTable

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1528/files) to review incremental changes.
- [**stack/create_table_last_ts**](https://github.com/unitycatalog/unitycatalog/pull/1528) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1528/files)]

---------
**PR Checklist**

**Description of changes**
Delta: last-commit-timestamp-ms on createTable

Adds a required top-level `last-commit-timestamp-ms` field to the createTable request so engines hand UC the v0 commit's timestamp out-of-band rather than smuggling it through `properties`. Server derives `delta.lastCommitTimestamp` from this field at create-time.

### DIFF
--- a/api/delta-docs/Models/CreateTableRequest.md
+++ b/api/delta-docs/Models/CreateTableRequest.md
@@ -13,6 +13,7 @@
 | **protocol** | [**DeltaProtocol**](DeltaProtocol.md) | Delta protocol version and feature requirements | [default to null] |
 | **properties** | **Map** | Delta table properties | [default to null] |
 | **domain-metadata** | [**DomainMetadataUpdates**](DomainMetadataUpdates.md) |  | [optional] [default to null] |
+| **last-commit-timestamp-ms** | **Long** | Timestamp of version 0 (the commit the client wrote  before calling this endpoint), in epoch milliseconds.  | [default to null] |
 | **uniform** | [**UniformMetadata**](UniformMetadata.md) | Optional UniForm conversion metadata. When present, the table is registered as UniForm-enabled, so readers can access it via either the Delta or Iceberg REST Catalog. The engine generates the Iceberg metadata file at the supplied metadata-location before calling createTable.  | [optional] [default to null] |
 
 [[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
@@ -15,10 +15,9 @@ import java.util.List;
 import java.util.Optional;
 
 /**
- * Converts a Delta REST Catalog {@link CreateTableRequest} (with typed Delta columns and kebab-case
- * field names) into the UC {@link CreateTable} (with UC {@link ColumnInfo}s and
- * partition-index-per-column). The server holds path params for catalog and schema; the rest comes
- * from the request body.
+ * Converts a Delta {@link CreateTableRequest} (with typed Delta columns and kebab-case field names)
+ * into the UC {@link CreateTable} (with UC {@link ColumnInfo}s and partition-index-per-column). The
+ * server holds path params for catalog and schema; the rest comes from the request body.
  *
  * <p>Required-field checks (name, location, columns, protocol, table-type, data-source-format) and
  * the DELTA-only format rule apply to all tables. The full UC catalog-managed contract ({@link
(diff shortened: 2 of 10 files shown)


timothyw553 (approved):
LGTM.

Please update description to reflect code changes though

api/delta-docs/Models/CreateTableRequest.md
```diff
@@ -13,6 +13,7 @@
 | **protocol** | [**DeltaProtocol**](DeltaProtocol.md) | Delta protocol version and feature requirements | [default to null] |
 | **properties** | **Map** | Delta table properties | [default to null] |
 | **domain-metadata** | [**DomainMetadataUpdates**](DomainMetadataUpdates.md) |  | [optional] [default to null] |
+| **last-commit-timestamp-ms** | **Long** | Timestamp of version 0 (the commit the client wrote  before calling this endpoint), in epoch milliseconds.  | [default to null] |
```
openinx, line 16: How could the client know the last-commit-timestamp-ms ?   and what's the usage for adding this field ?

server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
```diff
@@ -49,6 +48,10 @@ public static CreateTable toCreateTable(String catalog, String schema, CreateTab
     if (req.getProtocol() == null) {
       throw new BaseException(ErrorCode.INVALID_ARGUMENT, "protocol is required.");
     }
+    if (req.getLastCommitTimestampMs() == null) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT, "last-commit-timestamp-ms is required.");
+    }
```
openinx, line 67: Is this field required ?  I just see the CreateTableRequest.md says, the default value is null .

openinx: https://github.com/unitycatalog/unitycatalog/pull/1528/changes#diff-273d56bdb9e73b77809a065870e23c144e6b9ce0a6cb860acae582969f4eba3bR16

server/src/test/java/io/unitycatalog/server/service/delta/DeltaPropertyMapperTest.java
```diff
@@ -137,27 +141,44 @@ public void mergeDerivedPropertiesWinOnConflict() {
...
-            .deltaRowTracking(new RowTrackingDomainMetadata().rowIdHighWaterMark(100L));
-    Map<String, String> client = Map.of("custom.key", "custom.value");
-    Map<String, String> merged = DeltaPropertyMapper.buildStoredProperties(protocol, dm, client);
+    CreateTableRequest req =
+        new CreateTableRequest()
+            .protocol(
+                new DeltaProtocol()
+                    .minReaderVersion(3)
+                    .minWriterVersion(7)
+                    .writerFeatures(List.of(TableFeature.ROW_TRACKING.specName())))
```
openinx, line 156: do we have the timestamp also ?

api/delta-docs/Models/CreateTableRequest.md
openinx: I just read the PR description. 
> Adds a required top-level last-commit-timestamp-ms field to the createTable request so engines hand UC the v0 commit's timestamp out-of-band rather than smuggling it through properties. Server derives delta.lastCommitTimestamp from this field at create-time and from commit.commitInfo.timestamp on every metadata-changing addCommit, keeping properties a faithful round-trip of the Delta log's metaData.configuration

Could you pls add more context about `engines hand UC the v0 commit's timestamp out-of-band rather than smuggling it through properties` ?    what's the before and after differences?

yili-db: Before: clients prepare this property `delta.lastCommitTimestamp` with the same timestamp.
After: clients simply send the same timestamp via a dedicated `last-commit-timestamp-ms` field in createTable request.
This timestamp is essentially the commit timestamp of v0.json.
I hesitate to add this before/after comparison because this new API makes it clear and simply so the comment shouldn't talk about the old API that is more complex.

openinx: And if this is a required, do we need an extra column in this table to clearly say it's a required/optional column ?

server/src/main/java/io/unitycatalog/server/service/delta/DeltaCreateTableMapper.java
yili-db: Yes it is required. Optional fields have "[optional]".
This is defined as required in yaml. OpenAPI generated table is out of our control.

api/delta-docs/Models/CreateTableRequest.md
yili-db: In this auto generated table, `[optional]` means optional. Not having `[optional]` means required. It's always been like that.

server/src/test/java/io/unitycatalog/server/service/delta/DeltaPropertyMapperTest.java
yili-db: Just added.
This is a unit test of `buildStoredProperties` where it doesn't check (upper layer checks for timestamp presence).

api/delta-docs/Models/CreateTableRequest.md
openinx: Okay,  basically,  we move it from the table properties to the top-level field.  that make sense to me. 

As we make it as a required field ( since we did not see the [optional] for this field ),  that what's the reason that make it as a required field ? and usually for an external engine integrator, do they need to explicitly call the delta filesystem, and parse the log to get this required timestamp, which actually make the create table api call be more complicated ?

openinx (approved):

tdas (approved):

PR #1530, opened 2026-05-01
[Hadoop] Add unitycatalog-hadoop connector module scaffold

openinx:
**PR Checklist**

**Description of changes**

Introduces the `unitycatalog-hadoop` connector module under `connectors/hadoop/`, following the same structure as `unitycatalog-spark`.

- **`build.sbt`**: adds `lazy val hadoop` project definition that depends on `client`, uses `javaOnlyReleaseSettings` with Java 11, and declares Hadoop/cloud-provider dependencies (hadoop-client-api, hadoop-azure, util-hadoop, AWS SDK) as `Provided`, with standard test dependencies (JUnit 5, Mockito, hadoop-aws, gcs-connector). The module is included in the root `.aggregate()`.
- **`.github/CODEOWNERS`**: adds `@tdas @openinx @TimothyW553` as codeowners for `/connectors/hadoop/`.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,6 +6,9 @@
 # Spark Connector
 /connectors/spark/ @tdas @openinx
 
+# Hadoop Connector
+/connectors/hadoop/ @tdas @openinx @TimothyW553
+
 # Java clients.
 /clients/java/ @tdas @openinx
 
--- a/build.sbt
+++ b/build.sbt
@@ -692,6 +692,37 @@ lazy val spark = (project in file("connectors/spark"))
     }
   )
 
+lazy val hadoop = (project in file("connectors/hadoop"))
+  .dependsOn(client)
+  .enablePlugins(CheckstylePlugin)
+  .settings(
+    name := s"$artifactNamePrefix-hadoop",
+    commonSettings,
+    javaOnlyReleaseSettings,
+    javafmtCheckSettings(),
+    javaCheckstyleSettings("dev/checkstyle-config.xml"),
+    Compile / compile / javacOptions ++= javacRelease11,
+    libraryDependencies ++= Seq(
+      "org.apache.hadoop" % "hadoop-client-api" % hadoopVersion % Provided,
+      "com.google.cloud.bigdataoss" % "util-hadoop" % "3.0.2" % Provided,
+      "org.apache.hadoop" % "hadoop-azure" % hadoopVersion % Provided,
+      "software.amazon.awssdk" % "auth" % "2.25.37" % Provided,
+    ),
+    libraryDependencies ++= Seq(
+      // Test dependencies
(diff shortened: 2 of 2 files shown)


build.sbt
```diff
@@ -692,6 +692,37 @@ lazy val spark = (project in file("connectors/spark"))
...
+    commonSettings,
+    javaOnlyReleaseSettings,
+    javafmtCheckSettings(),
+    javaCheckstyleSettings("dev/checkstyle-config.xml"),
+    Compile / compile / javacOptions ++= javacRelease11,
+    libraryDependencies ++= Seq(
+      "org.apache.hadoop" % "hadoop-client-api" % hadoopVersion % Provided,
+      "com.google.cloud.bigdataoss" % "util-hadoop" % "3.0.2" % Provided,
+      "org.apache.hadoop" % "hadoop-azure" % hadoopVersion % Provided,
+      "software.amazon.awssdk" % "auth" % "2.25.37" % Provided,
```
openinx, line 709: The three are for importing the the token provider interfaces ( from the cloud filesystem ), so that we can implement those interfaces.

tdas: how are we going to make sure that things wont fail if a user does not have these in the classpath at runtime?

my worry is that we will accidentally attempt to load the aws-related classes in a azure env where aws SDK is not present at runtime. have you thought of that?

openinx: I think the ultimate approach is explicitly separating the cloud vendor's token providers into separate modules. Then it will introduces several moduels:

* unitycatalog-hadoop-aws
* unitycatalog-hadoop-azure
* unitycatalog-hadoop-gcp
* unitycatalog-hadoop-common

But considering there are very few code in those separate modules, I'm doubting it's worth to spawn so many modules.

I will incline to use the integration test coverage to coverage the classes mixes.  I mean we use the built artifacts, and only load the `hadoop-aws` to test the aws unitycatalog integration, and if all the tests are happy.  then we should be fine.

openinx: Besides, we should also use the `io.unitycatalog.spark.auth.storage.AbfsVendedTokenProvider` rather than the `AbfsVendedTokenProvider.class.getName()` in the https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/java/io/unitycatalog/spark/auth/CredPropsUtil.java#L364 , to avoid the loading class  problem as you said.

.github/CODEOWNERS
```diff
@@ -6,6 +6,9 @@
 # Spark Connector
 /connectors/spark/ @tdas @openinx
 
+# Hadoop Connector
+/connectors/hadoop/ @tdas @openinx @TimothyW553
```
tdas, line 10: hadoopUtils

tdas: its not a connector by itself. so its better to communicate in the name.

tdas (approved):
comments can be addressed in the follow up PR.. maybe in @TimothyW553 #1527

PR #1532, opened 2026-05-04
Honor aws.endpoint for STS and S3 clients

chrislusf:
## What this fixes

Today UC's STS and S3 clients always target real AWS, regardless of \`aws.endpoint\` in \`server.properties\` or any of the AWS SDK env vars / system properties. That makes UC unusable end-to-end against any S3-compatible self-hosted backend (SeaweedFS, MinIO, Ceph RGW, LocalStack), even though the rest of the configuration is wired up.

Two builders inside the server need an \`endpointOverride\`:

\`AwsCredentialGenerator.StsAwsCredentialGenerator\` (\`v0.4.0\`):

\`\`\`java
//                                                                              ^ no .endpointOverride
\`\`\`

\`FileIOFactory.getS3Client\`:

\`\`\`java
return S3Client.builder()
    .region(Region.of(region))
    .credentialsProvider(awsCredentialsProvider)
    .forcePathStyle(false)
    .build();
\`\`\`

The AWS Java SDK's generic env-var fallback (\`AWS_ENDPOINT_URL_STS\`, \`aws.endpointUrlSts\` system property, the catch-all \`AWS_ENDPOINT_URL\`) doesn't kick in for that builder shape. I confirmed by pointing \`AWS_ENDPOINT_URL_STS\` at port 1 (deliberately unreachable) and getting back the same real-AWS \`InvalidClientTokenId\` 403, with zero traffic to the configured port (HTTP sniffer in front of the local STS endpoint recorded nothing).

## What this PR does

1. Adds an \`AWS_ENDPOINT(\"aws.endpoint\")\` property and per-bucket \`s3.endpoint.N\`. Plumbs both through \`S3StorageConfig\`. Empty -> SDK default resolution (unchanged AWS path).
2. \`StsAwsCredentialGenerator\`: applies \`endpointOverride(URI.create(config.getEndpoint()))\` when set.
3. \`FileIOFactory.getS3Client\`: applies \`endpointOverride\` and flips to \`forcePathStyle(true)\` when an endpoint is set, since self-hosted backends typically can't issue wildcard certs for \`bucket.host\` (real AWS keeps virtual-hosted-style).
4. \`etc/conf/server.properties\`: documents the two new keys.
5. Two new unit tests in \`CloudCredentialVendorTest\`:
   - \`testStsClientHonorsAwsEndpointOverride\` -- endpoint set, expects \`endpointOverride(URI)\` to be called with that URI.
   - \`testStsClientSkipsEndpointOverrideWhenUnset\` -- endpoint unset, expects \`endpointOverride\` never to be called.

After this change, pointing UC at SeaweedFS / MinIO / Ceph RGW / LocalStack works end-to-end:

- catalog/schema/EXTERNAL Delta CRUD already worked
- \`/temporary-table-credentials\` now reaches the local STS, gets a real session token back, and \`uc table read\` works without forcing clients to bypass UC and use static keys directly

Repro and trace details for the original failure: <https://github.com/seaweedfs/seaweedfs/pull/9308> (SeaweedFS-side integration test that runs UC v0.4.0 in Docker and currently logs the broken handoff via \`t.Logf\` instead of asserting it).

## Compatibility

- \`aws.endpoint\` and \`s3.endpoint.N\` are both optional and default to empty.
- When empty, all builder shapes are byte-identical to before this PR (no \`endpointOverride\` call, \`forcePathStyle(false)\`).
- No breaking change to \`server.properties\` or to the existing test fixtures.

I wasn't able to run the full test suite locally (Maven Central is returning 404 for everything from this network -- \`net.bytebuddy\`, \`org.junit.platform\`, etc.), so this relies on the project's CI to validate. Happy to iterate on any feedback.

### DIFF
--- a/etc/conf/server.properties
+++ b/etc/conf/server.properties
@@ -48,6 +48,8 @@ aws.accessKey=
 aws.secretKey=
 # The region of AWS. If running on-premises, pick the nearest region.
 aws.region=
+# Optional STS/S3 endpoint override for self-hosted S3-compatible backends.
+aws.endpoint=
 
 #### The following "s3.*" configs are legacy per-bucket hardcoded credential config
 
@@ -60,6 +62,8 @@ s3.accessKey.0=
 s3.secretKey.0=
 # Test Only (If you provide a session token, it will just use those session creds, no downscoping)
 s3.sessionToken.0=
+# Optional per-bucket endpoint override.
+s3.endpoint.0=
 
 ## ADLS Storage Config (Multiple configs can be added by incrementing the index)
 adls.storageAccountName.0=
--- a/examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
+++ b/examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
@@ -168,6 +168,17 @@ public class DeltaKernelUtils {
       conf.set("fs.s3a.session.token", awsTempCredentials.getSessionToken());
       conf.set("fs.s3a.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem");
       conf.set("fs.s3a.path.style.access", "true");
+      // Hadoop S3A doesn't read AWS_ENDPOINT_URL; it only honors fs.s3a.endpoint.
+      String s3Endpoint = System.getenv("AWS_ENDPOINT_URL_S3");
+      if (s3Endpoint == null || s3Endpoint.isEmpty()) {
+        s3Endpoint = System.getenv("AWS_ENDPOINT_URL");
+      }
+      if (s3Endpoint == null || s3Endpoint.isEmpty()) {
+        s3Endpoint = System.getProperty("aws.endpoint");
+      }
+      if (s3Endpoint != null && !s3Endpoint.isEmpty()) {
+        conf.set("fs.s3a.endpoint", s3Endpoint);
+      }
     } else if (scheme.equals(Constants.URI_SCHEME_FILE)) {
       conf.set("fs.file.impl", "org.apache.hadoop.fs.LocalFileSystem");
     } else {
--- a/examples/cli/src/test/java/io/unitycatalog/cli/delta/DeltaKernelUtilsTest.java
+++ b/examples/cli/src/test/java/io/unitycatalog/cli/delta/DeltaKernelUtilsTest.java
@@ -1,0 +1,60 @@
+package io.unitycatalog.cli.delta;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.unitycatalog.client.model.AwsCredentials;
+import io.unitycatalog.client.model.TemporaryCredentials;
+import java.net.URI;
+import org.apache.hadoop.conf.Configuration;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class DeltaKernelUtilsTest {
+
+  private static final String AWS_ENDPOINT_PROP = "aws.endpoint";
+  private String savedEndpoint;
+
+  private TemporaryCredentials creds() {
+    return new TemporaryCredentials()
+        .awsTempCredentials(
+            new AwsCredentials()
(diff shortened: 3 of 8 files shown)


Linhares015:
Independent confirmation from a different angle: I hit the exact same root cause (`StsAwsCredentialGenerator` builds its `StsClient` with no `endpointOverride`, so any self-hosted STS is unreachable) while integrating UC v0.6.0 with SeaweedFS. Validated end-to-end, not just unit-level: with a minimal endpoint-override patch (narrower than this PR — STS only, restricted to loopback/private hosts by a validator, no `FileIOFactory`/`forcePathStyle` change), UC really does `AssumeRole` against a local STS and vends a working session credential that Spark/Delta then reads and writes with (including a session-policy-scoped `AccessDenied` outside the granted prefix). Details, commands and evidence: https://github.com/Linhares015/minerva/blob/main/docs/adr/0002-uc-sts-s3-governanca.md.

Re-ran just now against current `main` (`2725aee9af41ed4f5c98b4215034ff85d771f39d`) with the same shape of change to `StsAwsCredentialGenerator`/`S3StorageConfig`/`ServerProperties`: compiles clean, `server/testOnly io.unitycatalog.server.utils.ServerPropertiesTest` — 16/16 passed.

This PR's approach is more complete than what I tested (also covers `FileIOFactory.getS3Client`/`forcePathStyle`, which matters for UC's own S3 access, not only STS/credential vending), so no need for a competing PR from my side.

One likely reason this hasn't gotten review traction: the DCO check is currently failing (https://github.com/unitycatalog/unitycatalog/pull/1532/checks). @chrislusf, might be worth an amended commit with `Signed-off-by` per CONTRIBUTING.md — that's probably what's keeping this off reviewers' radar rather than the content itself, given it fixes a real, reproducible gap for every self-hosted-S3 deployment (SeaweedFS/MinIO/Ceph RGW/LocalStack) trying to use UC's credential vending instead of static keys.

Linhares015:
Follow-up from the earlier comment: did a broader security review of the endpoint-override mechanism (SSRF, metadata enumeration, lateral credential scope), since that's a real threat-model question once an endpoint override exists at all.

Relevant finding for this PR: my narrower patch added a loopback/private-host allowlist validator (`host.startsWith("10.")`/`"192.168."`/`"127.0.0.1"`) that I assumed was a safety improvement. It isn't — it's a string-prefix check, not a real IP/CIDR parse, so a hostname like `10.attacker.example` passes the validator (starts with `"10."`) but resolves via DNS to whatever the domain owner wants. Confirmed empirically: config with that hostname boots fine and `AssumeRole` genuinely succeeds against a target completely outside the intended private range, while the same target given as a literal non-matching IP is correctly rejected at config load.

Since this PR doesn't add an allowlist at all (`aws.endpoint`/`s3.endpoint.N` are treated as trusted admin config, same threat model as the AWS SDK's own `AWS_ENDPOINT_URL_STS` env var — never attacker-influenced input in UC's actual data plane), I'd suggest **not** adding one in the merged version either. A validator that looks like it restricts to private/loopback hosts but doesn't reliably do so is worse than no validator, because it creates false confidence for anyone reviewing the config surface. Your current design (trust the operator-supplied endpoint, no allowlist) is the safer choice here.

Separately, confirmed metadata-enumeration and cross-table credential-scope isolation are both solid in UC v0.6.0 core (unrelated to this PR): list endpoints correctly filter out objects the caller has no grant on, and a table-scoped vended credential can't reach a sibling table's prefix or the bucket root, even within the same external location. Evidence and full methodology: https://github.com/Linhares015/minerva/blob/main/spikes/003-catalog-auth/README.md#v03n

chrislusf:
Merge conflicts resolved. The FileIOFactory changes were ported onto the consolidated FileOperations path: per-bucket s3.endpoint.N now maps to Iceberg's S3FileIOProperties.ENDPOINT plus PATH_STYLE_ACCESS in getS3Config, and aws.endpoint still drives the STS client's endpointOverride. Added a FileOperationsTest case covering the endpoint config.

On the allowlist question: agreed with your conclusion — endpoint config stays trusted administrator configuration here, matching how AWS_ENDPOINT_URL/SDK endpoint env vars behave. A prefix-based host check would be both spoofable (10.attacker.example) and a false sense of security.

PR #1533, opened 2026-05-05
[Hadoop] Move UCHadoopConf and CredPropsUtil to hadoop module

openinx:
**PR Checklist**

**Description of changes**

- Moves `UCHadoopConf` and `CredPropsUtil` from the `spark` module into a new `hadoop` module so they can be reused by any connector (Spark, Flink, Trino, etc.) without pulling in Spark dependencies.
- Renames `CredentialSetting` → `HadoopCredentialConf` for clarity.
- Makes `catalogUri` and `scheme` required constructor parameters on `HadoopCredentialConf.builder(catalogUri, scheme)` so misconfiguration is caught at construction time rather than at build time.
- Moves `CredPropsUtil` to `io.unitycatalog.hadoop.internal` (not public API); public entry point is `HadoopCredentialConf`.
- Replaces `sessionHadoopFsImplProps()` (which used `session.conf.getOption` to avoid Hadoop defaults) with `sessionHadoopConf()`, which passes the full `sparkContext.hadoopConfiguration` directly to `HadoopCredentialConf.Builder#hadoopConf`. This is simpler and correct because `CredScopedFileSystem` is never written back into `sparkContext.hadoopConfiguration`, so recursive-wrapping was never possible.
- Updates all call sites in `UCSingleCatalog.scala` and all tests.


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConf.java
```diff
@@ -1,11 +1,11 @@
-package io.unitycatalog.spark;
+package io.unitycatalog.hadoop.internal;
 
 import io.unitycatalog.client.retry.JitterDelayRetryPolicy;
 import io.unitycatalog.client.retry.RetryPolicy;
 import org.apache.hadoop.conf.Configuration;
 
 public class UCHadoopConf {
-  private UCHadoopConf() {}
+  protected UCHadoopConf() {}
```
yili-db: Why protected? This class is full of static members.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -440,35 +436,16 @@ object UCSingleCatalog {
...
-            conf.getOption(key).orElse(conf.getOption("spark.hadoop." + key))
-              .filter(_ != credScopedFsClass) // skip if already CredScopedFileSystem (prevents recursive wrapping)
-              .map(key -> _)
-          }
-          .toMap.asJava
-    }
+  def sessionHadoopConf(): Configuration = {
+    SparkSession.getActiveSession
+      .map(_.sparkContext.hadoopConfiguration)
+      .getOrElse(new Configuration())
```
yili-db, line 445: why is this `new Configuration(/*true*/)` while everywhere else has `new Configuration(false)`?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -483,27 +504,27 @@ public static Map<String, String> createPathCredProps(
       case "s3":
```
timothyw553, line 497: just realized while reading this, we're missing "s3a" right?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredentialSetting.java
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * <pre>{@code
+ * Map<String, String> props = CredentialSetting.builder()
+ *     .catalogUri(uri)
+ *     .tokenProvider(tokenProvider)
+ *     .initialCredentials(creds)
+ *     .scheme("s3")
+ *     .enableCredentialRenewal(true)
+ *     .enableCredentialScopedFs(true)
+ *     .existingFsImplProperties(sessionFsImplProps)
```
timothyw553: nit: should be `.hadoopConf(Configuration)` instead of `.existingFsImplProperties(sessionFsImplProps)` right?

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/CredentialSettingTest.java
```diff
@@ -0,0 +1,245 @@
...
+        .isInstanceOf(IllegalStateException.class)
+        .hasMessageContaining("tokenProvider");
+  }
+
+  // ------- Helpers -------
+
+  /** Builder with credential renewal disabled (static creds). */
+  private static CredentialSetting.Builder staticBuilder() {
+    return CredentialSetting.builder().catalogUri("http://uc").enableCredentialRenewal(false);
+  }
```
timothyw553: All positive tests use this helper, so the renewal path is never tested. Can we add one test per scheme with `enableCredentialRenewal(true)` and a stub `TokenProvider`, asserting on the `fs.unitycatalog.*` keys?

timothyw553 (approved):
the public api makes sense to me. i've left a few comments (nit and add more tests)

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConf.java
openinx: by accident, reverted.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -11,22 +11,39 @@
...
 
+  private static final String CRED_SCOPED_FS_CLASS =
+      "io.unitycatalog.spark.fs.CredScopedFileSystem";
+  private static final String CRED_SCOPED_AFS_CLASS = "io.unitycatalog.spark.fs.CredScopedFs";
+  private static final String AWS_VENDED_TOKEN_PROVIDER_CLASS =
+      "io.unitycatalog.spark.auth.storage.AwsVendedTokenProvider";
+  private static final String GCS_VENDED_TOKEN_PROVIDER_CLASS =
+      "io.unitycatalog.spark.auth.storage.GcsVendedTokenProvider";
+  private static final String ABFS_VENDED_TOKEN_PROVIDER_CLASS =
+      "io.unitycatalog.spark.auth.storage.AbfsVendedTokenProvider";
```
openinx, line 40: Once we move those classes from the unitycatalog spark connector to this hadoop module, then we will need to rename those class names.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -11,22 +11,39 @@
...
+  private static final String GCS_ACCESS_TOKEN_KEY = "fs.gs.auth.access.token.credential";
+  private static final String GCS_ACCESS_TOKEN_EXPIRATION_KEY =
+      "fs.gs.auth.access.token.expiration";
+  private static final String ABFS_FIXED_SAS_TOKEN_KEY = "fs.azure.sas.fixed.token";
+
   private abstract static class PropsBuilder<T extends PropsBuilder<T>> {
-    private final ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+    private final HashMap<String, String> builder = new HashMap<>();
```
openinx, line 47: Since we don't have the guava now ( as the dependencies),  so we have to use the HashMap here directly.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: Because for the spark job classpath,  we should always use the classpath's hadoop conf, to initialize the filesystem.   Other places use `Configuration(false)` , because they are testing code, and we should exclude the classpath hadoop conf impact to the test correctness.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/CredentialSetting.java
openinx: updated, done.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/CredentialSettingTest.java
openinx: yes, that make sense to me. thanks for the comment.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/HadoopCredentialConf.java
```diff
@@ -0,0 +1,161 @@
...
+ *     .initialCredentials(creds)
+ *     .enableCredentialRenewal(true)
+ *     .enableCredentialScopedFs(true)
+ *     .hadoopConf(hadoopConf)
+ *     .buildForTable(tableId, TableOperation.READ_WRITE);
+ * }</pre>
+ *
+ * @since 0.5.0
+ */
+public final class HadoopCredentialConf {
```
tdas: didnt i say HadoopCredentialConf**s**?

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/CredPropsUtilTest.java
```diff
@@ -7,15 +7,15 @@
...
- * Verifies that {@link CredPropsUtil} saves the original {@code fs.<scheme>.impl} values under
- * {@code fs.<scheme>.impl.original} before overriding them with {@link
- * io.unitycatalog.spark.fs.CredScopedFileSystem}, so that the real delegate can be restored in
- * {@code CredScopedFileSystem#newFileSystem}.
+ * Verifies that {@link io.unitycatalog.hadoop.internal.CredPropsUtil} saves the original {@code
+ * fs.<scheme>.impl} values under {@code fs.<scheme>.impl.original} before overriding them with
+ * CredScopedFileSystem, so that the real delegate can be restored in {@code
+ * CredScopedFileSystem#newFileSystem}.
  */
 class CredPropsUtilTest {
```
tdas, line 20: shouldnt you match this file name to with util class name?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/HadoopCredentialConf.java
tdas: Lets just do UCCredentialHadoopConfs. 
UC in the name makes it clear in any user code in any connector that this is about UCCredentials.

tdas (approved):
Approved. but please rename before merging.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/CredPropsUtilTest.java
openinx: Hi @tdas ,  this is the test for CredPropsUtil ( the internal class), not for the public API class.  and the test class name is matching the `CredPropsUtilTest`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/HadoopCredentialConf.java
openinx: Yes,  thanks for the comment.  Let me update it.

openinx: Done via https://github.com/unitycatalog/unitycatalog/pull/1533/commits/f8ef823595f6f2c5c3664bcd29981c34493668d7

PR #1534, opened 2026-05-05
Run external-table and staging-table access-control suites against both UC REST and Delta REST surfaces

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1534/files) to review incremental changes.
- [**stack/create_table_tests**](https://github.com/unitycatalog/unitycatalog/pull/1534) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1534/files)]

---------
**PR Checklist**

**Description of changes**
Run external-table and staging-table access-control suites against both UC REST and Delta REST surfaces

`SdkExternalLocationAccessControlTest` parameterizes its existing 7-case table-creator loop and runs it for UC REST and Delta REST. `SdkStagingTableAccessControlTest` becomes abstract, with `SdkUCStagingTableAccessControlTest` and `SdkDeltaStagingTableAccessControlTest` providing surface-specific `createStaging` / `finalizeManagedTable`.

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/sdk/access/SdkExternalLocationAccessControlTest.java
+++ b/server/src/test/java/io/unitycatalog/server/sdk/access/SdkExternalLocationAccessControlTest.java
@@ -1,7 +1,6 @@
 package io.unitycatalog.server.sdk.access;
 
 import static io.unitycatalog.server.utils.TestUtils.assertApiException;
-import static io.unitycatalog.server.utils.TestUtils.assertPermissionDenied;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import io.unitycatalog.client.ApiClient;
@@ -22,7 +21,6 @@ import io.unitycatalog.client.model.DataSourceFormat;
 import io.unitycatalog.client.model.ExternalLocationInfo;
 import io.unitycatalog.client.model.ListExternalLocationsResponse;
 import io.unitycatalog.client.model.SecurableType;
-import io.unitycatalog.client.model.TableInfo;
 import io.unitycatalog.client.model.TableType;
 import io.unitycatalog.client.model.UpdateExternalLocation;
 import io.unitycatalog.client.model.VolumeInfo;
@@ -37,6 +35,7 @@ import lombok.Singular;
 import lombok.SneakyThrows;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.function.Executable;
 
 public class SdkExternalLocationAccessControlTest extends SdkAccessControlBaseCRUDTest {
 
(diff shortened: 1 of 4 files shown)


server/src/test/java/io/unitycatalog/server/sdk/access/SdkExternalLocationAccessControlTest.java
```diff
@@ -467,13 +493,47 @@ public void testCreateExternalTableVolumePermissions() throws Exception {
...
+   * different message conventions: {@code PERMISSION_DENIED} vs {@code PermissionDeniedException}).
+   */
+  private static void assertPermissionDenied(Executable executable) {
+    TestUtils.assertPermissionDenied(executable, "");
+  }
+
+  /** Minimum valid Delta CreateTableRequest for an EXTERNAL Delta table. */
+  private static io.unitycatalog.client.delta.model.CreateTableRequest deltaExternalTableRequest(
+      String name, String location) {
+    return new io.unitycatalog.client.delta.model.CreateTableRequest()
```
openinx, line 511: Why do we use the full qualified class name in this PR ?  does it conflicts with the existing classes ? this actually make the code hard to read.

yili-db: Many models in `delta.model.` are duplicating with existing non-Delta ones. So in many places they are used with FQCN.
Here CreateTableRequest might not have the same counterpart from all.yaml, but using FQCN actually helps so that we know it's for Delta.
Otherwise it will be `CreateTable` (for general UC) vs `CreateTableRequest` (for Delta).

server/src/test/java/io/unitycatalog/server/sdk/access/SdkStagingTableAccessControlTest.java
```diff
@@ -80,33 +81,24 @@ public void testManagedTableCreationByDifferentUserShouldFail() throws Exception
...
+        createStaging(
+            userAConfig, TestUtils.CATALOG_NAME, TestUtils.SCHEMA_NAME, TestUtils.TABLE_NAME);
+    assertThat(staging.location()).isNotNull();
 
     // Step 2: Test temporary credentials generation
     // User A (owner) should be able to get temporary credentials
     GenerateTemporaryTableCredential tempCredRequest =
         new GenerateTemporaryTableCredential()
-            .tableId(stagingTableInfo.getId())
+            .tableId(staging.id())
```
openinx: For the temp credentials generating API,  we also have two version of the API, right ?  for this test, do we also need to make it abstracted, so that we can use the corresponding APIs to do the test ? 

Seems like this PR is focusing on the staging related API, rather than the temporary credential API ?

server/src/test/java/io/unitycatalog/server/sdk/access/SdkStagingTableAccessControlTest.java
openinx: I see, we do have the permission check for the temporary table credentials generation.  then  I will recommend to cover them in both subclass tests.

openinx (approved):
LGTM.

tdas (approved):

server/src/test/java/io/unitycatalog/server/sdk/access/SdkExternalLocationAccessControlTest.java
```diff
@@ -467,13 +493,47 @@ public void testCreateExternalTableVolumePermissions() throws Exception {
...
       }
     }
   }
 
+  /**
+   * 403-only check that works for both UC REST and Delta error responses (their JSON bodies use
+   * different message conventions: {@code PERMISSION_DENIED} vs {@code PermissionDeniedException}).
+   */
+  private static void assertPermissionDenied(Executable executable) {
+    TestUtils.assertPermissionDenied(executable, "");
```
tdas: [AI Assisted] This file-local helper silently weakens assertions across the entire file, not just on the new Delta path.

`TestUtils.assertPermissionDenied(Executable)` enforces the literal `"PERMISSION_DENIED"` marker in the error body. By redirecting through `TestUtils.assertPermissionDenied(executable, "")`, the marker check is dropped — *any* `PERMISSION_DENIED` error code now passes, including ones thrown from a completely unrelated cause.

The unfortunate consequence is that this helper is also used by every existing UC-only call site in `testExternalLocationPermissions` (lines 221, 237, 264, 272, 285), so the assertion strength regresses for tests that have nothing to do with Delta.

Two options that both keep the UC arm strong:

- **Option A:** keep `TestUtils.assertPermissionDenied` (the strong, marker-checking version) imported as before for the UC-only call sites, and use a differently-named helper only inside `runExternalTableVolumePermissionsTest` for the surface-agnostic assertions.
- **Option B:** make this helper accept either surface's marker (`"PERMISSION_DENIED"` || `"PermissionDeniedException"`) so it stays strict on both surfaces. Probably worth lifting into `TestUtils` so other surface-agnostic suites can reuse it.

yili-db: Removed this helper. Now runExternalTableVolumePermissionsTest call TestUtils.assertPermissionDenied directly with different messages for old and new API.

PR #1536, opened 2026-05-05
Remove sbt-coursier and add sbt cache

yili-db:
**PR Checklist**

**Description of changes**

Remove sbt-coursier and add sbt cache

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -52,6 +52,7 @@ jobs:
       with:
         java-version: '17'
         distribution: 'temurin'
+        cache: 'sbt'
 
     - name: Resolve Delta master version and commit SHA
       if: matrix.delta-version == 'master-SNAPSHOT'
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -20,7 +20,10 @@ addSbtPlugin("com.github.sbt" % "sbt-pgp" % "2.1.2")
 
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "2.2.0")
 
-addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.3")
+// sbt-coursier 1.0.3 (pre-sbt-1.3 standalone plugin) intentionally omitted: sbt 1.9.x ships
+// coursier as the default resolver, with modern retry/backoff against Maven Central. Adding
+// the old plugin pins coursier to a 2018-era build that mishandles Sonatype's throttling and
+// breaks resolution under load.
 
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "3.11.3")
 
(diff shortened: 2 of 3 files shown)


project/plugins.sbt
```diff
@@ -20,7 +20,10 @@ addSbtPlugin("com.github.sbt" % "sbt-pgp" % "2.1.2")
 
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "2.2.0")
 
-addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.3")
+// sbt-coursier 1.0.3 (pre-sbt-1.3 standalone plugin) intentionally omitted: sbt 1.9.x ships
```
openinx: If we want to move this plugin, then we don't need add the extra comments.  just remove it.

.github/workflows/unit-tests.yml
```diff
@@ -52,6 +52,7 @@ jobs:
       with:
         java-version: '17'
         distribution: 'temurin'
+        cache: 'sbt'
```
openinx, line 55: why need this ?

yili-db: It's a feature of actions/setup-java@v4 that enables caching of sbt's dependency directories across workflow runs. It reduces requests to maven and speeds up build process.

It is not the root cause and real fix for this problem, but just an easy optimization that I picked up on the way.

project/plugins.sbt
yili-db: removed

project/plugins.sbt
```diff
@@ -20,8 +20,6 @@ addSbtPlugin("com.github.sbt" % "sbt-pgp" % "2.1.2")
 
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "2.2.0")
 
-addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.3")
```
openinx, line 23: For future reviewer context: 


sbt-coursier is an SBT plugin that replaces SBT's default dependency resolution (Apache Ivy) with Coursier (https://get-coursier.io/), a faster, parallel dependency resolver written in Scala.

  What it provides:
  - Faster dependency downloads via parallel resolution
  - Better artifact caching
  - More reliable resolution (handles eviction conflicts better than Ivy)
  - Support for alternative artifact repositories

  Why it's now redundant:

  Starting with SBT 1.3.0 (released 2019), Coursier is the default dependency resolver — it was merged directly into SBT core. The sbt-coursier plugin was only needed for SBT 0.13.x projects that wanted to opt
   in to Coursier early.

  So in any project on SBT 1.3+, installing the sbt-coursier plugin and setting useCoursier := true is a no-op at best, and at worst can cause version conflicts between the bundled Coursier and the one the
  plugin tries to bring in.

openinx (approved):
LGMT,  thanks for the work.

tdas (approved):

PR #1537, opened 2026-05-06
Delta: enforce case-insensitive column names

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1537/files) to review incremental changes.
- [**stack/column_case_ins**](https://github.com/unitycatalog/unitycatalog/pull/1537) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1537/files)]

---------
**PR Checklist**

**Description of changes**
Delta: enforce case-insensitive column names

Per Delta protocol ("All column names must be unique regardless of casing"). Fixes three case-sensitive sites in `ColumnUtils`: StructType duplicate detection and the duplicate-detection / lookup of partition-columns. Spec note added in `delta.yaml`.

### DIFF
--- a/api/delta.yaml
+++ b/api/delta.yaml
@@ -1330,12 +1330,16 @@ components:
     StructField:
       type: object
       description: |
-        A field in a StructType. Contains the field name, type, nullability,
-        and metadata matching the Delta schema field format.
+        A field in a StructType. Contains the field name, type, nullability, and metadata
+        matching the Delta schema field format.
+        Field names are case-preserving but case-insensitive: within a StructType no two fields
+        may have names equal under ASCII lowercase comparison. Every API reference to a column
+        name (partition-columns, delta.clustering.clusteringColumns, etc.) matches
+        case-insensitively. Per Delta [PROTOCOL.md](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#schema-serialization-format).
       properties:
         name:
           type: string
-          description: Column name
+          description: Column name. Case-preserving; unique case-insensitively within the parent StructType.
         type:
           $ref: '#/components/schemas/DeltaType'
         nullable:
--- a/server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
+++ b/server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
@@ -489,24 +489,22 @@ public class ColumnUtils {
     if (partitionColumns == null || partitionColumns.isEmpty()) {
       return;
     }
-    // Duplicate detection within partition-columns and the column-name lookup against the schema
-    // both compare case-insensitively, matching the Delta protocol rule that column names are
-    // unique regardless of casing.
-    Set<String> seenLower = new HashSet<>();
-    for (String partName : partitionColumns) {
-      if (!seenLower.add(partName.toLowerCase(Locale.ROOT))) {
-        throw new BaseException(
-            ErrorCode.INVALID_ARGUMENT,
-            "partition-columns contains duplicate entry (case-insensitive): " + partName);
-      }
-    }
     Map<String, ColumnInfo> columnsByLowerName =
         columns.stream()
             .collect(
                 Collectors.toMap(c -> c.getName().toLowerCase(Locale.ROOT), Function.identity()));
+    Set<String> seenLower = new HashSet<>();
     for (int i = 0; i < partitionColumns.size(); i++) {
       String partName = partitionColumns.get(i);
-      ColumnInfo match = columnsByLowerName.get(partName.toLowerCase(Locale.ROOT));
+      String lowerPartName = partName.toLowerCase(Locale.ROOT);
+      // Duplicate detection within partition-columns, comparing case-insensitively.
+      if (!seenLower.add(lowerPartName)) {
+        throw new BaseException(
+            ErrorCode.INVALID_ARGUMENT,
+            "partition-columns contains duplicate entry (case-insensitive): " + partName);
+      }
+      // Column-name lookup against the schema, comparing case-insensitively.
+      ColumnInfo match = columnsByLowerName.get(lowerPartName);
       if (match == null) {
         throw new BaseException(
             ErrorCode.INVALID_ARGUMENT,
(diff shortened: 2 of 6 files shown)


server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -484,19 +489,24 @@ public static void applyPartitionColumns(
...
-            "partition-columns contains duplicate entry: " + partName);
+            "partition-columns contains duplicate entry (case-insensitive): " + partName);
       }
     }
-    Map<String, ColumnInfo> columnsByName =
-        columns.stream().collect(Collectors.toMap(ColumnInfo::getName, Function.identity()));
+    Map<String, ColumnInfo> columnsByLowerName =
+        columns.stream()
+            .collect(
+                Collectors.toMap(c -> c.getName().toLowerCase(Locale.ROOT), Function.identity()));
```
timothyw553, line 495: same case-insensistive names (i.e. col and COL) will error deeper , could we just throw here?
```suggestion
                Collectors.toMap(c -> c.getName().toLowerCase(Locale.ROOT), Function.identity()));
                 Collectors.toMap(
                     c -> c.getName().toLowerCase(Locale.ROOT),
                     Function.identity(),
                     (a, b) -> {
                       throw new BaseException(
                           ErrorCode.INVALID_ARGUMENT,
                           "Duplicate column name (case-insensitive): " + b.getName());
                     }));
```

timothyw553 (approved):

api/delta.yaml
```diff
@@ -1330,12 +1330,16 @@ components:
     StructField:
       type: object
       description: |
-        A field in a StructType. Contains the field name, type, nullability,
-        and metadata matching the Delta schema field format.
+        A field in a StructType. Contains the field name, type, nullability, and metadata
+        matching the Delta schema field format.
+        Field names are case-preserving but case-insensitive: within a StructType no two fields
```
openinx, line 1335: so we use the `case-insentitive`,  because we want to deduplicate the `field` that with the same lowercase name ?

api/delta.yaml
```diff
@@ -1330,12 +1330,16 @@ components:
...
+        may have names equal under ASCII lowercase comparison. Every API reference to a column
+        name (partition-columns, delta.clustering.clusteringColumns, etc.) matches
+        case-insensitively. Per Delta [PROTOCOL.md](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#schema-serialization-format).
```
openinx, line 1338: Just to confirm, the delta properties and values I cleary see: `Widely supported values (matched case-insensitively): uncompressed/none (no compression), snappy, gzip, lz4 (deprecated, Hadoop framing), lz4_raw ([LZ4 block format](https://parquet.apache.org/docs/file-format/data-pages/compression/#lz4_raw)), zstd.` in here https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-properties

About the field name, I did not see the case insensitive description.  Are they also case-insensitive in delta protocol ?

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -233,15 +237,16 @@ public static void validateStructType(StructType structType, String path) {
...
     }
-    Set<String> seen = new HashSet<>();
+    Set<String> seenLower = new HashSet<>();
     for (int i = 0; i < fields.size(); i++) {
       String fieldPath = path + ".fields[" + i + "]";
       requireNonNull(fields.get(i), fieldPath);
       validateStructField(fields.get(i), fieldPath);
       String name = fields.get(i).getName();
-      if (!seen.add(name)) {
+      if (!seenLower.add(name.toLowerCase(Locale.ROOT))) {
```
openinx, line 246: Could you pls check all the `StructField.getName` places which need toLowerCase or not ?  I just see this may also need the lower case: https://github.com/unitycatalog/unitycatalog/blob/d7b6d7a9ff9f42a57dfb43b7379925ff709b60b2/server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java#L186-L191

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -484,19 +489,24 @@ public static void applyPartitionColumns(
...
       return;
     }
-    Set<String> seen = new HashSet<>();
+    // Duplicate detection within partition-columns and the column-name lookup against the schema
+    // both compare case-insensitively, matching the Delta protocol rule that column names are
+    // unique regardless of casing.
+    Set<String> seenLower = new HashSet<>();
     for (String partName : partitionColumns) {
-      if (!seen.add(partName)) {
+      if (!seenLower.add(partName.toLowerCase(Locale.ROOT))) {
```
openinx: Instead of lowercase of `partitionName` everywhere, why not just lower case when we firstly construct the `partitionColumns`  ?

openinx (approved):
Left few comments.  But pls address the comments,  my only concern is: lower the case everywhere can easily introduce the bug by missing attention, and if we could do it in general way then it's less risky to create the bugs.

api/delta.yaml
yili-db: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#schema-serialization-format:~:text=All%20column%20names%20must%20be%20unique%20regardless%20of%20casing
```
All column names must be unique regardless of casing
```

api/delta.yaml
yili-db: Yes that is defined by Delta.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
yili-db: This one doesn't. It's comparing 
`ColumnInfo.name` to `ColumnInfo.typeJson.field.name` of the same `ColumnInfo`. Both are set by clients in the old UC API so there is no reason the casing should be different.
This is the `case-preserving` part.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
yili-db: Merged into the next loop.

server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
yili-db: This function `applyPartitionColumns` only runs when there are partition columns.
The actual dedup of columns are done by `ColumnUtils.validateStructType(columns, "columns");`

tdas (approved):

PR #1538, opened 2026-05-06
[Hadoop] Move credential providers and CredScopedFileSystem to hadoop module

timothyw553:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1538/files) to review incremental changes.
- [**stack/uc-hadoop-move-only**](https://github.com/unitycatalog/unitycatalog/pull/1538) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1538/files)]
  - [stack/uc-hadoop-drc-creds](https://github.com/unitycatalog/unitycatalog/pull/1527) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1527/files/77fa2204d48d4313d8d1f85a07cffe4d15179b52..37edbff97427b0173b8ffdb3805eb76961f0ab3a)]

---------
**PR Checklist**

**Description of changes**

Moves the Hadoop credential providers and credential-scoped filesystem from `connectors/spark` to `connectors/hadoop/internal/auth` and `connectors/hadoop/internal/fs` so non-Spark connectors (Flink, Trino, Delta OSS) can reuse them without a Spark dependency.

Moved files (no behavior change):
- `auth/{Aws,Abfs,Gcs}VendedTokenProvider`, `auth/GenericCredential`, `auth/GenericCredentialProvider`
- `fs/CredScopedFileSystem`, `fs/CredScopedFs`, `fs/CredScopedKey`

Plumbing:
- `internal/CredPropsUtil` class-name string constants point at the relocated classes.
- Spark module deletes the originals and updates imports.
- `GenericCredentialProvider` builds its `TemporaryCredentialsApi` via `ApiClientBuilder` directly (drops the Spark-only `ApiClientFactory` for the renewal path).
- Replaces Spark-shaded Guava cache with `BoundedKeyedCache` from #1539.


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/credentials/AbfsVendedTokenProvider.java
```diff
@@ -1,10 +1,10 @@
-package io.unitycatalog.spark.auth.storage;
+package io.unitycatalog.hadoop.internal.credentials;
```
openinx: nit: will it be better for us to keep the `internal.auth`, since the `credentials` is the name the can cover all the classes under the `unitycatalog-hadoop` package ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/credentials/GenericCredentialProvider.java
```diff
@@ -102,7 +101,11 @@ protected TemporaryCredentialsApi temporaryCredentialsApi() {
           RetryPolicy retryPolicy = UCHadoopConf.createRequestRetryPolicy(conf);
           tempCredApi =
               new TemporaryCredentialsApi(
-                  ApiClientFactory.createApiClient(retryPolicy, ucUri, tokenProvider));
+                  ApiClientBuilder.create()
```
openinx: Critical:  this change actually is losing the `spark` and `delta` version that we intentionally added in the `ApiClientFactory`, which will make us hard to trace the client version info from the unitycatalog server side. 

we should figure out how to fix this.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -31,14 +30,15 @@
...
+ *       io.unitycatalog.hadoop.internal.CredPropsUtil} sets {@code
+ *       fs.<scheme>.impl.disable.cache=true} so that Hadoop always instantiates a fresh {@code
+ *       CredScopedFileSystem} for each file access. Because {@code CredScopedFileSystem} is a thin,
+ *       stateless wrapper, this is cheap.
  *   <li><b>Global credential-scoped cache for the real delegate.</b> {@code CredScopedFileSystem}
  *       maintains a static {@link #CACHE} keyed by {@link CredScopedKey}, which encodes the
  *       credential scope (table ID + operation, or path + operation). On each {@link
  *       #initialize(URI, Configuration)} call the key is derived from the Hadoop {@link
- *       Configuration} injected by {@link io.unitycatalog.spark.auth.CredPropsUtil}, and the
+ *       Configuration} injected by {@link io.unitycatalog.hadoop.internal.CredPropsUtil}, and the
```
openinx, line 41: Thanks for the fixing this broken link.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
+
+public final class BoundedKeyedCache<K, V> {
```
openinx, line 11: Can you file a separate PR which focusing on this cache class only ?  this need to be well-reviewed and carefully tested.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -0,0 +1,134 @@
...
+      ref.lock.unlock();
+      releaseLock(key, ref);
+    }
+  }
+
+  public void put(K key, V value) {
+    Objects.requireNonNull(key, "key cannot be null");
+    Objects.requireNonNull(value, "value cannot be null");
+    List<V> evicted = new ArrayList<>();
+    synchronized (cacheLock) {
```
openinx, line 54: be careful,  you will need the global cache lock to do the cache insertion, which will limit the entire process throughput, and will become the performence bottleneck, since every io request that requirs new filessystem will go through this code path.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/credentials/GenericCredentialProvider.java
openinx: one approach comes in my mind is: define a separate API in the `UCCredentialHadoopConfs`, and with this you can set the engine version info ( eventually it will turn into the hadoop conf in the engine side ). 

and inside the token provider implementation,  you can parse those hadoop conf keys directly, to get those version info, and inject to the http requests.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/credentials/AbfsVendedTokenProvider.java
timothyw553: done. renamed to `io.unitycatalog.hadoop.internal.auth`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/credentials/GenericCredentialProvider.java
timothyw553: done. added `Builder.addEngineVersion(name, version)`; the renewal client reads the conf and calls `addAppVersion`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
timothyw553: done. split into #1539.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
timothyw553: done. moved to #1539 with per-key locking instead of a global lock.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/credentials/GenericCredentialProvider.java
timothyw553: done. implemented as you suggested.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
```diff
@@ -100,9 +100,18 @@ protected TemporaryCredentialsApi temporaryCredentialsApi() {
...
+              ApiClientBuilder.create()
+                  .uri(ucUri)
+                  .tokenProvider(tokenProvider)
+                  .retryPolicy(retryPolicy);
+          // Propagate engine version metadata into the User-Agent header so the UC server can
+          // trace which engine versions are calling. Engines register these via
+          // UCCredentialHadoopConfs.Builder#addEngineVersion, which writes them into the Hadoop
+          // configuration under UC_ENGINE_VERSION_PREFIX.
+          conf.getPropsWithPrefix(UCHadoopConf.UC_ENGINE_VERSION_PREFIX)
+              .forEach(builder::addAppVersion);
```
openinx: For here,  actually,  we are duplicating the same code in `ApiClientFactory#createApiClient`,  they may not exact the same, but most of the code are the same. 

the difference is 
1.  Here,  basically we use the hadoop conf to initialize the ApliClient. 
2.  Inside the ApiClientFactory, we use the concret parameters to create the ApiClient. 

My concern is:   if someone just change the ApliClientFactory, but forget to change here.  then  we will miss something, and will cause the serious bug. 

Any thought to make them to be a unified method, so that we can always share the same logic there ?

connectors/spark/src/main/java/io/unitycatalog/spark/ApiClientFactory.java
```diff
@@ -38,6 +39,32 @@ public static ApiClient createApiClient(
...
+      builder.addEngineVersion("Delta", deltaVersion);
+    }
+    String javaVersion = getJavaVersion();
+    if (javaVersion != null) {
+      builder.addEngineVersion("Java", javaVersion);
+    }
+    String scalaVersion = getScalaVersion();
+    if (scalaVersion != null) {
+      builder.addEngineVersion("Scala", scalaVersion);
+    }
```
openinx: Could you pls separate a method to get the app engine version info, so that we can make it share between `createApiClient` and `withEngineVersions` ? 

```
 public static Map<String, String> appEngines(){
     // Add Spark, Delta,Java, Scala version.
  }
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
```diff
@@ -111,6 +115,19 @@ public Builder hadoopConf(Configuration conf) {
...
 
+    /**
+     * Records an engine version (e.g. {@code addEngineVersion("Spark", "4.0.0")}) to be propagated
+     * to the User-Agent header on UC API calls so the server can trace which engine versions are
+     * calling. May be called multiple times. Engines should typically register their own version
+     * plus any relevant runtime versions (Delta, Java, Scala, etc.).
+     */
+    public Builder addEngineVersion(String name, String version) {
+      Preconditions.checkArgument(name != null && !name.isEmpty(), "engine version name required");
+      Preconditions.checkArgument(version != null, "engine version value for '%s' required", name);
```
openinx: the version cann't be empty string also ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
```diff
@@ -139,16 +157,27 @@ public Map<String, String> buildForTable(String tableId, TableOperation tableOpe
...
+    }
+
+    private Map<String, String> withEngineVersions(Map<String, String> props) {
+      if (engineVersions.isEmpty()) {
+        return props;
+      }
+      Map<String, String> merged = new HashMap<>(props);
+      engineVersions.forEach(
+          (name, version) -> merged.put(UCHadoopConf.UC_ENGINE_VERSION_PREFIX + name, version));
+      return java.util.Collections.unmodifiableMap(merged);
```
openinx: Nit:  Pls don't use the full qualified class name, unless we have a class name conflict in the same class. otherwise it makes some noisy to read.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
```diff
@@ -111,6 +115,19 @@ public Builder hadoopConf(Configuration conf) {
       return this;
     }
 
+    /**
+     * Records an engine version (e.g. {@code addEngineVersion("Spark", "4.0.0")}) to be propagated
+     * to the User-Agent header on UC API calls so the server can trace which engine versions are
+     * calling. May be called multiple times. Engines should typically register their own version
+     * plus any relevant runtime versions (Delta, Java, Scala, etc.).
+     */
+    public Builder addEngineVersion(String name, String version) {
```
openinx: As you added the new API in this UCCredentialHadoopConfs,  you will need to add the tests in the `UCCredentialHadoopConfs`, so that we can confirm the `buildForTable` and `buildForPath` are correctly generating the expected hadoop props.

connectors/spark/src/main/java/io/unitycatalog/spark/ApiClientFactory.java
timothyw553: done. `createApiClient` and `withEngineVersions` now share the same version helper.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
timothyw553: done. empty engine version values are now rejected.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
timothyw553: done. replaced the fully qualified `java.util.Collections` usage with an import.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
timothyw553: done. added tests for engine version propagation in both `buildForTable` and `buildForPath`.

connectors/spark/src/main/java/io/unitycatalog/spark/ApiClientFactory.java
timothyw553: done. added spark-side test coverage for `withEngineVersions`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
timothyw553: done. shared `ApiClient` setup now lives in `ApiClients` and is used by both spark and hadoop renewal paths.

clients/java/src/main/java/io/unitycatalog/client/ApiClients.java
```diff
@@ -0,0 +1,24 @@
+package io.unitycatalog.client;
+
+import io.unitycatalog.client.auth.TokenProvider;
+import io.unitycatalog.client.internal.Preconditions;
+import io.unitycatalog.client.retry.RetryPolicy;
+import java.net.URI;
+import java.util.Map;
+
+public final class ApiClients {
```
openinx: we shouldn't expose this as a module level public API,  pls move this into `io.unitycatalog.client.internal` package , name it as `ApiClientUtils`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
```diff
@@ -139,16 +159,27 @@ public Map<String, String> buildForTable(String tableId, TableOperation tableOpe
...
+              initialCredentials));
+    }
+
+    private Map<String, String> withEngineVersions(Map<String, String> props) {
+      if (engineVersions.isEmpty()) {
+        return props;
+      }
+      Map<String, String> merged = new HashMap<>(props);
+      engineVersions.forEach(
+          (name, version) -> merged.put(UCHadoopConf.UC_ENGINE_VERSION_PREFIX + name, version));
```
openinx: If you always want to add the `UCHadoopConf.UC_ENGINE_VERSION_PREFIX` to the engine name, will it be better to add the prefix in `addEngineVersion` directly ?  

I mean for every added engine name, it will automatically and always add the prefix, then it will simplify your following work. and everywhere will always have the consistent view about them.

connectors/spark/src/main/java/io/unitycatalog/spark/ApiClientFactory.java
```diff
@@ -13,29 +17,33 @@ private ApiClientFactory() {}
   public static ApiClient createApiClient(
       RetryPolicy retryPolicy, URI uri, TokenProvider tokenProvider) {
```
openinx: nit: the extra blank line can be removed.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -257,7 +257,7 @@ class UCSingleCatalog
       new GenerateTemporaryTableCredential()
         .tableId(tableId).operation(TableOperation.READ_WRITE))
     val tableUriScheme = new Path(tableLocation).toUri.getScheme
-    val credentialProps = UCCredentialHadoopConfs.builder(uri.toString, tableUriScheme)
+    val credentialProps = ApiClientFactory.withEngineVersions(UCCredentialHadoopConfs.builder(uri.toString, tableUriScheme))
```
openinx: This actually make the code a bit more complicated to read. 

How about exposing the `addEngineVersions(Map<String, String>)` rather than the `addEngineVersion(String, String)`  ( you can still keep the private `addEngineVersion(String, String)` inside the `UCCredentialHadoopConfs.Builder` for implement the `addEngineVersions`) ? 

with this,  you can just chain the builder here: 

```java
UCCredentialHadoopConfs
      .builder(uri.toString, CatalogUtils.stringToURI(stagingLocation).getScheme)
      .addEngineVersions(ApiClientFactory.appEngineVersions())
      .tokenProvider(tokenProvider)
      .initialCredentials(temporaryCredentials)
      .enableCredentialRenewal(renewCredEnabled)
      .enableCredentialScopedFs(credScopedFsEnabled)
      .hadoopConf(UCSingleCatalog.sessionHadoopConf())
      .buildForTable(stagingTableId, TableOperation.READ_WRITE)
```

clients/java/src/main/java/io/unitycatalog/client/ApiClients.java
timothyw553: done. moved this to internal `ApiClientUtils` and updated callers.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
timothyw553: done. engine version keys are now prefixed inside the builder.

connectors/spark/src/main/java/io/unitycatalog/spark/ApiClientFactory.java
timothyw553: done. removed the extra blank line.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: done. added `addEngineVersions(Map<String, String>)` and updated spark call sites to chain it directly.

openinx (approved):
LGTM,  thanks @TimothyW553 for the work.

PR #1539, opened 2026-05-07
[Hadoop] Add BoundedKeyedCache and CloseableUtils

timothyw553:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1539/files) to review incremental changes.
- [**stack/uc-hadoop-cache**](https://github.com/unitycatalog/unitycatalog/pull/1539) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1539/files)]
  - [stack/uc-hadoop-move-only](https://github.com/unitycatalog/unitycatalog/pull/1538) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1538/files/f3fe43e3b95da1f0047f07f5a4b6e4fdf8a16ce8..633daf3c4edc2983c31682a1d74d0e952b742b82)]
    - [stack/uc-hadoop-drc-creds](https://github.com/unitycatalog/unitycatalog/pull/1527) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1527/files/633daf3c4edc2983c31682a1d74d0e952b742b82..5c73ae379e949cdeb60a4b2f73482456e3b881da)]

---------
**PR Checklist**

**Description of changes**

Adds two utility classes to the `unitycatalog-hadoop` module so subsequent stacked PRs can move credential providers off Spark-shaded Guava:

- `internal/util/BoundedKeyedCache` — bounded LRU keyed cache built on JDK collections, with per-key locking via `withKeyLock` / `getOrLoad` and an optional eviction listener (used to close evicted FileSystems).
- `internal/util/CloseableUtils` — `closeQuietly(Closeable)` helper used as the default eviction action.

Both classes are package-private under `internal/util`. Tests cover LRU eviction order, eviction-listener invocation (null-safe), and per-key serialization in `getOrLoad`.

Pulled out into its own PR per @openinx's request so the cache can be reviewed and tested in isolation before #1538 starts using it.

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
@@ -1,0 +1,87 @@
+package io.unitycatalog.hadoop.internal.util;
+
+import io.unitycatalog.client.internal.Preconditions;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Consumer;
+
+public final class BoundedKeyedCache<K, V> {
+  private final int maxSize;
+  private final Consumer<V> evictionListener;
+  private final Object cacheLock = new Object();
+  private final LinkedHashMap<K, V> cache = new LinkedHashMap<>(16, 0.75f, true);
+  private final IdLockMap<K> keyLocks = new IdLockMap<>();
+
+  public BoundedKeyedCache(int maxSize) {
+    this(maxSize, null);
+  }
+
+  public BoundedKeyedCache(int maxSize, Consumer<V> evictionListener) {
+    Preconditions.checkArgument(maxSize > 0, "maxSize must be positive, got %s", maxSize);
+    this.maxSize = maxSize;
+    this.evictionListener = evictionListener == null ? value -> {} : evictionListener;
+  }
+
+  public V getIfPresent(K key) {
+    synchronized (cacheLock) {
+      return cache.get(key);
+    }
+  }
+
+  public <E extends Exception> V getOrLoad(K key, CheckedSupplier<V, E> loader) throws E {
+    V cached = getIfPresent(key);
+    if (cached != null) {
+      return cached;
+    }
+    try (IdLockMap<K>.IdLock ignored = keyLocks.acquire(key)) {
+      V lockedCached = getIfPresent(key);
+      if (lockedCached != null) {
+        return lockedCached;
+      }
+      V loaded = Objects.requireNonNull(loader.get(), "loader returned null");
+      put(key, loaded);
+      return loaded;
+    }
+  }
+
+  public void put(K key, V value) {
+    Objects.requireNonNull(key, "key cannot be null");
+    Objects.requireNonNull(value, "value cannot be null");
+    List<V> evicted = new ArrayList<>();
+    synchronized (cacheLock) {
+      V previous = cache.put(key, value);
+      if (previous != null && previous != value) {
+        evicted.add(previous);
+      }
+      while (cache.size() > maxSize) {
+        Map.Entry<K, V> eldest = cache.entrySet().iterator().next();
+        cache.remove(eldest.getKey());
+        evicted.add(eldest.getValue());
+      }
+    }
+    evicted.forEach(evictionListener);
+  }
+
+  public void clear() {
+    List<V> evicted;
+    synchronized (cacheLock) {
+      evicted = new ArrayList<>(cache.values());
+      cache.clear();
+    }
+    evicted.forEach(evictionListener);
+  }
+
+  int size() {
+    synchronized (cacheLock) {
+      return cache.size();
+    }
+  }
+
+  @FunctionalInterface
+  public interface CheckedSupplier<T, E extends Exception> {
+    T get() throws E;
+  }
+}
--- a/connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/util/IdLockMapTest.java
+++ b/connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/util/IdLockMapTest.java
@@ -1,0 +1,164 @@
+package io.unitycatalog.hadoop.internal.util;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.jupiter.api.Test;
+
+class IdLockMapTest {
+
+  @Test
+  void acquireRejectsNullKey() {
+    IdLockMap<String> locks = new IdLockMap<>();
+
+    assertThatThrownBy(() -> locks.acquire(null))
+        .isInstanceOf(NullPointerException.class)
+        .hasMessageContaining("key cannot be null");
+  }
+
+  @Test
+  void closeRemovesUnusedLock() {
+    IdLockMap<String> locks = new IdLockMap<>();
+
+    try (IdLockMap<String>.IdLock ignored = locks.acquire("k")) {
+      assertThat(locks.lockCount()).isEqualTo(1);
+      assertThat(locks.referenceCount("k")).isEqualTo(1);
+    }
+
+    assertThat(locks.lockCount()).isZero();
+    assertThat(locks.referenceCount("k")).isZero();
+  }
+
+  @Test
+  void sameKeyWaiterDoesNotEnterUntilCurrentHolderCloses() throws Exception {
+    IdLockMap<String> locks = new IdLockMap<>();
+    CountDownLatch waiterStarted = new CountDownLatch(1);
+    CountDownLatch waiterEntered = new CountDownLatch(1);
+    ExecutorService executor = Executors.newSingleThreadExecutor();
+
+    try (IdLockMap<String>.IdLock ignored = locks.acquire("k")) {
+      Future<?> waiter =
+          executor.submit(
+              () -> {
+                waiterStarted.countDown();
+                try (IdLockMap<String>.IdLock ignoredWaiter = locks.acquire("k")) {
+                  waiterEntered.countDown();
+                }
+                return null;
+              });
+
+      assertThat(waiterStarted.await(5, TimeUnit.SECONDS)).isTrue();
+      waitUntil(() -> locks.referenceCount("k") == 2);
+      assertThat(waiterEntered.await(100, TimeUnit.MILLISECONDS)).isFalse();
+
+      ignored.close();
+      waiter.get(5, TimeUnit.SECONDS);
+    } finally {
+      executor.shutdownNow();
+    }
+
+    assertThat(locks.lockCount()).isZero();
+  }
+
+  @Test
+  void differentKeysProgressIndependently() throws Exception {
+    IdLockMap<String> locks = new IdLockMap<>();
+    CountDownLatch otherKeyEntered = new CountDownLatch(1);
+    ExecutorService executor = Executors.newSingleThreadExecutor();
+
+    try (IdLockMap<String>.IdLock ignored = locks.acquire("a")) {
+      Future<?> otherKey =
+          executor.submit(
+              () -> {
+                try (IdLockMap<String>.IdLock ignoredOther = locks.acquire("b")) {
+                  otherKeyEntered.countDown();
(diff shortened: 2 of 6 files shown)


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.function.Consumer;
+
+public final class BoundedKeyedCache<K, V> {
+  private final int maxSize;
+  private final Consumer<V> evictionListener;
+  private final Object cacheLock = new Object();
+  private final LinkedHashMap<K, V> cache = new LinkedHashMap<>(16, 0.75f, true);
+  private final ConcurrentHashMap<K, RefCountedLock> keyLocks = new ConcurrentHashMap<>();
+
+  public BoundedKeyedCache(long maxSize, Consumer<V> evictionListener) {
```
openinx: ideally, the maxSize would be integer,  unlikely we will have more keys which exceed the integer.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -0,0 +1,134 @@
...
+    if (maxSize <= 0 || maxSize > Integer.MAX_VALUE) {
+      throw new IllegalArgumentException("maxSize must be between 1 and " + Integer.MAX_VALUE);
+    }
```
openinx: we can use the `Preconditions.checkArgument` directly here.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
timothyw553: done changed to interger

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
timothyw553: sure, change to preconditions

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
+
+public final class BoundedKeyedCache<K, V> {
+  private final int maxSize;
+  private final Consumer<V> evictionListener;
+  private final Object cacheLock = new Object();
```
openinx, line 14: If you want a instance level lock, you can always use the `synchroized(this)` ,right ? then we actually don't need the extra `Object cacheLock` .

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -0,0 +1,133 @@
...
+  private final LinkedHashMap<K, V> cache = new LinkedHashMap<>(16, 0.75f, true);
+  private final ConcurrentHashMap<K, RefCountedLock> keyLocks = new ConcurrentHashMap<>();
+
+  public BoundedKeyedCache(int maxSize, Consumer<V> evictionListener) {
+    Preconditions.checkArgument(maxSize > 0, "maxSize must be positive, got %s", maxSize);
+    this.maxSize = maxSize;
+    this.evictionListener = evictionListener == null ? value -> {} : evictionListener;
```
openinx, line 25: nit: will it be more clear if we introduce a separate `BoundedKeyedCache(int maxSize)` constructor ?  I think you are using it for testing purpose.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  public V getIfPresent(K key) {
```
openinx, line 28: nit: this should remove the `public` modifier, we'd better to make it as package-access, I understand you want to use it for testing.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -0,0 +1,133 @@
...
+    synchronized (cacheLock) {
+      return cache.get(key);
+    }
+  }
+
+  public <E extends Exception> V getOrLoad(K key, CheckedSupplier<V, E> loader) throws E {
```
openinx, line 34: I don't think it's worth to introduce an extra generic `E` for abstracting the Exception class.   you can always throw the default runtime exception. then we don't even need to add the `throws E` in the method signature.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -0,0 +1,133 @@
...
+    V cached = getIfPresent(key);
+    if (cached != null) {
+      return cached;
+    }
+    return withKeyLock(
```
openinx: Actually, you can make the `IdLock/keyLock extends the AutoClosable` interface, then you can use the following directly to simplify the code. 

```java
try(IdLock ignored =  getIdLock(key)) {
         V lockedCached = getIfPresent(key);
        if (lockedCached != null) {
          return lockedCached;
        }
        V loaded = Objects.requireNonNull(loader.get(), "loader returned null");
        put(key, loaded);
        return loaded;
}
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -0,0 +1,133 @@
...
+  @FunctionalInterface
+  public interface CheckedSupplier<T, E extends Exception> {
+    T get() throws E;
+  }
+
+  private static final class RefCountedLock {
+    private final ReentrantLock lock = new ReentrantLock();
+    private int references;
+  }
+}
```
openinx: I will suggest you to separate the id lock map as a separate class, so that you can add every comprehensive test cases for this id lock map, and accurate assert every ref count change.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
openinx: And this `getOrLoad` is the only public API , even for inernal usage.  all others should be package-access, for internal testing purpose.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
openinx: I just tried to write this IdLockMap.

```java
public static class IdLockMap<K>{
    public final ConcurrentHashMap<K, IdLock> map = new ConcurrentHashMap<>();

    public IdLock get(K key) {
      return map.compute(key, (k, v) -> {
        if(v == null) {
          return new IdLock(key);
        } else {
          v.ref().incrementAndGet();
          return v;
        }
      });
    }

    public class IdLock implements AutoCloseable {
      private final K key;
      private final ReentrantLock lock = new ReentrantLock();
      private final AtomicInteger ref = new AtomicInteger(1);

      public IdLock(K key) {
        lock.lock();
        this.key = key;
      }

      @Override
      public void close() {
        map.computeIfPresent(key, (k, v) -> {
          if (v != this) {
            return v;
          } else {
            lock.unlock();
            return ref.decrementAndGet() != 0 ? v : null;
          }
        });
      }

      public AtomicInteger ref(){
        return ref;
      }
    }
  }

  {
    IdLockMap<Integer> locks = new IdLockMap<>();
    try(IdLockMap<Integer>.IdLock ignored = locks.get(1)){
      // do something within lock context.
    }
  }
```

openinx: The sample usage is: 

```java
    IdLockMap<Integer> locks = new IdLockMap<>();
    try(IdLockMap<Integer>.IdLock ignored = locks.get(1)){
      // do something within lock context.
    }
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
timothyw553: done. Added `BoundedKeyedCache(int maxSize)` with a no-op eviction listener.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
timothyw553: done. Made the test-only helpers package-private and kept required cache APIs public for cross-package use.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
timothyw553: done. `getOrLoad` now uses try-with-resources for the per-key lock.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
timothyw553: done. Extracted `IdLockMap` and added direct tests for ref counts and same-key locking. and the new `IdLockMap` supports this try-with-resources usage.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
timothyw553: I kept the private `cacheLock` because `synchronized(this)` lets outside code block cache internals; the private lock is safer and still simple.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
timothyw553: I kept the checked exception type so callers can propagate `IOException` directly instead of wrapping it in `RuntimeException`.

timothyw553: I made test-only helpers package-private, but kept `getIfPresent`, `put`, and `clear` public because the stacked PR uses them across internal packages.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -0,0 +1,94 @@
...
+    synchronized (cacheLock) {
+      return cache.size();
+    }
+  }
+
+  Collection<V> values() {
+    synchronized (cacheLock) {
+      return new ArrayList<>(cache.values());
+    }
+  }
```
openinx: Is this method still useful?  seems like we can remove it now.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/util/IdLockMapTest.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    assertThat(locks.lockCount()).isZero();
+  }
+
+  @Test
+  void differentKeysProgressIndependently() throws Exception {
+    IdLockMap<String> locks = new IdLockMap<>();
+    CountDownLatch otherKeyEntered = new CountDownLatch(1);
+    ExecutorService executor = Executors.newSingleThreadExecutor();
```
openinx, line 77: I think we'd better to have at least one case to cover the multi-threads to create the intensive contention while accessing the id lock map.

openinx (approved):
Generally looks good to me.  left one comment for adding the contention test via multi-threads.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
timothyw553: removed

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/util/IdLockMapTest.java
timothyw553: added a high-contention test where 64 threads queue on the same key lock and verifies only one thread enters at a time plus lock refs clean up.

PR #1177, opened 2025-11-10
[java client] Handle unknown enum values gracefully in the java client

tdas:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

For various enums generated from the open api spec, its possible that an implementation of the spec may have more values than the ones defined in the spec. so its better to make the client not fail on encountaring such unknown values in RPC responses. Rather it should handle gracefully to produce the enum value of "UNKNOWN".

PR #1541, opened 2026-05-07
[Spark] Fall back to BinaryType for unrecognized UC column types

### LINKED ISSUE #1077: GEOMETRY column type breaks all access to table from Spark
**Describe the bug**

I have a Databricks table with a column using the new [GEOMETRY type](https://docs.databricks.com/aws/en/sql/language-manual/data-types/geometry-type). When I try to access this table from a Spark workload, I am not able to describe the table or operate on any of its columns.

**To Reproduce**

1. Spark config:
```
spark.sql.catalog.spark_catalog org.apache.spark.sql.delta.catalog.DeltaCatalog
spark.sql.catalog.my_catalog io.unitycatalog.spark.UCSingleCatalog
spark.sql.catalog.my_catalog.uri <Databricks workspace URL>/api/2.1/unity-catalog
spark.sql.catalog.my_catalog.token <Databricks PAT>
```
2. Try to run SQL queries in Spark that access the table or its metadata:
    1. `LIST TABLES in my_catalog.default`
    2. `DESCRIBE TABLE my_catalog.default.my_table`
    3. `SELECT id FROM my_catalog.default.my_table`
3. Receive error from Spark:
```
io.unitycatalog.client.ApiException: com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of `io.unitycatalog.client.model.ColumnTypeName`, problem: Unexpected value 'GEOMETRY'
 at [Source: (jdk.internal.net.http.ResponseSubscribers$HttpResponseInputStream); line: 1, column: 8354] (through reference chain: io.unitycatalog.client.model.TableInfo["columns"]->java.util.ArrayList[13]->io.unitycatalog.client.model.ColumnInfo["type_name"])
	at io.unitycatalog.client.api.TablesApi.getTableWithHttpInfo(TablesApi.java:273)
	at io.unitycatalog.client.api.TablesApi.getTable(TablesApi.java:241)
	at io.unitycatalog.spark.UCProxy.loadTable(UCSingleCatalog.scala:237)
	at org.apache.spark.sql.connector.catalog.DelegatingCatalogExtension.loadTable(DelegatingCatalogExtension.java:73)
	at org.apache.spark.sql.delta.catalog.DeltaCatalog.super$loadTable(DeltaCatalog.scala:229)
	at org.apache.spark.sql.delta.catalog.DeltaCatalog.$anonfun$loadTable$1(DeltaCatalog.scala:229)
	at org.apache.spark.sql.delta.metering.DeltaLogging.recordFrameProfile(DeltaLogging.scala:171)
	at org.apache.spark.sql.delta.metering.DeltaLogging.recordFrameProfile$(DeltaLogging.scala:169)
	at org.apache.spark.sql.delta.catalog.DeltaCatalog.recordFrameProfile(DeltaCatalog.scala:67)
	at org.apache.spark.sql.delta.catalog.DeltaCatalog.loadTable(DeltaCatalog.scala:228)
	at io.unitycatalog.spark.UCSingleCatalog.loadTable(UCSingleCatalog.scala:73)
	at org.apache.spark.sql.connector.catalog.CatalogV2Util$.getTable(CatalogV2Util.scala:363)
	at org.apache.spark.sql.connector.catalog.CatalogV2Util$.loadTable(CatalogV2Util.scala:337)
	at org.apache.spark.sql.catalyst.analysis.Analyzer$ResolveRelations$.$anonfun$lookupTableOrView$2(Analyzer.scala:1228)
[...]
```

**Expected behavior**

While I understand that the GEOMETRY type is brand new and cannot yet be operated on via `UCSingleCatalog`, I would expect table operations that don't directly operate on the column to continue to work.

james-willis:
**PR Checklist**

**Description of changes**

Fixes #1077.

When Unity Catalog returns a column whose `type_text` is not understood by the version of Spark in use (the original report is `GEOMETRY(4326)` from Databricks UC against Spark 4.0, but the same will happen for any future UC type), `DataType.fromDDL` throws and `UCProxy.loadTable` propagates that exception. The result is that even metadata-only operations like `LIST TABLES`, `DESCRIBE TABLE`, and queries that don't reference the unknown column fail for the entire table.

This change wraps `DataType.fromDDL` in a `try/catch` in `UCProxy.loadTable` and, on `AnalysisException` / `IllegalArgumentException`, logs a warning and falls back to `BinaryType` — the Parquet physical representation for most such logical types — so the table remains usable for metadata operations and for queries that don't reference the unknown column. Operations that do reference the column directly will still fail (with a Spark-side error rather than a deserialization failure).

**Spark 4.1 forward-compatibility:** Spark 4.1 added native `GEOMETRY(srid|ANY)` / `GEOGRAPHY(...)` types. Once the connector is built against Spark 4.1, `DataType.fromDDL("GEOMETRY(4326)")` will parse natively and the fallback won't trigger for that case — the fallback only catches forms the grammar doesn't accept (e.g. `geometry(EPSG:4326)`, bare `GEOMETRY`) or future types.

**Related:** PR #1177 (Nov 2025) already fixed the Jackson side so that an unknown `ColumnTypeName` enum value (`GEOMETRY`) deserializes to `UNKNOWN_DEFAULT_OPEN_API` instead of throwing. That fix is necessary but not sufficient — without this change, `loadTable` still fails on the `fromDDL` step.


james-willis:
On the `BinaryType` choice for the fallback: there's no way to programmatically obtain the column's actual physical type from UC's `ColumnInfo` (`api/all.yaml:1635`) — it exposes `type_text` / `type_json` / `type_name` / precision / scale / interval-type, but no `physical_type` or equivalent. `type_json`'s `nameToType` regex doesn't help on Spark ≤ 4.0 either (no GEOMETRY case there), so it doesn't widen coverage for the case in the issue.

Two ways to get a true physical type, both requiring I/O — but only on the failure path inside the `catch`, not the fast path:

- **Read the Delta log** (`_delta_log/*.json`) for Delta tables — accurate for Delta. In practice though, when Delta itself doesn't recognize a logical type, its stored type tends to be `binary`, so the upside over a hardcoded `BinaryType` fallback is small.
- **Read the Parquet footer** for external Parquet tables — most accurate, only paid when `fromDDL` fails. Doesn't work for empty tables, and again most unknown logical types are `BINARY` underneath.

So the cost is bounded to the rare unknown-type case rather than every `loadTable`. The argument for keeping a hardcoded `BinaryType` fallback is now mostly about complexity and per-format branching (Delta vs Parquet vs others) rather than latency.

james-willis:
@tdas @openinx please take a look! thank you :)

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -643,7 +643,18 @@ private class UCProxy(
...
         partitionCols += col.getName -> index
       }
-      StructField(col.getName, DataType.fromDDL(col.getTypeText), col.getNullable)
+      // UC may return column types newer than this version of Spark understands (e.g. GEOMETRY).
+      // Fall back to BinaryType — the Parquet physical type for most such logical types — so the
+      // table is still usable for metadata ops and queries that don't reference the column.
+      val dataType = try {
+        DataType.fromDDL(col.getTypeText)
+      } catch {
+        case _: org.apache.spark.sql.AnalysisException | _: IllegalArgumentException =>
```
timothyw553, line 652: this seems a bit dangerous, where malformed/corrupt metadata will look like a valid BINARY column. perhaps before this case we should add
```suggestion
        case e: org.apache.spark.sql.AnalysisException
            if col.getTypeName != ColumnTypeName.UNKNOWN_DEFAULT_OPEN_API => throw e
        case e: IllegalArgumentException
            if col.getTypeName != ColumnTypeName.UNKNOWN_DEFAULT_OPEN_API => throw e
        case _: org.apache.spark.sql.AnalysisException | _: IllegalArgumentException =>
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553, line 1: can we add tests for
  + recognized type remains unchanged
  + unknown scalar type follows the intended fallback
  + malformed known type still throws
  + nested unknown type is not flattened to scalar BINARY
  + Spark 4.0 uses the fallback
  + Spark 4.1 parses native GEOMETRY
  + direct reads and writes have explicitly asserted behavior

PR #1544, opened 2026-05-11
New endpoint for Delta: updateTable

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1544/files) to review incremental changes.
- [**stack/drc_update_table**](https://github.com/unitycatalog/unitycatalog/pull/1544) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1544/files)]
  - [stack/drc_update_table_columns](https://github.com/unitycatalog/unitycatalog/pull/1545) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1545/files/b0433125fd1cd3863666f5337912a104e7d59e22..2747d9d5f00e784f060e7a587047c5c46026e389)]
    - [stack/drc_commit](https://github.com/unitycatalog/unitycatalog/pull/1546) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1546/files/2747d9d5f00e784f060e7a587047c5c46026e389..01d122f22f62be2e2f2d04adb67d4d0ae41f3522)]
      - [stack/phaseout](https://github.com/unitycatalog/unitycatalog/pull/1555) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1555/files/01d122f22f62be2e2f2d04adb67d4d0ae41f3522..98bbc249650319957064c75ce6d976e6f82c6210)]

---------
**PR Checklist**

**Description of changes**
New endpoint for Delta: updateTable

Implements POST /delta/v1/catalogs/{c}/schemas/{s}/tables/{t}: an all-or-nothing transaction that applies a list of metadata edits guarded by optional assert-* pre-conditions. Supports set-properties, remove-properties, set-protocol, set-table-comment, set-domain-metadata, remove-domain-metadata, and update-metadata-snapshot-version (EXTERNAL-only); requires assert-table-uuid and accepts an optional assert-etag for optimistic concurrency. All property mutations share one MutablePropertyMap.load / flush so a multi-action request issues a single batched DB write.

Not in this PR (follow-ups): set-columns / set-partition-columns on updateTable, and add-commit, set-latest-backfilled-version, and UniForm-Iceberg sidecar registration on the commit endpoint.


server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -251,21 +254,72 @@ public LoadTableResponse loadTableForDelta(String catalog, String schema, String
...
+   */
+  public LoadTableResponse updateTableForDelta(
+      String catalog, String schema, String table, UpdateTableRequest request) {
+    DeltaUpdateTableMapper.CollectedRequest collected =
+        DeltaUpdateTableMapper.collectRequest(request);
+    String callerId = IdentityUtils.findPrincipalEmailAddress();
+    return TransactionManager.executeWithTransaction(
+        sessionFactory,
+        session -> {
+          TableInfoDAO dao = findTableOrThrow(session, catalog, schema, table);
```
tdas, line 275: [AI Assisted] **Missing Delta-format guard.** `loadTableForDelta` rejects non-Delta entries (line 251-256), but `updateTableForDelta` does not. A Delta REST client that knows the UUID can hit this endpoint against a parquet/csv/metric-view row in the same schema; the property mutation commits, then `buildLoadTableResponse` → `DataSourceFormat.fromValue(...)` throws on the way out — i.e. corruption already happened.

Please add the same guard right after `findTableOrThrow` and pin it with a test.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
```diff
@@ -0,0 +1,338 @@
...
+          "assert-etag failed: expected " + e.getEtag() + " but table has " + currentEtag);
+    }
+  }
+
+  /** Shared by the {@code loadTable} response builder and the {@code assert-etag} check. */
+  public static String computeEtag(TableInfoDAO dao) {
+    return dao.getUpdatedAt() != null
+        ? "etag-" + dao.getUpdatedAt().getTime()
+        : "etag-" + dao.getId();
+  }
```
tdas, line 250: [AI Assisted] **Etag is millisecond-precision and not race-safe.**

1. Two updates in the same millisecond yield the same etag, so a third request pinning that etag passes `assert-etag` even though state changed twice.
2. `updateTableForDelta` runs at default isolation (not `REPEATABLE_READ`) and `TableInfoDAO` has no `@Version`. Two concurrent updaters can both read the same DAO, both pass `assert-etag`, both flush — last writer wins, first update silently lost.

Options: include a monotonic row-version in the etag, or `PESSIMISTIC_WRITE`-lock the DAO right after `findTableOrThrow`. If acceptable for now, please document the weakness explicitly.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
```diff
@@ -0,0 +1,338 @@
...
+ * <p>The mapper does no DB I/O; the caller ({@link io.unitycatalog.server.persist.TableRepository})
+ * loads state once and flushes the diff.
+ */
+public final class DeltaUpdateTableMapper {
+
+  /** Domain metadata name to which derived property key it owns. */
+  private static final Map<String, String> DOMAIN_TO_PROPERTY_KEY =
+      Map.of(
+          DomainMetadataNames.CLUSTERING, TableProperties.CLUSTERING_COLUMNS,
+          DomainMetadataNames.ROW_TRACKING, TableProperties.ROW_TRACKING_ROW_ID_HIGH_WATER_MARK);
```
tdas: [AI Assisted] Set-side projection lives in `DeltaPropertyMapper.deriveFromDomainMetadata`; remove-side projection lives in this local `DOMAIN_TO_PROPERTY_KEY`. Two sources of truth — adding a new domain requires touching both, and drift is silent. Centralize into one mapping that both sides read.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
```diff
@@ -0,0 +1,338 @@
...
+    DeltaPropertyMapper.deriveFromProtocol(derived, protocol);
+    properties.putAll(derived);
+  }
+
+  private static void applySetDomainMetadata(
+      MutablePropertyMap properties, DomainMetadataUpdates updates) {
+    if (updates == null) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT, "set-domain-metadata requires an updates block.");
+    }
```
tdas: [AI Assisted] Only rejects `updates == null`. A request with `SetDomainMetadataUpdate.updates(new DomainMetadataUpdates())` (block present, every subfield null) is accepted and silently no-ops. Either reject empty payloads or pin the no-op in a test.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
```diff
@@ -0,0 +1,338 @@
...
+    properties.removeAll(toRemove);
+  }
+
+  /** Full replacement of the protocol block only; other stored property keys are left alone. */
+  private static void applySetProtocol(MutablePropertyMap properties, DeltaProtocol protocol) {
+    if (protocol == null) {
+      throw new BaseException(ErrorCode.INVALID_ARGUMENT, "set-protocol requires a protocol.");
+    }
+    properties.removeMatchingPrefix(TableProperties.FEATURE_PREFIX);
```
tdas, line 297: [AI Assisted] `set-protocol` on a MANAGED table doesn't re-validate against `UcManagedDeltaContract`. After this call, a MANAGED table can lose `catalogManaged` / required features and end up internally inconsistent. Either re-validate for MANAGED here, or add a comment explaining why divergence is acceptable.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
```diff
@@ -0,0 +1,338 @@
...
+  /** Actions run in canonical order, not request order. */
+  public static void applyUpdates(
+      TableInfoDAO dao, MutablePropertyMap properties, CollectedRequest collected) {
+    CollectedUpdates c = collected.updates();
+    c.setProtocol.ifPresent(u -> applySetProtocol(properties, u.getProtocol()));
+    c.setProperties.ifPresent(u -> applySetProperties(properties, u.getUpdates()));
+    c.removeProperties.ifPresent(u -> applyRemoveProperties(properties, u.getRemovals()));
+    c.setDomainMetadata.ifPresent(u -> applySetDomainMetadata(properties, u.getUpdates()));
+    c.removeDomainMetadata.ifPresent(u -> applyRemoveDomainMetadata(properties, u.getDomains()));
+    c.setTableComment.ifPresent(u -> dao.setComment(u.getComment()));
```
tdas: [AI Assisted] `u.getComment()` may be null and is pushed straight to `dao.setComment(null)`. If null is the intended way to clear, document it; otherwise reject.

server/src/main/java/io/unitycatalog/server/persist/MutablePropertyMap.java
```diff
@@ -0,0 +1,122 @@
...
+   * {@code delta.feature.*} entries in one call.
+   */
+  public void removeMatchingPrefix(String keyPrefix) {
+    current.keySet().removeIf(k -> k.startsWith(keyPrefix));
+  }
+
+  /**
+   * Persist only the keys that changed since {@link #load}: delete keys removed, insert keys added,
+   * update keys whose value differs. No-op for keys whose value is unchanged. This is where every
+   * property mutation in the transaction becomes a single batch of DB writes.
```
tdas: [AI Assisted] Doc says "single batch of DB writes", but `flush` issues per-key `session.persist`/`merge`/`remove`. Whether these are actually batched depends on `hibernate.jdbc.batch_size`. Suggest softening to "a single end-of-transaction flush", or confirm batching is configured.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
```diff
@@ -0,0 +1,338 @@
...
+  private static <T extends TableUpdate> Optional<T> fillOnce(Optional<T> slot, T value) {
+    if (slot.isPresent()) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT,
+          "At most one " + value.getAction() + " is allowed per request.");
+    }
+    return Optional.of(value);
+  }
+
+  private static <T extends TableRequirement> Optional<T> fillOnce(Optional<T> slot, T value) {
```
tdas: [AI Assisted] The two `fillOnce` overloads are identical except for the `getAction()` vs `getType()` call. Collapse into one generic helper that takes a `Function<T, String> nameFn`.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
```diff
@@ -0,0 +1,338 @@
...
+      properties.remove(propertyKey);
+    }
+  }
+
+  private static void applyUpdateSnapshotVersion(
+      TableInfoDAO dao, MutablePropertyMap properties, UpdateSnapshotVersionUpdate update) {
+    if (!TableType.EXTERNAL.toString().equals(dao.getType())) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT,
+          "update-metadata-snapshot-version is only supported for EXTERNAL tables.");
```
tdas: [AI Assisted] Error message doesn't tell a MANAGED-table user what to do instead. Suggest: `"update-metadata-snapshot-version is only supported for EXTERNAL Delta tables; for MANAGED tables, use the Delta commit endpoint."`

tdas (commented):
[AI Assisted] Comments below are from an AI-assisted review pass. Flagging two correctness concerns and a few suggestions; please use your judgement.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: Added a gate `requireDeltaTable` for  `updateTableForDelta`. Also a test `non-Delta table is rejected before any mutation commits`

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
yili-db: Documented the weakness of ms in comments.
Added the `session.refresh(dao, PESSIMISTIC_WRITE)` to prevent concurrent updates.
Eventually when the followup https://github.com/unitycatalog/unitycatalog/pull/1546 is landed, the commit version becomes the authoritative mechanism.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
yili-db: Fixed.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
yili-db: Fixed.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
yili-db: Fixed.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
yili-db: Fixed.

server/src/main/java/io/unitycatalog/server/persist/MutablePropertyMap.java
yili-db: Removed this comment.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
yili-db: Consolidated into one.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
yili-db: Added.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
```diff
@@ -0,0 +1,349 @@
...
+   * re-validate against the UC catalog-managed contract so required features can't be stripped.
+   */
+  private static void applySetProtocol(
+      TableInfoDAO dao, MutablePropertyMap properties, DeltaProtocol protocol) {
+    ValidationUtils.checkNotNull(protocol, "set-protocol requires a protocol.");
+    properties.removeMatchingPrefix(TableProperties.FEATURE_PREFIX);
+    Map<String, String> derived = new HashMap<>();
+    DeltaPropertyMapper.deriveFromProtocol(derived, protocol);
+    properties.putAll(derived);
+    if (TableType.MANAGED.toString().equals(dao.getType())) {
```
tdas: [AI Assisted] **Residual consistency gap on `set-protocol`** (non-blocking, possibly fine as a follow-up).

`UcManagedDeltaContract.validate(protocol, /* domainMetadata= */ null, properties.asMap())` skips `validateDomainMetadataAgainstProtocol` (it returns early on null). But the row may still carry domain-metadata-derived property keys from a previous `set-domain-metadata` (e.g. `delta.rowTracking.rowIdHighWaterMark`). If the new protocol drops the `rowTracking` writer feature, those projected property keys survive and the table ends up referencing a feature the protocol no longer declares.

Options: (a) synthesize the effective `DomainMetadataUpdates` from the current properties before validating, (b) sweep DM-derived property keys whose backing feature is dropped, or (c) document that this gap is acceptable until a follow-up enforces it.

tdas (commented):
[AI Assisted] Second-pass review. All nine prior comments are addressed; verified the fixes against the new commit. One residual subtle concern below, non-blocking. Overall the design looks solid — pessimistic lock + documented etag weakness + MANAGED contract re-validation is the right shape. Still needs human / domain-expert sign-off (CCv2 / catalog-managed semantics).

tdas (approved):
LGTM.. with the need for a follow up. 

But pelase make sure tests pass.

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
yili-db: Added validation for residual DM after removal of features. This validation is now done after both protocol and DM are updated, at the end of `applyUpdates`

PR #1545, opened 2026-05-11
New endpoint for Delta: updateTable supports set-columns / set-partition-columns

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1545/files) to review incremental changes.
- [**stack/drc_update_table_columns**](https://github.com/unitycatalog/unitycatalog/pull/1545) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1545/files)]
  - [stack/drc_commit](https://github.com/unitycatalog/unitycatalog/pull/1546) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1546/files/54dea8f73ea105edb5afffb93d9e786163e5752f..5425098d6a18abac44fa16a58f3acc148e17f0c2)]
    - [stack/phaseout](https://github.com/unitycatalog/unitycatalog/pull/1555) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1555/files/5425098d6a18abac44fa16a58f3acc148e17f0c2..063190cae2c6ff52e04de2faaad3c4f459f954f6)]

---------
**PR Checklist**

**Description of changes**
New endpoint for Delta: updateTable supports set-columns / set-partition-columns

Adds schema and partition-column mutation actions to the Delta updateTable endpoint. set-columns replaces the column definitions; set-partition-columns replaces the partition list. The two are independent per spec, so a column-only request preserves the existing partition columns by name (rejecting if a previously-partition column is missing from the new schema), and a partition-only request leaves the schema untouched. Both actions share one combined pass that builds the post-update column list and swaps the DAO column collection once.

Not in this PR (follow-up): add-commit / set-latest-backfilled-version on the commit endpoint.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
@@ -351,6 +351,10 @@ public final class DeltaUpdateTableMapper {
       List<StructField> fields =
           ValidationUtils.checkNotNull(
               columns.getFields(), "set-columns requires columns.fields.");
+      if (fields.isEmpty()) {
+        throw new BaseException(
+            ErrorCode.INVALID_ARGUMENT, "set-columns requires at least one column.");
+      }
       newColumns = ColumnUtils.toColumnInfos(fields);
     } else {
       newColumns = ColumnInfoDAO.toList(dao.getColumns());
@@ -358,10 +362,15 @@ public final class DeltaUpdateTableMapper {
     }
     // Source of the partition list: the request when set-partition-columns is present, otherwise
     // the existing DAO's partition columns (preserved by name across a column-only action).
-    List<String> partitionNames =
-        setPartition
-            .map(SetPartitionColumnsUpdate::getPartitionColumns)
-            .orElseGet(() -> currentPartitionColumnNames(dao));
+    List<String> partitionNames;
+    if (setPartition.isPresent()) {
+      partitionNames =
+          ValidationUtils.checkNotNull(
+              setPartition.get().getPartitionColumns(),
+              "set-partition-columns requires a partition-columns list.");
+    } else {
+      partitionNames = currentPartitionColumnNames(dao);
+    }
     ColumnUtils.applyPartitionColumns(newColumns, partitionNames);
     List<ColumnInfoDAO> newColumnDAOs = ColumnInfoDAO.fromList(newColumns);
     newColumnDAOs.forEach(
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
@@ -7,21 +7,29 @@ import io.unitycatalog.server.delta.model.DomainMetadataUpdates;
 import io.unitycatalog.server.delta.model.RemoveDomainMetadataUpdate;
 import io.unitycatalog.server.delta.model.RemovePropertiesUpdate;
 import io.unitycatalog.server.delta.model.SetDomainMetadataUpdate;
+import io.unitycatalog.server.delta.model.SetPartitionColumnsUpdate;
 import io.unitycatalog.server.delta.model.SetPropertiesUpdate;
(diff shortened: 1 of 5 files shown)


server/src/main/java/io/unitycatalog/server/utils/ColumnUtils.java
```diff
@@ -268,6 +269,19 @@ public static ColumnInfo toColumnInfo(StructField field, int position) {
     return column;
   }
 
+  /**
+   * Project a list of Delta {@link StructField}s into UC {@link ColumnInfo}s, stamping each
+   * column's {@code position} from its index in the list. Used by both the create and update paths
+   * of the Delta REST Catalog so the wire-order-to-position mapping stays in one place.
```
tdas: lets stop using Delta REST catalog. the entire UC API is REST.. this does not make sense. lets just start calling it the UC Delta APIs

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
```diff
@@ -0,0 +1,421 @@
...
+    } else {
+      newColumns = ColumnInfoDAO.toList(dao.getColumns());
+      newColumns.forEach(c -> c.setPartitionIndex(null));
+    }
+    // Source of the partition list: the request when set-partition-columns is present, otherwise
+    // the existing DAO's partition columns (preserved by name across a column-only action).
+    List<String> partitionNames =
+        setPartition
+            .map(SetPartitionColumnsUpdate::getPartitionColumns)
+            .orElseGet(() -> currentPartitionColumnNames(dao));
```
tdas: [AI Assisted] Null `partition-columns` field is silently treated as "preserve existing partitions," which is asymmetric with the `SetSchemaUpdate.getColumns()` null check above. The chain `setPartition.map(SetPartitionColumnsUpdate::getPartitionColumns).orElseGet(...)` collapses to `Optional.empty()` when the getter returns null, so a client sending `{"action":"set-partition-columns"}` (no `partition-columns` field — `required` per `api/delta.yaml`, but generated Java models don't enforce it on the wire) silently no-ops on the partition side instead of erroring out.

Suggest the same `ValidationUtils.checkNotNull` treatment for parity:

```java
List<String> partitionNames;
if (setPartition.isPresent()) {
  partitionNames = ValidationUtils.checkNotNull(
      setPartition.get().getPartitionColumns(),
      "set-partition-columns requires a partition-columns list.");
} else {
  partitionNames = currentPartitionColumnNames(dao);
}
```

server/src/main/java/io/unitycatalog/server/service/delta/DeltaUpdateTableMapper.java
```diff
@@ -0,0 +1,421 @@
...
+    // Source of the new schema: the request when set-columns is present, otherwise the existing
+    // DAO with partition indices cleared so applyPartitionColumns can re-stamp them below.
+    List<ColumnInfo> newColumns;
+    if (setSchema.isPresent()) {
+      StructType columns =
+          ValidationUtils.checkNotNull(
+              setSchema.get().getColumns(), "set-columns requires a columns block.");
+      List<StructField> fields =
+          ValidationUtils.checkNotNull(
+              columns.getFields(), "set-columns requires columns.fields.");
```
tdas, line 353: [AI Assisted] `ValidationUtils.checkNotNull` guards against a missing `fields` array but not against `fields: []`. With an empty list, `ColumnUtils.toColumnInfos([])` returns an empty list; the partition list carried over from the existing DAO will usually then error with `partition-columns references unknown column`, but a non-partitioned table would happily end up with zero columns.

Suggest rejecting an empty `fields` block at the same checkpoint:

```java
if (fields.isEmpty()) {
  throw new BaseException(
      ErrorCode.INVALID_ARGUMENT, "set-columns requires at least one column.");
}
```

tdas (approved):

tdas:
tests are failing

yili-db:
> tests are failing

@tdas This is an unrelated non-blocking test failing. I have https://github.com/unitycatalog/unitycatalog/pull/1560 to fix it.

PR #1546, opened 2026-05-11
New endpoint for Delta: updateTable supports add-commit / set-latest-backfilled-version

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1546/files) to review incremental changes.
- [**stack/drc_commit**](https://github.com/unitycatalog/unitycatalog/pull/1546) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1546/files)]
  - [stack/phaseout](https://github.com/unitycatalog/unitycatalog/pull/1555) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1555/files/9448fa1ac2a338f0518637c624442f144a5b6eb2..681072621f178318d82dad0db78a4b162ced8f9f)]

---------
**PR Checklist**

**Description of changes**
New endpoint for Delta: updateTable supports add-commit / set-latest-backfilled-version

Adds the CCv2 commit-log actions to the Delta updateTable endpoint. add-commit appends a new commit (with optional uniform metadata for UniForm tables); set-latest-backfilled-version records that a prior commit has been backfilled. Both are restricted to MANAGED Delta tables and go through the same onboarding / normal-commit / backfill-only dispatch as the UC REST commit endpoint, so commit-log progression is identical between the two surfaces. When add-commit is bundled with metadata-changing sibling actions, delta.lastUpdateVersion and delta.lastCommitTimestamp are stamped automatically from the commit's version / timestamp.

A new gRPC code COMMIT_VERSION_CONFLICT (HTTP 409, ErrorType COMMIT_VERSION_CONFLICT_EXCEPTION) replaces ALREADY_EXISTS for the commit-version replay case so Delta clients get the spec-correct error type.

With this PR all 11 actions defined in the Delta REST spec are now wired through the updateTable endpoint.


server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -253,39 +253,59 @@ public void postCommit(DeltaCommit commit) {
...
+      DeltaCommitDAO lastCommitDAO = firstAndLastCommits.get(1);
+      ValidationUtils.checkArgument(
+          firstCommitDAO.getCommitVersion() <= lastCommitDAO.getCommitVersion(),
+          "Inconsistent commit log: first commit version > last commit version.");
+      if (commit.getCommitInfo() == null) {
+        // latestBackfilledVersion non-null guaranteed upstream (UC REST: validateCommit;
+        // Delta update: DeltaUpdateTableMapper checkNotNull).
+        ValidationUtils.checkNotNull(
+            commit.getLatestBackfilledVersion(),
+            "latest-backfilled-version is required when commit-info is absent.");
```
tdas: [AI Assisted] Non-blocking, but worth a follow-up: `ValidationUtils.checkArgument` and `checkNotNull` throw `ErrorCode.INVALID_ARGUMENT` (HTTP 400). Both of these checks are internal invariants -- the comment on the `checkNotNull` even says "guaranteed upstream", and the first <= last commit ordering is a server-side consistency property the user can't influence. If either ever fires, the caller will get a 400 with internal-corruption text when the right answer is 500 / `INTERNAL`.

The previous form (`assert`) was correct in spirit for invariants -- fail-fast under `-ea`, no-op in prod -- just hard to surface in tests. Two reasonable options:
1. Restore `assert` here and reserve `ValidationUtils.*` for user-supplied data.
2. Add a `ValidationUtils.checkState(boolean, String)` (or similar) that throws `ErrorCode.INTERNAL`, and use it for these two sites.

OK to handle in a follow-up.

tdas (approved):
[AI Assisted] LGTM — nice follow-through on the prior round (UniForm consistency tests, stamping via non-property changes, EXTERNAL × commit gate, backfill-past-last-commit, the Delta-API authz pin, and the `requireConvertedDeltaVersionEquals` factor-out). One non-blocking concern inline; happy to land as a follow-up.

Two other items from the prior round that didn't need code changes but deserve a one-line answer in the PR description, not blocking:
- Was the `postCommit` error-type change from `ALREADY_EXISTS_EXCEPTION` to `COMMIT_VERSION_CONFLICT_EXCEPTION` checked against existing clients (delta-spark / kernel)? It's a wire-format change for an already-released endpoint, so worth a sentence in the description confirming no client branches on the old type, or gating it behind `server.managed-table.use-delta-api-only` from #1555.
- The fully-qualified `io.unitycatalog.server.delta.model.DeltaCommit` co-exists with the file's `io.unitycatalog.server.model.DeltaCommit` import. Pushing `toUcCommitInfo` into a small adapter in `service/delta/` would let the repo speak only the UC wire type. Follow-up.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: Revert back to `assert`

yili-db:
> [AI Assisted] LGTM — nice follow-through on the prior round (UniForm consistency tests, stamping via non-property changes, EXTERNAL × commit gate, backfill-past-last-commit, the Delta-API authz pin, and the `requireConvertedDeltaVersionEquals` factor-out). One non-blocking concern inline; happy to land as a follow-up.
> 
> Two other items from the prior round that didn't need code changes but deserve a one-line answer in the PR description, not blocking:
> 
> * Was the `postCommit` error-type change from `ALREADY_EXISTS_EXCEPTION` to `COMMIT_VERSION_CONFLICT_EXCEPTION` checked against existing clients (delta-spark / kernel)? It's a wire-format change for an already-released endpoint, so worth a sentence in the description confirming no client branches on the old type, or gating it behind `server.managed-table.use-delta-api-only` from #1555.
> * The fully-qualified `io.unitycatalog.server.delta.model.DeltaCommit` co-exists with the file's `io.unitycatalog.server.model.DeltaCommit` import. Pushing `toUcCommitInfo` into a small adapter in `service/delta/` would let the repo speak only the UC wire type. Follow-up.

`ALREADY_EXISTS_EXCEPTION` to `COMMIT_VERSION_CONFLICT_EXCEPTION`:
This is a better except to surface for existing clients. The behavior was never part of standard or spec. It was used only because it returns the correct http error code.

The fully-qualified `io.unitycatalog.server.delta.model.DeltaCommit`:
Moving it and the conversion to `DeltaUpdateTableMapper`

PR #1547, opened 2026-05-11
Add DeltaApiException for typed Delta error envelope access

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1547/files) to review incremental changes.
- [**stack/client_error**](https://github.com/unitycatalog/unitycatalog/pull/1547) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1547/files)]

---------
**PR Checklist**

**Description of changes**
Add DeltaApiException for typed Delta error envelope access

A specialization of ApiException that parses the Delta error envelope ({error: {code, type, message, stack}}) and exposes typed accessors (getErrorCode, getErrorType, getErrorMessage). Construct via DeltaApiException.from(ApiException) when a Delta error body is expected; falls back to Optional.empty() if the body isn't parseable. TestUtils.assertDeltaApiException now goes through it, replacing the inline ObjectMapper.readValue. Sets up the path to a future ApiClient-level auto-wrap without touching call sites.

### DIFF
--- a/clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
+++ b/clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
@@ -1,0 +1,100 @@
+package io.unitycatalog.client.delta;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.unitycatalog.client.ApiException;
+import io.unitycatalog.client.delta.model.ErrorModel;
+import io.unitycatalog.client.delta.model.ErrorResponse;
+import io.unitycatalog.client.delta.model.ErrorType;
+import java.util.Optional;
+
+/**
+ * Delta-specific specialization of {@link ApiException}. Parses the spec-defined error envelope
+ * ({@code {"error": {"code", "type", "message", "stack"}}}) once at construction and exposes typed
+ * accessors so call sites don't repeat the parse-from-responseBody dance.
+ *
+ * <p>Designed to wrap an existing {@link ApiException}: callers that catch the generated base type
+ * can upgrade it with {@link #from(ApiException)} when they want the typed view. The wrapping is
+ * caller-driven for now; the structure is set up so a future {@code ApiClient} interceptor can do
+ * the wrap automatically without touching call sites that already use this class. Until then, both
+ * shapes co-exist:
+ *
+ * <pre>{@code
+ * try {
+ *   deltaTablesApi.updateTable(...);
+ * } catch (ApiException ex) {
+ *   DeltaApiException.from(ex).ifPresent(d -> {
+ *     ErrorType type = d.getErrorType();      // spec-defined enum
+ *     String message = d.getErrorMessage();   // spec-defined message
+ *     // ...
+ *   });
+ *   throw ex;
+ * }
+ * }</pre>
+ */
+public class DeltaApiException extends ApiException {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  /** {@code null} if the response body wasn't parseable as a Delta error envelope. */
+  private final ErrorModel error;
+
+  /**
+   * Wrap an existing {@link ApiException}, copying its HTTP code, headers, body, cause, and message
+   * through, and parsing the response body into the Delta error envelope. If the body isn't
+   * parseable, {@link #getError()} (and the typed accessors derived from it) return null; callers
+   * wanting a hard parse failure should use {@link #from(ApiException)} instead and react to an
+   * empty {@link Optional}.
+   */
+  public DeltaApiException(ApiException source) {
+    super(
+        source.getMessage(),
+        source.getCause(),
+        source.getCode(),
+        source.getResponseHeaders(),
+        source.getResponseBody());
+    this.error = parse(source.getResponseBody());
+  }
+
+  /**
+   * Lenient upgrade: returns a {@link DeltaApiException} only when the response body parses cleanly
+   * as a Delta error envelope. {@link Optional#empty()} typically means the response wasn't from
+   * the Delta surface, or the server emitted a non-spec body -- in either case the caller can fall
+   * back to the generic {@link ApiException} it was given.
+   */
+  public static Optional<DeltaApiException> from(ApiException ex) {
+    DeltaApiException upgraded = new DeltaApiException(ex);
+    return upgraded.error == null ? Optional.empty() : Optional.of(upgraded);
+  }
+
+  /** The full parsed error envelope. {@code null} if the body wasn't a Delta error response. */
+  public ErrorModel getError() {
+    return error;
+  }
+
+  /** {@code error.code}: spec-defined HTTP status code (typically matches {@link #getCode()}). */
+  public Integer getErrorCode() {
+    return error == null ? null : error.getCode();
+  }
+
+  /** {@code error.type}: spec-defined error-type enum, e.g. {@code TABLE_NOT_FOUND_EXCEPTION}. */
+  public ErrorType getErrorType() {
+    return error == null ? null : error.getType();
+  }
+
+  /** {@code error.message}: spec-defined human-readable message. */
+  public String getErrorMessage() {
+    return error == null ? null : error.getMessage();
+  }
+
+  private static ErrorModel parse(String body) {
+    if (body == null || body.isEmpty()) {
+      return null;
+    }
+    try {
+      ErrorResponse response = MAPPER.readValue(body, ErrorResponse.class);
+      return response == null ? null : response.getError();
+    } catch (Exception e) {
+      return null;
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
```diff
@@ -0,0 +1,100 @@
...
+ *     String message = d.getErrorMessage();   // spec-defined message
+ *     // ...
+ *   });
+ *   throw ex;
+ * }
+ * }</pre>
+ */
+public class DeltaApiException extends ApiException {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
```
timothyw553: ObjectMapper has `FAIL_ON_UNKNOWN_PROPERTIES=true` by default. is this intended or should we match ApiClient's objectmapper which has it as false?

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
```diff
@@ -0,0 +1,100 @@
...
+    return error == null ? null : error.getMessage();
+  }
+
+  private static ErrorModel parse(String body) {
+    if (body == null || body.isEmpty()) {
+      return null;
+    }
+    try {
+      ErrorResponse response = MAPPER.readValue(body, ErrorResponse.class);
+      return response == null ? null : response.getError();
```
timothyw553, line 103: what happens when `response.getError().[getCode()|getType()|getMessage()] == null`? then do we want to return null too? i didnt fully understand this case in the spec https://github.com/unitycatalog/unitycatalog/blob/main/api/delta-docs/Models/ErrorModel.md

clients/java/src/test/java/io/unitycatalog/client/delta/DeltaApiExceptionTest.java
```diff
@@ -0,0 +1,74 @@
...
+  public void fromReturnsEmptyOnMalformedJson() {
+    ApiException source = new ApiException(400, "msg", null, "this is not json");
+    assertThat(DeltaApiException.from(source)).isEmpty();
+  }
+
+  @Test
+  public void fromReturnsEmptyWhenErrorEnvelopeAbsent() {
+    // Well-formed JSON, but no "error" key -- the body isn't a Delta error response.
+    ApiException source = new ApiException(400, "msg", null, "{\"unrelated\":\"x\"}");
+    assertThat(DeltaApiException.from(source)).isEmpty();
```
timothyw553, line 49: kinda related to my 2nd comment if my assumption is correct - otherwise we can ignore. should we have a test for when we get `"{\"error\":null}"`? in that case the assertion is the same that we expect empty, right?

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
yili-db: If any of them is null, or even if response.getError() itself is null, this function returns null as the best effort.
Normally it shouldn't happen.

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
```diff
@@ -0,0 +1,108 @@
...
+  private final ErrorModel error;
+
+  /**
+   * Wrap an existing {@link ApiException}, copying its HTTP code, headers, body, cause, and message
+   * through, and parsing the response body into the Delta error envelope. If the body isn't
+   * parseable, {@link #getError()} (and the typed accessors derived from it) return null; callers
+   * wanting a hard parse failure should use {@link #from(ApiException)} instead and react to an
+   * empty {@link Optional}.
+   */
+  public DeltaApiException(ApiException source) {
```
tdas: should we hide the constructor and only provide a static method to create it? more flexible in the future maybe.

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
```diff
@@ -0,0 +1,108 @@
...
+ *   });
+ *   throw ex;
+ * }
+ * }</pre>
+ */
+public class DeltaApiException extends ApiException {
+
+  private static final ObjectMapper MAPPER = createObjectMapper();
+
+  protected static ObjectMapper createObjectMapper() {
```
tdas: does it need to be protected?

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
```diff
@@ -0,0 +1,108 @@
...
+  public ErrorModel getError() {
+    return error;
+  }
+
+  /** {@code error.code}: spec-defined HTTP status code (typically matches {@link #getCode()}). */
+  public Integer getErrorCode() {
+    return error == null ? null : error.getCode();
+  }
+
+  /** {@code error.type}: spec-defined error-type enum, e.g. {@code TABLE_NOT_FOUND_EXCEPTION}. */
```
tdas: [AI Assisted] nit: `TABLE_NOT_FOUND_EXCEPTION` isn't an actual enum value. Per `api/delta.yaml` the wire value is `NoSuchTableException` and the generated Java constant is `ErrorType.NO_SUCH_TABLE_EXCEPTION` (as used in the tests in this PR). Suggest changing the example to `NO_SUCH_TABLE_EXCEPTION` so anyone copy-pasting gets a value that actually exists.

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
```diff
@@ -0,0 +1,108 @@
...
+   * empty {@link Optional}.
+   */
+  public DeltaApiException(ApiException source) {
+    super(
+        source.getMessage(),
+        source.getCause(),
+        source.getCode(),
+        source.getResponseHeaders(),
+        source.getResponseBody());
+    this.error = parse(source.getResponseBody());
```
tdas, line 64: [AI Assisted] The constructor passes `source.getCause()` (rather than `source`) as the cause to `super(...)`, which drops the original `ApiException`'s stack trace and its cause chain. Today that's mostly fine because `from(ex)` callers still hold `ex` in scope. But the class Javadoc explicitly markets a future ApiClient-level auto-wrap path where `DeltaApiException` would be the only exception users see — in that world the network-layer stack trace is lost.

Consider chaining `source` itself:

```java
super(
    source.getMessage(),
    source,  // preserve original ApiException stack/cause
    source.getCode(),
    source.getResponseHeaders(),
    source.getResponseBody());
```

If the current behavior is intentional, a one-line comment explaining why would help the next reader. Either way, a small test that pins the cause/passthrough semantics would lock the contract.

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private static ErrorModel parse(String body) {
+    if (body == null || body.isEmpty()) {
+      return null;
+    }
+    try {
+      ErrorResponse response = MAPPER.readValue(body, ErrorResponse.class);
+      return response == null ? null : response.getError();
+    } catch (Exception e) {
```
tdas: [AI Assisted] nit: `catch (Exception e)` is wider than needed here. Jackson throws `IOException` / `JsonProcessingException`; catching `Exception` will also swallow NPEs / other `RuntimeException`s that would otherwise indicate a real bug in this code (or in the generated model). Suggest narrowing to `IOException` (or `JacksonException`).

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
```diff
@@ -0,0 +1,108 @@
...
+        source.getResponseBody());
+    this.error = parse(source.getResponseBody());
+  }
+
+  /**
+   * Lenient upgrade: returns a {@link DeltaApiException} only when the response body parses cleanly
+   * as a Delta error envelope. {@link Optional#empty()} typically means the response wasn't from
+   * the Delta surface, or the server emitted a non-spec body -- in either case the caller can fall
+   * back to the generic {@link ApiException} it was given.
+   */
```
tdas, line 71: [AI Assisted] The tests in this PR now lock four distinct empty-return cases (null body, malformed JSON, missing `error` key, `"error": null`). Worth enumerating them here so the contract is explicit:

> Returns empty when the body is `null`/empty, isn't valid JSON, is missing the `error` envelope, or has `"error": null`.

This makes the Javadoc match what the tests actually pin.

tdas (approved):
[AI Assisted] LGTM with a few non-blocking suggestions inline. Nice job pinning the `FAIL_ON_UNKNOWN_PROPERTIES=false` and `{"error":null}` / `{"error":{}}` edge cases in tests.

clients/java/src/main/java/io/unitycatalog/client/delta/DeltaApiException.java
yili-db: Changed to JacksonException

PR #1548, opened 2026-05-12
[Hadoop] Revisit Hadoop credential API for Delta tables

timothyw553:
**PR Checklist**

**Description of changes**

This updates the Hadoop credential configuration API for UC Delta table support, adds a public Delta table builder overload using catalog/schema/table/operation/location, keeps `UCDeltaTableIdentifier` internal, factors Delta table credential property helpers into `CredPropsUtil`, and updates credential-scoped filesystem cache keys so Delta table scopes include the table location instead of the random credential UID. The location in the Delta table scoped key fixes stale credential-cache reuse after drop and recreate while avoiding excess filesystem instances from per-call random UIDs.

Validation:

- `./build/sbt -mem 4096 hadoop/test` passed: 132 tests, 0 failures.

No related issue was identified in the source branch.

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
@@ -1,6 +1,7 @@
 package io.unitycatalog.hadoop;
 
 import io.unitycatalog.client.auth.TokenProvider;
+import io.unitycatalog.client.delta.model.CredentialOperation;
 import io.unitycatalog.client.delta.model.StorageCredential;
 import io.unitycatalog.client.internal.Preconditions;
 import io.unitycatalog.client.model.PathOperation;
@@ -8,6 +9,7 @@ import io.unitycatalog.client.model.TableOperation;
 import io.unitycatalog.client.model.TemporaryCredentials;
 import io.unitycatalog.hadoop.internal.CredPropsUtil;
 import io.unitycatalog.hadoop.internal.DeltaStorageCredentialUtil;
+import io.unitycatalog.hadoop.internal.UCDeltaTableIdentifier;
 import io.unitycatalog.hadoop.internal.UCHadoopConfConstants;
 import java.util.Collections;
 import java.util.HashMap;
@@ -63,10 +65,10 @@ public final class UCCredentialHadoopConfs {
     private Configuration hadoopConf = new Configuration(false);
     private final Map<String, String> engineVersionProps = new LinkedHashMap<>();
 
-    private Builder(String uri, String scheme) {
-      Preconditions.checkArgument(uri != null, "catalogUri is required");
+    private Builder(String catalogUri, String scheme) {
+      Preconditions.checkArgument(catalogUri != null, "catalogUri is required");
       Preconditions.checkArgument(scheme != null, "scheme is required");
-      this.catalogUri = uri;
+      this.catalogUri = catalogUri;
       this.scheme = scheme;
     }
 
@@ -183,47 +185,39 @@ public final class UCCredentialHadoopConfs {
     /**
      * Builds Hadoop properties for a UC Delta table's storage location.
      *
-     * @param tableId the UC table UUID, used to scope the credential cache and filesystem cache
      * @param catalog the UC catalog name
      * @param schema the UC schema name
      * @param table the UC table name
+     * @param operation the credential operation to request from the UC Delta credentials API
      * @param location the table storage location used to select a returned storage credential
      * @return unmodifiable map; empty if the scheme is unrecognized
-     * @throws IllegalArgumentException if a table identity field is missing
+     * @throws IllegalArgumentException if a table identity, operation, or location field is missing
      * @throws IllegalStateException if a required builder field is missing
      */
     public Map<String, String> buildForTable(
-        String tableId, String catalog, String schema, String table, String location) {
-      Preconditions.checkArgument(tableId != null && !tableId.isEmpty(), "tableId is required");
-      Preconditions.checkArgument(catalog != null && !catalog.isEmpty(), "catalog is required");
-      Preconditions.checkArgument(schema != null && !schema.isEmpty(), "schema is required");
-      Preconditions.checkArgument(table != null && !table.isEmpty(), "table is required");
+        String catalog,
+        String schema,
+        String table,
+        CredentialOperation operation,
+        String location) {
+      UCDeltaTableIdentifier identifier = UCDeltaTableIdentifier.of(catalog, schema, table);
+      Preconditions.checkArgument(operation != null, "operation is required");
       Preconditions.checkArgument(location != null && !location.isEmpty(), "location is required");
       validateUcDeltaApi();
       TemporaryCredentials tempCreds =
           DeltaStorageCredentialUtil.toTemporaryCredentials(initialStorageCredential);
-      Map<String, String> props =
-          CredPropsUtil.createTableCredProps(
+      return withEngineVersionProps(
+          CredPropsUtil.createDeltaTableCredProps(
               credentialRenewalEnabled,
               credentialScopedFsEnabled,
               hadoopConf,
               scheme,
               catalogUri,
               tokenProvider,
-              tableId,
-              DeltaStorageCredentialUtil.toTableOperation(initialStorageCredential.getOperation()),
-              tempCreds);
-      if (props.isEmpty()) {
-        return props;
-      }
-      Map<String, String> deltaProps = new HashMap<>(props);
-      deltaProps.put(
-          UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_KEY, Boolean.TRUE.toString());
-      deltaProps.put(UCHadoopConfConstants.UC_DELTA_CATALOG_KEY, catalog);
-      deltaProps.put(UCHadoopConfConstants.UC_DELTA_SCHEMA_KEY, schema);
-      deltaProps.put(UCHadoopConfConstants.UC_DELTA_TABLE_NAME_KEY, table);
-      deltaProps.put(UCHadoopConfConstants.UC_DELTA_LOCATION_KEY, location);
-      return withEngineVersionProps(Collections.unmodifiableMap(deltaProps));
+              identifier,
+              location,
+              operation,
+              tempCreds));
     }
 
     /**
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -1,6 +1,7 @@
 package io.unitycatalog.hadoop.internal;
 
 import io.unitycatalog.client.auth.TokenProvider;
+import io.unitycatalog.client.delta.model.CredentialOperation;
 import io.unitycatalog.client.internal.Preconditions;
 import io.unitycatalog.client.model.AwsCredentials;
 import io.unitycatalog.client.model.AzureUserDelegationSAS;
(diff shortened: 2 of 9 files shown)


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCDeltaTableIdentifier.java
```diff
@@ -0,0 +1,58 @@
+package io.unitycatalog.hadoop;
+
+import com.google.common.base.Objects;
+import com.google.common.base.Preconditions;
```
openinx: okay, we cannot use the google guava here. the `Preconditions` should come from `io.unitycatalog.client.internal.Preconditions`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCDeltaTableIdentifier.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class UCDeltaTableIdentifier {
```
openinx: As this is a public API, we will need to add the simply description for the javadoc.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedKey.java
```diff
@@ -25,14 +28,24 @@ public interface CredScopedKey {
...
+      if (conf.getBoolean(
+          UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_KEY,
+          UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_DEFAULT_VALUE)) {
+        return new DeltaTableCredScopedKey(
+            UCDeltaTableIdentifier.of(
+                conf.get(UCHadoopConfConstants.UC_DELTA_CATALOG_KEY),
+                conf.get(UCHadoopConfConstants.UC_DELTA_SCHEMA_KEY),
+                conf.get(UCHadoopConfConstants.UC_DELTA_TABLE_NAME_KEY)),
+            conf.get(UCHadoopConfConstants.UC_TABLE_OPERATION_KEY),
+            conf.get(UCHadoopConfConstants.UC_CREDENTIALS_UID_KEY));
```
openinx: As we use `operation`, `catalog`, `schema`, `tableName` to request the credentials, from the hadoop conf:
* UC_TABLE_OPERATION_KEY
* UC_DELTA_CATALOG_KEY
* UC_DELTA_SCHEMA_KEY
* UC_DELTA_TABLE_NAME_KEY

And we also use the `location` to differenciate the credentials.  I I think we also need to add the `location` in the `DeltaTableCredScopedKey`,  so that we can distinguish the credential keys from the case drop + re-create table with different locations.

openinx: Pls also help to add the unit test, this is important.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedKey.java
```diff
@@ -95,6 +108,45 @@ public String toString() {
     }
   }
 
+  class DeltaTableCredScopedKey implements CredScopedKey {
+    private final UCDeltaTableIdentifier identifier;
+    private final String tableOperation;
+    private final String uid;
```
openinx: `uid` is not good enough for adding it as key, since every `CredPropsUtil#createDeltaTableCredProps` will create a separate random uuid, with this context, it may spawn too many S3AFileSystem inside the CredScopedFileSystem cache.  So I suggest to remove this .

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -419,6 +432,67 @@ private static Map<String, String> abfsPathTempCredProps(
...
+  public static Map<String, String> createDeltaTableCredProps(
+      boolean renewCredEnabled,
+      boolean credScopedFsEnabled,
+      Configuration hadoopConf,
+      String scheme,
+      String uri,
+      TokenProvider tokenProvider,
+      UCDeltaTableIdentifier identifier,
+      String location,
+      TableOperation tableOp,
```
openinx: Here we shouldn't use the old client's `TableOperation` , we should use the new client's `CredentialOperation` directly.

openinx: This may require us to add a new `credentialOperation` inside the `PropsBuilder` in `CredPropsUtil`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
```diff
@@ -183,47 +183,31 @@ public Map<String, String> buildForTable(String tableId, TableOperation tableOpe
...
+    public Map<String, String> buildForTable(UCDeltaTableIdentifier identifier, String location) {
+      Preconditions.checkArgument(identifier != null, "identifier is required");
       Preconditions.checkArgument(location != null && !location.isEmpty(), "location is required");
       validateUcDeltaApi();
       TemporaryCredentials tempCreds =
           DeltaStorageCredentialUtil.toTemporaryCredentials(initialStorageCredential);
-      Map<String, String> props =
-          CredPropsUtil.createTableCredProps(
+      TableOperation tableOp =
+          DeltaStorageCredentialUtil.toTableOperation(initialStorageCredential.getOperation());
```
openinx: With this context (https://github.com/unitycatalog/unitycatalog/pull/1548/changes#r3223441388),  I think we can remove this now.  since we can pass the `CredentialOperation` directly.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedKey.java
```diff
@@ -25,14 +28,24 @@ public interface CredScopedKey {
   static CredScopedKey create(URI uri, Configuration conf) {
     String type = conf.get(UCHadoopConfConstants.UC_CREDENTIALS_TYPE_KEY);
     if (UCHadoopConfConstants.UC_CREDENTIALS_TYPE_PATH_VALUE.equals(type)) {
-      String path = conf.get(UCHadoopConfConstants.UC_PATH_KEY);
-      String pathOperation = conf.get(UCHadoopConfConstants.UC_PATH_OPERATION_KEY);
-
-      return new PathCredScopedKey(path, pathOperation);
+      return new PathCredScopedKey(
```
openinx: This change is unrelated, and unnecessary for this.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedKey.java
```diff
@@ -25,14 +28,24 @@ public interface CredScopedKey {
...
+          conf.get(UCHadoopConfConstants.UC_PATH_KEY),
+          conf.get(UCHadoopConfConstants.UC_PATH_OPERATION_KEY));
     } else if (UCHadoopConfConstants.UC_CREDENTIALS_TYPE_TABLE_VALUE.equals(type)) {
-      String tableId = conf.get(UCHadoopConfConstants.UC_TABLE_ID_KEY);
-      String tableOperation = conf.get(UCHadoopConfConstants.UC_TABLE_OPERATION_KEY);
-      return new TableCredScopedKey(tableId, tableOperation);
+      if (conf.getBoolean(
+          UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_KEY,
+          UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_DEFAULT_VALUE)) {
```
openinx, line 38: Pls define a separate local variable for parsing the hadoop conf key, that's better for readibility.

openinx: Same as the following hadoop keys.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -82,6 +83,18 @@ public T tableId(String tableId) {
       return self();
     }
 
+    public T ucDeltaTableIdentifier(UCDeltaTableIdentifier identifier, String location) {
```
openinx, line 89: the `tableId` and `ucDeltaTableIdentifier` are exclusive to each other. the former is for old client api, and the latter one is for new client api. 

so we'd better to add the Precondition.check for each of them, so that people won't mess up to set both.  This is required to add to the unit tests.

openinx:
Hi @TimothyW553 , pls remove the `[codex]` in the PR tittle, and use the `[Hadoop]` instead.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCDeltaTableIdentifier.java
timothyw553: done. replaced Guava Preconditions with UC Preconditions.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCDeltaTableIdentifier.java
timothyw553: done. added Javadocs for the public API.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedKey.java
timothyw553: done. added location to the Delta table cache key and tests.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedKey.java
timothyw553: done. removed UID from the filesystem cache key.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
timothyw553: done. used CredentialOperation directly in Delta credential props.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
timothyw553: done. removed the TableOperation conversion.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedKey.java
timothyw553: done. removed the unrelated change.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedKey.java
timothyw553: done. added local variables for Hadoop conf keys.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
timothyw553: done. added precondition checks and tests for exclusive table ID and Delta identifier.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
```diff
@@ -183,47 +184,30 @@ public Map<String, String> buildForTable(String tableId, TableOperation tableOpe
     /**
      * Builds Hadoop properties for a UC Delta table's storage location.
      *
-     * @param tableId the UC table UUID, used to scope the credential cache and filesystem cache
-     * @param catalog the UC catalog name
-     * @param schema the UC schema name
-     * @param table the UC table name
```
openinx, line 189: We are not adding the javadoc for `UCDeltaTableIdentifier` in this public API.

timothyw553: oops yes, added.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/CredPropsUtilTest.java
```diff
@@ -164,6 +171,239 @@ void originalImplNotSetWhenCredScopedFsDisabled() {
...
+      throws Throwable {
+    Class<?> propsBuilderClass = builder.getClass().getSuperclass();
+    Method method = propsBuilderClass.getDeclaredMethod(methodName, parameterTypes);
+    method.setAccessible(true);
+    try {
+      return method.invoke(builder, args);
+    } catch (InvocationTargetException e) {
+      throw e.getCause();
+    }
+  }
```
openinx: Here, we can simplify it. actually we don't need to use the  java reflection to do the test. Just make the `S3PropsBuilder` to be package-access, should be good enough to do the entire tests. that's much simpler.

openinx (approved):
Looks good to me now, thanks @TimothyW553 for the revisiting work.

PR #1549, opened 2026-05-12
[Hadoop] Remove initial credential builder APIs

timothyw553:
**PR Checklist**

**Description of changes**

Remove the initial credential builder APIs from Hadoop credential configuration and route credential property generation through the temporary credential API path.

- `PathOperation` and `TableOperation` are now nested enums inside `UCCredentialHadoopConfs` (the single public API class) rather than standalone top-level classes. The enum `value()` method now delegates to `name()` since the string representation matches the enum constant name in every case, eliminating the redundant private field and constructor.
- `TempCredentialApi` is renamed to `GenericCredentialFetcher` (and its implementations `UCTempCredentialApi` → `UCGenericCredentialFetcher`, `UCDeltaTempCredentialApi` → `UCDeltaGenericCredentialFetcher`) to match the naming convention where `GenericCredentialProvider` handles renewal/caching and the fetcher handles one-shot credential retrieval. Static factory methods `forUc()` and `forUcDelta()` are added so tests and callers can construct fetchers without depending on the package-private implementation classes.
- `TempCredentialApiFactory` in `CredPropsUtil` is renamed to `GenericCredentialFetcherFactory`.
- The three `fetch*CredProps` methods in `CredPropsUtil` are refactored to share a single `fetchTemporaryCredentials` helper, eliminating duplicated API-call chains.
- `UCCredentialHadoopConfs.Builder` gains an optional `apiClient(ApiClient)` setter. When provided, the supplied client is reused for the initial credential fetch instead of creating a new HTTP connection pool. `UCSingleCatalog` passes its existing `apiClient` so the connection pool is shared.
- `UCCredentialHadoopConfs.Builder` gains an `addAppVersions(Map)` method. App versions (name → version strings, e.g. `{"Spark": "4.0.0"}`) are threaded through the `PropsBuilder` via `appVersions()` so vended-token providers can include them in the User-Agent header on credential renewal calls. `PropsBuilder` exposes the new `appVersions(Map)` builder method for this purpose.

Validation:
- `./build/sbt -mem 4096 hadoop/test`
- `./build/sbt -mem 4096 spark/compile`
- `./build/sbt -mem 4096 "spark/testOnly io.unitycatalog.spark.UCSingleCatalogStagingTableTest"`


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
```diff
@@ -13,10 +13,31 @@
  * Creates internal Hadoop credential wrappers from Unity Catalog temporary credential APIs.
  *
  * <p>This is an adapter over SDK-generated API clients, not a generated API client itself.
+ *
+ * <p><b>Internal API — not for external use. May change without notice.</b>
  */
-interface TempCredentialApi {
+public interface TempCredentialApi {
```
openinx: Talked to @tdas offline ,  and we agreed this class name will be more clear to use `GenericCredentialFetcher`, and the fetcher will fetch the one-time credential. while the `GenericCredenitalProvider` will handle the renewal and credential caching stuff. 

For `UCTempCredentialApi`, pls also rename it as `UCGenericCredentialFetcher`. 
For `UCDeltaTempCredentialApi`, pls rename it as `UCDeltaGenericCredentialFetcher`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
openinx: I think we should use the `conf.getBoolean` directly, which will simply the whole code piece. 
```java
conf.getBoolean(
            UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_KEY,
            UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_DEFAULT_VALUE)
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/TempCredentialApi.java
```diff
@@ -13,10 +13,31 @@
...
+            UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_DEFAULT_VALUE);
+    if (useDeltaCredentialsApi) {
+      return new UCDeltaTempCredentialApi(
+          conf, new io.unitycatalog.client.delta.api.TemporaryCredentialsApi(apiClient));
+    }
+    return new UCTempCredentialApi(
+        conf, new io.unitycatalog.client.api.TemporaryCredentialsApi(apiClient));
+  }
+
   static TempCredentialApi create(Configuration conf) {
```
openinx: I think this `create(Configuration conf)` should fully reuse and call `create(apiClient, conf)`, pls don't duplicate code in this method again.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/PathOperation.java
```diff
@@ -0,0 +1,22 @@
+package io.unitycatalog.hadoop;
+
+/**
+ * The access operation requested for a Unity Catalog external path credential.
+ *
+ * @since 0.5.0
+ */
+public enum PathOperation {
```
openinx: Talked to @tdas offline, we agreed to move this `PathOperation` inside the `UCCredentialHadoopConfs`. so that we only have one single public API class to maintain.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/TableOperation.java
```diff
@@ -0,0 +1,23 @@
+package io.unitycatalog.hadoop;
+
+/**
+ * The access operation requested for a Unity Catalog table credential.
+ *
+ * <p>Shared across both the UC REST credentials API and the UC Delta credentials API.
+ *
+ * @since 0.5.0
+ */
+public enum TableOperation {
```
openinx: Same as the `PathOperation`  API.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -610,4 +628,149 @@ public static Map<String, String> createPathCredProps(
...
+  }
+
+  private static TemporaryCredentials fetchTemporaryCredentials(
+      String catalogUri,
+      TokenProvider tokenProvider,
+      Map<String, String> engineVersionProps,
+      Configuration reqConf)
+      throws ApiException {
+    return genericCredFetcherFactory
+        .create(createApiClient(catalogUri, tokenProvider, engineVersionProps), reqConf)
```
openinx: we shouldn't create a separate api client for every temporary credential creation, that will spawn too many http connection pool.  instead we should fully reuse the api client that created in the UCSingleCatalog for the unitycatalog spark connector. 

That means we may need to add the apiClient inside the `UCCredentialHadoopConfs.Builder`, and with that builder set,  we can pass the `apiClient` that we already have in the `UCSingleCatlaog` and reuse it in the initial temporary credential process.

openinx (approved):
Looks good to me now,  let's see what will Github CI says.

PR #1551, opened 2026-05-12
[Build] Use GCS Maven Central mirror unconditionally to avoid HTTP 429s

seewishnew:
**PR Checklist**

**Description of changes**

GitHub Actions CI runners are getting HTTP 429 (Too Many Requests) responses from `repo1.maven.org` during dependency resolution, causing build failures. The GCS Maven Central mirror (`maven-central.storage-download.googleapis.com`) is already configured in `build/sbt-config/repositories`, but it is never activated because `build/sbt` only applies it when `JENKINS_URL` is set — which is not the case on GitHub Actions.

Changes:
- `build/sbt`: Always activate the `sbt-config/repositories` config instead of gating on `JENKINS_URL`. Appends to `SBT_OPTS` rather than overwriting, so callers can pass additional JVM flags via the environment.
- `build/sbt-config/repositories`: Fix stale GCS mirror URL (`/repos/central/data/` → `/maven2/`).
- `build/sbt-launch-lib.bash`: Default the sbt-launch.jar download to the GCS mirror, with automatic fallback to `repo1.maven.org` if the mirror is unavailable.

See #1550 for CI evidence of 429 failures without this fix.

### DIFF
--- a/build/sbt
+++ b/build/sbt
@@ -75,12 +75,12 @@ if [[ -n "$MAVEN_PROXY_URL" ]]; then
   maven-proxy: $MAVEN_PROXY_URL
   maven-proxy-ivy: $MAVEN_PROXY_URL, [organization]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext]
 EOF
-  export SBT_OPTS="-Dsbt.override.build.repos=true -Dsbt.repository.config=$SBT_REPOSITORIES_CONFIG"
-elif [[ "$JENKINS_URL" != "" ]]; then
-  # Make Jenkins use Google Mirror first as Maven Central may ban us
+else
+  # Default: use sbt-config/repositories which lists GCS Maven Central mirror before maven-central,
+  # avoiding HTTP 429 rate-limiting from repo1.maven.org on CI runners.
   SBT_REPOSITORIES_CONFIG="$(dirname "$(realpath "$0")")/sbt-config/repositories"
-  export SBT_OPTS="-Dsbt.override.build.repos=true -Dsbt.repository.config=$SBT_REPOSITORIES_CONFIG"
 fi
+export SBT_OPTS="${SBT_OPTS:-} -Dsbt.override.build.repos=true -Dsbt.repository.config=$SBT_REPOSITORIES_CONFIG"
 
 . "$(dirname "$(realpath "$0")")"/sbt-launch-lib.bash
 
--- a/build/sbt-config/repositories
+++ b/build/sbt-config/repositories
@@ -1,8 +1,9 @@
 [repositories]
   local
+  maven-local: file://${user.home}/.m2/repository
   local-preloaded-ivy: file:///${sbt.preloaded-${sbt.global.base-${user.home}/.sbt}/preloaded/}, [organization]/[module]/[revision]/[type]s/[artifact](-[classifier]).[ext]
   local-preloaded: file:///${sbt.preloaded-${sbt.global.base-${user.home}/.sbt}/preloaded/}
-  gcs-maven-central-mirror: https://maven-central.storage-download.googleapis.com/repos/central/data/
+  gcs-maven-central-mirror: https://maven-central.storage-download.googleapis.com/maven2/
   maven-central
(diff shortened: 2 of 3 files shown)


build/sbt
yili-db: So this branch is useless because our repo doesn't use jenkins?
then it should be turned into a pure `else` branch. And the change should be placed in this else branch.
Then the line `export SBT_OPTS="-Dsbt.override.build.repos=true -Dsbt.repository.config=$SBT_REPOSITORIES_CONFIG"` can be moved after `fi` as always executed.

yili-db:
```suggestion

if [[ -n "$MAVEN_PROXY_URL" ]]; then
# If MAVEN_PROXY_URL is set, use it (and local) as the sole repository for all dependencies.
SBT_REPOSITORIES_CONFIG=$(mktemp)
cat > "$SBT_REPOSITORIES_CONFIG" <<EOF
[repositories]
 local
 maven-proxy: $MAVEN_PROXY_URL
 maven-proxy-ivy: $MAVEN_PROXY_URL, [organization]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext]
EOF
else
# Default: use sbt-config/repositories which lists GCS Maven Central mirror before maven-central,
# avoiding HTTP 429 rate-limiting from repo1.maven.org on CI runners.
SBT_REPOSITORIES_CONFIG="$(dirname "$(realpath "$0")")/sbt-config/repositories"
fi
export SBT_OPTS="${SBT_OPTS:-} -Dsbt.override.build.repos=true -Dsbt.repository.config=$SBT_REPOSITORIES_CONFIG"
```

seewishnew: I'm not sure how JENKINS_URL plays a role here or in other infra.

yili-db: It doesn't matter anymore. With your change, the `else` behavior matches this `else if` anyway.

yili-db (approved):

build/sbt-config/repositories
```diff
@@ -1,8 +1,9 @@
 [repositories]
   local
+  maven-local: file://${user.home}/.m2/repository
   local-preloaded-ivy: file:///${sbt.preloaded-${sbt.global.base-${user.home}/.sbt}/preloaded/}, [organization]/[module]/[revision]/[type]s/[artifact](-[classifier]).[ext]
   local-preloaded: file:///${sbt.preloaded-${sbt.global.base-${user.home}/.sbt}/preloaded/}
-  gcs-maven-central-mirror: https://maven-central.storage-download.googleapis.com/repos/central/data/
+  gcs-maven-central-mirror: https://maven-central.storage-download.googleapis.com/maven2/
```
openinx, line 6: what's the difference for this line change ?

seewishnew: This is the newer, more preferred API, apparently.

openinx (approved):

build/sbt-launch-lib.bash
```diff
@@ -45,7 +45,7 @@ acquire_sbt_jar () {
   elif [[ "${SBT_VERSION}" == "1.5.5" ]] && [[ -n "${SBT_1_5_5_MIRROR_JAR_URL}" ]]; then
     URL1="${SBT_1_5_5_MIRROR_JAR_URL}"
   else
-    URL1=${DEFAULT_ARTIFACT_REPOSITORY:-${MAVEN_PROXY_URL:-https://repo1.maven.org/maven2}}/org/scala-sbt/sbt-launch/${SBT_VERSION}/sbt-launch-${SBT_VERSION}.jar
+    URL1=${DEFAULT_ARTIFACT_REPOSITORY:-${MAVEN_PROXY_URL:-https://maven-central.storage-download.googleapis.com/maven2}}/org/scala-sbt/sbt-launch/${SBT_VERSION}/sbt-launch-${SBT_VERSION}.jar
```
raveeram-db, line 48: @seewishnew can we add this as a fallback? IIRC, we could provide a list?

seewishnew: There's no straightforward list pattern, but in [delta](https://github.com/delta-io/delta/blob/0ab1e8d74dcfbbb55b1256799416c3c71adab998/build/sbt-launch-lib.bash#L68-L86) there's a backup url that we use and manually fall back to in the code if the first download attempt fails.

I'd initially thought of adding it here, but then I figured it may be overkill in this case and may involve overcrowding too many changes at once in this PR, making it harder to reason about.

Lemme add that here for good measure.

raveeram-db (approved):

tdas (approved):

PR #1553, opened 2026-05-14
[Spark] Cross-build connector per Spark major.minor version

chenwang-databricks:
**PR Checklist**

**Description of changes**

Add the SBT plumbing to publish the UC Spark connector as three separately-named artifacts -- `unitycatalog-spark_4.0_2.13`, `unitycatalog-spark_4.1_2.13`, `unitycatalog-spark_4.2_2.13` -- one per supported Spark major.minor. This mirrors Delta Lake's [`project/CrossSparkVersions.scala`](https://github.com/delta-io/delta/blob/master/project/CrossSparkVersions.scala) pattern.

This PR is purely build / release infra; the connector source code stays table-only at this stage. A follow-up PR will add the `TableViewCatalog` stub and the `MetricViewSupport` trait inside the per-version `scala-shims/spark-X.Y/` directories that this PR creates, and move the metric-view methods accordingly.

### 1. `project/CrossSparkVersions.scala` (new, ~95 lines)

Defines `SparkVersionSpec` for the three supported Spark versions (4.0.0, 4.1.0, 4.2.0-SNAPSHOT), the `_${shortVersion}` artifact-suffix helper (`sparkVersionedModuleName`), and `sparkSourceDirSettings` -- which appends `src/main/scala-shims/spark-X.Y/` and the matching test directory to `unmanagedSourceDirectories`. Modeled on Delta's `CrossSparkVersions.scala` but trimmed to the minimum UC needs today (no release-step plumbing, no MASTER spec, single-module focus).

The Spark version is selected via `-DsparkVersion=...` (full or short form -- both `4.0.0` and `4.0` resolve to the spark40 spec). Defaults to `4.0.0`. Setting `-DskipSparkSuffix=true` returns the legacy un-suffixed module name for the optional backward-compat release pass.

### 2. `build.sbt` -- `spark` module changes

- `Keys.moduleName := CrossSparkVersions.sparkVersionedModuleName(name.value)` so the published artifact becomes `unitycatalog-spark_${shortVersion}_2.13`.
- `CrossSparkVersions.sparkSourceDirSettings` to wire the per-version source directory into compile and test scopes.
- Drop the previous `4.0 || 4.2` conditional Delta dep gate -- under cross-build, `delta-spark_$sparkMajorMinorVersion` is uniform per spec. Same cleanup on the `integrationTests` module.

### 3. Empty `scala-shims/spark-{4.0,4.1,4.2}/` directories (six `.gitkeep` files)

Under both `connectors/spark/src/main/` and `connectors/spark/src/test/`. Empty in this PR so SBT can resolve the source-dir settings cleanly. The follow-up PR will land:

- `scala-shims/spark-4.0/org/apache/spark/sql/connector/catalog/TableViewCatalog.scala` and the same in `spark-4.1/` -- a stub trait so `class UCSingleCatalog ... with TableViewCatalog` compiles on Spark versions that don't have the real interface (Delta uses the same trick for `SupportsV1OverwriteWithSaveAsTable`).
- `scala-shims/spark-4.2/io/unitycatalog/spark/MetricViewSupport.scala` -- the real metric-view method bodies, mixed into `UCSingleCatalog` per Spark version.

### 4. `.github/workflows/unit-tests.yml` -- 3-cell test matrix

Replace the `spark-compile-only` Spark 4.2 cell that #1518 introduced (which excluded Delta entirely because no 4.2 artifact existed) with a full `spark-tests` cell that pulls `delta-spark_4.2_2.13:4.1.0-SNAPSHOT`. Final matrix:

| Spark              | Delta              | Validation mode | Blocking |
|--------------------|--------------------|------------------|----------|
| 4.0.0              | 4.1.0              | all-tests        | yes      |
| 4.1.0              | 4.1.0              | spark-tests      | yes      |
| 4.1.0              | master-SNAPSHOT    | spark-tests      | no       |
| 4.2.0-SNAPSHOT     | 4.1.0-SNAPSHOT     | spark-tests      | no       |

The 4.2 cell stays non-blocking until Delta publishes a stable `delta-spark_4.2_2.13` artifact; flip to blocking in a follow-up.

## Out of scope for this PR

- No new connector source code -- the metric-view feature stays out of this PR.
- The empty shim directories carry no view-side overrides yet; that's the follow-up.
- No consumer-facing artifact-name documentation changes -- those land with the follow-up that re-introduces metric views.

## Dependency

This PR builds on top of [#1518](https://github.com/unitycatalog/unitycatalog/pull/1518). #1518 adds the Apache Snapshots resolver and the `SparkCatalogCompatibility` reflection shim that the 4.2 cell needs. Please merge #1518 first.

## Test plan

- [x] `for v in 4.0.0 4.1.0 4.2.0-SNAPSHOT; do build/sbt -DsparkVersion=$v ";clean ;spark/package" ; done` produces the three suffixed JARs.
- [x] `build/sbt -DsparkVersion=$v spark/publishM2` lands each at its own Maven coordinates under `~/.m2/repository/io/unitycatalog/unitycatalog-spark_${shortVer}_2.13/`.
- [x] `spark-sql` smoke against UC OSS server with the locally-published artifact for each Spark major.minor passes 11/11 SQL statements covering `SHOW NAMESPACES`, `CREATE SCHEMA`, `CREATE TABLE`, `INSERT`, `SELECT`, `DESCRIBE EXTENDED`, `DROP TABLE`, `DROP SCHEMA`, `SHOW TABLES`. No JVM linkage errors on any version.
- [ ] CI matrix runs all three blocking cells green plus the two non-blocking cells.


connectors/spark/src/main/scala/io/unitycatalog/spark/compat/SparkCatalogCompatibility.scala
```diff
@@ -0,0 +1,43 @@
...
+object SparkCatalogCompatibility {
+
+  /**
+   * Builds a CatalogStorageFormat across Spark versions.
+   *
+   * Spark 4.2 adds a `serdeName` field to CatalogStorageFormat.copy, while Spark 4.0 and
+   * Spark 4.1 do not have that parameter. Keep the ABI-specific reflection in this shim so
+   * connector code can use one compatibility entry point.
+   */
+  def catalogStorageFormatWithLocation(
```
openinx, line 19: Hi @chenwang-databricks ,  thanks for the contribution.  I think this is not the correct approach how is shims layer working for supporting different spark engine compatibility.

openinx: If the difference for diff spark versions, comes from the `CatalogStorageFormat#copy`,  then the correct approach is: 

1.  define the `CatalogStorageFormat` in `scala-shims` dir's corresponding `spark-4.x` version, and then provide a the same method signature. 

2. Provide the corresponding method implementation for the same method signature. 

3. in the compilation, we should only one of the spark shims to build the project, with this, every spark version will be able to pass the compilation, since the `shims` layer clearly defines the same method signature, and in the runtime  it will eventually delegate to different spark impl. 

You can hack with delta's `scala-shims` , that's exactly how it works. https://github.com/delta-io/delta/tree/master/spark/src/main/scala-shims

PR #1554, opened 2026-05-14
[Spark] Add metric-view connector via per-Spark-version shims

chenwang-databricks:
**PR Checklist**

- [x] If there is a related issue, make sure it is linked to this PR. (#1513 -- this PR supersedes its connector content with a properly-shimmed version.)

**Description of changes**

Re-introduce the V2 view connector (`createView`, `loadView`, `loadTableOrView`, `dropView`, `listViews`, `renameView`, the `MetadataTable + ViewInfo` view-wrapping in `UCProxy.loadTable`, etc.) on top of the cross-Spark build infra now on `main`. Per-version `scala-shims/spark-X.Y/` directories carry the parts that touch Spark-4.2-only types; the shared sources stay free of any reference to those types so the connector compiles cleanly against Spark 4.0, 4.1, and 4.2.

The trait surface is generic to V2 view-side support and is not specific to metric views, even though metric views are the only view kind enabled today via `UCSingleCatalog.viewLikeUcTypes` -- future view kinds (e.g. `MATERIALIZED_VIEW`) will reuse the same trait without any further rename.

This mirrors Delta Lake's `scala-shims/spark-X.Y/` pattern (e.g. Delta's [`SupportsV1OverwriteWithSaveAsTable.scala`](https://github.com/delta-io/delta/blob/master/spark/src/main/scala-shims/spark-4.0/SupportsV1OverwriteWithSaveAsTable.scala) stub) and supersedes the metric-view content from [#1513](https://github.com/unitycatalog/unitycatalog/pull/1513), which was un-shimmed and only compiled on Spark 4.2.

### A. "Fake interface" stub for `TableViewCatalog`

- `connectors/spark/src/main/scala-shims/spark-4.0/org/apache/spark/sql/connector/catalog/TableViewCatalog.scala` and the same in `spark-4.1/`: empty `trait TableViewCatalog extends TableCatalog`. Lets `class UCSingleCatalog ... with TableViewCatalog` compile on all three Spark versions.
- No file in `scala-shims/spark-4.2/` -- the real Spark 4.2 interface is on the classpath.

### B. Per-version `ViewSupport` and `UCProxyViewSupport` traits

Mixed into `UCSingleCatalog` and `UCProxy` respectively. Self-typed against the host class so the trait can read the host's protected state (`delegate`, `ucProxy`, `tablesApi`, etc.).

- `scala-shims/spark-4.2/io/unitycatalog/spark/ViewSupport.scala` (real impl, ~85 lines): the view-side overrides on `UCSingleCatalog` plus the `rejectIfShimmedView` hook used by the shared `loadTable`.
- `scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala` (real impl, ~305 lines): view-side overrides on `UCProxy` (`createView`, `dropView`, `listViews`, `loadTableOrView`, `createTable(Identifier, TableInfo)`, the `wrapAsView` hook for `loadTable`, the `toViewInfo` builder, the dependency-list converters, and `buildColumnInfos` -- everything that touches `MetadataTable`, `ViewInfo`, `Dependency`, `DependencyList`, the new Spark 4.2 `TableInfo`).
- `scala-shims/spark-4.0/.../ViewSupport.scala`, `scala-shims/spark-4.1/.../ViewSupport.scala`: empty equivalents that expose a no-op `rejectIfShimmedView` (the catalog never sees view-shaped tables on those versions).
- `scala-shims/spark-4.0/.../UCProxyViewSupport.scala`, `scala-shims/spark-4.1/.../UCProxyViewSupport.scala`: empty equivalents that expose a `wrapAsView` hook throwing `NoSuchTableException` (the shared `UCProxy.loadTable` only ever calls `wrapAsView` for view-like UC table types -- which the shared `viewLikeUcTypes` map identifies, but those types only have UI/SQL meaning on Spark 4.2).

## Other supporting changes

- `UCSingleCatalog` mixes in `ViewSupport`. The shared `loadTable` no longer references `MetadataTable` / `ViewInfo` directly -- it calls `rejectIfShimmedView` instead.
- `UCProxy` becomes `private[spark]`; the constructor's `tablesApi` parameter and the `listUcTables` / `convertDataTypeToTypeName` / `getUcTable` helpers gain `protected[spark]` / `private[spark]` visibility so the `UCProxyViewSupport` trait can call into them via the self-type.
- `UCSingleCatalog.viewLikeUcTypes` now uses the literal `"METRIC_VIEW"` string instead of `TableSummary.METRIC_VIEW_TABLE_TYPE` (a Spark-4.2-only constant). Since `viewLikeUcTypes` is referenced from shared code on all three Spark versions, the literal lets it stay shared without requiring a per-version companion split.
- `UCTableProperties.PROP_TABLE_TYPE` is mirrored locally so we don't depend on `TableCatalog.PROP_TABLE_TYPE` (Spark 4.1+ only). Value tracks Spark's literal `"table_type"`.
- `build.sbt`: the spark module's **test** compile uses `--release 17` only on the 4.2 build (the 4.2-only `UCViewProxySuite` references Spark 4.2 Java records, e.g. the V2 `TableInfo` / `Dependency` surface); 4.0/4.1 stay on `--release 11`, and production `Compile` stays at `--release 11` on every version.

## Robustness hardening (from code review)

- `UCProxy.loadTable` routes every view-like row through `wrapAsView`, including listed-but-unmapped kinds (`MATERIALIZED_VIEW`, which maps to `None` in `viewLikeUcTypes`). `wrapAsView` now gates on `isViewCommandsSupportedTableType` and throws `NoSuchTableException` for unmapped kinds, so Spark's resolver sees a recognized control-flow exception instead of the `UnsupportedOperationException` that `toViewInfo -> ucTableTypeToSparkViewType` would otherwise raise during ordinary identifier resolution.
- `UCColumnJson.parseStructFieldJson` guards a null `ColumnInfo.type_json` (nullable on the wire for view-like rows written by an older writer or a non-Spark client) with a clear error instead of letting `json4s.parse(null)` throw a raw NPE out of `loadView` / `loadTable`. `toViewInfo` additionally guards the outer `getColumns` collection itself with `Option(...)` (defaulting to an empty schema), mirroring the existing `Option(t.getProperties)` handling, so a null columns list degrades gracefully rather than NPE-ing on the view-load path.
- `loadView` distinguishes three cases so the error matches what the user sees from `SHOW VIEWS`: a genuine non-view (a real table at the identifier) and a 404 both surface as `NoSuchViewException` (the latter translated from `getUcTable`'s `NoSuchTableException`, mirroring `dropView`, so `IF EXISTS` works); a listed-but-unmapped view kind (`MATERIALIZED_VIEW`, which `listViews` *does* return) surfaces as `UnsupportedOperationException` -- reporting `NoSuchViewException` there would contradict its presence in `SHOW VIEWS`. (`wrapAsView`, on the `loadTable` table-resolution path, still throws `NoSuchTableException` for unmapped kinds, since Spark's table resolver only understands that.)
- `toViewInfo` strips the `VIEW_SQL_CONFIG_PREFIX` keys from `properties()` after lifting them into `sqlConfigs()`, so each piece of view metadata lives in exactly one place and the prefixed copies aren't double-persisted on a `createView` / replace round-trip.

Known limitation (intentional): a listed-but-unmapped view kind (e.g. `MATERIALIZED_VIEW`, `None` in `viewLikeUcTypes`) appears in `listViews` but is not yet loadable, creatable, or droppable through this connector -- `loadView` reports `UnsupportedOperationException`, `createView` an `ApiException`, and `dropView` / `dropTable` passive-filter it (so `DROP ... IF EXISTS` no-ops). This keeps the surface consistent ("we list it but don't fully handle this kind yet"); it will be revisited when `MATERIALIZED_VIEW` becomes a mapped/supported kind.

## Review cleanup

- Removed the meaningless `== 200` check after `tablesApi.deleteTable` in both `UCProxy.dropTable` and `dropView` (it returns the empty response body, not an HTTP status; a real failure throws `ApiException`).
- `extractSqlConfigs` is hoisted to the shared `UCSingleCatalog` companion and reused from the 4.2 view shim instead of being duplicated.
- `UCColumnJson` moved to its own shared file `UCColumnJson.scala` (`private[spark]`, public-V2-API-only so it stays out of the shim dirs); deleted an empty orphaned `object UCProxy` and the unused `isViewCommandsSupportedTableType(String)` overload; dropped a now-unused json4s import.
- `buildColumnInfos` carries a comment explaining it intentionally mirrors `UCProxy.createTable`'s column loop but on the Spark-4.2-only V2 `Column`/`TableInfo` surface, so the two cannot be merged without breaking the 4.0/4.1 builds.
- Comment-placement and whitespace fixes (rename-stub note, `ViewSupport` scaladoc bullet, a double blank line).

## Tests

`UCProxySuite` (mock-backed unit tests for `UCProxy`) is split so the version-agnostic surface is exercised on every Spark version while the view tests stay 4.2-only:

- `src/test/java/io/unitycatalog/spark/UCProxyTestFixture.java` -- shared, non-test helper holding the reflective `UCProxy` construction + mock wiring. Exposes the proxy as a bare `Object` so it names no Spark-4.2-only type. Composed (not inherited) by both suites, so no `@Test` methods are shared via inheritance (which would otherwise double-run on the 4.2 build).
- `src/test/java/io/unitycatalog/spark/UCProxySuite.java` -- the **10 version-agnostic tests** (`listTables` pagination x5, `listNamespaces` pagination x3, the single-`getTable`-RPC guard, and `DROP TABLE`-on-metric-view safety). Lives in the shared test tree, so it compiles and runs on **Spark 4.0, 4.1, and 4.2** -- coverage that was previously 4.2-only.
- `src/test/scala-shims/spark-4.2/io/unitycatalog/spark/UCViewProxySuite.java` -- the **10 view / 4.2-only tests** (`createView`/`loadView`/`dropView`/`loadTableOrView`, the V2 `TableInfo` `createTable` overload, `MetadataTable`/`ViewInfo` assertions, and the unmapped-kind `loadView` error). It casts the proxy to `TableViewCatalog`/`ViewCatalog`, which only succeeds on the 4.2 build.

`testLoadTableTriggersExactlyOneGetTableRpc` is hermetic: `loadTable` issues its single `getTable` RPC before credential vending (which has no real endpoint under mocks), so the call is guarded and only the RPC count is asserted -- this both fixes a previously-failing 4.2 test and makes it safe to run on all versions.

## Validation

- `build/sbt -DsparkVersion=$v spark/compile` succeeds for all three versions (4.0.0, 4.1.0, 4.2.0-SNAPSHOT). Bytecode confirms the per-version method set: `javap -p UCSingleCatalog` shows zero view methods on the 4.0/4.1 builds and all view methods (`listViews`, `loadView`, `createView`, `replaceView`, `dropView`, `renameView`, `loadTableOrView`) on the 4.2 build; `wrapAsView` resolves to a `NoSuchTableException` stub on 4.0/4.1 and the real `MetadataTable` impl on 4.2.
- `spark-sql` smoke against UC OSS (Spark 4.0): table-side CRUD passes (`SHOW TABLES`, `CREATE TABLE` parquet, `INSERT`, `SELECT`, `DESCRIBE EXTENDED`, `DROP TABLE`), confirming the shared sources run with no JVM linkage errors when only the empty shims are active and Delta is on the classpath.

## Dependencies

Rebased directly onto `main`, which now carries the cross-Spark build infra ([#1556](https://github.com/unitycatalog/unitycatalog/pull/1556): `project/CrossSparkVersions.scala`, versioned artifact names, empty `scala-shims/spark-X.Y/` dirs) and the source-built Spark 4.2 CI from [#1590](https://github.com/unitycatalog/unitycatalog/pull/1590) (the 4.2 target is `4.2.0-SNAPSHOT`, built from the ref in `project/spark-versions.json` and cached; there is no `4.2.0-preview5` entry). No build / CI infra is duplicated in this diff.

Supersedes [#1553](https://github.com/unitycatalog/unitycatalog/pull/1553) (the earlier cross-build PoC PR with overlapping scope).

Cross-repo dependency: when `DeltaCatalog` wraps `UCProxy`, metric-view loading also needs the Delta-side fix in [delta-io/delta#7039](https://github.com/delta-io/delta/pull/7039), which stops `ServerSidePlannedTable.tryCreate` from capturing the `MetadataTable + ViewInfo` view shape in `DeltaCatalog.loadTable` (it now only attempts server-side planning for Delta tables). End-to-end metric-view loading through Delta requires a Delta build that includes that change.

## Test plan

- [x] Spark 4.0 / 4.1 / 4.2 compile cleanly with the per-version shim layout.
- [x] `spark-sql` table-side smoke passes on Spark 4.0.
- [x] `UCProxySuite` (version-agnostic, `src/test/java/`) runs and passes on Spark 4.0 and 4.2 (10/10); `UCViewProxySuite` (4.2-only) passes on 4.2 (10/10). The two suites run once each per version -- no double execution.
- [x] Blocking CI cells green: Build Test, `All tests (Spark 4.0.0)`, `Spark tests (Spark 4.1.0)`, plus lint / java-fmt / DCO / unit `test`.
- [ ] Non-blocking `Spark tests (Spark 4.2.0-SNAPSHOT, Delta master-SNAPSHOT)`: currently red on a source-built-Spark-master vs Delta-master drift (`NoSuchMethodError` on `CatalogTable.<init>` in table-side `*ReadWriteTest` / `*CredRenewITTest`) that reproduces on `main` HEAD; not introduced by this PR.


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -1,54 +1,78 @@
...
-import scala.collection.mutable.ArrayBuffer
-import scala.language.existentials
-
 /**
  * A Spark catalog plugin to get/manage tables in Unity Catalog.
  */
 class UCSingleCatalog
   extends StagingTableCatalog
   with SupportsNamespaces
+  with TableViewCatalog
```
openinx, line 49: Hi @chenwang-databricks ,  I did not see the package import for the `TableViewCatalog`,  what's the menchanism for us to load this `TableViewCatalog` ?

chenwang-databricks: It's in import org.apache.spark.sql.connector.catalog._

build.sbt
```diff
@@ -635,7 +635,14 @@ lazy val spark = (project in file("connectors/spark"))
         .files
         .filter(_.getName.contains("lombok"))
         .mkString(File.pathSeparator)
-      javacRelease11 ++ Seq(
+      // Spark 4.2 connector tests reference Java records (e.g. Dependency.table); require --release 17.
+      val testRelease =
+        if (CrossSparkVersions.getSparkVersionSpec().shortVersion == "4.2") {
+          javacRelease17
```
yili-db, line 641: This is fine. Spark 4.x doesn't support Java 11 anyway.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/ViewSupport.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  override def renameView(oldIdent: Identifier, newIdent: Identifier): Unit =
+    ucProxy.renameView(oldIdent, newIdent)
+
+  /**
+   * Single-RPC entry for `TableViewCatalog`. Forwards to `delegate.loadTable(ident)`,
+   * which (when `delegate` is Delta wrapping `UCProxy`) chains
+   * `Delta.loadTable -> super.loadTable (DelegatingCatalogExtension) -> UCProxy.loadTable`.
+   * `UCProxy.loadTable` issues exactly one UC `getTable` and returns either a `V1Table`
+   * for normal tables or a `MetadataTable + ViewInfo` for views; Delta's `case o => o`
```
yili-db, line 66: > Delta's `case o => o`
There's one problem though:
Before this `case o => o` there's:
```
      ServerSidePlannedTable.tryCreate(spark, ident, table, isUnityCatalog).foreach { sspt =>
        return sspt
      }
```
This will potentially capture and short circuit the view loading. Changes are likely needed on Delta side.

connectors/spark/src/test/java/io/unitycatalog/spark/UCProxySuite.java
```diff
@@ -1,228 +0,0 @@
...
- * Unit tests for UCProxy pagination using mocked API responses.
- *
- * <p>UCProxy is a private Scala class, so Java cannot instantiate it directly. We use reflection
- * with dynamic arg resolution to handle constructor changes across branches.
- *
- * <p>End-to-end integration tests that exercise pagination through UCSingleCatalog and a real UC
- * server are in {@link BaseTableReadWriteTest#testListTablesPagination} and {@link
- * SchemaOperationsTest#testListSchemasPagination}.
- */
-public class UCProxySuite {
```
yili-db, line 44: This test no longer runs with Spark 4.0 and 4.1?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -600,17 +633,84 @@ object UCSingleCatalog {
...
+   *   - `String` overload: caller has a Spark `TableSummary` view-type string (e.g. from
+   *     `ViewInfo.properties().get(PROP_TABLE_TYPE)` on the create path).
+   *   - `TableType` overload: caller has a UC `TableType` (e.g. from `getUcTable(...)` on
+   *     the drop path).
+   *
+   * Both return `true` iff the kind is in `viewLikeUcTypes` AND has a Spark mapping
+   * (i.e. `Some(_)` in the map). Listed-but-unmapped kinds (`None`) are filtered into
+   * `listViews` for surface correctness but rejected from view CRUD commands.
+   */
+  def isViewCommandsSupportedTableType(sparkType: String): Boolean =
```
yili-db: Is this ever used?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -725,6 +852,12 @@ private class UCProxy(
       .asInstanceOf[Table]
   }
 
+  // UC OSS does not currently expose a rename endpoint for tables or views; mirror the
```
yili-db: This comment belongs to where the renameTable is defined?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -942,3 +1107,130 @@ private class UCProxy(
     true
   }
 }
+
+private object UCProxy {
+
+  /**
```
yili-db: Are these change logs?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -942,3 +1107,130 @@ private class UCProxy(
...
+ *
+ * No `private[sql]` helpers, no Spark API additions. The shape is just
+ * `{name, type, nullable, metadata}`.
+ *
+ * Comment handling mirrors Spark's StructField wire convention: comments live inside
+ * `metadata` under the `"comment"` key on the wire, but on the in-memory `Column` they are
+ * exposed via the dedicated [[Column.comment]] accessor. The helpers merge / strip the
+ * `"comment"` key at the JSON boundary so callers never see it duplicated.
+ */
+private[spark] object UCColumnJson {
```
yili-db: this object isn't used in this file. It worth a separate file.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/ViewSupport.scala
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Spark-4.2-only mixin providing the view-side overrides that
+ * [[org.apache.spark.sql.connector.catalog.TableViewCatalog]] adds. Mixed into
+ * `UCSingleCatalog` and resolved per Spark version via the `scala-shims/spark-X.Y/`
+ * directory mechanism (see `project/CrossSparkVersions.scala`):
+ *
+ *   - `scala-shims/spark-4.0/io/unitycatalog/spark/ViewSupport.scala`: empty trait
+ *   - `scala-shims/spark-4.1/io/unitycatalog/spark/ViewSupport.scala`: empty trait
+ *   - this file:                                                              real impl
```
yili-db: indentation

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,307 @@
...
+    // Known limitation: a listed-but-unmapped view kind (e.g. MATERIALIZED_VIEW, `None` in
+    // `viewLikeUcTypes`) is rejected here AND by `UCProxy.dropTable` (which passive-filters all
+    // `isViewLikeTableType` rows), so such a row is not droppable through this connector and a
+    // `DROP ... IF EXISTS` no-ops. This is intentional: the connector treats unsupported view
+    // kinds as uniformly inert (no load / create / drop) rather than allowing a drop of a kind it
+    // can't otherwise handle. Revisit when MATERIALIZED_VIEW becomes a supported (mapped) kind.
+    if (!UCSingleCatalog.isViewCommandsSupportedTableType(t.getTableType)) {
+      return false
+    }
+    tablesApi.deleteTable(UCSingleCatalog.fullTableNameForApi(this.name, ident)) == 200
```
yili-db: This is the existing pattern but it's cosmetically wrong for both table and view.
The function `tablesApi.deleteTable` is not meant to return any useful things. If the deletion is successful it returns with no exception.
Its return value is actually the json response body that doesn't exist or is simply empty.
It worked simply because no one checks for the boolean return value.
While you are here, please remove this `== 200` from both places.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,307 @@
...
+  private[spark] def toViewInfo(t: UCTableInfo): ViewInfo = {
+    // Guard the columns collection itself (nullable on the wire for a row written by an older or
+    // non-Spark client), mirroring the `Option(t.getProperties)` handling below and the per-column
+    // `type_json` guard in `parseStructFieldJson` -- so a missing list degrades to an empty schema
+    // rather than NPE-ing out of `loadView` / `loadTable`.
+    val columns = Option(t.getColumns).map(_.asScala).getOrElse(Seq.empty).map { col =>
+      UCColumnJson.parseStructFieldJson(col.getTypeJson)
+    }.toArray
+
+    val props = new util.HashMap[String, String]()
```
yili-db, line 200:
```
val props = new util.HashMap[String, String](Option(t.getProperties).orElse(Map.of()))
```

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,307 @@
...
+    Option(t.getProperties).foreach(props.putAll)
+    val sqlConfigs = extractSqlConfigs(props)
+    // The VIEW_SQL_CONFIG_PREFIX keys are surfaced (un-prefixed) via `withSqlConfigs`; drop them
+    // from `props` so they don't also leak into the user-visible `properties()` map and get
+    // re-persisted (double-counted) on a createView/replace round-trip.
+    props.keySet().removeIf(_.startsWith(CatalogTable.VIEW_SQL_CONFIG_PREFIX))
```
yili-db, line 206: should this line be moved into `extractSqlConfigs`?

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,307 @@
...
+      .withSchemaMode("UNSUPPORTED")
+      .withQueryColumnNames(columns.map(_.name()))
+    Option(t.getComment).foreach(builder.withComment)
+    Option(t.getViewDependencies).foreach { ucDeps =>
+      builder.withViewDependencies(fromUcDependencyList(ucDeps))
+    }
+    builder.build()
+  }
+
+  private def extractSqlConfigs(properties: util.Map[String, String]): util.Map[String, String] = {
```
yili-db: This is a dup. Can it be reused?

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,307 @@
...
+    DependencyList.of(sparkDeps)
+  }
+
+  /**
+   * Builds the UC column-info list for `createView` from a Spark `TableInfo` (or its
+   * `ViewInfo` subtype). Preserves nullability, comment, partition position, and the
+   * canonical Spark column JSON in `typeJson` so column-level analyzer metadata
+   * (`metric_view.type`, `metric_view.expr`) survives the round-trip through Unity Catalog.
+   */
+  private def buildColumnInfos(
```
yili-db, line 271: This function is actually shared between table and view. It duplicates with the same code in UCProxy.createTable

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,307 @@
...
+   * `NoSuchTableException` -- so a kind we list but cannot build a Spark `ViewInfo` for (no
+   * `TableSummary` mapping) must surface as `NoSuchTableException`, not the
+   * `UnsupportedOperationException` that `toViewInfo` -> `ucTableTypeToSparkViewType` would
+   * otherwise throw.
+   */
+  protected def wrapAsView(t: UCTableInfo, ident: Identifier): Table =
+    if (UCSingleCatalog.isViewCommandsSupportedTableType(t.getTableType)) {
+      new MetadataTable(toViewInfo(t), ident.toString)
+    } else {
+      throw new NoSuchTableException(ident)
```
yili-db, line 69: should be `NoSuchViewException`?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -649,6 +757,25 @@ private class UCProxy(
...
+  // because no view types exist there). Everything else goes through `loadV1Table`.
+  override def loadTable(ident: Identifier): Table = {
+    val t = getUcTable(ident)
+    if (UCSingleCatalog.isViewLikeTableType(t.getTableType)) {
+      wrapAsView(t, ident)
+    } else {
+      loadV1Table(ident, t)
+    }
+  }
+
```
yili-db, line 781: double empty line

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: Removed

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: Moved the comment

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: OK deleted

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: Moved to a new file

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/ViewSupport.scala
chenwang-databricks: Fixed

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: Kept the explicit new util.HashMap + Option(t.getProperties).foreach(putAll) — the suggested Option(...).orElse(Map.of()) constructor would need the columns/null handling we already do, and the explicit form reads clearer alongside the adjacent null guards. Happy to switch if you feel strongly.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: Left it in toViewInfo. extractSqlConfigs is a pure read (returns the un-prefixed configs), whereas this line mutates the caller's props to strip the prefixed keys so they don't leak into properties(). Different concern + a side effect on the caller's map, so I kept them separate.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: They do look ~95% alike, but they can't be merged: buildColumnInfos operates on the Spark V2 TableInfo / Column surface that only exists on Spark 4.2, while the shared UCProxy.createTable works on StructType / StructField (via toStructFieldJson) and can't reference those V2 types without breaking the 4.0/4.1 builds. I added a comment documenting that the near-duplication is a consequence of the cross-Spark type split.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: Intentional NoSuchTableException here. wrapAsView is reached from loadTable (table-resolution path), where Spark's resolver expects Table / NoSuchTableException. The view-resolution path (loadView) uses NoSuchViewException. The existing scaladoc on wrapAsView explains this; I can expand it if it'd help.

connectors/spark/src/test/java/io/unitycatalog/spark/UCProxySuite.java
chenwang-databricks: OK I split it:
  - UCProxySuite (moved to src/test/java/) now holds the 10 version-agnostic tests — listTables pagination ×5, listNamespaces pagination ×3, the single-getTable-RPC guard, and the DROP TABLE-on-metric-view safety check. These compile and run on 4.0, 4.1, and 4.2.
  - UCViewProxySuite (stays under src/test/scala-shims/spark-4.2/) keeps the 9 view-only tests (createView/loadView/dropView/loadTableOrView, the V2-TableInfo createTable overload, MetadataTable/ViewInfo assertions) — those cast the proxy to TableViewCatalog/ViewCatalog, which only succeeds on the 4.2 build.
  - The reflective UCProxy construction is shared via a new non-test UCProxyTestFixture (composition, not inheritance — so JUnit doesn't double-run the shared tests on 4.2).

connectors/spark/src/main/scala-shims/spark-4.0/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,23 @@
+/*
```
yili-db: Code files under this repo do not seem to carry a copyright header.
Also this header appears to be incomplete. Maybe simply remove it. The same with other new files in this PR.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,306 @@
...
+    } catch {
+      case _: NoSuchTableException => throw new NoSuchViewException(ident)
+    }
+    // Gate on the command-supported predicate (excludes listed-but-unmapped kinds like
+    // MATERIALIZED_VIEW). A kind we list but have no Spark `TableSummary` mapping for cannot be
+    // turned into a `ViewInfo` -- `toViewInfo -> ucTableTypeToSparkViewType` would throw
+    // `UnsupportedOperationException`, which `DESCRIBE`/`SHOW`/view-resolution callers don't
+    // understand. Surface `NoSuchViewException` instead, mirroring `wrapAsView`.
+    if (!UCSingleCatalog.isViewCommandsSupportedTableType(t.getTableType)) {
+      throw new NoSuchViewException(ident)
```
yili-db, line 93: These views appear in `listViews` but when you load them it says `NoSuchViewException`. That is rather confusing. `UnsupportedOperationException` seems better in this case.

chenwang-databricks: Good point -- changed `loadView` to throw `UnsupportedOperationException` for a listed-but-unmapped view kind (e.g. `MATERIALIZED_VIEW`), so it is consistent with the kind appearing in `SHOW VIEWS`. A genuine non-view (a real table) at the identifier still throws `NoSuchViewException`, and `wrapAsView` on the `loadTable` table-resolution path still throws `NoSuchTableException` (Spark's table resolver only understands that). Added a unit test for the new path. (41c0e6d4)

connectors/spark/src/main/scala-shims/spark-4.0/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: Removed -- dropped the incomplete Apache header from all 8 new `scala-shims/` files to match the repo convention (existing source files here carry no copyright header). The other new files in this PR (`UCColumnJson.scala` and the test files) never had one. (41c0e6d4)

connectors/spark/src/main/scala/io/unitycatalog/spark/UCColumnJson.scala
```diff
@@ -0,0 +1,100 @@
...
+   * The `"comment"` key (if present in the wire metadata) is lifted out and passed to
+   * [[Column.create]] as the dedicated `comment` arg, NOT left in `metadataInJSON` --
+   * matches what Spark's own `CatalogV2Util.structFieldToV2Column` does on the v1 path.
+   */
+  def parseStructFieldJson(jsonStr: String): Column = {
+    // `ColumnInfo.type_json` is nullable on the wire (e.g. a view-like row created by an older
+    // writer or a non-Spark client that did not round-trip the field). Fail with a clear,
+    // actionable error rather than letting `json4s.parse(null)` throw a raw NPE out of the
+    // view-load path.
+    if (jsonStr == null) {
```
yili-db, line 68: [AI Generated] This null-`type_json` guard (added as review hardening) has no test. Suggest a unit test asserting `parseStructFieldJson(null)` throws `IllegalArgumentException` with this exact message. The two sibling guards below are also untested individually: the non-string `name` case (L75) and the non-boolean `nullable` case (L82). A small dedicated `UCColumnJson` suite (happy-path round-trip + these three malformed inputs) would lock the wire-format contract that reviewers otherwise have to reason about by hand.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,306 @@
...
+   * Builds a Spark `ViewInfo` from a UC `UCTableInfo` for view-like rows. Round-trips
+   * dataType / nullable / comment / metadata via the in-connector [[UCColumnJson]]
+   * helper.
+   */
+  private[spark] def toViewInfo(t: UCTableInfo): ViewInfo = {
+    // Guard the columns collection itself (nullable on the wire for a row written by an older or
+    // non-Spark client), mirroring the `Option(t.getProperties)` handling below and the per-column
+    // `type_json` guard in `parseStructFieldJson` -- so a missing list degrades to an empty schema
+    // rather than NPE-ing out of `loadView` / `loadTable`.
+    val columns = Option(t.getColumns).map(_.asScala).getOrElse(Seq.empty).map { col =>
```
yili-db, line 196: [AI Generated] The null-`getColumns` guard (degrade to an empty schema) is untested. Every `UCViewProxySuite` fixture sets non-null columns, so this branch never executes. Suggest a `loadView`/`toViewInfo` test with a `TableInfo` whose `columns` is null, asserting the returned `ViewInfo` has an empty schema rather than NPE-ing out of the view-load path.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,306 @@
...
+    try {
+      tablesApi.createTable(ct)
+    } catch {
+      case e: ApiException if e.getCode == 409 =>
+        throw new ViewAlreadyExistsException(ident)
+    }
+    loadView(ident)
+  }
+
+  override def replaceView(ident: Identifier, info: ViewInfo): ViewInfo = {
```
yili-db, line 149: [AI Generated] `replaceView` throwing `UnsupportedOperationException` is untested. A one-line `assertThatThrownBy(() -> proxyViews.replaceView(ident, info)).isInstanceOf(UnsupportedOperationException.class)` in `UCViewProxySuite` would cover it.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,306 @@
...
+    if (!UCSingleCatalog.isViewCommandsSupportedTableType(t.getTableType)) {
+      return false
+    }
+    // `deleteTable` returns the (empty) response body, not an HTTP status; a real failure throws
+    // ApiException. Issue the delete for its side effect and report success.
+    tablesApi.deleteTable(UCSingleCatalog.fullTableNameForApi(this.name, ident))
+    true
+  }
+
+  override def renameView(oldIdent: Identifier, newIdent: Identifier): Unit = {
```
yili-db, line 174: [AI Generated] Same gap as `replaceView`: `renameView`'s `UnsupportedOperationException` is untested. A one-line assertion in `UCViewProxySuite` covers it.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,306 @@
...
+    } catch {
+      case _: NoSuchTableException => return false
+    }
+    // Known limitation: a listed-but-unmapped view kind (e.g. MATERIALIZED_VIEW, `None` in
+    // `viewLikeUcTypes`) is rejected here AND by `UCProxy.dropTable` (which passive-filters all
+    // `isViewLikeTableType` rows), so such a row is not droppable through this connector and a
+    // `DROP ... IF EXISTS` no-ops. This is intentional: the connector treats unsupported view
+    // kinds as uniformly inert (no load / create / drop) rather than allowing a drop of a kind it
+    // can't otherwise handle. Revisit when MATERIALIZED_VIEW becomes a supported (mapped) kind.
+    if (!UCSingleCatalog.isViewCommandsSupportedTableType(t.getTableType)) {
```
yili-db, line 165: [AI Generated] `UCViewProxySuite.testDropViewReturnsFalseForRegularTable` covers the real-table branch, but this listed-but-unmapped branch (e.g. `MATERIALIZED_VIEW` -> `false`) is untested, and it's the subtler one. Add a `dropView` test with a `MATERIALIZED_VIEW` `TableInfo` asserting it returns `false` and that `deleteTable` is never called, mirroring the existing regular-table test.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,306 @@
...
+   * matching the wire format already used by the UC server.
+   */
+  private def toUcDependencyList(sparkDeps: DependencyList): UCDependencyList = {
+    val ucDeps = new java.util.ArrayList[UCDependency]()
+    sparkDeps.dependencies().foreach {
+      case td: TableDependency =>
+        ucDeps.add(new UCDependency()
+          .table(new UCTableDependency()
+            .tableFullName(td.nameParts().mkString("."))))
+      case _ =>
```
yili-db, line 238: [AI Generated] The silent drop of non-`TableDependency` (function) deps here is untested. Worth a `createView` test with a `ViewInfo` carrying a function dependency, asserting the resulting `CreateTable` payload contains only the table dependencies, so the lossy drop is intentional and pinned rather than incidental.

yili-db (commented):
[AI Generated] Reviewed against the team code-review guidelines. The implementation looks solid; the one consistent gap I see is test coverage for the error/robustness paths added during review. Flagging each one inline next to the code so it's actionable. None are blocking on their own, but together they leave the new failure-handling branches unverified (guideline: "verify that it's failing as expected with the correct error message").

chenwang-databricks:
@yili-db thanks — good catch on the untested review-hardening branches. Added coverage in ea056ecf:

**New `UCColumnJsonSuite`** (shared `src/test/java/`):
- `parseStructFieldJson(null)` → `IllegalArgumentException` ("type_json is missing")
- non-string `name` → `IllegalArgumentException`
- non-boolean `nullable` → `IllegalArgumentException`
- happy-path round-trip, to pin the wire-format contract

**`UCViewProxySuite`** (+5):
- `loadView` with null `columns` degrades to an empty schema (no NPE)
- `replaceView` / `renameView` throw `UnsupportedOperationException`
- `dropView` on a listed-but-unmapped kind (`MATERIALIZED_VIEW`) returns `false` and never calls `deleteTable`
- `createView` with a function dependency drops it from the `CreateTable` payload, keeping only table deps

`UCColumnJsonSuite` passes locally (4/4); `UCViewProxySuite` is Spark-4.2-only and verified via CI.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -942,3 +1105,4 @@ private class UCProxy(
     true
   }
 }
+
```
yili-db, line 1108: Is this an extra empty line?

yili-db (approved):

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,306 @@
...
+   * otherwise throw.
+   */
+  protected def wrapAsView(t: UCTableInfo, ident: Identifier): Table =
+    if (UCSingleCatalog.isViewCommandsSupportedTableType(t.getTableType)) {
+      new MetadataTable(toViewInfo(t), ident.toString)
+    } else {
+      throw new NoSuchTableException(ident)
+    }
+
+  override def listViews(namespace: Array[String]): Array[Identifier] = {
```
openinx, line 72: this method is overidding the interface from where ? I didn't find the original interface definition in the code.

openinx: okay, seems like the `TableViewCatalog` is defined in the spark 4.2.  Could you pls share me the code pointer where defines the `TableViewCatalog` ?  I searched the spark code, and seems like failed to get the pointer.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,306 @@
...
+
+    val builder = new ViewInfo.Builder()
+      .withColumns(columns)
+      .withProperties(props)
+      .withTableType(UCSingleCatalog.ucTableTypeToSparkViewType(t.getTableType))
+      .withQueryText(t.getViewDefinition)
+      .withCurrentCatalog(t.getCatalogName)
+      .withCurrentNamespace(Array(t.getSchemaName))
+      .withSqlConfigs(sqlConfigs)
+      .withSchemaMode("UNSUPPORTED")
```
openinx, line 216: what is this field used for ?

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * Converts Spark's typed `DependencyList` into the wire-format UC `DependencyList`.
+   * Only `TableDependency` is currently translated; UC OSS does not persist function
+   * dependencies. The conversion is dot-flatten lossy for identifiers with literal `.`,
+   * matching the wire format already used by the UC server.
+   */
+  private def toUcDependencyList(sparkDeps: DependencyList): UCDependencyList = {
+    val ucDeps = new java.util.ArrayList[UCDependency]()
+    sparkDeps.dependencies().foreach {
```
openinx, line 233: what are those dependencies used for, usually ?

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,306 @@
...
+    }
+  }
+
+  private def extractPartitionColumnNames(partitions: Array[Transform]): Seq[String] = {
+    if (partitions == null) return Seq.empty
+    partitions.flatMap { t =>
+      t.name() match {
+        case "identity" =>
+          val fieldNames = t.references().flatMap(_.fieldNames())
+          require(fieldNames.length == 1,
```
openinx, line 296: why we are limited to support only one field of partition columns ? I didn't get....

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,306 @@
...
+            s"Expected single-field partition reference but got: ${fieldNames.mkString(".")}")
+          Some(fieldNames.head)
+        case "cluster_by" =>
+          None
```
openinx, line 300: why `None` ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -83,6 +106,7 @@ class UCSingleCatalog
     val proxy = new UCProxy(uri, tokenProvider, renewCredEnabled, credScopedFsEnabled,
       serverSidePlanningEnabled, apiClient, tablesApi)
     proxy.initialize(name, options)
+    ucProxy = proxy
```
openinx, line 109: nit: why not just remove the redundant `proxy` local variable, and use the `ucProxy` always directly ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -137,7 +161,15 @@ class UCSingleCatalog
 
   override def listTables(namespace: Array[String]): Array[Identifier] = delegate.listTables(namespace)
 
-  override def loadTable(ident: Identifier): Table = delegate.loadTable(ident)
+  // The view-rejection check (`MetadataTable + ViewInfo` flowing back through Delta's
+  // `case o => o` fallthrough on top of `UCProxy.loadTable`) lives in the per-Spark-
+  // version `ViewSupport.rejectIfShimmedView` hook -- a no-op on Spark 4.0/4.1
+  // (no view types exist there) and the real check on Spark 4.2.
```
openinx, line 167: nit: This comments seems to be un-readable.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -600,17 +632,92 @@ object UCSingleCatalog {
     checkUnsupportedNestedNamespace(ident.namespace())
     Seq(catalogName, ident.namespace()(0), ident.name()).mkString(".")
   }
+
+  /**
+   * Single source of truth for which UC `TableType` values the connector treats as
```
openinx, line 637: Pls don't use the markdown comment in the java/scala code, instead, we suggest to use the javadoc or scala doc. that's more easier to read in the code.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCColumnJson.scala
```diff
@@ -0,0 +1,100 @@
...
+ * `"comment"` key at the JSON boundary so callers never see it duplicated.
+ *
+ * Uses only public V2 APIs that exist on all supported Spark versions, so this lives in the
+ * shared source tree (not a `scala-shims/` dir) even though its only caller today is the
+ * Spark-4.2 `UCProxyViewSupport` view shim.
+ */
+private[spark] object UCColumnJson {
+
+  /** Spark V2 -> wire: build the `StructField`-shape JSON for one [[Column]]. */
+  def buildStructFieldJson(col: Column): String = {
```
openinx, line 37: we already have the `toStructFieldJson` method for the normal table, how is this different with that method ?  https://github.com/unitycatalog/unitycatalog/blob/14e13707c4d3652d9b17513fee70a666c5521090/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala#L939-L945

Should we consider to unify them into one ?

openinx (commented):
Hi @chenwang-databricks,  I skimmed this PR. and this seems like a huge PR which involves lots of context and changes. 

will you be able to split this into several smaller ones, so that we can very focus on each indivisual changes and iterate this feature ? 


After skimmed the PR, the work that can be split into at least 5 PRs in my mind: 
1.  unify the `toStructFieldJson` for both normal table and views . 
2.  introduce the shim classes. 
3. glue the shim classes into the UCSingleCatalog, to support view operations. 
4. the e2e local-run tests. 
5. integration tests run-against the remote uc server.

PR #1555, opened 2026-05-14
New experimental flag: server.managed-table.use-delta-api-only (default off)

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1555/files) to review incremental changes.
- [**stack/phaseout**](https://github.com/unitycatalog/unitycatalog/pull/1555) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1555/files)]

---------
**PR Checklist**

**Description of changes**
New experimental flag: server.managed-table.use-delta-api-only (default off).

When on, UC endpoints that have a Delta equivalent reject MANAGED tables so clients are forced through the Delta surface. Wired at following entry points and with the gate shared on ServerProperties.checkDeltaApiOnlyForManagedTable and checkDeltaApiOnlyEnabled:
- StagingTableService.createStagingTable (always blocked)
- TableService.createTable (request-typed)
- TemporaryTableCredentialsService.generateTemporaryTableCredential (request-typed)
- DeltaCommitsService.getCommits (always blocked)
- DeltaCommitsService.postCommit (always blocked)

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -76,32 +76,45 @@ public class TableRepository {
   }
 
   /**
-   * Retrieves the storage location for a table or staging table by its ID. First attempts to find a
-   * regular table with the given ID, then falls back to searching for a staging table if no regular
-   * table is found. NOTE: This function is specially needed by generateTemporaryTableCredential
-   * during the short window when a staging table is just created and the initial data is being
-   * written but before the actual table is already created. Reading of a staging table is not a
-   * common supplemental of an actual table but only a special case.
+   * Returned by {@link #getStorageLocationForTableOrStagingTable} so callers (today the temp-creds
+   * service) can vend and gate from a single DB read. Staging rows project as MANAGED because they
+   * always finalize to MANAGED Delta.
+   */
+  public record TableStorageLocationInfo(NormalizedURL url, TableType tableType) {}
+
+  /**
+   * Retrieves the storage location for a table or staging table by its ID, plus the table type.
+   * Staging tables are counted as Managed tables too. First attempts to find a regular table with
+   * the given ID, then falls back to searching for a staging table if no regular table is found.
+   * NOTE: This function is specially needed by generateTemporaryTableCredential during the short
+   * window when a staging table is just created and the initial data is being written but before
+   * the actual table is already created. Reading of a staging table is not a common supplemental of
+   * an actual table but only a special case.
    *
    * @param tableId the ID of the table or staging table
-   * @return the normalized URL of the storage location
+   * @return the normalized URL of the storage location, plus table type. Both in
+   *     TableStorageLocationInfo
    * @throws BaseException with ErrorCode.TABLE_NOT_FOUND if neither a table nor staging table is
    *     found with the given ID
    */
-  public NormalizedURL getStorageLocationForTableOrStagingTable(UUID tableId) {
+  public TableStorageLocationInfo getStorageLocationForTableOrStagingTable(UUID tableId) {
     return TransactionManager.executeWithTransaction(
         sessionFactory,
         session -> {
           LOGGER.debug("Getting storage location of table by id: {}", tableId);
           TableInfoDAO tableInfoDAO = session.get(TableInfoDAO.class, tableId);
           if (tableInfoDAO != null) {
-            return NormalizedURL.from(tableInfoDAO.getUrl());
+            return new TableStorageLocationInfo(
+                NormalizedURL.from(tableInfoDAO.getUrl()),
+                TableType.fromValue(tableInfoDAO.getType()));
           }
 
           LOGGER.debug("Getting storage location of staging table by id: {}", tableId);
           StagingTableDAO stagingTableDAO = session.get(StagingTableDAO.class, tableId);
           if (stagingTableDAO != null) {
-            return NormalizedURL.from(stagingTableDAO.getStagingLocation());
+            // Staging rows always become MANAGED Delta on finalize, so project them as MANAGED.
+            return new TableStorageLocationInfo(
+                NormalizedURL.from(stagingTableDAO.getStagingLocation()), TableType.MANAGED);
           }
           throw new BaseException(
               ErrorCode.TABLE_NOT_FOUND,
--- a/server/src/main/java/io/unitycatalog/server/service/DeltaCommitsService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/DeltaCommitsService.java
@@ -43,6 +43,10 @@ public class DeltaCommitsService extends AuthorizedService {
   public HttpResponse postCommit(
       @AuthorizeResourceKey(value = TABLE, key = "table_id")
       DeltaCommit commit) {
+    // The UC REST commit endpoint only accepts MANAGED Delta tables (enforced downstream by
+    // validateTableForCommit), so the gate applies unconditionally here.
+    serverProperties.checkDeltaApiOnlyEnabled(
+        "POST /delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table} with action add-commit");
     deltaCommitRepository.postCommit(commit);
     return HttpResponse.of(HttpStatus.OK);
   }
@@ -57,6 +61,9 @@ public class DeltaCommitsService extends AuthorizedService {
   public HttpResponse getCommits(
     @AuthorizeResourceKey(value = TABLE, key = "table_id")
     DeltaGetCommits rpc) {
+    // Same as postCommit above -- managed-Delta-only by contract; gate unconditionally.
+    serverProperties.checkDeltaApiOnlyEnabled(
+        "GET /delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table} (commits[])");
     return HttpResponse.ofJson(deltaCommitRepository.getCommits(rpc));
   }
 }
--- a/server/src/main/java/io/unitycatalog/server/service/TableService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/TableService.java
@@ -87,6 +87,9 @@ public class TableService extends AuthorizedService {
       @AuthorizeKey(key = "table_type")
       CreateTable createTable) {
     assert createTable != null;
+    serverProperties.checkDeltaApiOnlyForManagedTable(
+        createTable.getTableType(),
+        "POST /delta/v1/catalogs/{catalog}/schemas/{schema}/tables");
     TableInfo tableInfo = tableRepository.createTable(createTable);
 
     SchemaInfo schemaInfo =
--- a/server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
+++ b/server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
@@ -2,6 +2,7 @@ package io.unitycatalog.server.utils;
 
 import io.unitycatalog.server.exception.BaseException;
 import io.unitycatalog.server.exception.ErrorCode;
+import io.unitycatalog.server.model.TableType;
 import io.unitycatalog.server.service.credential.aws.S3StorageConfig;
 import io.unitycatalog.server.service.credential.azure.ADLSStorageConfig;
 import io.unitycatalog.server.service.credential.gcp.GcsStorageConfig;
@@ -197,6 +198,8 @@ public class ServerProperties {
     REDIRECT_PORT("server.redirect-port", POSITIVE_INTEGER_VALIDATOR),
     COOKIE_TIMEOUT("server.cookie-timeout", "P5D", DURATION_VALIDATOR),
     MANAGED_TABLE_ENABLED("server.managed-table.enabled", "false", BOOLEAN_VALIDATOR),
+    MANAGED_TABLE_USE_DELTA_API_ONLY(
+        "server.managed-table.use-delta-api-only", "false", BOOLEAN_VALIDATOR),
     // `storage-root.*` are replaced by managed storage locations of catalog and schema.
     MODEL_STORAGE_ROOT("storage-root.models", STORAGE_PATH_VALIDATOR), // Deprecated
     TABLE_STORAGE_ROOT("storage-root.tables", STORAGE_PATH_VALIDATOR), // Deprecated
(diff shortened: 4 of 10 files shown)


server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -453,6 +456,27 @@ public void checkManagedTableEnabled() {
...
 
+  /**
+   * Reject the UC request when MANAGED_TABLES_USE_DELTA_API_ONLY is on and the targeted table is
+   * MANAGED. Call from UC endpoints whose Delta equivalent should be used instead.
+   *
+   * @param tableType the table type to check; only {@link TableType#MANAGED} triggers the gate.
+   * @param deltaEndpoint full Delta endpoint to suggest in the error, e.g. {@code "POST
+   *     /delta/v1/catalogs/{catalog}/schemas/{schema}/staging-tables"}.
+   */
+  public void checkNotManagedDeltaUcEndpoint(TableType tableType, String deltaEndpoint) {
```
tdas: [AI Assisted]

**S2: `checkNotManagedDeltaUcEndpoint` is hard to parse.**

The name reads as "check (that this is) not (a) managed-delta UC endpoint", but the method actually means "if the use-delta-api-only flag is on AND the table is MANAGED, reject this request and tell the caller to use the supplied Delta endpoint instead." The "Delta" in the name refers to the *redirect target* while "Managed" refers to the *table type* — two unrelated concepts compressed into one camel-case run.

Consider e.g. `rejectManagedTableIfDeltaApiOnly(TableType, alternativeDeltaEndpoint)` or `requireDeltaApiForManagedTable(...)`. Whatever you pick, the verb should make clear it's an unconditional throw, not a boolean query.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -197,6 +198,8 @@ public enum Property {
     REDIRECT_PORT("server.redirect-port", POSITIVE_INTEGER_VALIDATOR),
     COOKIE_TIMEOUT("server.cookie-timeout", "P5D", DURATION_VALIDATOR),
     MANAGED_TABLE_ENABLED("server.managed-table.enabled", "false", BOOLEAN_VALIDATOR),
+    MANAGED_TABLES_USE_DELTA_API_ONLY(
```
tdas: [AI Assisted]

**S3: Enum constant name is inconsistent with the property key and the sibling enum.**

The property key is singular `server.managed-table.use-delta-api-only` and the sibling enum a line above (`MANAGED_TABLE_ENABLED` → `server.managed-table.enabled`) is singular `MANAGED_TABLE_...`, but this new constant is plural `MANAGED_TABLES_...`.

Suggest renaming to `MANAGED_TABLE_USE_DELTA_API_ONLY` so the enum, the key, and the sibling all agree.

server/src/main/java/io/unitycatalog/server/service/DeltaCommitsService.java
```diff
@@ -43,6 +44,11 @@ public DeltaCommitsService(
   public HttpResponse postCommit(
       @AuthorizeResourceKey(value = TABLE, key = "table_id")
       DeltaCommit commit) {
+    // The UC REST commit endpoint only accepts MANAGED Delta tables (enforced downstream by
+    // validateTableForCommit), so the gate applies unconditionally here.
+    serverProperties.checkNotManagedDeltaUcEndpoint(
```
tdas: [AI Assisted]

**S4: `TableType` parameter is dead at three of the five callsites.**

Here, in `getCommits` below, and in `StagingTableService.createStagingTable`, you pass the literal `TableType.MANAGED` along with a comment like *"gate applies unconditionally here"*. The `TableType` argument has no semantic content at these sites — the code would be clearer as two methods:

- `checkDeltaApiOnlyEnabled(deltaEndpoint)` — for endpoints that are MANAGED-only by contract (staging-create, delta-commits)
- `checkDeltaApiOnlyForManagedTable(tableType, deltaEndpoint)` — for request-typed endpoints (createTable, temp-creds)

This removes the need for the "always pass MANAGED" explanatory comments and makes the gating intent legible from the callsite alone.

server/src/main/java/io/unitycatalog/server/service/DeltaCommitsService.java
```diff
@@ -57,6 +63,10 @@ public HttpResponse postCommit(
   public HttpResponse getCommits(
     @AuthorizeResourceKey(value = TABLE, key = "table_id")
     DeltaGetCommits rpc) {
+    // Same as postCommit above -- managed-Delta-only by contract; gate unconditionally.
+    serverProperties.checkNotManagedDeltaUcEndpoint(
```
tdas: [AI Assisted]

**S6: EXTERNAL table on `postCommit`/`getCommits` produces a misleading error when the flag is on.**

Because you pass `TableType.MANAGED` literally, the gate fires for any `tableId` — including EXTERNAL — when the flag is on. An EXTERNAL caller gets "Use the Delta endpoint ... instead", but that Delta endpoint also doesn't accept EXTERNAL tables. With the flag *off* the downstream `validateTableForCommit` would give them a correct EXTERNAL-rejected message.

Low impact (these endpoints are MANAGED-Delta-only by contract so the EXTERNAL caller is already in error territory), but worth deciding: should the gate check the *actual* table type here instead of unconditionally rejecting, so the error message stays accurate? Posting as a question rather than a finding.

server/src/main/java/io/unitycatalog/server/service/TableService.java
```diff
@@ -87,6 +87,9 @@ public HttpResponse createTable(
       @AuthorizeKey(key = "table_type")
       CreateTable createTable) {
     assert createTable != null;
+    serverProperties.checkNotManagedDeltaUcEndpoint(
```
tdas: [AI Assisted]

**S5: Asymmetry between `createTable` (gated) and `getTable` (not gated) — please clarify intent.**

This PR gates `createTable` for MANAGED but leaves `getTable` (line 105) un-gated. The test explicitly notes "UC getTable on the MANAGED Delta table -- allowed for now", and the commit message originally claimed `getTable` *was* gated, which suggests the design intent shifted mid-PR.

If the goal is to force MANAGED clients onto the Delta surface, why is the metadata-read path still allowed while the credential-read path (`TemporaryTableCredentialsService`) is blocked? Either:

- it's intentional (some clients legitimately need UC `getTable` for metadata even when DML must go through Delta) — please add a one-line comment to `getTable` explaining the deliberate carve-out, and update the PR description to call this out, or
- it's an oversight — wire it up like the others.

The "for now" wording in the test comment leaves this ambiguous. Posting as a question.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -76,32 +76,45 @@ public TableRepository(
...
-   * regular table with the given ID, then falls back to searching for a staging table if no regular
-   * table is found. NOTE: This function is specially needed by generateTemporaryTableCredential
-   * during the short window when a staging table is just created and the initial data is being
-   * written but before the actual table is already created. Reading of a staging table is not a
-   * common supplemental of an actual table but only a special case.
+   * Returned by {@link #getStorageLocationForTableOrStagingTable} so callers (today the temp-creds
+   * service) can vend and gate from a single DB read. Staging rows project as MANAGED because they
+   * always finalize to MANAGED Delta.
+   */
+  public record TableStorageLocationInfo(NormalizedURL url, TableType tableType) {}
```
tdas, line 83: [AI Assisted]

**S7: `TableStorageLocationInfo` is exposed publicly even though there's exactly one caller.**

The record is `public` on a `public` class, but `getStorageLocationForTableOrStagingTable` has a single in-repo caller (`TemporaryTableCredentialsService`). Confirmed via repo-wide grep that nothing else uses this method or type.

If you don't expect external server-extension callers, consider package-private to keep the surface area tight — per code-review rule 11, public APIs are sticky.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -453,6 +456,27 @@ public void checkManagedTableEnabled() {
...
+  /**
+   * Reject the UC request when MANAGED_TABLES_USE_DELTA_API_ONLY is on and the targeted table is
+   * MANAGED. Call from UC endpoints whose Delta equivalent should be used instead.
+   *
+   * @param tableType the table type to check; only {@link TableType#MANAGED} triggers the gate.
+   * @param deltaEndpoint full Delta endpoint to suggest in the error, e.g. {@code "POST
+   *     /delta/v1/catalogs/{catalog}/schemas/{schema}/staging-tables"}.
+   */
+  public void checkNotManagedDeltaUcEndpoint(TableType tableType, String deltaEndpoint) {
+    if (isTrueOrEnable(get(Property.MANAGED_TABLES_USE_DELTA_API_ONLY))
```
tdas: [AI Assisted]

**TG1: No direct unit test for `checkNotManagedDeltaUcEndpoint`.**

The four branches (flag on/off × MANAGED/non-MANAGED) are only exercised transitively via the E2E test, which only runs with flag=ON. A small unit test on `ServerProperties` would give explicit branch coverage without the cost of the SDK harness. Not a hard blocker since the gate is small, but worth a few lines.

tdas (commented):
[AI Assisted]

Left suggestions and testing-gap comments inline. One PR-level item that doesn't bind to a specific line:

**S1: Commit message is stale and contradicts the code.** The commit message states the gate is wired at **four** entry points and lists `TableService.getTable (post-load)` as one of them. Actual diff: `TableService.getTable` is **not** touched (the test at step 5 explicitly says "UC getTable on the MANAGED Delta table -- allowed for now"), and two endpoints that *are* gated — `DeltaCommitsService.postCommit` and `DeltaCommitsService.getCommits` — are not mentioned. The PR description on GitHub lists the correct five entry points, but the commit message lives forever in git history. Please rewrite the commit message to match the actual five gated endpoints, and clarify whether `getTable` being un-gated is intentional (see the inline comment on `TableService.createTable`).

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -453,6 +456,27 @@ public void checkManagedTableEnabled() {
...
+        && tableType == TableType.MANAGED) {
```
tdas: [AI Assisted]

**TG2: `METRIC_VIEW` table type behavior is not exercised.**

`TableRepository.java:866` shows `METRIC_VIEW` is a known table type. The gate is `tableType == MANAGED`, so `METRIC_VIEW` passes through. That's likely correct, but worth a one-line test (or at least a code comment on this conditional) confirming that any non-MANAGED type — including `METRIC_VIEW` — is intentionally not gated.

server/src/test/java/io/unitycatalog/server/sdk/delta/SdkUseDeltaApiOnlyFlagTest.java
```diff
@@ -0,0 +1,172 @@
...
+  public void setUp() {
+    super.setUp();
+    ApiClient apiClient = TestUtils.createApiClient(serverConfig);
+    ucTablesApi = new TablesApi(apiClient);
+    ucCredsApi = new TemporaryCredentialsApi(apiClient);
+    ucCommitsApi = new DeltaCommitsApi(apiClient);
+  }
+
+  @Test
+  public void testManagedTablesUseDeltaApiOnlyFlag() throws ApiException {
```
tdas, line 85: [AI Assisted]

**TG3: Flag-OFF path is not asserted in this test class.**

The PR description argues it's covered by "every other UC REST CRUD / commits test in this module". That's true for the non-gated endpoints, but it does *not* prove that *with this PR applied*, the gated endpoints still work when the flag is off (i.e. that nothing accidentally fires when the flag is unset).

One smoke test toggling the flag off and exercising the five gated endpoints would close that loop and give explicit conditional-branch coverage (per code-review rule 2g).

tdas (approved):
No blockers, so putting a high level approval. @yili-db please address the suggestions in this PR or follow up PRs.

server/src/test/java/io/unitycatalog/server/sdk/delta/SdkUseDeltaApiOnlyFlagTest.java
yili-db: All the gated end points are commonly exercised in all CRUD tests. They continue to pass even with this PR adding this new flag and with the this flag turned off by default. They need no change in this PR to continue to work but they are covering the "Flag-OFF" path by default already.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: This can't be done in Java. It is defined in `io.unitycatalog.server.persist` and used in `io.unitycatalog.server.service`. Unlike Scala, Java doesn't support limiting it to parent level `io.unitycatalog.server`.

server/src/main/java/io/unitycatalog/server/service/TableService.java
yili-db: Yes this is intentional. Added comments in tests:
```
    // 5. UC getTable on the MANAGED Delta table -- allowed for now as non-Delta clients and
    // older clients supporting only EXTERNAL tables still needs to gracefully recognize MANAGED
    // Delta tables by loading them.
```

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
yili-db: Renamed to `checkDeltaApiOnlyForManagedTable` and `checkDeltaApiOnlyEnabled` per the other comment

yili-db:
> [AI Assisted]
> 
> Left suggestions and testing-gap comments inline. One PR-level item that doesn't bind to a specific line:
> 
> **S1: Commit message is stale and contradicts the code.** The commit message states the gate is wired at **four** entry points and lists `TableService.getTable (post-load)` as one of them. Actual diff: `TableService.getTable` is **not** touched (the test at step 5 explicitly says "UC getTable on the MANAGED Delta table -- allowed for now"), and two endpoints that _are_ gated — `DeltaCommitsService.postCommit` and `DeltaCommitsService.getCommits` — are not mentioned. The PR description on GitHub lists the correct five entry points, but the commit message lives forever in git history. Please rewrite the commit message to match the actual five gated endpoints, and clarify whether `getTable` being un-gated is intentional (see the inline comment on `TableService.createTable`).

PR message correctly lists all 5 endpoints.

PR #1556, opened 2026-05-20
[Spark] Cross-build connector per Spark major.minor version

seewishnew:
## Summary

Add the SBT plumbing to publish the UC Spark connector as three separately-named artifacts — `unitycatalog-spark_4.0_2.13`, `unitycatalog-spark_4.1_2.13`, `unitycatalog-spark_4.2_2.13` — one per supported Spark major.minor. Mirrors Delta Lake's `CrossSparkVersions.scala` pattern.

This PR is purely build/release infra; the connector source code stays table-only at this stage. A follow-up PR will add compile-time shims inside the per-version `scala-shims/spark-X.Y/` directories.

## CI Flow

Each cell in the CI matrix follows the same ordered pipeline:

```mermaid
graph TD
    A["Checkout UC PR"] --> B["publishLocal + publishM2<br/>(with -DskipDeltaSpark=true<br/>for Delta-master cells)"]
    B --> C{"Delta version?"}
    C -->|"Released (4.1.0)"| E["Set DELTA_VERSION=4.1.0"]
    C -->|"master-SNAPSHOT"| D["build_delta_from_source.sh<br/>clone delta-io/delta master<br/>publishM2 with<br/>-DunityCatalogVersion=0.5.0-SNAPSHOT"]
    D --> E
    E --> F["integrationTests/test:compile"]
    F --> G["spark/jacoco or +jacoco"]
```

**Key design points:**
- UC jars are published to both ivy2 (`publishLocal`) and M2 (`publishM2`) before Delta builds. Delta resolves UC from ivy2 (its resolver chain lacks `maven-local`).
- Delta publishes to M2 only (not `publishLocal`). Delta's ivy.xml lists internal modules (`delta-spark-v1`, `delta-spark-v2`) that aren't published separately; the M2 POM correctly filters them via `pomPostProcess`. UC resolves Delta from M2 (its resolver chain includes `maven-local`).
- `-DunityCatalogVersion` overrides Delta's pinned UC coordinate, making `ensurePinnedUnityCatalog` a no-op.

## CI Matrix

| Spark | Delta | Mode | Blocking |
|-------|-------|------|----------|
| 4.0.0 | 4.1.0 | all-tests | yes |
| 4.1.0 | 4.1.0 | spark-tests | yes |
| 4.1.0 | master-SNAPSHOT | spark-tests | no |
| 4.2.0-SNAPSHOT | master-SNAPSHOT | spark-tests | no |

All 4 cells pass on the latest commit.

### `project/CrossSparkVersions.scala` (new)
- `SparkVersionSpec` case class — version metadata loaded from `project/spark-versions.json`
- `sparkVersionedModuleName`: appends `_X.Y` suffix; `-DskipSparkSuffix=true` suppresses (backward compat)
- `sparkDependentSettings`: wires settingKey + per-version shim source dirs

### `project/spark-versions.json` (new)
Single source of truth for supported Spark versions. Read by both SBT and the Python publish test.

### `project/scripts/build_delta_from_source.sh` (new)
Clones Delta master, reads UC version from `version.sbt`, builds Delta against the locally-published UC jars. Supports `--meta-only` for version discovery without building.

### `project/tests/test_cross_spark_publish.py` + `.github/workflows/cross-spark-build.yml` (new)
Validates JAR file names: default publish (with suffix), backward-compat (no suffix), per-version, and full cross-version workflow.

### `build.sbt` changes
- `sparkVersion`/`sparkMajorMinorVersion` derived from `CrossSparkVersions.getSparkVersionSpec()`
- `deltaSparkTestDeps`: conditional on `-DskipDeltaSpark=true` to break the circular dep during publish-only paths
- `spark` module: `Keys.moduleName` with versioned suffix, `sparkDependentSettings`

### `.github/workflows/unit-tests.yml`
- Pre-Delta `publishLocal + publishM2` step with `-DskipDeltaSpark=true`
- Delta-master cells build Delta from source via `build_delta_from_source.sh`
- Removed `setup_unitycatalog_main.sh` indirection (UC publishes its own jars first)

### `UCSingleCatalog.scala` — SPARK-55250 compat fix
Spark 4.2 removed the `namespaceExists()` pre-check from `CreateNamespaceExec`. UC now catches `NamespaceAlreadyExistsException` for IF NOT EXISTS semantics.

### Empty `scala-shims/spark-{4.0,4.1,4.2}/` directories
Placeholder structure for per-version shim code (main + test).

## Dependency

Builds on top of #1518 (merged). Apache Snapshots resolver + SparkCatalogCompatibility shim.

## Relationship to #1553

Supersedes #1553 (Chen's PoC). Credit to @chenwang-databricks and @murali-db for the original PoC.

## Test plan

- [x] `build/sbt spark/publishM2` → `unitycatalog-spark_4.1_2.13-*.jar` (default)
- [x] `build/sbt -DsparkVersion=4.0 spark/publishM2` → `unitycatalog-spark_4.0_2.13-*.jar`
- [x] `build/sbt -DskipSparkSuffix=true spark/publishM2` → `unitycatalog-spark_2.13-*.jar`
- [x] Cross-Spark publish test (4 tests, all pass locally and in CI)
- [x] CI matrix green for all 4 cells (blocking + non-blocking)
- [x] Shadow PR #1558 validates full 4.2 stack E2E (UC suffixed artifacts → Delta shadow → UC tests)

Signed-off-by: Vishnu Chandrashekhar <vishnu.c@databricks.com>

### DIFF
--- a/.github/workflows/cross-spark-build.yml
+++ b/.github/workflows/cross-spark-build.yml
@@ -1,0 +1,37 @@
+name: "Cross-Spark Build"
+on:
+  push:
+    branches: [main, branch-*]
+    paths-ignore:
+      - '**.md'
+      - '**.txt'
+  pull_request:
+    branches: [main, branch-*]
+    paths-ignore:
+      - '**.md'
+      - '**.txt'
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    name: "Build Test"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@c1e323688fd81a25caa38c78aa6df2d33d3e20d9 # v4.8.0
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'sbt'
+
+      - name: Set up Python
+        uses: actions/setup-python@7f4fc3e22c37d6ff65e88745f38bd3157c663f7c # v4.9.1
+        with:
+          python-version: '3.9'
+
+      - name: Run cross-Spark build test
+        run: python project/tests/test_cross_spark_publish.py
--- a/.github/workflows/cross-spark-build.yml
+++ b/.github/workflows/cross-spark-build.yml
@@ -25,13 +25,8 @@ jobs:
         uses: actions/setup-java@c1e323688fd81a25caa38c78aa6df2d33d3e20d9 # v4.8.0
         with:
           java-version: '17'
-          distribution: 'temurin'
+          distribution: 'zulu'
           cache: 'sbt'
 
-      - name: Set up Python
-        uses: actions/setup-python@7f4fc3e22c37d6ff65e88745f38bd3157c663f7c # v4.9.1
-        with:
-          python-version: '3.9'
-
       - name: Run cross-Spark build test
-        run: python project/tests/test_cross_spark_publish.py
+        run: python3 project/tests/test_cross_spark_publish.py
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -67,26 +67,10 @@ jobs:
         distribution: 'zulu'
         cache: 'sbt'
 
-    - name: Clone Delta master and resolve version
+    - name: Build Delta SNAPSHOT from source
       if: matrix.delta-version == 'master-SNAPSHOT'
-      id: delta-meta
+      id: delta-build
       run: |
-        export DELTA_DIR=/tmp/delta
-        export SPARK_MAJOR_MINOR=$(echo "${{ matrix.spark-version }}" | cut -d. -f1,2)
-        bash project/scripts/build_delta_from_source.sh --meta-only
-
-    - name: Cache Delta SNAPSHOT local Maven artifacts
-      if: matrix.delta-version == 'master-SNAPSHOT'
-      id: delta-cache
-      uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
-      with:
-        path: ~/.m2/repository/io/delta
-        key: delta-snapshot-spark-${{ matrix.spark-version }}-${{ steps.delta-meta.outputs.sha }}
-
-    - name: Build Delta SNAPSHOT and publish to local Maven
-      if: matrix.delta-version == 'master-SNAPSHOT' && steps.delta-cache.outputs.cache-hit != 'true'
-      run: |
-        export DELTA_DIR=/tmp/delta
         export SPARK_MAJOR_MINOR=$(echo "${{ matrix.spark-version }}" | cut -d. -f1,2)
         export SKIP_UC_AUTO_BUILD="${{ matrix.delta-skip-uc-auto-build || 'false' }}"
         bash project/scripts/build_delta_from_source.sh
@@ -94,7 +78,7 @@ jobs:
     - name: Set effective Delta version
       run: |
         if [[ "${{ matrix.delta-version }}" == "master-SNAPSHOT" ]]; then
-          echo "DELTA_VERSION=${{ steps.delta-meta.outputs.version }}" >> $GITHUB_ENV
+          echo "DELTA_VERSION=${{ steps.delta-build.outputs.version }}" >> $GITHUB_ENV
         else
           echo "DELTA_VERSION=${{ matrix.delta-version }}" >> $GITHUB_ENV
         fi
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -49,8 +49,11 @@ jobs:
             non-blocking: true
           - spark-version: "4.2.0-SNAPSHOT"
             delta-version: "master-SNAPSHOT"
-            # SNAPSHOT Spark: disable Delta's UC auto-build (we pre-publish UC
-            # at the expected coordinate in the build step below instead).
+            # SNAPSHOT Spark requires building Delta from source, but Delta's
+            # ensurePinnedUnityCatalog auto-build would clone and build UC at a
+            # pinned coordinate that doesn't match what this CI job publishes.
+            # Disable auto-build and run Delta's setup script explicitly instead
+            # (see the Build Delta SNAPSHOT step below).
             delta-skip-uc-auto-build: true
             validation-mode: "spark-tests"
             non-blocking: true
@@ -68,18 +71,16 @@ jobs:
       if: matrix.delta-version == 'master-SNAPSHOT'
       id: delta-meta
       run: |
-        DELTA_REPO="${{ matrix.delta-repo || 'delta-io/delta' }}"
-        DELTA_REF="${{ matrix.delta-ref || 'master' }}"
-        DELTA_SHA=$(git ls-remote "https://github.com/$DELTA_REPO.git" "$DELTA_REF" | cut -f1)
+        DELTA_SHA=$(git ls-remote "https://github.com/delta-io/delta.git" "master" | cut -f1)
         if [ -z "$DELTA_SHA" ]; then
-          DELTA_SHA=$(git ls-remote "https://github.com/$DELTA_REPO.git" HEAD | cut -f1)
+          DELTA_SHA=$(git ls-remote "https://github.com/delta-io/delta.git" HEAD | cut -f1)
         fi
-        DELTA_VER=$(curl -fsSL "https://raw.githubusercontent.com/$DELTA_REPO/$DELTA_REF/version.sbt" \
+        DELTA_VER=$(curl -fsSL "https://raw.githubusercontent.com/delta-io/delta/master/version.sbt" \
               | grep -oP '(?<=")[^"]*-SNAPSHOT(?=")') || \
-          DELTA_VER=$(curl -fsSL "https://raw.githubusercontent.com/$DELTA_REPO/master/version.sbt" \
+          DELTA_VER=$(curl -fsSL "https://raw.githubusercontent.com/delta-io/delta/master/version.sbt" \
               | grep -oP '(?<=")[^"]*-SNAPSHOT(?=")')
         if [ -z "$DELTA_VER" ]; then
-          echo "::error::Failed to resolve Delta SNAPSHOT version from $DELTA_REPO/$DELTA_REF"
+          echo "::error::Failed to resolve Delta SNAPSHOT version from delta-io/delta/master"
           exit 1
         fi
         echo "sha=$DELTA_SHA" >> $GITHUB_OUTPUT
@@ -91,16 +92,14 @@ jobs:
       uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
       with:
         path: ~/.m2/repository/io/delta
-        key: delta-snapshot-${{ matrix.delta-repo || 'delta-io/delta' }}-${{ steps.delta-meta.outputs.sha }}
+        key: delta-snapshot-${{ steps.delta-meta.outputs.sha }}
 
     - name: Build Delta SNAPSHOT from source and publish to local Maven
       if: matrix.delta-version == 'master-SNAPSHOT' && steps.delta-cache.outputs.cache-hit != 'true'
       run: |
-        DELTA_REPO="${{ matrix.delta-repo || 'delta-io/delta' }}"
-        DELTA_REF="${{ matrix.delta-ref || 'master' }}"
         SPARK_MAJOR_MINOR=$(echo "${{ matrix.spark-version }}" | cut -d. -f1,2)
-        git clone --depth=1 --branch="$DELTA_REF" "https://github.com/$DELTA_REPO.git" /tmp/delta || \
-          git clone --depth=1 "https://github.com/$DELTA_REPO.git" /tmp/delta
+        git clone --depth=1 --branch=master "https://github.com/delta-io/delta.git" /tmp/delta || \
+          git clone --depth=1 "https://github.com/delta-io/delta.git" /tmp/delta
         cd /tmp/delta
         SKIP_UC_FLAG=""
         if [[ "${{ matrix.delta-skip-uc-auto-build }}" == "true" ]]; then
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -51,7 +51,7 @@ jobs:
             delta-version: "master-SNAPSHOT"
             # SNAPSHOT Spark requires building Delta from source, but Delta's
             # ensurePinnedUnityCatalog auto-build would clone and build UC at a
-            # pinned coordinate that doesn't match what this CI job publishes.
+            # pinned UC version that doesn't match what this CI job publishes.
             # Disable auto-build and run Delta's setup script explicitly instead
             # (see the Build Delta SNAPSHOT step below).
             delta-skip-uc-auto-build: true
@@ -64,7 +64,7 @@ jobs:
       uses: actions/setup-java@c1e323688fd81a25caa38c78aa6df2d33d3e20d9 # v4.8.0
       with:
         java-version: '17'
-        distribution: 'temurin'
+        distribution: 'zulu'
         cache: 'sbt'
 
     - name: Clone Delta master
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -20,10 +20,10 @@ jobs:
 
     # Build a human-readable job name from the matrix entry, e.g.:
     #   "All tests (Spark 4.0.0, Delta 4.1.0)"
-    #   "[Non Blocking] Spark compile (Spark 4.2.0-SNAPSHOT)"
+    #   "[Non Blocking] Spark tests (Spark 4.2.0-SNAPSHOT, Delta master-SNAPSHOT)"
     name: >-
       ${{ matrix.non-blocking && '[Non Blocking]' || '' }}
-      ${{ matrix.validation-mode == 'spark-compile-only' && 'Spark compile' || matrix.validation-mode == 'spark-tests' && 'Spark tests' || 'All tests' }}
+      ${{ matrix.validation-mode == 'spark-tests' && 'Spark tests' || 'All tests' }}
       (Spark ${{ matrix.spark-version }}${{ matrix.delta-version != '' && format(', Delta {0}', matrix.delta-version) || '' }})
     continue-on-error: ${{ matrix.non-blocking || false }}
     strategy:
@@ -48,10 +48,10 @@ jobs:
             # indicate upcoming incompatibilities but should not block PRs.
             non-blocking: true
           - spark-version: "4.2.0-SNAPSHOT"
-            delta-version: ""
-            validation-mode: "spark-compile-only"
-            # Delta does not yet publish a Spark 4.2 artifact. This mode applies
-            # only to Spark 4.2; Spark 4.0/4.1 continue to run compile + tests.
+            delta-version: "master-SNAPSHOT"
+            delta-repo: "seewishnew/delta"
+            delta-ref: "enable-spark-master-cross-build"
+            validation-mode: "spark-tests"
             non-blocking: true
 
     steps:
@@ -67,9 +67,20 @@ jobs:
       if: matrix.delta-version == 'master-SNAPSHOT'
       id: delta-meta
       run: |
-        DELTA_SHA=$(git ls-remote https://github.com/delta-io/delta.git HEAD | cut -f1)
-        DELTA_VER=$(curl -fsSL https://raw.githubusercontent.com/delta-io/delta/master/version.sbt \
+        DELTA_REPO="${{ matrix.delta-repo || 'delta-io/delta' }}"
+        DELTA_REF="${{ matrix.delta-ref || 'master' }}"
+        DELTA_SHA=$(git ls-remote "https://github.com/$DELTA_REPO.git" "$DELTA_REF" | cut -f1)
+        if [ -z "$DELTA_SHA" ]; then
+          DELTA_SHA=$(git ls-remote "https://github.com/$DELTA_REPO.git" HEAD | cut -f1)
+        fi
+        DELTA_VER=$(curl -fsSL "https://raw.githubusercontent.com/$DELTA_REPO/$DELTA_REF/version.sbt" \
+              | grep -oP '(?<=")[^"]*-SNAPSHOT(?=")') || \
+          DELTA_VER=$(curl -fsSL "https://raw.githubusercontent.com/$DELTA_REPO/master/version.sbt" \
               | grep -oP '(?<=")[^"]*-SNAPSHOT(?=")')
+        if [ -z "$DELTA_VER" ]; then
+          echo "::error::Failed to resolve Delta SNAPSHOT version from $DELTA_REPO/$DELTA_REF"
+          exit 1
+        fi
         echo "sha=$DELTA_SHA" >> $GITHUB_OUTPUT
         echo "version=$DELTA_VER" >> $GITHUB_OUTPUT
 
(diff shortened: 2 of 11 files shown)


seewishnew:
## Spark 4.2 compatibility fixes added to this branch

While validating the cross-build infra against Spark 4.2.0-SNAPSHOT (via the shadow PR #1558), two Spark 4.2 behavioral changes surfaced that required fixes in the UC connector. These are on `cross-build-infra` and will merge with this PR.

### 1. `createNamespace` exception mapping (commit `61f7907`)

**Spark change:** [SPARK-55250](https://issues.apache.org/jira/browse/SPARK-55250) — *Reduce Hive client calls on CREATE NAMESPACE* (branch-4.2 only, Jan 2026)

**What changed:** `CreateNamespaceExec` no longer pre-checks `catalog.namespaceExists()` before calling `createNamespace`. It now always calls `createNamespace` and catches `NamespaceAlreadyExistsException` for `IF NOT EXISTS` handling.

**Impact on UC:** `UCProxy.createNamespace` was throwing a raw `ApiException` (HTTP 400, `SCHEMA_ALREADY_EXISTS`) on duplicate schemas. Spark 4.2 doesn't catch `ApiException`, so `CREATE DATABASE IF NOT EXISTS` failed when the schema already existed.

**Fix:** Wrap `schemasApi.createSchema()` to catch `ApiException` with `SCHEMA_ALREADY_EXISTS` in the response body and re-throw as `NamespaceAlreadyExistsException`. Safe on Spark 4.0/4.1 — the pre-check prevents `createNamespace` from being called when the namespace exists, so the new catch is never hit.

**File:** `connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala`

### 2. `DESC EXTENDED` output format change (commit `957dd71`)

**Spark change:** [SPARK-56678](https://issues.apache.org/jira/browse/SPARK-56678) — *Use structured Catalog/Namespace/Database/Table rows in DESCRIBE TABLE EXTENDED for v2 tables and views* (branch-4.2 only)

**What changed:** `DescribeTableExec` replaced the single `Name` row (containing the fully-qualified table name) with four structured rows: `Catalog`, `Namespace`, `Database`, `Table`.

**Impact on UC:** `DeltaManagedTableReadWriteTest.testCreateManagedDeltaTable` asserted `describeResult.get("Name")`, which returned `null` on Spark 4.2 since the `Name` key no longer exists.

**Fix:** Version-aware assertion — check for `Name` (4.0/4.1) or fall back to `Table` (4.2). Test-only change.

**File:** `connectors/spark/src/test/java/io/unitycatalog/spark/DeltaManagedTableReadWriteTest.java`

.github/workflows/cross-spark-build.yml
```diff
@@ -0,0 +1,37 @@
...
+    name: "Build Test"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@c1e323688fd81a25caa38c78aa6df2d33d3e20d9 # v4.8.0
+        with:
+          java-version: '17'
+          distribution: 'temurin'
```
openinx: nit:  usually we just use the `zulu` jdk distribution.

.github/workflows/cross-spark-build.yml
```diff
@@ -0,0 +1,37 @@
...
+          cache: 'sbt'
+
+      - name: Set up Python
+        uses: actions/setup-python@7f4fc3e22c37d6ff65e88745f38bd3157c663f7c # v4.9.1
+        with:
+          python-version: '3.9'
+
+      - name: Run cross-Spark build test
+        run: python project/tests/test_cross_spark_publish.py
```
openinx: do we need the `uv` and `python` hardern work ?

.github/workflows/unit-tests.yml
```diff
@@ -43,15 +43,17 @@ jobs:
...
             validation-mode: "spark-tests"
             # Cross-version testing with Delta master is informational; failures
             # indicate upcoming incompatibilities but should not block PRs.
             non-blocking: true
           - spark-version: "4.2.0-SNAPSHOT"
-            delta-version: ""
-            validation-mode: "spark-compile-only"
-            # Delta does not yet publish a Spark 4.2 artifact. This mode applies
-            # only to Spark 4.2; Spark 4.0/4.1 continue to run compile + tests.
+            delta-version: "master-SNAPSHOT"
```
openinx, line 51: Question:   as we discussed offline,  i think we will need to carefully define the test matrix that we want to cover the combos of spark/uc/delta.  I don't want to block this PR,  but pls file a separate issue to track this work.

.github/workflows/unit-tests.yml
```diff
@@ -43,15 +43,17 @@ jobs:
...
+            delta-repo: "seewishnew/delta"
+            delta-ref: "enable-spark-master-cross-build"
```
openinx: why not use the offical delta-io/delta repo ?

.github/workflows/unit-tests.yml
```diff
@@ -43,15 +43,17 @@ jobs:
...
+            delta-skip-uc-auto-build: true
```
openinx: what is this used for ?

.github/workflows/unit-tests.yml
```diff
@@ -67,9 +69,20 @@ jobs:
...
+          DELTA_SHA=$(git ls-remote "https://github.com/$DELTA_REPO.git" HEAD | cut -f1)
+        fi
+        DELTA_VER=$(curl -fsSL "https://raw.githubusercontent.com/$DELTA_REPO/$DELTA_REF/version.sbt" \
+              | grep -oP '(?<=")[^"]*-SNAPSHOT(?=")') || \
+          DELTA_VER=$(curl -fsSL "https://raw.githubusercontent.com/$DELTA_REPO/master/version.sbt" \
               | grep -oP '(?<=")[^"]*-SNAPSHOT(?=")')
+        if [ -z "$DELTA_VER" ]; then
+          echo "::error::Failed to resolve Delta SNAPSHOT version from $DELTA_REPO/$DELTA_REF"
+          exit 1
+        fi
```
openinx: I'm inclined to use the offical repo and `master` to simplify all those, unless we have solid reason to have to do this.

.github/workflows/unit-tests.yml
```diff
@@ -79,56 +92,49 @@ jobs:
...
+        git clone --depth=1 --branch="$DELTA_REF" "https://github.com/$DELTA_REPO.git" /tmp/delta || \
+          git clone --depth=1 "https://github.com/$DELTA_REPO.git" /tmp/delta
         cd /tmp/delta
-        # Publish only delta-spark and the modules its POM references at SNAPSHOT:
-        # storage (dependsOn) and the kernel artifacts re-added via pomPostProcess.
-        # This avoids compiling unrelated Delta projects (hudi, iceberg, flink, connect, ...).
+        SKIP_UC_FLAG=""
+        if [[ "${{ matrix.delta-skip-uc-auto-build }}" == "true" ]]; then
+          SKIP_UC_FLAG="-Ddelta.autoBuildPinnedUnityCatalog=false"
+        fi
```
openinx: this is some kind of duplicated with the above git clone script.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -828,7 +828,13 @@ private class UCProxy(
     createSchema.setCatalogName(this.name)
     createSchema.setName(namespace.head)
     createSchema.setProperties(metadata)
-    schemasApi.createSchema(createSchema)
+    try {
+      schemasApi.createSchema(createSchema)
+    } catch {
+      case e: ApiException if e.getResponseBody != null &&
+        e.getResponseBody.contains("SCHEMA_ALREADY_EXISTS") =>
```
openinx, line 840: pls also give the comment say why change this

openinx (changes requested):

.github/workflows/unit-tests.yml
seewishnew: Good catch, I think this snuck in as I was checking out between the shadow PR branches. Fixed.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
tdas: Why does the production code need to change because of this build infra change???

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -828,7 +828,13 @@ private class UCProxy(
...
+        throw new NamespaceAlreadyExistsException(namespace)
```
tdas, line 841: [AI Assisted] This new mapping is the only place in this file that matches an `ApiException` by substring on the raw response body — every other case in `UCProxy` uses the structured `e.getCode` (lines 400, 401, 559, 808). That makes this brittle: any future tweak to the server's error JSON (key name, structured wrapping, casing) will silently break `CREATE NAMESPACE IF NOT EXISTS` on Spark 4.2.

Looking at `server/src/main/java/io/unitycatalog/server/exception/ErrorCode.java:49-54`, `SCHEMA_ALREADY_EXISTS` maps to HTTP 400 (with a comment noting the new Delta API returns 409). At minimum, please also gate on `e.getCode == 400 || e.getCode == 409` so an error-format refactor doesn't quietly regress this path. Better: parse the server error body's `errorCode` field properly the way other UC clients do.

Also — there's no test covering this mapping. I grepped the spark test tree and the only `CREATE DATABASE IF NOT EXISTS` test runs against `spark_catalog`, not the UC catalog, so this code path isn't exercised. Given the Spark 4.2 cell is `non-blocking: true`, a regression here can ship undetected. Please add a unit/integration test that creates a UC schema twice with `IF NOT EXISTS` and asserts it does not throw — that test will work on Spark 4.0/4.1 too (the pre-check just makes the catch unreachable there).

Finally, please add a 3-4 line comment above the try/catch explaining (a) why this exists (SPARK-55250), (b) which Spark versions are affected, and (c) why the body string is being matched. This is exactly the kind of cross-Spark-version domain knowledge that needs to live in the code.

project/CrossSparkVersions.scala
```diff
@@ -0,0 +1,198 @@
...
+
+      val task = args.mkString(" ")
+      val extracted = sbt.Project.extract(state)
+      val sparkVersionKey = autoImport.sparkVersion
+      val publishArtifactKey = SettingKey[Boolean]("publishArtifact")
+
+      val sparkModules = extracted.structure.allProjectRefs.filter { ref =>
+        val hasSpark = (ref / sparkVersionKey).get(extracted.structure.data).isDefined
+        val publishable = (ref / publishArtifactKey).get(extracted.structure.data).getOrElse(true)
+        hasSpark && publishable
```
tdas: [AI Assisted] Constructing `SettingKey[Boolean]("publishArtifact")` by string label is an anti-pattern — please use `Keys.publishArtifact` directly (already in scope via `import sbt.Keys._` at the top of this file). This currently happens to work only because sbt's `AttributeKey` identifies by label+manifest; any sbt or convention change would break it silently with no compile error.

One related caveat worth a one-line comment for the next reader: `(ref / publishArtifactKey)` reads the **project-scoped** `publishArtifact`, but `build.sbt:73` defines `Test / publishArtifact := false`. That's fine for the current intent, but without a comment a future reader may wrongly conclude the discovery is filtering out test artifacts here.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
seewishnew: Justification here: https://github.com/unitycatalog/unitycatalog/pull/1556#issuecomment-4510063292.

This missing gap was caught by the shadow PRs' failures:
1. https://github.com/delta-io/delta/pull/6824
2. https://github.com/unitycatalog/unitycatalog/pull/1558

Even though the 4.2.0 check is failing here for other delta-incompatibility reasons, I wanted certainty that the missing gaps are addressed here before landing this.

The delta shadow PR is primed for becoming the proper fix in delta for accommodating uc artifacts with spark version specific suffixes.

Overall, this gets us a lot closer to the point where we can block on failing 4.2.0 tests and catch drifts and need for shims more readily.

.github/workflows/cross-spark-build.yml
seewishnew: Switched to `zulu` in fdd817b.

.github/workflows/cross-spark-build.yml
seewishnew: The test script (`project/tests/test_cross_spark_publish.py`) only uses Python standard library modules (`re`, `subprocess`, `sys`, `pathlib`, `shutil`) — no third-party deps, no `uv`/`pyproject.toml` needed. Removed the explicit `setup-python` step since `ubuntu-latest` already ships with `python3`. Fixed in fdd817b.

.github/workflows/unit-tests.yml
seewishnew: Filed #1565 to track defining the comprehensive Spark/UC/Delta test matrix.

.github/workflows/unit-tests.yml
seewishnew: Expanded the comment to explain the mechanism: Delta's `ensurePinnedUnityCatalog` auto-build would clone and build UC at a pinned coordinate that doesn't match what this CI job publishes, so we disable it and run the setup script explicitly instead. Fixed in fdd817b.

.github/workflows/unit-tests.yml
seewishnew: Agreed — simplified. Removed the `delta-repo`/`delta-ref` parameterization and hardcoded `delta-io/delta` / `master`. Also restructured the flow to clone Delta once upfront and read `version.sbt` + SHA directly from the clone, eliminating the separate `curl` + `git ls-remote` steps. The build step now just `cd`s into the existing clone. Fixed in fdd817b + 670f0dd.

.github/workflows/unit-tests.yml
seewishnew: Good catch. The duplication existed because `version.sbt` was fetched via `curl` before the clone, and then Delta was cloned again in the build step. Restructured to clone Delta once upfront — the build step now just `cd`s into the existing clone (670f0dd).

The setup script inside Delta (`setup_unitycatalog_main.sh`) does clone UC separately from the checkout, but that's a different repo at a different ref with version-coordinate suffixing that Delta's build contract requires (the Ivy coordinate includes a 7-char SHA suffix). Once the follow-up Delta PR lands with per-Spark-version UC dependency support, we can revisit this to pre-publish UC from the checkout instead of re-cloning via the setup script.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
seewishnew: Added a SPARK-55250 comment explaining the cross-version behavior, strengthened the catch guard with `e.getCode == 400 || e.getCode == 409` (matching the rest of the file), and added `testCreateSchemaIfNotExists` in `SchemaOperationsTest.java`. Fixed in fdd817b.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
seewishnew: Good points. Addressed all three:

1. **HTTP status guard**: Added `e.getCode == 400 || e.getCode == 409` alongside the body check, consistent with the rest of the file (lines 400, 401, 559, 808). The 400 matches current UC server behavior; 409 future-proofs against the `deltaHttpStatus` defined in `ErrorCode.java:51`.

2. **Test coverage**: Added `testCreateSchemaIfNotExists` in `SchemaOperationsTest.java` — creates a schema, then runs `CREATE DATABASE IF NOT EXISTS` on the same name, asserts no exception. Exercises the catch on Spark 4.2; on 4.0/4.1 the pre-check makes it a no-op.

3. **Comment**: Added a 4-line SPARK-55250 comment above the `try` block explaining why, which versions are affected, and why body matching is needed.

Fixed in fdd817b.

project/CrossSparkVersions.scala
seewishnew: Good catch — replaced `SettingKey[Boolean]("publishArtifact")` with `Keys.publishArtifact` (already in scope via `import sbt.Keys._`). Added a comment noting this reads the project-scoped setting (not `Test/publishArtifact`). Fixed in fdd817b.

.github/workflows/unit-tests.yml
```diff
@@ -81,54 +90,44 @@ jobs:
...
       if: matrix.delta-version == 'master-SNAPSHOT' && steps.delta-cache.outputs.cache-hit != 'true'
       run: |
         SPARK_MAJOR_MINOR=$(echo "${{ matrix.spark-version }}" | cut -d. -f1,2)
-        git clone --depth=1 https://github.com/delta-io/delta.git /tmp/delta
         cd /tmp/delta
-        # Publish only delta-spark and the modules its POM references at SNAPSHOT:
-        # storage (dependsOn) and the kernel artifacts re-added via pomPostProcess.
-        # This avoids compiling unrelated Delta projects (hudi, iceberg, flink, connect, ...).
+        SKIP_UC_FLAG=""
+        if [[ "${{ matrix.delta-skip-uc-auto-build }}" == "true" ]]; then
```
openinx: Instead of spreading out the delta build and install script piece everywhere, could you pls compact them into a bash script so that we can always execute the bash in one line command directly ? that help us to better understand the core job orchestration in the workflow.

.github/workflows/unit-tests.yml
```diff
@@ -81,54 +90,44 @@ jobs:
...
+          SKIP_UC_FLAG="-Ddelta.autoBuildPinnedUnityCatalog=false"
+          SPARK_MAJOR_MINOR=$SPARK_MAJOR_MINOR bash project/scripts/setup_unitycatalog_main.sh
+        fi
         build/sbt -DsparkVersion=$SPARK_MAJOR_MINOR \
+          $SKIP_UC_FLAG \
```
openinx: oh....  that have to inject the `delta.autoBuildPinnedUnityCatalog` inside the delta `build.sbt` ?  that's so coupled....  why do we need that system property ?  is it possible to simplify this ?

seewishnew: No, it is fundamentally coupled. I went through a lot of iterations last night trying to get this all to work end to end and validate through the shadow PRs. I think something like this is really needed to overcome the circularity and avoid maven cache pollution in UC CI.

This also takes us one step closer to making this check blocking.

project/tests/test_cross_spark_publish.py
```diff
@@ -0,0 +1,387 @@
...
+
+    def __init__(self, uc_root: Path):
+        self.uc_root = uc_root
+        self.uc_version = self._get_uc_version()
+
+    def _get_uc_version(self) -> str:
+        with open(self.uc_root / "version.sbt", "r") as f:
+            for line in f:
+                if "version :=" in line:
+                    return line.split('"')[1]
```
openinx: Could you pls use the regex to match the specifc version number ?   the exact parsing is easily to get bug.

project/tests/test_cross_spark_publish.py
```diff
@@ -0,0 +1,387 @@
...
+        sys.exit("Error: Could not parse version from version.sbt")
+
+    def clean_maven_cache(self) -> None:
+        import shutil
```
openinx: why do we use the `method` level package import ?  could we unify it to use the package import ?

project/CrossSparkVersions.scala
```diff
@@ -0,0 +1,191 @@
...
+
+  val spark41 = SparkVersionSpec(
+    fullVersion = "4.1.0",
+    additionalSourceDir = "scala-shims/spark-4.1"
+  )
+
+  val spark42Snapshot = SparkVersionSpec(
+    fullVersion = "4.2.0-SNAPSHOT",
+    additionalSourceDir = "scala-shims/spark-4.2"
+  )
```
openinx: Could you pls maintain those versions information in a json file directly under the `project` dir (source of truth) , since I don't want to use any special python script to execute the sbt command to parse the version spec any more ( i mean in the github workflow yaml files ) ,  that's unnecessary.

tdas (approved):
LGTM

tdas:
but please do get signoff from @openinx

project/CrossSparkVersions.scala
```diff
@@ -0,0 +1,191 @@
...
+   * Release steps that publish artifacts for all Spark versions.
+   *
+   * Step 1: publish ALL modules WITHOUT Spark suffix (backward compat)
+   * Step 2+: publish Spark-dependent modules WITH suffix for each non-snapshot version
+   *
+   * Each step runs as a subprocess so SBT reloads with the correct settings.
+   */
+  def crossSparkReleaseSteps(task: String): Seq[ReleaseStep] = {
+
+    def runSbtSubprocess(state: State, sbtArgs: Seq[String], description: String): State = {
```
openinx: wait, do we have to use the sbt subprocess to run all those steps ? since I remembered the `MAVEN_PROXY_URL` env variable cannot propogate from the parent process to subprocess, which actually bring us a huge trouble when we do the release work.

with that, we have to manually split the sbt into several sbt commands, and execute them one by one to finish the release, rather than execute this native command directly..

build.sbt
```diff
@@ -602,6 +602,12 @@ lazy val spark = (project in file("connectors/spark"))
   .enablePlugins(CheckstylePlugin)
   .settings(
     name := s"$artifactNamePrefix-spark",
+    Keys.moduleName := CrossSparkVersions.sparkVersionedModuleName(name.value),
+    CrossSparkVersions.sparkDependentSettings,
+    Compile / unmanagedJars ++= {
+      val sparkAssemblyDir = sys.props.get("sparkAssemblyDir").map(file).filter(_.exists)
+      sparkAssemblyDir.toSeq.flatMap(d => (d ** "*.jar").get.classpath)
```
openinx: what is this used for ?  i did not get.

openinx: seems like it's an unrelated change to this PR,  maybe we could remove it.

build.sbt
```diff
@@ -651,7 +657,12 @@ lazy val spark = (project in file("connectors/spark"))
       "org.apache.hadoop" % "hadoop-aws" % hadoopVersion % Test,
       "org.projectlombok" % "lombok" % "1.18.32" % Test,
       "com.google.cloud.bigdataoss" % "gcs-connector" % "3.0.2" % Test classifier "shaded",
-      "io.delta" %% s"delta-spark_$sparkMajorMinorVersion" % deltaVersion % Test,
+    ) ++ (
+      // delta-spark is only needed for tests. When callers only need publishLocal/publishM2
+      // (e.g. Delta's setup_unitycatalog_main.sh), the matching Delta artifact may not exist
+      // yet. Pass -Duc.skipTestDeps=true to exclude it and avoid resolution failures.
+      if (sys.props.getOrElse("uc.skipTestDeps", "false").toBoolean) Seq.empty
```
openinx: `uc.skipTestDeps` -> `skipDeltaSpark`

build.sbt
seewishnew: **`sparkAssemblyDir` (line 613-615):** Lets you point SBT at a locally-built Spark for compilation via `-DsparkAssemblyDir=/path/to/spark/assembly`. Useful for developing against Spark SNAPSHOT locally without publishing Spark to `~/.m2` first. Not used in CI today, but valuable for local cross-build iteration.

build.sbt
```diff
@@ -739,10 +750,12 @@ lazy val integrationTests = (project in file("integration-tests"))
       "org.assertj" % "assertj-core" % "3.26.3" % Test,
       "org.projectlombok" % "lombok" % "1.18.32" % Provided,
       "org.apache.spark" %% "spark-sql" % sparkVersion % Test,
-      "io.delta" %% s"delta-spark_$sparkMajorMinorVersion" % deltaVersion % Test,
       "org.apache.hadoop" % "hadoop-aws" % hadoopVersion % Test,
       "org.apache.hadoop" % "hadoop-azure" % hadoopVersion % Test,
       "com.google.cloud.bigdataoss" % "gcs-connector" % "3.0.2" % Test classifier "shaded",
+    ) ++ (
+      if (sys.props.getOrElse("uc.skipTestDeps", "false").toBoolean) Seq.empty
+      else Seq("io.delta" %% s"delta-spark_$sparkMajorMinorVersion" % deltaVersion % Test)
```
openinx: Could you pls just define a `deltaSparkDeps` method to get the Seq pls ?  so that we don't have to duplicate the `if ... else` everywhere.

project/CrossSparkVersions.scala
seewishnew: Forwarding the env vars below: 
```scala
      val envVars = sys.env.toSeq
      val exitCode = scala.sys.process.Process(cmd, baseDir, envVars: _*).!
```

project/scripts/build_delta_from_source.sh
```diff
@@ -0,0 +1,65 @@
...
+  esac
+done
+
+# ── Clone ────────────────────────────────────────────────────────────────────
+if [ -d "$DELTA_DIR/.git" ]; then
+  echo "Delta already cloned at $DELTA_DIR -- skipping clone."
+else
+  echo "Cloning delta-io/delta into $DELTA_DIR ..."
+  git clone --depth=1 --branch=master "https://github.com/delta-io/delta.git" "$DELTA_DIR" || \
+    git clone --depth=1 "https://github.com/delta-io/delta.git" "$DELTA_DIR"
```
openinx: Let us remove this fallback git clone.

project/scripts/build_delta_from_source.sh
```diff
@@ -0,0 +1,65 @@
...
+
+emit_output() {
+  local key="$1" value="$2"
+  if [ -n "${GITHUB_OUTPUT:-}" ]; then
+    echo "$key=$value" >> "$GITHUB_OUTPUT"
+  fi
+  echo "$key=$value"
+}
+
+emit_output "sha" "$DELTA_SHA"
```
openinx: sha -> `DELTA_SHA`.  and version -> `DELTA_VERSION`.

openinx (approved):
LGTM now,  thanks @seewishnew for the work, and thanks @tdas for the reviewing.

PR #1557, opened 2026-05-20
Stop requiring inCommitTimestampEnablement{Version,Timestamp}

yili-db:
**PR Checklist**

**Description of changes**
Stop requiring inCommitTimestampEnablement{Version,Timestamp}

Per Delta PROTOCOL.md §4 these properties only apply when ICT was enabled mid-table. UC catalog-managed Delta tables enable ICT at v0, so the values are inapplicable.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
@@ -70,12 +70,9 @@ public final class UcManagedDeltaContract {
   /**
    * Required properties whose value the engine must compute at commit time. The staging response
    * sends them with {@code null} values; the createTable request must echo them back with non-null
-   * values the engine substituted.
+   * values the engine substituted. Currently empty.
    */
-  public static final List<String> ENGINE_GENERATED_PROPERTY_KEYS =
-      List.of(
-          TableProperties.IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION,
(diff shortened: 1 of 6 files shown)


server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
```diff
@@ -70,12 +70,9 @@ private UcManagedDeltaContract() {}
   /**
    * Required properties whose value the engine must compute at commit time. The staging response
    * sends them with {@code null} values; the createTable request must echo them back with non-null
-   * values the engine substituted.
+   * values the engine substituted. Currently empty.
```
tdas, line 73: nit: we dont have to give a commentary on the contents of the list when its right below. :)

yili-db: This is to answer "yes" to the confusion of "Is it intentionally empty?"
"Currently" also means that it could be non-empty in the future and that's why it's still here despite being empty.

tdas: > This is to answer "yes" to the confusion of "Is it intentionally empty?"

And we expect someone also modify the comment to remove it when this list is not empty any more?

if that change is AI generated.. sure AI can handle it. but quite a bit of chance that it will diverge and cause confusion.

you can say "this is placeholder for adding properties in the future" as an "Is it intentionally empty?" .. that  wont go stale like this even if its not empty.

tdas (approved):

tdas (approved):

PR #1562, opened 2026-05-21
Route managed Delta staging of CREATE / CTAS to the UC Delta API

yili-db:
**PR Checklist**

**Description of changes**
Route managed Delta staging of CREATE / CTAS to the UC Delta API

UCSingleCatalog skips its legacy createStagingTable for managed-Delta CREATE and CTAS when the bundled Delta library is new enough (>= 4.3.0) and the catalog opts in via deltaRestApi.enabled, letting Delta stage and finalize via the UC Delta API instead.
New DeltaAvailability helper detects readiness;
BaseSparkIntegrationTest gains a -Duc.test.deltaRestApi.enabled toggle.
Pure-helper unit tests cover compareVersions, deltaVersionAtLeast, and the isDeltaRestApiReady predicate.


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -65,6 +74,7 @@ class UCSingleCatalog
     val serverSidePlanningEnabled = OptionsUtil.getBoolean(options,
       OptionsUtil.SERVER_SIDE_PLANNING_ENABLED,
       OptionsUtil.DEFAULT_SERVER_SIDE_PLANNING_ENABLED)
+    deltaRestApiEnabled = options.getBoolean("deltaRestApi.enabled", false)
```
openinx: Could you pls define it as constant value like other catalog config ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -48,6 +48,15 @@ class UCSingleCatalog
   private[this] var apiClient: ApiClient = null;
   private[this] var tablesApi: TablesApi = null
 
+  // When true, callers that target managed Delta tables skip the legacy UC staging step here and
+  // let the Delta catalog stage the table via the UC Delta API instead.
+  private[this] var deltaRestApiEnabled: Boolean = false
```
openinx, line 56: i think the comment may have some confusion.   I mean this catalog config currently actually only control the `createStagingTable` behavior,  even though we set it to `true`, for other apis we are still using the old api.  I think that's the confusion.

openinx: either we make it more clear in the javadoc to explicit say that we only control the `createStagingTable`, or  we use this to control all the apis switch. 

for the latter one,  seems like it's more complicated work.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -79,6 +89,7 @@ class UCSingleCatalog
         delegate.asInstanceOf[DelegatingCatalogExtension].setDelegateCatalog(proxy)
         delegate.initialize(name, options)
         UCSingleCatalog.DELTA_CATALOG_LOADED.set(true)
+        deltaCatalogLoaded = true
```
openinx, line 93: but we already have the `UCSingleCatalog.DELTA_CATALOG_LOADED`, why do we still need to duplicate another variable to represent the same meaning ?  I assume it's not worth.

connectors/spark/src/main/scala/io/unitycatalog/spark/DeltaAvailability.scala
```diff
@@ -0,0 +1,73 @@
...
+    val version =
+      try {
+        val pkg = Class.forName("io.delta.package$")
+        val module = pkg.getField("MODULE$").get(null)
+        pkg.getMethod("VERSION").invoke(module).asInstanceOf[String]
+      } catch {
+        case _: Throwable => return false
+      }
+    compareVersions(version, target) >= 0
+  }
```
openinx: for the delta version parsing,  i think we should fully reuse the `ApiClientFactory#getDeltaVersion`,  maybe we can just move it out of the `ApiCilentFactory`,  and move all them (including the `compareVersions` ) into the `DeltaUtils`.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
```diff
@@ -61,6 +61,10 @@ protected SparkSession createSparkSessionWithCatalogs(
             .master("local[*]")
             .config("spark.sql.shuffle.partitions", "4")
             .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension");
+    // Opt this test process into the UC Delta API path for managed Delta creates via
+    // `-Duc.test.deltaRestApi.enabled=true`. Off by default so existing tests keep their
```
openinx, line 65: is this a `-D` table property for it ?  I think it's a catalog config, which we cannot specify it via `-D`.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
```diff
@@ -70,7 +74,8 @@ protected SparkSession createSparkSessionWithCatalogs(
               .config(catalogConf + "." + OptionsUtil.TOKEN, serverConfig.getAuthToken())
               .config(catalogConf + "." + OptionsUtil.WAREHOUSE, catalog)
               .config(catalogConf + "." + OptionsUtil.RENEW_CREDENTIAL_ENABLED, renewCred)
-              .config(catalogConf + "." + OptionsUtil.CRED_SCOPED_FS_ENABLED, credScopedFsEnabled);
+              .config(catalogConf + "." + OptionsUtil.CRED_SCOPED_FS_ENABLED, credScopedFsEnabled)
+              .config(catalogConf + ".deltaRestApi.enabled", deltaRestApiEnabled);
```
openinx, line 78: pls use the the constant variable, rather the hardcode value.

openinx (changes requested):

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
yili-db: It does say `legacy UC staging step ` and `stage the table` in the comment

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
yili-db: UCSingleCatalog.DELTA_CATALOG_LOADED is a static ThreadLocal boolean for **testing purpose only**.
Likely it can be used for this purpose but it seems fragile as multiple UCSingleCatalog instances, when executed in the same thread.
When a Spark command line that loads two catalogs using two UCSingleCatalog instances, with one of them loading Delta and the other doesn't, then UCSingleCatalog.DELTA_CATALOG_LOADED is likely having trouble because it's per-thread, not per-object.
Another problem is that even if one single catalog is loaded, UCSingleCatalog.DELTA_CATALOG_LOADED becomes `false` when reading from anther thread.

connectors/spark/src/main/scala/io/unitycatalog/spark/DeltaAvailability.scala
yili-db: Combined to `DeltaVersionUtils.java` which is now moved from `test` to `main`

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
yili-db: This is reading from `System.getProperty` and setting it as a catalog conf.
Anyway this is just a placeholder for now.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -90,6 +99,24 @@ class UCSingleCatalog
     }
   }
 
+  /** See [[DeltaVersionUtils.isDeltaRestApiReady]] for the predicate. */
+  private def deltaRestApiReady: Boolean =
```
tdas: shouldUseDeltaAPI?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -90,6 +99,24 @@ class UCSingleCatalog
...
+    DeltaVersionUtils.isDeltaRestApiReady(deltaCatalogLoaded, deltaRestApiEnabled)
```
tdas, line 106: follow up ... please put as much of the delta related context into a separate interface that this class extends.

tdas (approved):
high level lgtm. no red flags. @openinx please approve too.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: sounds good, thanks @yili-db !

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: Hi @yili-db , for my understanding, I think we can fully reuse the two. if we are running in the same process, there is no reason that one thread can successfully load the delta classes, while another thread cannot load. 

I think in the original code,  we used the `ThreadLocal` variable, because we wanted to isolate the impact and disruption during run many spark session test cases in parallel. 

And we already have two flags, one is `LOAD_DELTA_CATALOG`,  another is`DELTA_CATALOG_LOADED`,   I don't want to introduce another 3rd flag to make this class loading logic even more complicated, and hard to understand.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -118,12 +145,20 @@ class UCSingleCatalog
...
       delegate.createTable(ident, columns, partitions, newProps)
     } else if (hasLocationClause) {
       val newProps = prepareExternalTableProperties(properties)
       delegate.createTable(ident, columns, partitions, newProps)
     } else {
+      // Managed (no LOCATION, no EXTERNAL) but not Delta: UC doesn't store non-Delta managed
+      // tables. Surface the same friendly error the legacy staging path used to produce.
+      // Path-based identifiers (e.g. `delta`.`/tmp/foo`) fall through to the delegate -- their
+      // ident shape is what `isManagedDeltaTable` excluded above.
+      val isPathTable = ident.namespace().length == 1 && new Path(ident.name()).isAbsolute
```
openinx: could you pls share the same method from the `isManagedDeltaTable`, rather than duplicate the code here ?  that's easy to diverge.

openinx (approved):
No blocker to merge, but I have several comments. Pls solve the comments before merging the PR.

timothyw553 (approved):

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
yili-db: @openinx 
But a `static ThreadLocal` has two problems:
1. It is the same boolean for all catalogs, when accessed in the same thread: you have explained that this is fine because they should all be true or all be false. That's fair.
2. It's a **different** boolean for the same catalog, when accessed from a different thread: any call to createTable will see it as **`false`** unless it happens to be called from the **same thread** where the instance was created.

It is only a mechanism for tests. If we want to reuse it for real work, at the minimum it needs to be changed from `ThreadLocal` to `AtomicBoolean`. But I am not sure how that affects concurrent tests runs. I'd prefer to leave it as is but introduce another different non-static boolean.

PR #1567, opened 2026-05-23
Fix derive table properties and make deletionVectors not required

yili-db:
**PR Checklist**

**Description of changes**
Fix derive table properties and make deletionVectors not required:
- Derive table properties delta.minReaderVersion, delta.minWriterVersion, delta.lastUpdateVersion
- Fix delta.clusteringColumns to clusteringColumns
- Change feature deletionVectors from required to recommended because it's not compatible with IcebergV2

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaConsts.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaConsts.java
@@ -98,7 +98,7 @@ public final class DeltaConsts {
      * of segment names, so nested columns stay as arrays rather than collapsing to dotted strings).
      * Mirrors the {@code delta.clustering} domain-metadata entry.
      */
-    public static final String CLUSTERING_COLUMNS = "delta.clusteringColumns";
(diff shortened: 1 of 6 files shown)


timothyw553 (approved):

server/src/main/java/io/unitycatalog/server/service/delta/DeltaConsts.java
```diff
@@ -98,7 +98,7 @@ private TableProperties() {}
      * of segment names, so nested columns stay as arrays rather than collapsing to dotted strings).
      * Mirrors the {@code delta.clustering} domain-metadata entry.
      */
-    public static final String CLUSTERING_COLUMNS = "delta.clusteringColumns";
+    public static final String CLUSTERING_COLUMNS = "clusteringColumns";
```
tdas, line 101: can we identify which properties are derived vs not in some way?
need not be in this PR.

tdas (approved):
lgTM. with one follow up comment.

PR #1569, opened 2026-05-26
[Hadoop] Add buildForStagingTable to UCCredentialHadoopConfs

openinx:
**PR Checklist**

**Description of changes**

Wire the Delta REST `getStagingTableCredentials` endpoint
(`GET /delta/v1/staging-tables/{table_id}/credentials`) into the Hadoop
credential-vending layer so connectors (Spark, Flink, Trino, etc.) can obtain
staging-table credentials through the same `UCCredentialHadoopConfs` builder API.

### New public API

```java
Map<String, String> props = UCCredentialHadoopConfs.builder(uri, "s3")
    .tokenProvider(tokenProvider)
    .enableCredentialRenewal(true)
    .enableCredentialScopedFs(true)
    .hadoopConf(hadoopConf)
    .buildForStagingTable(stagingTableId, location);
```

### Changes

| File | What |
|------|------|
| `UCHadoopConfConstants` | New `UC_DELTA_STAGING_TABLE_ID_KEY` config key |
| `CredPropsUtil.PropsBuilder` | New `deltaStagingTableId()` method |
| `CredPropsUtil` | New `createDeltaStagingTableCredProps` + `fetchDeltaStagingTableCredProps` |
| `GenericCredentialFetcher` | New `forUcDeltaStagingTable()` factory + routing in `create()` |
| `UCDeltaStagingTableCredentialFetcher` | New fetcher that calls `getStagingTableCredentials` |
| `UCCredentialHadoopConfs.Builder` | New `buildForStagingTable(stagingTableId, location)` |

### Tests

- `UCCredentialHadoopConfsTest`: builder validation for `buildForStagingTable` (missing fields, missing token provider)
- `CredPropsUtilTest`: `createDeltaStagingTableCredProps` for S3/GCS/ABFS, static creds, unknown scheme, cred-scoped FS, `PropsBuilder` conflict guards, `fetchDeltaStagingTableCredProps` orchestration
- `UCDeltaStagingTableCredentialFetcherTest`: happy path, immutable-after-construction, null response, missing config keys

All 133 hadoop module tests pass locally.

### Note

`UCSingleCatalog` is **not** changed — it continues using the existing `buildForTable(stagingTableId, READ_WRITE)` path via the UC REST API. The new `buildForStagingTable` is intended for future Delta REST API integration.

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
@@ -71,7 +71,7 @@ public final class UCCredentialHadoopConfs {
 
   /**
    * Collects credential settings and produces Hadoop configuration properties via {@link
-   * #buildForTable} or {@link #buildForPath}.
+   * #buildForTable}, {@link #buildForStagingTable}, or {@link #buildForPath}.
    */
   public static final class Builder {
 
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
@@ -217,6 +217,37 @@ public final class UCCredentialHadoopConfs {
     }
 
     /**
+     * Builds Hadoop properties for a <em>Delta staging table</em> using the UC Delta staging table
+     * credentials API ({@code GET /delta/v1/staging-tables/{table_id}/credentials}).
+     *
+     * @param stagingTableId the staging table UUID
+     * @param location the staging table storage location, used to select the matching storage
+     *     credential from the response
+     * @return unmodifiable map; empty if the scheme is unrecognized
+     * @throws IllegalArgumentException if {@code stagingTableId}, {@code location}, or {@code
+     *     tokenProvider} is missing
+     * @throws ApiException if the credential fetch from the UC Delta API fails
+     */
+    public Map<String, String> buildForStagingTable(String stagingTableId, String location)
+        throws ApiException {
+      Preconditions.checkArgument(
+          stagingTableId != null && !stagingTableId.isEmpty(), "stagingTableId is required");
+      Preconditions.checkArgument(location != null && !location.isEmpty(), "location is required");
+      Preconditions.checkArgument(tokenProvider != null, "tokenProvider is required");
+      return CredPropsUtil.fetchDeltaStagingTableCredProps(
+          credentialRenewalEnabled,
+          credentialScopedFsEnabled,
+          hadoopConf,
+          scheme,
+          apiClient,
+          catalogUri,
+          tokenProvider,
+          stagingTableId,
+          location,
+          appVersions);
+    }
+
+    /**
      * Builds Hadoop properties for an <em>external path</em> using the UC REST credentials API.
      *
      * @return unmodifiable map; empty if the scheme is unrecognized
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -98,16 +98,18 @@ public class CredPropsUtil {
 
     public T tableId(String tableId) {
       Preconditions.checkState(
-          !builder.containsKey(UCHadoopConfConstants.UC_DELTA_CATALOG_KEY),
-          "tableId cannot be set with UC Delta table identifier");
+          !builder.containsKey(UCHadoopConfConstants.UC_DELTA_CATALOG_KEY)
+              && !builder.containsKey(UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_ID_KEY),
+          "tableId cannot be set with UC Delta table or staging table identifier");
       builder.put(UCHadoopConfConstants.UC_TABLE_ID_KEY, tableId);
       return self();
     }
 
     public T ucDeltaTableIdentifier(UCDeltaTableIdentifier identifier, String location) {
       Preconditions.checkState(
-          !builder.containsKey(UCHadoopConfConstants.UC_TABLE_ID_KEY),
-          "UC Delta table identifier cannot be set with tableId");
+          !builder.containsKey(UCHadoopConfConstants.UC_TABLE_ID_KEY)
+              && !builder.containsKey(UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_ID_KEY),
+          "UC Delta table identifier cannot be set with tableId or staging table identifier");
       builder.put(UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_KEY, "true");
       builder.put(
           UCHadoopConfConstants.UC_CREDENTIALS_TYPE_KEY,
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -126,6 +126,7 @@ public class CredPropsUtil {
           !builder.containsKey(UCHadoopConfConstants.UC_TABLE_ID_KEY)
               && !builder.containsKey(UCHadoopConfConstants.UC_DELTA_CATALOG_KEY),
           "deltaStagingTableId cannot be set with tableId or UC Delta table identifier");
+      builder.put(UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_KEY, "true");
       builder.put(UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_ID_KEY, stagingTableId);
       builder.put(UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_LOCATION_KEY, location);
       return self();
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -860,6 +860,7 @@ public class CredPropsUtil {
       Map<String, String> appVersions)
       throws ApiException {
     Configuration reqConf = new Configuration(false);
+    reqConf.set(UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_KEY, "true");
     reqConf.set(UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_ID_KEY, stagingTableId);
     reqConf.set(UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_LOCATION_KEY, location);
     TemporaryCredentials creds =
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -121,13 +121,13 @@ public class CredPropsUtil {
       return self();
     }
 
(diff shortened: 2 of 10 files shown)


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialFetcher.java
```diff
@@ -31,13 +31,24 @@ static GenericCredentialFetcher forUcDelta(
...
    * configuration containing only the credential-request keys (type, table/path id, operation, and
    * for Delta: catalog/schema/table/location). Auth keys are not required — the provided {@code
    * apiClient} already carries authentication.
    */
   static GenericCredentialFetcher create(ApiClient apiClient, Configuration conf) {
+    String stagingTableId = conf.get(UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_ID_KEY);
+    if (stagingTableId != null) {
+      return forUcDeltaStagingTable(
+          conf, new io.unitycatalog.client.delta.api.TemporaryCredentialsApi(apiClient));
+    }
```
openinx: It may be better for us to move this uc delta staging table credential create into the `if(useDeltaCredentialsApi)  ` code block, since it's used for ucDelta api.

tdas (approved):
high level LGTM.. @yili-db to do a detailed review and approve

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConfConstants.java
```diff
@@ -80,6 +80,10 @@ private UCHadoopConfConstants() {}
   public static final String UC_DELTA_TABLE_NAME_KEY = "fs.unitycatalog.delta.table.name";
   public static final String UC_DELTA_LOCATION_KEY = "fs.unitycatalog.delta.location";
 
+  // Key for UC Delta staging table credential requests.
+  public static final String UC_DELTA_STAGING_TABLE_ID_KEY =
```
yili-db: maybe you can still reuse `UC_TABLE_ID_KEY`? The staging table ID is the same table ID when it becomes a real table.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialFetcher.java
yili-db: Yes, it should be inside `useDeltaCredentialsApi` block.
Then you can even reuse the `UC_TABLE_ID_KEY` as:
```
if (useDeltaCredentialsApi) {
  if (it has UC_TABLE_ID_KEY) this is a staging table
  else this is a table
} ...
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
```diff
@@ -0,0 +1,45 @@
...
+        response != null,
+        "UC Delta API returned no credentials response for staging table '%s'.",
+        stagingTableId);
+    return new GenericCredential(
+        DeltaStorageCredentialUtil.toTemporaryCredentials(
+            DeltaStorageCredentialUtil.selectForLocation(
+                location, response.getStorageCredentials())));
+  }
+
+  private static String require(Configuration conf, String key) {
```
yili-db: This function can be made a helper. It's done the same way in 3 fetchers now.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.hadoop.conf.Configuration;
+
+/** Adapts the UC Delta staging table credentials SDK API for Hadoop token providers. */
+final class UCDeltaStagingTableCredentialFetcher implements GenericCredentialFetcher {
+  private final TemporaryCredentialsApi api;
+  private final UUID stagingTableId;
+  private final String location;
+
+  UCDeltaStagingTableCredentialFetcher(Configuration conf, TemporaryCredentialsApi api) {
+    Preconditions.checkNotNull(api, "api is required");
```
yili-db: If `Preconditions.checkNotNull` is made to return the value, this can be made:
`this.api = Preconditions.checkNotNull(api, "api is required");`

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -549,6 +561,67 @@ public static Map<String, String> createDeltaTableCredProps(
...
+   *     automatically refreshes credentials before expiry; when {@code false}, embeds the initial
+   *     credentials as static keys.
+   * @param credScopedFsEnabled when {@code true}, overrides {@code fs.<scheme>.impl} with
+   *     CredScopedFileSystem so that filesystem instances are reused per credential scope rather
+   *     than created anew for every file access.
+   * @param hadoopConf the engine's existing Hadoop configuration, used to read any previously
+   *     configured {@code fs.<scheme>.impl} values before they are overridden by
+   *     CredScopedFileSystem.
+   */
+  public static Map<String, String> createDeltaStagingTableCredProps(
```
yili-db, line 578: It looks like the framework of all 4 `createXCredProps` functions are the same, except that each version has a customized way to specify the "identity" when `renewCredEnabled==true`:

- In this function it's `.deltaStagingTableId(stagingTableId, location)` for all three clouds
- In `createTableCredProps` it's `.credentialType(UCHadoopConfConstants.UC_CREDENTIALS_TYPE_TABLE_VALUE).tableId(tableId).tableOperation(tableOp)` etc.

So it appears that they can be made a "Base" function that does this framework for all 3 clouds, then each `createXCredProps` function provide a lambda that sets their customized keys when `renewCredEnabled==true`?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
openinx: yes,  the `guava`'s Preconditions.checkNotNull is made to return the checked value, but here we are using our `unitycatalog-client` 's `Preconditions, which just return the `void` value.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
```diff
@@ -216,6 +216,37 @@ public Map<String, String> buildForTable(
...
+     *
+     * @param stagingTableId the staging table UUID
+     * @param location the staging table storage location, used to select the matching storage
+     *     credential from the response
+     * @return unmodifiable map; empty if the scheme is unrecognized
+     * @throws IllegalArgumentException if {@code stagingTableId}, {@code location}, or {@code
+     *     tokenProvider} is missing
+     * @throws ApiException if the credential fetch from the UC Delta API fails
+     */
+    public Map<String, String> buildForStagingTable(String stagingTableId, String location)
```
yili-db, line 231: This function will be called as the entry point right after creating a staging table, right?
Then the caller, which holds a `StagingTableResponse`, has the `StagingTableResponse.storageCredentials` already so that it should reuse the returned credential without sending another call in `fetchDeltaStagingTableCredProps`

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
```diff
@@ -0,0 +1,45 @@
...
+    this.api = api;
+    String rawId = require(conf, UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_ID_KEY);
+    this.stagingTableId = UUID.fromString(rawId);
+    this.location = require(conf, UCHadoopConfConstants.UC_DELTA_LOCATION_KEY);
```
openinx: Hi @yili-db ,  Here I think we should use the `UC_DELTA_STAGING_TABLE_LOCATION_KEY`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -119,6 +121,16 @@ public T ucDeltaTableIdentifier(UCDeltaTableIdentifier identifier, String locati
       return self();
     }
 
+    public T deltaStagingTableId(String stagingTableId, String location) {
```
openinx: nit:  change to use the `ucDeltaStagingTableId`,  align with the `ucDeltaTableIdentifier`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -119,6 +121,16 @@ public T ucDeltaTableIdentifier(UCDeltaTableIdentifier identifier, String locati
...
+      Preconditions.checkState(
+          !builder.containsKey(UCHadoopConfConstants.UC_TABLE_ID_KEY)
+              && !builder.containsKey(UCHadoopConfConstants.UC_DELTA_CATALOG_KEY),
+          "deltaStagingTableId cannot be set with tableId or UC Delta table identifier");
+      builder.put(UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_ID_KEY, stagingTableId);
+      builder.put(UCHadoopConfConstants.UC_DELTA_LOCATION_KEY, location);
```
openinx: use the `UC_DELTA_STAGING_TABLE_LOCATION_KEY`, rather than the `UC_DELTA_LOCATION_KEY`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConfConstants.java
openinx: Hi @yili-db ,  I think we'd better to explicit those keys to differentiate from the normal table,  since the CredScopedFS fully relies on those keys to cache the filesystems.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialFetcher.java
openinx: Done — the routing is now:
```
if (useDeltaCredentialsApi) {
  if (UC_TABLE_ID_KEY is set) → staging table
  else → regular Delta table
}
```
Exactly as you suggested.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java
openinx: Good catch! Added an overload `buildForStagingTable(String stagingTableId, String location, List<StorageCredential> storageCredentials)` that accepts the pre-fetched credentials directly (e.g. from `StagingTableResponse.getStorageCredentials()`). It delegates to `createDeltaStagingTableCredProps` and skips the network call, while still configuring credential renewal for when the initial creds expire.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
openinx: yes, make sense.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
openinx: I think that could be a follow-up,  which is not a blocker for this PR.

yili-db (approved):

PR #1571, opened 2026-05-26
[Hadoop] Default GCS conflict check to false for downscoped token compatibility

seewishnew:
**PR Checklist**

**Description of changes**

## Problem

When using Unity Catalog with a GCP workspace, UC vends **downscoped GCS OAuth tokens** scoped to a specific table's storage path prefix (e.g., `gs://bucket/.../tables/<uuid>/`). The GCS connector's `create()` method calls `checkNoFilesConflictingWithDirs()`, which stats every ancestor directory. These ancestor paths fall outside the token's scope, so each stat returns **403 Forbidden**, breaking all Delta writes.

The root cause: `CredPropsUtil.GcsPropsBuilder` hardcoded `fs.gs.create.items.conflict.check.enable=true` as a per-table property, ignoring any user-provided Hadoop/Spark configuration.

## Fix

Read the user's Hadoop configuration for `fs.gs.create.items.conflict.check.enable` instead of hardcoding `true`, and default to `false`. This makes downscoped tokens work out of the box. Users with broader GCS credentials who want the conflict check can opt back in by setting the key to `true` in their Hadoop/Spark config.

This is safe because Delta and Iceberg handle file/directory conflicts at the commit protocol layer, making the filesystem-level check redundant for table formats that UC supports.

## Changes

- **`CredPropsUtil.java`**: Extracted `GCS_CONFLICT_CHECK_KEY` constant; changed `GcsPropsBuilder` to read from `hadoopConf` (default `"false"`) instead of hardcoding `"true"`.
- **`CredPropsUtilTest.java`**: Added 4 test methods covering:
  - Default is `"false"` with an empty config
  - User-set `"true"` is respected (opt-in)
  - `createPathCredProps` and `createDeltaTableCredProps` honor the default
  - `renewCredEnabled=true` (renewal code path) also defaults to `"false"`

## How was this tested

- **Unit tests:** `sbt "hadoop/testOnly io.unitycatalog.hadoop.internal.CredPropsUtilTest"` — 32 tests pass (28 existing + 4 new)
- **Manual end-to-end:** Spark 4.1.0 + Delta 4.2.0 + patched UC 0.5.0-SNAPSHOT against a GCP workspace with downscoped GCS tokens. Verified `INSERT INTO` succeeds where it previously failed with 403.

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -57,6 +57,7 @@ public class CredPropsUtil {
   private static final String GCS_ACCESS_TOKEN_KEY = "fs.gs.auth.access.token.credential";
   private static final String GCS_ACCESS_TOKEN_EXPIRATION_KEY =
       "fs.gs.auth.access.token.expiration";
+  private static final String GCS_CONFLICT_CHECK_KEY = "fs.gs.create.items.conflict.check.enable";
   private static final String ABFS_FIXED_SAS_TOKEN_KEY = "fs.azure.sas.fixed.token";
 
   private abstract static class PropsBuilder<T extends PropsBuilder<T>> {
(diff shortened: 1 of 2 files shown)


timothyw553 (approved):

yili-db (approved):

openinx (approved):
LGTM, thanks for the finding.

PR #1574, opened 2026-05-28
[Build] Support Spark commit-based builds

murali-db:
## Summary
- Add an opt-in `sparkCommit` hook that resolves UC Spark dependencies from locally built Spark artifacts for an exact SHA.
- Add a manual Spark Source Artifact Cache workflow that defaults to the same Spark source ref used by Delta, while allowing maintainers to pass another Spark branch, tag, or commit.
- Update the existing non-blocking `Delta master-SNAPSHOT + Spark 4.2.0-SNAPSHOT` Unit Tests lane to consume the cached source-built Spark artifacts instead of adding a new Delta-master check.
- Mark the source-built Spark compatibility spec as `4.2.0-SNAPSHOT` so release tooling treats it as pre-release/source-built.

## Test plan
- `bash -n project/scripts/build_spark_from_source.sh`
- `bash -n project/scripts/build_delta_from_source.sh`
- Parsed `.github/workflows/spark_commit_tests.yml` and `.github/workflows/unit-tests.yml` with PyYAML.
- `git diff --check`
- `build/sbt -DsparkVersion=4.2.0-SNAPSHOT -DsparkCommit=b6bd005ac7549411ec4e7dc944d7a0e19fd56561 -DsparkArtifactVersion=4.2.0-b6bd005ac754-SNAPSHOT "show spark/sparkVersion" "show spark/libraryDependencies"` shows `spark/sparkVersion = 4.2.0-SNAPSHOT` and `org.apache.spark:spark-sql:4.2.0-b6bd005ac754-SNAPSHOT:provided`.
- Verified release metadata treats only `4.0.0` and `4.1.0` as release-like versions and skips `4.2.0-SNAPSHOT`.
- Verified changed files contain no internal Maven/proxy references.

Full Spark source build/test was not run locally because it requires building Apache Spark from source.

### DIFF
--- a/.github/workflows/spark_commit_tests.yml
+++ b/.github/workflows/spark_commit_tests.yml
@@ -1,4 +1,4 @@
-name: "UC Spark Source Ref"
+name: "Spark Source Artifact Cache"
 
 on:
   workflow_dispatch:
@@ -11,25 +11,15 @@ on:
       spark-version:
         description: "Spark compatibility line used for UC and Delta shims"
         required: true
-        default: "4.2"
-        type: string
-      delta-repo:
-        description: "Delta repository to build for the UC test"
-        required: true
-        default: "https://github.com/delta-io/delta.git"
-        type: string
-      delta-ref:
-        description: "Delta ref to build for the UC test"
-        required: true
-        default: "master"
+        default: "4.2.0-preview5"
         type: string
 
 permissions:
   contents: read
 
 jobs:
-  test:
-    name: "UC against Spark ${{ inputs['spark-source-ref'] }}"
+  build-cache:
+    name: "Cache Spark ${{ inputs['spark-source-ref'] }}"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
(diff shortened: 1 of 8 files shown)


project/spark-versions.json
```diff
@@ -3,6 +3,7 @@
   "versions": [
     {"version": "4.0.0", "sourceDir": "scala-shims/spark-4.0"},
     {"version": "4.1.0", "sourceDir": "scala-shims/spark-4.1"},
-    {"version": "4.2.0-preview5", "sourceDir": "scala-shims/spark-4.2"}
+    {"version": "4.2.0-preview5", "sourceDir": "scala-shims/spark-4.2"},
+    {"version": "4.2.0", "sourceDir": "scala-shims/spark-4.2", "requiresSparkCommit": true}
```
seewishnew: This is critical for our release process: we need to qualify these source-built refs of Spark with preview or snapshot suffix here. This is not just for visual clarity, but also for our release pipelines to ignore these artifacts of UC compiled against the pre-release versions of Spark.

seewishnew: Same goes for Delta; but I think over there we're piggybacking off of the preview target's name. Worth double-checking all the same.

.github/workflows/spark_commit_tests.yml
```diff
@@ -0,0 +1,79 @@
...
+          java-version: "17"
+          distribution: "zulu"
+          cache: "sbt"
+
+      - name: Build Spark source ref locally
+        id: spark-build
+        env:
+          SPARK_SOURCE_REF: ${{ inputs['spark-source-ref'] }}
+          SPARK_VERSION: ${{ inputs['spark-version'] }}
+        run: bash project/scripts/build_spark_from_source.sh
```
seewishnew: We're not caching this build?

project/spark-versions.json
murali-db: Done. I renamed the source-built Spark spec from the release-looking `4.2.0` to `4.2.0-SNAPSHOT` with `requiresSparkCommit=true`, so release tooling can treat it as pre-release/source-built and ignore it. The derived source artifact still strips the snapshot base and resolves to `4.2.0-<sha>-SNAPSHOT`.

murali-db: Checked. Delta does not add a plain release-looking source-built version to a release metadata file; it uses the existing 4.2 compatibility lane and overlays the commit-qualified artifact version only for source-built runs. UC now follows the same release-safety principle by naming the source-built spec `4.2.0-SNAPSHOT`.

.github/workflows/spark_commit_tests.yml
murali-db: Done. The manual workflow is now a Spark Source Artifact Cache producer: it restores/saves `~/.m2/repository/org/apache/spark` keyed by runner, Scala line, Spark compatibility version, resolved Spark SHA, and the Spark helper hash. The existing non-blocking Delta master + Spark 4.2 lane consumes that cache.

murali-db:
Superseded by the split stack:

- #1590 adds the Spark source artifact cache producer workflow.
- #1591 consumes the cached Spark artifacts in the existing non-blocking Unit Tests lane.

This keeps the UC shape aligned with the Delta split.

PR #1579, opened 2026-05-31
Update UC catalog-managed Delta contract

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1579/files) to review incremental changes.
- [**stack/table_property_contract**](https://github.com/unitycatalog/unitycatalog/pull/1579) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1579/files)]

---------
**PR Checklist**

**Description of changes**
Update UC catalog-managed Delta contract

Features:
- Move deletionVectors from suggested to required.

Suggested properties added:
- delta.parquet.compression.codec = "zstd"
- delta.randomizeFilePrefixes = "true"
- delta.columnMapping.mode = "name"
- delta.columnMapping.maxColumnId (engine-generated)

Required properties added:
- delta.checkpoint.writeStatsAsStruct = "true"
- delta.checkpoint.writeStatsAsJson = "true"

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
@@ -65,10 +65,7 @@ public final class UcManagedDeltaContract {
       Map.of(
           TableProperties.ENABLE_DELETION_VECTORS, "true",
           TableProperties.CHECKPOINT_POLICY, "v2",
-          TableProperties.ENABLE_IN_COMMIT_TIMESTAMPS, "true",
-          TableProperties.PARQUET_COMPRESSION_CODEC, "zstd",
-          TableProperties.CHECKPOINT_WRITE_STATS_AS_STRUCT, "true",
-          TableProperties.CHECKPOINT_WRITE_STATS_AS_JSON, "true");
+          TableProperties.ENABLE_IN_COMMIT_TIMESTAMPS, "true");
 
   /**
    * Required properties whose value the engine must compute at commit time. The staging response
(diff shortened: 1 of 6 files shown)


server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
```diff
@@ -68,8 +63,11 @@ private UcManagedDeltaContract() {}
    */
   public static final Map<String, String> REQUIRED_FIXED_PROPERTIES =
       Map.of(
+          TableProperties.ENABLE_DELETION_VECTORS, "true",
           TableProperties.CHECKPOINT_POLICY, "v2",
-          TableProperties.ENABLE_IN_COMMIT_TIMESTAMPS, "true");
+          TableProperties.ENABLE_IN_COMMIT_TIMESTAMPS, "true",
```
openinx, line 68: Are those 3 properties duplicated with the table features ? what are the differences ?

server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
```diff
@@ -87,10 +85,13 @@ private UcManagedDeltaContract() {}
 
   private static Map<String, String> buildSuggestedProperties() {
     Map<String, String> props = new HashMap<>();
-    props.put(TableProperties.ENABLE_DELETION_VECTORS, "true");
+    props.put(TableProperties.COLUMN_MAPPING_MODE, "name");
+    props.put(TableProperties.COLUMN_MAPPING_MAX_COLUMN_ID, null);
```
openinx, line 89: what does the value `null` mean here ?  means there is no `COLUMN_MAPPING_MAX_COLUMN_ID` ?

openinx: My real question is :  how should the client interpret the `null` ?

openinx (commented):
Left few comments,  thanks @yili-db .

server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
yili-db: As explained here:
```
        suggested-properties:
          type: object
          description: Table properties that should be set whenever possible; null values mean client generates the value
```
`null` means that it's up to the engine to generate a value and send to server.

"Please generate a COLUMN_MAPPING_MAX_COLUMN_ID and send to me if you can. It should be the reasonable value if you know how to. But I don't really know how many columns you are going to have yet so I can't dictate the value here."

server/src/main/java/io/unitycatalog/server/service/delta/UcManagedDeltaContract.java
yili-db: They aren't dup. Delta has a protocol that lists all the features that a table can support. But these properties are the real switches saying "yes I actually use DV, or ICT etc".
It's just the Delta protocol.

tdas (approved):

PR #1580, opened 2026-05-31
chore: make tarball/dist launch relocatable; stage jars per component

roeap:
This PR prepares improvements to our docker build. The naive approach to generating the tarball inside docker lead to errors due to ordering of classpath and then got dirty doing sorting during build etc.

@tdas @yili-db - it looks like the test failures are due to upstream delta? Also there seems to be a disabled tarball build test. The invoked script is gone. There seem to be no credentials in the workflow though. Should we try and re-enable this?

in #1581 (based on this) we enable tarball tarball tests in the CI again.

There are some other improvements, so here is AI to tell you what happened :):

## Summary

First of a 4-PR stack that reworks the Docker build. This PR is standalone (no Docker changes yet) and de-risks the rest.

The tarball launch path in `bin/start-uc-server` and `bin/uc` previously read a generated `classpath` file containing **absolute host paths** into the Coursier cache. That forced any consumer (notably the Docker image) to reproduce those exact paths — the root cause of the current image shipping the entire dependency cache.

This PR launches the server/CLI with a JVM classpath wildcard over a **per-component** jars directory (`-cp jars/server/*`, `-cp jars/cli/*`). The JVM expands the glob at startup, so the layout is fully relocatable and no classpath file is needed.

Server and CLI classpaths are staged into **separate** directories on purpose: the server needs `antlr4-runtime` 4.13.x (via Hibernate) while the CLI pulls in old ANTLR 4.5.1 (via `hadoop-client`). A single merged dir let the wildcard pick the wrong ANTLR and crash server startup (`InvalidClassException: Could not deserialize ATN`). Per-module dirs are each internally consistent.

`Tarball.scala` gains a `stageDist` task producing the relocatable `jars/{server,cli}/ bin/ etc/` tree without tarring (so the Docker build can `COPY` it directly); `createTarball` reuses it and no longer writes the absolute-path classpath file.

## Stack
1. **This PR** — relocatable launchers + tarball staging
2. Rebuild server Dockerfile (glibc multi-stage)
3. CI: publish to ghcr.io keylessly + cosign
4. Distroless production image

Validated end-to-end in a fork (build, multi-arch publish, server boots and serves).

AI-assisted by: Isaac

### DIFF
--- a/bin/start-uc-server
+++ b/bin/start-uc-server
@@ -41,8 +41,10 @@ run_in_source_code() {
 
 # This function runs the server using the tarball
 run_in_tarball() {
-  SERVER_CLASSPATH_FILE=$(find "$TARBALL_JAR_DIR" -name "classpath" | head -n 1)
-  run_uc_server_java_command "$SERVER_CLASSPATH_FILE" "$@"
+  # Relocatable classpath: a wildcard over the server jars dir (server and cli
+  # jars are staged separately; see project/Tarball.scala). Quote the glob so the
+  # shell leaves it for the JVM to expand.
+  java -cp "$TARBALL_JAR_DIR/server/*" io.unitycatalog.server.UnityCatalogServer "$@" || exit
 }
 
 # Check if TARBALL_JAR_DIR exists, then we are running in the tarball
--- a/bin/uc
+++ b/bin/uc
@@ -44,9 +44,10 @@ run_in_source_code() {
 
 # This function runs the CLI using the tarball
 run_in_tarball() {
-  CLI_CLASSPATH_FILE=$(find "$TARBALL_JAR_DIR" -name "classpath" | head -n 1)
-
-  run_cli_java_command "$CLI_CLASSPATH_FILE" "$@"
+  # Relocatable classpath: a wildcard over the cli jars dir (server and cli jars
+  # are staged separately; see project/Tarball.scala). Quote the glob so the
+  # shell leaves it for the JVM to expand.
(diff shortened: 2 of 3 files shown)


yili-db:
One thing worth catching in this PR: `docs/server/deployment.md` still references the `jars/classpath` file that this change removes.

Both the MySQL and PostgreSQL setup steps say:

> Modify the `jars/classpath` file and add path to your JDBC driver.

- MySQL: https://github.com/unitycatalog/unitycatalog/blob/9a1fcc9c07fd7a741df1b4adf3f6af597b539d61/docs/server/deployment.md?plain=1#L82
- PostgreSQL: https://github.com/unitycatalog/unitycatalog/blob/9a1fcc9c07fd7a741df1b4adf3f6af597b539d61/docs/server/deployment.md?plain=1#L109

After this PR there's no `classpath` file anymore, and the launchers expand a JVM wildcard over `jars/server/*`. So the new instruction should be something like:

> Copy your JDBC driver `.jar` into the `jars/server/` directory (the launcher picks up all jars in that dir via the classpath wildcard).

Two notes for the doc wording:

- It needs to be an actual `.jar` dropped into `jars/server/`, since the JVM wildcard only matches `*.jar` in that directory. The old "edit a text file and add an arbitrary path" approach is gone.
- The driver only needs to be in `jars/server/` for the server to find it (not `jars/cli/`).

Could we update those two lines in this PR? Otherwise external JDBC driver users following the docs will be broken after the change lands.

PR #1586, opened 2026-06-01
[Hadoop] Use the general CredId as the global cache key rather than the uuid generated per query.

openinx:
## Summary

This PR stores the initial vended credential in the driver-side global credential cache and switches the cache key from a random per-build UID to the credential **scope** (`CredId`).

- **Cache initial credentials globally**: the credential initialized from `initGenericCredential` is now placed into `GenericCredentialProvider.globalCache`, so it can be reused across Spark queries instead of being re-fetched.
- **Key the cache by `CredId`**: the cache was previously keyed by `UC_CREDENTIALS_UID_KEY`, a fresh `UUID.randomUUID()` generated on every props build. That made each provisioning event a distinct entry, so credentials were never actually reused. We now key by the credential scope (`CredId`) — mirroring how `CredScopedFileSystem` already keys its `FileSystem` cache — so requests targeting the same resource share a vended credential.

`UC_CREDENTIALS_UID_KEY` is removed entirely, including the construction-time UID validation in `GenericCredentialProvider`. Misconfiguration now surfaces lazily when the fetcher requires the UC URI at credential-access time.

> Rebased onto the latest `main`. As part of the rebase, the code was adapted to the renamed UC Delta client types (`TemporaryCredentialsApi` → `DeltaTemporaryCredentialsApi`, `CredentialOperation` → `DeltaCredentialOperation`, `CredentialsResponse` → `DeltaCredentialsResponse`, `StorageCredential[Config]` → `DeltaStorageCredential[Config]`).

## Test plan

- [x] `sbt hadoop/test` — 140 tests pass.
- [x] `sbt hadoop/Test/compile` (checkstyle) — no issues.
- [x] The existing `testGlobalCredCache` test still asserts 4 distinct cache entries, now satisfied by 4 distinct `CredId`s (tableA/tableB/pathA/pathB) rather than 4 random UIDs.


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/CredId.java
```diff
@@ -0,0 +1,106 @@
...
+ *       table-level temporary credentials via the UC Delta credentials API.
+ *   <li>{@link DeltaStagingTableCredId} — keyed by staging table ID and location; used for
+ *       staging-table-level temporary credentials via the UC Delta credentials API.
+ *   <li>{@link PathCredId} — keyed by path and operation; used for path-level temporary
+ *       credentials.
+ *   <li>{@link DefaultCredId} — keyed by URI scheme and authority; used as a fallback when no Unity
+ *       Catalog credential type is present in the configuration.
+ * </ul>
+ */
+public interface CredId {
```
openinx, line 46: Renamed the `CredScopedKey` to `CredId`, since we are going to use this `CredId` as the general cache key for both CredScopedFileSystem and `globalCache` in `GenericCredentialsProvider`. 

It make more sense for us to define it as a short `CredId`, rather than the `CredScopedKey`. Also as other places will use the subclasses of `CredId` directly after this PR, so we move those subclasses out of the interface, so that any usage/reference can directly use the subclass name, rather than always having the redundant `CredId.DeltaTableCredId`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/DefaultCredId.java
```diff
@@ -0,0 +1,53 @@
...
+      this.scheme = defaultUri.getScheme();
+      this.authority = defaultUri.getAuthority();
+    } else {
+      this.scheme = uri.getScheme();
+      this.authority = uri.getAuthority();
+    }
+  }
+
+  @Override
+  public Map<String, String> props() {
```
openinx, line 29: For this `DefaultCredId`,  we are not relying on the `props` to construct the `DefaultCredId`.  since the `CredPropsUtil` will never use this `DefaultCredId`, so it's fine to set the empty map here. 

the only place that will use the `DefaultCredId` , is the CredScopedFileSystem here.

openinx: https://github.com/unitycatalog/unitycatalog/pull/1586/changes#diff-8a05225861fa66d9440f3b86c0ad8e6dfa4ce66327149773c4738c33ae16e76bR87

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaGenericCredentialFetcher.java
```diff
@@ -6,54 +6,40 @@
...
-  private final String schema;
-  private final String tableName;
-  private final String location;
 
-  UCDeltaGenericCredentialFetcher(Configuration conf, DeltaTemporaryCredentialsApi api) {
-    Preconditions.checkNotNull(api, "api is required");
-    this.api = api;
-    String rawOp = require(conf, UCHadoopConfConstants.UC_TABLE_OPERATION_KEY);
-    DeltaCredentialOperation op = DeltaCredentialOperation.fromValue(rawOp);
+  UCDeltaGenericCredentialFetcher(DeltaTableCredId credId, DeltaTemporaryCredentialsApi api) {
```
openinx, line 18: Here we use the concrete CredId impl rather than the hadoop conf, will greatly simplify the code. and we make the entire `*CredentialFetcher` don't need to perceive the hadoop conf properties,  which make the code more cohesive.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
```diff
@@ -16,20 +15,20 @@ final class UCDeltaStagingTableCredentialFetcher implements GenericCredentialFet
...
 
-  UCDeltaStagingTableCredentialFetcher(Configuration conf, DeltaTemporaryCredentialsApi api) {
-    Preconditions.checkNotNull(api, "Temporary credentials API is required");
-    this.api = api;
-    this.stagingTableId =
-        UUID.fromString(require(conf, UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_ID_KEY));
-    this.stagingTableLocation =
-        require(conf, UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_LOCATION_KEY);
+  UCDeltaStagingTableCredentialFetcher(
+      DeltaStagingTableCredId credId, DeltaTemporaryCredentialsApi api) {
```
openinx, line 19: Same as the above.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
```diff
@@ -47,11 +48,9 @@ protected void initialize(Configuration conf) {
...
             UCHadoopConfConstants.UC_RENEWAL_LEAD_TIME_DEFAULT_VALUE);
 
-    this.credUid = conf.get(UCHadoopConfConstants.UC_CREDENTIALS_UID_KEY);
-    Preconditions.checkState(
-        credUid != null && !credUid.isEmpty(),
-        "Credential UID cannot be null or empty, '%s' is not set in hadoop configuration",
-        UCHadoopConfConstants.UC_CREDENTIALS_UID_KEY);
+    // Identify the credential scope; used as the global cache key so that requests targeting the
+    // same scope can share a vended credential.
+    this.cacheKey = CredId.create(conf);
```
timothyw553, line 53: we never put the initial credential into `globalCache` here, only into `this.credential`, so other providers for the same `CredId` can't reuse it. could we add `globalCache.put(cacheKey, credential)`?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
```diff
@@ -16,20 +15,20 @@ final class UCDeltaStagingTableCredentialFetcher implements GenericCredentialFet
...
+    this.stagingTableId = UUID.fromString(credId.stagingTableId());
+    this.stagingTableLocation = credId.location();
   }
 
   @Override
   public GenericCredential createCredential() throws ApiException {
     DeltaCredentialsResponse response = api.getStagingTableCredentials(stagingTableId);
-    Preconditions.checkArgument(
-        response != null,
+    Preconditions.checkNotNull(
```
timothyw553, line 31: the sibling fetcher throws `IllegalArgumentException` for this same case — should we match it instead of `NullPointerException`?

timothyw553 (commented):
a few comments

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/CredId.java
```diff
@@ -0,0 +1,112 @@
...
+
+  Supplier<CredId> NO_FALLBACK =
+      () -> {
+        throw new IllegalArgumentException(
+            "Cannot recognize the correct hadoop config to initialize the CredId.");
```
timothyw553: could we add the config type/keys we actually saw? hard to debug otherwise.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialFetcher.java
```diff
@@ -21,43 +24,54 @@
...
-            UCHadoopConfConstants.UC_DELTA_CREDENTIALS_API_ENABLED_DEFAULT_VALUE);
-    if (useDeltaCredentialsApi) {
-      DeltaTemporaryCredentialsApi deltaApi = new DeltaTemporaryCredentialsApi(apiClient);
-      String stagingTableId = conf.get(UCHadoopConfConstants.UC_DELTA_STAGING_TABLE_ID_KEY);
-      if (stagingTableId != null && !stagingTableId.isEmpty()) {
-        return forUcDeltaStagingTable(conf, deltaApi);
-      } else {
-        return forUcDelta(conf, deltaApi);
-      }
+  static GenericCredentialFetcher create(ApiClient apiClient, CredId credId) {
```
timothyw553, line 57: this repeats the routing `CredId.create` already does. should the API selection live on `CredId` instead?

openinx: Hi @timothyw553 ,  I think they are different because: 
1.  here we are use the `CredId` to select the correct UC API to call. 
2. while in the `CredId.create`, we are constructing the `CredId` based on the conf, so that we can get the produce the correct `CredId` instance. 

the two selection should be irrelevant.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
openinx: yes,  this PR actually is trying to address the foundemental infra problem, I mean use the `credKey` as the key to cache the credentials, rather than the credUid issued from the `loadTable`. 

Adding the initial credential into the global cache will be a separate PR to address it, which I don't want to overcomplicate this foundermental PR.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
openinx: I checked all the *Fetcher,  all are using `checkNotNull` now.  in this context,  i think it make more sense to use `checkNotNull` here, since we are checking a returned value from a api, not the passing argument from the method.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/CredId.java
openinx: let me think about how to add the context here.

openinx: the comment looks good to me.

timothyw553 (approved):
Looks good for the main logic. The bulk is tests and refactoring byproduct

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCGenericCredentialFetcher.java
```diff
@@ -8,57 +8,43 @@
...
+    Preconditions.checkNotNull(credId, "credId is required");
     Preconditions.checkNotNull(api, "api is required");
-    this.api = api;
-    String type = conf.get(UCHadoopConfConstants.UC_CREDENTIALS_TYPE_KEY);
-    Preconditions.checkArgument(
-        UCHadoopConfConstants.UC_CREDENTIALS_TYPE_PATH_VALUE.equals(type)
-            || UCHadoopConfConstants.UC_CREDENTIALS_TYPE_TABLE_VALUE.equals(type),
-        "Unsupported unity catalog temporary credentials type '%s', please check '%s'",
-        type,
-        UCHadoopConfConstants.UC_CREDENTIALS_TYPE_KEY);
```
openinx, line 29: Those checks already moved to the `CredId.create`,  so there is no need to keep here now.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCGenericCredentialFetcher.java
```diff
@@ -8,57 +8,43 @@
...
-      this.tableRequest = null;
-    } else {
-      this.pathRequest = null;
-      this.tableRequest =
-          new GenerateTemporaryTableCredential()
-              .tableId(require(conf, UCHadoopConfConstants.UC_TABLE_ID_KEY))
-              .operation(
-                  TableOperation.fromValue(
-                      require(conf, UCHadoopConfConstants.UC_TABLE_OPERATION_KEY)));
-    }
```
openinx, line 46: After we use the `CredId`, it actually greatly simplifies the code here.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedKey.java
```diff
@@ -1,248 +0,0 @@
...
- *       for table-level temporary credentials via the UC Delta credentials API.
- *   <li>{@link DeltaStagingTableCredScopedKey} — keyed by staging table ID and location; used for
- *       staging-table-level temporary credentials via the UC Delta credentials API.
- *   <li>{@link PathCredScopedKey} — keyed by path and operation; used for path-level temporary
- *       credentials.
- *   <li>{@link DefaultCredScopedKey} — keyed by URI scheme and authority; used as a fallback when
- *       no Unity Catalog credential type is present in the configuration.
- * </ul>
- */
-public interface CredScopedKey {
```
openinx, line 43: This class is replaced with the `CredId`, and its subclasses.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/CredId.java
```diff
@@ -0,0 +1,118 @@
...
+          observed.put(
+              UC_DELTA_CREDENTIALS_API_ENABLED_KEY, conf.get(UC_DELTA_CREDENTIALS_API_ENABLED_KEY));
+          observed.put(UC_DELTA_STAGING_TABLE_ID_KEY, conf.get(UC_DELTA_STAGING_TABLE_ID_KEY));
+
+          throw new IllegalArgumentException(
+              "Cannot recognize the hadoop config to initialize the CredId. Observed: " + observed);
+        });
+  }
+
+  static CredId create(Configuration conf, Supplier<CredId> defaultCredId) {
```
openinx, line 63: The hadoop conf  is coming from the props that set via this CredPropsUtils#credId: https://github.com/unitycatalog/unitycatalog/blob/a1d535e37d81a0bcaae0d99d391a011e74fac01d/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java#L85-L89

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -82,79 +82,9 @@ public T tokenProvider(TokenProvider tokenProvider) {
...
-    public T path(String path) {
-      builder.put(UCHadoopConfConstants.UC_PATH_KEY, path);
-      return self();
-    }
-
-    public T pathOperation(PathOperation pathOp) {
-      builder.put(UCHadoopConfConstants.UC_PATH_OPERATION_KEY, pathOp.getValue());
+    /** Applies the credential-scope identity properties carried by {@code credId}. */
+    public T credId(CredId credId) {
+      credId.props().forEach(this::set);
```
openinx, line 87: After we moved the props inside the CredId subclasses,  actually it greatly simplifies the code here.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -312,7 +242,6 @@ private static S3PropsBuilder s3TempCredPropsBuilder(
             .set(UCHadoopConfConstants.S3A_CREDENTIALS_PROVIDER, AWS_VENDED_TOKEN_PROVIDER_CLASS)
             .uri(uri)
             .tokenProvider(tokenProvider)
-            .uid(UUID.randomUUID().toString())
```
openinx, line 315: we will have a follow up PR to add the UniqueCredId to support the per-query cache, rather than the cross-query in this PR, that help us to still support the per-query case in this release.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConfConstants.java
```diff
@@ -60,11 +60,6 @@ private UCHadoopConfConstants() {}
   // Key for specifying the manual clock, for testing purpose.
   public static final String UC_TEST_CLOCK_NAME = "fs.unitycatalog.test.clock.name";
 
-  // Key representing a unique credential ID. It identifies a job-level credential for a specific
-  // table, meaning that the same job–table combination shares the same credential. Cached
-  // credentials are indexed by this key and are not reused across different jobs.
-  public static final String UC_CREDENTIALS_UID_KEY = "fs.unitycatalog.credentials.uid";
```
openinx, line 66: As we said before in that comment: https://github.com/unitycatalog/unitycatalog/pull/1586/changes#r3515547625

we will have a follow up to add the UniqueCredId to support the per query cache rather than the cross-queries cache.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -467,72 +326,28 @@ private static AbfsPropsBuilder abfsTempCredPropsBuilder(
...
-   * @param credScopedFsEnabled when {@code true}, overrides {@code fs.<scheme>.impl} with
-   *     CredScopedFileSystem so that filesystem instances are reused per credential scope rather
-   *     than created anew for every file access.
-   * @param hadoopConf the engine's existing Hadoop configuration, used to read any previously
-   *     configured {@code fs.<scheme>.impl} values before they are overridden by
-   *     CredScopedFileSystem.
-   */
-  public static Map<String, String> createDeltaTableCredProps(
+  /** Fetches table credentials from the UC REST API and builds Hadoop configuration properties. */
+  public static Map<String, String> createTableCredProps(
```
openinx, line 330: In the new PR, we compact the `fetchTableCredProps` and `createTableCredProps` into a single method `createTableCredProps`, which greatly simplifies the code.  and removed lots of duplicated code.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/CredId.java
```diff
@@ -0,0 +1,118 @@
...
+    String type = conf.get(UC_CREDENTIALS_TYPE_KEY);
+    boolean isDeltaApi =
+        conf.getBoolean(
+            UC_DELTA_CREDENTIALS_API_ENABLED_KEY, UC_DELTA_CREDENTIALS_API_ENABLED_DEFAULT_VALUE);
```
yili-db, line 67: `UC_DELTA_CREDENTIALS_API_ENABLED_DEFAULT_VALUE` is still `false`. Please change it to true.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/DeltaStagingTableCredId.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  public String location() {
+    return location;
+  }
+
+  @Override
+  public Map<String, String> props() {
+    Map<String, String> props = new HashMap<>();
+    props.put(UC_DELTA_CREDENTIALS_API_ENABLED_KEY, "true");
```
yili-db: `Map.of`?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/DeltaTableCredId.java
```diff
@@ -0,0 +1,88 @@
...
+    return tableOperation;
+  }
+
+  public String location() {
+    return location;
+  }
+
+  @Override
+  public Map<String, String> props() {
+    Map<String, String> props = new HashMap<>();
```
yili-db: `Map.of`?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialFetcher.java
```diff
@@ -21,43 +24,54 @@
 public interface GenericCredentialFetcher {
   GenericCredential createCredential() throws ApiException;
 
-  /** Creates a fetcher backed by the standard UC temporary credentials API. */
-  static GenericCredentialFetcher forUc(Configuration conf, TemporaryCredentialsApi api) {
-    return new UCGenericCredentialFetcher(conf, api);
+  /** Creates a fetcher backed by the standard UC temporary credentials API for a table. */
+  static GenericCredentialFetcher forUc(
+      TableCredId credId, io.unitycatalog.client.api.TemporaryCredentialsApi api) {
```
yili-db: No need to use FQCN. It used to have `import io.unitycatalog.client.api.TemporaryCredentialsApi;`

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -467,72 +326,28 @@ private static AbfsPropsBuilder abfsTempCredPropsBuilder(
...
+      String tableId,
+      UCCredentialHadoopConfs.TableOperation tableOp,
+      Map<String, String> appVersions)
+      throws ApiException {
+    TableCredId credId = new TableCredId(tableId, tableOp.value());
+    TemporaryCredentials tempCreds =
+        fetchTemporaryCredentials(apiClient, catalogUri, tokenProvider, appVersions, credId);
     switch (scheme) {
       case "s3":
         if (renewCredEnabled) {
```
yili-db: btw is there any benefit top keep this `renewCredEnabled` switch? Can we always have it enabled?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/CredId.java
openinx: Hi @yili-db ,  we only think this `isDeltaApi` is true if it's explicitly set to `true`, otherwise we need to think it as `false` so that we can fallback to the legacy UC API.  This is important,  because for other non-delta table, we will need this to enable the path-based credential vending. 

For UC Delta API,  those `DeltaStagingTableCredId` and `DeltaTableCredId`  actually already explicitly set to true in their props, so we definitely will use the `isDeltaApi=true` in the following branches. 
1. https://github.com/unitycatalog/unitycatalog/blob/ceb042d1775524da6f1bd9a74d3cbb56aa4eb5d1/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/DeltaTableCredId.java#L53
2. https://github.com/unitycatalog/unitycatalog/blob/ceb042d1775524da6f1bd9a74d3cbb56aa4eb5d1/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/DeltaStagingTableCredId.java#L39

TL;DR, I suggest to keep as it is, so that for non uc managed delta table, we can still correctly build the `PathCredId` without explicitly set the `UC_DELTA_CREDENTIALS_API_ENABLED_KEY` props.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/DeltaStagingTableCredId.java
openinx: okay, i'm fine with either.  seems like `Map.of` naturally make it as immutable. so may be better.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
openinx: I agreed with you, yes, actually it's meaningless to keep the fixed credential for now, since it has the problem which will spawn so many threads/connection pools. 

I will suggest to file a separate PR to remove that, since we are close to get the bugfix release out, and for the bugfix release, it does not make sense to change the user-behaivor.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialFetcher.java
openinx: Yes, make sense.  thanks.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -567,35 +380,31 @@ public static Map<String, String> createDeltaTableCredProps(
...
-      String stagingTableId,
+      UCDeltaTableIdentifier identifier,
       String location,
-      TemporaryCredentials tempCreds,
-      Map<String, String> appVersions) {
+      UCCredentialHadoopConfs.TableOperation tableOp,
+      Map<String, String> appVersions)
+      throws ApiException {
+    DeltaTableCredId credId = new DeltaTableCredId(identifier, tableOp.value(), location);
+    TemporaryCredentials tempCreds =
```
yili-db: Now it seems the only difference between different versions of `create*CredProps` is `DeltaTableCredId credId = new DeltaTableCredId(identifier, tableOp.value(), location);`
Everything else starting from `fetchTemporaryCredentials` and `switch (scheme) {` is exactly the same? More dedup can be done?

openinx: Great finding.

openinx: Yes, make sense.  I'd love to dedup more.

openinx: Done.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialFetcher.java
openinx: Done.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/DeltaTableCredId.java
openinx: Done.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/DeltaStagingTableCredId.java
openinx: Done.

yili-db (approved):

PR #1590, opened 2026-06-02
[Build] Test source-built Spark automatically in CI

murali-db:
## Summary
- Makes UC's Unit Tests workflow test source-built Spark automatically on every PR, building Spark from a pinned ref and caching it so normal runs stay fast.
- Removes the separate manual Spark cache-warmer workflow; the Spark build is now an inline, build-on-cache-miss step in the test path.
- Keeps the test matrix as a clear inline definition in `unit-tests.yml` (same combos as before), with the 4.2 lane switched from published `4.2.0-preview5` to source-built `4.2.0-SNAPSHOT` and a `source-build` flag per row.
- Supersedes #1591; the useful consumer changes are folded into this PR.

## How this works
- The test matrix is inline YAML. Each row carries `spark-version`, `delta-version`, `validation-mode`, `non-blocking`, and `source-build`. Published rows set `source-build: false`; the `4.2.0-SNAPSHOT` row sets `source-build: true`.
- Two composite actions:
  - `.github/actions/prepare-spark-env` prepares the environment: sets up the JDK and, for `source-build: true`, resolves the Spark SHA/cache key, restores `~/.m2/repository/org/apache/spark`, builds Spark via `project/scripts/build_spark.sh` and saves the cache on a miss, then verifies the Spark Maven artifacts.
  - `.github/actions/run-uc-unit-tests` calls `prepare-spark-env`, then publishes UC locally, builds Delta `master-SNAPSHOT` when needed, compiles integration tests, runs the connector/unit tests, and runs the license check.
- `project/scripts/get_spark_version_info.py` does one thing: `--resolve-source-build` reads the source-build fields from `project/spark-versions.json` (`sourceBuildDefaultRef`) and emits the Spark SHA, the commit-qualified Maven version (`4.2.0-<sha12>-SNAPSHOT`), and an exact Maven cache key.
- Cache keys are exact (no `restore-keys`); when the Spark SHA or `build_spark.sh` changes, a new key is used and older entries are evicted by GitHub's LRU.

## Test plan
- Parsed `.github/workflows/unit-tests.yml` and both composite `action.yml` files with PyYAML.
- `python3 -m py_compile project/scripts/get_spark_version_info.py project/tests/test_cross_spark_publish.py`.
- Ran the source-build metadata + deterministic cache-key tests in `project/tests/test_cross_spark_publish.py`.
- Confirmed the manual `spark_commit_tests.yml` warmer is removed.
- Full CI on this PR exercises the published rows and the non-blocking source-built Spark 4.2 + Delta `master-SNAPSHOT` row end to end.


project/spark-versions.json
```diff
@@ -3,6 +3,13 @@
   "versions": [
     {"version": "4.0.0", "sourceDir": "scala-shims/spark-4.0"},
     {"version": "4.1.0", "sourceDir": "scala-shims/spark-4.1"},
-    {"version": "4.2.0-preview5", "sourceDir": "scala-shims/spark-4.2"}
+    {"version": "4.2.0-preview5", "sourceDir": "scala-shims/spark-4.2"},
```
tdas: seems like in UC we are doing this other way round compared to the Delta? 
In Delta CrossSparkVersions is source of truth that generates the json.. where as in UC the json is the source of truth and CrossSparkVersions just reads that. 

WHY?

project/scripts/get_spark_version_info.py
```diff
@@ -0,0 +1,150 @@
...
+"""
+
+import argparse
+import json
+import os
+import subprocess
+from pathlib import Path
+
+
+def run_command(args, cwd):
```
tdas, line 20: In the Delta repo, isnt this file more like get_spark_version_info or something that is not tied to just about source building?

.github/workflows/spark_commit_tests.yml
```diff
@@ -0,0 +1,72 @@
+name: "Spark Source Artifact Cache"
```
tdas: what is the purpose of this workflow? wouldnt the workflow that runs the spark tests automatically test whether the spark version build (even if its compiled from source) is working or not?

project/spark-versions.json
tdas: which one is better?

tdas: not a blocker.. dont want to change that in this PR

murali-db: yeah UC already had the json as the source of truth, so I kept that local convention here. agree not worth flipping in this PR.

project/scripts/get_spark_version_info.py
murali-db: done, renamed/generalized this to get_spark_version_info.py so it matches the Delta pattern and is not source-build-only.

.github/workflows/spark_commit_tests.yml
murali-db: yep fair point, removed the manual warmer. CI now restores the exact cache and builds/saves Spark automatically on miss before running the Spark 4.2 test lane.

.github/workflows/unit-tests.yml
```diff
@@ -15,12 +15,71 @@ permissions:
   contents: read
 
 jobs:
+  prepare_spark_source_cache:
```
tdas: please please add comments explaning what is goign on here.

.github/workflows/unit-tests.yml
```diff
@@ -15,12 +15,71 @@ permissions:
...
+    name: Prepare Spark 4.2.0-SNAPSHOT source cache
```
tdas: why is the specific to a particular spark version. what is this not parameterized to any or all spark versions in the spark version json that has a snapshot + commit hash?

.github/workflows/unit-tests.yml
```diff
@@ -15,12 +15,71 @@ permissions:
...
+        echo "cache_key=spark-m2-ubuntu-latest-scala-2.13-${{ steps.spark-source.outputs.spark_version }}-${{ steps.spark-source.outputs.artifact_base_version }}-${{ steps.spark-source.outputs.spark_sha }}-${{ hashFiles('project/scripts/build_spark.sh') }}" >> "$GITHUB_OUTPUT"
+
+    - name: Restore Spark Maven cache
+      id: spark-source-cache
+      uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+      with:
+        path: ~/.m2/repository/org/apache/spark
+        # Exact-match cache only. Spark source builds must not fall back to a
+        # different SHA or artifact base version.
+        key: ${{ steps.spark-cache-key.outputs.cache_key }}
```
tdas: what will happen to previous cache key builds after hash has changed and old build are not relevant any more?

.github/workflows/unit-tests.yml
```diff
@@ -15,12 +15,71 @@ permissions:
...
+        path: ~/.m2/repository/org/apache/spark
+        key: ${{ steps.spark-cache-key.outputs.cache_key }}
+
   test:
     runs-on: ubuntu-latest
 
     # Build a human-readable job name from the matrix entry, e.g.:
     #   "All tests (Spark 4.0.0, Delta 4.2.0)"
-    #   "[Non Blocking] Spark tests (Spark 4.2.0-preview5, Delta master-SNAPSHOT)"
+    #   "[Non Blocking] Spark tests (Spark 4.2.0-SNAPSHOT, Delta master-SNAPSHOT)"
```
tdas, line 39: can remove all these hardcoding of versions and make sure 100s of lines dont need to change everything we do version updates??

.github/workflows/unit-tests.yml
```diff
@@ -104,3 +159,72 @@ jobs:
     - name: Run license check
       if: matrix.validation-mode == 'all-tests'
       run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION licenseCheck
+
+  source-spark-test:
```
tdas: i dont understand why this is outside the test matrix - https://github.com/unitycatalog/unitycatalog/pull/1590/changes#diff-cd0effce06f425599e96952f4d5f684641a683f46d9909e7c74a9f8b2f03ccd3R90-R108

that means you are duplicating all of this code and harding coding it for specific spark verison that have to be updated every time spark version is upgraded.

.github/workflows/unit-tests.yml
```diff
@@ -104,3 +159,72 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+    - name: Set up JDK 17
+      uses: actions/setup-java@c1e323688fd81a25caa38c78aa6df2d33d3e20d9 # v4.8.0
+      with:
+        java-version: '17'
+        distribution: 'zulu'
+        cache: 'sbt'
+
+    - name: Restore prepared Spark Maven cache
```
tdas: havent you already restored the cache above??

.github/workflows/unit-tests.yml
murali-db: Expanded the workflow header to document the four-job split, metadata ownership (spark-versions.json + get_spark_version_info.py), and the cache producer/consumer model.

.github/workflows/unit-tests.yml
murali-db: Source-build Spark versions now come from sourceBuildDefaultRef in spark-versions.json; build-spark-source matrixes over --source-build-spark-versions (no hardcoded 4.2).

.github/workflows/unit-tests.yml
murali-db: Cache keys are exact-match (no restore-keys). When the Spark SHA or build_spark.sh changes, a new key is used; older entries are orphaned and GitHub evicts them by LRU.

.github/workflows/unit-tests.yml
murali-db: Published Spark×Delta combos now live in ciTestMatrix.published in spark-versions.json; the workflow has no version literals.

.github/workflows/unit-tests.yml
murali-db: Source-build rows are generated from ciTestMatrix.sourceBuild + metadata; test-spark-source uses the same matrix-driven composite action as test (./.github/actions/run-uc-unit-tests).

.github/workflows/unit-tests.yml
murali-db: Jobs do not share disk: build-spark-source saves the cache; test-spark-source recomputes the same cache_key from metadata and restores it in its own runner.

.github/workflows/unit-tests.yml
murali-db: Per the follow-up discussion: collapsed this into a single metadata-driven test matrix. `--ci-test-matrix` now emits one flat array over all Spark rows, each tagged `source-build: true|false`. The single `test` job loops over every row and only source-build rows restore the Spark cache (building+saving on a miss); published rows skip the build entirely, so nothing waits behind a source build. Dropped the separate build-spark-source and test-spark-source jobs - no hardcoded Spark version anywhere now.

.github/workflows/unit-tests.yml
```diff
@@ -14,13 +14,47 @@ on:
...
+    name: Generate CI test matrix
     runs-on: ubuntu-latest
+    outputs:
+      test_matrix: ${{ steps.generate.outputs.test_matrix }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - name: Generate CI test matrix from metadata
+        id: generate
+        run: |
+          TEST_MATRIX=$(python3 project/scripts/get_spark_version_info.py --ci-test-matrix)
```
tdas: is this generating the same test matrix as before?

.github/workflows/unit-tests.yml
```diff
@@ -29,78 +63,14 @@ jobs:
...
-      run: build/sbt -DsparkVersion=${{ matrix.spark-version }} -DdeltaVersion=$DELTA_VERSION licenseCheck
+      # Local composite actions require checkout before uses: ./.github/actions/...
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - name: Run UC unit tests
+        uses: ./.github/actions/run-uc-unit-tests
+        with:
+          spark-version: ${{ matrix.spark-version }}
+          delta-version: ${{ matrix.delta-version }}
+          validation-mode: ${{ matrix.validation-mode }}
+          use-spark-source-cache: ${{ matrix.source-build }}
```
tdas, line 85: is this the right cut point to move the code to a custom action? right abstraction between these two yml files? 

challenge me if you think otherwise, but isnt a better split of the code the following
- action file 1 - contains all the code to "prepare" the environment for tests to run with source building and caching stuf
- action file 2 - uses file 1 to prepare the env and then run the tests.

.github/workflows/unit-tests.yml
murali-db: yeah same combos.. killed the generation, it's inline yaml again. only diff is 4.2 is source-built 4.2.0-SNAPSHOT now instead of preview5

.github/workflows/unit-tests.yml
murali-db: done.. split into prepare-spark-env (jvm + build/cache spark) and run-uc-unit-tests (uses it then runs tests)

tdas (approved):

PR #1592, opened 2026-06-02
fix: make uc_delta_commits Postgres-compatible (UUID columns + DELETE…

roeap:
… batch)

**PR Checklist**

**Description of changes**

Drop the H2-specific columnDefinition='BINARY(16)' from DeltaCommitDAO's
UUID columns so Hibernate maps them to Postgres 'uuid' (matching the other
working DAOs); without this the CREATE for uc_delta_commits silently fails
under hbm2ddl=update on Postgres and the table is never created.

Rewrite the two DELETE ... LIMIT native queries in DeltaCommitRepository as
portable 'DELETE ... WHERE id IN (SELECT id ... LIMIT n)' subqueries, which
Postgres accepts (it rejects DELETE ... LIMIT). Batch semantics and the
loop-until-zero callers are unchanged.

Co-authored-by: Isaac

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
@@ -696,7 +696,9 @@ public class DeltaCommitRepository {
   private static int deleteCommitsUpTo(Session session, UUID tableId, long upToCommitVersion) {
     NativeQuery<?> query =
         session.createNativeQuery(
-            "DELETE FROM uc_delta_commits WHERE table_id = :tableId AND commit_version <= :upToCommitVersion LIMIT :numCommitsPerBatch");
+            "DELETE FROM uc_delta_commits WHERE id IN ("
+                + "SELECT id FROM uc_delta_commits WHERE table_id = :tableId "
+                + "AND commit_version <= :upToCommitVersion LIMIT :numCommitsPerBatch)");
     query.setParameter("tableId", tableId);
     query.setParameter("upToCommitVersion", upToCommitVersion);
     query.setParameter("numCommitsPerBatch", NUM_COMMITS_PER_BATCH);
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -696,7 +696,9 @@ private static void saveCommit(Session session, UUID tableId, DeltaCommitInfo co
   private static int deleteCommitsUpTo(Session session, UUID tableId, long upToCommitVersion) {
     NativeQuery<?> query =
         session.createNativeQuery(
-            "DELETE FROM uc_delta_commits WHERE table_id = :tableId AND commit_version <= :upToCommitVersion LIMIT :numCommitsPerBatch");
+            "DELETE FROM uc_delta_commits WHERE id IN ("
```
yili-db, line 699: @roeap 
This new query would break MYSQL.
A simpler fix is to simply remove `LIMIT :numCommitsPerBatch`.
The query `DELETE FROM uc_delta_commits WHERE table_id = :tableId AND commit_version <= :upToCommitVersion` is good enough as the max possible rows are <= 10.

yili-db:
@roeap The `DCO` failure was due to lack of `-s` when `git commit`.
The `Delta master-SNAPSHOT` related failures are expected for now as we haven't fix the tests.

roeap:
@yili-db - added a sign-off, sorry about that.

Any chance you can also have a look at the stack 
- #1580
- #1582
- #1583 

I tried to keep them small. At the end of that satck, we can build a (hopefully new and improved) Docker image. With that I could add some tests to this PR based on a simple compose setup.

yili-db:
https://github.com/unitycatalog/unitycatalog/pull/1446 is addressing the same issue

PR #1601, opened 2026-06-08
[Spark] Fix tests for Delta master + bump CI to Delta 4.2.0

yili-db:
**PR Checklist**

**Description of changes**
[Spark] Fix tests for Delta master + bump CI to Delta 4.2.0

- Bump the blocking CI matrix from Delta 4.1.0 to 4.2.0 (already released).
- Make spark connector tests green on Delta master by version-gating the assertions Delta has changed, rather than weakening them.
- Turn on `server.managed-table.use-delta-api-only=true` on the Delta 4.3+ entry so the matrix exercises the production UC Delta API path.

Before:
- CI runs Delta **4.1** and master (4.3)
- CI **fails** for Delta master

After:
- CI runs Delta **4.2** and master (4.3)
- CI **passes** for Delta master

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
@@ -3,6 +3,7 @@ package io.unitycatalog.spark;
 import static io.unitycatalog.server.utils.TestUtils.CATALOG_NAME;
 import static io.unitycatalog.server.utils.TestUtils.SCHEMA_NAME;
 import static io.unitycatalog.server.utils.TestUtils.createApiClient;
+import static io.unitycatalog.spark.DeltaVersionUtils.MIN_DELTA_VERSION_FOR_UC_DELTA_API;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
@@ -247,8 +248,9 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
   protected final boolean canUpdateColumnsToUC() {
     if (tableFormat().equalsIgnoreCase("DELTA")) {
       // Delta external tables don't update columns to UC yet.
-      // Delta managed tables can update columns starting from Delta 4.2.
-      return isManagedTable() && DeltaVersionUtils.isDeltaAtLeast("4.2.0");
+      // Delta managed tables can update columns starting from Delta 4.3.
+      return isManagedTable()
+          && DeltaVersionUtils.isDeltaAtLeast(MIN_DELTA_VERSION_FOR_UC_DELTA_API);
     }
     return true;
   }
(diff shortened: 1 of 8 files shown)


connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -687,21 +705,21 @@ public void testTableWithSupportedDataTypes() throws ApiException {
                 "CHAR(10)",
                 "'char_test'",
                 "char_test ",
-                ColumnTypeName.CHAR,
-                "char(10)",
+                testingDelta() ? ColumnTypeName.STRING : ColumnTypeName.CHAR,
```
tdas, line 708: is there anything test that the `__CHAR_VARCHAR_TYPE_STRING` is present? shouldnt that be covered in tests the Delta OSS UC integration tests as well?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -846,7 +864,11 @@ public void testTableWithSupportedDataTypes() throws ApiException {
       assertThat(col.getNullable())
           .as("nullable for %s", spec.getName())
           .isEqualTo(spec.isNullable());
-      assertThat(col.getTypeJson())
+      // Delta >= 4.3.0 enables columnMapping by default on managed Delta tables, which injects
+      // `delta.columnMapping.id` and `delta.columnMapping.physicalName` into per-column metadata
+      // that the test fixtures do not carry. The strip is a no-op when those keys are absent
+      // (older Delta and non-Delta), so apply it unconditionally and keep the equality strict.
+      assertThat(stripColumnMappingMetadata(col.getTypeJson()))
```
tdas, line 886: is there any other tests verifying that the expected column mapping fields are present when column mapping is enabled??

without that we have test gaps.

tdas (approved):

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: Yes, implicitly: `charVarcharMetadata("char(10)")` returns `{"__CHAR_VARCHAR_TYPE_STRING":"char(10)"}`, which is embedded into the expected typeJson. The per-column `stripColumnMappingMetadata(col.getTypeJson()).isEqualTo(spec.getTypeJson())` further down would fail if Delta ever stopped emitting the marker (the strip only touches `delta.columnMapping.*` entries). So no separate assertion needed for the marker.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: Good catch -- you're right, the strip silently hides whether the fields are actually present. Added an explicit pre-equality assertion on the managed-Delta-4.3+ path that every column's typeJson contains both `delta.columnMapping.id` and `delta.columnMapping.physicalName`. If Delta ever stops emitting them, this assertion fires before the equality check has a chance to pass on a no-op strip.

Issue #1603, opened 2026-06-09
Fix permission GET endpoints to work with authorization enabled


newfront:
**Describe the bug**

When server.authorization=enable, the GET endpoints on PermissionService were not protected by an @AuthorizeExpression, so the UnityAccessDecorator had no authorization rule to evaluate. This change adds an authentication gate to every permission read endpoint and wires up the resource-key resolution they were missing, then adds an end-to-end integration test suite covering both authorization-enabled and authorization-disabled modes.

**To Reproduce**
Start up Unity Catalog with Authentication enabled and hit `api/2.1/unity-catalog/permissions/{securable}/{name}?principal=...` - it will return a 500

```
=== GET /api/2.1/unity-catalog/permissions/catalog/managed_demo ===
{"error_code":"INTERNAL","details":[{"reason":"INTERNAL","@type":"google.rpc.ErrorInfo"}],"message":"No authorization expression found."}
[HTTP 500]
```

**Expected behavior**
I would expect my own permissions to be returned for a securable when hitting the get endpoints.

**System [please complete the following information]:**

-   OS: e.g. Linux on ECS and Mac OSX

**Additional context**
Nope

PR #1604, opened 2026-06-09
Fix permission GET endpoints to work with authorization enabled

newfront:
**PR Checklist**

**Issue**
https://github.com/unitycatalog/unitycatalog/issues/1603

**Description of changes**

PermissionService.java

Added @AuthorizeExpression to ensure the GET routes would authenticate as they were throwing 500s before when authenticated.

PermissionServiceTest was added to cover the GET permissions route behaviors across metastore, catalog, schema, table, and others.


server/src/main/java/io/unitycatalog/server/service/PermissionService.java
```diff
@@ -80,54 +80,65 @@ public PermissionService(UnityCatalogAuthorizer authorizer, Repositories reposit
 
   // TODO: Refactor these endpoints to use a common method with dynamic resource id lookup
   @Get("/metastore/{name}")
+  @AuthorizeExpression("#principal != null")
```
yili-db: You can make this as a common string const in `src/main/java/io/unitycatalog/server/auth/AuthorizeExpressions.java`.
This string is very short and simple, but the value is that it provides a shared comment to explain why the auth can be skipped:
```
/** get*Authorization endpoints will check the authorization themselves and return result differently according to if principal is owner or not */ 
public static final String GET_RESOURCE_AUTHORIZATION = "#principal != null"
```

server/src/main/java/io/unitycatalog/server/service/PermissionService.java
```diff
@@ -80,54 +80,65 @@ public PermissionService(UnityCatalogAuthorizer authorizer, Repositories reposit
...
   public HttpResponse getMetastoreAuthorization(
       @Param("name") String name) {
     return getAuthorization(METASTORE, name);
   }
 
   @Get("/catalog/{name}")
+  @AuthorizeExpression("#principal != null")
   public HttpResponse getCatalogAuthorization(
-      @Param("name") String name) {
+      @Param("name") @AuthorizeResourceKey(CATALOG) String name) {
```
yili-db: `@AuthorizeResourceKey(` isn't needed here. It's only used by AuthorizeExpression if the expression refers to it.

server/src/test/java/io/unitycatalog/server/service/PermissionServiceTest.java
```diff
@@ -0,0 +1,211 @@
...
+ * <p>The suite runs in two modes via the concrete subclasses below:
+ *
+ * <ul>
+ *   <li>{@link PermissionServiceWithAuthorizationTest} — authorization enabled, so the
+ *       {@code @AuthorizeExpression} rules and token authentication are enforced.
+ *   <li>{@link PermissionServiceWithoutAuthorizationTest} — authorization disabled (the server
+ *       default), so the decorators are not installed and an {@code AllowingAuthorizer} is used.
+ * </ul>
+ */
+public abstract class PermissionServiceTest extends BaseServerTest {
```
yili-db: Use `SdkAccessControlBaseCRUDTest` as base class?

server/src/test/java/io/unitycatalog/server/service/PermissionServiceWithoutAuthorizationTest.java
```diff
@@ -0,0 +1,47 @@
...
+import com.linecorp.armeria.common.AggregatedHttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Runs the {@link PermissionServiceTest} suite with authorization disabled (the server default). No
+ * authentication/authorization decorators are installed and the server uses an {@code
+ * AllowingAuthorizer}, so the endpoints are reachable without a token and grants are no-ops.
+ */
+public class PermissionServiceWithoutAuthorizationTest extends PermissionServiceTest {
```
yili-db: Is this test coverage useful?

newfront: This was only to make sure the changes I made didn't break the paths for no-auth

yili-db: Adding an auth expression shouldn't break any no-auth tests. Otherwise it's a basic fundamental thing being broken.
I suggest skipping it.

newfront: but when moving to the SdkAccessControlBaseCRUDTest base, auth is always on, so we know it will work, so I'll leave this out (remove the NoAuth test)

server/src/test/java/io/unitycatalog/server/service/PermissionServiceTest.java
```diff
@@ -0,0 +1,211 @@
...
+  protected static final String CONTROL_PATH = "/api/1.0/unity-control/";
+  protected static final String CATALOGS_ENDPOINT = BASE_PATH + "catalogs";
+  protected static final String SCIM_USERS_ENDPOINT = CONTROL_PATH + "scim2/Users";
+
+  protected SecurityConfiguration securityConfiguration;
+  protected SecurityContext securityContext;
+  protected WebClient client;
+
+  /** Whether the server under test should boot with {@code server.authorization=enable}. */
+  protected abstract boolean authorizationEnabled();
```
yili-db: SdkAccessControlBaseCRUDTest (and its base class) has auth always turned on.
This `authorizationEnabled` seems unnecessary since the no-auth test is already removed.

server/src/test/java/io/unitycatalog/server/service/PermissionServiceTest.java
```diff
@@ -0,0 +1,211 @@
...
+      serverProperties.setProperty(Property.AUTHORIZATION_ENABLED.getKey(), "enable");
+    }
+  }
+
+  @BeforeEach
+  @Override
+  public void setUp() {
+    super.setUp();
+
+    Path configurationFolder = Path.of("etc", "conf");
```
yili-db: These are already covered by `SdkAccessControlBaseCRUDTest` and its base class `BaseAccessControlCRUDTest`.
Most, if not all, of the setup, tear down, helper functions etc already exist there.

server/src/main/java/io/unitycatalog/server/service/PermissionService.java
newfront: thanks for showing me where these live. This has been updated

server/src/main/java/io/unitycatalog/server/service/PermissionService.java
newfront: noted. That was left over when I was backing out the changes when I was using the PATCH annotations. I'll be more careful next time.

server/src/test/java/io/unitycatalog/server/service/PermissionServiceWithoutAuthorizationTest.java
newfront: this has been deleted

server/src/test/java/io/unitycatalog/server/service/PermissionServiceTest.java
newfront: this is gone, anything inherited from the base class is now left there.

server/src/test/java/io/unitycatalog/server/service/PermissionServiceTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+  private AggregatedHttpResponse send(HttpMethod method, String path, String token, String body) {
+    if (body == null) {
+      return client.execute(headers(method, path, token)).aggregate().join();
+    }
+    return client.execute(headers(method, path, token), HttpData.ofUtf8(body)).aggregate().join();
+  }
+
+  private AggregatedHttpResponse getCatalogPermissions(String catalogName, String token) {
+    return send(HttpMethod.GET, BASE_PATH + "permissions/catalog/" + catalogName, token, null);
```
yili-db: This test should avoid sending raw http requests. Instead a better way is to exercise the SDK which sends the same requests:
```
    // Base class already provides adminApiClient and grantsApi but you can create more for different user identities
    // adminApiClient = TestUtils.createApiClient(adminConfig);
    // grantsApi = new GrantsApi(adminApiClient);
    ServerConfig regular1Config = createTestUserServerConfig(REGULAR_1);
    GrantsApi grantsApi = new GrantsApi(TestUtils.createApiClient(regular1Config));
...
      PermissionsChange permissionsChange =
          new PermissionsChange().principal(userEmail).add(privilegeList).remove(List.of());
      UpdatePermissions updatePermissions =
          new UpdatePermissions().changes(java.util.List.of(permissionsChange));
      grantsApi.update(resourceSecurableType, resourceFullName, updatePermissions);
...
    grantsApi.get(SecurableType.CATALOG, "catalogname", null);
```

yili-db:
The PR description needs some cleanup.

server/src/test/java/io/unitycatalog/server/service/PermissionServiceTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+    // Reading back as the owner exposes the full ACL, including regular-1's grant.
+    PermissionsList getResponse = grantsApi.get(SecurableType.CATALOG, CATALOG_NAME, null);
+    assertThat(privilegesFor(getResponse, REGULAR_1)).containsExactly(Privilege.USE_CATALOG);
+  }
+
+  @Test
+  @SneakyThrows
+  public void nonOwnerCannotUpdateCatalogPermissions() {
+    createTestUser(REGULAR_1);
```
yili-db: If you remove these 6 lines the same test can continue with testing a deny of REGULAR_1, right?

server/src/test/java/io/unitycatalog/server/service/PermissionServiceTest.java
```diff
@@ -0,0 +1,120 @@
...
+            regularGrantsApi.update(
+                SecurableType.CATALOG,
+                CATALOG_NAME,
+                addPrivilege(REGULAR_1, Privilege.USE_CATALOG)));
+  }
+
+  @Test
+  @SneakyThrows
+  public void nonOwnerCanReadCatalogPermissions() {
+    createTestUser(REGULAR_1);
```
yili-db: Same thing here, the test can simply continue with the same setup.
And if previously you have granted permission to both REGULAR_1 and REGULAR_2, here you can validate that response has REGULAR_1 only.

server/src/test/java/io/unitycatalog/server/service/PermissionServiceTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+  @Test
+  @SneakyThrows
+  public void ownerCanGrantAndReadCatalogPermissions() {
+    createTestUser(REGULAR_1);
+
+    // Admin (metastore owner) grants USE CATALOG to regular-1.
+    PermissionsList grantResponse =
+        grantsApi.update(
+            SecurableType.CATALOG, CATALOG_NAME, addPrivilege(REGULAR_1, Privilege.USE_CATALOG));
```
yili-db: Grant to both REGULAR_1 and REGULAR_2

server/src/test/java/io/unitycatalog/server/service/PermissionServiceTest.java
newfront: yes

yili-db (approved):

PR #1606, opened 2026-06-10
[Feat] Add HEAD table exists endpoint for Delta REST Catalog

foss-contributor:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1606/files) to review incremental changes.
- [**stack/delta-api-table-exists-head-endpoint**](https://github.com/unitycatalog/unitycatalog/pull/1606) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1606/files)]
  - [stack/delta-api-table-delete-endpoint](https://github.com/unitycatalog/unitycatalog/pull/1607) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1607/files/698fe12a84f1f76b4ef2054d636ddba59ce533d5..e2cf8146bcaa72b80af6d632754767b7273f96cd)]

---------
**Description of changes**

Adds a tableExists endpoint to the Delta REST Catalog so Delta clients can cheaply check whether a UC-managed or external Delta table exists.

```
Endpoint: HEAD /delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}
Operation ID: tableExists
Auth: same as loadTable (AuthorizeExpressions.GET_TABLE).
Responses: 204 No Content if the table exists, 404 Not Found if it doesn't.
```

Note that while the `delta.yaml` file describes an error body, the HTTP standard was implemented instead as any response body is removed automatically for the HEAD endpoint (see RFC 9110 §9.3.2).

Supersedes #1552, recreated on current main to resolve merge conflicts.

**Testing**

The new HEAD endpoint was tested against both managed and external tables to return status 204. The non-existent 404 error code was also tested.

PR #1559, opened 2026-05-20
[Feat] Add DELETE table endpoint for Delta REST Catalog

foss-contributor:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1559/files/95e47d2bb9bcd51955314f91e869336fa9c6f01d..4286a5241ce18f029f7fb9a99bf3e736fb68aa96) to review incremental changes.

- [foss-contributor:feat/delta-table-exists-head-endpoint](https://github.com/unitycatalog/unitycatalog/pull/1552) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1552/files)]
- [**foss-contributor:stack/delta-table-delete-endpoint**](https://github.com/unitycatalog/unitycatalog/pull/1559) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1559/files/95e47d2bb9bcd51955314f91e869336fa9c6f01d..4286a5241ce18f029f7fb9a99bf3e736fb68aa96)]

---

**Description of changes**

Adds a delete table endpoint to the Delta REST Catalog so Delta clients can delete UC-managed or external Delta tables through the Delta REST API.

```
Endpoint: DELETE /delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}
Operation ID: deleteTable
Auth: same as UC table delete (AuthorizeExpressions.DELETE_TABLE).
Responses: 204 No Content if the table is deleted, 404 Not Found if it doesn't exist.
```

**Testing**

The new DELETE endpoint was tested against both managed and external tables to return status 204 and make the table no longer loadable. The non-existent 404 error code was also tested.

PR #1607, opened 2026-06-10
[Feat] Add DELETE table endpoint for Delta REST Catalog

foss-contributor:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1607/files) to review incremental changes.
- [stack/delta-api-table-exists-head-endpoint](https://github.com/unitycatalog/unitycatalog/pull/1606) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1606/files)] [MERGED]
  - [**stack/delta-api-table-delete-endpoint**](https://github.com/unitycatalog/unitycatalog/pull/1607) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1607/files)]

---------
**Description of changes**

Adds a delete table endpoint to the Delta REST Catalog so Delta clients can delete UC-managed or external Delta tables through the Delta REST API.

```
Endpoint: DELETE /delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}
Operation ID: deleteTable
Auth: same as UC table delete (AuthorizeExpressions.DELETE_TABLE).
Responses: 204 No Content if the table is deleted, 404 Not Found if it doesn't exist.
```

Supersedes #1559, recreated on current main to resolve merge conflicts.

**Testing**

The new DELETE endpoint was tested against both managed and external tables to return status 204 and make the table no longer loadable. The non-existent 404 error code was also tested.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -865,6 +865,23 @@ public class TableRepository {
         /* readOnly = */ false);
   }
 
+  /**
+   * Deletes a table from a DAO the caller already holds (e.g. from {@link #findTableOrThrow}),
+   * sparing callers the dotted-name round-trip and schema lookup of {@link #deleteTable(String)}.
+   *
+   * @throws BaseException with ErrorCode.TABLE_NOT_FOUND if the table no longer exists.
+   */
+  public void deleteTable(TableInfoDAO dao) {
+    TransactionManager.executeWithTransaction(
+        sessionFactory,
+        session -> {
+          deleteTable(session, dao.getSchemaId(), dao.getName());
+          return null;
+        },
+        "Failed to delete table " + dao.getName(),
+        /* readOnly = */ false);
+  }
+
   public void deleteTable(Session session, UUID schemaId, String tableName) {
     TableInfoDAO tableInfoDAO = findBySchemaIdAndName(session, schemaId, tableName);
     if (tableInfoDAO == null) {
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -867,15 +867,13 @@ public class TableRepository {
 
   /**
    * Deletes a table from a DAO the caller already holds (e.g. from {@link #findTableOrThrow}),
-   * sparing callers the dotted-name round-trip and schema lookup of {@link #deleteTable(String)}.
-   *
-   * @throws BaseException with ErrorCode.TABLE_NOT_FOUND if the table no longer exists.
+   * sparing callers the dotted-name round-trip and lookups of {@link #deleteTable(String)}.
    */
   public void deleteTable(TableInfoDAO dao) {
     TransactionManager.executeWithTransaction(
         sessionFactory,
         session -> {
-          deleteTable(session, dao.getSchemaId(), dao.getName());
+          deleteTable(session, dao);
           return null;
         },
         "Failed to delete table " + dao.getName(),
@@ -887,6 +885,10 @@ public class TableRepository {
     if (tableInfoDAO == null) {
       throw new BaseException(ErrorCode.TABLE_NOT_FOUND, "Table not found: " + tableName);
     }
+    deleteTable(session, tableInfoDAO);
+  }
+
+  private void deleteTable(Session session, TableInfoDAO tableInfoDAO) {
     if (TableType.MANAGED.getValue().equals(tableInfoDAO.getType())) {
       try {
         FileOperations.deleteDirectory(NormalizedURL.from(tableInfoDAO.getUrl()));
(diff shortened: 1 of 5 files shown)


server/src/main/java/io/unitycatalog/server/service/delta/DeltaApiService.java
```diff
@@ -140,6 +142,28 @@ public HttpResponse tableExists(
...
+   */
+  @Delete("/delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}")
+  @AuthorizeExpression(AuthorizeExpressions.DELETE_TABLE)
+  @AuthorizeResourceKey(METASTORE)
+  public HttpResponse deleteTable(
+      @Param("catalog") @AuthorizeResourceKey(CATALOG) String catalog,
+      @Param("schema") @AuthorizeResourceKey(SCHEMA) String schema,
+      @Param("table") @AuthorizeResourceKey(TABLE) String table) {
+    String fullName = catalog + "." + schema + "." + table;
+    TableInfo tableInfo = tableRepository.getTable(fullName);
```
yili-db: similar to https://github.com/unitycatalog/unitycatalog/pull/1606/changes#r3390586342

server/src/main/java/io/unitycatalog/server/service/delta/DeltaApiService.java
```diff
@@ -140,6 +142,28 @@ public HttpResponse tableExists(
...
+    tableRepository.deleteTable(fullName);
```
yili-db: Once you have the TableInfoDAO, you can make a simpler version of delete like this:
```
public void deleteTable(TableInfoDAO dao) {
```

foss-contributor: It seems like we already have `public void deleteTable(String fullName)` in `TableRepository` that checks for existence before deleting. Thoughts on using this?

yili-db: The purpose of having this dao here is not for checking existence, but for `removeHierarchicalAuthorizations`

foss-contributor: Agree, updated to use the DAO's schema ID to prevent another round trip to get the table's schema ID for the removeAuthorization

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -865,6 +865,23 @@ public void deleteTable(String fullName) {
         /* readOnly = */ false);
   }
 
+  /**
+   * Deletes a table from a DAO the caller already holds (e.g. from {@link #findTableOrThrow}),
+   * sparing callers the dotted-name round-trip and schema lookup of {@link #deleteTable(String)}.
+   *
+   * @throws BaseException with ErrorCode.TABLE_NOT_FOUND if the table no longer exists.
+   */
+  public void deleteTable(TableInfoDAO dao) {
```
yili-db, line 872: Sorry I didn't see this earlier. Actually this function can *almost* directly call the following one:
```
  public void deleteTable(Session session, UUID schemaId, String tableName) {
    TableInfoDAO tableInfoDAO = findBySchemaIdAndName(session, schemaId, tableName);
```

So this one can be refactored into two functions. The remaining part after `findBySchemaIdAndName` can be made as a new `void deleteTable(Session session, TableInfoDAO dao)` which can be called from here directly without table lookup again.

foss-contributor: refactoring to keep the null check within `deleteTable(Session session, UUID schemaId, String tableName)` since once you have a dao it should not be null

server/src/main/java/io/unitycatalog/server/service/delta/DeltaApiService.java
```diff
@@ -140,6 +142,27 @@ public HttpResponse tableExists(
...
+   */
+  @Delete("/delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}")
+  @AuthorizeExpression(AuthorizeExpressions.DELETE_TABLE)
+  @AuthorizeResourceKey(METASTORE)
+  public HttpResponse deleteTable(
+      @Param("catalog") @AuthorizeResourceKey(CATALOG) String catalog,
+      @Param("schema") @AuthorizeResourceKey(SCHEMA) String schema,
+      @Param("table") @AuthorizeResourceKey(TABLE) String table) {
+    TableInfoDAO tableInfo = tableRepository.findTableOrThrow(catalog, schema, table);
+    tableRepository.deleteTable(tableInfo);
```
yili-db, line 160: while you are here, please also address this multi-txn problem.
The deletion should be done in a single txn like this:
```
   TableInfoDAO dao = tableRepository.deleteTable(catalog, schema, table);
```
Inside `deleteTable(catalog, schema, table)` it starts a single txn to do the findTable and delete and return the dao before deletion.
Please figure out if dao can be captured before deletion and returned.
The same pattern also applies to the TableService.java. The whole deletion flow can be simplified.

foss-contributor:
Reopened by https://github.com/unitycatalog/unitycatalog/pull/1623

PR #1617, opened 2026-06-12
Advance table updatedAt only on metadata changes in the Delta update endpoint

yili-db:
**PR Checklist**

**Description of changes**
updateTable advanced updatedAt on every request. updatedAt tracks metadata changes, so the etag (derived from it) must stay stable under data ingest, per the managed tables spec.

- Gate the updatedAt/updatedBy roll on the request carrying a metadata action; data-only add-commit and backfill-only requests no longer advance it.
- Exempt set-domain-metadata touching only the delta.rowTracking high-water mark: the Delta protocol makes writers mirror the HWM on every fresh-row commit, so it is data-stream bookkeeping, not a metadata change. The HWM property is still persisted; any other domain or sibling action still counts.
- Tests piggyback on existing blocks: metadata changes advance updated-time and etag; data-only, backfill-only, and HWM-only requests leave them (and the lastUpdateVersion stamp) unchanged; mixed requests still roll.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -294,8 +294,9 @@ public class TableRepository {
    * diff-flush at the end applies only the keys that actually changed. Request classification runs
    * before opening the transaction (see {@link DeltaUpdateTableMapper#collectRequest}); inside,
    * requirement checks run first so a stale-snapshot conflict short-circuits before any write is
-   * persisted. The DAO's {@code updatedAt}/{@code updatedBy} advance once at the end; the resulting
-   * etag naturally rolls.
+   * persisted. The DAO's {@code updatedAt}/{@code updatedBy} advance at the end only when the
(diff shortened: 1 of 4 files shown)


server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -322,8 +323,13 @@ public DeltaLoadTableResponse updateTableForDelta(
                               d.uniformFields(),
                               d.latestBackfilledVersion()));
           properties.flush(session, dao.getId());
-          dao.setUpdatedAt(new Date());
-          dao.setUpdatedBy(callerId);
+          // updatedAt/updatedBy track the last change to the table's catalog-visible metadata.
+          // Data-only add-commit and backfill-only requests change no metadata, so they must
+          // not advance these fields.
+          if (collected.updates().changesTableMetadata()) {
```
yili-db, line 329: This is the bug fix. Previously all commits bumps up the `updatedAt`

tdas (approved):

PR #1622, opened 2026-06-16
[Refactor] Return deleted TableInfoDAO from TableRepository.deleteTable

foss-contributor:
**Description of changes**

Refactors table deletion so `TableRepository.deleteTable(...)` returns the deleted `TableInfoDAO` instead of `void`. The DAO already carries the table and schema UUIDs, so the delete handler can remove the table's hierarchical authorizations from a **single** lookup — dropping the extra `getTable` + `getSchema` reads the UC delete handler previously made, and closing the time-of-check/time-of-use window between a separate lookup and the delete.

- `TableRepository`:
  - `void deleteTable(String fullName)` → `TableInfoDAO deleteTable(String fullName)` (thin wrapper that splits the dotted name, mirroring `getTable(String)`).
  - Adds `TableInfoDAO deleteTable(String catalog, String schema, String table)` for callers that already hold the three name parts separately.
  - The in-transaction `deleteTable(Session, UUID, String)` overload now returns the DAO and is package-private (only the schema cascade delete uses it).
- `TableService.deleteTable` now makes a single `deleteTable(fullName)` call and removes the hierarchical authorizations using the returned DAO.

No change to the UC REST `DELETE /tables/{full_name}` contract.

**Testing**

Pure refactor — covered by the existing suite. Verified locally: `server/Test/compile`, `javafmtCheckAll`, and Checkstyle (main + test) clean; targeted tests pass (`SdkTableCRUDTest`, `SdkTableAccessControlCRUDTest` — delete + auth removal end-to-end, `SdkMetricViewCRUDTest`, `SdkSchemaCRUDTest` — schema cascade delete); full `server/test`: 285/285.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -877,7 +877,7 @@ public class TableRepository {
         /* readOnly = */ false);
   }
 
-  public TableInfoDAO deleteTable(Session session, UUID schemaId, String tableName) {
+  TableInfoDAO deleteTable(Session session, UUID schemaId, String tableName) {
     TableInfoDAO tableInfoDAO = findBySchemaIdAndName(session, schemaId, tableName);
     if (tableInfoDAO == null) {
       throw new BaseException(ErrorCode.TABLE_NOT_FOUND, "Table not found: " + tableName);
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -849,29 +849,35 @@ public class TableRepository {
     return new ListTablesResponse().tables(result).nextPageToken(nextPageToken);
   }
 
-  public void deleteTable(String fullName) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -849,29 +849,35 @@ public ListTablesResponse listTables(
...
-          return null;
+              repositories.getSchemaRepository().getSchemaIdOrThrow(session, catalog, schema);
+          return deleteTable(session, schemaId, table);
         },
         "Failed to delete table",
         /* readOnly = */ false);
   }
 
-  public void deleteTable(Session session, UUID schemaId, String tableName) {
+  public TableInfoDAO deleteTable(Session session, UUID schemaId, String tableName) {
```
yili-db: This can be package private.

yili-db (approved):

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
foss-contributor: fixed

PR #1623, opened 2026-06-16
[Feat] Add DELETE table endpoint for Delta REST API

foss-contributor:
**Description of changes**

Adds a delete table endpoint to the Delta REST API so Delta clients can delete UC-managed or external Delta tables, wiring up the endpoint already advertised in the Delta REST API config.

```
Endpoint: DELETE /delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}
Operation ID: deleteTable
Auth: same as UC table delete (AuthorizeExpressions.DELETE_TABLE).
Responses: 204 No Content if the table is deleted, 404 Not Found if it doesn't exist.
```

**Testing**

The new DELETE endpoint was tested against both managed and external tables to return status 204 and make the table no longer loadable, plus the 404 case for a missing table (`SdkDeleteTableTest`).

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/delta/DeltaApiService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/delta/DeltaApiService.java
@@ -8,6 +8,7 @@ import static io.unitycatalog.server.model.SecurableType.TABLE;
 
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.server.annotation.Delete;
 import com.linecorp.armeria.server.annotation.ExceptionHandler;
 import com.linecorp.armeria.server.annotation.Get;
 import com.linecorp.armeria.server.annotation.Head;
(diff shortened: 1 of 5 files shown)


server/src/main/java/io/unitycatalog/server/service/delta/DeltaApiService.java
```diff
@@ -140,6 +142,27 @@ public HttpResponse tableExists(
...
+  /**
+   * Delete a table by three-part name. Mirrors {@link
+   * io.unitycatalog.server.service.TableService#deleteTable}, but returns 204 No Content per {@code
+   * delta.yaml} (the UC counterpart returns 200). {@code deleteTable} returns the deleted table's
+   * DAO, so the table and schema UUIDs for the authorization cleanup come from the same lookup that
+   * removed the table.
+   */
+  @Delete("/delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}")
+  @AuthorizeExpression(AuthorizeExpressions.DELETE_TABLE)
+  @AuthorizeResourceKey(METASTORE)
```
yili-db: `METASTORE` isn't needed because it doesn't appear in AuthorizeExpression.

yili-db (commented):
Just a minor comment

server/src/main/java/io/unitycatalog/server/service/delta/DeltaApiService.java
foss-contributor: fixed

yili-db (approved):

PR #1632, opened 2026-06-25
Make the OAuth token scope configurable via oauth.scope

timothyw553:
## What

The OAuth token provider (client-credentials flow) hardcoded `scope=all-apis` in the token request and ignored any configured scope. This adds an optional `oauth.scope` (set via `spark.sql.catalog.<name>.auth.oauth.scope` for the Spark connector), defaulting to `all-apis` so existing behavior is unchanged.

## Why

`all-apis` is the correct scope for the Databricks/UC OIDC token endpoint, but it is not a valid scope for other OAuth identity providers. Unity Catalog supports federating to external IdPs (Google, Okta, Azure AD — see `docs/server/auth.md`), whose token endpoints require their own scope — e.g. Azure AD's `<resource>/.default` or an Okta custom scope. With the scope hardcoded and no way to override it, pointing the connector's OAuth at a non-Databricks token endpoint cannot succeed, and there is no workaround.

## Changes

- `AuthConfigs`: add the `oauth.scope` key and the `all-apis` default.
- `OAuthTokenProvider`: read `oauth.scope` in `initialize()` (default `all-apis`), send it URL-encoded in the token request, and include it in `configs()` so it round-trips.
- Tests: assert the configured scope is sent in the token request and round-trips through `configs()`; the default remains `all-apis`.

## Compatibility

No behavior change when `oauth.scope` is unset (defaults to `all-apis`). `configs()` now includes `oauth.scope`; the existing OAuth assertions were updated accordingly.


clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
```diff
@@ -49,9 +51,29 @@ class OAuthTokenProvider implements TokenProvider {
...
+      String oauthClientId,
+      String oauthClientSecret,
+      String oauthScope,
+      long leadRenewalTimeSeconds,
+      ApiClient apiClient,
+      Clock clock) {
     Preconditions.checkNotNull(oauthUri, "OAuth URI must not be null");
     Preconditions.checkNotNull(oauthClientId, "OAuth client ID must not be null");
     Preconditions.checkNotNull(oauthClientSecret, "OAuth client secret must not be null");
+    Preconditions.checkNotNull(oauthScope, "OAuth scope must not be null");
```
seewishnew, line 58: Nit: Perhaps also worth falling-back to default with a warn log in case it's empty post-trim?

clients/java/src/test/java/io/unitycatalog/client/auth/OAuthTokenProviderTest.java
```diff
@@ -183,6 +189,68 @@ public void testConstructor() {
         .hasMessageContaining("Lead renewal time must be non-negative");
   }
 
+  @Test
+  public void testAccessTokenUsesConfiguredScope() throws Exception {
```
seewishnew, line 206: Can we also add a test case for multiple space-separated scopes?

seewishnew: One more thing to test would be fine-grained sub-scope representations, for e.g.: `catalog.tables:read`

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
```diff
@@ -93,6 +116,11 @@ public void initialize(Map<String, String> configs) {
         AuthConfigs.OAUTH_CLIENT_SECRET);
     this.oauthClientSecret = oauthClientSecret;
 
+    // Optional OAuth scope; defaults to all-apis when unset.
```
tdas, line 101: [AI-assisted] docs · normal

the description leans on docs/server/auth.md for the federation story, but this adds a user-facing config (oauth.scope / spark...auth.oauth.scope) and doesn't document it anywhere. can we add it in the same PR?

an undocumented key is invisible to the azure/okta users who need it, and the why (all-apis only works for the UC endpoint) drifts out of memory if it isn't written next to the feature.

clients/java/src/test/java/io/unitycatalog/client/auth/OAuthTokenProviderTest.java
tdas, line 206: [AI-assisted] tests · normal

the expected value recomputes the encoding with the same URLEncoder.encode prod uses — if the encoding were wrong the test would be wrong the same way and still pass. can we assert the literal wire string (scope=https%3A%2F%2Fexample.com%2F.default) instead?

the whole point of this PR is sending a non-trivial scope correctly on the wire; a test that mirrors the implementation can't catch an encoding bug.

tdas (changes requested):

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
```diff
@@ -93,6 +116,11 @@ public void initialize(Map<String, String> configs) {
...
+    String oauthScope = configs.get(AuthConfigs.OAUTH_SCOPE);
```
openinx: As we  will use  the  default value of `oauth.scope` to be `all-apis`, then I think we can use the `map.getOrDefault` directly here. 

```
this.oauthScope = configs.getOrDefault(AuthConfigs.OAUTH_SCOPE, AuthConfigs.DEFAULT_OAUTH_SCOPE);
```

openinx: What are the candidate options that we can set for the `oauth.scope` ?  where should we follow the doc the choose the selection ?

openinx: I mean if the values are limited set, then it will be better for us to have a strick check for the parse value.

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
```diff
@@ -128,7 +157,9 @@ private TempToken renewToken() {
           Base64.getEncoder().encodeToString(credentials.getBytes(StandardCharsets.UTF_8));
 
       // Prepare form data
-      String formData = "grant_type=client_credentials&scope=all-apis";
+      String formData =
+          "grant_type=client_credentials&scope="
+              + URLEncoder.encode(oauthScope, StandardCharsets.UTF_8);
```
openinx, line 142: will the oauthScope have any special char that need to use the URLEncoder.encode to translate ?

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
timothyw553: oauth.scope is now trimmed, and empty values use all-apis.

clients/java/src/test/java/io/unitycatalog/client/auth/OAuthTokenProviderTest.java
timothyw553: Added a test for multiple space-separated scopes, including catalog.tables:read.

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
timothyw553: Added docs for oauth.scope and the Spark auth.oauth.scope config.

clients/java/src/test/java/io/unitycatalog/client/auth/OAuthTokenProviderTest.java
timothyw553: The test now checks the exact encoded request body instead of using URLEncoder for the expected value.

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
timothyw553: Kept explicit normalization instead of getOrDefault because blank values still need all-apis, and custom scopes come from the OAuth provider.

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
timothyw553: URLEncoder is still needed because OAuth scopes can contain special characters and the token request is form-encoded.

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
openinx: Could we just remove this constructor and keep the widen one ?

docs/server/auth.md
```diff
@@ -67,6 +67,38 @@ server.allowed-issuers=https://accounts.google.com,https://login.microsoftonline
...
+
+For example, Java client configuration can include:
+
+```java
+Map<String, String> configs = Map.of(
+    "type", "oauth",
+    "oauth.uri", "https://login.example.com/oauth2/v2.0/token",
+    "oauth.clientId", "<client ID>",
+    "oauth.clientSecret", "<client secret>",
+    "oauth.scope", "https://example.com/.default");
```
openinx: Is this a correct example ?

timothyw553: fixed.

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
timothyw553: done.

openinx (approved):
LGTM, thanks @timothyw553

yili-db:
The doc update has been split to https://github.com/unitycatalog/unitycatalog/pull/1747

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
```diff
@@ -161,6 +172,13 @@ private TempToken renewToken() {
     }
   }
 
+  private static String normalizeOAuthScope(String oauthScope) {
+    if (oauthScope == null || oauthScope.trim().isEmpty()) {
+      return AuthConfigs.DEFAULT_OAUTH_SCOPE;
+    }
+    return oauthScope.trim();
```
yili-db: nit: this will trim it twice?

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
yili-db: I agree that "absent" config should fallback to default.
But what is the best outcome for "empty post-trim" config value here (and especially in `configs.get(AuthConfigs.OAUTH_SCOPE)`). In this case users specifically put `oauth.scope=` in config, falling back to `all-apis` may not be the intended result.
Instead, how about omitting `&scope=` parameter at all if it's empty? Or fail if the config is specified as empty?

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
```diff
@@ -161,6 +172,13 @@ private TempToken renewToken() {
     }
   }
 
+  private static String normalizeOAuthScope(String oauthScope) {
+    if (oauthScope == null || oauthScope.trim().isEmpty()) {
```
yili-db: nit:   One replaceAll alongside the trim:
```
  // RFC 6749 §3.3: scope is space-delimited, single SP between tokens. Collapse any
  // interior whitespace (double spaces, tabs from copy-pasted config) so it does not
  // go out form-encoded as +++ or %09, which a strict IdP rejects as invalid_scope.
  return oauthScope.trim().replaceAll("\\s+", " ");
```

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
timothyw553: changed it to just trim once during normalization

clients/java/src/main/java/io/unitycatalog/client/auth/OAuthTokenProvider.java
timothyw553: done

timothyw553: and added tests with spaces and tabs.

yili-db (approved):

tdas (approved):

PR #1635, opened 2026-06-26
Emit per-scope credential properties from CredPropsUtil

foss-contributor:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1635/files/8f16cd2ef54183bf67f7d3662944e2bd5298da2e..461be13f9b828b142affc5ca910105020052dc06) to review incremental changes.
- [stack/hadoop-cred-scopes-extract](https://github.com/unitycatalog/unitycatalog/pull/1633) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1633/files)]
  - [stack/hadoop-cred-scopes-overlay](https://github.com/unitycatalog/unitycatalog/pull/1634) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1634/files/d696d40fd9b56df3db85dada833afed8bc1fb3d2..8f16cd2ef54183bf67f7d3662944e2bd5298da2e)]
    - [**stack/hadoop-cred-scopes-emit**](https://github.com/unitycatalog/unitycatalog/pull/1635) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1635/files/8f16cd2ef54183bf67f7d3662944e2bd5298da2e..461be13f9b828b142affc5ca910105020052dc06)] ← _this PR_

---------

**PR Checklist**

**Description of changes**

`CredPropsUtil` now writes each additional vended credential into the `fs.unitycatalog.credscope.*` keys the filesystem reads back, activating multi-scope credential use end to end.

- `createDeltaTableCredProps` fetches the table's vend once through the existing (optionally cached) credential path; the additional scopes ride on the returned credential, so a cache hit still exposes the whole vend.
- For each additional scope, it builds that scope's namespaced properties — keyed by the scope's own prefix so it renews independently, like the table's own credential — and encodes them via `CredentialScopes.encode`.
- A single-credential vend emits no `credscope.*` keys, so its output is unchanged.
- Extracted a `buildSchemeProps` helper so the primary and each scope build props from an already-fetched credential without a second fetch.

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -13,6 +13,8 @@ import io.unitycatalog.hadoop.internal.auth.CredentialCache;
 import io.unitycatalog.hadoop.internal.auth.CredentialCache.RenewableCredential;
 import io.unitycatalog.hadoop.internal.auth.GenericCredential;
 import io.unitycatalog.hadoop.internal.auth.GenericCredentialFetcher;
+import io.unitycatalog.hadoop.internal.auth.ScopedCredential;
+import io.unitycatalog.hadoop.internal.fs.CredentialScopes;
 import io.unitycatalog.hadoop.internal.id.CredId;
 import io.unitycatalog.hadoop.internal.id.DeltaStagingTableCredId;
 import io.unitycatalog.hadoop.internal.id.DeltaTableCredId;
@@ -22,8 +24,11 @@ import io.unitycatalog.hadoop.internal.util.ClockUtil;
 import java.net.URI;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import org.apache.hadoop.conf.Configuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Internal utility that builds cloud-provider specific Hadoop configuration properties for Unity
@@ -34,6 +39,8 @@ import org.apache.hadoop.conf.Configuration;
  */
 public class CredPropsUtil {
   private CredPropsUtil() {}
+
+  private static final Logger LOG = LoggerFactory.getLogger(CredPropsUtil.class);
 
   /**
    * Factory seam for {@link GenericCredentialFetcher#create(ApiClient, CredId)}, swappable from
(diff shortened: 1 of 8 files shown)


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/ScopedCredential.java
```diff
@@ -0,0 +1,30 @@
+package io.unitycatalog.hadoop.internal.auth;
+
+import io.unitycatalog.client.model.TemporaryCredentials;
+
+/** A vended credential scoped to a storage location prefix other than the table's own. */
+public final class ScopedCredential {
```
openinx, line 6: I didn't get why do we need this `ScopedCredential` ?  Could you pls explain more about the reason ? Thanks.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/DeltaStorageCredentialUtil.java
```diff
@@ -36,6 +39,35 @@ public static TemporaryCredentials toTemporaryCredentials(DeltaStorageCredential
...
+      DeltaStorageCredential primary, List<DeltaStorageCredential> creds) {
+    List<ScopedCredential> scoped = new ArrayList<>();
+    for (DeltaStorageCredential cred : creds) {
+      if (cred == null || cred == primary || cred.getPrefix() == null) {
+        continue;
+      }
+      if (!isSupportedCloudScheme(URI.create(cred.getPrefix()).getScheme())) {
+        continue;
+      }
+      String operation = cred.getOperation() == null ? "READ" : cred.getOperation().getValue();
```
openinx, line 65: What's the case that `cred.getOperation()` will be null ?  and why is it correct to use the `READ` as the operation by default if there is no specified operation ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -380,16 +387,62 @@ public static Map<String, String> createDeltaTableCredProps(
...
+    // Key each scope's props by its own prefix so that, in renewal mode, its refresh re-selects
+    // this scope's entry from the table's vend and it renews like the table's own credential.
+    int index = 0;
+    for (ScopedCredential scope : additionalScopes) {
+      DeltaTableCredId scopeCredId =
+          new DeltaTableCredId(identifier, scope.operation(), scope.prefix());
+      Map<String, String> scopeProps =
+          buildSchemeProps(
+              renewCredEnabled,
+              /* credScopedFsEnabled= */ false,
```
openinx, line 433: wait .... why you want to explicitly disable the `credScopedFsEnabled` ?  As we are going to fully enable the the credScoped filesystem for all the code path, and this change seems a very hacky approach that you are using, and this makes it really hard to track.

openinx: for context, In future release, we will even remove the `credScopedFsEnabled` switch, since we will going to set all credential vending path to use the credScopedFS by default, so there will be no use to retain this switch any more.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -380,16 +387,62 @@ public static Map<String, String> createDeltaTableCredProps(
...
+          identifier.catalog(),
+          identifier.schema(),
+          identifier.table(),
+          additionalScopes.size());
+    }
+    Map<String, String> merged = new HashMap<>(props);
+    // Key each scope's props by its own prefix so that, in renewal mode, its refresh re-selects
+    // this scope's entry from the table's vend and it renews like the table's own credential.
+    int index = 0;
+    for (ScopedCredential scope : additionalScopes) {
```
openinx, line 427: why will we need multiple additional scope credentials for the cloned table ?  I didn't understand that.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -380,16 +387,62 @@ public static Map<String, String> createDeltaTableCredProps(
...
+      DeltaTableCredId scopeCredId =
+          new DeltaTableCredId(identifier, scope.operation(), scope.prefix());
```
openinx, line 429: will it happen the case that the scope cred id has the same identifier/operation/prefix with the `credId` that we already created in the `createDeltaTableCredProps` code ? 

If those two kinds of credid are conflicting, then it will seriously pollute the credentials cache, which make this entire cache unavailable, and causing serious bug.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -380,16 +387,62 @@ public static Map<String, String> createDeltaTableCredProps(
...
+              renewCredEnabled,
+              /* credScopedFsEnabled= */ false,
+              hadoopConf,
+              URI.create(scope.prefix()).getScheme(),
+              catalogUri,
+              tokenProvider,
+              appVersions,
+              scopeCredId,
+              new GenericCredential(scope.credentials()));
+      CredentialScopes.encode(merged, index, scope.prefix(), scopeProps);
```
openinx, line 441: The multiple additional scope credentials, actually make the hadoop conf setting even more complex, which I'm wondoring if it's worth to introduce so...  we will have to discuss to simplify the design, otherwise it will make us extremely hard to maintain those code in the critical credential vending path.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/DeltaStorageCredentialUtil.java
```diff
@@ -36,6 +39,35 @@ public static TemporaryCredentials toTemporaryCredentials(DeltaStorageCredential
     return out;
   }
 
+  /** Cloud schemes the connector can translate into Hadoop credential properties. */
+  private static boolean isSupportedCloudScheme(String scheme) {
```
openinx, line 43: be careful to hard those supported cloud scheme in multiple places.  as if we want to support a new scheme, then we will have to change all the places and cut a new release to support, that's hard to maintain.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/DeltaStorageCredentialUtil.java
foss-contributor: Should never be null, defensive degrade to READ. I think better to drop the null check and use the operation field as is.

PR #1642, opened 2026-07-02
[Hadoop] Add the initial credential into the global cache to reuse cred across queries. 

openinx:
**PR Checklist**

**Description of changes**

In this PR, we introduce a initCredentialCache, so that we can cache the initial credential at the driver side,  which can accelerate the cross queries credential access. 

For example.  Q1 access the table , and Q2 access the same table again,  then the credential will be shared for the two queries.  that can be helpful to accelerate the query performance.

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -48,22 +48,7 @@ public class CredPropsUtil {
   public static volatile GenericCredentialFetcherFactory genericCredFetcherFactory =
       GenericCredentialFetcher::create;
 
-  // Caches the initial credential fetched here so that different queries targeting the same
-  // credential scope ({@link CredId}) reuse the same vended credential instead of re-fetching from
-  // the UC server. It is re-fetched only once the cached credential is about to expire. Kept
-  // separate from the token-provider renewal cache since this runs on the query-planning side (e.g.
-  // the Spark driver), which is engine-agnostic here.
-  private static final String UC_INITIAL_CREDENTIAL_CACHE_MAX_SIZE =
-      "unitycatalog.initial.credential.cache.maxSize";
-  private static final int UC_INITIAL_CREDENTIAL_CACHE_MAX_SIZE_DEFAULT = 1024;
-  static final CredentialCache initialCredCache;
-
-  static {
-    int maxSize =
-        Integer.getInteger(
-            UC_INITIAL_CREDENTIAL_CACHE_MAX_SIZE, UC_INITIAL_CREDENTIAL_CACHE_MAX_SIZE_DEFAULT);
-    initialCredCache = new CredentialCache(maxSize);
-  }
+  static final CredentialCache initialCredCache = CredentialCache.createInitialCredentialCache();
 
   private static final String CRED_SCOPED_FS_CLASS =
       "io.unitycatalog.hadoop.internal.fs.CredScopedFileSystem";
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -4,7 +4,6 @@ import io.unitycatalog.client.ApiClient;
 import io.unitycatalog.client.ApiException;
 import io.unitycatalog.client.auth.TokenProvider;
 import io.unitycatalog.client.internal.ApiClientUtils;
-import io.unitycatalog.client.internal.Clock;
 import io.unitycatalog.client.model.AwsCredentials;
 import io.unitycatalog.client.model.AzureUserDelegationSAS;
 import io.unitycatalog.client.model.GcpOauthToken;
@@ -19,6 +18,7 @@ import io.unitycatalog.hadoop.internal.id.DeltaStagingTableCredId;
 import io.unitycatalog.hadoop.internal.id.DeltaTableCredId;
 import io.unitycatalog.hadoop.internal.id.PathCredId;
 import io.unitycatalog.hadoop.internal.id.TableCredId;
+import io.unitycatalog.hadoop.internal.util.ClockUtil;
 import java.net.URI;
 import java.util.Collections;
 import java.util.HashMap;
@@ -536,14 +536,13 @@ public class CredPropsUtil {
         hadoopConf.getLong(
             UCHadoopConfConstants.UC_RENEWAL_LEAD_TIME_KEY,
             UCHadoopConfConstants.UC_RENEWAL_LEAD_TIME_DEFAULT_VALUE);
-    Clock clock = clock(hadoopConf);
 
     return initialCredCache.access(
         credId,
         () ->
             new RenewableCredential(
                 renewalLeadTimeMillis,
-                clock,
+                ClockUtil.resolveClock(hadoopConf),
                 createCredential(apiClient, catalogUri, tokenProvider, appVersions, credId)));
   }
 
@@ -557,12 +556,6 @@ public class CredPropsUtil {
     ApiClient client =
         apiClient != null ? apiClient : createApiClient(catalogUri, tokenProvider, appVersions);
     return genericCredFetcherFactory.create(client, credId).createCredential();
-  }
-
-  private static Clock clock(Configuration hadoopConf) {
-    // Use the test clock if one is intentionally configured for testing.
-    String clockName = hadoopConf.get(UCHadoopConfConstants.UC_TEST_CLOCK_NAME);
-    return clockName != null ? Clock.getManualClock(clockName) : Clock.systemClock();
   }
 
   private static ApiClient createApiClient(
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -4,11 +4,15 @@ import io.unitycatalog.client.ApiClient;
 import io.unitycatalog.client.ApiException;
 import io.unitycatalog.client.auth.TokenProvider;
 import io.unitycatalog.client.internal.ApiClientUtils;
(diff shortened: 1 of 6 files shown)


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -506,46 +534,82 @@ public static Map<String, String> createPathCredProps(
...
+      TokenProvider tokenProvider,
+      Map<String, String> appVersions,
+      CredId credId)
+      throws ApiException {
+    boolean credCacheEnabled =
+        hadoopConf.getBoolean(
+            UCHadoopConfConstants.UC_CREDENTIAL_CACHE_ENABLED_KEY,
+            UCHadoopConfConstants.UC_CREDENTIAL_CACHE_ENABLED_DEFAULT_VALUE);
+    if (!credCacheEnabled) {
+      return createCredential(apiClient, catalogUri, tokenProvider, appVersions, credId);
```
openinx, line 517: In this case, we intentionally disabled the credential cache, so let's just skip to access the cache, and request the API to get a new credential directly.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -221,7 +241,8 @@ protected AbfsPropsBuilder self() {
   }
 
   private static Map<String, String> s3FixedCredProps(
-      boolean credScopedFsEnabled, Configuration hadoopConf, TemporaryCredentials tempCreds) {
+      boolean credScopedFsEnabled, Configuration hadoopConf, GenericCredential cred) {
```
openinx, line 230: Here we change it from `TemporaryCredentials` to the `GenericCredential`, because the `GenericCredential` is a more abstracted class, which actually is not bind to the unitycatalog client generated models. 

And also, that help us to align with the `globalCache` key and value types.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -449,46 +475,82 @@ private static Map<String, String> createCredProps(
...
+      GenericCredential cached = initialCredCache.getIfPresent(credId);
+      // Reuse the cached credential while it's still valid; otherwise fetch and cache a fresh one.
+      if (cached != null && !cached.readyToRenew(clock, renewalLeadTimeMillis)) {
+        return cached;
+      }
+      GenericCredential created =
+          createCredential(apiClient, catalogUri, tokenProvider, appVersions, credId);
+      initialCredCache.put(credId, created);
+      return created;
+    }
```
openinx: maybe we can abstract the common part from there and `GenericCredentialProvider`, so that we can share the same code in the common part.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
```diff
@@ -0,0 +1,76 @@
...
+import io.unitycatalog.hadoop.internal.util.BoundedKeyedCache;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Caches vended {@link GenericCredential}s keyed by their scope ({@link CredId}). A cached
+ * credential is reused while it is still valid and re-fetched via the supplied factory only once it
+ * is about to expire.
+ */
+public class CredentialCache {
```
openinx, line 15: As both the `CredPropsUtil` and `GenericCredentialProvider` has the credential cache, so I just abstracted a common `CredentialCache` here, so that both classes can reuse it, at least for the credential renewal process.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
```diff
@@ -0,0 +1,76 @@
...
+  private final BoundedKeyedCache<CredId, RenewableCredential> cache;
+
+  public CredentialCache(int maxSize) {
+    this.cache = new BoundedKeyedCache<>(maxSize);
+  }
+
+  public GenericCredential access(CredId credId, RenewableCredentialFactory factory)
+      throws ApiException {
+    synchronized (cache) {
```
yili-db, line 55: This will lock this global cache for all table loading, when one table is fetching cred from UC via `factory.create()`.
In general this `synchronized(cache)` should not exist anymore because `BoundedKeyedCache` is already lock aware. It has a `getOrLoad` that locks on only one key that can almost do this, **except** that it doesn't have a concept of expiration. If you add expiration control to `BoundedKeyedCache` it will be the most optimal locking for this cache.

Of course I see that existing code is already doing `synchronized (globalCache) {` so your code is just doing the same. Above optimization could be done as a separate follow up. Since now we are trying to improve the performance it may be worthy.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
openinx: Also addressed this comment here: https://github.com/unitycatalog/unitycatalog/pull/1642#discussion_r3516378917

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
openinx: if there are multiple accessor which are accessing the same cred based on the `CredId`. there are the cases:

1.  if cache hit and cred is still valid.  then  the lock duration will be very short.  no much impact to the latency. 
2. if cache hit but cred is expired.  then need to wait the duration to re-access the new credential,  the lock duration maybe 100ms.   but that's exactly the case that we want to intentionally design,  since we just want only one accessor can renew the cred, and others will reuse it. 
3.  if cache miss,  same as the case.2 

I think your point is,  we should make it to be a more fine-grain lock, so that one `credId-A` won't affect another `credId-B` access, right ? 

Yes,  I agree that can be a follow-up optimization, which may need some refactor for the `BoundedKeyedCache`.

openinx: Filed a separate issue for the follow-up work: https://github.com/unitycatalog/unitycatalog/issues/1651

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
```diff
@@ -0,0 +1,76 @@
...
+ * is about to expire.
+ */
+public class CredentialCache {
+  private final BoundedKeyedCache<CredId, RenewableCredential> cache;
+
+  public CredentialCache(int maxSize) {
+    this.cache = new BoundedKeyedCache<>(maxSize);
+  }
+
+  public GenericCredential access(CredId credId, RenewableCredentialFactory factory)
```
yili-db, line 53: `getOrRenew` or `getOrFetch`

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
```diff
@@ -0,0 +1,76 @@
...
+      cache.put(credId, created);
+      return created.credential();
+    }
+  }
+
+  public void clear() {
+    cache.clear();
+  }
+
+  public int size() {
```
yili-db: can this be package private for testing?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
```diff
@@ -0,0 +1,76 @@
...
+    return cache.size();
+  }
+
+  public List<GenericCredential> values() {
```
yili-db: same here

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
openinx: the access have 3 cases: 
1.  cache hit and cred is valid --> get. 
2. cache hit and cred is invalid --> renew.
3. cache miss -> create. 

so actually, it has `get`, `renew`, and `create` 3 different meaning.  seems like neither  the names meet the meaning.  let me add a more easy to understand doc so that people won't get lost.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
openinx: Those are testing code,  let me push those three methods `size()`, `clear()`, `values()` to test code directly.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
```diff
@@ -0,0 +1,98 @@
...
+ * is about to expire.
+ */
+public class CredentialCache {
+  private static final int DEFAULT_MAX_SIZE = 1024;
+  private static final String GLOBAL_CACHE_MAX_SIZE_KEY = "unitycatalog.credential.cache.maxSize";
+  private static final String INITIAL_CACHE_MAX_SIZE_KEY =
+      "unitycatalog.initial.credential.cache.maxSize";
+
+  // Visible for testing so tests can inspect and reset the underlying cache directly.
+  public final BoundedKeyedCache<CredId, RenewableCredential> cache;
```
yili-db: Ah this is actually worse. The cache itself shouldn't be public (even when commented as testing only).

Instead, let's revert back to the 3 test related method:
clear(): make it public. It's called from tests of parent package
size() / credentials(): make them package-private. They are only called from BaseTokenProviderTest.

This is better than having the entire cache public.

openinx: Done.

yili-db (approved):

PR #1644, opened 2026-07-02
[Spark] Extract table-create column/partition conversions into UCColumnConversions

chenwang-databricks:
**Stack 1/4** — first of a 4-PR split of #1554 (metric-view connector). Review bottom-up; merge in order.

### What this PR does
Pure refactor (no behavior change): extract the table-create column/partition conversions out of `UCSingleCatalog` into a shared `UCColumnConversions` object.

- `toStructFieldJson(StructField)` — the `{name,type,nullable,metadata}` `type_json` serializer, moved out of `UCSingleCatalog`.
- `partitionColumnNames(Array[Transform])` — the `identity`/`cluster_by` partition-transform logic, moved out and documented (answers the "why single field / why drop cluster_by" review questions).

This gives the upcoming view-create path one serializer to reuse instead of duplicating the `StructField`/`Column` → `type_json` logic (addresses @openinx's "unify `toStructFieldJson`" comment). The `Column` overload lands in stack PR 3.

### Stack
1. **this PR** — extract `UCColumnConversions`
2. per-Spark-version shim scaffolding
3. wire metric-view support + move view-type mapping to `UCViewTypes` + unit tests
4. local-run e2e tests

Verified: compiles on Spark 4.0/4.1/4.2; `UCProxySuite` (normal-table create path) green.

This pull request and its description were written by Isaac.

This PR is assisted by Isaac Autopilot: [View the full session spec, test plan, and artifacts](http://go/isaac-autopilot/cli-viewer/1782162701477/chen-wang_data).

PR #1645, opened 2026-07-02
[Spark] Per-Spark-version view-shim scaffolding.

chenwang-databricks:
**Stack 2/4** — split of #1554. Stacked on #1644.

> This PR is based on #1644. GitHub shows the cumulative diff (fork-based stacks can't use a fork branch as the PR base); **review only this PR's own commit** `57a6c3bb`, and merge after #1644.

### What this PR does
Adds the per-Spark-version shim source files so the connector compiles against Spark 4.0/4.1/4.2 while the view-catalog APIs (`TableViewCatalog`, `ViewInfo`, `MetadataTable`) only exist on 4.2. **No behavior** — no traits are mixed in yet.

- `scala-shims/spark-{4.0,4.1}`: a compile-time `TableViewCatalog` stub (the real interface arrives in Spark 4.2 via [SPARK-52729](https://github.com/apache/spark/pull/51419)) + no-op `ViewSupport` / `UCProxyViewSupport` traits.
- `scala-shims/spark-4.2`: placeholder `ViewSupport` / `UCProxyViewSupport` (still no-ops here; the real 4.2 impl lands in stack PR 3).
- Widens `UCProxy` to `private[spark]` so the shim trait can self-type to it.

Answers @openinx's "where is `TableViewCatalog`?" — it's the SPARK-52729 interface; 4.0/4.1 get the vendored stub, 4.2 uses real Spark.

### Stack
1. #1644 — extract `UCColumnConversions`
2. **this PR** — shim scaffolding
3. wire metric-view support + `UCViewTypes` + unit tests
4. local-run e2e tests

Verified: compiles on Spark 4.0/4.1/4.2; no behavior change.

This PR is assisted by Isaac Autopilot: [View the full session spec, test plan, and artifacts](http://go/isaac-autopilot/cli-viewer/1782162701477/chen-wang_data).

### DIFF
--- a/connectors/spark/src/main/scala-shims/spark-4.0/io/unitycatalog/spark/UCProxyViewSupport.scala
+++ b/connectors/spark/src/main/scala-shims/spark-4.0/io/unitycatalog/spark/UCProxyViewSupport.scala
@@ -1,0 +1,16 @@
+package io.unitycatalog.spark
+
+import io.unitycatalog.client.model.{TableInfo => UCTableInfo}
+import org.apache.spark.sql.catalyst.analysis.NoSuchTableException
+import org.apache.spark.sql.connector.catalog.{Identifier, Table}
+
+/**
+ * Spark 4.0 does not support the metric-view API, so this shim leaves views unsupported: the
+ * `wrapAsView` hook throws because no view types exist on this Spark version. The shared
+ * `UCProxy.loadTable` only calls this hook for a view-like UC `tableType`, which cannot occur on a
+ * Spark version without view support.
+ */
+trait UCProxyViewSupport { self: UCProxy =>
+  protected def wrapAsView(t: UCTableInfo, ident: Identifier): Table =
+    throw new NoSuchTableException(ident)
+}
--- a/connectors/spark/src/main/scala-shims/spark-4.0/io/unitycatalog/spark/ViewSupport.scala
+++ b/connectors/spark/src/main/scala-shims/spark-4.0/io/unitycatalog/spark/ViewSupport.scala
@@ -1,0 +1,15 @@
+package io.unitycatalog.spark
+
+import org.apache.spark.sql.connector.catalog.{Identifier, Table}
+
+/**
+ * Spark 4.0 does not have the metric-view APIs (`TableViewCatalog`, `loadTableOrView`,
+ * `createView`, `loadView`, ...), so this shim provides `rejectIfShimmedView` as a no-op. It
+ * lets the shared [[UCSingleCatalog.loadTable]] body compile and run without referencing any
+ * view-only types. The `with TableViewCatalog` mixin on `UCSingleCatalog` resolves to the
+ * [[org.apache.spark.sql.connector.catalog.TableViewCatalog]] stub trait also defined in this
+ * shim directory.
+ */
+trait ViewSupport { self: UCSingleCatalog =>
+  protected def rejectIfShimmedView(t: Table, ident: Identifier): Unit = ()
+}
--- a/connectors/spark/src/main/scala-shims/spark-4.1/io/unitycatalog/spark/UCProxyViewSupport.scala
+++ b/connectors/spark/src/main/scala-shims/spark-4.1/io/unitycatalog/spark/UCProxyViewSupport.scala
@@ -1,0 +1,16 @@
+package io.unitycatalog.spark
+
+import io.unitycatalog.client.model.{TableInfo => UCTableInfo}
+import org.apache.spark.sql.catalyst.analysis.NoSuchTableException
+import org.apache.spark.sql.connector.catalog.{Identifier, Table}
+
+/**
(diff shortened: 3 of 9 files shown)


connectors/spark/src/main/scala-shims/spark-4.0/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,18 @@
+package io.unitycatalog.spark
+
+import io.unitycatalog.client.model.{TableInfo => UCTableInfo}
+import org.apache.spark.sql.catalyst.analysis.NoSuchTableException
+import org.apache.spark.sql.connector.catalog.{Identifier, Table}
+
+/**
+ * Empty Spark-4.0 shim for [[UCProxyViewSupport]] (the real implementation lives
+ * in `scala-shims/spark-4.2/`). Provides the `wrapAsView` hook that throws on Spark 4.0
```
openinx: > (the real implementation lives in `scala-shims/spark-4.2/`)

I suggest to remove this, as if we plan to add the spark-4.3 support for the oss unitycatalog, we will have to change all the spark-4.x `scala-shims` layer's comment. 

Instead, we can simply say `spark-4.0 does not support the metric view API, so left it as unsupported`.

connectors/spark/src/main/scala-shims/spark-4.0/io/unitycatalog/spark/ViewSupport.scala
```diff
@@ -0,0 +1,18 @@
+package io.unitycatalog.spark
+
+import org.apache.spark.sql.connector.catalog.{Identifier, Table}
+
+/**
+ * Empty Spark-4.0 shim for [[ViewSupport]] (the real implementation lives in
+ * `scala-shims/spark-4.2/`). Provides the `rejectIfShimmedView` hook as a no-op so that
+ * the shared [[UCSingleCatalog.loadTable]] body compiles and runs on Spark 4.0 without
+ * any reference to Spark-4.2-only types (`MetadataTable`, `ViewInfo`).
```
openinx: Pls don't hardcode the spark4.2 here, since if we plan to support spark 4.3 , we will have to change all those comments also.

connectors/spark/src/main/scala-shims/spark-4.1/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,18 @@
+package io.unitycatalog.spark
+
+import io.unitycatalog.client.model.{TableInfo => UCTableInfo}
+import org.apache.spark.sql.catalyst.analysis.NoSuchTableException
+import org.apache.spark.sql.connector.catalog.{Identifier, Table}
+
+/**
+ * Empty Spark-4.0 shim for [[UCProxyViewSupport]] (the real implementation lives
```
openinx: [AI Generated] nit: Scaladoc says "Empty Spark-4.0 shim" but this file is under `scala-shims/spark-4.1/`. Looks like a copy-paste from the 4.0 directory — same issue in `ViewSupport.scala:6`. Please update both 4.1 files to say Spark-4.1.

connectors/spark/src/main/scala-shims/spark-4.0/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -0,0 +1,18 @@
...
+ * because no view types exist there -- the shared `UCProxy.loadTable` calls into this
+ * hook only when a UC row's `tableType` is in [[UCSingleCatalog.viewLikeUcTypes]], which
```
openinx: [AI Generated] nit: References `[[UCSingleCatalog.viewLikeUcTypes]]`, which does not exist until stack PR 3. The scaladoc link will be broken until then. Consider plain text (e.g. "added in a follow-up PR") instead of a wiki link. Same applies to the 4.1 copy of this file.

openinx (commented):
LGTM. Minor suggestions from TD review.

chenwang-databricks:
Thanks! Addressed all four scaladoc nits in the latest revision:

1. Dropped the hardcoded `scala-shims/spark-4.2/` pointer in the 4.0/4.1 `UCProxyViewSupport`/`ViewSupport` docs — the wording is now version-agnostic ("Spark 4.x does not support the metric-view API, so this shim leaves views unsupported"), so a future spark-4.3 shim won't need comment edits.
2. Same for the `ViewSupport` docs — no more spark-4.2 reference.
3. Fixed the copy-paste: the spark-4.1 files now say "Spark 4.1" instead of "Spark-4.0".
4. Removed the broken wiki-link from this PR's docs (the referenced member only exists once the glue PR lands), using plain text here; the follow-up PR reintroduces the `[[UCViewTypes.isViewLikeTableType]]` link where the symbol actually exists.

openinx (approved):
LGTM, thanks @chenwang-databricks for the work.

PR #1646, opened 2026-07-02
[Spark] Wire metric-view support into the connector + UCViewTypes + tests (3/4)

chenwang-databricks:
**Stack 3/4** — split of #1554. Stacked on #1645.

> Based on #1645 (which is based on #1644). GitHub shows the cumulative diff; **review only this PR's own commit** `3b76bc27`, and merge after #1645.

### What this PR does
Wires metric-view support into the connector on Spark 4.2 and applies the review-driven refactors.

- **Real 4.2 `ViewSupport` / `UCProxyViewSupport`**: `listViews` / `loadView` / `createView` / `dropView` / `renameView` (+ `loadTableOrView`), wrapping view-like UC rows as `MetadataTable + ViewInfo`; 4.0/4.1 stay no-ops.
- **`UCSingleCatalog`** mixes in `TableViewCatalog` + `ViewSupport`, keeps a `ucProxy` handle, and calls the per-version `rejectIfShimmedView` hook from `loadTable`.
- **`UCProxy`** view dispatch: `listUcTables` / `getUcTable` / `loadV1Table` extractions, view filtering in `listTables` / `dropTable`, 409 → already-exists mapping.
- **`UCViewTypes`** (new): the view-type mapping (`viewLikeUcTypes` + predicates) moved **out of `UCSingleCatalog`** so the catalog stays focused on table management (@zhenghu's Slack ask).
- **`UCColumnConversions`**: the view path's `Column` serializer folded in alongside the `StructField` one (from #1644), and the separate `UCColumnJson` deleted (@openinx's unify ask).
- Comment/scaladoc cleanups (#3/#4/#10/#11) and the redundant-`proxy` fix (#9).

### Tests
`UCViewProxySuite` (Spark 4.2 view CRUD), `UCProxySuite` view-filtering additions + shared `UCProxyTestFixture`, `UCColumnConversionsSuite`, and new `UCViewUnsupportedSuite` on **4.0/4.1** asserting metric views are unsupported there (explicit per-version coverage per @zhenghu).

### Stack
1. #1644 — extract `UCColumnConversions`
2. #1645 — shim scaffolding
3. **this PR** — view glue + `UCViewTypes` + unit tests
4. local-run e2e tests

Verified: compiles on Spark 4.0/4.1/4.2; unit suites green on 4.0/4.1; the Spark-4.2 view suites compile against real 4.2 and run green in CI (they require a real Spark 4.2 runtime).

This PR is assisted by Isaac Autopilot: [View the full session spec, test plan, and artifacts](http://go/isaac-autopilot/cli-viewer/1782162701477/chen-wang_data).


connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
```diff
@@ -41,6 +43,77 @@ public static String toStructFieldJson(StructField field) {
...
   }
 
+  /**
+   * Spark V2 {@link Column} -&gt; wire: build the same {@code {name, type, nullable, metadata}}
+   * JSON as {@link #toStructFieldJson} but from a V2 catalog {@code Column}. Unlike a {@code
+   * StructField} (whose {@code metadata} already contains any comment), a V2 {@code Column} exposes
+   * its comment via the dedicated {@code Column.comment()} accessor, so the comment is merged back
+   * under the {@code "comment"} metadata key here to keep the wire shape identical.
+   */
+  public static String buildStructFieldJson(Column col) {
```
openinx: will this be better to use `buildColumnJson` ?

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
```diff
@@ -41,6 +43,77 @@ public static String toStructFieldJson(StructField field) {
...
+  }
+
+  /**
+   * Wire -&gt; Spark V2 {@link Column}: parse a {@code StructField}-shape JSON string back into a
+   * {@code Column}. The {@code "comment"} key (if present in the wire metadata) is lifted out and
+   * passed to {@code Column.create} as the dedicated {@code comment} arg, NOT left in {@code
+   * metadataInJSON} -- matches what {@code CatalogV2Util.structFieldToV2Column} does on the v1
+   * path.
+   */
+  public static Column parseStructFieldJson(String jsonStr) {
```
openinx: parseColumnJson

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
openinx: Also I want to comment, if it's possible for us the do the `Column` -> `StructField`, or vice versa.  so that  can keep only one `buildStructFieldJson/buildColumnJson`,  then we don't have to build the methods for different classes ( StructField or Column), and duplicate the code. 

Just a question, not a blocker

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
```diff
@@ -41,6 +43,77 @@ public static String toStructFieldJson(StructField field) {
...
+    // `ColumnInfo.type_json` is nullable on the wire (e.g. a view-like row created by an older
+    // writer or a non-Spark client that did not round-trip the field). Fail with a clear,
+    // actionable error rather than NPE-ing out of the view-load path.
+    if (jsonStr == null) {
+      throw new IllegalArgumentException(
+          "Column type_json is missing; cannot reconstruct the Spark column for this view");
+    }
+    JsonNode parsed = readTree(jsonStr);
+    JsonNode nameNode = parsed.get("name");
```
openinx: can you move those hardcoded string as a constant , so that different places can always use the same value, rather than hardcode everywhere ?

connectors/spark/src/main/java/io/unitycatalog/spark/UCTableProperties.java
```diff
@@ -13,14 +13,26 @@ private UCTableProperties() {}
   public static final String DELTA_CATALOG_MANAGED_KEY = "delta.feature.catalogManaged";
   public static final String DELTA_CATALOG_MANAGED_VALUE = "supported";
 
-  // These properties were added by `V1Table.addV2TableProperties` in package spark-catalyst.
-  // They are used for constructing the CreateTable rpc.
+  // Spark 4.2 added the `PROP_TABLE_TYPE = "table_type"` constant on `TableCatalog`. We
```
openinx: I think this comment is wrong, since I found the `PROP_TABLE_TYPE` was added from spark 4.1 : https://github.com/apache/spark/blob/branch-4.1/sql/catalyst/src/main/java/org/apache/spark/sql/connector/catalog/TableCatalog.java#L65-L68 

The spark 4.0 actually does not introduce this: https://github.com/apache/spark/blob/branch-4.0/sql/catalyst/src/main/java/org/apache/spark/sql/connector/catalog/TableCatalog.java

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -1,17 +1,298 @@
...
+
+  override def loadView(ident: Identifier): ViewInfo = {
+    // `getUcTable` throws `NoSuchTableException` on a 404. On the view API surface
+    // (`DESCRIBE`/`SHOW`/`DROP VIEW IF EXISTS`) callers branch on `NoSuchViewException`, so
+    // translate the not-found case to the view-typed exception, mirroring `dropView`.
+    val t = try {
+      getUcTable(ident)
+    } catch {
+      case _: NoSuchTableException => throw new NoSuchViewException(ident)
+    }
```
openinx: do you think it will be more simplified if we make the `getUcTable` to return the `Option[UCTableInfo]`,  because we won't need to explicitly catch the `NoSuchTableException` to see if the table is present or absent.  the `Option` already implies it.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -1,17 +1,298 @@
...
+    if (!UCViewTypes.isViewLikeTableType(t.getTableType)) {
```
openinx, line 83: How is the `isViewLikeTableType` differenciating from the `isViewCommandsSupportedTableType`  ?  why do we need two methods to support the similar check ?

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -1,17 +1,298 @@
...
+      // A real table sits at this identifier -- it genuinely isn't a view.
+      throw new NoSuchViewException(ident)
+    }
+    if (!UCViewTypes.isViewCommandsSupportedTableType(t.getTableType)) {
+      // A listed-but-unmapped view kind (e.g. MATERIALIZED_VIEW, `None` in `viewLikeUcTypes`):
```
openinx, line 87: will we handle the `MATERIALZIED_VIEW` in this PR ?

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -1,17 +1,298 @@
...
+  override def replaceView(ident: Identifier, info: ViewInfo): ViewInfo = {
+    throw new UnsupportedOperationException("Replacing a view is not supported yet")
+  }
+
+  override def dropView(ident: Identifier): Boolean = {
+    val t = try {
+      getUcTable(ident)
+    } catch {
+      case _: NoSuchTableException => return false
+    }
```
openinx: Same as above,  if we have the Option return value from the `getUcTable`, then  here will be simplfied.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -1,17 +1,298 @@
...
+          require(fieldNames.length == 1,
+            s"Expected single-field partition reference but got: ${fieldNames.mkString(".")}")
+          Some(fieldNames.head)
+        case "cluster_by" =>
+          None
+        case other =>
+          throw new ApiException(s"Unsupported partition transform: $other")
+      }
+    }.toSeq
+  }
```
openinx: don't we already have the `partitionColumnNames` in the `UCColumnConversoins` ?  why do we need a dup one here .

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -1,9 +1,23 @@
...
+import scala.language.existentials
+
 import io.unitycatalog.client.api.{SchemasApi, TablesApi}
 import io.unitycatalog.client.auth.TokenProvider
-import io.unitycatalog.client.model._
-import io.unitycatalog.client.model.TableInfo
+import io.unitycatalog.client.model.{
+  Dependency => UCDependency,
+  DependencyList => UCDependencyList,
+  TableDependency => UCTableDependency,
```
openinx: do we have real usage in this class  ?  can we remove the 3 references here ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -14,30 +28,24 @@ import io.unitycatalog.spark.utils.OptionsUtil
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
 import org.apache.spark.internal.Logging
-import org.apache.spark.sql.catalyst.TableIdentifier
-import org.apache.spark.sql.catalyst.analysis.{NamespaceAlreadyExistsException, NoSuchNamespaceException, NoSuchTableException}
-import org.apache.spark.sql.catalyst.catalog.{CatalogTable, CatalogTableType, CatalogUtils}
 import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.TableIdentifier
+import org.apache.spark.sql.catalyst.analysis.{NamespaceAlreadyExistsException, NoSuchNamespaceException, NoSuchTableException, TableAlreadyExistsException, ViewAlreadyExistsException}
+import org.apache.spark.sql.catalyst.catalog.{CatalogStorageFormat, CatalogTable, CatalogTableType, CatalogUtils}
```
openinx: any usage of `CatalogStorageFormat`, can we remove ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -81,6 +104,7 @@ class UCSingleCatalog
     val proxy = new UCProxy(uri, tokenProvider, renewCredEnabled, credScopedFsEnabled,
```
openinx: if we plan to introduce this `ucProxy` member in this class, then we can remove this local `proxy` variable, and assign it to the `ucProxy` directly.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -135,7 +159,15 @@ class UCSingleCatalog
 
   override def listTables(namespace: Array[String]): Array[Identifier] = delegate.listTables(namespace)
 
-  override def loadTable(ident: Identifier): Table = delegate.loadTable(ident)
+  // The view-rejection check (`MetadataTable + ViewInfo` flowing back through Delta's
+  // `case o => o` fallthrough on top of `UCProxy.loadTable`) lives in the per-Spark-
+  // version `ViewSupport.rejectIfShimmedView` hook -- a no-op on Spark 4.0/4.1
+  // (no view types exist there) and the real check on Spark 4.2.
```
openinx: why views is exsiting in the catalog right ? not exist in the spark engine. so if the unitycatalog, has the `view` type field for those entities returned from catalog,  then we should always reject the request, and throw the exception. 

then do you think we should remove this `rejectIfShimmedView` as a shimmed methods across different spark versions ?  should we add the code in the UCSingleCatalog directly as for every spark engine we should always reject it.

build.sbt
```diff
@@ -647,7 +647,14 @@ lazy val spark = (project in file("connectors/spark"))
         .files
         .filter(_.getName.contains("lombok"))
         .mkString(File.pathSeparator)
-      javacRelease11 ++ Seq(
+      // Spark 4.2 connector tests reference Java records (e.g. Dependency.table); require --release 17.
+      val testRelease =
+        if (CrossSparkVersions.getSparkVersionSpec().shortVersion == "4.2") {
```
openinx: if the spark version is higher than or equals to 4.2.0, then we should use jdk17, not limited to 4.2.

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
chenwang-databricks: Done — renamed to `buildColumnJson`.

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
chenwang-databricks: Done — renamed to `parseColumnJson`.

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
chenwang-databricks: Good question. It's the cross-Spark type split: `Column` (V2 catalog) only exists on Spark 4.2, while the 4.0/4.1 table-create path only has `StructField`. `UCColumnConversions` lives in the shared source tree (compiled on all three versions), so it can't convert `StructField` <-> `Column` — that would pull a 4.2-only type into the 4.0/4.1 build. So the two serializers are kept deliberately (the `buildColumnInfos` comment notes this). Contrast with `partitionColumnNames`, which *is* shared — precisely because it works on `Transform[]`, a type present on all three versions.

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
chenwang-databricks: Done — added `FIELD_NAME`/`FIELD_TYPE`/`FIELD_NULLABLE`/`FIELD_METADATA`/`FIELD_COMMENT` constants and used them across `toStructFieldJson`/`buildColumnJson`/`parseColumnJson`.

connectors/spark/src/main/java/io/unitycatalog/spark/UCTableProperties.java
chenwang-databricks: You're right — verified against branch-4.1 (present) and branch-4.0 (absent). Fixed the comment to "Spark 4.1 added `PROP_TABLE_TYPE`; we mirror it so the connector compiles on Spark 4.0 too". Also corrected the matching wrong comments in the 4.0/4.1 `UCViewProxySuite` tests.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: Done — `getUcTable` now returns `Option[UCTableInfo]` (`None` on 404). `loadView`/`dropView`/`dropTable` branch on the Option instead of catching `NoSuchTableException`; `loadTable` re-raises `NoSuchTableException` on `None` (Spark's resolver expects that).

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: They gate different stages: `isViewLikeTableType` = surface routing (is this view-like at all — a key in the map, incl. listed-but-unmapped kinds like `MATERIALIZED_VIEW`); `isViewCommandsSupportedTableType` = CRUD gating (stricter — only kinds with a `Some(_)` Spark mapping can actually be loaded/dropped). The gap between them is exactly the listed-but-unmapped kinds, which appear in `listViews` for surface correctness but are rejected from view CRUD. I've tightened the scaladoc on both defs to state the contrast directly.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: No — only `METRIC_VIEW` is mapped in this PR. `MATERIALIZED_VIEW` is intentionally listed-but-unmapped (`None` in `viewLikeUcTypes`): it shows up in `listViews` for surface correctness (so it isn't mis-classified as a table) but load/create/drop reject it as "not supported yet". Turning it on later is a one-line map change once Spark exposes a `TableSummary.MATERIALIZED_VIEW_TABLE_TYPE`.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: Done — `dropView` now branches on the `Option` from `getUcTable` (see the getUcTable change above).

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: Good catch — removed the local `extractPartitionColumnNames` and now reuse `UCColumnConversions.partitionColumnNames(tableInfo.partitions())`, mirroring the shared table path. It's reusable because it operates on `Transform[]`, which exists on all three Spark versions.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: Done — removed the unused `Dependency`/`DependencyList`/`TableDependency` aliases (only `TableInfo => UCTableInfo` is used in this file; the `_` wildcard stays).

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: Done — removed `CatalogStorageFormat` from the import (the other three in that group are used).

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: Done — `initialize` now assigns `ucProxy = new UCProxy(...)` directly and uses `ucProxy` throughout.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: Agreed on the premise (views should always be rejected on every Spark version), but the hook can't be fully inlined: the 4.2 body pattern-matches on `MetadataTable` + `ViewInfo`, types that don't exist on Spark 4.0/4.1 — inlining would break the 4.0/4.1 compile. On 4.0/4.1 a view can never come back as `MetadataTable + ViewInfo` anyway (`UCProxy.loadTable` never produces them), so the no-op is correct there and the shim boundary is load-bearing for compilation, not just behavior. I've expanded the call-site comment to spell this out.

build.sbt
chenwang-databricks: Done — added an `isAtLeast(major, minor)` helper on `SparkVersionSpec` (parses numeric major/minor, avoiding the lexicographic `"4.10" < "4.2"` trap) and changed the gate to `isAtLeast(4, 2)`.

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
```diff
@@ -32,15 +41,86 @@ private UCColumnConversions() {}
...
+   * path.
+   */
+  public static Column parseColumnJson(String jsonStr) {
+    // `ColumnInfo.type_json` is nullable on the wire (e.g. a view-like row created by an older
+    // writer or a non-Spark client that did not round-trip the field). Fail with a clear,
+    // actionable error rather than NPE-ing out of the view-load path.
+    if (jsonStr == null) {
+      throw new IllegalArgumentException(
+          "Column type_json is missing; cannot reconstruct the Spark column for this view");
+    }
```
openinx: I suggest to use the `Precondition.checkNotNull(jsonstr, ...)` directly, which is from the `io.unitycatalog.client.internal.Preconditions`.

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
```diff
@@ -32,15 +41,86 @@ private UCColumnConversions() {}
...
+    JsonNode parsed = readTree(jsonStr);
+    JsonNode nameNode = parsed.get(FIELD_NAME);
+    if (nameNode == null || !nameNode.isTextual()) {
+      throw new IllegalArgumentException(
+          "Expected string `name` in StructField JSON, got: " + nameNode);
+    }
```
openinx: nit: pls use the `Preconditions.checkArgument` .

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
```diff
@@ -32,15 +41,86 @@ private UCColumnConversions() {}
...
+    String name = nameNode.asText();
+
+    // Re-serialize the type sub-node to a compact string for DataType.fromJson.
+    DataType dataType = DataType.fromJson(writeValueAsString(parsed.get(FIELD_TYPE)));
+
+    JsonNode nullableNode = parsed.get(FIELD_NULLABLE);
+    if (nullableNode == null || !nullableNode.isBoolean()) {
+      throw new IllegalArgumentException(
+          "Expected boolean `nullable` in StructField JSON, got: " + nullableNode);
+    }
```
openinx: ditto.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -624,29 +660,58 @@ private[spark] class UCProxy(
...
+    val t = getUcTable(ident).getOrElse(throw new NoSuchTableException(ident))
+    if (UCViewTypes.isViewLikeTableType(t.getTableType)) {
+      wrapAsView(t, ident)
+    } else {
+      loadV1Table(ident, t)
     }
+  }
+
+  private def loadV1Table(
+      ident: Identifier,
```
openinx: nit: is this field useful ? can we remove it ?

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -1,17 +1,278 @@
...
+      .map(table => Identifier.of(namespace, table.getName))
+      .toArray
+  }
+
+  override def loadTableOrView(ident: Identifier): Table = loadTable(ident)
+
+  override def loadView(ident: Identifier): ViewInfo = {
+    // `getUcTable` returns `None` on a 404. On the view API surface
+    // (`DESCRIBE`/`SHOW`/`DROP VIEW IF EXISTS`) callers branch on `NoSuchViewException`, so
+    // translate the not-found case to the view-typed exception, mirroring `dropView`.
```
openinx: nit: this comment seems to be redundent, since the one line code is already straightforward.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -1,17 +1,278 @@
...
+    val t = getUcTable(ident).getOrElse(throw new NoSuchViewException(ident))
+    if (!UCViewTypes.isViewLikeTableType(t.getTableType)) {
+      // A real table sits at this identifier -- it genuinely isn't a view.
```
openinx: ditto.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
openinx: okay, the `isViewLikeTableType` is only used for identify if it's a view. but the `isViewCommandsSupportedTableType` is used for checking if the CRUD are supported.   

For the former one, will it support the broader views such as permenant view, global view, temp view ?

openinx: we will have a plan to suppor those three basic view in the coming release.

openinx (changes requested):
few minor comments.

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
chenwang-databricks: Done — switched to `Preconditions.checkNotNull(jsonStr, ...)` (from `io.unitycatalog.client.internal.Preconditions`). Note this now throws `NullPointerException` (its documented behavior) rather than `IllegalArgumentException`; updated the null-case unit test accordingly.

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
chenwang-databricks: Done — `Preconditions.checkArgument(nameNode != null && nameNode.isTextual(), ...)`.

connectors/spark/src/main/java/io/unitycatalog/spark/UCColumnConversions.java
chenwang-databricks: Done — `Preconditions.checkArgument(nullableNode != null && nullableNode.isBoolean(), ...)`.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: Good catch — `loadV1Table` never used its `ident` param (it builds the `TableIdentifier` from the `UCTableInfo`). Dropped the param and updated the caller.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: Done — removed; the `.getOrElse(throw new NoSuchViewException(ident))` is self-explanatory.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: Done — removed. (Kept the comment on the third branch, since the "listed-but-unmapped view kind -> UnsupportedOperationException, not NoSuchView" rationale is less obvious.)

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCProxyViewSupport.scala
chenwang-databricks: Those are Spark engine-side view kinds; `isViewLikeTableType` only classifies UC `TableType`s returned by the catalog (`METRIC_VIEW`, `MATERIALIZED_VIEW`). Permanent/global/temp views aren't UC catalog table types, so they never reach this predicate — it's specifically about UC view-like table types.

chenwang-databricks: Sounds good — that's out of scope for this PR (metric views only). The `viewLikeUcTypes` map + these predicates are set up so adding a kind later is a one-line change.

openinx (approved):
looks good to me now.  thanks @chenwang-databricks for the work.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -723,6 +786,10 @@ private[spark] class UCProxy(
       .asInstanceOf[Table]
   }
 
+  // Spark 4.2's `TableViewCatalog` made `createTable(ident, TableInfo)` the canonical entry
```
tdas: method docs should use /** */

tdas: or move this to be inside the method

tdas:
the `spark-4.0` and `spark-4.1` shim trees are byte-identical copies — `UCProxyViewSupport` / `ViewSupport` / `TableViewCatalog` + `UCViewProxySuite`, ~200 lines. this PR even edited both `UCProxyViewSupport` copies in lockstep, which is the smell. the real axis here is capability (pre-view-API vs 4.2), not version.

can we collapse 4.0 and 4.1 into one shared shim dir? `additionalSourceDir` is just a path string, so both versions can point at e.g. `scala-shims/spark-below-4.2` in `spark-versions.json`, keep one copy, and delete the dupes — no change to `CrossSparkVersions`. if 4.0/4.1 ever need to diverge we go tiered (a list of dirs: shared + version-specific — the delta pattern the header says this was trimmed from), but thats YAGNI till then. only constraint: one class per included dir, else dup-class errors.

also while you're in there: this suite's comment says `PROP_TABLE_TYPE` "was added in Spark 4.1" but `UCTableProperties` says "Spark 4.2 added it" — one of them is wrong.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: Done — converted the `//` block to a `/** */` Scaladoc comment above the override (kept it as method doc rather than moving it inside the body).

chenwang-databricks:
Done — collapsed the byte-identical `spark-4.0` and `spark-4.1` shim trees into a single shared `scala-shims/spark-4.0-4.1` dir (main + test), following Yi's Delta dedup pattern ([delta-io/delta#7202](https://github.com/delta-io/delta/pull/7202)). `UCProxyViewSupport` / `ViewSupport` / `TableViewCatalog` + `UCViewProxySuite` now live once, with the version-literal comments generalized to "Spark 4.0/4.1".

Rather than repointing a single `sourceDir` string, I restored the tiered form `CrossSparkVersions` was originally trimmed from: `sourceDir: String` → `sourceDirs: Seq[String]`, and each version lists its shim dirs (4.0 and 4.1 → `["scala-shims/spark-4.0-4.1"]`, 4.2 → `["scala-shims/spark-4.2"]`). Keeps the "capability, not version" axis explicit and makes future divergence or 3-way sharing (a `spark-4.1-4.2` dir) a one-line JSON change — matching the Delta convention. `build.sbt` needs no change.

On the `PROP_TABLE_TYPE` note: the collapse removes the duplicate copy, so it's now single-sourced. Both `UCTableProperties` and the (now shared) test comment say it was added in **Spark 4.1** — verified against Spark sources (present on branch-4.1, absent on branch-4.0).

Verified compile + Test/compile + testJavastyle + unit suites green on all three versions (4.0/4.1 now pull shims from the shared dir; 4.2 from its own): 4.2 29/29, 4.1 22/22, 4.0 22/22.

connectors/spark/src/main/scala-shims/spark-4.1/org/apache/spark/sql/connector/catalog/TableViewCatalog.scala
```diff
@@ -1,16 +0,0 @@
-package org.apache.spark.sql.connector.catalog
```
tdas, line 1: why are we removing this?

connectors/spark/src/test/scala-shims/spark-4.0-4.1/io/unitycatalog/spark/UCViewProxySuite.java
```diff
@@ -0,0 +1,155 @@
...
+ * {@code UCProxyViewSupport} shim is a no-op whose {@code wrapAsView} hook throws, and the
+ * {@code with TableViewCatalog} mixin resolves to a bare stub with no view CRUD surface.
+ *
+ * <p>So on Spark 4.0/4.1 there is nothing like {@code createView}/{@code loadView} to call; this
+ * suite pins that view-like UC rows ({@code METRIC_VIEW} and {@code MATERIALIZED_VIEW}) are inert
+ * through every {@code TableCatalog} operation, and that the generic table operations offer no
+ * back-door to view creation. On Spark 4.2 the mapped {@code METRIC_VIEW} behaves differently (see
+ * the 4.2 {@code UCViewProxySuite}).
+ */
+public class UCViewProxySuite {
```
tdas: since this is testing completely different thing that the 4.2 version .. pelase name it differently

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/ViewSupport.scala
```diff
@@ -1,15 +1,70 @@
...
+ * `delegate` and `ucProxy` state (declared `protected[spark]` for that purpose).
+ *
+ * The companion `rejectIfShimmedView` hook is the Spark-4.2-only half of the catalog's
+ * `loadTable` orthogonality check: when Delta's `case o => o` fallthrough returns a
+ * `MetadataTable + ViewInfo` for a view-like row, this hook converts it to the
+ * `NoSuchTableException` the Spark resolver expects from a strict table-side load.
  */
-trait ViewSupport { self: UCSingleCatalog =>
-  protected def rejectIfShimmedView(t: Table, ident: Identifier): Unit = ()
+trait ViewSupport extends TableViewCatalog { self: UCSingleCatalog =>
```
tdas: rename to UCSingleCatalogViewSupport

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -624,29 +660,56 @@ private[spark] class UCProxy(
...
   }
 
-  override def loadTable(ident: Identifier): Table = {
-    val t = try {
-      tablesApi.getTable(
+  // Returns `None` on a 404 so callers can branch on presence/absence without catching an
+  // exception. `loadTable` re-raises `NoSuchTableException` on `None` (Spark's resolver expects
+  // that); the passive-filter paths (`dropTable`, `dropView`) and `loadView` map `None` to their
+  // own not-found result.
+  private[spark] def getUcTable(ident: Identifier): Option[UCTableInfo] = {
```
tdas: getUCTableLike

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -624,29 +660,56 @@ private[spark] class UCProxy(
...
     UCSingleCatalog.checkUnsupportedNestedNamespace(namespace)
+    // Per the TableViewCatalog contract, listTables must return tables only -- view-like UC
+    // table types (metric views, materialized views) belong on listViews.
+    listUcTables(namespace)
+      .filterNot(t => UCViewTypes.isViewLikeTableType(t.getTableType))
+      .map(table => Identifier.of(namespace, table.getName))
+      .toArray
+  }
 
+  private[spark] def listUcTables(namespace: Array[String]): Seq[UCTableInfo] = {
```
tdas: listUCTableLikes

tdas (approved):
lgtm, but please address the comments.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: Done — renamed to `getUCTableLike` (it resolves view-like rows too, not just tables).

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
chenwang-databricks: Done — renamed to `listUCTableLikes`.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/ViewSupport.scala
chenwang-databricks: Done — renamed the trait (and its file) to `UCSingleCatalogViewSupport` in both the 4.2 and 4.0-4.1 shims, and updated the `with ...` mixin on `UCSingleCatalog`. Now symmetric with the sibling `UCProxyViewSupport` (self: UCProxy) — this one is self: UCSingleCatalog.

connectors/spark/src/test/scala-shims/spark-4.0-4.1/io/unitycatalog/spark/UCViewProxySuite.java
chenwang-databricks: Done — renamed to `UCViewUnsupportedSuite`, since it pins that views are unsupported/inert through the TableCatalog surface on pre-view-API Spark. The 4.2 `UCViewProxySuite` keeps its name (it exercises the real view CRUD).

connectors/spark/src/main/scala-shims/spark-4.1/org/apache/spark/sql/connector/catalog/TableViewCatalog.scala
chenwang-databricks: It's not removed — it moved. When the byte-identical `spark-4.0` and `spark-4.1` shim dirs were collapsed into the shared `scala-shims/spark-4.0-4.1/`, the `TableViewCatalog` stub went with them (now at `scala-shims/spark-4.0-4.1/org/apache/spark/sql/connector/catalog/TableViewCatalog.scala`). GitHub paired the *4.0* copy as the rename source, so the *4.1* copy renders as a bare deletion — but both 4.0 and 4.1 still resolve `with TableViewCatalog` from the shared dir (verified: all three versions compile green). Same rename-pairing artifact as the earlier `.gitkeep` deletions.

openinx (approved):

PR #1647, opened 2026-07-02
[Spark] Local-run e2e test for metric views (Spark 4.2)

chenwang-databricks:
Now that the Spark 4.2.0 connector port has merged (#1675), this PR is rebased directly onto `main` and its diff is just this test plus a small build fix it surfaced. Verified passing end-to-end against published Spark 4.2.0.

### What this PR does
Adds `MetricViewE2ETest` — a local-run end-to-end test that exercises the full connector → UC-server round-trip for metric views against an **embedded** `UnityCatalogServer` via `UCSingleCatalog` (real HTTP + persistence, not mocks).

Metric views are created with the **actual SQL DDL** a user runs — `CREATE VIEW ... WITH METRICS LANGUAGE YAML AS $$...$$` (SPARK-56920). On a DS v2 catalog like `UCSingleCatalog`, Spark's `DataSourceV2Strategy` routes the parsed `CreateMetricView` plan through `CreateV2MetricViewExec`, which builds the Spark 4.2 V2 `View` (metric-view type + query text + typed table dependencies) and calls the connector's `createView`. The test drives that whole path — parse → analyze → route → persist — plus `SHOW VIEWS`, `SHOW TABLES`, a `SELECT` through the view, and `DROP VIEW`.

The `SELECT` reads a measure via the metric-view `measure(...)` function with `GROUP BY` on the dimension (`SELECT region, measure(cnt_sum) FROM mv GROUP BY region`), which Spark rewrites into the underlying aggregation over the source table.

Spark-4.2-only (lives under `src/test/scala-shims/spark-4.2/`), where the metric-view DDL and view-catalog APIs exist; it therefore compiles/runs only on the Spark 4.2 build, not 4.0/4.1.

### Also included: a jackson-core fix surfaced by the test
On published Spark 4.2.0 the spark module's `jackson-core` `dependencyOverride` (pinned to `2.15.0`) is too old for Spark 4.2's `jackson-dataformat-yaml` (2.21.2), which calls `YAMLParser._updateToken(JsonToken)` — a method added in jackson-core 2.18 — while parsing the metric-view YAML. This throws `NoSuchMethodError` on `CREATE VIEW ... WITH METRICS`. Fix: bump only the spark-module `jackson-core` override to `2.19.2` (the highest core with a matching published `jackson-annotations`, and already what Spark 4.0/4.1/4.2 resolve); the other jackson modules stay at 2.15.0.

### Note on `SHOW TABLES`
On Spark 4.2 a `RelationCatalog` resolves `SHOW TABLES` through `listRelationSummaries`, whose default unions tables and views (views carry a `VIEW` table-type in the summary). So a metric view correctly appears on **both** `SHOW VIEWS` and `SHOW TABLES`; the test asserts that.

### Stack
1. #1644 — extract `UCColumnConversions` (merged)
2. #1645 — shim scaffolding (merged)
3. #1646 — view glue + `UCViewTypes` + unit tests (merged)
4. #1675 — upgrade the connector to Apache Spark 4.2.0 (`RelationCatalog`/`View`/`loadRelation`) (merged)
5. **this PR** — local-run e2e test for metric views + the jackson-core fix above

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -699,7 +699,8 @@ lazy val spark = (project in file("connectors/spark"))
       "com.fasterxml.jackson.core" % "jackson-databind" % "2.15.0",
       "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.15.0",
       "com.fasterxml.jackson.core" % "jackson-annotations" % "2.15.0",
-      "com.fasterxml.jackson.core" % "jackson-core" % "2.15.0",
+      // jackson-core >= 2.18 required by Spark 4.2's jackson-dataformat-yaml (YAMLParser._updateToken).
+      "com.fasterxml.jackson.core" % "jackson-core" % "2.19.2",
       "com.fasterxml.jackson.dataformat" % "jackson-dataformat-xml" % "2.15.0",
       "org.antlr" % "antlr4-runtime" % "4.13.1",
       "org.antlr" % "antlr4" % "4.13.1",
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java
@@ -1,0 +1,140 @@
+package io.unitycatalog.spark;
+
+import static io.unitycatalog.server.utils.TestUtils.CATALOG_NAME;
+import static io.unitycatalog.server.utils.TestUtils.SCHEMA_NAME;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.io.File;
+import java.util.List;
+import org.apache.spark.sql.Row;
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * End-to-end test for metric views: drives the real {@code CREATE VIEW ... WITH METRICS} SQL DDL
+ * (SPARK-56920) through the connector against an embedded {@code UnityCatalogServer}. Skipped on
+ * Spark below 4.2, where the metric-view DDL does not exist.
+ */
+public class MetricViewE2ETest extends BaseSparkIntegrationTest {
+
+  @TempDir private File sourceTableDir;
+
+  /** Metric views require Spark 4.2+; skip (rather than fail) the whole class on older runtimes. */
+  @BeforeAll
+  public static void assumeMetricViewsSupported() {
+    Assumptions.assumeTrue(
+        isSparkAtLeast(4, 2),
+        "Metric views require Spark 4.2+, but runtime is "
+            + org.apache.spark.package$.MODULE$.SPARK_VERSION());
+  }
+
+  private String tbl(String name) {
+    return CATALOG_NAME + "." + SCHEMA_NAME + "." + name;
+  }
+
+  /**
+   * Creates the UC-backed source table the metric view reads from. The metric-view CREATE path
+   * eagerly resolves the YAML {@code source:} relation, so this table must exist first. A plain
+   * parquet external table keeps the source setup independent of the Delta create path.
+   */
+  private void createEventsSource() {
+    sql(
+        "CREATE TABLE %s (region STRING, cnt INT) USING parquet LOCATION '%s'",
+        tbl("events"), sourceTableDir.toURI());
+    sql("INSERT INTO %s VALUES ('us', 1), ('us', 2), ('eu', 3)", tbl("events"));
+  }
+
+  /** {@code CREATE VIEW ... WITH METRICS LANGUAGE YAML AS $$...$$} against the UC v2 catalog. */
+  private void createMetricView(String name) {
+    sql(
+        "CREATE VIEW %s\n"
+            + "WITH METRICS\n"
+            + "LANGUAGE YAML\n"
+            + "AS\n"
+            + "$$\n"
+            + "version: \"0.1\"\n"
+            + "source: %s\n"
+            + "dimensions:\n"
+            + "  - name: region\n"
+            + "    expr: region\n"
+            + "measures:\n"
+            + "  - name: cnt_sum\n"
+            + "    expr: sum(cnt)\n"
+            + "$$",
+        tbl(name), tbl("events"));
+  }
+
+  @Test
+  public void testMetricViewOperations() throws Exception {
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG, CATALOG_NAME);
+    createEventsSource();
+
+    // CREATE VIEW ... WITH METRICS -> CreateV2MetricViewExec -> connector.createView, persisted
+    // by the embedded UC server.
+    createMetricView("mv_e2e");
+
+    // SHOW VIEWS shows views only: the metric view is listed, the source table `events` is not.
+    List<String> views =
+        sql("SHOW VIEWS IN %s.%s", CATALOG_NAME, SCHEMA_NAME).stream()
+            .map(r -> r.getString(1))
+            .collect(java.util.stream.Collectors.toList());
+    assertThat(views).contains("mv_e2e").doesNotContain("events");
+
+    // SHOW TABLES unions tables and views on Spark 4.2 (a RelationCatalog resolves it through
+    // `listRelationSummaries`), so both the source table and the metric view are listed.
+    List<String> tables =
+        sql("SHOW TABLES IN %s.%s", CATALOG_NAME, SCHEMA_NAME).stream()
+            .map(r -> r.getString(1))
+            .collect(java.util.stream.Collectors.toList());
+    assertThat(tables).contains("events", "mv_e2e");
+
+    // DESCRIBE resolves the ident through the connector's `loadRelation`: `delegate.loadTable`
+    // throws NoSuchTableException for a view, so it falls back to `ucProxy.loadView` and returns
+    // the metric view's columns (the dimension + measure) with their resolved data types.
+    List<Row> describe = sql("DESCRIBE %s", tbl("mv_e2e"));
+    assertThat(describe.stream().map(r -> r.getString(0) + ":" + r.getString(1)))
+        .contains("region:string", "cnt_sum:bigint");
+
+    // DESCRIBE EXTENDED additionally reports the detailed view info -- notably that the object is
+    // a METRIC_VIEW -- alongside the same column rows.
+    List<Row> descExt = sql("DESCRIBE EXTENDED %s", tbl("mv_e2e"));
+    assertThat(descExt.stream().map(r -> r.getString(0) + ":" + r.getString(1)))
+        .contains("region:string", "cnt_sum:bigint", "Type:METRIC_VIEW");
+
+    // SELECT through the metric view aggregates the source rows (eu=3, us=1+2=3). A measure is read
+    // via the `measure(...)` function with GROUP BY on the dimensions.
+    List<Row> rows =
+        sql(
+            "SELECT region, measure(cnt_sum) FROM %s GROUP BY region ORDER BY region",
+            tbl("mv_e2e"));
+    assertThat(rows).hasSize(2);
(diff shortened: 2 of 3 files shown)


connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java
```diff
@@ -0,0 +1,114 @@
+package io.unitycatalog.spark;
```
openinx, line 1: Pls don't add the test in the `scala-shims` dir,  that does not make sense, since those are not shimmed code. 

instead, we'd move those to the `unitycatalog/connectors/spark/src/test/java/io/unitycatalog/spark` dir, and detect the spark runtime version,  if the version is below 4.2, then we should skip the tests directly.  

with this, we won't need to duplicate this code for every spark versions, (if we want to support the spark 4.3, 4.4, etc).

connectors/spark/src/test/scala-shims/spark-4.2/io/unitycatalog/spark/MetricViewE2ETest.java
```diff
@@ -0,0 +1,114 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.File;
+import java.util.List;
+import org.apache.spark.sql.Row;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * Spark-4.2-only end-to-end test for metric views: exercises the full connector -&gt; UC-server
```
openinx: This javadoc is verbose, I suggest to remove it entirely.

build.sbt
```diff
@@ -699,7 +699,12 @@ lazy val spark = (project in file("connectors/spark"))
       "com.fasterxml.jackson.core" % "jackson-databind" % "2.15.0",
       "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.15.0",
       "com.fasterxml.jackson.core" % "jackson-annotations" % "2.15.0",
-      "com.fasterxml.jackson.core" % "jackson-core" % "2.15.0",
+      // jackson-core must be >= 2.18 so Spark 4.2's jackson-dataformat-yaml (2.21.2) finds
+      // YAMLParser._updateToken(JsonToken); the older 2.15.0 pin (kept for the other jackson
+      // modules) lacks it and breaks `CREATE VIEW ... WITH METRICS LANGUAGE YAML` parsing. 2.19.2
+      // is the highest core with a matching published jackson-annotations, and is already what
+      // Spark 4.0/4.1/4.2 resolve, so it is consistent across all cross-Spark lanes.
```
openinx: comment looks verbose, I think we can remove it.

connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java
```diff
@@ -0,0 +1,114 @@
...
+            + "    expr: region\n"
+            + "measures:\n"
+            + "  - name: cnt_sum\n"
+            + "    expr: sum(cnt)\n"
+            + "$$",
+        tbl(name), tbl("events"));
+  }
+
+  @Test
+  public void testMetricViewCreateQueryDropRoundTripViaSql() throws Exception {
```
openinx: This e2e test coverage may not enough,  after I took a look for this code again. 
https://github.com/unitycatalog/unitycatalog/blob/bebab910ed2836b295b0319b3507d570e588f8e3/connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCSingleCatalogViewSupport.scala#L25-L58

the extra cases that come in my mind should be 

1.  rename a metric view. 
2.  the e2e test case to cover the `loadRelation`, I mean if the loadTable throw tableNotFound, and then fallback to use the `loadView`.

chenwang-databricks:
Thanks @openinx — addressed the comments in the latest push:

- **Moved the test out of `scala-shims`** into `connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java` and gated it at runtime on `SPARK_VERSION >= 4.2` via JUnit `Assumptions.assumeTrue` (using `org.apache.spark.package$.MODULE$.SPARK_VERSION()`). It now compiles on all Spark versions and simply **skips** below 4.2 — no per-version duplication. Verified: 4.2.0 runs (4 tests pass), 4.0.0 / 4.1.0 compile + checkstyle clean + all tests skipped (aborted assumption), not failed.
- **Removed the verbose class javadoc** and **trimmed the `build.sbt` jackson-core comment** to one line.
- **Added coverage**: (1) `DESCRIBE` on a metric view, which exercises `loadRelation`'s `delegate.loadTable` → `NoSuchTableException` → `ucProxy.loadView` fallback; (2) `ALTER VIEW ... RENAME TO` on a metric view, asserted to fail (the connector's `renameView` is intentionally unsupported — no UC rename endpoint yet).

**On the two red checks** — both are `[Non Blocking]` Delta-master lanes and are **not caused by this PR** (all blocking lanes are green):
- `Spark 4.2.0 × Delta master-SNAPSHOT`: the Delta source build (`build_delta.sh`) fails resolving `org.apache.spark:spark-catalyst_2.13:4.2.0-SNAPSHOT`, which no longer exists now that Spark 4.2.0 is released — an upstream Delta-master issue; the build dies before the connector tests run.
- `Spark 4.1.0 × Delta master-SNAPSHOT`: the failures are all `ServerSidePlanningTest`, the known Delta-master flake.

openinx (approved):
LGMT,  thanks @chenwang-databricks

connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java
```diff
@@ -0,0 +1,138 @@
...
+ * End-to-end test for metric views: drives the real {@code CREATE VIEW ... WITH METRICS} SQL DDL
+ * (SPARK-56920) through the connector against an embedded {@code UnityCatalogServer}. Skipped on
+ * Spark below 4.2, where the metric-view DDL does not exist.
+ */
+public class MetricViewE2ETest extends BaseSparkIntegrationTest {
+
+  @TempDir private File sourceTableDir;
+
+  /** Metric views require Spark 4.2+; skip (rather than fail) on older runtimes. */
+  @BeforeEach
```
yili-db: Can this be `@BeforeAll`?

connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java
```diff
@@ -0,0 +1,138 @@
...
+  public void assumeMetricViewsSupported() {
+    Assumptions.assumeTrue(
+        isSparkAtLeast(4, 2),
+        "Metric views require Spark 4.2+, but runtime is "
+            + org.apache.spark.package$.MODULE$.SPARK_VERSION());
+  }
+
+  private static boolean isSparkAtLeast(int major, int minor) {
```
yili-db: Please move this function to the base class

connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Creates the UC-backed source table the metric view reads from. The metric-view CREATE path
+   * eagerly resolves the YAML {@code source:} relation, so this table must exist first. A plain
+   * parquet external table keeps the source setup independent of the Delta create path.
+   */
+  private void createEventsSource() {
+    sql(
+        "CREATE TABLE %s (region STRING, cnt INT) USING parquet LOCATION '%s'",
```
yili-db, line 45: Any reason using parquet? Can it use Delta? And a managed Delta table?

connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java
```diff
@@ -0,0 +1,138 @@
...
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG, CATALOG_NAME);
+    createEventsSource();
+
+    // CREATE VIEW ... WITH METRICS -> CreateV2MetricViewExec -> connector.createView, persisted
+    // by the embedded UC server.
+    createMetricView("mv_e2e");
+
+    // SHOW VIEWS surfaces the metric view on the view surface.
+    List<Row> views = sql("SHOW VIEWS IN %s.%s", CATALOG_NAME, SCHEMA_NAME);
+    assertThat(views.stream().anyMatch(r -> r.getString(1).equals("mv_e2e"))).isTrue();
```
yili-db: `show views` should only show views? Then it should assert that the source table ins't in result.

connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    // SHOW TABLES also surfaces it: on Spark 4.2 a RelationCatalog resolves SHOW TABLES through
+    // `listRelationSummaries`, which unions tables and views, so a metric view appears on both.
+    List<Row> tables = sql("SHOW TABLES IN %s.%s", CATALOG_NAME, SCHEMA_NAME);
+    assertThat(tables.stream().anyMatch(r -> r.getString(1).equals("mv_e2e"))).isTrue();
```
yili-db: assert both the table and view.

connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java
```diff
@@ -0,0 +1,138 @@
...
+    assertThat(rows.get(0).getString(0)).isEqualTo("eu");
+    assertThat(rows.get(0).getLong(1)).isEqualTo(3L);
+    assertThat(rows.get(1).getString(0)).isEqualTo("us");
+    assertThat(rows.get(1).getLong(1)).isEqualTo(3L);
+
+    // ALTER VIEW ... RENAME TO delegates to UCSingleCatalogViewSupport.renameView ->
+    // ucProxy.renameView, which is unsupported today, so the rename must surface an error rather
+    // than silently succeed -- the original view stays and the target is not created.
+    assertThatThrownBy(() -> sql("ALTER VIEW %s RENAME TO %s", tbl("mv_e2e"), tbl("mv_renamed")))
+        .isInstanceOf(Exception.class);
```
yili-db: better assert on the details of failure

connectors/spark/src/test/java/io/unitycatalog/spark/MetricViewE2ETest.java
yili-db: I see. It seems Delta isn't ready for 4.2 yet.

tdas (approved):

chenwang-databricks:
Thanks @yili-db — addressed all the comments in the latest push:

- **`@BeforeAll`**: changed the version gate from `@BeforeEach` to a static `@BeforeAll`, so the whole class is aborted once (`Canceled`) on Spark < 4.2 instead of per-test.
- **Move the version helper to the base class**: `isSparkAtLeast(major, minor)` now lives in `BaseSparkIntegrationTest` (reusable by other version-gated tests).
- **parquet vs Delta**: kept `USING parquet` — as you noted, no published Delta targets Spark 4.2 yet, so a managed Delta source can't be used in this test today. Parquet also keeps the source setup independent of the Delta create path. Can revisit once Delta-for-4.2 ships.
- **`SHOW VIEWS` should only show views**: now asserts the metric view is listed **and** the source table `events` is **not** (`.contains("mv_e2e").doesNotContain("events")`).
- **`SHOW TABLES` — assert both table and view**: now asserts it contains **both** `events` (table) and `mv_e2e` (view), matching the Spark 4.2 `listRelationSummaries` union.
- **Assert on the rename failure detail**: now asserts the specific `UnsupportedOperationException` with message containing `"Renaming a view is not supported"`, rather than a bare `Exception`.

Also folded the three metric-view test methods into a single `testMetricViewOperations` and added richer `DESCRIBE` / `DESCRIBE EXTENDED` assertions (column data types + `Type:METRIC_VIEW`). Verified: 4.2.0 passes; 4.0.0 / 4.1.0 compile + checkstyle clean + the class is skipped.

yili-db (approved):

PR #1652, opened 2026-07-03
[Hadoop] Key credential caches by full credential context (catalogUri, scheme, auth)

openinx:
**PR Checklist**

**Description of changes**

Multi-tenant Spark Connect can route queries from different tenants through the same gateway with different `TokenProvider` auth configs. Previously, `CredId` was keyed only by table/path/operation, so `CredentialCache` and `CredScopedFileSystem` could reuse vended credentials and filesystem delegates across tenants targeting the same resource — a permission leak (e.g. tenant B receiving tenant A's READ_WRITE credential for the same table).

This change keys every credential cache by the full credential-vending *context* rather than the resource alone. A vended credential is a function of the catalog endpoint, the storage scheme, and the caller's auth identity, so all three are hashed into a single `credContextId`:

- Adds `MapIdGenerator`, which derives a stable id from a `Map<String, String>` using an order-independent, length-prefixed SHA-256 digest computed incrementally (no external dependency).
- Adds `credContextId` as the first field on `TableCredId`, `PathCredId`, `DeltaTableCredId`, and `DeltaStagingTableCredId`, computed from `catalogUri`, `scheme`, and `TokenProvider.configs()`.
- Wires `credContextId` generation into all four `CredPropsUtil.create*CredProps` entry points.
- Persists the context id in Hadoop props under `fs.unitycatalog.cred.context.id` and round-trips it via `CredId.create(conf)`. The key is intentionally all-lowercase and dot-separated (matching `table.id`, `staging.table.id`) because these props flow through Spark's case-insensitive table option handling, which lowercases option keys.

Folding `catalogUri` in (not just the auth config) also isolates the case where one identity targets two different catalogs — e.g. cross-catalog data sync — which an auth-only id could not distinguish for `TableCredId`, since it carries no location.

`readCredContextId` fails fast if the context id is absent from the config, rather than falling back to a shared default id that could silently let unrelated sessions reuse each other's credentials. Every `CredId` variant persists the context id via `props()`, so all runtime paths supply it.

**Testing**

`./build/sbt -mem 4096 hadoop/test` (164 tests pass)

- Unit: `MapIdGeneratorTest` (stability, order-independence, boundary/unicode/null cases); `CredIdTest` (context dimension in equals/hashCode/props round-trip for all four variants, plus fast-fail when the context id is missing).
- Driver cache: `CredPropsUtilTest` isolates credentials across auth configs and across catalog URIs (all four `CredId` variants), and verifies `credContextId` varies by catalogUri, scheme, and auth.
- Worker global cache: `BaseTokenProviderTest` across S3/ABFS/GCS providers.
- Delegate cache: `CredScopedFileSystemCacheTest`.
- End-to-end: verified `ParquetExternalTableReadWriteTest` and `DeltaExternalTableReadWriteTest` pass, exercising the context id round-trip through Spark's table option handling for both native Parquet and Delta reads.

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -516,9 +516,8 @@ public class CredPropsUtil {
   /**
    * Derives a stable id for the credential context so caches only reuse a vended credential across
    * requests that would receive the same one. A vended credential is a function of the catalog
-   * endpoint, storage scheme, and auth identity, so all three are folded into the hash. The {@code
-   * TokenProvider} config keys ({@code type}, {@code token}, {@code oauth.*}, ...) never overlap
-   * with the catalog and scheme keys.
+   * endpoint, storage scheme, and auth identity, so all three are folded into the hash alongside
+   * the auth configs from {@code tokenProvider}.
    */
   static String credContextId(String catalogUri, String scheme, TokenProvider tokenProvider) {
     Objects.requireNonNull(catalogUri, "catalogUri is required");
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -19,10 +19,12 @@ import io.unitycatalog.hadoop.internal.id.DeltaTableCredId;
 import io.unitycatalog.hadoop.internal.id.PathCredId;
 import io.unitycatalog.hadoop.internal.id.TableCredId;
 import io.unitycatalog.hadoop.internal.util.ClockUtil;
+import io.unitycatalog.hadoop.internal.util.MapIdGenerator;
 import java.net.URI;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import org.apache.hadoop.conf.Configuration;
 
 /**
@@ -360,7 +362,7 @@ public class CredPropsUtil {
         catalogUri,
         tokenProvider,
         appVersions,
-        new TableCredId(tableId, tableOp.value()));
+        new TableCredId(authUniqueId(tokenProvider), tableId, tableOp.value()));
   }
 
   /**
@@ -389,7 +391,7 @@ public class CredPropsUtil {
         catalogUri,
         tokenProvider,
         appVersions,
-        new DeltaTableCredId(identifier, tableOp.value(), location));
+        new DeltaTableCredId(authUniqueId(tokenProvider), identifier, tableOp.value(), location));
   }
 
   /**
@@ -417,7 +419,7 @@ public class CredPropsUtil {
         catalogUri,
         tokenProvider,
         appVersions,
-        new DeltaStagingTableCredId(stagingTableId, location));
+        new DeltaStagingTableCredId(authUniqueId(tokenProvider), stagingTableId, location));
   }
 
   /** Fetches path credentials from the UC REST API and builds Hadoop configuration properties. */
@@ -442,7 +444,7 @@ public class CredPropsUtil {
         catalogUri,
         tokenProvider,
         appVersions,
-        new PathCredId(path, pathOp.value()));
+        new PathCredId(authUniqueId(tokenProvider), path, pathOp.value()));
   }
 
   /**
@@ -499,6 +501,11 @@ public class CredPropsUtil {
       default:
         return Collections.emptyMap();
     }
+  }
+
+  private static String authUniqueId(TokenProvider tokenProvider) {
+    Objects.requireNonNull(tokenProvider, "tokenProvider is required");
+    return MapIdGenerator.generateId(tokenProvider.configs());
   }
 
   private static GenericCredential fetchGenericCredential(
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -516,11 +516,11 @@ public class CredPropsUtil {
   /**
    * Derives a stable id for the credential context so caches only reuse a vended credential across
    * requests that would receive the same one. A vended credential is a function of the catalog
-   * endpoint, storage scheme, and auth identity, so all three are folded into the hash. The auth
-   * configs are namespaced under {@code auth.} to avoid colliding with the other two keys.
+   * endpoint, storage scheme, and auth identity, so all three are folded into the hash. The {@code
+   * TokenProvider} config keys ({@code type}, {@code token}, {@code oauth.*}, ...) never overlap
+   * with the catalog and scheme keys.
    */
-  private static String credContextId(
-      String catalogUri, String scheme, TokenProvider tokenProvider) {
+  static String credContextId(String catalogUri, String scheme, TokenProvider tokenProvider) {
     Objects.requireNonNull(catalogUri, "catalogUri is required");
     Objects.requireNonNull(scheme, "scheme is required");
     Objects.requireNonNull(tokenProvider, "tokenProvider is required");
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -68,6 +68,10 @@ public class CredPropsUtil {
   private static final String GCS_CONFLICT_CHECK_KEY = "fs.gs.create.items.conflict.check.enable";
   private static final String ABFS_FIXED_SAS_TOKEN_KEY = "fs.azure.sas.fixed.token";
 
+  // Keys used to build the credential-context id (see #credContextId).
+  private static final String CRED_CONTEXT_CATALOG_URI_KEY = "catalogUri";
+  private static final String CRED_CONTEXT_SCHEME_KEY = "scheme";
+
   private abstract static class PropsBuilder<T extends PropsBuilder<T>> {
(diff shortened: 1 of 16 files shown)


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/MapIdGenerator.java
```diff
@@ -0,0 +1,78 @@
...
+  private static String hashCanonical(byte[] canonical) {
+    try {
+      MessageDigest digest = MessageDigest.getInstance("SHA-256");
+      return toHex(digest.digest(canonical));
+    } catch (NoSuchAlgorithmException e) {
+      throw new IllegalStateException("SHA-256 not available", e);
+    }
+  }
+
+  private static String toHex(byte[] bytes) {
```
yili-db, line 57:
```
import org.apache.commons.codec.binary.Hex;
String hex = Hex.encodeHexString(bytes);
```
It's likely already part of the dep due to hadoop dependency

yili-db: Or maybe the entire hash and hex can be done with:
```
import org.apache.commons.codec.digest.DigestUtils;

// One line, zero boilerplate, internally optimized
String sha256Hex = DigestUtils.sha256Hex(input);
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -501,6 +503,11 @@ private static Map<String, String> createCredProps(
     }
   }
 
+  private static String authUniqueId(TokenProvider tokenProvider) {
```
yili-db: Taking OAuth as example, it has:
```
TYPE, OAUTH_URI, OAUTH_CLIENT_ID, OAUTH_CLIENT_SECRET
```
It does **not** have `catalogUri`.

This is mostly working except that there could be an edge case: Spark happens to connect to two different UC catalogs, with the same user identity. For example this Spark is dealing with data sync between two catalogs. This authUniqueId will be exactly the same across two different catalogs.

Then in some cases for each of the `CredId`:
- `DeltaStagingTableCredId`: this is fine as staging table id will be unique for sure, and it has `location` which will be unique as well.
- `DeltaTableCredId`: this is fine as it contains catalog name. And it has a unique `location`
- `PathCredId`: this may be OK as the `location` is likely unique
- `TableCredId`: this is questionable as it's unknown if two different catalog can have the same table id. And it has no `location`.

So it's very likely it will continue to work. But if you want to get rid of the guess, adding the `catalogUri` to the hashing would be the best way.
An alternative is to add `location` into `TableCredId` as well.
WDYT?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/MapIdGenerator.java
openinx: I inclined to simplify the deps here, and not using the the transitive dep from the hadoop, which may introduce the compatibility issues if we upgrade the hadoop and indirectly used the incompatibility issue.  

And currently, we have no plan to add the hadoop suffix for the `unitycatalog-hadoop` module.  and those `toHex` code actually only few lines of code, which it's not worth to introduce a separate dep.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
openinx: Thought this for a while, and also talked with @tdas offline. 

@tdas  suggested us to use the entire catalog props `map<string, string>` as the part of the `credId`.   I just took a look, and find that our abstracted `unitycatalog-hadoop` is a common module that accommodate to different engines, such as flink, trino, spark.  and there is no `catalog` or extra `map<string, string>` to represent the entire so-called `catalog` props.   

All the available members from the UCCredentialsHadoopConfs are here: https://github.com/unitycatalog/unitycatalog/blob/main/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/UCCredentialHadoopConfs.java#L78-L87

so, that's not feasible for us to use the catalog props. 

Then the available fields to choose as a key would be:
* catalogUri
* scheme
* TokenProvider.props()

with those 3 configs (rather than the `authConfig`),  actually we can fully identify the credentials  for different user cases.  so  i will update this PR with this context.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/CredId.java
```diff
@@ -70,36 +82,40 @@ static CredId create(Configuration conf, Supplier<CredId> defaultCredId) {
     boolean hasUcDeltaStagingTableId = stagingTableId != null && !stagingTableId.isEmpty();
 
     if (UC_CREDENTIALS_TYPE_TABLE_VALUE.equals(type) && !isDeltaApi && !hasUcDeltaStagingTableId) {
-      // Case 1: UC table (legacy API) — keyed by table ID + operation.
+      // Case 1: UC table (legacy API) — keyed by cred context + table ID + operation.
+      String credContextId = readCredContextId(conf);
```
yili-db, line 82: move this to before `if` block so that it doesn't have to repeate?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/DeltaStagingTableCredId.java
```diff
@@ -9,20 +10,27 @@
 import java.util.Objects;
 
 /**
- * {@link CredId} keyed by staging table ID and location; used for staging-table-level temporary
- * credentials via the UC Delta credentials API.
+ * {@link CredId} keyed by credential context, staging table ID, and location; used for
+ * staging-table-level temporary credentials via the UC Delta credentials API.
  */
 public class DeltaStagingTableCredId implements CredId {
+  private final String credContextId;
```
yili-db, line 17: How about moving this field into `CredId`? That makes it impossible to introduce a new cred id type without credContextId. Someone in the future might miss it when making a new type.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -501,6 +513,23 @@ private static Map<String, String> createCredProps(
...
+   * endpoint, storage scheme, and auth identity, so all three are folded into the hash alongside
+   * the auth configs from {@code tokenProvider}.
+   */
+  static String credContextId(String catalogUri, String scheme, TokenProvider tokenProvider) {
+    Objects.requireNonNull(catalogUri, "catalogUri is required");
+    Objects.requireNonNull(scheme, "scheme is required");
+    Objects.requireNonNull(tokenProvider, "tokenProvider is required");
+
+    Map<String, String> context = new HashMap<>(tokenProvider.configs());
+    context.put(CRED_CONTEXT_CATALOG_URI_KEY, catalogUri);
```
yili-db, line 528: as a precondition, the context map shouldn't already contain these two keys. Maybe do an assert here?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/CredId.java
```diff
@@ -115,4 +131,9 @@ static CredId create(Configuration conf, Supplier<CredId> defaultCredId) {
    *     carries no Unity Catalog credential-request properties (e.g. {@link DefaultCredId}).
    */
   Map<String, String> props();
+
+  static String readCredContextId(Configuration conf) {
+    String credContextId = conf.get(UC_CRED_CONTEXT_ID_KEY);
+    return credContextId != null ? credContextId : EMPTY_CRED_CONTEXT_ID;
```
yili-db: A fixed `EMPTY_CRED_CONTEXT_ID` isn't future proof. Just in case a future change produces empty context id, it will invalidate the isolation. Instead if a random key is used as a fallback, it will safely fallback to not share creds between sessions.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/CredId.java
openinx: that's not okay,  as the for fallbacking to the defaultCredId case ( case.5) ,  we are not even needing a extra `credContextId`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/DeltaStagingTableCredId.java
openinx: it works for the `TableCredId`, `PathCredId`, `DeltaTableCredId`, and `DeltaStagingTableCreId`, but it does not apply to the `DefaultCredId`, which is used for the case that disabled the `renewCredential.enabled`, essentially we won't set any kind of the hadoop props via the `CredPropsUtil` if the credential renewal is disabled. 

In more long term, we can fully removed the `renewalCredential.enabled=false` branch, which will involve how non-delta table should work with the credential.  but that's a totally separate question.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
openinx: Hi @yili-db , that's already impossible for now, since the `tokenprovider.configg` is always starting with the `auth.*` prefix,  while the catalogUri is using the `catalogUri`, which will never intersect with each other.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/CredId.java
openinx: okay, that make sense, it will be better to throw if we unexpectedly parse a null credContextId,  even though for now it's unlikely to happen as we always set the credContextId in the variants of CredId implementation.  but i'd say it's good to fast fail.

yili-db (approved):

PR #1653, opened 2026-07-03
Vend path credentials for bare cloud-path Spark SQL

### LINKED ISSUE #1075: Spark Parquet file path syntax is not accepted
**Describe the bug**

When trying to create a Delta table in Unity Catalog from a Parquet file using Spark SQL, the `parquet.` file path syntax is not accepted. According to the documentation and typical Spark behavior, using `parquet. `should be supported in CTAS (Create Table As Select) statements, but Unity Catalog currently rejects it.

**To Reproduce**

```
CREATE OR REPLACE TABLE `unity`.`demo_pool`.`import_table1`
USING DELTA
AS (SELECT *
    FROM parquet.`s3:/data-bucket/file.parquet`);
```
The query fails with the following error:
```
io.unitycatalog.client.ApiException: getTable call failed with: 400 - {"error_code":"INVALID_ARGUMENT","details":[{"reason":"INVALID_ARGUMENT","metadata":{},"@type":"google.rpc.ErrorInfo"}],
#"stack_trace":null,"message":"Invalid table name: unity.parquet.`s3://data-bucket/file.parquet`"}
```

**Expected behavior**

The query should create a Delta table in Unity Catalog using the data in the specified Parquet file. The `parquet. `file source syntax should be fully supported in CTAS statements, as it is in standard Spark.

- Spark version: 3.5.3
- Unity Catalog version: 0.2.1

s-maling-celonis:
## Summary

Vend Unity Catalog temporary path credentials for bare cloud-path Spark SQL (for example `` SELECT * FROM parquet.`s3://bucket/dir` `` and `INSERT OVERWRITE DIRECTORY`) via a new `ResolvePathCredentials` analyzer rule.

Also fixes catalog resolution for bare paths when a UC catalog is current: `UCProxy.getUCTableLike` now treats identifiers UC cannot express as `catalog.schema.table` (e.g. parquet dotted cloud paths) as absent, so Spark falls through to `ResolveSQLOnFile` instead of failing analysis with `Invalid table name` (HTTP 400). Fixes #1075.

- **ResolvePathCredentials** injects vended `fs.*` Hadoop options onto bare cloud paths during analysis, before `ResolveSQLOnFile` lists the path for schema inference. Eligible schemes come from `UCCredentialHadoopConfs.isSupportedScheme` / `CloudType` (same set as name-based table vending, including `s3a`).
- **UCSparkSessionExtensions** registers the rule via `injectHintResolutionRule`. The Hints batch is one of the analyzer's `earlyBatches`, so it runs strictly before the Resolution batch that holds `ResolveRelations` and `ResolveSQLOnFile`; `injectResolutionRule` would land in that same batch and could run after the path is already listed. The parser is untouched and stays side-effect free.
- **SQL only.** This rule matches `UnresolvedRelation` / `InsertIntoDir` from Spark SQL. `spark.read.parquet("s3://...")` and `df.write.parquet("s3://...")` build a file source without those nodes and do not vend path credentials here.
- **Current catalog.** The rule is a no-op unless the session's current catalog is a `UCSingleCatalog`. Register `spark.sql.extensions=io.unitycatalog.spark.UCSparkSessionExtensions` and `SET CATALOG` (or `spark.sql.defaultCatalog`) to that catalog.
- **Idempotence** — analyzer batches run to a fixed point, so the rule is re-applied until the plan stops changing. A node is skipped only after a lookup for this URI: a successful vend (`fs.unitycatalog.credentials.type=path` and `fs.unitycatalog.path`) or, on an allowed miss, Spark-only skip markers (`fs.unitycatalog.path.vending.attempted` / `...location`). Ambient `fs.s3a.*` keys do not skip vending. Miss markers are not a Hadoop `PathCredId` (no `credentials.type=path`), so credential-scoped filesystems can still use ambient credentials.
- **UCSingleCatalog.vendPathCredentialConf** vends credentials for a path; **vendPathCredentialConfWithFallback** tries `PATH_READ_WRITE` then falls back to `PATH_READ` for ambiguous bare `format.path` relations (reads and INSERT INTO targets).
- **UCSingleCatalog.isAddressableTableName** / **getUCTableLike guard** — skip REST lookup for dotted schema/table names Spark probes during path resolution; return absent so `CatalogV2Util.loadTable` can fall through.
- **vendPathCredentialConfOrEmpty** falls back to ambient storage credentials only for HTTP 400, 403, and 404 (unmanaged path / no permission; 400 covers OSS `FAILED_PRECONDITION`). Other `ApiException`s are logged and rethrown.
- **INSERT OVERWRITE DIRECTORY** always requests `PATH_READ_WRITE`.
- **INSERT INTO parquet bare paths**: the target `UnresolvedRelation` lives in `InsertIntoStatement.table`, which is not a tree child (only `query` is), so `plan.resolveOperators` never reached it. An explicit `InsertIntoStatement` case now injects credentials into that target. Note Spark 4.0–4.2 reject executing `INSERT INTO` on a bare path at analysis (`TABLE_OR_VIEW_NOT_FOUND`); the rule still injects credentials during analysis.
- **Delta exclusion (temporary):** Bare `delta` cloud path tables are excluded from UC path credential vending in this PR. Delta's analysis path does not propagate relation options into `DeltaLog`, so vended credentials would not reach execution and can interfere with ambient credentials (`spark.hadoop.fs.s3a.*`, instance profile). Delta bare-path support is tracked in #1754.

## Motivation

Path-based relations like parquet cloud paths are resolved by Spark's built-in `ResolveSQLOnFile`, which bypasses the Unity Catalog catalog entirely. No credentials are vended and cloud storage access fails, while the same read works once credentials flow through `UCSingleCatalog.loadTable`.

When a UC catalog is the session's current catalog, Spark calls `loadTable` on bare parquet path relations first. UC rejects the resulting dotted name with HTTP 400; only a not-found answer lets analysis continue to path-based resolution and credential injection.

## Configuration

Register the extension in sessions that query bare cloud paths:

```properties
spark.sql.extensions=io.unitycatalog.spark.UCSparkSessionExtensions
```

The session's current catalog must be a `UCSingleCatalog`.

Optional per-catalog toggle (default true):

```properties
spark.sql.catalog.<catalog>.vendPathCredentials.enabled=false
```

## Test plan

- [x] UCProxySuite — unit tests for unaddressable identifiers
- [x] UCSingleCatalogPathCredentialTest — ambient HTTP allowlist and miss skip markers
- [x] PathCredentialReadWriteTest
  - testWriteAndReadBarePath — parameterized happy path (s3 / s3a)
  - testWriteAndReadBareDeltaPath — local Delta e2e; cloud asserts UC credentials are not injected for delta cloud paths
  - testInsertIntoBarePathInjectsCredentials
  - testRepeatedApplicationVendsOnceAndLeavesPlanUnchanged — asserts a single vend and a fixed point with the credential cache disabled
  - testAllowedMissStampsIdentityAndSecondApplyLeavesPlanUnchanged — miss skip markers, not PathCredId keys
  - testFallsBackToAmbientWhenPathNotManagedByUc
  - testDisabledByFlag
  - testBarePathUsesCurrentCatalogAfterSetCatalog
  - testCountFromParquetSubquery — parameterized over CountSubqueryLayout (set-catalog, default-catalog-at-startup, Delta spark_catalog, runtime default catalog)
  - testCountFromParquetSubqueryFailsWhenCurrentCatalogIsDeltaCatalog
  - testMergeIntoUsingParquetSubquery
  - testMergeIntoParquetSubqueryWithDefaultCatalog
  - testPositionalSqlParameters / testNamedSqlParameters
  - testVendPathCredentialsForScheme / testVendPathCredentialsWithoutActiveSession
- [x] Compiles against Spark 4.0.0, 4.1.0 and 4.2.0
- [ ] Full spark/test (CI)


connectors/spark/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/UCSparkSqlExtensionsParser.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  override def parseFunctionIdentifier(sqlText: String): FunctionIdentifier =
+    delegate.parseFunctionIdentifier(sqlText)
+
+  override def parseMultipartIdentifier(sqlText: String): Seq[String] =
+    delegate.parseMultipartIdentifier(sqlText)
+
+  override def parseTableSchema(sqlText: String): StructType = delegate.parseTableSchema(sqlText)
+
+  override def parseRoutineParam(sqlText: String): StructType = delegate.parseTableSchema(sqlText)
```
yili-db: `parseRoutineParam` delegates to `parseTableSchema`

connectors/spark/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/UCSparkSqlExtensionsParser.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  override def parsePlan(sqlText: String): LogicalPlan = {
+    val spark = SparkSession.active
```
yili-db: This is fragile. It could break in multi-thread environment.
Instead `extensions.injectParser` provides it and this class can save it as a member var.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,120 @@
...
+ * home of the parser hook that invokes the rule), and use the {@link S3CredentialTestFileSystem}
+ * fake filesystem to assert the vended credentials reach S3A. The test principal is the metastore
+ * owner, so path authorization passes and credentials fall back to the per-bucket server config
+ * ({@code accessKey0}/... for {@code s3://test-bucket0}) configured in {@link
+ * BaseSparkIntegrationTest#setUpProperties()}.
+ */
+public class PathCredentialReadWriteTest extends BaseSparkIntegrationTest {
+
+  private static final String VEND_ENABLED_CONF =
+      "spark.sql.unitycatalog.vendPathCredentials.enabled";
```
yili-db: All other UC configs are using `spark.sql.catalog.<name>.*` and they are listed in `OptionsUtil`

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,120 @@
...
+    sql("INSERT OVERWRITE DIRECTORY '%s' USING parquet SELECT 1 AS i, 'a' AS s", location);
+    assertSingleRow(sql("SELECT * FROM parquet.`%s`", location));
+  }
+
+  /**
+   * Same as above but with credential renewal + credential-scoped filesystem enabled, so the vended
+   * credentials flow through {@code AwsVendedTokenProvider} + {@code CredScopedFileSystem} rather
+   * than static keys.
+   */
+  @Test
```
yili-db, line 407: This and `testWriteAndReadBareS3PathStaticCreds` should be a single `ParameterizedTest`

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,120 @@
...
+  public void testDisabledByFlag() throws IOException {
+    session = createUcSparkSession(false, false, SPARK_CATALOG);
+    String location = bucketPath("import_disabled");
+    // Write with the feature on so the data exists.
+    sql("INSERT OVERWRITE DIRECTORY '%s' USING parquet SELECT 1 AS i, 'a' AS s", location);
+
+    session.conf().set(VEND_ENABLED_CONF, "false");
+    assertThatThrownBy(() -> sql("SELECT * FROM parquet.`%s`", location));
```
yili-db: The failure test should assert on an specific error that is lack of credential, rather than any other random error

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,120 @@
...
+   * credentials flow through {@code AwsVendedTokenProvider} + {@code CredScopedFileSystem} rather
+   * than static keys.
+   */
+  @Test
+  public void testWriteAndReadBareS3PathVendedProvider() throws IOException {
+    session = createUcSparkSession(true, true, SPARK_CATALOG);
+    String location = bucketPath("import_vended");
+
+    sql("INSERT OVERWRITE DIRECTORY '%s' USING parquet SELECT 1 AS i, 'a' AS s", location);
+    assertSingleRow(sql("SELECT * FROM parquet.`%s`", location));
```
yili-db, line 447: Should also try at least one other operations `INSERT INTO`

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
```diff
@@ -0,0 +1,131 @@
...
+    if (!isEnabled) return plan
+    currentUcCatalog match {
+      case None => plan
+      case Some(uc) =>
+        plan.resolveOperators {
+          // Read: a two-part `format`.`<cloud path>` identifier, still unresolved.
+          case u: UnresolvedRelation
+              if u.multipartIdentifier.length == 2 && isCloudPath(u.multipartIdentifier.last) =>
+            val path = u.multipartIdentifier.last
+            val conf = uc.vendPathCredentialConf(path, PathOperation.PATH_READ)
```
yili-db: **UnresolvedRelation** doesn't always mean read. `INSERT INTO parquet.path` hits here as well.
So something like this:
```
    val conf = try {
      uc.vendPathCredentialConf(path, PathOperation.PATH_READ_WRITE)
    } catch {
      case e: ApiException =>
        logWarning(s"PATH_READ_WRITE credential generation failed...")
        uc.vendPathCredentialConf(path, PathOperation.PATH_READ)
```

s-maling-celonis: Thanks for catching this. Two follow-ups after digging in:

1. Bare `format.`path`` relations now use `vendPathCredentialConfWithFallback` (try `PATH_READ_WRITE`, fall back to `PATH_READ` on `ApiException`, mirroring `loadTable`) — `e0f09140`.

2. The `INSERT INTO parquet.`s3://...`` target does **not** actually reach the `UnresolvedRelation` case above: its relation lives in `InsertIntoStatement.table`, which is not a tree child (only `query` is), so `plan.resolveOperators` skipped it and no credentials were vended. Added an explicit `InsertIntoStatement` case that injects credentials into that target — `8e409371`. Note Spark 4.1 rejects *executing* `INSERT INTO` on a bare path at analysis (`TABLE_OR_VIEW_NOT_FOUND`, since INSERT DML requires a registered table identifier), so this is pinned by a parse-time test rather than an executed integration test.

connectors/spark/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/UCSparkSqlExtensionsParser.scala
s-maling-celonis: Thanks for the catch — fixed in 37abe6db: `parseRoutineParam` now delegates to `delegate.parseRoutineParam(sqlText)`.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: Done in 0207c900. The flag is now `spark.sql.catalog.<name>.vendPathCredentials.enabled` (constant in `OptionsUtil`), read at catalog init in `UCSingleCatalog` and checked by `ResolvePathCredentials`. The disabled test recreates the session with the catalog option set to `false` since catalog options are not mutable at runtime.

connectors/spark/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/UCSparkSqlExtensionsParser.scala
s-maling-celonis: Done in 9a8fbf2e. `UCSparkSessionExtensions` now passes the `SparkSession` from `injectParser` into `UCSparkSqlExtensionsParser`, which stores it as a member and uses it in `parsePlan` instead of `SparkSession.active`.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: Fixed in e0f09140.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: Fixed in e0f09140.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: `INSERT INTO parquet.`s3://...`` is now covered by `testInsertIntoBarePathInjectsCredentialsAtParseTime` (`8e409371`). It can't run as an executed integration test on Spark 4.1 — INSERT DML on a bare path is rejected at analysis with `TABLE_OR_VIEW_NOT_FOUND` — so the test parses the statement, asserts the credential is injected into the `UnresolvedRelation` target, then asserts execution still throws. This also uncovered that the target lives in `InsertIntoStatement.table` (not reached by the existing `UnresolvedRelation` rule), fixed in the same commit.

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
```diff
@@ -0,0 +1,144 @@
...
+                  if u.multipartIdentifier.length == 2 &&
+                    isCloudPath(u.multipartIdentifier.last) =>
+                i.copy(table = injectUnresolvedRelationCredentials(u, uc))
+              case _ => i
+            }
+
+          // Write: INSERT OVERWRITE DIRECTORY '<cloud path>' USING <format> ...
+          case i: InsertIntoDir if i.storage.locationUri.exists(u => isCloudPath(u.toString)) =>
+            val location = i.storage.locationUri.get.toString
+            val conf = uc.vendPathCredentialConf(location, PathOperation.PATH_READ_WRITE)
```
yili-db: There are two possible ways to access path based tables:
1. Using UC credential like this PR
2. Before this PR, it can be done with **ambient credentials** already (instance profile, spark.hadoop.fs.s3a.*, etc.)

`uc.vendPathCredentialConf` can fail with an exception when:
1. path isn't managed by UC
2. path is managed by UC but user lacks of permission etc

When a failure happens in this line, it should return back the original `plan` and fallback to **ambient credentials**.
The same applies to `injectUnresolvedRelationCredentials`

This fallback can be tested by inserting into a path that is different than the designated testing S3 path and observe a corresponding S3 access failure due to lack of cred.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,162 @@
...
+  @ParameterizedTest
+  @CsvSource({"false, false", "true, true"})
+  public void testWriteAndReadBareS3Path(boolean renewCred, boolean credScopedFsEnabled)
+      throws IOException {
+    stopSession();
+    session = createUcSparkSession(renewCred, credScopedFsEnabled, true, SPARK_CATALOG);
+    String location = bucketPath("write_directory_" + renewCred + "_" + credScopedFsEnabled);
+
+    sql("INSERT OVERWRITE DIRECTORY '%s' USING parquet SELECT 1 AS i, 'a' AS s", location);
+    assertSingleRow(sql("SELECT * FROM parquet.`%s`", location));
```
yili-db, line 341: try `INSERT INTO` as well?

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,162 @@
...
+    String insertSql = String.format("INSERT INTO parquet.`%s` SELECT 2 AS i, 'b' AS s", location);
+
+    LogicalPlan plan = session.sessionState().sqlParser().parsePlan(insertSql);
+    assertThat(plan).isInstanceOf(InsertIntoStatement.class);
+    LogicalPlan table = ((InsertIntoStatement) plan).table();
+    assertThat(table).isInstanceOf(UnresolvedRelation.class);
+    UnresolvedRelation target = (UnresolvedRelation) table;
+    assertThat(target.options().get("fs.s3a.access.key")).isEqualTo("accessKey0");
+
+    assertThrows(Exception.class, () -> session.sql(insertSql).collectAsList());
```
yili-db: I see that you have the matching of `InsertIntoStatement` already. Is this test outdated?

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  /**
+   * When the feature is disabled, no credentials are injected for the bare path, so the read fails
+   * (this is the pre-fix behavior). Confirms the rule is what enables direct path access.
+   */
+  @Test
+  public void testDisabledByFlag() throws IOException {
```
yili-db, line 593: This is the same kind of test I would expect for "path based table access but path isn't managed by UC". It makes sure that in that case the code execution continues instead of error out by UC.

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
s-maling-celonis: Done in 8f5583ca. When UC path credential vending fails, `vendPathCredentialConfOrEmpty` / `vendPathCredentialConfWithFallback` return an empty map and the plan node is left unchanged so Spark can use ambient storage credentials. Logging is at **debug** only (missing external location is expected, not a warning). Covered by `testFallsBackToAmbientWhenPathNotManagedByUc`.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: `INSERT INTO` is covered at parse time in `testInsertIntoBarePathInjectsCredentialsAtParseTime` (the `InsertIntoStatement.table` branch). Executed `INSERT INTO` on a bare path is not testable on Spark 4.1 — analysis rejects it with `TABLE_OR_VIEW_NOT_FOUND`. Executed write coverage is via `INSERT OVERWRITE DIRECTORY` in `testWriteAndReadBareS3Path`.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: Not outdated — it is still the only test that hits the `InsertIntoStatement.table` branch (not a tree child, so the generic `UnresolvedRelation` case never sees it). Removed the execution `assertThrows` on line 135; it only documented Spark 4.1 analysis failure, not credential behavior.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: Added `testFallsBackToAmbientWhenPathNotManagedByUc` in 8f5583ca — same shape as `testDisabledByFlag`: feature enabled, path UC cannot vend (`NO_CREDS_BUCKET`), assert no `ApiException` in the cause chain and a filesystem-level failure instead.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -375,6 +378,66 @@ class UCSingleCatalog
...
+    val scheme = new Path(location).toUri.getScheme
+    if (scheme == null) return util.Collections.emptyMap[String, String]()
+    UCCredentialHadoopConfs
+      .builder(uri.toString, scheme)
+      .addAppVersions(ApiClientFactory.appEngineVersions())
+      .tokenProvider(tokenProvider)
+      .apiClient(apiClient)
+      .enableCredentialRenewal(renewCredEnabled)
+      .enableCredentialScopedFs(credScopedFsEnabled)
+      .hadoopConf(UCSingleCatalog.sessionHadoopConf())
```
yili-db: This function `sessionHadoopConf` also calls `SparkSession.getActiveSession.sessionState.newHadoopConf` and has the same problem as https://github.com/unitycatalog/unitycatalog/pull/1653#discussion_r3606627938

This is an existing function that has been called by other `UCSingleCatalog` functions. But here is the difference on why it is OK for those:
1. It is called indirectly by functions like `UCSingleCatalog.createTable` which is guaranteed to happen inside `SparkSession.withActive`
2. Here it may not when called like `session.sessionState().sqlParser().parsePlan(insertSql)`. `SparkSession.getActiveSession` may or may not work in this case. The test just happened to work.

So a better way is to use the captured spark session by parser, and add it to the function signature of vendPathCredentialConf etc. Then a separate function sessionHadoopConf that does take a spark session as input (while keeping the other one as overloaded without input session parameter).

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,192 @@
...
+  @ParameterizedTest
+  @CsvSource({"false, false", "true, true"})
+  public void testWriteAndReadBareS3Path(boolean renewCred, boolean credScopedFsEnabled)
+      throws IOException {
+    stopSession();
+    session = createUcSparkSession(renewCred, credScopedFsEnabled, true, SPARK_CATALOG);
+    String location = bucketPath("write_directory_" + renewCred + "_" + credScopedFsEnabled);
+
+    sql("INSERT OVERWRITE DIRECTORY '%s' USING parquet SELECT 1 AS i, 'a' AS s", location);
+    assertSingleRow(sql("SELECT * FROM parquet.`%s`", location));
```
yili-db, line 341: Can you also try these two queries?
```
  // Positional
  session.sql("SELECT ? AS c", new Object[]{42})

  // Named
  session.sql("SELECT :x AS c", java.util.Map.of("x", 42))
```
We never tested queries like this because up until this PR we didn't change any parser.

But now AI told me:
`UCSparkSqlExtensionsParser` overrides `parsePlan`/`parseQuery` but not `parsePlanWithParameters`. `SparkSession.sql(text, args)` routes through `parsePlanWithParameters`; the inherited default drops the `parameterContext`, so `?`/`:name` markers never bind in any session that registers the extension. Existing tests miss it because the test `sql()` helper uses `String.format`, not real Spark params.

**Fails only with the extension:** `session.sql("SELECT ? AS c", new Object[]{42})`

**Fix:** override the missing method (and delegate `parseQuery` to the real one):

```scala
override def parsePlanWithParameters(sqlText: String, ctx: ParameterContext): LogicalPlan =
  ResolvePathCredentials(spark).apply(delegate.parsePlanWithParameters(sqlText, ctx))

override def parseQuery(sqlText: String): LogicalPlan =
  ResolvePathCredentials(spark).apply(delegate.parseQuery(sqlText))
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: Done in f2d83dec.

- Added `UCSingleCatalog.sessionHadoopConf(spark)` and kept the no-arg overload for existing catalog call sites (`loadTable`, `createTable`, …).
- Threaded `spark` through `vendPathCredentialConf`, `vendPathCredentialConfOrEmpty`, and `vendPathCredentialConfWithFallback`; `ResolvePathCredentials` passes its parser session into all vend calls.
- Also resolve the UC catalog via `spark.sessionState.conf` + `catalogManager.catalog(name)` instead of `currentCatalog`, which goes through `SQLConf.get` and can break when there is no active session on the thread.

Added `testVendPathCredentialsWithoutActiveSession`: clears the active session, calls `vendPathCredentialConfWithFallback(session, path)` directly, and asserts `fs.s3a.access.key` is vended.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: Done in 76c040a2.

- Override `parsePlanWithParameters` to delegate with `ParameterContext` intact, then apply `ResolvePathCredentials`.
- Fix `parseQuery` to call `delegate.parseQuery` (was incorrectly routing through `parsePlan`).
- Added `testPositionalSqlParameters` and `testNamedSqlParameters` using a UC-only session (`SELECT ?` / `SELECT :x`).

Note: with **Delta + UC** extensions registered together, parameterized SQL still depends on the Delta parser wrapper also forwarding `parsePlanWithParameters` (same class of bug Iceberg fixed in apache/iceberg#16626). UC-only and UC-as-outermost-over-SparkSqlParser paths are covered by the tests.

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
```diff
@@ -0,0 +1,159 @@
...
+      relation: UnresolvedRelation,
+      uc: UCSingleCatalog): UnresolvedRelation = {
+    val path = relation.multipartIdentifier.last
+    val conf = uc.vendPathCredentialConfWithFallback(spark, path)
+    if (conf.isEmpty) relation else relation.copy(options = mergeOptions(relation.options, conf))
+  }
+
+  private def currentUcCatalog: Option[UCSingleCatalog] = {
+    val manager = spark.sessionState.catalogManager
+    val catalogName = spark.sessionState.conf.getConf(SQLConf.DEFAULT_CATALOG)
```
yili-db: Reading `DEFAULT_CATALOG` directly ignores `USE CATALOG`. Spark's `currentCatalog` is `_currentCatalogName.getOrElse(conf.getConf(DEFAULT_CATALOG))`, so this is just the fallback branch — after `USE CATALOG my_uc` (not the default), bare-path reads silently won't get UC credentials.

Use the session's own `CatalogManager`, pinning the conf so the fallback is thread-safe at parse time:

    private def currentUcCatalog: Option[UCSingleCatalog] = {
      val manager = spark.sessionState.catalogManager
      val catalog =
        try SQLConf.withExistingConf(spark.sessionState.conf) {
          manager.currentCatalog
        } catch { case _: CatalogNotFoundException => null }
      catalog match {
        case uc: UCSingleCatalog => Some(uc)
        case _ => None
      }
    }

`withExistingConf` is present on `SQLConf` in all cross-built versions (4.0.0/4.1.0/4.2.0), verified via javap.

connectors/spark/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/UCSparkSqlExtensionsParser.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  override def parseTableSchema(sqlText: String): StructType = delegate.parseTableSchema(sqlText)
+
+  override def parseRoutineParam(sqlText: String): StructType =
+    delegate.parseRoutineParam(sqlText)
+
+  override def parsePlan(sqlText: String): LogicalPlan =
+    applyPathCredentials(delegate.parsePlan(sqlText))
+
+  override def parsePlanWithParameters(
```
yili-db: # Fix Spark 4.0.0 compile break in `UCSparkSqlExtensionsParser`

## Problem

The parser is in the common `src/main/scala/` tree, so it compiles for all cross-built versions (4.0.0, 4.1.0, 4.2.0-SNAPSHOT). It imports `ParameterContext` and overrides `parsePlanWithParameters` — both are 4.1+ only, **absent in 4.0.0** (verified via `javap`). So 4.0.0 fails to build.

The override is needed on 4.1/4.2 (`sql(text, args)` routes through `parsePlanWithParameters` there). On 4.0 it's not — `sql(text, args)` goes through plain `parsePlan`, and the type doesn't exist to name.

## Proposal

Isolate just that one method in a small `ParserInterfaceShim` trait — empty on 4.0, carrying the override on 4.1/4.2 — and keep the whole parser class in the common tree. Trait needs hooks (`delegateParser`, `applyPathCredentials`) since Scala 2 traits take no constructor params.

**`scala-shims/spark-4.0/`** — empty, no `ParameterContext` reference:
```scala
trait ParserInterfaceShim extends ParserInterface {
  protected def delegateParser: ParserInterface
  protected def applyPathCredentials(plan: LogicalPlan): LogicalPlan
}
```

**`scala-shims/spark-4.1-4.2/`** — same trait plus:
```scala
  override def parsePlanWithParameters(sqlText: String, parameterContext: ParameterContext): LogicalPlan =
    applyPathCredentials(delegateParser.parsePlanWithParameters(sqlText, parameterContext))
```

**Common `UCSparkSqlExtensionsParser`** — `extends ParserInterfaceShim`, implements the two hooks, keeps all the plain delegates + `parsePlan`/`parseQuery`. Never names `parsePlanWithParameters`.

**`project/spark-versions.json`** — add the second shim dir per version (`sourceDirs` is already an additive array, no build.sbt change):
```json
{"version": "4.0.0", "sourceDirs": ["scala-shims/spark-4.0-4.1", "scala-shims/spark-4.0"]},
{"version": "4.1.0", "sourceDirs": ["scala-shims/spark-4.0-4.1", "scala-shims/spark-4.1-4.2"]},
{"version": "4.2.0-SNAPSHOT", "sourceDirs": ["scala-shims/spark-4.2", "scala-shims/spark-4.1-4.2"], ...}
```

## Notes

- 4.0 loses nothing: `sql(text, args)` calls `parsePlan` then wraps the result, so overriding `parsePlan` (common class) already covers parameterized queries.

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
s-maling-celonis: Done in 85177a01.

`currentUcCatalog` now resolves `manager.currentCatalog` inside `SQLConf.withExistingConf(spark.sessionState.conf)` so `USE CATALOG` is honored and the conf fallback is thread-safe at parse time.

Added `testBarePathUsesCurrentCatalogAfterUseCatalog`: default stays the built-in `spark_catalog`, session selects the UC catalog via `USE CATALOG`, then asserts credentials are injected at parse time and on executed read.

connectors/spark/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/UCSparkSqlExtensionsParser.scala
s-maling-celonis: Done in 21f3cc4d, following your proposal.

- Added `ParserInterfaceShim` in `scala-shims/spark-4.0/` (empty; no `ParameterContext`) and `scala-shims/spark-4.1-4.2/` (provides `parsePlanWithParameters`).
- `UCSparkSqlExtensionsParser` now extends `ParserInterfaceShim`, implements `delegateParser` / `applyPathCredentials` hooks, and no longer names `ParameterContext`.
- Updated `project/spark-versions.json` with the second shim dir per Spark version.

4.0 keeps parameterized SQL via plain `parsePlan`; 4.1+ keep `parsePlanWithParameters` via the shim.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,284 @@
...
+   * After {@code USE CATALOG}, bare-path credential injection must follow the session's current
+   * catalog, not {@code SQLConf.DEFAULT_CATALOG}. Default remains the built-in {@code spark_catalog}
+   * while the UC catalog is selected explicitly.
+   */
+  @Test
+  public void testBarePathUsesCurrentCatalogAfterUseCatalog() throws IOException, ParseException {
+    stopSession();
+    session = createUcSparkSession(false, false, true, CATALOG_NAME);
+    String location = bucketPath("use_catalog");
+    sql("USE CATALOG %s", CATALOG_NAME);
```
yili-db: Should be `SET CATALOG %s`

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
yili-db: Thanks for explaining. It seems that this is true for 4.0, 4.1 and 4.2.
Just for future proof, let's still keep the `InsertIntoStatement` branch but add a comment to explain that it doesn't reach there at all.
Also please update the comment where it says "blocked in 4.1" to remove the mentioning of "4.1" since it's not limited to 4.1.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: Done in 5f88fc7a — switched to `SET CATALOG %s` and renamed the test to `testBarePathUsesCurrentCatalogAfterSetCatalog`.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: Done in 5f88fc7a — kept the `InsertIntoStatement` branch in `ResolvePathCredentials` with an expanded comment (generic `UnresolvedRelation` never visits `InsertIntoStatement.table`; executed INSERT INTO on bare paths fails at analysis on Spark 4.0–4.2). Updated test javadocs to say Spark 4.x / 4.0–4.2 instead of pinning to 4.1.

connectors/spark/src/main/scala-shims/spark-4.0/org/apache/spark/sql/catalyst/parser/extensions/ParserInterfaceShim.scala
```diff
@@ -0,0 +1,33 @@
+/*
```
yili-db: Please remove the license header from all these files. The convention of this repo is no license header.

s-maling-celonis: Done in b0e3ced2 — removed the ASF license header from all five path-credential Spark sources (both ParserInterfaceShim shims, ResolvePathCredentials, UCSparkSessionExtensions, and UCSparkSqlExtensionsParser).

yili-db (approved):

s-maling-celonis:
@yili-db 1 extra fix was required here https://github.com/unitycatalog/unitycatalog/pull/1653/changes/cc9cc17947d0ad32cbc049214236f48b25e8c616. Let me know if you would prefer this fix as a separate PR.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -736,6 +736,22 @@ object UCSingleCatalog {
...
+   *
+   * Spark asks the current catalog to load every relation before falling back to path-based
+   * resolution, so `SELECT * FROM parquet.`s3://bucket/dir/part.parquet`` arrives here as schema
+   * `parquet` with the path as the table name. Callers treat these as absent, which lets Spark
+   * carry on to `ResolveSQLOnFile` -- see [[UCProxy.getUCTableLike]].
+   */
+  def isAddressableTableName(ident: Identifier): Boolean = {
+    ident.namespace().length == 1 &&
+      !ident.namespace()(0).contains(".") &&
+      !ident.name().contains(".")
```
yili-db: This is a good change. I think it should guard against `/` instead so that paths like s3://bucket/dir/part without extension would also be recognized.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,549 @@
...
+
+  @TempDir protected File dataDir;
+
+  /**
+   * Builds a Spark session that registers {@code UCSparkSessionExtensions} (so the parser-level
+   * {@link ResolvePathCredentials} hook is active) and points the given catalogs at the test UC
+   * server. Mirrors {@link BaseSparkIntegrationTest#createSparkSessionWithCatalogs} but adds the UC
+   * extension. The catalogs are expected to already exist (created in {@code setUp}).
+   */
+  private SparkSession createUcSparkSession(
```
yili-db: Now in this test there are four `create*Session` functions. They are doing similar task with a little bit customization. That's not only duplicate code but also it's hard to understand how each one of them differs with the rest.
I suggest consolidate all four of them into one function with customization using parameters.

Then I believe the tests calling these functions are also exercising identical test coverage, but with different types of session. The test coverage can be consolidated into a function to accept the same kind of session customization.

For example:
```
@Test
void someTest() {
List<SessionCustomizationParameters> parameters = List.of(...);
  for (p : parameters) {
    testPathBasedAccess(p, ...);
  }
}

void testPathBasedAccess(p, ...) {
  session = createSparkSession(p);
  // Run all the queries
  stopSession();
}

Comment (yili-db) [reply connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java]: This could be a follow up item

File: connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java [hunk 20]
```diff
@@ -0,0 +1,549 @@
...
+    String location = bucketPath("storium_runtime_default");
+    sql("INSERT OVERWRITE DIRECTORY '%s' USING parquet SELECT 1 AS i, 'a' AS s", location);
+
+    assertThat(sql("SELECT current_catalog()").get(0).getString(0)).isEqualTo(CATALOG_NAME);
+    List<Row> rows = sql("SELECT COUNT(*) AS c FROM (SELECT * FROM parquet.`%s`)", location);
+    assertThat(rows.get(0).getLong(0)).isEqualTo(1);
+  }
+
+  /** Mirrors celospark spark-uc.properties + Storium per-session hiveconf catalog layout. */
+  private SparkSession createStoriumLikeThriftSession(String tenantCatalog) {
```
yili-db: this is never called.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,549 @@
...
+    String location = bucketPath("storium_like_thrift");
+    sql("INSERT OVERWRITE DIRECTORY '%s' USING parquet SELECT 1 AS i, 'a' AS s", location);
+    sql("SET CATALOG %s", SPARK_CATALOG);
+
+    assertThat(sql("SELECT current_catalog()").get(0).getString(0)).isEqualTo(SPARK_CATALOG);
+    assertThat(session.conf().get("spark.sql.defaultCatalog")).isEqualTo(CATALOG_NAME);
+
+    assertThatThrownBy(
+            () -> sql("SELECT COUNT(*) AS c FROM (SELECT * FROM parquet.`%s`)", location))
+        .isNotNull();
```
yili-db: Can you please do a more specific error assertion?

yili-db (commented):
@s-maling-celonis Please take a look

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: Addressed in 0e7d780e: `isAddressableTableName` now checks for `/` instead of `.`, covering extensionless paths like `s3://bucket/dir/part`. Added `testLoadTableSkipsRpcForExtensionlessBarePath` in `UCProxySuite`. Also rebased onto current `main`.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: Updated in c863d4c5: `testCountFromParquetSubqueryFailsWhenCurrentCatalogIsDeltaCatalog` now asserts the cause chain contains `"accessKey0"` (missing vended credentials), matching `testDisabledByFlag`.

s-maling-celonis:
Aug 1 feedback addressed in `0e7d780e` and `c863d4c5`. Also rebased onto current `main` (PR is now 12 files, path-credential scope only) and cleaned up fork-only commits so DCO passes.

Thanks for the comments, ready for another look @yili-db

s-maling-celonis:
Re: [Aug 1 review](https://github.com/unitycatalog/unitycatalog/pull/1653#pullrequestreview-4834354745) — all items addressed in `0e7d780e` and `c863d4c5` (rebased onto current `main`; PR is 12 path-credential files; DCO green).

**1. `isAddressableTableName` — guard on `/` (`UCSingleCatalog.scala`)**
Extensionless bare paths like `s3://bucket/dir/part` have no `.` but carry `/` in the table name. The guard now checks `/` instead of `.`, and `UCProxySuite` adds `testLoadTableSkipsRpcForExtensionlessBarePath` to assert the UC RPC is skipped.

**2. Consolidate four `create*Session` helpers (`PathCredentialReadWriteTest.java`)**
Session setup is consolidated into layered factories: shared `configureUcCatalog*` / `configureUcCatalogWithPathCredOptions`, plus `createPathCredSession*` (with named variants for default-catalog-at-startup and Delta `spark_catalog` layout) and `createUcExtensionOnlySession` for parser-delegate tests.

**3. Session-consolidation follow-up**
Done in the same refactor rather than deferred — count-subquery catalog-selection cases are parameterized via a `CountSubqueryLayout` enum (`@EnumSource`), each owning its session open + post-setup steps.

**4. Unused session helper**
Removed the dead overload.

**5. Specific failure assertion (`testCountFromParquetSubqueryFailsWhenCurrentCatalogIsDeltaCatalog`)**
Now asserts the cause chain contains `"accessKey0"` (missing vended credentials), consistent with `testDisabledByFlag`.

Also rebased onto upstream `main` to drop unrelated fork/workflow and duplicate test diffs already landed via #1723.

Ready for another look @yili-db

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+}
+
+object ResolvePathCredentials {
+
+  /**
+   * URI schemes for which UC can vend credentials. These match the schemes UC external locations
+   * are registered with and the ones [[UCCredentialHadoopConfs]] understands; other schemes (local
+   * paths, `s3a`, HDFS, ...) are left untouched.
```
yili-db: Why is `s3a` ignored? I see other comments mentioning `s3a`.

s-maling-celonis:
Re: https://github.com/unitycatalog/unitycatalog/pull/1653#discussion_r3702944784 — @yili-db

Good catch — the other `s3a` mentions refer to **Hadoop config keys** (`fs.s3a.access.key`, etc.), i.e. the S3A filesystem implementation UC vends into, not the **`s3a://` URI scheme** in SQL.

Previously `s3a://` was excluded from `CLOUD_SCHEMES` because UC external locations and the path-credentials API only accept canonical `s3://` URLs (`UriScheme.S3` / `NormalizedURL` on the server). That was misleading in the comment.

Fixed in 670dc32e:

1. **`ResolvePathCredentials`** — add `s3a` to `CLOUD_SCHEMES` so bare `` parquet.`s3a://…` `` paths get credential injection at parse time.
2. **`UCSingleCatalog.vendPathCredentialConf`** — rewrite `s3a://` → `s3://` for UC API lookup only; Spark still executes against the original `s3a://` URI.
3. **Injected options** — same as `s3://`: `fs.s3a.*` Hadoop keys (not `fs.s3.*` credential keys).
4. **Tests** — parameterized `testWriteAndReadBarePath` / `testVendPathCredentialsForScheme` cover both schemes.

So to Spark: path stays `s3a://`, credentials arrive as `fs.s3a.*`; only the UC server lookup uses `s3://`.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -386,6 +389,87 @@ class UCSingleCatalog
...
+  /**
+   * UC external locations and the temporary path-credentials API use canonical `s3://` URLs on the
+   * server. Hadoop clients may still reference the same storage as `s3a://`; rewrite to `s3://`
+   * for credential lookup only.
+   */
+  private def pathForCredentialApi(location: String): (String, String) = {
+    val scheme = new Path(location).toUri.getScheme
+    if (scheme == null) {
+      (location, null)
+    } else if (scheme.equalsIgnoreCase("s3a")) {
```
yili-db: `equalsIgnoreCase` is a good catch. The same as `isCloudPath` where it has `toLowerCase`.

However the other upper case schemas like "S3://" and "ABFS://" won't be lower cased. They will not be correctly handled.

s-maling-celonis: Agreed — `isCloudPath` already lowercases for matching, but `pathForCredentialApi` was passing the raw scheme through to the UC API (`CloudType` / `UriScheme` / `NormalizedURL` are all case-sensitive).

Fixed in `da5b1891`: `pathForCredentialApi` now lowercases every scheme for lookup and rewrites the URI prefix to the canonical form (`S3://` → `s3://`, `S3A://` → `s3://`, `ABFS://` → `abfs://`, etc.).

`testVendPathCredentialsForScheme` extended with `S3` and `S3A` rows. End-to-end write/read stays lowercase-only because the fake S3 filesystem matches schemes case-sensitively — unrelated to the UC lookup path.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -0,0 +1,561 @@
...
+ * external table.
+ *
+ * <p>Unlike the other Spark integration tests, these register {@code UCSparkSessionExtensions} (the
+ * home of the parser hook that invokes the rule), and use the {@link S3CredentialTestFileSystem}
+ * fake filesystem to assert the vended credentials reach S3A. The test principal is the metastore
+ * owner, so path authorization passes and credentials fall back to the per-bucket server config
+ * ({@code accessKey0}/... for {@code s3://test-bucket0}) configured in {@link
+ * BaseSparkIntegrationTest#setUpProperties()}.
+ */
+public class PathCredentialReadWriteTest extends BaseSparkIntegrationTest {
```
yili-db, line 45: Can this test exercise `Delta` path tables instead? I see that it loads `DeltaCatalog` and extension, but exercising parquet tables. Exercising `Delta` path tables makes sure this change work nicely with Delta catalog and extension.

yili-db (commented):
@s-maling-celonis One more comment regarding testing Delta tables. Thank you!

s-maling-celonis:
> @s-maling-celonis One more comment regarding testing Delta tables. Thank you!

@yili-db thanks for suggestion.
Re: Delta path tables — added testWriteAndReadBareDeltaPath (local Delta path e2e + cloud parse-time cred injection). Delta cloud-path reads are not working end-to-end yet: credentials are injected at parse time on `delta.\s3://...`, but Delta’s LogStore opens S3 without those `fs.s3a.*` keys, so execution fails while parquet bare paths work in the same session. Root cause is Delta’s analysis path not propagating relation options into DeltaLog’s Hadoop conf (unlike ResolveSQLOnFile` for parquet). Suggest addressing Delta execution in a follow-up PR rather than expanding this one.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
s-maling-celonis: Thanks for the suggestion, Parquet is still important, but we can still exercise Delta path tables — I added testWriteAndReadBareDeltaPath with two parts:

**Local delta.\file://...** — CREATE + read under DeltaCatalog+SET CATALOG to the UC catalog. This passes end-to-end and confirms the Delta extension/catalog layout works with path-credential sessions.
**Cloud delta.\s3://...** — parse-time credential injection is verified (the parsed UnresolvedRelationhas the expectedfs.s3a.* options). Execution-time read does not work yet.
What works: ResolvePathCredentials correctly vends credentials and attaches them to the parsed plan for delta.\s3://...the same way it does forparquet.s3://...``.

What does not work: actually running SELECT * FROM delta.\s3://...fails when Delta opens S3 to read_delta_log:

java.io.IOException: missing vended credential fs.s3a.access.key: expected accessKey0 but was null
at S3SingleDriverLogStore.listFrom(...)
at DeltaLog / snapshot loading ...
Why: Parquet bare paths go through Spark’s ResolveSQLOnFile, which passes UnresolvedRelation.options (including fs.s3a.) into the DataSource Hadoop config — so execution works. Delta path reads take a different analysis path (DeltaAnalysis → DeltaTableV2 → DeltaLog / LogStore) and the vended fs. options from the parsed relation are not propagated into the Hadoop Configuration used at execution time. Catalog-backed Delta tables are unaffected (UCSingleCatalog.loadTable vends creds separately).

I’d suggest treating Delta cloud-path execution as a follow-up (likely a resolution-rule extension to ResolvePathCredentials for Delta’s path-table plan nodes). This PR’s scope remains bare-path credential vending for the paths that work end-to-end today (parquet + parse-time delta coverage).

s-maling-celonis:
> @s-maling-celonis One more comment regarding testing Delta tables. Thank you!

@yili-db Here is the PoC https://github.com/unitycatalog/unitycatalog/pull/1754/changes/838ca09d90570eff1b668140f42415eae986830f, submitted as a separate PR, and includes commits from this PR already as the base

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
yili-db: I see. Thank you for putting up a followup PR.
As a temporary gating in this PR, can you please add a blocking to ignore delta path tables so that they don't ask UC for any credential and they continue to work using ambient credentials? That is: only parquet tables go through UC cred vending in this PR? Also with the PR description updated.

s-maling-celonis: @yili-db Addressed in 7e70896d — added a Delta exclusion gate rather than parquet-only:

- Bare delta cloud paths (and INSERT OVERWRITE DIRECTORY ... USING delta on cloud paths) skip UC path credential vending, so those paths continue using ambient storage credentials.
- Parquet and other file formats are unchanged.
- testWriteAndReadBareDeltaPath updated: local Delta e2e unchanged; cloud parse asserts credentials are not injected.
- PR description updated. Delta execution support remains in #1754.

yili-db (approved):

tdas (changes requested):
I have taken a look at this PR... and based on my Spark knowledge... hijacking the parser like this to inject credentials is not the right way. Parser should be fast and side-effect free. Analyzer is where all the resolution happens and is design for all the external IO like talkign to catalog, etc. For name-based table references, all the loadTable() happenign in the analyzer. We should follow the same model for path based tables as well. This parser based approach is NOT RECOMMENDED.

s-maling-celonis:
Re: https://github.com/unitycatalog/unitycatalog/pull/1653#pullrequestreview-4952760105 — thanks @tdas, agreed. The parser hook is gone; credential vending now happens in the analyzer.

**`ad4fea78`** — `ResolvePathCredentials` is registered via `injectHintResolutionRule`. Not `injectResolutionRule`, because `ResolveSQLOnFile` lists the path for schema inference as soon as it matches and lands in the same `Resolution` batch; the `Hints` batch is an `earlyBatch` and runs strictly before it, so the `fs.*` options are on the `UnresolvedRelation` before anything touches storage.

**`8e23389a`** — vending once per path per query. A fixed-point batch re-applies the rule until the plan stops changing, so vending on every pass isn't safe: with `fs.unitycatalog.credential.cache.enabled=false` each pass mints a fresh session token, the options never stabilize, and the batch spends a UC request per iteration until it hits its limit. The rule now skips nodes that already carry `fs.*` options.

Verified on Spark 4.0.0, 4.1.0 and 4.2.0; full `spark/test` passes.

Note: `loadTable` was not used because a bare path has no UC table row to build a `Table` from — no schema or provider, since the path-credentials API vends credentials for a location, not metadata — so returning one would mean UC taking over file-format resolution from `ResolveSQLOnFile`. It is still called and declines via `isAddressableTableName`, so Spark falls through as before.

Ready for another look.

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
```diff
@@ -0,0 +1,173 @@
...
+}
+
+object ResolvePathCredentials {
+
+  /**
+   * URI schemes for which UC can vend credentials. These match [[CloudType]] (including Hadoop's
+   * `s3a://` alias for S3). UC external locations and the path-credentials API use canonical
+   * `s3://` URLs; [[UCSingleCatalog.vendPathCredentialConf]] rewrites `s3a://` for lookup only.
+   */
+  private val CLOUD_SCHEMES = Set("s3", "s3a", "gs", "abfs", "abfss")
```
tdas: are these constants defined somewhere else in the code base for name-based table resolution? how can we make sure that the list is same? can we use the same list so that they never diverge?

tdas: cc @yili-db

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
```diff
@@ -0,0 +1,173 @@
...
+    isCloudPath(relation.multipartIdentifier.last) &&
+    !hasCredentialOptions(relation.options.keySet.asScala)
+  }
+
+  /**
+   * True when the option map already carries Hadoop filesystem credentials, either vended by an
+   * earlier pass of this rule or supplied explicitly by the user.
+   */
+  private def hasCredentialOptions(keys: Iterable[String]): Boolean =
+    keys.exists(_.toLowerCase.startsWith("fs."))
```
tdas: @yili-db are you sure this `fs.*` is the right check? will it have false positive or false negatives? 

fs.* can be any credentials. what we want is to identify whether the UC related credentials have been injected or not 
I dont know there is any way for the analyzed plan to get other non-UC fs.* or not that can confuse this check. but to minimize the assumptions and make it robust, more exact check means less likely things will get confused. 

so is there a way to make this more exactly match UC-vended credentials that we have injected?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -418,6 +421,90 @@ class UCSingleCatalog
...
+      (location, null)
+    } else {
+      val canonicalScheme =
+        if (scheme.equalsIgnoreCase("s3a")) "s3" else scheme.toLowerCase
+      val apiLocation = location.replaceFirst(
+        s"(?i)^${java.util.regex.Pattern.quote(scheme)}://",
+        s"$canonicalScheme://")
+      (apiLocation, canonicalScheme)
+    }
+  }
```
tdas: [AI-assisted]

this swallows all ApiException and only logDebugs before falling back to ambient. a transient 500 or an auth/config error then looks identical to 'not managed by UC' and surfaces downstream as a confusing s3 'no credentials' failure. can we narrow the fallback to the expected not-found/permission cases and log louder for the rest?

eating an unexpected UC failure at debug makes a real outage look like a missing external location — the FS-layer error the user sees points nowhere near the actual cause.

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
```diff
@@ -0,0 +1,173 @@
...
+              i
+            } else {
+              i.copy(storage =
+                i.storage.copy(properties = i.storage.properties ++ conf.asScala))
+            }
+        }
+    }
+  }
+
+  private def injectUnresolvedRelationCredentials(
```
tdas, line 109: [AI-assisted]

when vending returns empty (path not managed) you return the relation unchanged, so hasCredentialOptions never trips. doesnt the hints batch then re-run this rule on every fixed-point iteration and re-issue the failing READ_WRITE+READ rpcs each pass until resolution finishes? the class doc claims idempotence 'even when the cache is disabled', but testRepeatedApplicationVendsOnceAndLeavesPlanUnchanged only covers the success path. am i missing something?

every query touching an unmanaged bare path could re-hit UC with failing rpcs per analyzer iteration, and the doc overstates a guarantee that only holds once creds are actually injected.

connectors/spark/src/test/java/io/unitycatalog/spark/UCProxySuite.java
```diff
@@ -188,6 +190,63 @@ public void testLoadTableTriggersExactlyOneGetTableRpc() throws Exception {
         .getTable(eq("test_catalog.test_schema.t1"), eq(true), eq(true));
   }
 
+  /**
```
tdas, line 193: [AI-assisted]

testLoadTableReportsNotFoundForNameUcCannotAddress and testLoadTableSkipsRpcForNameUcCannotAddress are byte-for-byte the same — same path, same NoSuchTableException assert, same verify(never getTable). the 'reports not found' one doesnt actually test not-found reporting any differently. merge them, or make one distinct (e.g. drop the never-called verify from the reporting test).

two tests asserting the same thing give false confidence — one can rot silently and the count still looks green.

tdas (changes requested):
This looks much better overall. just a few more comments. ready to merge after that.

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
s-maling-celonis: Yes — name-based table vending already goes through `CloudType` via `UCCredentialHadoopConfs`. The local `CLOUD_SCHEMES` set is gone.

`isCloudPath` now uses `UCCredentialHadoopConfs.isSupportedScheme`, which lowercases then `CloudType.fromScheme` (same aliases: `s3` / `s3a` / `gs` / `abfs` / `abfss`). The `s3a://` → `s3://` rewrite stays only in `pathForCredentialApi` for the UC REST lookup.

Covered in `UCCredentialHadoopConfsTest.isSupportedSchemeMatchesCloudTypeIncludingAliasesAndCasing`. See 3118ce8c.

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
s-maling-celonis: Agreed — matching any `fs.*` key was too broad. Ambient Hadoop options (`fs.s3a.access.key`, `fs.s3a.impl`, …) could skip vending, and a miss that left the node unchanged never tripped the check.

Skip is now `hasUcPathCredentials`: `fs.unitycatalog.credentials.type=path` **and** `fs.unitycatalog.path` equal to this URI (or the canonical s3a→s3 form stored by `vendPathCredentialConf`).

`CredPropsBuilder.build()` always emits `PathCredId.props()`, including when renewal is disabled, so successful vends still carry that identity. Ambient `fs.s3a.*` keys no longer skip. See 3118ce8c.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: `vendPathCredentialConfOrEmpty` now falls back to ambient only for HTTP 400, 403, and 404. 400 covers OSS `FAILED_PRECONDITION` when no external location / bucket config covers the path.

Any other `ApiException` is `logWarning` (includes the HTTP code) and rethrown so a 500 / auth failure does not look like an unmanaged path.

Covered in `UCSingleCatalogPathCredentialTest`. See 3118ce8c.

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
s-maling-celonis: You were not missing anything — an allowed miss used to leave the node unchanged, so the hints batch re-issued PATH_READ_WRITE + PATH_READ on every fixed-point pass.

On an allowed miss we now stamp path-cred identity (`credentials.type=path`, `fs.unitycatalog.path`, operation) **without** cloud secrets. Later analyzer passes skip via `hasUcPathCredentials`. Successful vends already included that identity.

Added `testAllowedMissStampsIdentityAndSecondApplyLeavesPlanUnchanged` with the credential cache disabled. See 3118ce8c.

connectors/spark/src/test/java/io/unitycatalog/spark/UCProxySuite.java
s-maling-celonis: Merged them. Dropped `testLoadTableReportsNotFoundForNameUcCannotAddress` and kept `testLoadTableSkipsRpcForNameUcCannotAddress` (plus `testLoadTableSkipsRpcForExtensionlessBarePath` for the `/` guard). See 3118ce8c.

s-maling-celonis:
@tdas Addressed the five notes from the Aug 24 review in 3118ce8c:

1. Shared scheme list via `UCCredentialHadoopConfs.isSupportedScheme` / `CloudType`
2. Skip only on UC path-cred identity, not any `fs.*` key
3. Ambient fallback only for HTTP 400/403/404; other `ApiException`s rethrow
4. Stamp identity on an allowed miss so the hints batch does not re-RPC
5. Merged the duplicate `UCProxySuite` tests

Replies are on the threads.

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
tdas: wait wait.... i am thinking a bit harder about this. 

if the user actually specifies ambient hadoop options with the intention of it being used......what will this do now? will UC override those credentials? If so, that is not something we want .. isnt it? if the user has intentionally specified credentials that has some intended permissions (something allowed and somethign denied), we dont want to ignore that intent by replacing with another credentials which will have different permissoins..isnt it? 

cc @yili-db as well what do you think?

yili-db: Yes @tdas you have spotted a real problem: it overrides ambient credential if a cred can be vended from UC.
And a way to "stop UC vending if ambient cred is configured" can't reliably work. The following ones aren't detectable from Spark:
  - IAM instance profile / assumed role via IMDS.
  - Env-var credentials (AWS_ACCESS_KEY_ID, ...).
  - The default AWS provider chain and cloud metadata identity.

One option is to make `vendPathCredentials.enabled` default `false` so that users must explicitly opt-in. And when they do that, it's clear that they do want to override ambient cred.

s-maling-celonis:
Delta 4.3.1+ performs the UC table lookup in the outer `DeltaCatalog` before delegating to `UCProxy`. As a result, the `UCProxy`-only guard in this PR is bypassed and bare Parquet paths can still fail with HTTP 400.

I opened #1810 to reject non-Delta path identifiers in the outer `UCSingleCatalog`, while preserving delegation for bare Delta paths. That fix, or an equivalent change, is required for the credential-vending functionality in this PR to work with Delta 4.3+.

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
tdas: any who wants path-based stuff to work will enable this... and the behavior on that should be sane and not cause surprises. and for some basic functionality like path-based stuff working, that should just work out of the box, enabled by default. so i think a better proposal is the following: 

1. by default we DO NOT fetch UC credentials if user has already provided credentials. There should be a test for that.  This is basically reverting back to how it was. but please ensure that there is a test for this. 
2. in future, if needed, we can add a config which allows UC credentials to override user-provided credentials. that will be opt-in... only very few users will need to choose that.

things work out-of-the-box without surprises. 

does this make sense @s-maling-celonis @yili-db ?

tdas:
@s-maling-celonis please note this thread - https://github.com/unitycatalog/unitycatalog/pull/1653#discussion_r3866580358

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
s-maling-celonis: If the requirement is that we never overwrite ambient credentials — including instance profile, env, and the default AWS/GCP/Azure chain, not only explicit `fs.s3a.*` keys — then skipping on Hadoop secrets is not enough. Those identities never show up as `fs.*`, and we cannot probe them without either disabling vending on typical cloud VMs or failing to fall through after S3 AccessDenied.

The only way to guarantee no overwrite is **not to vend unless the user asks.** So `vendPathCredentials.enabled` should default to **false**. When they enable it, they are opting into UC replacing ambient for bare-path SQL (same as named tables).

Honoring explicit `fs.*` secrets when the flag is on is still worth doing, but it does not satisfy a no-ambient-overwrite requirement by itself.

Does that match the intended product behavior @tdas @yili-db?

tdas (approved):
LGTM. thank you for fixing this gap

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
s-maling-celonis: @tdas confirming so we don’t merge with mixed signals.

Your note here (skip UC path creds when the user already provided credentials; keep vending on by default) is **not** in this PR. Current behavior is still: `vendPathCredentials.enabled` defaults to **true**, and ambient `fs.s3a.*` (or other `fs.*`) does **not** skip vending — only a prior UC path-cred lookup does.

Given the LGTM, is that the behavior you want to ship, with the skip-on-user-creds change as a follow-up? Or should we do that skip in this PR before merge?

yili-db: > The only way to guarantee no overwrite is not to vend unless the user asks. So vendPathCredentials.enabled should default to false. When they enable it, they are opting into UC replacing ambient for bare-path SQL (same as named tables)

@s-maling-celonis
Yes, @tdas and I are aligned with this **default to false** approach so that it never overrides ambient credentials unintentionally. This also makes it safe to merge even with more things to fix in follow up PRs.
@tdas already approved. Please adjust this PR to default to false and then we can merge this.
Thank you!

yili-db (changes requested):
Please adjust this PR to default to false and then we can merge this.
Thank you!

connectors/spark/src/main/scala/io/unitycatalog/spark/ResolvePathCredentials.scala
s-maling-celonis: @yili-db thanks — default is now `false` (`7f75a339`), and the branch is rebased onto current `main`. Ready for another look.

yili-db (approved):

PR #1660, opened 2026-07-06
fix(server): close Hibernate SessionFactory when the server is disposed

tiennguyen-onehouse:
## Summary

`UnityCatalogServer` creates a `HibernateConfigurator` — and therefore a Hibernate `SessionFactory` with a JDBC connection pool — internally, but nothing ever closes it. `stop()` only shuts down the Armeria HTTP server, so the pooled physical database connections stay open until the JVM exits.

This is invisible with the default in-memory H2, but against a real database every discarded server instance leaks its pool. The sharpest case is the test suite: `BaseServerTest` boots a fresh server per test method, so a test JVM accumulates leaked pools and can eventually exhaust the database's connection limit (e.g. PostgreSQL `max_connections`).

## Stress test — reproducing the leak and proving the fix

A throwaway harness boots a `UnityCatalogServer` and immediately discards it, 40 times in a row, against a real PostgreSQL 16 instance (Testcontainers) started with `max_connections=40`, while a separate probe connection watches `SELECT count(*) FROM pg_stat_activity WHERE datname = current_database()` after each iteration. The dispose step calls `close()` when the server implements `AutoCloseable` (fix applied) and `stop()` otherwise, so the identical harness demonstrates both sides.

**Before (dispose via `stop()`)** — exactly one leaked connection per discarded server, then hard failure:

```
[stress] mode=stop() [unfixed] max_connections=40 baseline connections=1
[stress] iteration 1:  connections=2
[stress] iteration 2:  connections=3
...
[stress] iteration 38: connections=39
[stress] iteration 39: connections=40
[stress] FAILED at iteration 40 (connections=40): PSQLException: FATAL: sorry, too many clients already
```

**After (dispose via `close()`)** — flat at the probe's single connection for all 40 iterations:

```
[stress] mode=close() [fix applied] max_connections=40 baseline connections=1
[stress] iteration 1:  connections=1
...
[stress] iteration 40: connections=1
[stress] COMPLETED 40 iterations; final connections=1
```

This is also the trajectory of the real test suite: one server per test method means a same-JVM run against a default PostgreSQL (`max_connections=100`) dies at roughly the 100th test without this fix.

## Changes

- `UnityCatalogServer` implements `AutoCloseable`; the `HibernateConfigurator` is promoted from a local in `initializeServer()` to a field so its lifecycle can be managed (same single instance as before — previously the reference was simply unreachable after construction, which is why nothing could ever close it).
- `stop()` is unchanged and stays restartable — `BaseMetastoreSummaryTest` exercises `stop()` → `start()` on the same instance (closing the factory in `stop()` fails that test with `EntityManagerFactory is closed`).
- New `close()` stops the server and closes the `SessionFactory` in a `finally` block, so a failing `stop()` cannot leak the pool. Idempotent, and safe before `start()` (Armeria's `stop()` no-ops when not started; Hibernate's native `SessionFactory.close()` tolerates repeat calls).
- The constructor closes the factory if `initializeServer()` throws after the factory was built (e.g. `initMetastoreIfNeeded()` failing on a broken schema), so a failed boot does not leak.
- `BaseServerTest.tearDown()` disposes the per-test server with `close()`, also closes the test harness's own factory, and nulls the server so tearDown is idempotent.
- `BaseAccessControlCRUDTest`'s casbin cleanup moved from a separate `@AfterEach` (`cleanUp()`) into the `tearDown()` override chain. JUnit runs the most-derived `tearDown` override (e.g. `SdkAccessControlBaseCRUDTest`) before less-derived `@AfterEach` methods, so once tearDown closes the factory, the old `cleanUp()` failed with `EntityManagerFactory is closed`. `casbin_rule` is created by casbin's `JDBCAdapter` outside Hibernate's `create-drop`, so it must still be cleared before the factory closes.

## Known adjacent gaps (intentionally out of scope)

- `JCasbinAuthorizer` opens its own raw `DriverManager` connection via casbin's `JDBCAdapter`; it is not covered by `close()` and still leaks one connection per auth-enabled server instance. Fixing it needs a lifecycle method on `UnityCatalogAuthorizer` — happy to do as a follow-up.
- The server does not configure Armeria's `gracefulShutdownTimeout`, so `stop()` does not formally wait for in-flight handlers. Safe today because all DB-touching services are synchronous event-loop handlers, but worth revisiting if async/`@Blocking` services are added.

## How we tested

```
build/sbt "server/testOnly *SdkDeltaCommitsCRUDTest *MetastoreSummaryTest *AccessControl* *AuthCRUD*"
Passed: Total 26, Failed 0, Errors 0, Passed 26
```

Covers the per-test boot/dispose cycle, the stop/start restart path, and the auth-enabled hierarchy with the double-teardown pattern — plus the connection-exhaustion stress test above.

## PR Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
+++ b/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
@@ -66,7 +66,7 @@ import org.apache.logging.log4j.core.config.Configurator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class UnityCatalogServer {
+public class UnityCatalogServer implements AutoCloseable {
   private static final Logger LOGGER = LoggerFactory.getLogger(UnityCatalogServer.class);
   private static final String BASE_PATH = "/api/2.1/unity-catalog/";
   private static final String CONTROL_PATH = "/api/1.0/unity-control/";
@@ -75,6 +75,9 @@ public class UnityCatalogServer {
   private final Server server;
   private final ServerProperties serverProperties;
   private final SecurityContext securityContext;
+  private final HibernateConfigurator hibernateConfigurator;
+  /** True when this server built the configurator itself and must therefore close it. */
+  private final boolean ownsHibernateConfigurator;
 
   static {
     System.setProperty("log4j.configurationFile", "etc/conf/server.log4j2.properties");
@@ -89,7 +92,36 @@ public class UnityCatalogServer {
     this.securityContext =
         new SecurityContext(configurationFolder, securityConfiguration, "server", INTERNAL);
     this.serverProperties = unityCatalogServerBuilder.serverProperties;
-    this.server = initializeServer(unityCatalogServerBuilder);
+    // An injected configurator stays the caller's to close; only a self-built one is ours.
+    this.ownsHibernateConfigurator = unityCatalogServerBuilder.hibernateConfigurator == null;
+    this.hibernateConfigurator =
+        ownsHibernateConfigurator
+            ? new HibernateConfigurator(serverProperties)
+            : unityCatalogServerBuilder.hibernateConfigurator;
+    try {
+      this.server = initializeServer(unityCatalogServerBuilder);
+    } catch (Throwable t) {
+      // Construction failed after the SessionFactory was built; close it so a failed boot does
+      // not leak its connection pool. Errors matter as much as RuntimeExceptions here: a
+      // NoClassDefFoundError out of initializeServer() would leak the pool just the same.
+      closeOwnedSessionFactory(t);
+      throw t;
+    }
+  }
+
+  /**
+   * Closes the SessionFactory if this server created it, leaving an injected one to its owner. A
+   * failure to close is attached to {@code primaryFailure} so it cannot mask the original error.
+   */
+  private void closeOwnedSessionFactory(Throwable primaryFailure) {
+    if (!ownsHibernateConfigurator) {
+      return;
+    }
+    try {
+      hibernateConfigurator.getSessionFactory().close();
+    } catch (Throwable closeFailure) {
+      primaryFailure.addSuppressed(closeFailure);
+    }
   }
 
   private void setDefaults(UnityCatalogServer.Builder unityCatalogServerBuilder) {
@@ -107,11 +139,6 @@ public class UnityCatalogServer {
             .http(unityCatalogServerBuilder.port)
             .serviceUnder("/docs", new DocService());
 
-    // Init hibernate
-    HibernateConfigurator hibernateConfigurator =
-        unityCatalogServerBuilder.hibernateConfigurator != null
-            ? unityCatalogServerBuilder.hibernateConfigurator
-            : new HibernateConfigurator(unityCatalogServerBuilder.serverProperties);
     // Init all repositories
     Repositories repositories =
         new Repositories(hibernateConfigurator.getSessionFactory(), serverProperties);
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -89,7 +90,15 @@ private UnityCatalogServer(UnityCatalogServer.Builder unityCatalogServerBuilder)
     this.securityContext =
         new SecurityContext(configurationFolder, securityConfiguration, "server", INTERNAL);
     this.serverProperties = unityCatalogServerBuilder.serverProperties;
-    this.server = initializeServer(unityCatalogServerBuilder);
+    this.hibernateConfigurator = new HibernateConfigurator(serverProperties);
+    try {
+      this.server = initializeServer(unityCatalogServerBuilder);
+    } catch (RuntimeException e) {
```
yili-db: should this be `Throwable`?

yili-db (commented):
Thank you for fixing this real bug.
Worth noting is that this leak only happens in test suites where server is repeatedly started and stoped between different tests.
In prod this is not a problem as the server never stops (until JVM stops).

tiennguyen-onehouse:
Note on interaction with #1446 (raised by @yili-db there): #1446 adds an optional injected `HibernateConfigurator` on `UnityCatalogServer.Builder`, documented as caller-owned. This PR's `close()` is consistent with the *create-owns-close* contract as written, because on this base the server always builds its own configurator. The two PRs conflict textually in `UnityCatalogServer` (same hunk), so whichever merges second gets a reconciliation commit that makes the contract explicit: an `ownsHibernateConfigurator` flag set only when the server built the configurator itself, with `close()` closing the session factory only in that case (an injected one stays owned by its creator — `BaseServerTest.tearDown()` here already closes it explicitly).

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
tiennguyen-onehouse: Good call — taken, in `8de9ca74`.

You're right that `RuntimeException` is too narrow. This is a cleanup-and-rethrow rather than a swallow, so the usual "don't catch `Throwable`" caution doesn't apply here: an `Error` out of `initializeServer()` — a `NoClassDefFoundError` or `ExceptionInInitializerError` during boot is entirely plausible — would leak the connection pool exactly like a `RuntimeException` would. And because we rethrow, Java's precise rethrow keeps the constructor's signature unchanged, so widening costs nothing.

One extra wrinkle while I was in there: the close itself could throw and mask the original failure, so it's now attached as suppressed instead.

```java
try {
  this.server = initializeServer(unityCatalogServerBuilder);
} catch (Throwable t) {
  // Construction failed after the SessionFactory was built; close it so a failed boot does
  // not leak its connection pool. Errors matter as much as RuntimeExceptions here: a
  // NoClassDefFoundError out of initializeServer() would leak the pool just the same.
  closeOwnedSessionFactory(t);
  throw t;
}
```

where `closeOwnedSessionFactory` closes only a self-built factory and does `primaryFailure.addSuppressed(closeFailure)` if closing fails.

tiennguyen-onehouse:
Heads-up on the latest push (`8de9ca74`): **this PR is now stacked on top of #1446.**

**Why.** As @yili-db noted on #1446, the two PRs collide: #1446 lets callers inject a `HibernateConfigurator` via `UnityCatalogServer.Builder` and documents that *"the server never closes this factory, so the caller owns its lifecycle"* — while this PR closed the factory unconditionally. Merged independently, this one would close an **injected** factory and break #1446's shared-ownership harness. Rebasing here was the only way to make the two consistent.

**What changed vs the previous revision:**

1. **Ownership is now explicit.** The server closes the configurator only when it built one itself; an injected one is left to its owner. That makes #1446's builder javadoc true rather than aspirational, and it also answers @tdas's point on #1446 ([review](https://github.com/unitycatalog/unitycatalog/pull/1446#pullrequestreview-4687409493)) — the server now genuinely closes the one it builds, so the ownership contrast in the docs is real. The permanent fix he asked for lands here, as @yili-db anticipated.
2. **`catch (Throwable)` instead of `RuntimeException`**, per @yili-db's review comment, with the close failure attached as suppressed so it can't mask the original error.
3. **`BaseServerTest`** closes the factory it injected (the server won't), and a stale comment claiming otherwise is corrected.

**Reviewing this PR:** the diff currently shows #1446's changes as well, because GitHub only allows a PR base in the upstream repo — a fork PR can't be retargeted at another fork branch. **Only the last commit (`8de9ca74`, 3 files) belongs to this PR.** Once #1446 merges I'll rebase and the diff will collapse to just those files.

**Verified locally** (JDK 17): compiles, `javafmtCheck` + checkstyle pass, and 46/46 tests green — including `SdkMetastoreSummaryTest` (the `stop()` → `start()` restart path that must keep working now that `close()` is the terminal operation), the access-control tearDown chain, and all three DeltaCommits suites (H2 + real PostgreSQL 16 + real MySQL 8.4).

yili-db:
https://github.com/unitycatalog/unitycatalog/pull/1446 is merged. Please rebase.

tiennguyen-onehouse:
Rebased onto current `main` now that #1446 has merged, and force-pushed. The branch is reduced to a single commit containing only this PR's change — the 3 session-factory files (`UnityCatalogServer`, `BaseServerTest`, `BaseAccessControlCRUDTest`); the #1446 content it was previously stacked on is gone. `catch (Throwable)` is carried in per @yili-db's review comment above. `MERGEABLE`, and locally: `MetastoreSummaryTest` (the `stop()`/`start()` restart path) + the access-control tearDown suites pass, format/checkstyle clean.

yili-db (approved):

yili-db:
Merged. Thank you for your contribution!

PR #1657, opened 2026-07-06
Support wildcard patterns in server.allowed-issuers

s-maling-celonis:
## Summary
- Add `IssuerAllowlist` to match `server.allowed-issuers` entries with `*` (single DNS label)
- Use wildcard matching in `AuthService` token exchange issuer validation
- Document wildcard usage in `docs/server/auth.md`

Example: `server.allowed-issuers=https://*.dev.example.com` accepts tokens from `https://dev.dev.example.com`, `https://backend.dev.example.com`, etc.

## Test plan
- [x] `IssuerAllowlistTest` unit tests (exact, wildcard, mixed patterns)

Made with [Cursor](https://cursor.com)

PR #1661, opened 2026-07-07
Wildcard issuer/audience matching and subject-token principal resolution

s-maling-celonis:
## Summary

Extends token exchange for multi-tenant OAuth deployments with four related changes:

### Wildcard `server.allowed-issuers` ([supersedes #1657](https://github.com/unitycatalog/unitycatalog/pull/1657))
- Add `IssuerAllowlist` for single-DNS-label wildcard issuer patterns (e.g. `https://*.dev.example.com`)
- Reuse the same matcher in `AudienceAllowlist`

### Wildcard `server.audiences`
- Validate JWT `aud` after signature verification (any token audience may match any configured entry)
- Support `server.audiences=*` alone to disable audience checks for dynamic OAuth client UUID audiences

### Subject-token principal resolution (service principals)
UC resolves the principal from the **subject token** on `POST /auth/tokens` — no OAuth client credentials on the UC request:

1. **`id_token`** (CLI/browser): `email` claim or `sub` → `getUserByEmail`, then `externalId` fallback via `azp`/non-URL `aud`
2. **`access_token`** (M2M): if no `email` claim, `externalId` via `azp`/non-URL `aud` first, then `sub`; otherwise same as `id_token`

The issued UC access token always uses the resolved user's **email** as `sub` (for grants/authorization).

Additional behavior:
- **Audience bypass**: signed subject tokens whose `azp` or non-URL `aud` matches a registered UC user `externalId` are accepted even when the client UUID is not in `server.audiences`
- **Unique index** on `uc_users.external_id` (multiple `NULL` values still allowed)
- Prefer the `externalId` flow over `server.audiences=*` for per-tenant service principals

### UC access token expiry
- Add `server.access-token-timeout` (default `PT24H`) for access tokens issued by token exchange
- Set JWT `exp` via `withExpiresAt` in `SecurityContext.createAccessToken`
- Return `expires_in` in the `/auth/tokens` response so clients can cache and refresh correctly

### Chunked JSON payload authorization
`UnityAccessDecorator` previously ran `beforeRequest` on every `peekData` chunk, before the full JSON body was buffered and parsed. For chunked `POST` requests (e.g. `POST /schemas`), authorization could run on a partial payload and fail to resolve payload-derived resource keys such as `catalog_name`, causing:

```
Cannot invoke "java.util.UUID.toString()" because "resource" is null
```

This fix:
- Defers authorization until `processPeekData` has successfully parsed the complete JSON payload
- Checks the accumulated payload buffer (not the current chunk) for JSON completion

Example multi-tenant configuration:

```properties
server.allowed-issuers=https://*.dev.example.com
server.audiences=unity-catalog-local,*.dev.example.com
server.access-token-timeout=PT24H
```

Bootstrap UC users with `email` for grants and `externalId` = OAuth `client_id`.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
@@ -183,9 +183,10 @@ public class UnityAccessDecorator implements DecoratingHttpServiceFunction {
         // This code block is not called immediately. It is called later as part of delegate.serve()
         LOGGER.debug("Authorization peekData invoked.");
 
-        peekDataHandler.processPeekData(data);
-        Map<SecurableType, UUID> resourceIds = mapResourceKeys(resourceKeys, nonResourceValues);
-        evaluateAction.beforeRequest(principal, expression, resourceIds, nonResourceValues);
+        if (peekDataHandler.processPeekData(data)) {
+          Map<SecurableType, UUID> resourceIds = mapResourceKeys(resourceKeys, nonResourceValues);
+          evaluateAction.beforeRequest(principal, expression, resourceIds, nonResourceValues);
+        }
       });
     }
 
@@ -426,7 +427,8 @@ public class UnityAccessDecorator implements DecoratingHttpServiceFunction {
         // Unfortunately we don't appear to get a signal that we've got all the data, so we have to
         // resort to attempting to parse the data whenever it _looks_ complete.
         // TODO: try to optimize this using Jackson streaming or something else.
-        if (data.array()[data.array().length - 1] == '}') {
+        byte[] accumulated = dataStream.toByteArray();
+        if (accumulated.length > 0 && accumulated[accumulated.length - 1] == '}') {
           try {
             Map<String, Object> payload = MAPPER.readValue(
                 dataStream.toByteArray(),
--- a/docs/server/auth.md
+++ b/docs/server/auth.md
@@ -56,7 +56,7 @@ server.audiences=<Client ID provided earlier>
 When authorization is enabled, the server validates incoming identity tokens against configured issuers and audiences:
 
 - **server.allowed-issuers**: Comma-separated list of allowed token issuers (exact match). Tokens from issuers not in this list will be rejected. This prevents attackers from using their own identity provider to forge tokens.
-- **server.audiences**: Comma-separated list of expected JWT audience values. Tokens must contain one of these audience values. This is typically your application's client ID.
+- **server.audiences**: Comma-separated list of expected JWT audience values. Tokens must contain an `aud` value matching one of these entries (exact match or wildcard with `*`). A single value of `*` disables audience validation (issuer and user checks still apply); that sentinel cannot be combined with other values.
 
 #### Multiple Identity Providers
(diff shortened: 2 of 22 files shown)


shkelzeen (approved):

shkelzeen (approved):

shkelzeen (approved):

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -183,9 +183,10 @@ private HttpResponse authorizeByRequest(
         // This code block is not called immediately. It is called later as part of delegate.serve()
         LOGGER.debug("Authorization peekData invoked.");
 
-        peekDataHandler.processPeekData(data);
-        Map<SecurableType, UUID> resourceIds = mapResourceKeys(resourceKeys, nonResourceValues);
-        evaluateAction.beforeRequest(principal, expression, resourceIds, nonResourceValues);
+        if (peekDataHandler.processPeekData(data)) {
```
yili-db, line 186: We are aware of this. But adding back this `if` alone will bring back the auth skipping vulnerability. `processPeekData` could always return false for a valid request.
I am revamping https://github.com/unitycatalog/unitycatalog/pull/1494 to fix it.
Can you revert these 3 lines for now please?

yili-db:
This is valuable work, but the PR bundles several independent changes on the auth path. Since these are all security-sensitive, each needs to be reviewed carefully on its own — bundled together it's hard to reason about any one of them, and they get tied to the same merge. Can we split it into different PRs?

Rough breakdown, in dependency order:

1. **Access-token expiry** (`server.access-token-timeout`, `withExpiresAt`, `expires_in`). Small and self-contained.

2. **Wildcard issuer/audience matching** (`IssuerAllowlist`, `AudienceAllowlist`, `audiences=*` handling).

3. **Subject-token principal resolution** (`TokenExchangePrincipalResolver`, `externalId` lookup + unique constraint, audience-bypass-via-`externalId`). The most sensitive part — it relaxes audience binding during signature verification and accepts cross-audience tokens for registered clients, so it needs its own focused review and sign-off. Depends on (2).

This lets the expiry and wildcard changes land quickly and keeps each PR focused.

Thank you!

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
s-maling-celonis: Sure, i will be closing this PR anyway in favor of the 3 new independent changes. Wrt. this change, i think we can just manually apply the changes from the linked PR onto our fork.

s-maling-celonis:
Closing in favor of three focused PRs (per review feedback to split security-sensitive auth changes):

1. **Access token expiry** — https://github.com/unitycatalog/unitycatalog/pull/1689
2. **Wildcard issuer/audience matching** — https://github.com/unitycatalog/unitycatalog/pull/1690
3. **Subject-token principal resolution** — https://github.com/unitycatalog/unitycatalog/pull/1691

The chunked JSON authorization fix (`UnityAccessDecorator`) is deferred until #1494 lands.

PR #1663, opened 2026-07-08
Make auth config lookups case-insensitive in TokenProvider.create

Bhavya1119:
## What
Normalize auth configuration keys to be case-insensitive in `TokenProvider.create`,
so all `TokenProvider` implementations (OAuth, static, custom) resolve their config
keys regardless of key case.

## Why
When the Spark/Delta Unity Catalog integration loads a table, catalog options are
routed through Spark's `CaseInsensitiveStringMap`, which lower-cases all keys. Those
options are then handed to `TokenProvider.create`. Because `OAuthTokenProvider.initialize`
reads keys case-sensitively (`configs.get("oauth.clientId")`), the lower-cased
`oauth.clientid` is not found and OAuth fails with:

    java.lang.IllegalArgumentException: Configuration key 'oauth.clientId' is missing or empty
      at io.unitycatalog.client.auth.OAuthTokenProvider.initialize(OAuthTokenProvider.java:82)
      at io.unitycatalog.client.auth.TokenProvider.create(TokenProvider.java:143)
      at org.apache.spark.sql.delta.coordinatedcommits.UCTokenBasedRestClientFactory$.createUCClient(...)
      at org.apache.spark.sql.delta.catalog.UCDeltaCatalogClientImpl$.fromCatalogOptions(...)
      ...

Observed with: unitycatalog-spark/client 0.5.0, delta-spark 4.3.0, Apache Spark 4.1.0,
authenticating an external Spark client to Databricks Unity Catalog via OAuth M2M
(client credentials).

Config keys generally should not be case-sensitive, so normalizing once at the factory
fixes this for every provider without changing the public config contract.

## Change
- `TokenProvider.create` normalizes the incoming config map to be case-insensitive
  before dispatching to the provider's `initialize`.

## Tests
- Added `testCreateResolvesOAuthConfigCaseInsensitively` to `TokenProviderTest`:
  passes OAuth keys lower-cased (`oauth.clientid`, `oauth.clientsecret`) and asserts
  `create` builds the `OAuthTokenProvider` and exposes the canonical config keys.
- `build/sbt client/test` and `build/sbt testJavastyle` pass.

## Notes
The root interaction originates in Delta case-folding pass-through auth keys; this PR
hardens the client side so it's robust to that. Happy to also file a Delta-side issue
if useful.

### DIFF
--- a/clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
+++ b/clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
@@ -2,6 +2,7 @@ package io.unitycatalog.client.auth;
 
 import io.unitycatalog.client.internal.Preconditions;
 import java.util.Map;
+import java.util.TreeMap;
 
 /**
  * Interface for providing access tokens to authenticate with Unity Catalog.
@@ -107,7 +108,9 @@ public interface TokenProvider {
    *     not found, no default constructor, or instantiation failure)
    */
   static TokenProvider create(Map<String, String> configs) {
-    String authType = configs.get(AuthConfigs.TYPE);
+    Map<String, String> ci = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+    ci.putAll(configs);
+    String authType = ci.get(AuthConfigs.TYPE);
     Preconditions.checkArgument(
         authType != null && !authType.trim().isEmpty(),
         "Required configuration key '%s' is missing or empty. "
@@ -140,7 +143,7 @@ public interface TokenProvider {
     }
 
     // Initialize the TokenProvider with configs.
-    tokenProvider.initialize(configs);
+    tokenProvider.initialize(ci);
     return tokenProvider;
   }
 }
--- a/clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
+++ b/clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
@@ -108,9 +108,11 @@ public interface TokenProvider {
    *     not found, no default constructor, or instantiation failure)
    */
   static TokenProvider create(Map<String, String> configs) {
-    Map<String, String> ci = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
-    ci.putAll(configs);
-    String authType = ci.get(AuthConfigs.TYPE);
+    // Try normalizing to a case-insensitive view if possible so downstream lookups regardless of
+    // casing
+    Map<String, String> normalizedConfigs = toCaseInsensitiveMap(configs);
+
+    String authType = normalizedConfigs.get(AuthConfigs.TYPE);
     Preconditions.checkArgument(
         authType != null && !authType.trim().isEmpty(),
         "Required configuration key '%s' is missing or empty. "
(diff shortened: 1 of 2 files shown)


clients/java/src/main/java/io/unitycatalog/client/auth/TokenProvider.java
```diff
@@ -107,7 +108,9 @@ public interface TokenProvider {
    *     not found, no default constructor, or instantiation failure)
    */
   static TokenProvider create(Map<String, String> configs) {
-    String authType = configs.get(AuthConfigs.TYPE);
+    Map<String, String> ci = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
yili-db: What the difference between `new TreeMap<>(String.CASE_INSENSITIVE_ORDER);` and `return new CaseInsensitiveStringMap(newConfigs);` (from AuthConfigUtils.java).
And the input to this function could be a CaseInsensitiveStringMap already:
```
tokenProvider = TokenProvider.create(AuthConfigUtils.buildAuthConfigs(options))
```

I propose a helper that does this to avoid duplicate wrapping:
```
if the map is a CaseInsensitiveStringMap: return as-is
otherwise: return CaseInsensitiveStringMap(the input map)
```

Bhavya1119: The difference is the module boundary. CaseInsensitiveStringMap here is org.apache.spark.sql.util.CaseInsensitiveStringMap which is a Spark class. AuthConfigUtils can use it because it lives in the Spark connector module . But TokenProvider is in clients/java, the standalone UC Java client, which is intentionally Spark-independent. 

Using Spark's CaseInsensitiveStringMap there would pull a spark-sql dependency into the client. new TreeMap<>(String.CASE_INSENSITIVE_ORDER) gives the same case-insensitive lookup with only the JDK, so the client stays dependency-free. For lookups they're functionally equivalent, and TreeMap (via compareToIgnoreCase) also avoids the locale pitfalls of lower-casing keys.

On the double-wrapping: correct , through the connector path the input is already a CaseInsensitiveStringMap (from buildAuthConfigs), so create re-wraps it. That's a one-time copy of a ~4-entry map, so the cost is negligible, and it's correct since a CaseInsensitiveStringMap is still a Map<String, String>.

On the proposed instanceof CaseInsensitiveStringMap ? as-is : wrap helper: that would require importing Spark's CaseInsensitiveStringMap into the client module, which re-introduces the coupling we're trying to avoid. Also worth noting why the fix belongs in TokenProvider.create and not AuthConfigUtils: the OAuth failure originates on the Delta path (UCTokenBasedRestClientFactory.extractAuthConfig → TokenProvider.create), which builds a plain, lower-cased Map and does not go through AuthConfigUtils , so it isn't a CaseInsensitiveStringMap there. create is the single chokepoint both the connector and Delta callers share, so normalizing there makes it robust regardless of the map type the caller passes.

If avoiding the extra allocation matters, I'm happy to add a small JDK-only guard in the client rather than depend on spark's type

yili-db: You are right. This code should not use `CaseInsensitiveStringMap`.

I still think `AuthConfigUtils` should be responsible to return an `CaseInsensitiveStringMap`. Because the responsibility boundary between `AuthConfigUtils` and `TokenProvider.create` should be like:
- Each function should expect their **input** map be either a case sensitive map with correct casing, or a case insensitive map.
- Each function should return **output** as either a case sensitive map with correct casing, or a case insensitive map.
- No function should returning case sensitive map with incorrect casing and assume downstream to tolerant that.

With that, I still think your change is a good defense in depth. You can make it a helper and avoid a double wrapping without importing `CaseInsensitiveStringMap`. Just a simple trick to check the class name: `obj.getClass().getSimpleName().equals("CaseInsensitiveStringMap")`.

Thank you!

Bhavya1119: Thanks for the review! Addressed both points:

Extracted the normalization into a private static toCaseInsensitiveMap helper on TokenProvider. It returns the map as-is when it's already case-insensitive  detected via getClass().getSimpleName().equals("CaseInsensitiveStringMap") as you suggested, so the client keeps no compile-time dependency on Spark's CaseInsensitiveStringMap and otherwise copies into a TreeMap(String.CASE_INSENSITIVE_ORDER). This avoids the double-wrapping on the connector path.

Agree with the responsibility boundary. AuthConfigUtils already returns a CaseInsensitiveStringMap, so it's compliant; the map with incorrect casing comes from the Delta path (UCTokenBasedRestClientFactory.extractAuthConfig), which doesn't go through AuthConfigUtils. So this create-level normalization is the defense-in-depth for callers that bypass it.

Also added testCreatePassesThroughAlreadyCaseInsensitiveMap to cover the pass-through branch (existing testCreateResolvesOAuthConfigCaseInsensitively covers the wrap branch). Full client suite is green.

yili-db (approved):

Bhavya1119:
Lets add this to latest release as well.

PR #1666, opened 2026-07-10
[Spec] Clarify commit retry and rebase rules in ManagedTablesSpec

yili-db:
**PR Checklist**

**Description of changes**
Expand the "Commit recovery" section under "Write a Commit" to spell out when a client may safely retry versus rebase an `add-commit`.


spec/protocols/ManagedTablesSpec.md
```diff
@@ -862,14 +862,35 @@ Publishing (copying a ratified commit file to `_delta_log/<v>.json`) and reporti
...
+`add-commit` carries no idempotency token in this protocol version. Instead, the server deduplicates retries using:
+
+- the UUID embedded in `file-name` (the primary handle), and
+- the commit file's content (a backup handle the server may consult once the commit for that version has been backfilled).
+
+A retry is deduplicated only when these are unchanged from the original attempt.
+
+**Retrying the same RPC on `5xx` or timeout is safe.** A `5xx` or timeout leaves the commit's outcome unknown. The client may simply resend the same RPC. Because the dedup handle is unchanged, the server resolves the retry to the actual outcome of the original attempt instead of treating it as a new commit — so a retry is never harmful. Each retry returns exactly one of:
+
+- **`200`** — the commit landed. This includes the case where an earlier attempt already landed and the dedup handle recognizes this retry as the same commit.
```
yili-db: Should state that in this case the response body is not the post-commit table state, but the newest table state that may contain modifications from later commits as well.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -862,14 +862,35 @@ Publishing (copying a ratified commit file to `_delta_log/<v>.json`) and reporti
...
+
+- **`200`** — the commit landed. This includes the case where an earlier attempt already landed and the dedup handle recognizes this retry as the same commit. In that recognized-retry case the response body is not the post-commit table state but the newest table state, which may also contain modifications from later commits.
+- **`409` conflict** — the commit definitively did *not* land; the client must rebase (see below) before trying again. This does not necessarily mean a different commit occupies the target version — it may be a version conflict (`CommitVersionConflictException`: another writer won the version) or a precondition failure (`UpdateRequirementConflictException`: an `assert-etag` / `assert-table-uuid` requirement no longer holds because the table changed underneath the proposal, e.g. a metadata-only change).
```
timothyw553, line 875: we're saying that if the client tries to commit $X$ and receives a 409 then the client must rebase to the latest table version (maybe $X$) and then recreate the new version-numbered commit (version $X+1$). but actually 409 doesn't always mean that the original commit $X$ did not land.

`lockTableForCommit` maps `PessimisticLockException` to `UpdateRequirementConflictException (409)`.
```
1. Client sends commit X.
2. The server starts commit X and holds the table lock.
3. The client times out, but the server continues.
4. The client sends an identical retry for X.
5. The retry cannot get the lock.
6. The server returns UpdateRequirementConflictException (409).
7. The original request finishes and commits X.
8. The client treats 409 as a real conflict.
9. The client rebuilds the change as X+1.
10. The same logical change is committed twice.
```
either 409 must mean “outcome unknown,” not “rebase.” or we have to change the error code for the server response; maybe just map `PessimisticLockException` to `CommitStateUnknownException`

spec/protocols/ManagedTablesSpec.md
```diff
@@ -862,14 +862,35 @@ Publishing (copying a ratified commit file to `_delta_log/<v>.json`) and reporti
 
 ##### Commit recovery
 
-`add-commit` carries no idempotency token in this protocol version; the UUID embedded in `file-name` is the deduplication handle. When a commit outcome is unknown (`CommitStateUnknownException`, a network failure, or a process restart mid-commit), the client recovers as follows:
+`add-commit` carries no idempotency token in this protocol version. Instead, the server deduplicates retries using:
+
+- the UUID embedded in `file-name` (the primary handle), and
+- the commit file's content (a backup handle the server may consult once the commit for that version has been backfilled).
```
timothyw553: A table name can be reused. A table UUID cannot.

Example:
```
1. Client loads table "orders".
       UUID = AAA
2. Client prepares a commit for that table.
3. Someone deletes "orders".
4. Someone creates a new table named "orders".
       UUID = BBB
5. Client sends its old commit with:
       assert-table-uuid = AAA
6. Server finds UUID BBB and returns 409.
```
The UUID check prevented the client from changing the wrong table.

here we're saying the client should rebase after that 409. A careless client could reload orders, use UUID BBB, and apply the old change to the new table. That is unsafe.

The correct rule is:

If only the etag changed, rebuild the commit. If the table UUID changed, stop. Do not commit to the new table automatically.
```diff
- - the commit file's content (a backup handle the server may consult once the commit
- for that version has been backfilled).
+ - the commit file's content. Once the catalog record has been removed, the server
+ must compare the staged and published commit contents before returning 200 or 409.
+ If it cannot complete the comparison, it must return `CommitStateUnknownException`.
```

spec/protocols/ManagedTablesSpec.md
```diff
@@ -862,14 +862,35 @@ Publishing (copying a ratified commit file to `_delta_log/<v>.json`) and reporti
...
+- **`429`** (`ResourceExhaustedException`) — the unbackfilled-commit limit is reached. The commit did not land, but retrying (identically *or* rebased) will keep failing until the client publishes and reports backfill for pending versions; back off and backfill first.
+- **another `5xx` / timeout** (`CommitStateUnknownException`, or a transient server error) — the outcome is still unknown; keep retrying the same RPC.
+
+**A retry without rebasing must be byte-for-byte identical.** To retry an unknown-outcome commit *without* rebasing, resend the exact same commit — same target `version`, same `file-name` (same UUID), and same staged commit file content. Never regenerate the file, allocate a new UUID, or bump the version. Any such change makes the request a new, distinct commit proposal and defeats deduplication.
```
timothyw553:
```diff
-To retry an unknown-outcome commit *without* rebasing, resend the exact same commit —
-same target `version`, same `file-name` (same UUID), and same staged commit file content.
+To retry without rebasing, resend the exact same `updateTable` request body, including
+all requirements, update actions, commit fields, and the same staged file content.
+After resolving the original result, send any new metadata or backfill actions in a
+separate request.
```

spec/protocols/ManagedTablesSpec.md
yili-db: Yes this is a real bug. I just sent https://github.com/unitycatalog/unitycatalog/pull/1828 to fix.
409 must mean "rebase".

spec/protocols/ManagedTablesSpec.md
yili-db: This is now rephrased as `Reloading must confirm the table is still the same object the commit targeted...`
Therefore 409 is still "rebase and retry" error, but rebase needs to compare the table ID.
It's generally hard and impractical to force clients to compare different error messages caused by table id change. Most clients feel comfortable to compare HTTP code only.

timothyw553 (approved):
LGTM after #1828 merges and this branch is rebased.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -862,14 +862,35 @@ Publishing (copying a ratified commit file to `_delta_log/<v>.json`) and reporti
...
+- **checking table status** — reloading the table and seeing that the proposed version was won by another writer, or is otherwise not its own commit.
+
+**Warning:** Never rebase to *learn* the outcome. Submitting a rebased commit (a new version / commit file / UUID) while the outcome is still unknown — or when the attempt actually landed — creates a **duplicate commit**: the rebased proposal re-commits the same logical change on top of a state that already includes it. Always resolve the outcome first (below), then decide whether to resend identically or to rebase.
+
+When the outcome is unknown (`CommitStateUnknownException`, a network failure, or a process restart mid-commit), the client resolves it in one of two ways. The simplest is to resend the identical RPC (safe, as above) and read its response: `200` means it landed; a `409` (`CommitVersionConflictException` / `UpdateRequirementConflictException`) means it did not, so rebase; another `5xx` / timeout leaves it unknown, so keep resending. Alternatively, the client can inspect table state directly:
```
zhenlineo, line 888: Isn't `CommitStateUnknownException` an end state? The server has checked and failed to tell the state, retrying will just continuously receive the same result. We should not retry this error any more.

yili-db: It could be caused by a transient problem on server side, therefore client should retry (with backoff and limit).
Eventually if it really can't get a different response and client can give up and fail.
This is similar to "connection timeout" etc.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -862,14 +862,35 @@ Publishing (copying a ratified commit file to `_delta_log/<v>.json`) and reporti
...
 
 1. Reload the table via [`loadTable`](#load-a-table).
 2. If `latest-table-version` >= `v` (the proposed version), look up version `v`: in the `commits` array, or in `_delta_log` if already published. If its `file-name` matches the UUID filename the client generated, the commit was accepted; resume from publishing. If it differs, a concurrent writer won version `v`; rebuild the snapshot and retry at the new latest version + 1.
 3. If `latest-table-version` is `v-1`, the proposal was not accepted; re-send the same `add-commit`.
```
tdas: [AI-assisted]

step 2 (unchanged) still tells the client to resolve by matching version v's `file-name` against the UUID name it generated. but once the version is published+backfilled the file in `_delta_log` is `<v>.json`, and per your new paragraph the server may have purged the UUID record entirely — so file-name will never match. if content is now the backup handle, shouldnt the client inspection compare content too? otherwise how does a client ever resolve a retry of a purged commit — resend returns CommitStateUnknownException, loadTable can't match by name, and it loops.

you added a content-based handle for the server but left the client-side resolution keyed on the UUID name, so the two halves of the section describe different worlds and a real recovery path terminates in a livelock.

spec/protocols/ManagedTablesSpec.md
```diff
@@ -862,14 +862,35 @@ Publishing (copying a ratified commit file to `_delta_log/<v>.json`) and reporti
...
+- **checking table status** — reloading the table and seeing that the proposed version was won by another writer, or is otherwise not its own commit.
+
+**Warning:** Never rebase to *learn* the outcome. Submitting a rebased commit (a new version / commit file / UUID) while the outcome is still unknown — or when the attempt actually landed — creates a **duplicate commit**: the rebased proposal re-commits the same logical change on top of a state that already includes it. Always resolve the outcome first (below), then decide whether to resend identically or to rebase.
+
+When the outcome is unknown (`CommitStateUnknownException`, a network failure, or a process restart mid-commit), the client resolves it in one of two ways. The simplest is to resend the identical RPC (safe, as above) and read its response: `200` means it landed; a `409` (`CommitVersionConflictException` / `UpdateRequirementConflictException`) means it did not, so rebase; another `5xx` / timeout leaves it unknown, so keep resending. Alternatively, the client can inspect table state directly:
```
tdas: [AI-assisted]

the 200/409/429/5xx outcomes are spelled out in the bullet list up top, then re-listed again in the closing 'resolves it in one of two ways' paragraph, and both restate the errors table below (1117-1124). can we tighten? pick one place for the per-code outcomes and link the rest.

same fact stated in three places drifts the moment a code's meaning changes, and the tripled length reads like an essay rather than a spec.

PR #1668, opened 2026-07-13
[Spark] Preserve V2 columns in Delta staging operations

timothyw553:
## Problem

Spark carries generated expressions in Array[Column]. UCSingleCatalog must keep those columns intact when it starts a staged Delta operation. Converting through StructType loses the generated expression before Delta receives it.

## Change

The Array[Column] overloads for stageCreate, stageReplace, and stageCreateOrReplace now delegate directly to Delta when the Delta API is enabled. Other paths keep their existing behavior.

This intentionally supports only the deprecated Array[Column] API. It does not add TableInfo shims.

Delta still injects delta.generationExpression in delta-io/delta#7186.

## Tests

UCSingleCatalogStagingTableTest passes on Spark 4.0: 23 tests passed.


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -400,17 +400,32 @@ class UCSingleCatalog
   }
 
   /** Only called for REPLACE TABLE and RTAS */
+  override def stageReplace(
+      ident: Identifier,
+      columns: Array[Column],
```
timothyw553, line 424: notice that this is columns not StructType

timothyw553:
also see https://github.com/delta-io/delta/pull/7062#discussion_r3484232299 for relevant discussion

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: this interface is deprecated spark >= 4.1.0

timothyw553:
@openinx could you please take a look?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: As we are diverging the same `stageReplace` interface into two different implementations, so I will be very cautious to see those divesified code. 

`Only called for REPLACE TABLE and RTAS` this assumption actually highly depends on the Apache Spark implementation details, and if someday the Apache Spark impl changed, then it will produce the bug from oss uc spark connector side, which is not the result that we expected. 

Because essentially,  this is a public API, which we should only follow the API from Spark, rather than relying on the spark implementation details. 

Can you tell why you want to diverge the same `stageReplace` a bit more ? and why is not possible for us to unify or delegate the `stageReplace` to another one ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -443,6 +456,19 @@ class UCSingleCatalog
   }
 
   /** Only called for CREATE OR REPLACE TABLE ... [AS SELECT] */
+  override def stageCreateOrReplace(
```
openinx, line 459: same question,  let's not try to diverge the same API implementation, which is very easy for us to create a new bug and introduce maintenance problem.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
openinx: I took a look at the spark code, https://github.com/apache/spark/blob/cabbcabb4f4853f079a18377b7a0f311098a847a/sql/catalyst/src/main/java/org/apache/spark/sql/connector/catalog/StagingTableCatalog.java#L142,  it will use the `CatalogV2Util.v2ColumnsToStructType` to do the `Array[Collumn]` -> `StructType` conversion with keeping the generated columns in the field. 

so why the default implementation from `StagingTableCatalog` cannot peserve the generated props ?

timothyw553: actually I don't think `CatalogV2Util.v2ColumnsToStructType` preserves the generated columns in the field. I do see it on Spark **master** yes -- in fact it was added only on July 6 https://github.com/apache/spark/commit/4553429c621044bb30d20b901f61bed8b2089817
```java
  private def v2ColumnToStructField(col: Column): StructField = {
    val metadata = Option(col.metadataInJSON()).map(Metadata.fromJson).getOrElse(Metadata.empty)
    var f = StructField(col.name(), col.dataType(), col.nullable(), metadata)
    Option(col.comment()).foreach { comment =>
      f = f.withComment(comment)
    }
    Option(col.defaultValue()).foreach { default =>
      f = encodeDefaultValue(default, f)
    }
    Option(col.id()).foreach { id =>
      f = f.withId(id)
    }
    Option(col.generationExpression()).foreach { genExpr =>
      f = f.copy(metadata = new MetadataBuilder()
        .withMetadata(f.metadata)
        .putString(GeneratedColumn.GENERATION_EXPRESSION_METADATA_KEY, genExpr)
        .build())
    }
    f
  }
```

but this `.putString(GeneratedColumn.GENERATION_EXPRESSION_METADATA_KEY, genExpr)` is actually not in spark 4.0, 4.1, or 4.2.

timothyw553: see here in `branch-4.2` https://github.com/apache/spark/blob/branch-4.2/sql/catalyst/src/main/scala/org/apache/spark/sql/connector/catalog/CatalogV2Util.scala#L662

PR #1669, opened 2026-07-13
[Spark] Extract create property preparation

timothyw553:
## Summary

UCSingleCatalog prepares table properties inside the Column[] create path. Spark 4.1 and 4.2 also need this logic from their TableInfo paths.

This PR extracts property preparation into a shared helper. Existing Column[] behavior does not change.

This is the base PR for #1668.

## Tests

Covered by the TableInfo tests in #1668.


timothyw553:
Closing because we decided to support only the Array[Column] path. The shared property helper is no longer needed.

PR #1670, opened 2026-07-15
[server] Add temporary credential URL

foss-contributor:
## Summary
Add the normalized URL to temporary credentials and use it for Delta credential prefixes. Matches the URL location field currently used by the Delta REST API.

## Tests
Updated existing SDK tests to verify URL field. Also added to API test.

### DIFF
--- a/api/Models/TemporaryCredentials.md
+++ b/api/Models/TemporaryCredentials.md
@@ -7,6 +7,7 @@
 | **azure\_user\_delegation\_sas** | [**AzureUserDelegationSAS**](AzureUserDelegationSAS.md) |  | [optional] [default to null] |
 | **gcp\_oauth\_token** | [**GcpOauthToken**](GcpOauthToken.md) |  | [optional] [default to null] |
 | **expiration\_time** | **Long** | Server time when the credential will expire, in epoch milliseconds. The API client is advised to cache the credential given this expiration time.  | [optional] [default to null] |
+| **url** | **String** | The normalized URL of the storage path the temporary credential was generated for. | [optional] [default to null] |
 
 [[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
 
--- a/api/all.yaml
+++ b/api/all.yaml
@@ -2657,6 +2657,9 @@ components:
             The API client is advised to cache the credential given this expiration time.
           type: integer
           format: int64
+        url:
+          description: The normalized URL of the storage path the temporary credential was generated for.
+          type: string
     PathOperation:
       enum:
         - UNKNOWN_PATH_OPERATION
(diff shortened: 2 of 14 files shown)


api/Models/TemporaryCredentials.md
```diff
@@ -7,6 +7,7 @@
 | **azure\_user\_delegation\_sas** | [**AzureUserDelegationSAS**](AzureUserDelegationSAS.md) |  | [optional] [default to null] |
 | **gcp\_oauth\_token** | [**GcpOauthToken**](GcpOauthToken.md) |  | [optional] [default to null] |
 | **expiration\_time** | **Long** | Server time when the credential will expire, in epoch milliseconds. The API client is advised to cache the credential given this expiration time.  | [optional] [default to null] |
+| **url** | **String** | The normalized URL of the storage path the temporary credential was generated for. | [optional] [default to null] |
```
tdas, line 10: dont we always have a url?

yili-db (approved):

tdas (commented):
`server/src/main/java/io/unitycatalog/server/service/delta/DeltaStagingTableMapper.java:48`
[AI-assisted]

createStagingTable vends from NormalizedURL.from(staging.getStagingLocation()), so prefix is now the normalized url. but .location(info.getStagingLocation()) is still the raw string. before, prefix == location (both raw). are staging locations always already normalized? if not, location and prefix now disagree — does the delta client still match the credential to the path?

if a staging location ever carries a trailing slash / uppercase host / etc., the response hands the client a location and a credential prefix in different forms and credential matching silently breaks — the old code was byte-identical by construction.

api/Models/TemporaryCredentials.md
tdas: i thought about it i think it's okay.

tdas (approved):

foss-contributor:
> `server/src/main/java/io/unitycatalog/server/service/delta/DeltaStagingTableMapper.java:48` [AI-assisted]
> 
> createStagingTable vends from NormalizedURL.from(staging.getStagingLocation()), so prefix is now the normalized url. but .location(info.getStagingLocation()) is still the raw string. before, prefix == location (both raw). are staging locations always already normalized? if not, location and prefix now disagree — does the delta client still match the credential to the path?
> 
> if a staging location ever carries a trailing slash / uppercase host / etc., the response hands the client a location and a credential prefix in different forms and credential matching silently breaks — the old code was byte-identical by construction.

As part of https://github.com/unitycatalog/unitycatalog/issues/1694, this will be covered. The credential selection is done entirely within the `location` field within the credential list., not the table location itself. See `DeltaStagingTableResponse.storage-credentials`.

Before the new changes, we might run into issues where if the vended credential location does not match the table location, an error will be thrown. However, with the new changes, if there is only one credential, no location matching is done, eliminating this problem. If there is more than one credential, matching is done at access time to ensure accurately using the correct credential based on the actual requested URI not the table's recorded storage location.

PR #1677, opened 2026-07-16
Default provider-less managed CREATE TABLE to spark.sql.sources.default

s-maling-celonis:
## What
A managed `CREATE TABLE` with no `USING` clause was rejected by `UCSingleCatalog` as a non-Delta managed table, even when the session had `spark.sql.sources.default=delta`. A named V2 catalog receives no `provider` property when `USING` is omitted (unlike Spark's built-in `V2SessionCatalog`), so `hasDeltaProvider` returned false.

## Change
- `hasDeltaProvider` falls back to `SQLConf.get.defaultDataSourceName` (reads `spark.sql.sources.default`) when no explicit `provider` is set.
- `managedDeltaCreatePropsForDelegate` materializes `provider=delta` before delegating, so the Delta catalog and `UCProxy.requireProviderSpecified` see an explicit provider.

## Tests
`DeltaManagedTableReadWriteTest`:
- default=delta + `CREATE TABLE ...(name STRING)` (no USING) → creates a managed Delta table (insert/read-back).
- default=parquet → still rejected as non-Delta (no regression).

## Notes
No behavior change for explicit `USING <x>` or for Spark's out-of-the-box `parquet` default.


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -595,10 +607,16 @@ object UCSingleCatalog {
...
+   * Returns true when the effective provider is `delta` -- i.e. `USING delta`, or no `USING`
+   * clause while `spark.sql.sources.default` resolves to delta. Unlike Spark's built-in
+   * `V2SessionCatalog`, a named V2 catalog receives no `provider` property when `USING` is
+   * omitted, so we apply the same default-source fallback Spark itself uses.
+   */
   def hasDeltaProvider(properties: util.Map[String, String]): Boolean =
     Option(properties.get(TableCatalog.PROP_PROVIDER))
-      .exists(_.equalsIgnoreCase("delta"))
+      .getOrElse(SQLConf.get.defaultDataSourceName)
+      .equalsIgnoreCase("delta")
```
timothyw553: I noticed you added tests for `createTable`, but can you please add tests for `stageCreate`, `stageReplace`, and `stageCreateOrReplace`.

so, please add integration tests for provider-less `CREATE OR REPLACE TABLE t(...)` and CTAS with `default=delta`, and ensure the REPLACE path also materializes `provider=delta` -- prove they round-trip, don't just avoid throwing.

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaManagedTableReadWriteTest.java
```diff
@@ -111,6 +111,53 @@ public void testCreateManagedTableErrors() {
...
+    // catalog), the deterministic path this change targets.
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG, CATALOG_NAME);
+    ensureSparkCatalogSchemaExists();
+    String fullTableName = CATALOG_NAME + "." + SCHEMA_NAME + "." + DELTA_TABLE + "_defaultsrc";
+    String originalDefault = session.conf().get("spark.sql.sources.default", "parquet");
+    session.conf().set("spark.sql.sources.default", "delta");
+    try {
+      sql("CREATE TABLE %s(name STRING)", fullTableName); // no USING clause
+      sql("INSERT INTO %s VALUES ('a')", fullTableName);
+      assertThat(sql("SELECT name FROM %s", fullTableName)).hasSize(1);
```
timothyw553: hmm, this assertion of .hasSize(1) will pass for _any_ table creation. there's no assertion that it's Delta OR managed. perhaps add additional assertions such as:
```java
  TableInfo info = loadTableInfo(fullTableName);
  assertThat(info.getTableType()).isEqualTo(TableType.MANAGED);
  assertThat(info.getDataSourceFormat()).isEqualTo(DataSourceFormat.DELTA);
  assertManagedTableHasUcProperties(fullTableName, null);
```

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: another thing: I wonder if we should just keep this method pure and do something like `hasDeltaProvider(properties, defaultProvider)`? I have a small worry that a query-executing thread using`SQLConf.get` may fall back to a fresh/default conf where `defaultDataSourceName` returns `parquet`

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -146,7 +147,18 @@ class UCSingleCatalog
...
-    val validated = validateAndDefaultManagedDeltaCreateProperties(properties)
+    // `hasDeltaProvider` may resolve to Delta via `spark.sql.sources.default` when no `USING`
+    // clause was given, in which case the map has no `provider` key. Materialize it so the Delta
+    // delegate -- and `UCProxy.createTable`'s `requireProviderSpecified` check -- see an explicit
+    // provider, exactly as an explicit `USING delta` would produce.
+    val withProvider = if (properties.containsKey(TableCatalog.PROP_PROVIDER)) {
+      properties
+    } else {
+      val augmented = new util.HashMap[String, String](properties)
+      augmented.put(TableCatalog.PROP_PROVIDER, "delta")
```
timothyw553: hmm this is hardcoded to "delta". and below we look at the SQLConf and see if it's "delta". I wonder if it would be better to resolve it once and use that in both places?

imperfect but something like
```scala
  /** The effective provider: the explicit one, else the session default source. */
  def effectiveProvider(properties: util.Map[String, String]): String =
    Option(properties.get(TableCatalog.PROP_PROVIDER))
      .getOrElse(SQLConf.get.defaultDataSourceName)

  def hasDeltaProvider(properties: util.Map[String, String]): Boolean =
    effectiveProvider(properties).equalsIgnoreCase("delta")

  // Then Spot 2 writes that resolved value rather than a literal:
  augmented.put(TableCatalog.PROP_PROVIDER, UCSingleCatalog.effectiveProvider(properties))
```

timothyw553 (commented):
```java
  override def createTable(ident: Identifier, schema: StructType, partitions: Array[Transform], properties: util.Map[String, String]): Table = {
    UCSingleCatalog.requireProviderSpecified("CREATE TABLE", properties)
```

can you also please check if this method is reachable and should apply the same fallback?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
timothyw553: or maybe just a simple comment saying these two should be consistent

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: Done in e12de2d8 — added integration tests for provider-less CTAS (`testCtasManagedTableDefaultsToDeltaWhenDefaultSourceIsDelta`), CREATE OR REPLACE (`testCreateOrReplaceManagedTableDefaultsToDeltaWhenDefaultSourceIsDelta`), and REPLACE (`testReplaceManagedTableDefaultsToDeltaWhenDefaultSourceIsDelta`), each asserting MANAGED + DELTA + UC properties round-trip via `loadTableInfo`.

s-maling-celonis: Done in e12de2d8 — `hasDeltaProvider(properties, defaultProvider)` is now pure; `sessionDefaultProvider` is read once per catalog operation and passed through at each call site (no `SQLConf.get` inside the helper).

connectors/spark/src/test/java/io/unitycatalog/spark/DeltaManagedTableReadWriteTest.java
s-maling-celonis: Done in e12de2d8 — `testCreateManagedTableDefaultsToDeltaWhenDefaultSourceIsDelta` now asserts `TableType.MANAGED`, `DataSourceFormat.DELTA`, and UC properties via `assertManagedDeltaTableMetadata`; the new staging-path tests use the same helper.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: Done in e12de2d8 — added `effectiveProvider` / `withMaterializedProvider`; materialization uses the same resolved value as `hasDeltaProvider` (no hardcoded `"delta"`).

s-maling-celonis: Addressed by the same refactor — gate and materialization both go through `effectiveProvider(properties, sessionDefaultProvider)`.

s-maling-celonis:
Re: **UCProxy.createTable reachability** (review comment Jul 24) — investigated, no UCProxy change needed for this PR.

**Reachability:** SQL table DDL always enters via outer `UCSingleCatalog.createTable(ident, columns: Array[Column], partitions, properties)` / `stageCreate*`, which materializes `provider` via `withMaterializedProvider` before delegating to Delta/UCProxy. `UCProxy.createTable(ident, schema: StructType, partitions, properties)` is the finalize step that registers the table with UC REST and intentionally keeps `requireProviderSpecified` — provider-less props are not expected there.

**Not live paths:** outer `UCSingleCatalog.createTable(..., StructType, ...)` throws `AssertionError` ("deprecated … should not be called"). Direct `UCProxy.createTable` calls without `provider` fail as intended (`UCViewUnsupportedSuite.testCreateTableIgnoresViewTableTypeAndStillRequiresProvider`).

**Evidence:** integration tests for provider-less CREATE / CTAS / CREATE OR REPLACE / REPLACE with `spark.sql.sources.default=delta` all round-trip MANAGED + DELTA + UC metadata — if `UCProxy` saw a map without `provider`, `requireProviderSpecified` would throw before UC registration.

Default-source resolution belongs in outer `UCSingleCatalog` (Steps 1–3); UCProxy remains downstream and assumes callers supply an explicit `provider` + `location`.

s-maling-celonis:
@timothyw553 Any update on this PR? Do you have any more requested changes? Or is another reviewer required?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -140,8 +141,13 @@ class UCSingleCatalog
    *     Delta-side `buildReplaceProps` after loading; we do not pre-check it here.
    *   - Non-Delta REPLACE stays on the legacy path; its rejection error is unchanged.
    */
```
openinx, line 172: Are those javadoc actually related to the `sessionDefaultProvider` ?  or should we move it the following `shouldDelegateReplaceToDeltaApi` method ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -140,8 +141,13 @@ class UCSingleCatalog
...
+  /** Session default format (`spark.sql.sources.default`); read once per catalog operation. */
+  private def sessionDefaultProvider: String =
+    SQLConf.get.getConf(SQLConf.DEFAULT_DATA_SOURCE_NAME)
+
   private def shouldDelegateReplaceToDeltaApi(properties: util.Map[String, String]): Boolean = {
-    shouldUseDeltaAPI && UCSingleCatalog.hasDeltaProvider(properties)
+    val defaultProvider = sessionDefaultProvider
+    shouldUseDeltaAPI && UCSingleCatalog.hasDeltaProvider(properties, defaultProvider)
```
openinx: nit: could we just use one line `shouldUseDeltaAPI && UCSingleCatalog.hasDeltaProvider(properties, defaultProvider)` pls, without introducing an extra local variable named `defaultProvider` ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -630,10 +642,41 @@ object UCSingleCatalog {
     !hasExternalClause && !hasLocationClause && !isPathTable
   }
 
-  /** Returns true when `USING <format>` is `delta`. */
-  def hasDeltaProvider(properties: util.Map[String, String]): Boolean =
-    Option(properties.get(TableCatalog.PROP_PROVIDER))
```
openinx, line 635: I think we can use the following here

```
Option(properties.get(TableCatalog.PROP_PROVIDER)).getOrElse(sessionDefaultProvider).exists(..)
```

with this, we will won't to introduce extra `effectiveProvider` methods and adding the extra `defaultProvider` parameter in the method.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -154,8 +160,12 @@ class UCSingleCatalog
    */
   private def managedDeltaCreatePropsForDelegate(
       ident: Identifier,
-      properties: util.Map[String, String]): util.Map[String, String] = {
-    val validated = validateAndDefaultManagedDeltaCreateProperties(properties)
+      properties: util.Map[String, String],
+      defaultProvider: String): util.Map[String, String] = {
```
openinx: we don't even need to introduce this extra `defaultProvider` parameter here, since we can always get it from the current spark session via the `sessionDefaultProvider`.  so I'd prefer to remove it directly from it, and then callers no need to pass it .

openinx (changes requested):

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: Done in 5df09ac7 — moved `sessionDefaultProvider` javadoc onto the field itself, ahead of `hasDeltaProvider` / `withMaterializedProvider`, and kept the REPLACE-gate javadoc on `shouldDelegateReplaceToDeltaApi`.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: Done in 5df09ac7 — `shouldDelegateReplaceToDeltaApi` is now a one-liner: `shouldUseDeltaAPI && hasDeltaProvider(properties)`.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: Done in 5df09ac7 — removed companion-object `effectiveProvider` / `hasDeltaProvider(properties, defaultProvider)` / `withMaterializedProvider(properties, defaultProvider)`. `hasDeltaProvider` and `withMaterializedProvider` are now private instance methods that inline `Option(...).getOrElse(sessionDefaultProvider)`.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: Done in 5df09ac7 — dropped `defaultProvider` from `managedDeltaCreatePropsForDelegate` and all call sites; it reads `sessionDefaultProvider` via the instance helpers above.

s-maling-celonis:
Follow-up on the `hasDeltaProvider` purity thread: per @openinx's Aug 6 feedback, the companion-object `hasDeltaProvider(properties, defaultProvider)` helpers are gone. Resolution now lives on the catalog instance via `sessionDefaultProvider` → `hasDeltaProvider(properties)` / `withMaterializedProvider(properties)`. Call sites no longer thread a `defaultProvider` argument.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -132,6 +133,35 @@ class UCSingleCatalog
...
+  /** Ensures `provider` is present using the same resolution as {@link #hasDeltaProvider}. */
+  private def withMaterializedProvider(
+      properties: util.Map[String, String]): util.Map[String, String] = {
+    if (properties.containsKey(TableCatalog.PROP_PROVIDER)) {
+      properties
+    } else {
+      val augmented = new util.HashMap[String, String](properties)
+      augmented.put(
+        TableCatalog.PROP_PROVIDER,
+        Option(properties.get(TableCatalog.PROP_PROVIDER)).getOrElse(sessionDefaultProvider))
```
openinx, line 160: nit: this can be simplified as `augmented.put(TableCatalog.PROP_PROVIDER, sessionDefaultProvider)` directly,   as we are pretty sure the `properties` has no `TableCatalog.PROP_PROVIDER` config key in this  branch.

openinx (approved):
This is looking good to me now,  just left one nit comment which won't block the code merge,  thanks @s-maling-celonis for the great contribution !

PR #1680, opened 2026-07-17
[spark] Avoid duplicate UC getTable call during view relation fallback

timothyw553:
## Summary

This is stacked on #1675's commit which will be removed prior to merging. Please just review commit https://github.com/unitycatalog/unitycatalog/pull/1680/commits/469e3ac10616954a8dc6f561f23c9fcbc46c47ea

It adds one optimization commit on top of the Spark 4.2 upgrade. When Spark 4.2 loadRelation first tries delegate.loadTable and that path discovers the identifier is a view, UCProxy.loadTable temporarily saves that UC row. The fallback loadView then reuses the saved row instead of calling UC getTable a second time.

## Testing

- build/sbt -J-Xmx3G -DsparkVersion=4.2 -DskipDeltaSpark=true "spark/Compile/compile"
- Focused local invocation of UCViewProxySuite.testUCSingleCatalogLoadRelationReusesViewFromDelegateLoadTable: passed

Depends on #1675.

### DIFF
--- a/connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCSingleCatalogViewSupport.scala
+++ b/connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCSingleCatalogViewSupport.scala
@@ -1,6 +1,5 @@
 package io.unitycatalog.spark
 
-import org.apache.spark.sql.catalyst.analysis.{NoSuchTableException, NoSuchViewException}
 import org.apache.spark.sql.connector.catalog.{
   Identifier,
   Relation,
@@ -41,24 +40,21 @@ trait UCSingleCatalogViewSupport extends RelationCatalog { self: UCSingleCatalog
     ucProxy.renameView(oldIdent, newIdent)
 
   /**
-   * Keep normal table loading on the delegate path. If the delegate rejects the identifier as
-   * a table, try the UC view path with any view row from that load memoized.
+   * Keep normal table loading on the delegate path. If the UC table-only path finds a view, reuse
+   * the view metadata carried by its `NoSuchTableException` instead of issuing another UC lookup.
    */
   override def loadRelation(ident: Identifier): Relation = {
-    ucProxy.memoizeViewOnLoadTable(ident)
     try {
-      try {
-        delegate.loadTable(ident)
-      } catch {
-        case tableNotFound: NoSuchTableException =>
-          try {
-            ucProxy.loadView(ident)
-          } catch {
-            case _: NoSuchViewException => throw tableNotFound
-          }
-      }
-    } finally {
-      ucProxy.clearViewMemo()
+      delegate.loadTable(ident)
+    } catch {
+      case viewFound: ViewFoundDuringTableLoadException =>
+        val t = viewFound.tableInfo
+        if (UCViewTypes.isViewCommandsSupportedTableType(t.getTableType)) {
+          ucProxy.toView(t)
+        } else {
+          throw new UnsupportedOperationException(
+            s"Loading a ${t.getTableType} view is not supported yet")
+        }
     }
   }
 }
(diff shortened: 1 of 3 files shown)


connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCSingleCatalogViewSupport.scala
```diff
@@ -42,18 +42,23 @@ trait UCSingleCatalogViewSupport extends RelationCatalog { self: UCSingleCatalog
...
+      try {
+        delegate.loadTable(ident)
+      } catch {
+        case tableNotFound: NoSuchTableException =>
+          try {
+            ucProxy.loadView(ident)
+          } catch {
+            case _: NoSuchViewException => throw tableNotFound
+          }
+      }
```
openinx: why do we need the double nested `try...catch..`,  could we use the one `try...catch...`, and add the `finally` directly to it ?

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCSingleCatalogViewSupport.scala
```diff
@@ -42,18 +42,23 @@ trait UCSingleCatalogViewSupport extends RelationCatalog { self: UCSingleCatalog
 
   /**
    * Keep normal table loading on the delegate path. If the delegate rejects the identifier as
-   * a table, try the UC view path.
+   * a table, try the UC view path with any view row from that load memoized.
    */
   override def loadRelation(ident: Identifier): Relation = {
+    ucProxy.memoizeViewOnLoadTable(ident)
```
openinx: I won't suggest to use the ThreadLocal to catch this loaded table in one method, and without understandable context then use it in another method,  that's actually making it hard to track and over-complicated. 

If we really want to reuse the results from the `delegate.loadTable` via one RPC,  then why not just call the `getUCTableLike` directly here, and use that `Option[UCTableInfo]` to decide whether to reuse this loaded table info or not?   won't that simplify the whole code and make it more readable ?

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCSingleCatalogViewSupport.scala
timothyw553: Agreed. I removed the nested `try/catch` and the `finally` cleanup. `loadRelation` now uses a single `try/catch` around `delegate.loadTable`.

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCSingleCatalogViewSupport.scala
timothyw553: Agreed that the `ThreadLocal` made the metadata flow difficult to follow, so I removed it. I considered calling `getUCTableLike` directly from `loadRelation`, but that would add a preliminary UC lookup before `delegate.loadTable`, reintroducing the extra `getTable` RPC on the managed Delta table path.

Instead, when `UCProxy.loadTable` has already fetched a view row, it throws a private `NoSuchTableException` subtype carrying that `UCTableInfo`. The Delta delegate rethrows the same exception, and `loadRelation` converts the attached metadata directly into a view. This keeps table resolution delegate-first and uses exactly one UC lookup for both tables and views.

openinx (approved):

connectors/spark/src/main/scala-shims/spark-4.2/io/unitycatalog/spark/UCSingleCatalogViewSupport.scala
yili-db: https://github.com/unitycatalog/unitycatalog/pull/1675#discussion_r3599692745
Either ThreadLocal or a customized exception could carry the cached result.

PR #1682, opened 2026-07-18
perf(hadoop): per-key locking for the credential global cache

### LINKED ISSUE #1651: Support fine grain lock to improve the concurrency in the credential global cache
Paste the discussion from https://github.com/unitycatalog/unitycatalog/pull/1642#discussion_r3516620608

### From @yili-db

This will lock this global cache for all table loading, when one table is fetching cred from UC via `factory.create()`.
In general this `synchronized(cache)` should not exist anymore because `BoundedKeyedCache` is already lock aware. It has a `getOrLoad` that locks on only one key that can almost do this, **except** that it doesn't have a concept of expiration. If you add expiration control to `BoundedKeyedCache` it will be the most optimal locking for this cache.

Of course I see that existing code is already doing `synchronized (globalCache) {` so your code is just doing the same. Above optimization could be done as a separate follow up. Since now we are trying to improve the performance it may be worthy.

### From @openinx

if there are multiple accessor which are accessing the same cred based on the `CredId`. there are the cases:

1.  if cache hit and cred is still valid.  then  the lock duration will be very short.  no much impact to the latency. 
2. if cache hit but cred is expired.  then need to wait the duration to re-access the new credential,  the lock duration maybe 100ms.   but that's exactly the case that we want to intentionally design,  since we just want only one accessor can renew the cred, and others will reuse it. 
3.  if cache miss,  same as the case.2 

I think your point is,  we should make it to be a more fine-grain lock, so that one `credId-A` won't affect another `credId-B` access, right ? 

Yes,  I agree that can be a follow-up optimization, which may need some refactor for the `BoundedKeyedCache`.

tiennguyen-onehouse:
## Summary

Implements the follow-up agreed in #1651 (spawned from the review discussion on #1642): move the credential global cache from one JVM-wide lock to per-key locking, so one scope's renewal RPC no longer blocks every other scope's credential access.

- `CredentialCache.access()` previously wrapped the whole lookup-or-renew flow — including the renewal RPC to the UC server — in `synchronized (cache)`. While one scope renewed, all other scopes (even with valid cached credentials) blocked on that monitor.
- Per @yili-db's suggestion in the issue, `BoundedKeyedCache` (not Caffeine) is the right home: it already has per-key single-flight locking via `IdLockMap` in `getOrLoad` — it just had no notion of a cached-but-stale value.
- Freshness is modelled as a **cache-level policy, not a per-call argument** (per @openinx's review): `BoundedKeyedCache` takes an `isFresh` predicate at construction, the way Guava/Caffeine configure expiry on the builder and keep the loader separate. (The analogy is `expireAfterWrite` plus a loading `get`, not Caffeine's `refreshAfterWrite`, which is asynchronous and serves the stale value meanwhile.)
  - `getOrLoad(key, loader)` keeps its existing single-argument shape — no per-call predicate, no fused `BiFunction`.
  - A cached value the policy accepts is returned **without acquiring the key lock** (the map monitor is still taken briefly, so it is not fully lock-free). That fast path matters: a valid cached credential is the hot path on every storage operation. (This is why `compute(key, remappingFunction)` wasn't viable — the cache would have to *run* the function to learn whether a reload is needed, and running it lock-free would let two stale callers both fire the renewal RPC. The JDK draws the same line: `computeIfAbsent` can keep a lock-free hit path because "absent" is checkable without the function; `compute` cannot.)
  - The two-argument constructor delegates with `alwaysFresh()`, so `CredScopedFileSystem`'s load-once FS cache is unchanged. (The single-argument constructor had no production caller and is removed, so the three-argument one is the only way to build a cache with a policy.)
- `CredentialCache` declares its policy once, at construction, and `access()` becomes a one-liner:

```java
this.cache =
    new BoundedKeyedCache<>(
        maxSize, BoundedKeyedCache.noOpListener(), cached -> !cached.readyToRenew());
...
return cache.getOrLoad(key, factory::create).credential();
```

Renewal semantics are preserved exactly — same-scope accessors stay single-flight (one renews, waiters briefly block then reuse, unless the renewed credential is itself already due for renewal) — while different scopes never contend. The global `cacheLock` continues to guard only the light map operations (get/put/LRU eviction).

Both cache instances benefit: `GenericCredentialProvider.globalCache` and `CredPropsUtil.initialCredCache`.

Fixes #1651.

## How we tested

New `CredentialCacheTest` (5 tests) plus 8 new `BoundedKeyedCacheTest` cases, following the existing latch-based patterns in that file.

TDD verification — the headline test was written first and, re-run on the current rebase with `access()` restored to the global lock, fails by timing out exactly as the issue describes (scope B's access queues behind scope A's in-flight renewal):

```
[error] Test io.unitycatalog.hadoop.internal.auth.CredentialCacheTest.renewalOfOneScopeDoesNotBlockOtherScopes
        failed: java.util.concurrent.TimeoutException: null, took 7.118s
[info] Test run finished: 1 failed, 0 ignored, 3 total
```

With the fix, the whole hadoop module passes:

```
build/sbt "hadoop/test"
[info] Passed: Total 1477, Failed 0, Errors 0, Passed 1477
```

The lock-free hit path is pinned rather than assumed: `freshCachedValueIsReturnedWhileTheSameKeyIsLoading` parks a load on a key (holding that key's lock), publishes a fresh value, and asserts a concurrent `getOrLoad` on the same key returns it within 2s without invoking the loader. Deleting the pre-lock check makes it fail with `TimeoutException` in 2.0s.

Coverage, by file:

`CredentialCacheTest` (5 new) — a valid cached credential skips the factory; an expiring credential is renewed and replaced; a renewal in one scope does not block access in another scope; a stampede across scopes costs exactly one vend per scope and never serves a caller a neighbouring scope's credential; a failed vend caches nothing and leaves the scope usable.

`BoundedKeyedCacheTest` (8 new) — a cached value the policy rejects is reloaded; a value the policy accepts skips the loader; a rejected value is reloaded exactly once across threads (this is the same-key single-flight case you noted belongs here rather than in `CredentialCacheTest`); a policy-accepted value is served while a load on the *same* key is in flight, which is what pins the lock-free hit path; a null eviction listener is rejected; plus the three adversarial cases below.

### Adversarial regression tests

Added per @timothyw553's review. Each one targets a failure mode the per-key lock introduces, and each was verified to **fail** against a deliberately broken implementation rather than merely passing against this one:

| Test | Regression it catches | Verified by mutating |
|---|---|---|
| `loaderFailureDoesNotWedgeWaitersOnTheSameKey` | a loader that throws must not strand the callers queued behind it | dropping the try-with-resources so the key lock is never released → fails with `TimeoutException` |
| `everyCallerLoadsOnceWhenThePolicyRejectsEveryLoadedValue` | a value the policy still rejects must not be retried in a loop (a production hang) | wrapping the load in a *bounded* retry-until-fresh loop → fails on the load count. An unbounded loop, which is the actual hazard, instead fails the 10 s `future.get` and leaves the spinning thread behind, so the mutation was bounded to keep the signal clean |
| `concurrentLoadsHonourMaxSizeAndDisposeEachEvictedValueOnce` | the bound must hold under concurrent loads, and no value may be handed to the eviction listener twice (a double close of a live filesystem) | loosening the eviction bound by one → fails on size and on the evicted set |

The first is the concurrent counterpart of the pre-existing `keyLockReleasedAfterLoaderThrows`, which only covers the sequential case; it is what backs the "a failed renewal RPC cannot wedge a scope" claim below.

An earlier revision also asserted a `keyLockCount()` accessor was zero after each run. Measuring the two regressions it was meant to catch showed it was redundant, so it is gone rather than living in main code for the tests' benefit:

| Regression | Caught by | Needed the accessor |
|---|---|---|
| `IdLockMap` never drops the refcount | `IdLockMapTest`, 4 failures | no |
| `getOrLoad` mishandles the lock (an extra acquire) | 3 cases here, incl. the pre-existing `keyLockReleasedAfterLoaderThrows` | no |

Either the bug is in `IdLockMap`, where its own test already pins the count returning to zero, or it is in this class's use of it, where an unreleased hold deadlocks the next same-key caller. The accessor only made those failures report as an assertion rather than a timeout.

The pre-existing `BoundedKeyedCacheTest` cases are unchanged apart from how they spell the constructor, so they double as the regression check that the default behaviour is unchanged. The two-argument constructor's `alwaysFresh()` default — what `CredScopedFileSystem` relies on — is pinned by `constructorWithoutFreshnessPolicyKeepsEveryCachedValue`.

## Notes

- No new dependencies (the Caffeine alternative discussed in the issue turned out to be unnecessary, per the resolution in the comments).
- `GenericCredentialProvider`'s per-instance `synchronized(this)` blocks are untouched — they guard per-provider state, not the global cache, and are outside this issue's scope.

## Before/after benchmark

Re-measured on this revision with an uncommitted micro-benchmark that uses only the public `CredentialCache.access()` API (identical file run against `origin/main` and this branch; renewal RPC simulated with `Thread.sleep`):

| Scenario | Before (global lock) | After (per-key lock) |
|---|---|---|
| 8 independent scopes renewing concurrently (200 ms RPC each) | **1630 ms** (fully serialized, ~8 × 200 ms) | **205 ms** (parallel, ~1 × 200 ms) |
| Read of a valid cached scope while another scope's 1000 ms renewal is in flight | **950 ms** (blocked behind the renewal) | **0.36 ms** (unaffected) |

## Thread-model verification

Adversarial review of the lock model found no correctness issue; two constraints are now pinned in code/docs:

- **Loader re-entrancy**: loaders must not call back into the cache (potential lock cycle) — documented on `getOrLoad`.
- **Policy purity**: `isFresh` is applied both outside and inside the key lock, so it must be a pure function of the value and must not touch the cache — now documented on `getOrLoad`. In practice it just delegates to the value (`RenewableCredential.readyToRenew()`), so entries created under different renewal policies still behave correctly.
- **Single-flight is not unconditional**: a same-key waiter re-applies the policy and reloads if it rejects the loaded value (e.g. a credential vended with less remaining lifetime than the renewal lead time). Unchanged from the previous global-lock behaviour; the javadoc now says so rather than promising waiters always reuse the load.
- **Lock release on failure**: the per-key lock must release when a loader throws. Already guaranteed and covered on `main` by the pre-existing `keyLockReleasedAfterLoaderThrows` (a follow-up load succeeds instead of deadlocking); routing `access()` through `getOrLoad` inherits it, so a failed renewal RPC cannot wedge a scope.


openinx (commented):
nit: Several commits include `Co-authored-by: Claude ...` trailers. If you squash before merge, please drop those so history matches the repo DCO convention (`Signed-off-by` only).

connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AwsCredCacheConcurrencyITTest.java
```diff
@@ -0,0 +1,271 @@
...
+    sql("SELECT * FROM %s", WARM_TABLE);
+
+    // Advance the shared manual clock past the credential window so every credential vended
+    // during setup is stale and the next access per table must re-vend.
+    slowVendEnabled = true;
+    testClock().sleep(Duration.ofMillis(CRED_WINDOW_MILLIS));
+    ExecutorService executor = Executors.newFixedThreadPool(2);
+    try {
+      // First read of the slow table vends its READ credential; the generator holds that vend
+      // for SLOW_VEND_MILLIS.
```
openinx: nit: This comment still refers to `SLOW_VEND_MILLIS`; the slow vend is latch-driven now (`slowVendStarted` / `releaseSlowVend`). Please update the wording.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/CredentialCacheTest.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.jupiter.api.Test;
+
+class CredentialCacheTest {
```
openinx, line 19: nit: The PR description says four tests in this class; there are three here (same-scope single-flight is covered in `BoundedKeyedCacheTest`). Please align the PR description.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -32,13 +33,27 @@ public V getIfPresent(K key) {
...
+  /**
+   * Returns the cached value for {@code key} when present and accepted by {@code isValid},
+   * otherwise loads a fresh value and caches it. A cached-but-rejected value (e.g. an expiring
+   * credential) is treated as a miss. Loads are single-flight per key: the thread holding the key
+   * lock loads while same-key waiters block for the duration of the load and reuse its result;
+   * threads on different keys never block each other. A freshly loaded value is cached and returned
+   * even if {@code isValid} would reject it -- the loader's output is trusted. Loaders must not
+   * call back into this cache: a loader that loads other keys can form a lock cycle and deadlock.
+   */
+  public <E extends Exception> V getOrLoad(
```
openinx: Hi @tiennguyen-onehouse ,  can you help to adopt the similar interface that ConcurrentHashMap already commonly used to define the interface ? 

```java
    /**
     * Attempts to compute a mapping for the specified key and its
     * current mapped value (or {@code null} if there is no current
     * mapping). The entire method invocation is performed atomically.
     * The supplied function is invoked exactly once per invocation of
     * this method.  Some attempted update operations on this map by
     * other threads may be blocked while computation is in progress,
     * so the computation should be short and simple.
     *
     * <p>The remapping function must not modify this map during computation.
     *
     * @param key key with which the specified value is to be associated
     * @param remappingFunction the function to compute a value
     * @return the new value associated with the specified key, or null if none
     * @throws NullPointerException if the specified key or remappingFunction
     *         is null
     * @throws IllegalStateException if the computation detectably
     *         attempts a recursive update to this map that would
     *         otherwise never complete
     * @throws RuntimeException or Error if the remappingFunction does so,
     *         in which case the mapping is unchanged
     */
    public V compute(K key,
                     BiFunction<? super K, ? super V, ? extends V> remappingFunction) {
```

In this case,  I will suggest to expose a interface like this : 


```java
  public V get(K key, BiFunction<? super K, ? super V> remappingFunction){

  }
```

with this API, you actually can implement any kind of predicate inside your `remappingFunction`.  but we need to be very careful to write the javadoc for this API.  WDYT ?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AwsCredCacheConcurrencyITTest.java
```diff
@@ -0,0 +1,271 @@
...
+ *   <li>concurrent queries on the same table vend its credential exactly once (single flight).
+ * </ul>
+ *
+ * <p>Slowness is injected into the server-side credential generator (same JVM), gated on a static
+ * flag and the table location, so setup traffic is unaffected. The slow vend parks on a latch the
+ * test releases only after the fast-table query completes, so the pass/fail signal is an ordering
+ * fact rather than a timing margin; a bounded fallback await keeps the server worker from being
+ * parked indefinitely.
+ */
+public class AwsCredCacheConcurrencyITTest extends BaseCRUDTest {
```
openinx: This e2e test actually is not required, if we can add the enough coverage for the `BoundedKeyedCache`, since those two components are fully decoupled.

openinx: And we don't have to use the `AwsCredCacheConcurrencyITTest` to cover the `BoundedKeyedCache` change, that's not worthy.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
tiennguyen-onehouse: @openinx thanks — I agree the per-call `isValid` predicate reads as a bespoke shape, and I went down the `compute(key, remappingFunction)` path you suggested. It works, but I hit one snag worth surfacing before I commit to a direction.

Folding the validity check into a single `(key, current) -> value` remapping function means the cache has to *run* the function to decide whether a reload is needed, and running it can only happen under the per-key lock (running it lock-free would let two concurrent stale callers both fire the renewal RPC and lose single-flight). So every access — including the common "cached credential still valid, no RPC" case, which is the hot path on every storage op — would take the per-key lock. The JDK draws the same line: `computeIfAbsent` keeps a lock-free hit path because "absent" is checkable without the function, while `compute` can't and always locks.

Rather than a per-call predicate *or* losing that fast path, what do you think of following how Guava/Caffeine structure this — **freshness as a cache-level policy, not a per-call argument**? In Caffeine/Guava you configure `refreshAfterWrite` once on the builder and the loader is separate; `get(key)` stays a single clean call and the staleness check lives inside the cache. Our value already self-describes freshness (`RenewableCredential.readyToRenew()`), so it maps directly:

```java
// construction — freshness policy declared once, where the cache is built
new BoundedKeyedCache<>(maxSize, evictionListener, cached -> !cached.readyToRenew());

// a single, familiar getOrLoad(key, loader) — no per-call predicate, no fused BiFunction
public <E extends Exception> V getOrLoad(K key, CheckedSupplier<V, E> loader) throws E {
  V cached = getIfPresent(key);
  if (cached != null && isFresh.test(cached)) {
    return cached;                       // lock-free fast path
  }
  try (IdLockMap<K>.IdLock ignored = keyLocks.acquire(key)) {
    // double-check under the lock, then load + put
  }
}
```

Call sites collapse to the single method you were after:

```java
cache.getOrLoad(credId, factory::create);              // CredentialCache (freshness policy set at ctor)
CACHE.getOrLoad(key, () -> newFileSystem(uri, conf));   // load-once FS cache, unchanged
```

That gives us the single `getOrLoad(key, loader)` shape, keeps the valid read off the lock, and matches the established cache-design precedent rather than a one-off predicate. The freshness knob being a property of the cache (a credential cache renews when `readyToRenew`) also feels conceptually right — it doesn't vary per call, and because the predicate just delegates to the value, entries created with different renewal policies still behave correctly.

If that direction sounds good I'll rework it that way. And as a natural follow-up (not this PR), the fully Caffeine-style version would serve the still-valid credential immediately and renew on a background executor, so even the triggering caller never blocks on the RPC — happy to file that separately. WDYT?

connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AwsCredCacheConcurrencyITTest.java
tiennguyen-onehouse: @openinx agreed, and done in `194baf78`. The cross-scope non-blocking behavior is a property of `CredentialCache`/`BoundedKeyedCache`, and it's already covered deterministically at the unit level by `CredentialCacheTest.renewalOfOneScopeDoesNotBlockOtherScopes` (two mock scopes, a latch-blocked factory standing in for the renewal RPC, asserting the fast scope isn't blocked by the slow one's in-flight renewal). The Spark provider only delegates to `cache.access()`, so the e2e was re-verifying that same property through a much heavier stack. Removed `AwsCredCacheConcurrencyITTest` — which also drops the stale `SLOW_VEND_MILLIS` comment you flagged above. Thanks!

openinx:
Hi @tiennguyen-onehouse , thanks for the update,  could you pls update this PR again to resolve the code conflicts ?  Thank you.

tiennguyen-onehouse:
@openinx rebased on latest `main` and conflicts resolved — thanks for the ping.

The conflict was in `CredentialCache.access()`, which #1695 and #1701 had genericized to `CredentialCache<K, T>` / `RenewableCredential<T>` in the meantime. Resolved onto the new signature, so `access()` is now:

```java
return cache.getOrLoad(key, cached -> !cached.readyToRenew(), factory::create).credential();
```

`CredentialCacheTest` needed the same treatment (generic type args, and `new AwsCredential(...)` now that `GenericCredential.forAws` is gone). I also squashed the two commits into one: the second only deleted a test the first added, so the intermediate commit no longer built on current `main`. No `Co-authored-by` trailers, per your earlier note — `Signed-off-by` only.

Re-verified after the rebase: `hadoop/test` is 1467/1467 green, and reverting just `access()` to the global lock still makes `renewalOfOneScopeDoesNotBlockOtherScopes` time out at ~7s, so the test is still pinning the intended behavior.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
tiennguyen-onehouse: Implemented this way in the latest push (`08fce4c9`), so the interface is now the single `getOrLoad(key, loader)` you asked for — the per-call predicate is gone.

Freshness moved to the constructor:

```java
public BoundedKeyedCache(int maxSize, Consumer<V> evictionListener, Predicate<V> isFresh)
```

`null` (both existing constructors delegate with it) keeps every cached value usable until eviction, so `CredScopedFileSystem`'s load-once cache is untouched. `CredentialCache` declares its policy once at construction:

```java
this.cache = new BoundedKeyedCache<>(maxSize, null, cached -> !cached.readyToRenew());
...
return cache.getOrLoad(key, factory::create).credential();
```

The hit path stays lock-free, which is the part I was worried about losing with `compute`. That is now pinned by a test rather than left as a claim: `freshCachedValueIsReturnedWhileTheSameKeyIsLoading` parks a load on key `k` (holding `k`'s lock), publishes a fresh value via `put`, and asserts a concurrent `getOrLoad("k", ...)` returns it within 2s without invoking the loader. Removing the pre-lock check makes it fail with `TimeoutException` in 2.0s, so it genuinely guards the fast path.

`hadoop/test` is 1468/1468 green. Also rebased onto `main` and resolved the conflicts from #1695/#1701 (`CredentialCache` is generic now) — squashed to one commit, `Signed-off-by` only.

Still happy to file the background-refresh follow-up (serve the stale-but-valid credential immediately, renew off-thread) as a separate issue once this lands.

timothyw553 (approved):
LGTM. I ran some adversarial regression tests - everything looks good. though, maybe could we add some too in this PR?

tiennguyen-onehouse:
Thanks @timothyw553! Added five adversarial regression tests in 15618a86, aimed at the failure modes the per-key lock actually introduces rather than the happy path:

- **`loaderFailureDoesNotWedgeWaitersOnTheSameKey`** — eight callers on one key, the first loader throws. Only that caller sees the failure; the rest recover on a single retry. This is the concurrent counterpart of the existing `keyLockReleasedAfterLoaderThrows`, which only covers the sequential case, and it is what backs the "a failed renewal RPC cannot wedge a scope" claim in the description.
- **`everyCallerLoadsOnceWhenThePolicyRejectsEveryLoadedValue`** — a policy that rejects even what the loader just produced (a credential vended with less remaining lifetime than the renewal lead time). Each caller must settle for its own load; a retry-until-fresh loop here would hang a production thread.
- **`concurrentLoadsHonourMaxSizeAndDisposeEachEvictedValueOnce`** — 32 keys into a cache of 4. The bound holds, and every dropped value reaches the eviction listener exactly once. A double dispose would close a filesystem another caller still holds, since that listener is `CredScopedFileSystem`'s closer.
- **`concurrentAccessRenewsEachScopeExactlyOnce`** — 8 scopes × 8 threads: one vend per scope, and every caller gets its own scope's credential rather than a neighbour's.
- **`failedRenewalIsNotCachedAndLeavesTheScopeUsable`** — a failed vend caches nothing, and the retry runs with a timeout so a leaked lock fails the test instead of hanging the build.

Rather than just assert they pass, I mutation-checked the three cache-level ones against deliberately broken implementations — leaking the key lock, retrying the load in a loop, and loosening the eviction bound by one. Each mutation is caught by its intended test (details in the description).

Also rebased onto `main`, since CI here last ran 43 commits back. `hadoop/test` is 1476/1476 green.

tiennguyen-onehouse:
Follow-up in 5e22e2c1 from a review pass over the new tests:

- **A leaked key lock was only detectable as a timeout.** A regression that unlocks but never drops the `IdLockMap` refcount leaves every other assertion green and produces no hang, so nothing in `BoundedKeyedCacheTest` would have caught it. Added a package-private `keyLockCount()` and asserted it is zero at the end of the three concurrent cases; removing the refcount release now fails exactly those three and nothing else in the class.
- Dropped the duplicated null check in the two-argument constructor (the canonical three-argument one already does it) and pinned that a null eviction listener is rejected, since commit 2fde2418 turned the old null sentinel into an NPE and nothing covered it.
- Documented `maxSize` on the three-argument constructor.
- Corrected the description, which still showed the pre-2fde2418 API: it described the policy as optional and passed `null` for the eviction listener in the `CredentialCache` snippet, which would now throw. Also made the mutation descriptions exact and softened the Caffeine analogy, since this is synchronous reload-on-read (`expireAfterWrite` plus a loading `get`) rather than `refreshAfterWrite`, which is asynchronous and serves the stale value meanwhile.

`hadoop/test` is 1477/1477. @TimothyW553 sorry for the churn on your approval — the adversarial tests you asked for needed a push, so it went stale.

s-maling-celonis:
Thanks for this @tiennguyen-onehouse — we hit the same stall independently (concurrent Spark DDL for different tables blocking in `CredentialCache.access` on the driver). Closing https://github.com/unitycatalog/unitycatalog/pull/1876 in favor of this PR; the per-key locking here covers what we needed.

yili-db:
ping @openinx

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -7,38 +7,82 @@
...
 
 public final class BoundedKeyedCache<K, V> {
   private final int maxSize;
   private final Consumer<V> evictionListener;
+  private final Predicate<V> isFresh;
   private final Object cacheLock = new Object();
   private final LinkedHashMap<K, V> cache = new LinkedHashMap<>(16, 0.75f, true);
   private final IdLockMap<K> keyLocks = new IdLockMap<>();
 
   public BoundedKeyedCache(int maxSize) {
```
openinx: nit: As this constructor is visible for testing only,  could we remove this method and make the testing code to call the `BoundedKeyedCache(int maxSize, Consumer<V> evictionListener, Predicate<V> isFresh)` directly ?  that can prevent other to call this method in the prod code.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -7,38 +7,82 @@
...
+   *     #clear}; use {@link #noOpListener()} when evicted values need no disposal.
+   * @param isFresh applied by {@link #getOrLoad} to a cached value; one it rejects is treated as a
+   *     miss and reloaded. Must be a pure function of the value. Use {@link #alwaysFresh()} to keep
+   *     every cached value usable until it is evicted.
+   */
+  public BoundedKeyedCache(int maxSize, Consumer<V> evictionListener, Predicate<V> isFresh) {
     Preconditions.checkArgument(maxSize > 0, "maxSize must be positive, got %s", maxSize);
     this.maxSize = maxSize;
-    this.evictionListener = evictionListener == null ? value -> {} : evictionListener;
+    this.evictionListener = Objects.requireNonNull(evictionListener, "evictionListener");
```
openinx: nit:  I suggest to make this error message to be `evictionListener cannot be null`.  same to the `isFresh`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
```diff
@@ -7,38 +7,82 @@
...
+   * outside and inside the key lock.
+   */
   public <E extends Exception> V getOrLoad(K key, CheckedSupplier<V, E> loader) throws E {
     V cached = getIfPresent(key);
-    if (cached != null) {
+    if (cached != null && isFresh.test(cached)) {
       return cached;
     }
     try (IdLockMap<K>.IdLock ignored = keyLocks.acquire(key)) {
       V lockedCached = getIfPresent(key);
```
openinx, line 81: This `getIfPresent` is still using the global `cacheLock` which is `BoundedKeyedCache` level.   but as the  access time off the in-memory `LinkedHashMap.get` is extremely fast,  so there is no bottleneck for this. 

And before this PR,  we are using the global lock to call `loader.get()` , which is really expensive and time-consuming.  now we degrade the global lock to key level lock, so that's why it resolve the bottleneck. 

that make sense to me.

openinx (approved):
Sorry for the delayed review (I was distracted by few things in the past weeks ) and thanks for the patience. the core changes looks good to me, but left few non-blocking comments. @tiennguyen-onehouse @s-maling-celonis .

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
tiennguyen-onehouse: Done. The single-argument constructor had no production caller (`CredScopedFileSystem` uses the two-argument one, `CredentialCache` the three-argument one), so it is removed and the tests construct the canonical three-argument form directly, via static imports of `noOpListener()`/`alwaysFresh()` to keep the call sites readable.

One knock-on: `constructorWithoutEvictionListenerUsesNoOpListener` was testing the overload that just went away, and it only really re-checked eviction. I repurposed it as `constructorWithoutFreshnessPolicyKeepsEveryCachedValue`, which pins the default that still matters in production — the two-argument constructor's `alwaysFresh()`, i.e. `CredScopedFileSystem`'s load-once behaviour, asserting a cached value is never reloaded underneath a caller.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
tiennguyen-onehouse: Done, both now read `evictionListener cannot be null` and `isFresh cannot be null`. That also matches the wording `put()` already uses for its own key/value checks.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/util/BoundedKeyedCache.java
tiennguyen-onehouse: That is exactly the reasoning, thanks for spelling it out. `cacheLock` is only ever held for a `LinkedHashMap` get/put plus LRU bookkeeping, never across `loader.get()` — so the remaining global section is bounded by an in-memory map access, while the part that used to serialize every scope (the vend RPC) is now under the per-key lock.

tiennguyen-onehouse:
Thanks @openinx, no worries about the timing — and thanks @yili-db for the ping. All three comments addressed in 0439bad4:

- removed the single-argument constructor (no production caller) so the canonical three-argument one is the only way to build a cache with a policy;
- null-check messages now read `evictionListener cannot be null` / `isFresh cannot be null`;
- replied on the `getIfPresent` thread — your reading is right, `cacheLock` is never held across `loader.get()`.

`hadoop/test` is 1477/1477 and checkstyle is clean.

(Edited: I first wrote that this push had made your approval stale — it has not, the approval still stands on the branch. Sorry for the noise.) These were non-blocking, so if you would rather merge exactly the revision you reviewed (5e22e2c1), say the word and I'll drop this commit; it is only the constructor removal and the message wording.

@s-maling-celonis thanks for closing #1876 in favour of this and for confirming you hit the same stall independently — good to know the per-key locking covers your case too.

tiennguyen-onehouse:
One more cleanup in 6a2acd4c, in the same spirit as your constructor comment @openinx: I removed the package-private `keyLockCount()` accessor I had added on `BoundedKeyedCache`. It existed only so the concurrency tests could assert no per-key lock outlived the call that took it, which is test-only API in main code — the thing you flagged.

I checked whether anything was actually lost by mutating the two regressions it was meant to catch:

| Regression | `IdLockMapTest` | `BoundedKeyedCacheTest` |
|---|---|---|
| `IdLockMap.close()` skips the refcount release | 4 failures | 0 failures |
| `getOrLoad` acquires the key lock an extra time | 0 failures | 3 failures, incl. the pre-existing `keyLockReleasedAfterLoaderThrows` |

So the coverage holds either way. A lock-lifecycle bug is either in `IdLockMap`, where its own test already pins the count returning to zero, or in this class's use of it, where an unreleased hold deadlocks the next same-key caller and trips the timeout-bounded tests. The accessor only made those failures surface as an assertion instead of a timeout, which is not worth main-code API. Description updated with the same table, replacing the earlier claim that the accessor earned its keep.

`hadoop/test` is 1477/1477.

PR #1684, opened 2026-07-19
[hadoop] normalize REST credential responses

foss-contributor:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1684/files) to review incremental changes.
- [**stack/normalize-generic-credential**](https://github.com/unitycatalog/unitycatalog/pull/1684) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1684/files)] ← _this PR_
  - [stack/cloud-type-source-of-truth](https://github.com/unitycatalog/unitycatalog/pull/1686) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1686/files/33f58187235733e4553c7c443cfa52bb4abf6c4d..7f753c338b884203c30b219e76b71f94ba988369)]
    - [stack/cred-props-util-test-coverage](https://github.com/unitycatalog/unitycatalog/pull/1688) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1688/files/7f753c338b884203c30b219e76b71f94ba988369..92f93eebb67900fa4e3020ca41e3d1b70330f41a)]
      - [stack/credential-property-builders](https://github.com/unitycatalog/unitycatalog/pull/1692) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1692/files/92f93eebb67900fa4e3020ca41e3d1b70330f41a..329854fd63ebcbf5c686125afe63f88f5f0b7588)]
        - [stack/normalize-cloud-storage-prefixes](https://github.com/unitycatalog/unitycatalog/pull/1693) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1693/files/329854fd63ebcbf5c686125afe63f88f5f0b7588..1b079a53bc5117733ceaad3bd2ce8f7229ab2be9)]
          - [stack/location-aware-delegate-cache-keys](https://github.com/unitycatalog/unitycatalog/pull/1695) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1695/files/1b079a53bc5117733ceaad3bd2ce8f7229ab2be9..c4cf19b77c25d5486e6b1b6b5331b9e743a11382)]
            - [stack/credential-location-identity](https://github.com/unitycatalog/unitycatalog/pull/1699) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1699/files/c4cf19b77c25d5486e6b1b6b5331b9e743a11382..621dd3ea8d1460e4c419bba031a005245f4d7858)]
              - [stack/select-vended-credential-by-location](https://github.com/unitycatalog/unitycatalog/pull/1701) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1701/files/621dd3ea8d1460e4c419bba031a005245f4d7858..3220b7abc69ae612c5063402c4dfde2a5a79196b)]

---------
## Summary

Normalize Unity Catalog REST API and Delta REST API credential responses into a shared internal storage credential entity. This confines generated-model dependencies to the API boundary while preserving downstream credential handling.

## Testing

Added new tests for GenericCredential one-cloud only requirement to ensure that callers do not store more than one cloud credential.

## Issue


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredential.java
```diff
@@ -1,64 +1,81 @@
...
+      String awsAccessKeyId,
+      String awsSecretAccessKey,
+      String awsSessionToken,
+      String azureSasToken,
+      String gcsOauthToken,
+      Long expirationTimeMillis) {
+    boolean isAws = awsAccessKeyId != null || awsSecretAccessKey != null || awsSessionToken != null;
+    boolean isAzure = azureSasToken != null;
+    boolean isGcs = gcsOauthToken != null;
+    Preconditions.checkArgument(
```
foss-contributor: Do we want to keep this validation or move responsibility to the callers? Right now, from UC REST API --> GenericCredential and UC Delta API --> GenericCredential has validation. 

If we create this from Hadoop conf, there is not as strong validation so we will need to update the RenewalProviders.

clients/java/src/main/java/io/unitycatalog/client/internal/Preconditions.java
```diff
@@ -20,4 +20,9 @@ public static void checkState(boolean expression, String message, Object... args
       throw new IllegalStateException(String.format(message, args));
     }
   }
+
+  public static <T> T requireArgument(boolean expression, T value, String message, Object... args) {
```
openinx: I don't see there is any existing `requireArgument`  method in  the guava `Preconditions`. https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Preconditions.java#L333 

Seems like this method is ambiguous, as it named `requireArgument`,  but seems like it's checking the expression and then return a give value.  but the `expression` can be any kind of expression which is fully unrelated to the `value`, right ? 

I will suggest to remove this method, and instead,  let's use the `Preconditions.checkArgument` firstly, and then return the `value` directly in your caller. that's more straightforward.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredential.java
```diff
@@ -1,64 +1,81 @@
...
+ * <p>A plain, cloud-agnostic value type holding one cloud family's normalized credential values
+ * (AWS, Azure, or GCS) plus an optional expiration. Constructed via the {@link #forAws}, {@link
+ * #forAzure}, and {@link #forGcs} factories, each of which sets exactly one cloud family.
  */
-public class GenericCredential {
-  private final TemporaryCredentials tempCred;
-
-  public GenericCredential(TemporaryCredentials tempCred) {
-    this.tempCred = tempCred;
+public final class GenericCredential {
```
openinx: I will more suggest to make this `GenericCredential` as an abstracted class, and keep the `expirationTimeMillis` member, and `readyToRenew` method.  and make those aws/azure/gcp have their own subclasses named `Aws/Azure/Gcp/GenericCredential`.  

that will make each inheritence more clear, and nobody can use the same `GenericCredential` to ambiguously access the other cloud vender's credential fields.  Also if we want to support another cloud vendor, we won't need to touch the abstraction interface `GenericCredential`, that will be helpful for us to build the upper layer code.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredential.java
```diff
@@ -1,64 +1,81 @@
...
+  private GenericCredential(
+      String awsAccessKeyId,
+      String awsSecretAccessKey,
+      String awsSessionToken,
+      String azureSasToken,
+      String gcsOauthToken,
+      Long expirationTimeMillis) {
+    boolean isAws = awsAccessKeyId != null || awsSecretAccessKey != null || awsSessionToken != null;
+    boolean isAzure = azureSasToken != null;
+    boolean isGcs = gcsOauthToken != null;
```
openinx: It's actually not a good practice for us to have those concrete case-by-case in the abstracted `GenericCredential` class.  let's use the above comment to make it more clear.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCTemporaryCredentialUtil.java
```diff
@@ -0,0 +1,34 @@
...
+  private static final String DESCRIPTION = "UC temporary credentials";
+
+  private UCTemporaryCredentialUtil() {}
+
+  static GenericCredential toGenericCredential(TemporaryCredentials tempCred) {
+    Long expiration = tempCred.getExpirationTime();
+    if (tempCred.getAwsTempCredentials() != null) {
+      AwsCredentials aws = tempCred.getAwsTempCredentials();
+      return GenericCredential.forAws(
+          CredentialUtil.field(aws.getAccessKeyId(), DESCRIPTION, "AWS access key"),
```
openinx: The `CredentialUtil.field` looks wired with the constant `DESCRIPTION` ?  why not just simplify it as a separate method with the constant value directly ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCTemporaryCredentialUtil.java
```diff
@@ -0,0 +1,34 @@
...
+          CredentialUtil.field(aws.getSessionToken(), DESCRIPTION, "AWS session token"),
+          expiration);
+    } else if (tempCred.getAzureUserDelegationSas() != null) {
+      return GenericCredential.forAzure(
+          CredentialUtil.field(
+              tempCred.getAzureUserDelegationSas().getSasToken(), DESCRIPTION, "Azure SAS token"),
+          expiration);
+    } else {
+      var gcp = tempCred.getGcpOauthToken();
+      String oauthToken = gcp != null ? gcp.getOauthToken() : null;
```
openinx: we cannot fallback to use `null` if we are determined to use the the `Gcs` to wrap the credential.  it's more clear for us to throw rather than silently use `null`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CloudType.java
```diff
@@ -0,0 +1,17 @@
...
+
+public final class CloudType {
+  private CloudType() {}
+
+  public static boolean isSupportedScheme(String scheme) {
+    switch (scheme) {
+      case "s3":
+      case "gs":
+      case "abfs":
+      case "abfss":
```
openinx, line 11: Can make those a static list or set ?  that's more simpler to check. and we'd use it as  source of truth set to check what's the schema we currently support.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -473,6 +457,9 @@ private static Map<String, String> createCredProps(
       Map<String, String> appVersions,
       CredId credId)
       throws ApiException {
+    if (!CloudType.isSupportedScheme(scheme)) {
+      return Collections.emptyMap();
+    }
```
openinx, line 471: wait, why do we need to add this extra check and silently use the `emptyMap`, as we already have the default case in the `switch` branches to handle the `emptyMap` in line499,  will it be duplicated code ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/DeltaStorageCredentialUtil.java
```diff
@@ -3,49 +3,44 @@
...
+          CredentialUtil.field(config.getGcsOauthToken(), credDescription, "GCS OAuth token"),
+          expiry);
     }
-    return out;
   }
 
   /** Selects the credential whose prefix covers the requested location. */
   public static DeltaStorageCredential selectForLocation(
       String location, List<DeltaStorageCredential> creds) {
+    String responseDescription = "UC Delta API response for '" + location + "'";
```
openinx: Use string template with `%s` to fill the error message lazily only when the `expression` that we want to check is `true`,  otherwise it's not worthy to concat the string whenever the `expression` is `true` or `false`, which can be CPU bottleneck if this code path is hot enough. 

```java
Preconditions.checkArgument(creds != null && !creds.isEmpty(),  "UC Delta API response for '%s' has no storage credential", location) 
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/DeltaStorageCredentialUtil.java
```diff
@@ -107,12 +99,7 @@ private static boolean isAzureConfig(DeltaStorageCredentialConfig c) {
...
 
-  private static String field(String value, DeltaStorageCredential cred, String label) {
-    Preconditions.checkArgument(
-        value != null && !value.isEmpty(),
-        "UC Delta credential for '%s' is missing %s.",
-        cred.getPrefix(),
-        label);
-    return value;
+  private static String credDescription(DeltaStorageCredential cred) {
+    return "UC Delta credential for '" + cred.getPrefix() + "'";
```
openinx: no front-load error message concat even before we know the `checkArgument` result,  pls always use the error message template to fill the message when we are pretty sure we need to fill those error message when the condition of `checkArgument` is false.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/BaseTokenProviderTest.java
```diff
@@ -67,92 +69,44 @@ public void after() {
     GenericCredentialProvider.globalCache.clear();
   }
 
-  @Test
-  public void testTableTemporaryCredentialsRenew() throws Exception {
```
openinx, line 71: be very careful to remove those initial/renewal credential test cases, those are pretty important.   do you already have the equvalent test cases ??

openinx (changes requested):
Left some comments, pls take a look.  thank you @foss-contributor for taking this.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CloudType.java
foss-contributor: fixed in follow up PR - this is temporary fix to pass local file tests

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
foss-contributor: fix for failing local file tests - these tests receive no credentials from the UC server. before, we did not validate this response so the tests pass. Now we validate the response, and no creds = crash = fail.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/BaseTokenProviderTest.java
foss-contributor: removing the test simplification

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
```diff
@@ -0,0 +1,12 @@
+package io.unitycatalog.hadoop.internal;
+
+import io.unitycatalog.client.internal.Preconditions;
+
+public final class CredentialUtil {
+  private CredentialUtil() {}
+
+  public static String field(String value, String message, Object... args) {
+    Preconditions.checkArgument(value != null && !value.isEmpty(), message, args);
```
openinx: Can you remove this method, as I commented before this his ambigourous and unclear. 

in the real caller, let's directly use the 

```
# for value1
Preconditions.checkArgument(...)
this.value = value. 

# for value2
Preconditions.checkArgument(...)
this.value = value. 
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCTemporaryCredentialUtil.java
```diff
@@ -0,0 +1,24 @@
+package io.unitycatalog.hadoop.internal.auth;
+
+import io.unitycatalog.client.model.AwsCredentials;
+import io.unitycatalog.client.model.TemporaryCredentials;
+
+/** Converts a UC SDK {@link TemporaryCredentials} into an internal {@link GenericCredential}. */
+final class UCTemporaryCredentialUtil {
```
openinx: In this PR, we introduce several credential utils.  
* UCTemporaryCredentialUtil
* CredentialUtil

And we already have the `CredPropsUtil` and `DeltaStorageCredentialUtil`. 

actually, we don't need so much utility for the cred.  As I commented in the CredentialUtil,  I think it can be removed. 

and for `UCTemporaryCredentialUtil` and `DeltaStorageCredentialUtil`,  I can we can  unify them into a single `CredentialUtil`,  that will help us simply the whole work.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialTest.java
```diff
@@ -0,0 +1,43 @@
+package io.unitycatalog.hadoop.internal.auth;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+class GenericCredentialTest {
+
+  @Test
+  void awsCredentialExposesItsFields() {
```
openinx: If you wan to add the coverage for the `GenericCredential`,  then I think the `equals` and `hashCode` should be the 1st priority, for all the cloud creds, which we miss to cover actually. and the field assertion is trivial.

openinx: Since we already have this one https://github.com/unitycatalog/unitycatalog/pull/1684/changes#diff-dd975178e2a4b5761264eab6cf8478c7b54b9a60bb2f68cb02e07c909788dae5R15, then this class is unnecessary I think.

openinx (commented):
high-level looks good to me, but I left few comments. Please address those comments, then I will approve and merge.  Thanks @foss-contributor for the contribution !

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialTest.java
foss-contributor: added just hash and equals tests

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCTemporaryCredentialUtil.java
foss-contributor: moved them to CredentialUtil

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
foss-contributor: fixed

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/DeltaStorageCredentialUtilTest.java
foss-contributor, line 1: moved to CredentialUtilTest.java

openinx (approved):

PR #1685, opened 2026-07-19
UUID-based commit idempotency for CCv2

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1685/files) to review incremental changes.
- [stack/ccv2_idempotency_uuid](https://github.com/unitycatalog/unitycatalog/pull/1685) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1685/files)]

---------
**PR Checklist**

**Description of changes**

UUID-based commit idempotency for CCv2

Replaying a commit already accepted for a table is now an idempotent no-op success instead of a 409 conflict, so a client that lost the response to an accepted commit can safely resend.

- Detect replays via the per-commit UUID in the commit file name.
- Serialize per-table commit/backfill with a `PESSIMISTIC_WRITE` lock on the `uc_tables` row, shared by the REST and Delta paths.
- Short-circuit a replay with `CommitAlreadyAcceptedException` to roll the transaction back to a no-op; defer `assert-etag` past the apply so a stale-etag replay still succeeds.

Tested on both entry points, with full branch coverage of the replay check.

### DIFF
--- a/api/all.yaml
+++ b/api/all.yaml
@@ -1464,6 +1464,8 @@ paths:
         most cases the number of unbackfilled commits should be close to zero or one. But if
         clients misbehave and unbackfilled commits accumulate beyond the limit, server will
         reject further commits until more backfill is done.
+        Replaying an already-accepted commit (same version and file-name) is an idempotent no-op
+        that returns 200, so a client that lost the response can safely resend.
         WARNING: This API is experimental and may change in future versions.
       operationId: commit
       requestBody:
@@ -1491,7 +1493,7 @@ paths:
         '404':
           description: Not Found. The specified table does not exist.
         '409':
-          description: Conflict. Indicates that a commit with the same version already exists.
+          description: Conflict. Indicates that a different commit with the same version already exists.
         '429':
           description: The maximum number of unbackfilled commits per table has been reached. Please backfill some commits before retrying.
         '500':
--- a/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
@@ -765,6 +765,9 @@ public class DeltaCommitRepository {
    * conservatively reported as not-a-replay. The caller must hold the table lock (see {@link
    * #lockTableForCommit}).
    *
+   * <p>Filename with the client-generated, per-commit-unique UUID (or file content, if filename is
+   * no longer tracked in DB) is used for dedup.
+   *
    * @return {@code true} only if this matches an already-accepted commit; {@code false} for a fresh
    *     version, a genuine conflict, or a purged version
    */
--- a/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
@@ -4,6 +4,7 @@ import static java.sql.Connection.TRANSACTION_REPEATABLE_READ;
 
 import io.unitycatalog.server.exception.BaseException;
 import io.unitycatalog.server.exception.ErrorCode;
+import io.unitycatalog.server.exception.TransactionRollbackException;
 import io.unitycatalog.server.model.ColumnInfos;
 import io.unitycatalog.server.model.DataSourceFormat;
 import io.unitycatalog.server.model.DeltaCommit;
@@ -25,6 +26,7 @@ import io.unitycatalog.server.utils.IdentityUtils;
 import io.unitycatalog.server.utils.NormalizedURL;
 import io.unitycatalog.server.utils.ServerProperties;
 import io.unitycatalog.server.utils.ValidationUtils;
+import jakarta.persistence.PessimisticLockException;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
@@ -34,6 +36,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import org.hibernate.LockMode;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.query.NativeQuery;
@@ -65,6 +68,20 @@ import org.slf4j.LoggerFactory;
  * </ol>
  */
 public class DeltaCommitRepository {
+
+  /**
+   * Thrown when a commit request is recognized as an idempotent replay of an already-accepted
+   * commit, to roll the whole transaction back so nothing the request applied (the commit, and on
+   * the Delta path any sibling metadata) is persisted; the commit entry point then catches it and
+   * reports a no-op success.
+   *
+   * <p>Not a client-facing error, so it deliberately does not extend {@link BaseException}. As a
+   * {@link TransactionRollbackException}, {@code TransactionManager} rolls back and rethrows it
+   * as-is instead of wrapping it into an {@code INTERNAL} error; it is always caught within {@link
+   * #postCommit} / {@link io.unitycatalog.server.persist.TableRepository#updateTableForDelta} and
+   * never reaches the HTTP layer.
+   */
+  static class CommitAlreadyAcceptedException extends TransactionRollbackException {}
 
   private static final Logger LOGGER = LoggerFactory.getLogger(DeltaCommitRepository.class);
 
@@ -231,11 +248,23 @@ public class DeltaCommitRepository {
    * <p>The method validates the commit, ensures the table is a managed Delta table, and performs
    * the appropriate commit operation within a transaction.
    *
+   * <p>Replaying a commit already accepted for this table is an idempotent no-op success (not a
+   * conflict), so a client that lost the response can safely resend. See {@link
+   * #isAcceptedCommitReplay}.
+   *
    * @param commit the commit request containing version info, metadata, and backfill information
    * @throws BaseException if the commit is invalid, table is not found, or commit limits are
    *     exceeded
    */
   public void postCommit(DeltaCommit commit) {
+    try {
+      postCommitInTransaction(commit);
+    } catch (CommitAlreadyAcceptedException e) {
+      // Idempotent replay: the transaction rolled back to a no-op. Report success.
+    }
+  }
+
+  private void postCommitInTransaction(DeltaCommit commit) {
     serverProperties.checkManagedTableEnabled();
     validateCommit(commit);
     // Extract + shape-validate uniform fields outside the transaction. The subpath check (which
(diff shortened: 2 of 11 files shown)


server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -681,6 +755,78 @@ private static void saveCommit(Session session, UUID tableId, DeltaCommitInfo co
...
+      // Below the oldest tracked version, i.e. backfilled and purged. This could be a genuine
+      // replay, but the file name is no longer stored, so we can't confirm it. Conservatively
+      // treat it as a conflict.
+      // TODO: close this gap by comparing the incoming staged file against the published
+      // _delta_log/<v>.json.
+      return false;
+    }
+    // The file name embeds a client-generated, per-commit-unique UUID, used as the dedup handle: a
+    // match means the same commit replayed; a difference means another writer won that version.
+    return existing.getCommitFilename().equals(fileName);
```
timothyw553, line 816: it seems this is basically the crux of the dedup. I do wonder if there would be any case where filename matches but things like file_size, file_modification_timestamp, etc do not?

if its intended, then `isAcceptedCommitReplay` should perhaps very explicitly say so (that dedup is solely filename and not the rest).

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -681,6 +755,78 @@ private static void saveCommit(Session session, UUID tableId, DeltaCommitInfo co
...
+    if (version == lastCommitDAO.getCommitVersion()) {
+      existing = lastCommitDAO;
+    } else if (version == firstCommitDAO.getCommitVersion()) {
+      existing = firstCommitDAO;
+    } else if (version > firstCommitDAO.getCommitVersion()
+        && version < lastCommitDAO.getCommitVersion()) {
+      // Tracked commit versions are contiguous between the boundaries. The caller holds the table
+      // write lock, so no concurrent backfill can purge this version between reading the boundaries
+      // and this lookup; an in-range version must therefore have a row. A missing row means the
+      // uc_delta_commits table is internally inconsistent (a gap in the tracked range).
```
timothyw553, line 795: can we add two-thread integration test(s) that try to race each other? i don't know how expensive that may be

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -681,6 +755,78 @@ private static void saveCommit(Session session, UUID tableId, DeltaCommitInfo co
...
+                              + tableId
+                              + ": no row tracked at in-range commit version "
+                              + version));
+    } else {
+      // Below the oldest tracked version, i.e. backfilled and purged. This could be a genuine
+      // replay, but the file name is no longer stored, so we can't confirm it. Conservatively
+      // treat it as a conflict.
+      // TODO: close this gap by comparing the incoming staged file against the published
+      // _delta_log/<v>.json.
+      return false;
```
timothyw553, line 812: just a thought: do we need to update api/...yaml docs for this 409->200 behaviour?

timothyw553 (commented):
some comments

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: Two-thread racing test would be non-deterministic. Instead there're normal tests to cover both conflict and idempotent cases in `SdkDeltaCommitsCRUDTest`:
```
    // Replay old version 1 with a different file name -> genuine conflict.
    assertApiException(
        () -> deltaCommitsApi.commit(commit1_other_filename),
        ErrorCode.ALREADY_EXISTS,
        "Commit version already accepted.");
```
and 
```
    DeltaCommit commit1 =
        createCommitObject(tableInfo.getTableId(), 1L, tableInfo.getStorageLocation());
    deltaCommitsApi.commit(commit1);

    // UUID-based idempotency: replaying the identical commit (same version + same file name) is a
    // no-op success, not a conflict, so a client that lost the response is not misled into
    // rebasing and duplicating the commit. The table state is unchanged.
    deltaCommitsApi.commit(commit1);
    verifyDeltaCommits(
        /* expectedLatestTableVersion= */ 1, /* expectedCommits= */ commit1);
```

timothyw553 (approved):
Approved. I think there are some docs changes like `ManagedTablesSpec.md` that may need to be updated as well,? @openinx please have a look.

yili-db:
> Approved. I think there are some docs changes like `ManagedTablesSpec.md` that may need to be updated as well,? @openinx please have a look.

Thank you @timothyw553 . Doc update is in https://github.com/unitycatalog/unitycatalog/pull/1666

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -260,12 +293,37 @@ public void postCommit(DeltaCommit commit) {
...
+   * <p>A lock-wait timeout or deadlock surfaces as {@code UPDATE_REQUIREMENT_CONFLICT} (409) so the
+   * client retries, rather than as a generic 500.
+   */
+  public static void lockTableForCommit(Session session, TableInfoDAO dao, UUID tableId) {
+    try {
+      session.refresh(dao, LockMode.PESSIMISTIC_WRITE);
+    } catch (PessimisticLockException e) {
+      throw new BaseException(
+          ErrorCode.UPDATE_REQUIREMENT_CONFLICT,
+          "Concurrent commit in progress on table " + tableId + "; retry the request.");
```
openinx: nit: This shared message prints the table UUID, but the Delta update path previously reported `catalog.schema.table`. Since REST postCommit only has the tableId, consider passing a display label so the Delta path keeps the human-readable identifier.

server/src/test/java/io/unitycatalog/server/sdk/delta/SdkUpdateTableTest.java
```diff
@@ -640,6 +662,201 @@ h, new DeltaRemoveDomainMetadataUpdate().domains(List.of("delta.unknown"))),
...
+                      new DeltaCommit()
+                          .version(2L)
+                          .timestamp(1700000002L)
+                          .fileName("00000002.json")
+                          .fileSize(1024L)
+                          .fileModificationTimestamp(1700000002L)));
+      assertThat(rNext.getLatestTableVersion()).isEqualTo(2L);
+    }
+
+    // -------- a commit replay is a whole-request no-op, even with a stale etag and extra actions
```
openinx: nit: This comment wraps awkwardly (the section rule breaks onto a dangling `// --` line, and `backfill`/`set-latest-backfilled-version` split mid-word). Consider reflowing so the header stays on one line.

openinx (commented):
Left comments.

api/all.yaml
```diff
@@ -1464,6 +1464,8 @@ paths:
         most cases the number of unbackfilled commits should be close to zero or one. But if
         clients misbehave and unbackfilled commits accumulate beyond the limit, server will
         reject further commits until more backfill is done.
+        Replaying an already-accepted commit (same version and file-name) is an idempotent no-op
```
openinx, line 1467: why do we want to change the behvior as the described one ?

openinx: and different clients commit the same commit to the same versions, will one of the them get the  409 conflict ?

openinx, line 1467: **Blocking:** `spec/protocols/ManagedTablesSpec.md:871` still says re-sending an already-accepted `add-commit` fails with `CommitVersionConflictException`. That contradicts this PR. Please update the Commit recovery section (and call out whole-request no-op + the purged-version gap) so the protocol matches the server.

openinx (commented):
Left comments.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -681,6 +758,81 @@ private static void saveCommit(Session session, UUID tableId, DeltaCommitInfo co
...
+   * this table, rather than a fresh commit or a genuine conflict at an already-taken version. A
+   * client that lost the response to an accepted commit may safely resend the whole request;
+   * recognizing the replay lets the server report success instead of a spurious conflict.
+   *
+   * <p>Covers only versions still tracked in the DB; a replay of a backfilled-and-purged version is
+   * conservatively reported as not-a-replay. The caller must hold the table lock (see {@link
+   * #lockTableForCommit}).
+   *
+   * <p>Filename with the client-generated, per-commit-unique UUID (or file content, if filename is
+   * no longer tracked in DB) is used for dedup.
```
openinx: nit: This javadoc claims file-content dedup when the filename is no longer tracked, but the code still returns conflict + TODO below. Please drop the claim or implement it.

server/src/test/java/io/unitycatalog/server/sdk/delta/SdkUpdateTableTest.java
```diff
@@ -640,6 +662,199 @@ h, new DeltaRemoveDomainMetadataUpdate().domains(List.of("delta.unknown"))),
           "next version");
     }
 
+    // -------- idempotent replay of non-newest tracked versions (first-DAO + DB-lookup branches) --
```
openinx, line 665: nit: This section header wraps awkwardly (dangling `--`). Consider reflowing so it stays on one line.

api/all.yaml
yili-db: That's the commit idempotency from the same client for the same commit: RPC can timeout etc and client would retry. Client needs to receive a 200 if the commit being retried is already landed by previously timed out RPC.
Otherwise if the false signal of "commit conflict, your previous RPC did not land a commit" is received by client, client would rebase and commit a new commit with the same content and cause dup commit problem.

yili-db: > and different clients commit the same commit 

No they simply can't in the normal sense. The commit filename has a random UUID. Two clients should never have the same UUID for any commit.

yili-db: Doc update is in https://github.com/unitycatalog/unitycatalog/pull/1666

server/src/test/java/io/unitycatalog/server/sdk/delta/SdkUpdateTableTest.java
yili-db: It is already one line

api/all.yaml
openinx: Okay, thanks for the  explanation.  that's clear.

openinx (approved):

tdas (approved):

PR #1687, opened 2026-07-20
[Server] Consolidate Iceberg FileIO behind FileOperations

yili-db:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1687/files) to review incremental changes.
- [**stack/fileiofactory**](https://github.com/unitycatalog/unitycatalog/pull/1687) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1687/files)] ← _this PR_

---------
**PR Checklist**

**Description of changes**
Fold FileIOFactory into FileOperations as the single entry point for storage access, including credential-vended ResolvingFileIO for cloud paths and an expanded SimpleLocalFileIO for local paths.
- Move SimpleLocalFileIO from service/iceberg to persist/utils
- Wire MetadataService and TableConfigService through FileOperations
- Create FileOperations at server startup with StorageCredentialVendor
- Remove FileOperations from Repositories and unused repository fields


server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -1,24 +1,29 @@
...
+   * @param metadataLocation the normalized location of the {@code *.metadata.json} file
+   */
+  public TableMetadata readTableMetadata(NormalizedURL metadataLocation) {
     // TODO: cache fileIO
-    FileIO fileIO = fileIOFactory.getFileIO(NormalizedURL.from(metadataLocation));
-
-    return CompletableFuture.supplyAsync(() -> TableMetadataParser.read(fileIO, metadataLocation))
-        .join();
+    FileIO fileIO = fileOperations.getFileIO(metadataLocation);
+    return TableMetadataParser.read(fileIO, metadataLocation.toString());
```
timothyw553: I saw below `getFileIO` builds a `ResolvingFileIO`. here, should we be closing it by calling fileIO.close()?
or maybe something like
```java
  try (FileIO fileIO = fileOperations.getFileIO(metadataLocation)) {
    return TableMetadataParser.read(fileIO, metadataLocation.toString());
  }
```

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -79,4 +83,88 @@ private static void createLocalDirectory(NormalizedURL url) {
...
+      // deliberately do NOT route these through ResolvingFileIO: it resolves the file:// scheme to
+      // Iceberg's HadoopFileIO, which requires hadoop-client-runtime on the classpath. The server
+      // only depends on hadoop-client-api, and SimpleLocalFileIO covers the local read and
+      // directory operations we need without that heavy runtime dependency.
+      case FILE, NULL -> new SimpleLocalFileIO();
+      case S3, GS, ABFS, ABFSS -> {
+        ResolvingFileIO fileio = new ResolvingFileIO();
+        fileio.initialize(getFileIOConfig(path));
+        yield fileio;
+      }
```
timothyw553, line 104: I think this needs tests

timothyw553: something that lets `FileOperations.getFileIO("s3://…")` build a real `ResolvingFileIO` and read through it against the S3Mock

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -79,4 +83,88 @@ private static void createLocalDirectory(NormalizedURL url) {
...
+  }
+
+  private Map<String, String> getS3Config(NormalizedURL path, AwsCredentials awsCredentials) {
+    // TODO: if region isn't configured, use HEAD bucket to figure out
+    String s3Region = s3BucketRegionMap.get(path.getStorageBase());
+    return Map.of(
+        S3FileIOProperties.ACCESS_KEY_ID, awsCredentials.getAccessKeyId(),
+        S3FileIOProperties.SECRET_ACCESS_KEY, awsCredentials.getSecretAccessKey(),
+        S3FileIOProperties.SESSION_TOKEN, awsCredentials.getSessionToken(),
+        AwsClientProperties.CLIENT_REGION, s3Region);
```
timothyw553, line 180: > When you call the .get(key) method on a Java Map and the key is not found, the method returns null. It does not throw an exception.

in this case s3Region may be null, right? should we add a guard before this map

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -79,4 +83,88 @@ private static void createLocalDirectory(NormalizedURL url) {
...
+      // Local (file://) paths need no cloud credentials, so short-circuit before vending: the
+      // scheme is known here and vending would do a needless external-location lookup for local
+      // tables.
+      return Map.of();
+    }
+
+    // FIXME!! privileges are defaulted to READ only here for now as Iceberg REST impl doesn't
+    //  support write
+    TemporaryCredentials cred =
+        storageCredentialVendor.vendCredential(path, Set.of(CredentialContext.Privilege.SELECT));
```
timothyw553, line 127: recently I've discovered that there may be instances where deployments may actually just rely on ambient credentials for reading un-vended GCP tables. so I wonder, will that behaviour persist here?

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -79,4 +83,88 @@ private static void createLocalDirectory(NormalizedURL url) {
...
+    if (cred.getAzureUserDelegationSas() != null) {
+      return getADLSConfig(path, cred.getAzureUserDelegationSas());
+    } else if (cred.getGcpOauthToken() != null) {
+      return getGCSConfig(cred.getGcpOauthToken(), cred.getExpirationTime());
+    } else if (cred.getAwsTempCredentials() != null) {
+      return getS3Config(path, cred.getAwsTempCredentials());
+    } else {
+      // Cloud vend returned no recognized credential type; nothing to configure.
+      return Map.of();
```
timothyw553: is it better to throw directly here with a clear message?

timothyw553: I guess even if it's impossible, better to throw I think (up to you)

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
```diff
@@ -0,0 +1,181 @@
...
+   */
+  private static CloseableIterable<Path> walkPrefix(
+      String prefix, boolean includeDirectories, boolean bottomUp) {
+    Path dirPath = toPath(prefix);
+    if (!Files.exists(dirPath)) {
+      return CloseableIterable.empty();
+    }
+    Stream<Path> walk;
+    try {
+      walk = Files.walk(dirPath);
```
timothyw553, line 143: `Files.walk(dirPath, FileVisitOption.FOLLOW_LINKS)` to follow symlinks like in server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java

timothyw553 (commented):
Some comments.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: Yes ambient credentials on server can also be used for vending a temp table cred. This is where it is picked up for S3:
```
        // Otherwise just defer to DefaultCredentialsProvider
        credentialsProvider = DefaultCredentialsProvider.create();
```

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
yili-db: It's intentionally removed. `FOLLOW_LINKS` is dangerous.

timothyw553 (approved):
Generally looks good to me

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -2,31 +2,54 @@
...
+/**
+ * Single entry point for all storage/file access in the server. Covers both directory lifecycle
+ * management for managed storage locations (create/delete) and credential-vended Iceberg {@link
+ * FileIO} construction used by the Iceberg REST catalog.
+ */
 public class FileOperations {
 
-  private final ServerProperties serverProperties;
+  private final StorageCredentialVendor storageCredentialVendor;
+  private final Map<NormalizedURL, String> s3BucketRegionMap;
```
openinx, line 37: I understand it's the correct direction to unify all cloud storage access in a abstracted FileOperations, but seems like we are implementing all the cloud storage operations inside this FileOperations class ?  ultimately I assume we will need an abstraction to fully decouple those cloud storage access into several fully isolated classes, so that one cloud won't perceive and access any other cloud resource.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -2,31 +2,54 @@
...
 
   /** Delete entire directory recursively. Note that currently it does nothing for cloud FS */
-  public static void deleteDirectory(NormalizedURL url) throws IOException {
+  public static void deleteDirectory(NormalizedURL url) {
     switch (UriScheme.fromURI(url.toUri())) {
-      case FILE, NULL -> deleteLocalDirectory(url);
+      // Directory deletion for local paths is handled by SimpleLocalFileIO.
+      case FILE, NULL -> SimpleLocalFileIO.deleteDirectory(url.toString());
       // Currently we can NOT delete the path in cloud storage. We will update this in future
       // when UC OSS begins using the hadoopfs libraries.
```
openinx, line 54: this is unclear for me,   as the iceberg `FileIO` and hadoop `FileSystem` are the similar abstraction that we should use one of them to access the cloud storage entirely,  the case that mixing two to achieve the access is not the correct direction I think.    WDYT?

openinx: I also know the difference between `FileIO` and `FileSystem` is,  `FileIO`  clearly does not have any directory concepts, while `FileSystem` provides more comprehensive concepts for us.   It's more likely `FileSystem` would be better.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -2,31 +2,54 @@
...
       case S3 -> {}
```
openinx, line 55: Shouldn't we throw rather than silent return ?

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -79,4 +83,100 @@ private static void createLocalDirectory(NormalizedURL url) {
       throw new BaseException(ErrorCode.INTERNAL, "Failed to create directory: " + dirPath, e);
     }
   }
+
+  /**
+   * Returns an Iceberg {@link FileIO} for reading the given location. Local paths use {@link
+   * SimpleLocalFileIO}; cloud paths use a credential-vended {@link ResolvingFileIO}.
+   */
+  // TODO: Cache fileIOs
```
openinx, line 91: That's really important, otherwise, every access will exhaustively consume all the resources by creating massive fileIO instance.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: The core part of this file is the `getFileIO`, `getFileIOConfig` and `get*Config` functions which directly generates cloud specific FileIO objects. `s3BucketRegionMap` happens to be part of this process.

Other than that, all the other functions will be cloud agnostic. The abstraction is already handled by `getFileIOConfig`. Different cloud implementation of actual file IO is implemented by `ResolvingFileIO` (and its delegation).

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -79,4 +83,100 @@ private static void createLocalDirectory(NormalizedURL url) {
...
+    return switch (UriScheme.fromURI(path.toUri())) {
+      // Local paths are served by SimpleLocalFileIO (backed by java.nio + iceberg-core). We
+      // deliberately do NOT route these through ResolvingFileIO: it resolves the file:// scheme to
+      // Iceberg's HadoopFileIO, which requires hadoop-client-runtime on the classpath. The server
+      // only depends on hadoop-client-api, and SimpleLocalFileIO covers the local read and
+      // directory operations we need without that heavy runtime dependency.
+      case FILE, NULL -> new SimpleLocalFileIO();
+      case S3, GS, ABFS, ABFSS -> {
+        ResolvingFileIO fileio = new ResolvingFileIO();
+        fileio.initialize(getFileIOConfig(path));
```
openinx, line 102: Actually,  you can use HadoopFileIO directly to get a `FileIO` https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/hadoop/HadoopFileIO.java.  that will automatically address your fileIO cache problem, since the `FileSystem` will have the cache by default.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: This comment was put back in Feb. Later this will be migrated to `DelegateFileIO.deletePrefix`. My current plan is not using hadoop fs but just stick to Iceberg FileIO.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: If it throws, deleting a managed entity on cloud would simply fail. This is incomplete and pre-existing.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: Yes. But not in this PR yet.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: The downsides of using Hadoop FS (`hadoop-client-runtime`) is heavy dependency. It's a shaded uber-jar of an entire ecosystem that weighs ~30 MB, ~18,000 classes adding to the server runtime.
For now I don't see a good reason to adopt that yet.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
openinx: okay, if we are saying the `deletePrefix`, then that make sense to me.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -79,4 +83,100 @@ private static void createLocalDirectory(NormalizedURL url) {
...
+  public Map<String, String> getFileIOConfig(NormalizedURL path) {
+    UriScheme scheme = UriScheme.fromURI(path.toUri());
+    if (scheme == UriScheme.FILE || scheme == UriScheme.NULL) {
+      // Local (file://) paths need no cloud credentials, so short-circuit before vending: the
+      // scheme is known here and vending would do a needless external-location lookup for local
+      // tables.
+      return Map.of();
+    }
+
+    // FIXME!! privileges are defaulted to READ only here for now as Iceberg REST impl doesn't
```
openinx, line 124: If iceberg rest impl does not support vending write cred, then the IRC cannot support write for now right ?

openinx (commented):
I skimmed this PR again,  and I'm looking forward to seeing few answers from the questions: 

1.   Seems like we are agreed to reuse the existing iceberg `FileIO` as the fundermental storage abstraction.  then will we consider to use the abstracted approach to initialize the concrete `FileIO` instance ?   as  I like to see we fully isolate those cloud vender into separate and isolated classes, rather than mixing them together, which may distract us.

2. As long term we will require the server side cloud credential renewal and caching work,  without the 1st abstraction and  isolation,  it may not clearly solve and support server-side credential renwal and caching.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
yili-db: IRC write support is not implemented yet.

yili-db:
> I skimmed this PR again, and I'm looking forward to seeing few answers from the questions:
> 
> 1. Seems like we are agreed to reuse the existing iceberg `FileIO` as the fundermental storage abstraction.  then will we consider to use the abstracted approach to initialize the concrete `FileIO` instance ?   as  I like to see we fully isolate those cloud vender into separate and isolated classes, rather than mixing them together, which may distract us.
> 2. As long term we will require the server side cloud credential renewal and caching work,  without the 1st abstraction and  isolation,  it may not clearly solve and support server-side credential renwal and caching.

Thanks for the second pass @openinx. Answers to the two questions:

**1. Isolating the concrete `FileIO` per cloud.** The concrete per-cloud FileIO objects are already isolated — `getFileIO` returns Iceberg's `ResolvingFileIO`, which dispatches by scheme to `S3FileIO` / `GCSFileIO` / `ADLSFileIO`. No cloud accesses another cloud's resources at the FileIO level.

What's centralized in `FileOperations` today is only the *config assembly* (`getS3Config` / `getGCSConfig` / `getADLSConfig`) that turns vended credentials into the property map used to initialize each FileIO. They're already separate methods, and I agree they could be extracted into per-cloud classes (e.g. a scheme-keyed config provider with one impl per cloud). I'd prefer to do that as a follow-up — this PR is intentionally a focused consolidation: folding `FileIOFactory` into `FileOperations` and removing the config logic that was previously duplicated across both `FileIOFactory` and `TableConfigService`.

**2. Server-side credential renewal + caching.** Agreed, and neither is implemented in this PR (see the `// TODO: Cache fileIOs`). The relevant point is that all vending now goes through the single `getFileIOConfig` entry point, which is where caching (keyed by storage base, expiry-aware) and renewal naturally belong. The per-cloud config isolation from Q1 pairs with that work, so I'd like to tackle them together in the follow-up.

openinx (approved):

PR #1689, opened 2026-07-20
Add configurable expiry for UC access tokens issued by token exchange

s-maling-celonis:
## Summary

Split from #1661 (PR 1 of 3).

UC access tokens previously had no `exp` claim. This adds:

- `server.access-token-timeout` (default `PT24H`)
- JWT expiration via `withExpiresAt` in `SecurityContext.createAccessToken`
- `expires_in` in the `/auth/tokens` response so clients can cache and refresh correctly
- Cookie max-age capped to `min(server.cookie-timeout, server.access-token-timeout)` when `ext=cookie`
- Helm support via `auth.accessTokenTimeout`

## Behavior change

Previously, UC access tokens issued by `/auth/tokens` had **no `exp` claim** and did not expire.

This PR changes the default lifetime to **24 hours** (`server.access-token-timeout=PT24H`).

**Impact:**
- Clients must re-exchange tokens (or refresh using `expires_in`) at least once per day with default settings.
- Deployments that relied on non-expiring tokens must set `server.access-token-timeout` explicitly (for example to a longer duration) if they need backward-compatible behavior during migration.

**Scope:** Applies only to UC access tokens issued by `/auth/tokens` token exchange (`JwtTokenType.ACCESS`). The bootstrap admin token in `etc/conf/token.txt` (service token, `JwtTokenType.SERVICE`) is unchanged and still does not expire.

## Test plan

- [x] `sbt "server/testOnly io.unitycatalog.server.service.AuthServiceTest, io.unitycatalog.server.utils.ServerPropertiesTest"`
- [x] Token exchange returns `expires_in` matching configured timeout
- [x] Issued UC access token JWT includes valid `exp` claim (`exp ≈ iat + TTL`)
- [x] Expired access tokens are rejected with 401
- [x] Cookie max-age does not exceed access token lifetime

### DIFF
--- a/etc/conf/server.properties
+++ b/etc/conf/server.properties
@@ -28,6 +28,8 @@ server.audiences=
 
 # D-Days H-Hours M-Minutes S-Seconds (P5D = 5 days,PT5H = 5 hours, PT5M = 5 minutes, PT5S = 5 seconds)
 server.cookie-timeout=P5D
+# Lifetime of UC access tokens issued by token exchange. Previously these tokens had no exp claim.
+# Default is 24 hours; increase this value if you need longer-lived sessions during migration.
 server.access-token-timeout=PT24H
 
 # Enable MANAGED table
--- a/etc/conf/server.properties
+++ b/etc/conf/server.properties
@@ -28,6 +28,7 @@ server.audiences=
 
 # D-Days H-Hours M-Minutes S-Seconds (P5D = 5 days,PT5H = 5 hours, PT5M = 5 minutes, PT5S = 5 seconds)
 server.cookie-timeout=P5D
+server.access-token-timeout=PT24H
 
 # Enable MANAGED table
 # Default: true (enabled)
--- a/etc/conf/server.properties
+++ b/etc/conf/server.properties
@@ -28,8 +28,6 @@ server.audiences=
 
 # D-Days H-Hours M-Minutes S-Seconds (P5D = 5 days,PT5H = 5 hours, PT5M = 5 minutes, PT5S = 5 seconds)
 server.cookie-timeout=P5D
-# Lifetime of UC access tokens issued by token exchange. Previously these tokens had no exp claim.
-# Default is 24 hours; increase this value if you need longer-lived sessions during migration.
 server.access-token-timeout=PT24H
 
 # Enable MANAGED table
--- a/server/src/test/java/io/unitycatalog/server/service/AuthServiceTest.java
+++ b/server/src/test/java/io/unitycatalog/server/service/AuthServiceTest.java
@@ -22,6 +22,7 @@ import java.security.KeyPairGenerator;
 import java.security.NoSuchAlgorithmException;
 import java.security.interfaces.RSAPrivateKey;
 import java.security.interfaces.RSAPublicKey;
+import java.time.Duration;
 import java.util.Date;
 import java.util.UUID;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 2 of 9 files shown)


etc/conf/server.properties
```diff
@@ -28,6 +28,7 @@ server.audiences=
 
 # D-Days H-Hours M-Minutes S-Seconds (P5D = 5 days,PT5H = 5 hours, PT5M = 5 minutes, PT5S = 5 seconds)
 server.cookie-timeout=P5D
```
yili-db, line 30: The cookie still expires in 5 days, but the JWT inside expires in 1 day? A cookie with expired token is useless for the remaining 4 days, right?

I suggest changing the cookie timeout to be: `min(cookie_timeout, access_token_timeout)` to avoid sending out cookie that can outlast the access token.

yili-db:
Please call out the behavior change of changing from never expiring token to 1 day.

server/src/test/java/io/unitycatalog/server/service/AuthServiceTest.java
```diff
@@ -129,6 +130,8 @@ public void testTokenExchangeWithCorrectIssuerAndAudience() throws IOException {
     assertThat(body.get("issued_token_type").asText())
         .isEqualTo("urn:ietf:params:oauth:token-type:access_token");
     assertThat(body.get("token_type").asText()).isEqualTo("Bearer");
+    assertThat(body.get("expires_in").asLong()).isEqualTo(Duration.parse("PT24H").getSeconds());
+    assertThat(JWT.decode(body.get("access_token").asText()).getExpiresAt()).isNotNull();
```
yili-db: should validate that the expiration is expected `iat + ttl`. If changing wall clock is an issue, You can use a time range to validate.

yili-db: Similar to `// Logout without cookie should return 401 (no credentials provided)` can you also try to send any request with an expired token to make sure the expiration does work?

etc/conf/server.properties
s-maling-celonis: Done in f9f18260. Added `ServerProperties.getEffectiveCookieTimeout()` which returns `min(server.cookie-timeout, server.access-token-timeout)`, and `AuthService` now uses that when setting the `UC_TOKEN` cookie via `ext=cookie`. With the current defaults (`P5D` / `PT24H`), the cookie max-age is capped to 24h. Added `testEffectiveCookieTimeout` to cover both directions of the min.

server/src/test/java/io/unitycatalog/server/service/AuthServiceTest.java
s-maling-celonis: Done in d7e8d1af. `testTokenExchangeWithCorrectIssuerAndAudience` now decodes the access token and asserts `exp` is close to `iat + PT24H` (within 2 seconds), in addition to checking `expires_in` matches the configured TTL.

s-maling-celonis: Done in 6abf0c70. Added `testExpiredAccessTokenIsRejected`, which sends a signed UC access token with a past `exp` claim as a `UC_TOKEN` cookie to `/auth/logout` and asserts the server returns 401.

s-maling-celonis:
Documented in aaf319e0 and the updated PR description. UC access tokens now expire after 24 hours by default (`server.access-token-timeout=PT24H`); previously they had no `exp` claim. Config comments, Helm `auth.accessTokenTimeout`, and the PR "Behavior change" section describe the migration impact and how to tune the timeout.

yili-db (approved):
Nice and clean work. Thank you!

yili-db:
@tdas Please take a look.

tdas (approved):
LGTM. the fixes on @yili-db comment looks good too. thanks for those fixes.

PR #1690, opened 2026-07-20
Support wildcard patterns in server.allowed-issuers and server.audiences

s-maling-celonis:
## Summary

Adds wildcard matching for OAuth issuer and audience allowlists in multi-tenant deployments:

- **`WildcardAllowlist`** — shared matcher for exact entries and single-DNS-label `*` wildcards
- **`ServerProperties`** — volatile cached `getIssuerAllowlist()` / `getAudienceAllowlist()` getters that rebuild when the raw property value changes
- **`AuthService`** — validates issuer (before JWKS fetch) and audience during token exchange via those allowlists
- **`server.audiences=*`** alone disables audience validation for dynamic OAuth client UUID audiences

Example configuration:

```properties
server.allowed-issuers=https://*.dev.example.com
server.audiences=unity-catalog-local,https://*.dev.example.com
```

## Test plan

- [ ] `sbt "server/testOnly io.unitycatalog.server.utils.WildcardAllowlistTest io.unitycatalog.server.utils.ServerPropertiesTest io.unitycatalog.server.service.AuthServiceWildcardIssuerTest io.unitycatalog.server.service.AuthServiceWildcardAudienceTest io.unitycatalog.server.service.AuthServiceWildcardAudiencePatternTest io.unitycatalog.server.service.AuthServiceAllowlistConfigurationTest io.unitycatalog.server.service.AuthServiceTest"`
- [ ] Exact and wildcard issuer/audience entries match expected tokens
- [ ] Lone `*` audience entry disables audience validation
- [ ] Missing `server.allowed-issuers` / `server.audiences` returns 400 with clear error message


server/src/main/java/io/unitycatalog/server/utils/IssuerAllowlist.java
```diff
@@ -0,0 +1,52 @@
+package io.unitycatalog.server.utils;
+
+import java.util.List;
+import java.util.regex.Pattern;
+
+/** Matches token issuers against configured allowlist entries (exact or wildcard). */
+public final class IssuerAllowlist {
```
yili-db: Is this class a general wildcard pattern allowlist matching utility?
I see that it has nothing special to **issuer**. And it's used by AudienceAllowlist.
OTOH, it's not wired in for issuer matching.

server/src/main/java/io/unitycatalog/server/utils/IssuerAllowlist.java
```diff
@@ -0,0 +1,52 @@
...
+      if (c == '*') {
+        regex.append("[^./]+");
+      } else if ("\\.[]{}()+-^$|?".indexOf(c) >= 0) {
+        regex.append('\\').append(c);
+      } else {
+        regex.append(c);
+      }
+    }
+    regex.append('$');
+    return Pattern.compile(regex.toString());
```
yili-db: This happens for every single RPC call.
Consider make this class non-static, precompile, and save it as it constructs.
Then this class can be plugged into `ServerProperties.getAudiences` and the compiled regex can be cached in there. In ServerProperties things aren't guaranteed to remain unchanged, so it's better to employ a "one entry cache".

For example:
```
public final class WildcardAllowlist {
    private final String source;  <--- this is the raw string in `server.properties`
    private final List<String> exact = new ArrayList<>();  <--- exact match items
    private final List<Pattern> wildcards = new ArrayList<>();  <--- wildcard items
...
// Then in ServerProperties
  private volatile WildcardAllowlist cachedAudienceAllowlist; // one single entry to be cached

  public WildcardAllowlist getAudienceAllowlist() {
    String current = getProperty("server.audiences");          // live read
    WildcardAllowlist cached = cachedAudienceAllowlist;      // single volatile read
    if (cached == null || !cached.source.equals(current)) {
      cached = new WildcardAllowlist(current);
      cachedAudienceAllowlist = cached;                    // atomic publish
    }
    return cached;
  }
```

server/src/main/java/io/unitycatalog/server/utils/IssuerAllowlist.java
s-maling-celonis: Good catch — you're right on all three points about the original `IssuerAllowlist`.

**Issuer matching was a regression.** In the original combined PR (#1661), `AuthService` validated the token issuer against `server.allowed-issuers` before fetching JWKS. When the work was split into separate audience/issuer PRs, wildcard audience matching landed in `AuthService`, but issuer allowlist wiring did not — so `server.allowed-issuers` wildcards were parsed/configured but never enforced at token exchange. That is restored now.

**Changes made in the latest commit:**

1. **Renamed to a generic matcher** — `IssuerAllowlist` / `AudienceAllowlist` are replaced by `WildcardAllowlist`, a shared utility for exact + single-label `*` wildcard matching. Factory methods `forAllowedIssuers()` / `forAudiences()` carry the property-specific names for error messages.

2. **Issuer matching wired in `AuthService`** — `grantToken()` now calls `serverProperties.getIssuerAllowlist().isAllowed(issuer)` **before** JWKS lookup, same as pre-split behavior but with wildcard support.

3. **Precompiled patterns via `ServerProperties`** — `getIssuerAllowlist()` / `getAudienceAllowlist()` cache compiled allowlists keyed on the raw property string and rebuild only when the config changes.

4. **Tests** — added `AuthServiceWildcardIssuerTest` (integration) and `AuthServiceAllowlistConfigurationTest` (missing config → 400); unit tests in `WildcardAllowlistTest`.

server/src/main/java/io/unitycatalog/server/utils/IssuerAllowlist.java
s-maling-celonis: Thanks — implemented as suggested. Summary of what changed:

**Precompile at construction, not per RPC**

`WildcardAllowlist` now parses the raw CSV property value in its constructor and stores precompiled state: a `Set<String>` for exact entries and a `List<Pattern>` for wildcard entries. Matching is just set lookup + pattern checks — no regex compilation on the request path.

**One-entry cache in `ServerProperties`**

Added volatile cached fields and getters for both properties:

- `getIssuerAllowlist()` — `server.allowed-issuers`
- `getAudienceAllowlist()` — `server.audiences`

Each getter live-reads the raw property, compares it to `cached.source()`, and rebuilds only when the string changes. Steady state is a volatile read + string equality check.

**`source` tracks the raw config string**

`WildcardAllowlist` stores the raw `server.properties` value as `source` (via `source()`) specifically for cache invalidation, as in your example.

**`AuthService` uses the cached getters**

`grantToken()` calls `serverProperties.getIssuerAllowlist()` / `getAudienceAllowlist()` rather than building allowlists itself, so compilation happens once per distinct config value, not once per token exchange after the first build.

Also added `ServerPropertiesTest.testAllowlistCaching` and `testAllowlistRebuildsWhenPropertyChanges` to cover the cache hit and rebuild paths.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -511,28 +522,74 @@ public void checkDeltaApiOnlyEnabled(String deltaEndpoint) {
...
+   * #getAllowedIssuers()}).
+   *
+   * <p>A single entry of {@code *} disables audience validation (issuer and user checks still
+   * apply). That sentinel cannot be combined with other values.
    *
    * @return List of expected audience values
    */
   public List<String> getAudiences() {
-    return getCommaSeparatedList("server.audiences");
+    return getCommaSeparatedList(WildcardAllowlist.AUDIENCES_PROPERTY);
```
yili-db: This is a pre-existing problem. Please define `AUDIENCES_PROPERTY` as a enum (without `_PROPERTY`) in `public enum Property {` in this file.
Same for `ALLOWED_ISSUERS_PROPERTY`

etc/conf/server.properties
```diff
@@ -15,13 +15,14 @@ server.client-id=
 server.client-secret=
 server.redirect-port=
 
-# Required when authorization is enabled: Comma-separated list of allowed token issuers (exact match)
-# Tokens from issuers not in this list will be rejected
+# Required when authorization is enabled: Comma-separated list of allowed token issuers
+# (exact match or wildcard with *). Tokens from issuers not in this list will be rejected
```
yili-db: Comment of `single DNS label per wildcard` is missing from this file.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -177,14 +163,20 @@ public HttpResponse grantToken(
...
       decodedJWT = jwtVerifier.verify(decodedJWT);
     } catch (JWTVerificationException e) {
       LOGGER.debug("Token rejected: verification failed", e);
       throw new OAuthInvalidRequestException(
           ErrorCode.UNAUTHENTICATED, "Token verification failed: " + e.getMessage(), e);
     }
 
+    if (!serverProperties.isAudienceValidationDisabled()
+        && !serverProperties.getAudienceAllowlist().isAnyAllowed(decodedJWT.getAudience())) {
+      LOGGER.debug("Token rejected: audience not in allowlist");
```
yili-db: Better print out the audience in this debug log to aid debugging.

server/src/main/java/io/unitycatalog/server/utils/WildcardAllowlist.java
```diff
@@ -0,0 +1,126 @@
...
+      if (pattern.indexOf('*') >= 0) {
+        wildcardEntries.add(toRegex(pattern));
+      } else {
+        exactEntries.add(pattern);
+      }
+    }
+    if (exactEntries.isEmpty() && wildcardEntries.isEmpty()) {
+      throw misconfigured(allowlistName, propertyKey);
+    }
+    this.exact = exactEntries;
```
yili-db: `List.copyOf` looks good. This `exactEntries` looks also need a `Set.copyOf`

server/src/test/java/io/unitycatalog/server/service/AuthServiceAllowlistConfigurationTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+/** Integration tests for issuer/audience allowlist configuration during token exchange. */
+public class AuthServiceAllowlistConfigurationTest {
+
+  private static final String TOKEN_ENDPOINT = "/api/1.0/unity-control/auth/tokens";
+
+  abstract static class TokenExchangeTestBase extends BaseAuthCRUDTest {
```
yili-db: It looks like these 4 unit tests can be consolidated and deduped into one single file so that they share most of the setup. The only thing that would differ on setup is the issuer/audience allowlists.

So it seems that all tests that share the same config can share the same class as a subclass of TokenExchangeTestBase.

Also all tests exercising the same setup can be made into one test method to save on setup/teardown cost.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
s-maling-celonis: Done in ee9c6a87 — added ALLOWED_ISSUERS and AUDIENCES to ServerProperties.Property and switched all call sites to use Property.*.key instead of string constants in WildcardAllowlist.

etc/conf/server.properties
s-maling-celonis: Done in ee9c6a87 — added the single DNS label per wildcard note to both server.allowed-issuers and server.audiences comments in etc/conf/server.properties.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
s-maling-celonis: Done in ee9c6a87 — the debug log now includes the token audience: Token rejected: audience {} not in allowlist.

server/src/main/java/io/unitycatalog/server/utils/WildcardAllowlist.java
s-maling-celonis: Done in ee9c6a87 — exact is now assigned via Set.copyOf(exactEntries).

server/src/test/java/io/unitycatalog/server/service/AuthServiceAllowlistConfigurationTest.java
s-maling-celonis: Done in ee9c6a87 — consolidated the four integration test classes into AuthServiceAllowlistTest with shared TokenExchangeTestBase and one test method per config (@Nested subclasses).

server/src/main/java/io/unitycatalog/server/utils/IssuerAllowlist.java
s-maling-celonis: Addressed in prior commits — IssuerAllowlist was replaced by shared WildcardAllowlist, issuer matching is wired in AuthService, and both issuer/audience allowlists use the cached getters.

server/src/main/java/io/unitycatalog/server/utils/IssuerAllowlist.java
s-maling-celonis: Addressed in prior commits — patterns are precompiled in WildcardAllowlist at construction and cached via volatile one-entry caches in ServerProperties.getIssuerAllowlist() / getAudienceAllowlist().

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -177,14 +163,20 @@ public HttpResponse grantToken(
 
     try {
       JWTVerifier jwtVerifier =
-          jwksOperations.verifierForIssuerAndKey(issuer, keyId, alg, audiences);
+          jwksOperations.verifierForIssuerAndKey(issuer, keyId, alg, List.of());
```
yili-db: The `audiences` parameter of `verifierForIssuerAndKey` becomes dead code after this PR.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -268,6 +270,17 @@ private void validateProperties() {
       }
       property.validator.validate(property.key, value);
     }
+    validateAudienceConfiguration();
+  }
+
+  private void validateAudienceConfiguration() {
```
yili-db: Similar to this function, `server.allowed-issuers` should be check that it doesn't contain any lone *, especially `server.allowed-issuers=*`. Because it doesn't do what users would usually expect: `server.audiences` accepts `*` to disable check but `server.allowed-issuers` doesn't.

docs/server/auth.md
```diff
@@ -67,6 +67,29 @@ server.allowed-issuers=https://accounts.google.com,https://login.microsoftonline
 server.audiences=your-google-client-id,your-azure-client-id
 ```
 
+Wildcard issuers match a single DNS label per `*`:
+
+```properties
+server.allowed-issuers=https://*.dev.example.com
+```
+
+Wildcard audience patterns use the same `*` rules. For example, `https://*.dev.example.com`
```
yili-db, line 76: One thing worth calling out in the docs: with exact-match issuers the server only ever fetches JWKS from a fixed list of URLs. A wildcard like `https://*.dev.example.com` lets an unauthenticated caller pick which host we fetch from (the issuer check runs before signature verification), so it's effectively an SSRF surface.

Can you please add a note that wildcard issuers should only be used where necessary and only cover domains with no internal/sensitive hosts, and prefer `https://`? Not blocking, but people will copy these examples.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
s-maling-celonis: Done in 62da767f — removed the `audiences` parameter from `JwksOperations.verifierForIssuerAndKey`; audience validation stays in `AuthService` via `WildcardAllowlist` after signature verification.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
s-maling-celonis: Done in 62da767f — `validateAuthAllowlistConfiguration()` now rejects `server.allowed-issuers=*` (and any comma-separated entry that is exactly `*`) at startup with a clear error message. Added `ServerPropertiesTest.testRejectsLoneWildcardAllowedIssuers`.

docs/server/auth.md
s-maling-celonis: Done in 62da767f — added a **Wildcard issuer security** subsection to `docs/server/auth.md` covering JWKS fetch before signature verification and SSRF-style considerations, plus brief pointers in `etc/conf/server.properties` and Helm `values.yaml`.

yili-db (approved):

yili-db:
@tdas please take a look.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -165,7 +150,7 @@ public HttpResponse grantToken(
     String issuer = decodedJWT.getIssuer();
 
     // Validate issuer is in allowlist BEFORE fetching JWKS
-    if (!allowedIssuers.contains(issuer)) {
+    if (!serverProperties.getIssuerAllowlist().isAllowed(issuer)) {
       LOGGER.debug("Token rejected: invalid issuer '{}'", issuer);
```
tdas: shouldnt this be error logging?  
all error situations where some RPC has failed... should have error logging.. because in production where people will be running in warn or above, those logs should always be visible.

tdas: please make sure that is the case else where in this file as well.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -176,15 +161,20 @@ public HttpResponse grantToken(
...
       decodedJWT = jwtVerifier.verify(decodedJWT);
     } catch (JWTVerificationException e) {
       LOGGER.debug("Token rejected: verification failed", e);
       throw new OAuthInvalidRequestException(
           ErrorCode.UNAUTHENTICATED, "Token verification failed: " + e.getMessage(), e);
     }
 
+    if (!serverProperties.isAudienceValidationDisabled()
+        && !serverProperties.getAudienceAllowlist().isAnyAllowed(decodedJWT.getAudience())) {
+      LOGGER.debug("Token rejected: audience {} not in allowlist", decodedJWT.getAudience());
```
tdas: same comment as above.. .this should be error logging.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -176,15 +161,20 @@ public HttpResponse grantToken(
     LOGGER.debug("Validating token for issuer: {} and keyId: {}", issuer, keyId);
 
     try {
-      JWTVerifier jwtVerifier =
-          jwksOperations.verifierForIssuerAndKey(issuer, keyId, alg, audiences);
+      JWTVerifier jwtVerifier = jwksOperations.verifierForIssuerAndKey(issuer, keyId, alg);
       decodedJWT = jwtVerifier.verify(decodedJWT);
     } catch (JWTVerificationException e) {
       LOGGER.debug("Token rejected: verification failed", e);
```
tdas: same as above,

tdas (approved):
overall looks good but it would be good to fix the logging as i have commented before merging.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
s-maling-celonis: Fixed in 0ee25635 — all four token-rejection paths in `grantToken()` now use `LOGGER.error` (malformed token, invalid issuer, verification failed, audience not in allowlist). Success/diagnostic lines remain at debug.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
s-maling-celonis: Fixed in 0ee25635 — audience rejection now logs at error level.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
s-maling-celonis: Fixed in 0ee25635 — signature verification failure now logs at error level.

s-maling-celonis:
Thanks @tdas and @yili-db for the thorough review — all feedback is addressed now.

**Logging (tdas):** `0ee25635` promotes the four token-rejection logs in `AuthService.grantToken()` from `debug` to `error` (malformed token, invalid issuer, verification failed, audience not in allowlist).

**Earlier review items (yili-db):** issuer/audience wildcard matching, caching, config validation, docs, and test consolidation were addressed in prior commits on this branch.

Rebased onto latest `upstream/main`. Ready for another look when you have a moment.

PR #1691, opened 2026-07-20
Resolve token exchange principals from subject token claims

s-maling-celonis:
## Summary

This PR contains subject-token principal resolution for `POST /auth/tokens`.

UC resolves the principal from the **subject token** — no OAuth client credentials on the UC request:

1. **`id_token`** (CLI/browser): `email` claim or `sub` → `getUserByEmail`, then `externalId` fallback via `azp` or `client_id`
2. **`access_token`** (M2M): if no `email` claim, `externalId` via `azp`/`client_id` first, then `sub`; otherwise same as `id_token`

The issued UC access token always uses the resolved user's **email** as `sub` (for grants/authorization).

Additional behavior:

- **Audience bypass**: signed subject tokens whose `azp` or `client_id` matches a registered UC user `externalId` are accepted even when the client UUID is not in `server.audiences`
- **Unique index** on `uc_users.external_id` (multiple `NULL` values still allowed)
- Prefer the `externalId` flow over `server.audiences=*` for per-tenant service principals

Bootstrap UC users with `email` for grants and `externalId` = OAuth client id (same value as the token's `azp` or `client_id` claim).

## Test plan

- [x] `build/sbt "server/testOnly io.unitycatalog.server.service.TokenExchangePrincipalResolverTest io.unitycatalog.server.service.AuthServiceClientPrincipalTest io.unitycatalog.server.persist.UserRepositoryTest io.unitycatalog.server.service.AuthServiceTest io.unitycatalog.server.service.AuthServiceAllowlistTest"` (24/24 passed locally)
- [x] M2M/access_token flow resolves service principal via `externalId` (`AuthServiceClientPrincipalTest`)
- [x] Issued UC token `sub` claim is the resolved user email (`AuthServiceClientPrincipalTest`)
- [x] `extractOAuthClientId` reads `azp` then `client_id` (`TokenExchangePrincipalResolverTest`)
- [ ] Browser/id_token flow resolves user by email (manual)
- [ ] Full CI (Unit Tests workflow on fork PR)

### DIFF
--- a/docs/server/auth.md
+++ b/docs/server/auth.md
@@ -56,7 +56,18 @@ server.audiences=<Client ID provided earlier>
 When authorization is enabled, the server validates incoming identity tokens against configured issuers and audiences:
 
 - **server.allowed-issuers**: Comma-separated list of allowed token issuers (exact match or wildcard with `*`). Tokens from issuers not in this list will be rejected. This prevents attackers from using their own identity provider to forge tokens.
-- **server.audiences**: Comma-separated list of expected JWT audience values. Tokens must contain an `aud` value matching one of these entries (exact match or wildcard with `*`). A single value of `*` disables audience validation (issuer and user checks still apply); that sentinel cannot be combined with other values.
+- **server.audiences**: Comma-separated list of expected JWT audience values. Tokens must contain an `aud` value matching one of these entries (exact match or wildcard with `*`). A single value of `*` disables audience validation (issuer and user checks still apply); that sentinel cannot be combined with other values. Tokens whose `azp` or non-URL `aud` matches a registered UC user's `externalId` are also accepted even when the client UUID is not listed in `server.audiences`.
+
+#### Programmatic exchange (service principals)
+
+Service integrations exchange an upstream OAuth access token on `POST /auth/tokens` without client credentials on the UC request. UC resolves the principal in two steps:
+
+1. **Email mode** — look up the user by the token `email` claim (or `sub` fallback for `id_token`).
+2. **External id mode** — look up the user by OAuth client id from the subject token (`azp`, or the first non-URL `aud` value) mapped to `externalId`. For `access_token` subjects without an `email` claim, this is tried before `sub`.
+
+Create UC users with a human-readable `email` for grants and set `externalId` to the OAuth client id for programmatic exchange. The issued UC access token always uses the resolved user's `email`.
+
+CLI and browser login flows use email mode only.
 
 #### Multiple Identity Providers
 
@@ -81,9 +92,10 @@ accepts tokens whose `aud` includes `https://dev.dev.example.com` or
 server.audiences=unity-catalog-local,https://*.dev.example.com
 ```
 
-When the identity provider issues per-client UUID audiences that cannot be pre-listed (for example
-dynamic OAuth client IDs in `id_token.aud`), use `*` alone to skip audience validation while still
-enforcing issuer and user checks:
+When email-based exchange cannot list dynamic OAuth client UUID audiences, tokens whose `azp` or
+non-URL `aud` matches a registered UC user's `externalId` are accepted even when the client UUID is
+not listed in `server.audiences`. As a fallback for human login only, `*` alone skips audience
+allowlist validation:
 
 ```properties
 server.audiences=*
@@ -100,6 +112,7 @@ values under that pattern can steer JWKS fetches (an SSRF-style surface).
 
 Use wildcard issuers only where necessary, only for domains you control, and avoid patterns that
 cover internal or sensitive hosts. Prefer exact issuer URLs and `https://` issuers when possible.
+Prefer the subject-token `externalId` flow for per-tenant service principals instead of `audiences=*`.
 
 ### Restart the UC Server
 
--- a/docs/server/auth.md
+++ b/docs/server/auth.md
@@ -56,7 +56,18 @@ server.audiences=<Client ID provided earlier>
 When authorization is enabled, the server validates incoming identity tokens against configured issuers and audiences:
 
 - **server.allowed-issuers**: Comma-separated list of allowed token issuers (exact match or wildcard with `*`). Tokens from issuers not in this list will be rejected. This prevents attackers from using their own identity provider to forge tokens.
-- **server.audiences**: Comma-separated list of expected JWT audience values. Tokens must contain an `aud` value matching one of these entries (exact match or wildcard with `*`). A single value of `*` disables audience validation (issuer and user checks still apply); that sentinel cannot be combined with other values.
+- **server.audiences**: Comma-separated list of expected JWT audience values. Tokens must contain an `aud` value matching one of these entries (exact match or wildcard with `*`). A single value of `*` disables audience validation (issuer and user checks still apply); that sentinel cannot be combined with other values. Tokens whose `azp` or non-URL `aud` matches a registered UC user's `externalId` are also accepted even when the client UUID is not listed in `server.audiences`.
+
+#### Programmatic exchange (service principals)
+
+Service integrations exchange an upstream OAuth access token on `POST /auth/tokens` without client credentials on the UC request. UC resolves the principal in two steps:
+
+1. **Email mode** — look up the user by the token `email` claim (or `sub` fallback for `id_token`).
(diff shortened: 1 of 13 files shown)


docs/server/auth.md
```diff
@@ -56,7 +56,18 @@ server.audiences=<Client ID provided earlier>
 When authorization is enabled, the server validates incoming identity tokens against configured issuers and audiences:
 
 - **server.allowed-issuers**: Comma-separated list of allowed token issuers (exact match or wildcard with `*`). Tokens from issuers not in this list will be rejected. This prevents attackers from using their own identity provider to forge tokens.
-- **server.audiences**: Comma-separated list of expected JWT audience values. Tokens must contain an `aud` value matching one of these entries (exact match or wildcard with `*`). A single value of `*` disables audience validation (issuer and user checks still apply); that sentinel cannot be combined with other values.
+- **server.audiences**: Comma-separated list of expected JWT audience values. Tokens must contain an `aud` value matching one of these entries (exact match or wildcard with `*`). A single value of `*` disables audience validation (issuer and user checks still apply); that sentinel cannot be combined with other values. Tokens whose `azp` or non-URL `aud` matches a registered UC user's `externalId` are also accepted even when the client UUID is not listed in `server.audiences`.
```
yili-db: It seems that user's externalId should come from either `azp` or `client_id`, but never `aud`.
`aud` is meant for specifying "the receiving service that needs to validate user identity", which is UC in this case.
And often it's attacker-influenceable.

yili-db (changes requested):

docs/server/auth.md
s-maling-celonis: Done in edcac20f — `extractOAuthClientId` now reads `azp` then the `client_id` JWT claim only; `aud` is no longer used for `externalId` lookup (it stays in audience validation). Docs and tests updated; 24/24 targeted server tests passed locally.

s-maling-celonis: @yili-db Addressed the review feedback in edcac20f:

- `externalId` principal lookup uses subject-token `azp`, then `client_id` — never `aud`
- `aud` remains for audience validation only (`server.audiences`)
- Docs/javadoc and tests updated accordingly

**Tests:** 24/24 targeted server tests passed locally (`TokenExchangePrincipalResolverTest`, `AuthServiceClientPrincipalTest`, `UserRepositoryTest`, `AuthServiceTest`, `AuthServiceAllowlistTest`).

server/src/main/java/io/unitycatalog/server/security/SecurityContext.java
yili-db: This becomes dead code?

server/src/main/java/io/unitycatalog/server/service/TokenExchangeSubjectTokenHandler.java
```diff
@@ -0,0 +1,222 @@
...
+
+    if (audiences.isEmpty() && tokenClientId.isEmpty()) {
+      LOGGER.error("No audiences configured");
+      throw new OAuthInvalidRequestException(
+          ErrorCode.INVALID_ARGUMENT,
+          "No audiences configured. Set server.audiences in server.properties");
+    }
+
+    if (audiences.isEmpty()) {
+      if (tokenClientId.isPresent()
```
yili-db: This means that for a  matching `azp`/`client_id` tokens, audience allowlist is bypassed? May I ask why? It seems to defeat the purpose of audience allowlist.
If audience is truly unpredictable, the wildcard `*` audience can already disable it.

s-maling-celonis: Good catch — that bypass was intentional for dynamic client UUIDs, but you're right that it weakens the audience allowlist.

Fixed in 909af1d3: audience validation now uses `server.audiences` only (`aud` match or `*`). `azp` / `client_id` are used only for principal resolution via `externalId`, not to skip the allowlist. Docs and tests updated accordingly (including a case that rejects a mismatched `aud` even when `externalId` matches).

server/src/main/java/io/unitycatalog/server/security/SecurityContext.java
s-maling-celonis: Yes — after switching issuance to the resolved principal email, that overload was unused. Removed in 909af1d3.

server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -168,18 +174,14 @@ public HttpResponse grantToken(
           ErrorCode.UNAUTHENTICATED, "Token verification failed: " + e.getMessage(), e);
     }
 
-    if (!serverProperties.isAudienceValidationDisabled()
```
yili-db, line 175: The issuer validation remains in this function but audience validation moves to `resolvePrincipalEmail`?
Earlier in this PR you had audience validation conditional on if it's picking up email or id. Now that is no longer true. I think audience validation should be restored here to remain together with the other verification.

Then the `resolvePrincipalEmail` just do one thing: extract the email from a JWT token that is **already validated and trusted**.

s-maling-celonis: Done — audience validation is back in `grantToken` with issuer and signature checks. `resolvePrincipalEmail` only extracts the UC email from the already-validated token (`bb8ccd59`).

yili-db (approved):

server/src/main/java/io/unitycatalog/server/service/TokenExchangeSubjectTokenHandler.java
```diff
@@ -0,0 +1,162 @@
...
+
+    Claim clientId = decodedJWT.getClaim("client_id");
+    if (!clientId.isNull()) {
+      String clientIdValue = clientId.asString();
+      if (clientIdValue != null && !clientIdValue.isBlank()) {
+        return Optional.of(clientIdValue);
+      }
+    }
+
+    return Optional.empty();
```
tdas, line 96: [AI-assisted]

we should never eat this without logging. if the lookup fails, this turns a db/server issue into `User not allowed`, doesnt it? same below for email lookup.

future operational failures become indistinguishable from auth denial, and debugging token exchange gets much harder.

server/src/main/java/io/unitycatalog/server/service/TokenExchangeSubjectTokenHandler.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  /**
+   * Returns the UC principal email for the issued access token.
+   *
+   * <p>The subject token must already be issuer-, signature-, and audience-validated. For {@code
+   * id_token} subjects, resolution order is email (or {@code sub}) then OAuth client id from {@code
+   * azp} or {@code client_id} mapped to {@code externalId}. For {@code access_token} subjects
+   * without an {@code email} claim, {@code externalId} is tried before {@code sub}.
+   */
```
tdas, line 37: [AI-assisted]

blocker: why can an id_token fall back to azp/client_id? for browser tokens azp is the app, not the human, so this can issue a service principal token for the wrong user, right?

future cli/browser tokens that dont resolve by email could inherit grants from whichever UC user has externalId equal to the oauth client id.

tdas (changes requested):
couple of issues found.

server/src/main/java/io/unitycatalog/server/service/TokenExchangeSubjectTokenHandler.java
s-maling-celonis: Agreed — swallowing the lookup turned DB/server failures into `User not allowed`.

`dc54e569` only treats `NOT_FOUND` as a miss. Any other `BaseException` or runtime error is logged and rethrown (email and `externalId` lookups share that path).

server/src/main/java/io/unitycatalog/server/service/TokenExchangeSubjectTokenHandler.java
s-maling-celonis: Yes — for an `id_token`, `azp`/`client_id` is the app, not the human.

`dc54e569` drops that fallback: `id_token` resolves only from `email`/`sub`. `azp`/`client_id` → `externalId` remains for M2M `access_token` subjects. Covered by `idTokenDoesNotResolveViaOauthClientId` and `testIdTokenExchangeDoesNotFallBackToExternalId`.

tdas (approved):

PR #1692, opened 2026-07-20
refactor(hadoop): extract cloud credential property builders

foss-contributor:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1692/files) to review incremental changes.
- [**stack/credential-property-builders**](https://github.com/unitycatalog/unitycatalog/pull/1692) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1692/files)] ← _this PR_
  - [stack/normalize-cloud-storage-prefixes](https://github.com/unitycatalog/unitycatalog/pull/1693) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1693/files/082ff24468a7a1440012466f717219832a012c36..607ecf0161b3c1289ab89c48d4bd4587e1df1075)]
    - [stack/location-aware-delegate-cache-keys](https://github.com/unitycatalog/unitycatalog/pull/1695) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1695/files/607ecf0161b3c1289ab89c48d4bd4587e1df1075..8ea3692a327ccb67f7f380ef8a2e57a4984dc8d4)]
      - [stack/credential-location-identity](https://github.com/unitycatalog/unitycatalog/pull/1699) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1699/files/8ea3692a327ccb67f7f380ef8a2e57a4984dc8d4..9678b39a8b80f2cc49c501862f2a588465841d98)]
        - [stack/select-vended-credential-by-location](https://github.com/unitycatalog/unitycatalog/pull/1701) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1701/files/9678b39a8b80f2cc49c501862f2a588465841d98..aca6d1deba11a8e63a91443c7802dc9404d7b488)]
          - [stack/scoped-credential-namespaces](https://github.com/unitycatalog/unitycatalog/pull/1704) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1704/files/aca6d1deba11a8e63a91443c7802dc9404d7b488..03a9c5a58b4c484148abce4151a08679c00f9da7)]
            - [stack/encode-scoped-vended-credentials](https://github.com/unitycatalog/unitycatalog/pull/1705) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1705/files/03a9c5a58b4c484148abce4151a08679c00f9da7..775af5e9e15df81804d7575994444222ef44a9d3)]

---------
## Summary

- Extract provider-specific Hadoop credential-property construction from `CredPropsUtil` into dedicated S3, GCS, and ABFS builders.
- Select the builder through `CloudType` while preserving static and renewable credential wiring, credential-scoped filesystem overrides, and GCS configuration behavior.

## Issue

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
@@ -5,6 +5,8 @@ import io.unitycatalog.client.internal.Preconditions;
 import io.unitycatalog.hadoop.internal.CloudType;
 import io.unitycatalog.hadoop.internal.UCHadoopConfConstants;
 import io.unitycatalog.hadoop.internal.auth.GenericCredential;
+import io.unitycatalog.hadoop.internal.fs.CredScopedFileSystem;
+import io.unitycatalog.hadoop.internal.fs.CredScopedFs;
 import io.unitycatalog.hadoop.internal.id.CredId;
 import java.util.Collections;
 import java.util.HashMap;
@@ -14,11 +16,9 @@ import org.apache.hadoop.conf.Configuration;
 /** Builds the cloud-provider specific Hadoop configuration credential properties. */
 public abstract class CredPropsBuilder {
   /** CredScopedFileSystem implementation classes, swapped in when cred-scoped FS is enabled. */
-  protected static final String CRED_SCOPED_FS_CLASS =
-      "io.unitycatalog.hadoop.internal.fs.CredScopedFileSystem";
+  protected static final String CRED_SCOPED_FS_CLASS = CredScopedFileSystem.class.getName();
 
-  protected static final String CRED_SCOPED_AFS_CLASS =
-      "io.unitycatalog.hadoop.internal.fs.CredScopedFs";
+  protected static final String CRED_SCOPED_AFS_CLASS = CredScopedFs.class.getName();
 
   private final Configuration hadoopConf;
   private final HashMap<String, String> props = new HashMap<>();
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
@@ -1,0 +1,125 @@
+package io.unitycatalog.hadoop.internal.props;
+
+import io.unitycatalog.client.auth.TokenProvider;
+import io.unitycatalog.client.internal.Preconditions;
+import io.unitycatalog.hadoop.internal.CloudType;
+import io.unitycatalog.hadoop.internal.UCHadoopConfConstants;
+import io.unitycatalog.hadoop.internal.auth.GenericCredential;
+import io.unitycatalog.hadoop.internal.id.CredId;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.hadoop.conf.Configuration;
+
+/** Builds the cloud-provider specific Hadoop configuration credential properties. */
+public abstract class CredPropsBuilder {
+  /** CredScopedFileSystem implementation classes, swapped in when cred-scoped FS is enabled. */
+  protected static final String CRED_SCOPED_FS_CLASS =
+      "io.unitycatalog.hadoop.internal.fs.CredScopedFileSystem";
+
+  protected static final String CRED_SCOPED_AFS_CLASS =
+      "io.unitycatalog.hadoop.internal.fs.CredScopedFs";
+
+  private final Configuration hadoopConf;
+  private final HashMap<String, String> props = new HashMap<>();
+
+  private boolean renewCredEnabled;
+  private GenericCredential initialCredential;
+
+  protected CredPropsBuilder(Configuration hadoopConf) {
+    Preconditions.checkNotNull(hadoopConf, "hadoopConf is required");
+    this.hadoopConf = hadoopConf;
+  }
+
+  /** Returns a builder for {@code cloudType}, seeded with that cloud's static base properties. */
+  public static CredPropsBuilder forCloud(CloudType cloudType, Configuration hadoopConf) {
+    switch (cloudType) {
+      case S3:
+        return new S3CredPropsBuilder(hadoopConf);
+      case GCS:
+        return new GcsCredPropsBuilder(hadoopConf);
+      case ABFS:
+        return new AbfsCredPropsBuilder(hadoopConf);
+    }
+    throw new IllegalStateException("Unhandled cloud type: " + cloudType);
+  }
+
+  /**
+   * Records whether renewal is enabled and, when it is, writes the wiring the vended token provider
+   * needs to renew the credential: the provider-class keys, the catalog uri, the auth configs, the
+   * credential-scope identity, and the engine versions. The fixed path writes none of this.
+   */
+  public CredPropsBuilder renewCredEnabled(
+      boolean renewCredEnabled,
+      String catalogUri,
+      TokenProvider tokenProvider,
+      CredId credId,
+      Map<String, String> appVersions) {
+    this.renewCredEnabled = renewCredEnabled;
+    if (renewCredEnabled) {
+      applyVendedProviderKeys();
+      set(UCHadoopConfConstants.UC_URI_KEY, catalogUri);
+      // Only 'fs.*' properties propagate to the FileSystem, so prefix the auth configs.
+      tokenProvider
+          .configs()
+          .forEach((key, value) -> set(UCHadoopConfConstants.UC_AUTH_PREFIX + key, value));
+      credId.props().forEach(this::set);
+      appVersions.forEach(
+          (key, value) -> set(UCHadoopConfConstants.UC_ENGINE_VERSION_PREFIX + key, value));
+    }
+    return this;
+  }
+
+  /** When enabled, writes the cloud filesystem-impl overrides that install CredScopedFileSystem. */
+  public CredPropsBuilder credScopedFsEnabled(boolean credScopedFsEnabled) {
+    if (credScopedFsEnabled) {
+      writeImplOverrides();
+    }
+    return this;
+  }
+
+  /** Records the fetched credential whose secrets are written by {@link #build()}. */
+  public CredPropsBuilder initialCredential(GenericCredential initialCredential) {
+    this.initialCredential = initialCredential;
+    return this;
+  }
+
+  public Map<String, String> build() {
+    Preconditions.checkNotNull(initialCredential, "initialCredential is required");
+    if (renewCredEnabled) {
+      writeRenewableCredKeys(initialCredential);
+    } else {
+      writeFixedCredKeys(initialCredential);
+    }
+    return Collections.unmodifiableMap(new HashMap<>(props));
+  }
+
+  protected CredPropsBuilder set(String key, String value) {
+    props.put(key, value);
+    return this;
+  }
+
+  /**
+   * Saves the current value of {@code key} from the Hadoop conf (falling back to {@code
+   * defaultOriginal}) under {@code key + ".original"}, then overrides {@code key} with {@code
+   * newValue}. The saved {@code .original} value is the side channel a reader uses to restore the
+   * real delegate implementation after the wrapper has been installed.
+   */
+  protected CredPropsBuilder saveAndOverride(String key, String defaultOriginal, String newValue) {
+    props.put(key + ".original", hadoopConf.get(key, defaultOriginal));
+    props.put(key, newValue);
+    return this;
+  }
+
+  /** Cloud-specific filesystem-impl overrides that install CredScopedFileSystem. */
+  protected abstract void writeImplOverrides();
+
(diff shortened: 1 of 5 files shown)


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -0,0 +1,114 @@
...
+  protected void writeBaseConf(Configuration conf) {}
+
+  /** Cloud-specific filesystem-impl overrides installed. */
+  protected abstract void writeImplOverrides(Configuration conf);
+
+  /** Cloud-specific keys naming the vended token provider (renewable path only). */
+  protected abstract void applyVendedProviderKeys();
+
+  /** Writes a single credential's secrets for this cloud. */
+  public abstract CredPropsBuilder writeCredKeys(boolean renewable, GenericCredential cred);
```
openinx: From the user perspective,  those abstracted methods are unclear to me.  Think from the caller side,  the core arguments that the caller care most are those: 
*  cloudType: s3/gcs/azure.
*  hadoopConf:  to save and restore the original fs implemetations. 
*  credScopedFsEnabled:  true/false
*  renewalCredEnabled: true/false
*  initialCredential

with this context, then it will be much more clear to  design an `CredPropsBuilder` interface like this: 

```java
    return CredPropsBuilder.forCloud(cloudType.get(), hadoopConf)
        .renewCredEnabled(renewCredEnabled)
        .credScopedFsEnabled(credScopedFsEnabled)
        .initialCredential(cred)
        .build();
```

and all the internal props setting should be hidden inside the `CredPropsBuilder` and its subclasses, but not expose as `public` method.   only those methods above should be exposed as public.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/GcsCredPropsBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+    set("fs.gs.impl.disable.cache", "true");
+  }
+
+  @Override
+  protected void writeBaseConf(Configuration conf) {
+    // The upstream GCS connector defaults this to true which causes the connector to
+    // stat every ancestor directory on file creation. With UC-vended downscoped tokens
+    // (scoped to a table's path prefix) these ancestor stats return 403. Default to
+    // false; users with broader credentials can opt back in via Hadoop/Spark config.
+    set(GCS_CONFLICT_CHECK_KEY, conf.get(GCS_CONFLICT_CHECK_KEY, "false"));
```
openinx, line 25: If we need to set this for all the Gcs with no brain,  then why do we still need this `writeBaseConf` method, and instead why not just set it directly inside the `GcsCredPropsBuilder` constructor.   that's will be more simplified.

foss-contributor: I was thinking if we have writeBaseConf, we can use it for future additional base configurations. I agree let's put it in the ctor and pass in conf to the ctor.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
foss-contributor: Done

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.hadoop.conf.Configuration;
+
+/** Builds the cloud-provider specific Hadoop configuration credential properties. */
+public abstract class CredPropsBuilder {
+  /** CredScopedFileSystem implementation classes, swapped in when cred-scoped FS is enabled. */
+  protected static final String CRED_SCOPED_FS_CLASS =
+      "io.unitycatalog.hadoop.internal.fs.CredScopedFileSystem";
+
+  protected static final String CRED_SCOPED_AFS_CLASS =
+      "io.unitycatalog.hadoop.internal.fs.CredScopedFs";
```
openinx: As we will always load those classes regardless which cloud we will use, then I suggest to use the `CredScopedFileSystem.class.getName()` and `CredScopedFs.class.getName()` direclty. 

but this does not apply to the `io.unitycatalog.hadoop.internal.auth.AbfsVendedTokenProvider` etc,  since we want to selectively load the class depends on which cloud we are actually using. saying we don't want to load any extra abfs library during the aws accessing.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+    this.initialCredential = initialCredential;
+    return this;
+  }
+
+  public Map<String, String> build() {
+    Preconditions.checkNotNull(initialCredential, "initialCredential is required");
+    if (renewCredEnabled) {
+      writeRenewableCredKeys(initialCredential);
+    } else {
+      writeFixedCredKeys(initialCredential);
```
openinx: Do we actually need the `initialCredential` for the fixedCredKey case ?  as FixedCred case, is always using the static credential, there is no meaning for us to access the initial credential ( consume one RPC ) actually....

openinx: okay, as we need the the initial credential to set as the static credential for the first time. so we always need the `initialCredential`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+    throw new IllegalStateException("Unhandled cloud type: " + cloudType);
+  }
+
+  /**
+   * Records whether renewal is enabled and, when it is, writes the wiring the vended token provider
+   * needs to renew the credential: the provider-class keys, the catalog uri, the auth configs, the
+   * credential-scope identity, and the engine versions. The fixed path writes none of this.
+   */
+  public CredPropsBuilder renewCredEnabled(
```
openinx: This method actually confused me, consider the two cases: 

1.  `renewCredEnabled` is `true`,  then all the `catalogUri`, `tokenProvider`, `credId`, and `appVersions` actually are required.  that's correct. 

2. `renewCredEnabled` is `false`,  then all the args are not needed, but why do we still need to set all of those??

openinx: with this context,  I will suggest to change it to `renewCredEnabled(catalogUri, tokenProvider, credId, appVersions)`,   and we only call it when the `renewCredEnabled` is true. and for the static credential case, we should never call this method.

openinx: And in this `renewCredEnabled` method, we usually don't do any real actions except the precondition checks,  until we call the real `build()`.  Only in that `build()`, we actually have all the context, that what we need to build, and we eventually do all the props setting in that method.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+          (key, value) -> set(UCHadoopConfConstants.UC_ENGINE_VERSION_PREFIX + key, value));
+    }
+    return this;
+  }
+
+  /** When enabled, writes the cloud filesystem-impl overrides that install CredScopedFileSystem. */
+  public CredPropsBuilder credScopedFsEnabled(boolean credScopedFsEnabled) {
+    if (credScopedFsEnabled) {
+      writeImplOverrides();
+    }
```
openinx: Let's only focus on set the `credScopedFsEnabled` member in this builder chained method, and defer the real props building into that `build()` ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+    return this;
+  }
+
+  /** Records the fetched credential whose secrets are written by {@link #build()}. */
+  public CredPropsBuilder initialCredential(GenericCredential initialCredential) {
+    this.initialCredential = initialCredential;
+    return this;
+  }
+
+  public Map<String, String> build() {
```
openinx, line 82: Let's move all the props building in this method, rather than distributing them into every builder methods.  that has the global context, and actually always help us to consolidate those building into a single place.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+   * real delegate implementation after the wrapper has been installed.
+   */
+  protected CredPropsBuilder saveAndOverride(String key, String defaultOriginal, String newValue) {
+    props.put(key + ".original", hadoopConf.get(key, defaultOriginal));
+    props.put(key, newValue);
+    return this;
+  }
+
+  /** Cloud-specific filesystem-impl overrides that install CredScopedFileSystem. */
+  protected abstract void writeImplOverrides();
```
openinx: -> `setFsImplKeys`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+
+  /** Cloud-specific key naming the vended token provider (renewable path only). */
+  protected abstract void writeVendedProviderKeys();
```
openinx: -> `setVendedProviderKeys()`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+
+  /** Writes the renewable-path credential secrets (the {@code init.*} keys) for this cloud. */
+  protected abstract void writeRenewableCredKeys(GenericCredential cred);
```
openinx: -> `setRenewableCredKeys()`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+
+  /** Writes the fixed-path credential secrets for this cloud. */
+  protected abstract void writeFixedCredKeys(GenericCredential cred);
```
openinx: -> `setFixedCredKeys(..)`.

openinx (changes requested):
@foss-contributor , thanks for the contribution,  I left few comments, mostly in the abstracted `CredPropsBuilder`.  Please take a look, thanks.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/S3CredPropsBuilder.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @Override
+  protected void writeVendedProviderKeys() {
+    set(UCHadoopConfConstants.S3A_CREDENTIALS_PROVIDER, AWS_VENDED_TOKEN_PROVIDER_CLASS);
+  }
+
+  @Override
+  protected void writeRenewableCredKeys(GenericCredential cred) {
+    AwsCredential aws = (AwsCredential) cred;
```
openinx, line 45: Please do the `Precondition.checkArgument(cred instanceof AwsCredential, '....')` before we do the real casting,  so that we can report the error message if there is some real class mismatches.  this applies to other places.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -0,0 +1,132 @@
...
+   * real delegate implementation after the wrapper has been installed.
+   */
+  protected CredPropsBuilder saveAndOverride(String key, String defaultOriginal, String newValue) {
+    props.put(key + ".original", hadoopConf.get(key, defaultOriginal));
+    props.put(key, newValue);
+    return this;
+  }
+
+  /** Cloud-specific filesystem-impl overrides that install CredScopedFileSystem. */
+  protected abstract void setFsImplKeys();
```
openinx: oh, sorry, this should be even more clear,  `setCredScopedFsKeys()`.

openinx (approved):
Looks good for me now, thanks @foss-contributor for the contribution !

PR #1693, opened 2026-07-20
fix(hadoop): normalize cloud storage prefixes for credential selection

foss-contributor:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1693/files) to review incremental changes.
- [**stack/normalize-cloud-storage-prefixes**](https://github.com/unitycatalog/unitycatalog/pull/1693) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1693/files)] ← _this PR_
  - [stack/location-aware-delegate-cache-keys](https://github.com/unitycatalog/unitycatalog/pull/1695) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1695/files/6a8116af333f9d104b86a1efe8e27528a493900b..1a3cb6c5fc7b859072d81e4ad17bab73c196c2ca)]
    - [stack/credential-location-identity](https://github.com/unitycatalog/unitycatalog/pull/1699) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1699/files/1a3cb6c5fc7b859072d81e4ad17bab73c196c2ca..2e04b7332131fdfcc383c9ab944c9dda938270aa)]
      - [stack/select-vended-credential-by-location](https://github.com/unitycatalog/unitycatalog/pull/1701) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1701/files/2e04b7332131fdfcc383c9ab944c9dda938270aa..32a422c0a5243c46ebeac62803a95c227ddc570f)]
        - [stack/scoped-credential-namespaces](https://github.com/unitycatalog/unitycatalog/pull/1704) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1704/files/32a422c0a5243c46ebeac62803a95c227ddc570f..1b09b884dff5caf0669637cb3f971bcf8a581f3b)]
          - [stack/encode-scoped-vended-credentials](https://github.com/unitycatalog/unitycatalog/pull/1705) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1705/files/1b09b884dff5caf0669637cb3f971bcf8a581f3b..9e31d11e3269274d902bfdc549fd2095eba5f99b)]

---------
## Summary

- Normalize storage locations before selecting Delta storage credentials by centralizing prefix coverage and canonicalization in `CredentialUtil`.
- Canonicalize known CloudType aliases and casing, including `s3a` to `s3` and `abfss` to `abfs`, while retaining trailing-slash and path-boundary semantics.
- Compare canonical prefix lengths so the most specific matching credential remains selected.

## Testing

Added coverage for aliases, scheme casing, trailing slashes, cross-cloud non-matches, unknown schemes, and credential selection.

## Issue

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CloudType.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CloudType.java
@@ -1,6 +1,7 @@
 package io.unitycatalog.hadoop.internal;
 
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -11,6 +12,7 @@ public enum CloudType {
   GCS("gs"),
   ABFS("abfs", "abfss");
 
+  private final String canonicalScheme;
   private final Set<String> schemes;
 
   private static final Map<String, CloudType> BY_SCHEME = new HashMap<>();
@@ -23,12 +25,24 @@ public enum CloudType {
     }
   }
 
-  CloudType(String... schemes) {
-    this.schemes = Set.of(schemes);
+  /** The first scheme is the canonical one; the rest are aliases that normalize to it. */
+  CloudType(String canonicalScheme, String... aliases) {
+    Set<String> all = new LinkedHashSet<>();
+    all.add(canonicalScheme);
+    for (String alias : aliases) {
+      all.add(alias);
+    }
+    this.canonicalScheme = canonicalScheme;
+    this.schemes = Set.copyOf(all);
   }
 
   /** Resolves a scheme to its underlying {@link CloudType}. */
   public static Optional<CloudType> fromScheme(String scheme) {
     return Optional.ofNullable(BY_SCHEME.get(scheme));
+  }
+
+  /** The canonical scheme for this cloud; every alias normalizes to it. */
+  public String canonicalScheme() {
+    return canonicalScheme;
   }
 }
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CloudType.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CloudType.java
@@ -1,19 +1,15 @@
 package io.unitycatalog.hadoop.internal;
 
 import java.util.HashMap;
-import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 
 /** The connector's single source of truth for cloud storage types and their schemes. */
 public enum CloudType {
   S3("s3", "s3a"),
   GCS("gs"),
   ABFS("abfs", "abfss");
-
-  private final String canonicalScheme;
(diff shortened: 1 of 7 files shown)


connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/CredentialUtilTest.java
foss-contributor: Add a case sensitive does not match test

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CloudType.java
```diff
@@ -23,12 +21,20 @@ public enum CloudType {
...
   }
 
   /** Resolves a scheme to its underlying {@link CloudType}. */
   public static Optional<CloudType> fromScheme(String scheme) {
     return Optional.ofNullable(BY_SCHEME.get(scheme));
   }
+
+  /** The canonical scheme for this cloud; every alias normalizes to it. */
+  public String canonicalScheme() {
+    return schemes.get(0);
```
openinx: No, that's incorrect for `abfs` and `abfss`.  if someone use `abfs://bucket1/key`, and other one use `abfss://bucket1/key`,  are they actually the same ???

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
```diff
@@ -77,11 +78,31 @@ public static DeltaStorageCredential selectForLocation(
   }
 
   static boolean prefixCovers(String location, String prefix) {
-    String l = stripTrailingSlashes(location);
-    String p = stripTrailingSlashes(prefix);
+    String l = canonicalLocation(location);
```
openinx: Actually, the `org.apache.hadoop.fs.Path` and `URI` should already have all the fundamental methods,  I don't see any reason for us to re-build our own.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CloudType.java
foss-contributor: I thought about this. I don't think we should normalize anymore. The cloud URIs should be matched to be exactly as is, without any canonicalization. We should keep the trailing slash and other normalization, but this is not necessary.

The server should always vend `abfss` and not `abfs` as `abfss` uses TLS. The credentials associated with the location ie `sas token` should work for either `abfs` or `abfss`. While Azure storage may accept this, semantically on the Hadoop connector end, this is incorrect as these are different storage locations with different schemes. As such, the client should be notified to use the correct driver for the scheme, rather than accessing with the wrong scheme but correct credential. Same applies to `s3a -> s3`, the cred is scoped for the `s3` location, not. a `s3a` location. 

This semantic is preserved in future PRs. What do you think?

foss-contributor:
I thought about it some more. Actually, if we want to enforce strict URI prefix matching, then this PR is unnecessary...

PR #1618, opened 2026-06-12
Implement Iceberg REST catalog write endpoints

dataders:
## Implement Iceberg REST catalog write endpoints

> 🤖 Heads up: this PR was put together with Claude Fable (AI). A human (@dataders) is driving it and will own the review back-and-forth, but the code + this description were AI-generated. Flagging so reviewers know what they're looking at.

### What this does

Adds the write side of the Iceberg REST catalog to OSS Unity Catalog, so spec-compliant clients (PyIceberg, Spark `RESTCatalog`, the DuckDB iceberg extension, etc.) can actually create and commit to Iceberg tables — not just read UniForm. Covers the write-endpoint part of #3.

| Endpoint | Gist |
|---|---|
| `POST …/namespaces` | Create a schema (delegates to `SchemaService`; single-level only, like today). |
| `POST …/tables` | Create a native Iceberg table: build initial `TableMetadata`, write `00000-*.metadata.json` via `FileIO`, register in UC as `EXTERNAL` with a new `ICEBERG` format. `stage-create` supported and stateless (mirrors Iceberg's `CatalogHandlers`). |
| `POST …/tables/{table}` | Commit: check `UpdateRequirement`s, apply `MetadataUpdate`s, write the next `metadata.json`, then CAS the stored location under a row lock. Lost race → 409. `assert-create` commits materialize a staged create atomically. |
| `DELETE …/tables/{table}` | Drop a REST-created table. Returns **200** not 204 (Armeria stalls body-less 204s on HTTP/1.1; clients accept 200). |

New endpoints are advertised in `GET /v1/config` so clients that do capability detection (DuckDB) light up automatically.

A few supporting bits:
- **Server-assigned locations**: clients that don't send a `location` (DuckDB) get a `MANAGED` table at `<managed-root>/tables/<uuid>`, resolved like managed Delta staging. Explicit location → `EXTERNAL`.
- **UniForm stays read-only**: only `ICEBERG`-format tables accept commits, so the Delta log and Iceberg metadata can't diverge.
- **Privilege-aware credential vending**: `FileIOFactory`/`TableConfigService` no longer hardcode `SELECT` (drops the old `FIXME`); write creds are only vended to native Iceberg tables, and all routes still authorize as metastore owner.
- **UC-visible schema**: Iceberg schemas convert to UC `ColumnInfo`s via a new `IcebergSchemaConverter`, so tables show up in the UC API/UI. The metadata file stays source of truth for Iceberg clients.

[...]

PR #1696, opened 2026-07-21
Implement native Iceberg REST catalog write endpoints

iamkiayun:
## Summary

Unity Catalog's Iceberg REST catalog can now create and commit native Iceberg tables instead of limiting Iceberg clients to read-only Delta UniForm metadata. Spec-aware clients can discover the write endpoints, create managed or external tables, perform staged creates, commit new snapshots with conflict protection, and drop catalog-created tables.

This draft is a current-`main`, DCO-clean continuation of the implementation developed by @dataders in #1618. The implementation credit and history are called out here explicitly; the commits were replayed onto current upstream because the original branch is behind `main` and its DCO check cannot be repaired by a third-party sign-off.

## Design decisions

- Native `ICEBERG` tables are writable, while Delta UniForm tables remain read-only so Iceberg metadata cannot diverge from the Delta log.
- Commits write a new metadata file and then compare-and-swap the catalog's metadata pointer under a row lock. Concurrent losers receive a conflict instead of silently overwriting another snapshot.
- Location-less creates receive a managed location under Unity Catalog's configured managed root. Explicit locations remain external.
- Staged creates remain unregistered until the client's `assert-create` commit materializes the table.
- Storage credentials are vended according to the requested read or write privileges, and the write routes retain metastore-owner authorization.

The remaining endpoint families in the broader Iceberg REST specification—such as multi-table transactions, rename/register, namespace property updates, and view writes—remain outside this change.

## Validation

- Focused Iceberg, schema conversion, metadata, credential, and cloud-vendor suites: **24 passed, 0 failed**.
- Full server suite: **313 total, 311 passed, 2 skipped, 0 failed, 0 errors**.
- The two skipped classes are the MySQL and PostgreSQL Delta Commits Testcontainers suites. The local OrbStack endpoint requires Docker API 1.40+, while this repository's older Testcontainers client advertises API 1.32. Upstream CI should exercise these suites in its supported Docker environment.
- `serverShaded/assembly`: **passed**.
- Checkstyle and `git diff --check`: **passed**.
- The DCO rewrite changed commit metadata only; the source tree hash remained identical before and after the rewrite.

## Known interoperability checks

- Recheck the table `HEAD` endpoint with a real HTTP/1.1 Iceberg client; the Java integration suite passes, but an earlier local client probe could stall.
- Lowercase REST boolean values work for `purgeRequested`; mixed-case values such as `False` remain an optional compatibility improvement.

## Related

Related: #3

Continuation of and implementation credit: #1618 by @dataders.

## Post-deploy monitoring & validation

- **Window/owner:** PR author and maintainers during CI, release qualification, and the first 24 hours of any deployment enabling the new endpoints.
- **Exercise:** create a namespace and native Iceberg table, append two snapshots, force one concurrent stale commit, reload the table, and drop it through a released Iceberg client.
- **Healthy signals:** create/commit/load/drop return expected 2xx responses; stale commits return 409; the stored metadata location advances once per successful commit; data remains readable after server restart.
- **Watch logs for:** `CommitFailedException`, `UPDATE_REQUIREMENT_CONFLICT`, `Failed to update Iceberg metadata location`, credential-vending failures, and metadata read/write exceptions.
- **Failure/rollback trigger:** unexpected 5xx responses, catalog pointers referencing missing metadata, unauthorized write credentials, or lost concurrent updates. Stop advertising the write endpoints and revert this change while preserving the catalog database and Iceberg metadata for diagnosis.

## PR checklist

- [x] Description explains the behavior and design boundaries.
- [x] Unit and integration coverage is included.
- [x] Focused and full local server validation completed.
- [ ] Upstream CI validates the Docker-backed database suites.
- [ ] Maintainers confirm whether separate user documentation is required.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
+++ b/server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
@@ -36,6 +36,31 @@ public class IcebergRestExceptionHandler extends BaseExceptionHandler {
 
   @Override
   protected BaseException toBaseException(Throwable cause) {
+    if (cause instanceof BaseException baseException) {
+      return switch (baseException.getErrorCode()) {
+        case SCHEMA_NOT_FOUND ->
+            wrapException(
+                ErrorCode.NOT_FOUND,
+                baseException.getErrorMessage(),
+                new NoSuchNamespaceException("%s", baseException.getErrorMessage()));
+        case TABLE_NOT_FOUND ->
+            wrapException(
+                ErrorCode.NOT_FOUND,
+                baseException.getErrorMessage(),
+                new NoSuchTableException("%s", baseException.getErrorMessage()));
+        case SCHEMA_ALREADY_EXISTS, TABLE_ALREADY_EXISTS ->
+            wrapException(
+                ErrorCode.ALREADY_EXISTS,
+                baseException.getErrorMessage(),
+                new AlreadyExistsException("%s", baseException.getErrorMessage()));
+        case UPDATE_REQUIREMENT_CONFLICT ->
+            wrapException(
+                ErrorCode.ALREADY_EXISTS,
+                baseException.getErrorMessage(),
+                new CommitFailedException("%s", baseException.getErrorMessage()));
+        default -> baseException;
+      };
+    }
     if (cause instanceof NoSuchNamespaceException
         || cause instanceof NoSuchTableException
         || cause instanceof NoSuchViewException) {
--- a/server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
@@ -40,6 +40,17 @@ public class StagingTableRepository {
     return query.uniqueResult(); // Returns null if no result is found
   }
 
+  public boolean isUncommittedStagingLocation(NormalizedURL stagingLocation) {
+    return TransactionManager.executeWithTransaction(
+        sessionFactory,
+        session -> {
+          StagingTableDAO stagingTable = findByStagingLocation(session, stagingLocation);
+          return stagingTable != null && !stagingTable.isStageCommitted();
+        },
+        "Failed to look up staging table at " + stagingLocation,
+        /* readOnly = */ true);
+  }
+
   private void validateIfAlreadyExists(
       Session session, UUID schemaId, String tableName, NormalizedURL stagingLocation) {
     // Check if table by the same name already exists. It's OK if a staging table with the same name
--- a/server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
@@ -40,17 +40,6 @@ public class StagingTableRepository {
     return query.uniqueResult(); // Returns null if no result is found
   }
 
-  public boolean isUncommittedStagingLocation(NormalizedURL stagingLocation) {
-    return TransactionManager.executeWithTransaction(
-        sessionFactory,
-        session -> {
-          StagingTableDAO stagingTable = findByStagingLocation(session, stagingLocation);
-          return stagingTable != null && !stagingTable.isStageCommitted();
-        },
-        "Failed to look up staging table at " + stagingLocation,
-        /* readOnly = */ true);
-  }
-
   private void validateIfAlreadyExists(
       Session session, UUID schemaId, String tableName, NormalizedURL stagingLocation) {
     // Check if table by the same name already exists. It's OK if a staging table with the same name
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -52,9 +52,9 @@ import java.util.UUID;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.tuple.Pair;
 import org.hibernate.LockMode;
+import org.hibernate.PessimisticLockException;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
-import org.hibernate.PessimisticLockException;
 import org.hibernate.query.Query;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -80,6 +80,10 @@ public class TableRepository {
    * always finalize to MANAGED Delta.
    */
   public record TableStorageLocationInfo(NormalizedURL url, TableType tableType) {}
+
+  /** Detached catalog state needed to load or commit an Iceberg metadata file. */
+  public record IcebergTableState(
+      UUID tableId, DataSourceFormat dataSourceFormat, String metadataLocation) {}
 
   /**
    * Retrieves the storage location for a table or staging table by its ID, plus the table type.
@@ -385,8 +389,8 @@ public class TableRepository {
 
   /**
    * Acquire {@code SELECT ... FOR UPDATE} on the table row and refresh in-memory state, so two
-   * concurrent Iceberg REST commits serialize. Lock-wait timeouts and deadlock
-   * victims surface as {@code UPDATE_REQUIREMENT_CONFLICT} (409) instead of a generic 500.
+   * concurrent Iceberg REST commits serialize. Lock-wait timeouts and deadlock victims surface as
+   * {@code UPDATE_REQUIREMENT_CONFLICT} (409) instead of a generic 500.
    */
   private static void lockTableForUpdate(
       Session session, TableInfoDAO dao, String catalog, String schema, String table) {
@@ -588,25 +592,62 @@ public class TableRepository {
     return dao.getUniformIcebergMetadataLocation();
   }
 
+  public IcebergTableState getIcebergTableState(
+      String catalogName, String schemaName, String tableName) {
+    return getIcebergTableState(catalogName, schemaName, tableName, false);
+  }
+
+  public IcebergTableState getNativeIcebergTableState(
+      String catalogName, String schemaName, String tableName) {
+    return getIcebergTableState(catalogName, schemaName, tableName, true);
+  }
+
+  private IcebergTableState getIcebergTableState(
+      String catalogName, String schemaName, String tableName, boolean requireNativeIceberg) {
+    String fullName = catalogName + "." + schemaName + "." + tableName;
+    return TransactionManager.executeWithTransaction(
+        sessionFactory,
+        session -> {
+          TableInfoDAO dao = findTableOrThrow(session, catalogName, schemaName, tableName);
+          DataSourceFormat dataSourceFormat =
+              dao.getDataSourceFormat() == null
+                  ? null
+                  : DataSourceFormat.fromValue(dao.getDataSourceFormat());
+          if (requireNativeIceberg && dataSourceFormat != DataSourceFormat.ICEBERG) {
+            throw new BaseException(
+                ErrorCode.INVALID_ARGUMENT,
+                "Table is not a native Iceberg table and cannot be committed to via the Iceberg"
+                    + " REST catalog: "
+                    + fullName);
+          }
+          return new IcebergTableState(
+              dao.getId(), dataSourceFormat, dao.getUniformIcebergMetadataLocation());
+        },
+        "Failed to load Iceberg table state for " + fullName,
+        /* readOnly = */ true);
+  }
+
   /**
-   * Compare-and-swap the Iceberg metadata location of a native Iceberg table (created through the
-   * Iceberg REST catalog). This is the linearization point of an Iceberg REST commit: the metadata
-   * file for {@code newMetadataLocation} is written by the caller before this method runs, and the
-   * swap only succeeds when the stored location still matches {@code expectedMetadataLocation}
-   * ({@code null} for a freshly created table). A mismatch means a concurrent commit won and
-   * surfaces as {@link ErrorCode#UPDATE_REQUIREMENT_CONFLICT} so the caller can translate it into
-   * an Iceberg {@code CommitFailedException}.
+   * Commits a native Iceberg table's metadata pointer, UC columns, and UC properties in one locked
+   * transaction. This is the linearization point of an Iceberg REST commit: the metadata file for
+   * {@code newMetadataLocation} is written by the caller before this method runs, and the update
+   * only succeeds when the stored location still matches {@code expectedMetadataLocation}. A
+   * mismatch means a concurrent commit won and surfaces as {@link
+   * ErrorCode#UPDATE_REQUIREMENT_CONFLICT} so the caller can translate it into an Iceberg {@code
+   * CommitFailedException}.
    *
    * <p>Only tables with {@link DataSourceFormat#ICEBERG} can be committed to: UniForm-enabled Delta
    * tables also carry a uniform metadata location, but their Iceberg metadata is derived from the
    * Delta log and stays read-only through this API.
    */
-  public void setIcebergMetadataLocation(
+  public void commitIcebergTable(
       String catalogName,
       String schemaName,
       String tableName,
       String expectedMetadataLocation,
-      String newMetadataLocation) {
+      NormalizedURL newMetadataLocation,
+      List<ColumnInfo> columns,
+      Map<String, String> tableProperties) {
     String callerId = IdentityUtils.findPrincipalEmailAddress();
     String fullName = catalogName + "." + schemaName + "." + tableName;
     TransactionManager.executeWithTransaction(
@@ -631,7 +672,23 @@ public class TableRepository {
                     + " but found "
                     + dao.getUniformIcebergMetadataLocation());
           }
-          dao.setUniformIcebergMetadataLocation(newMetadataLocation);
+          List<ColumnInfoDAO> newColumns = ColumnInfoDAO.fromList(columns);
+          newColumns.forEach(
+              column -> {
+                column.setId(UUID.randomUUID());
+                column.setTable(dao);
+              });
+          dao.getColumns().clear();
+          session.flush();
+          dao.getColumns().addAll(newColumns);
+          dao.setColumnCount(newColumns.size());
+
+          MutablePropertyMap properties = MutablePropertyMap.load(session, dao.getId());
+          properties.removeAll(new ArrayList<>(properties.asMap().keySet()));
+          properties.putAll(tableProperties);
+          properties.flush(session, dao.getId());
+
+          dao.setUniformIcebergMetadataLocation(newMetadataLocation.toString());
           dao.setUpdatedAt(new Date());
           dao.setUpdatedBy(callerId);
           session.merge(dao);
@@ -655,7 +712,20 @@ public class TableRepository {
   }
 
   public TableInfo createTable(CreateTable createTable) {
-    return createTableImpl(createTable, Optional.empty(), (session, dao, tableInfo) -> tableInfo);
+    return createTableImpl(
+        createTable, Optional.empty(), Optional.empty(), (session, dao, tableInfo) -> tableInfo);
+  }
+
+  public TableInfo createTableForIceberg(CreateTable createTable, NormalizedURL metadataLocation) {
+    if (createTable.getDataSourceFormat() != DataSourceFormat.ICEBERG) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT, "createTableForIceberg requires ICEBERG format.");
+    }
+    return createTableImpl(
+        createTable,
+        Optional.empty(),
+        Optional.of(metadataLocation),
+        (session, dao, tableInfo) -> tableInfo);
   }
 
   /**
@@ -674,6 +744,7 @@ public class TableRepository {
     return createTableImpl(
         createTable,
         uniformFields,
+        Optional.empty(),
         (session, dao, tableInfo) ->
             buildLoadTableResponse(
                 session,
@@ -685,16 +756,17 @@ public class TableRepository {
   }
 
   /**
-   * Shared implementation for the two {@code create} entry points. Validates the name, opens a
+   * Shared implementation for the table {@code create} entry points. Validates the name, opens a
    * write transaction, builds the new {@link TableInfoDAO} (row, columns, properties), applies
-   * UniForm Iceberg metadata when {@code uniformFields} is non-empty, persists the DAO, then hands
-   * it and the built {@link TableInfo} to {@code mapper} which picks the return shape each caller
-   * needs. Keeps the create path as a single operation, and pins all DAO setters before {@code
-   * session.persist} so the create lands as a single INSERT.
+   * UniForm or native Iceberg metadata when supplied, persists the DAO, then hands it and the built
+   * {@link TableInfo} to {@code mapper} which picks the return shape each caller needs. Keeps the
+   * create path as a single operation, and pins all DAO setters before {@code session.persist} so
+   * the create lands as a single INSERT.
    */
   private <T> T createTableImpl(
       CreateTable createTable,
       Optional<DeltaUniformUtils.UniformIcebergFields> uniformFields,
+      Optional<NormalizedURL> nativeIcebergMetadataLocation,
       CreateResultMapper<T> mapper) {
     ValidationUtils.validateSqlObjectName(createTable.getName());
     String callerId = IdentityUtils.findPrincipalEmailAddress();
@@ -746,24 +818,26 @@ public class TableRepository {
           } else if (tableType == TableType.MANAGED) {
             storageLocation = NormalizedURL.from(createTable.getStorageLocation());
             serverProperties.checkManagedTableEnabled();
-            if (createTable.getDataSourceFormat() != DataSourceFormat.DELTA) {
+            if (createTable.getDataSourceFormat() != DataSourceFormat.DELTA
+                && createTable.getDataSourceFormat() != DataSourceFormat.ICEBERG) {
               throw new BaseException(
                   ErrorCode.INVALID_ARGUMENT,
-                  "Managed table creation is only supported for Delta format.");
+                  "Managed table creation is only supported for Delta and Iceberg formats.");
             }
-            // Find and commit staging table with the same staging location
             StagingTableDAO stagingTableDAO =
                 repositories
                     .getStagingTableRepository()
                     .commitStagingTable(session, callerId, storageLocation);
             tableUUID = stagingTableDAO.getId();
-            // MANAGED tables (created via either UC REST or Delta REST) must carry UC_TABLE_ID in
-            // their properties, matching the staging UUID. UC has the staging UUID as the source of
-            // truth; a request with a missing or mismatched UC_TABLE_ID gets rejected here
-            // instead of producing an internally-inconsistent UC table that subsequent commits
-            // would fail on.
-            UcManagedDeltaContract.validateTableIdProperty(
-                createTable.getProperties(), tableUUID.toString());
+            if (createTable.getDataSourceFormat() == DataSourceFormat.DELTA) {
+              // MANAGED tables (created via either UC REST or Delta REST) must carry UC_TABLE_ID
+              // in their properties, matching the staging UUID. UC has the staging UUID as the
+              // source of truth; a request with a missing or mismatched UC_TABLE_ID gets rejected
+              // here instead of producing an internally-inconsistent UC table that subsequent
+              // commits would fail on.
+              UcManagedDeltaContract.validateTableIdProperty(
+                  createTable.getProperties(), tableUUID.toString());
+            }
           } else if (tableType == TableType.METRIC_VIEW) {
             storageLocation = null;
             validateMetricView(session, createTable);
@@ -817,6 +891,8 @@ public class TableRepository {
           // UniForm Iceberg fields (when supplied by the Delta create path) are written while the
           // entity is still transient so they're folded into the single INSERT below.
           DeltaUniformUtils.applyToDao(tableInfoDAO, uniformFields);
+          nativeIcebergMetadataLocation.ifPresent(
+              location -> tableInfoDAO.setUniformIcebergMetadataLocation(location.toString()));
           session.persist(tableInfoDAO);
           if (RepositoryUtils.isViewLike(tableType.getValue())) {
             DependencyDAO.DependentType dependentType = DependencyDAO.DependentType.TABLE;
@@ -1041,43 +1117,5 @@ public class TableRepository {
         .forEach(session::remove);
     session.remove(tableInfoDAO);
     return tableInfoDAO;
-  }
-
-  /**
-   * Renames a table within the same schema. Looks up the source table by its three-part name
-   * (throwing {@link ErrorCode#TABLE_NOT_FOUND} if absent) and rejects a collision with an existing
-   * table of the target name in the same schema (throwing {@link ErrorCode#TABLE_ALREADY_EXISTS}).
-   * Renaming a table to its current name is an idempotent no-op. Metadata-only: the table UUID,
-   * schema, and storage location are all unchanged.
-   */
-  public void renameTable(String catalog, String schema, String table, String newName) {
-    ValidationUtils.validateSqlObjectName(newName);
-    String callerId = IdentityUtils.findPrincipalEmailAddress();
-    TransactionManager.executeWithTransaction(
-        sessionFactory,
-        session -> {
-          UUID schemaId =
-              repositories.getSchemaRepository().getSchemaIdOrThrow(session, catalog, schema);
-          TableInfoDAO tableInfoDAO = findBySchemaIdAndName(session, schemaId, table);
-          if (tableInfoDAO == null) {
-            throw new BaseException(ErrorCode.TABLE_NOT_FOUND, "Table not found: " + table);
-          }
-          // Do the no-op check after the table lookup. A missing table must still return
-          // TABLE_NOT_FOUND rather than succeed as a no-op.
-          if (table.equals(newName)) {
-            return null;
-          }
-          if (findBySchemaIdAndName(session, schemaId, newName) != null) {
-            throw new BaseException(
-                ErrorCode.TABLE_ALREADY_EXISTS, "Table already exists: " + newName);
-          }
-          tableInfoDAO.setName(newName);
-          tableInfoDAO.setUpdatedAt(new Date());
-          tableInfoDAO.setUpdatedBy(callerId);
-          session.merge(tableInfoDAO);
-          return null;
-        },
-        "Failed to rename table " + catalog + "." + schema + "." + table,
-        /* readOnly = */ false);
   }
 }
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -51,8 +51,6 @@ import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.tuple.Pair;
-import org.hibernate.LockMode;
-import org.hibernate.PessimisticLockException;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.query.Query;
@@ -83,7 +81,10 @@ public class TableRepository {
 
   /** Detached catalog state needed to load or commit an Iceberg metadata file. */
   public record IcebergTableState(
-      UUID tableId, DataSourceFormat dataSourceFormat, String metadataLocation) {}
+      UUID tableId,
+      DataSourceFormat dataSourceFormat,
+      String metadataLocation,
+      String storageLocation) {}
 
   /**
    * Retrieves the storage location for a table or staging table by its ID, plus the table type.
@@ -329,8 +330,7 @@ public class TableRepository {
         session -> {
           TableInfoDAO dao = findTableOrThrow(session, catalog, schema, table);
           requireDeltaTable(dao, catalog, schema, table);
-          DeltaCommitRepository.lockTableForCommit(
-              session, dao, dao.getId(), Optional.of(tableFullName));
+          RepositoryUtils.lockTableForCommit(session, dao, dao.getId(), Optional.of(tableFullName));
           // assert-table-uuid is stable identity, so check it up front. assert-etag is deferred to
           // after the apply, captured here against pre-apply state: a commit advances the etag, and
           // an idempotent replay must bypass the etag entirely (it throws mid-apply and rolls back
@@ -384,28 +384,6 @@ public class TableRepository {
       throw new BaseException(
           ErrorCode.UNSUPPORTED_TABLE_FORMAT,
           "Table is not a Delta table: " + catalog + "." + schema + "." + table);
-    }
-  }
-
-  /**
-   * Acquire {@code SELECT ... FOR UPDATE} on the table row and refresh in-memory state, so two
-   * concurrent Iceberg REST commits serialize. Lock-wait timeouts and deadlock victims surface as
-   * {@code UPDATE_REQUIREMENT_CONFLICT} (409) instead of a generic 500.
-   */
-  private static void lockTableForUpdate(
-      Session session, TableInfoDAO dao, String catalog, String schema, String table) {
-    try {
-      session.refresh(dao, LockMode.PESSIMISTIC_WRITE);
-    } catch (PessimisticLockException e) {
-      throw new BaseException(
-          ErrorCode.UPDATE_REQUIREMENT_CONFLICT,
-          "Concurrent update in progress on "
-              + catalog
-              + "."
-              + schema
-              + "."
-              + table
-              + "; retry the request.");
     }
   }
 
@@ -621,7 +599,7 @@ public class TableRepository {
                     + fullName);
           }
           return new IcebergTableState(
-              dao.getId(), dataSourceFormat, dao.getUniformIcebergMetadataLocation());
+              dao.getId(), dataSourceFormat, dao.getUniformIcebergMetadataLocation(), dao.getUrl());
         },
         "Failed to load Iceberg table state for " + fullName,
         /* readOnly = */ true);
@@ -661,7 +639,7 @@ public class TableRepository {
                     + " REST catalog: "
                     + fullName);
           }
-          lockTableForUpdate(session, dao, catalogName, schemaName, tableName);
+          RepositoryUtils.lockTableForCommit(session, dao, dao.getId(), Optional.of(fullName));
           if (!Objects.equals(dao.getUniformIcebergMetadataLocation(), expectedMetadataLocation)) {
             throw new BaseException(
                 ErrorCode.UPDATE_REQUIREMENT_CONFLICT,
@@ -684,6 +662,9 @@ public class TableRepository {
           dao.setColumnCount(newColumns.size());
 
           MutablePropertyMap properties = MutablePropertyMap.load(session, dao.getId());
+          // Native Iceberg metadata is the authoritative projection of UC-visible properties. A
+          // commit therefore replaces the complete property set; server-owned properties must not
+          // be persisted here because the next client snapshot would intentionally erase them.
           properties.removeAll(new ArrayList<>(properties.asMap().keySet()));
           properties.putAll(tableProperties);
           properties.flush(session, dao.getId());
@@ -712,14 +693,52 @@ public class TableRepository {
   }
 
   public TableInfo createTable(CreateTable createTable) {
+    requireNonIcebergFormat(createTable, "createTable");
     return createTableImpl(
         createTable, Optional.empty(), Optional.empty(), (session, dao, tableInfo) -> tableInfo);
   }
 
+  /**
+   * Renames a table within its schema. Looks up the source table first (throwing {@link
+   * ErrorCode#TABLE_NOT_FOUND} if absent) and rejects a collision with an existing table of the
+   * target name in the same schema (throwing {@link ErrorCode#TABLE_ALREADY_EXISTS}). Renaming a
+   * table to its current name is an idempotent no-op. Metadata-only: the table UUID, schema, and
+   * storage location are all unchanged.
+   */
+  public void renameTable(String catalog, String schema, String table, String newName) {
+    ValidationUtils.validateSqlObjectName(newName);
+    String callerId = IdentityUtils.findPrincipalEmailAddress();
+    TransactionManager.executeWithTransaction(
+        sessionFactory,
+        session -> {
+          UUID schemaId =
+              repositories.getSchemaRepository().getSchemaIdOrThrow(session, catalog, schema);
+          TableInfoDAO tableInfoDAO = findBySchemaIdAndName(session, schemaId, table);
+          if (tableInfoDAO == null) {
+            throw new BaseException(ErrorCode.TABLE_NOT_FOUND, "Table not found: " + table);
+          }
+          // Do the no-op check after the table lookup. A missing table must still return
+          // TABLE_NOT_FOUND rather than succeed as a no-op.
+          if (table.equals(newName)) {
+            return null;
+          }
+          if (findBySchemaIdAndName(session, schemaId, newName) != null) {
+            throw new BaseException(
+                ErrorCode.TABLE_ALREADY_EXISTS, "Table already exists: " + newName);
+          }
+          tableInfoDAO.setName(newName);
+          tableInfoDAO.setUpdatedAt(new Date());
+          tableInfoDAO.setUpdatedBy(callerId);
+          session.merge(tableInfoDAO);
+          return null;
+        },
+        "Failed to rename table " + catalog + "." + schema + "." + table,
+        /* readOnly = */ false);
+  }
+
   public TableInfo createTableForIceberg(CreateTable createTable, NormalizedURL metadataLocation) {
     if (createTable.getDataSourceFormat() != DataSourceFormat.ICEBERG) {
-      throw new BaseException(
-          ErrorCode.INVALID_ARGUMENT, "createTableForIceberg requires ICEBERG format.");
+      throw new BaseException(ErrorCode.INTERNAL, "createTableForIceberg requires ICEBERG format.");
     }
     return createTableImpl(
         createTable,
@@ -741,6 +760,7 @@ public class TableRepository {
    */
   public DeltaLoadTableResponse createTableForDelta(
       CreateTable createTable, Optional<DeltaUniformUtils.UniformIcebergFields> uniformFields) {
+    requireNonIcebergFormat(createTable, "createTableForDelta");
     return createTableImpl(
         createTable,
         uniformFields,
@@ -753,6 +773,14 @@ public class TableRepository {
                 createTable.getCatalogName(),
                 createTable.getSchemaName(),
                 createTable.getName()));
+  }
+
+  private static void requireNonIcebergFormat(CreateTable createTable, String caller) {
+    if (createTable.getDataSourceFormat() == DataSourceFormat.ICEBERG) {
+      throw new BaseException(
+          ErrorCode.UNSUPPORTED_TABLE_FORMAT,
+          caller + " cannot create native ICEBERG tables; use createTableForIceberg instead.");
+    }
   }
 
   /**
@@ -824,6 +852,8 @@ public class TableRepository {
                   ErrorCode.INVALID_ARGUMENT,
                   "Managed table creation is only supported for Delta and Iceberg formats.");
             }
+            // Find and commit the staging table with the same staging location. This single
+            // transaction validates ownership and prevents a staging location from being reused.
             StagingTableDAO stagingTableDAO =
                 repositories
                     .getStagingTableRepository()
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -698,44 +698,6 @@ public class TableRepository {
         createTable, Optional.empty(), Optional.empty(), (session, dao, tableInfo) -> tableInfo);
   }
 
-  /**
-   * Renames a table within its schema. Looks up the source table first (throwing {@link
-   * ErrorCode#TABLE_NOT_FOUND} if absent) and rejects a collision with an existing table of the
-   * target name in the same schema (throwing {@link ErrorCode#TABLE_ALREADY_EXISTS}). Renaming a
-   * table to its current name is an idempotent no-op. Metadata-only: the table UUID, schema, and
-   * storage location are all unchanged.
-   */
-  public void renameTable(String catalog, String schema, String table, String newName) {
-    ValidationUtils.validateSqlObjectName(newName);
-    String callerId = IdentityUtils.findPrincipalEmailAddress();
-    TransactionManager.executeWithTransaction(
-        sessionFactory,
-        session -> {
-          UUID schemaId =
-              repositories.getSchemaRepository().getSchemaIdOrThrow(session, catalog, schema);
-          TableInfoDAO tableInfoDAO = findBySchemaIdAndName(session, schemaId, table);
-          if (tableInfoDAO == null) {
-            throw new BaseException(ErrorCode.TABLE_NOT_FOUND, "Table not found: " + table);
-          }
-          // Do the no-op check after the table lookup. A missing table must still return
-          // TABLE_NOT_FOUND rather than succeed as a no-op.
-          if (table.equals(newName)) {
-            return null;
-          }
-          if (findBySchemaIdAndName(session, schemaId, newName) != null) {
-            throw new BaseException(
-                ErrorCode.TABLE_ALREADY_EXISTS, "Table already exists: " + newName);
-          }
-          tableInfoDAO.setName(newName);
-          tableInfoDAO.setUpdatedAt(new Date());
-          tableInfoDAO.setUpdatedBy(callerId);
-          session.merge(tableInfoDAO);
-          return null;
-        },
-        "Failed to rename table " + catalog + "." + schema + "." + table,
-        /* readOnly = */ false);
-  }
-
   public TableInfo createTableForIceberg(CreateTable createTable, NormalizedURL metadataLocation) {
     if (createTable.getDataSourceFormat() != DataSourceFormat.ICEBERG) {
       throw new BaseException(ErrorCode.INTERNAL, "createTableForIceberg requires ICEBERG format.");
@@ -779,7 +741,9 @@ public class TableRepository {
     if (createTable.getDataSourceFormat() == DataSourceFormat.ICEBERG) {
       throw new BaseException(
           ErrorCode.UNSUPPORTED_TABLE_FORMAT,
-          caller + " cannot create native ICEBERG tables; use createTableForIceberg instead.");
+          caller
+              + " cannot create native ICEBERG tables; use the Iceberg REST catalog endpoints"
+              + " instead.");
     }
   }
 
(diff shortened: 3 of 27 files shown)


yili-db:
Thank you for this PR. Will take a look.

server/src/main/java/io/unitycatalog/server/service/iceberg/FileIOFactory.java
```diff
@@ -32,10 +32,6 @@ public class FileIOFactory {
```
yili-db: Just FYI I have a PR to consolidate this file into `FileOperations` so there will be conflict.
https://github.com/unitycatalog/unitycatalog/pull/1687

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -142,6 +170,38 @@ public GetNamespaceResponse getNamespace(
...
+    request.validate();
+    if (request.namespace().levels().length != 1) {
+      throw new BadRequestException("Nested namespaces are not supported: %s", request.namespace());
+    }
+    String schemaName = request.namespace().level(0);
+    CreateSchema createSchema =
+        new CreateSchema().name(schemaName).catalogName(catalog).properties(request.properties());
+    String resp;
+    try {
+      resp = schemaService.createSchema(createSchema).aggregate().join().contentUtf8();
```
yili-db: This (service calling another service) is the pre-existing anti-pattern and it even complicates things.
Here it can simply be done with:
```
    SchemaInfo schemaInfo = schemaRepository.createSchema(createSchema);
    CatalogInfo catalogInfo = catalogRepository.getCatalog(catalog);
    initializeHierarchicalAuthorization(schemaInfo.getSchemaId(), catalogInfo.getId());
```
3 lines of dup code but it breaks the anti-pattern and saves the `.aggregate().join().contentUtf8()` plus `JsonUtils.getInstance().readValue(resp, SchemaInfo.class)`.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -142,6 +170,38 @@ public GetNamespaceResponse getNamespace(
...
+    } catch (BaseException e) {
```
yili-db: Catching this exception should be unnecessary. `IcebergRestExceptionHandler` should handle it already.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -183,14 +243,298 @@ public LoadTableResponse loadTable(
...
+      return true;
+    } catch (BaseException e) {
+      if (e.getErrorCode() == ErrorCode.TABLE_NOT_FOUND) {
+        return false;
+      }
+      throw e;
+    }
+  }
+
+  /** Looks up the UC table entry, translating UC's not-found error into the Iceberg shape. */
```
yili-db: Doesn't `IcebergRestExceptionHandler` already do this? If not, it should be fix instead.

server/src/main/java/io/unitycatalog/server/service/iceberg/TableConfigService.java
```diff
@@ -36,18 +32,24 @@ public TableConfigService(
   }
 
   public Map<String, String> getTableConfig(TableMetadata tableMetadata) {
```
yili-db: You can just change this function to make it explicit that privileges is always provided.
There are only two callers to it. Just change them to say READ_ONLY.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -183,14 +243,298 @@ public LoadTableResponse loadTable(
...
 
     TableMetadata tableMetadata = metadataService.readTableMetadata(metadataLocation);
-    Map<String, String> config = tableConfigService.getTableConfig(tableMetadata);
+    // Native Iceberg tables are writable through this catalog, so vend write-capable storage
+    // credentials; UniForm-derived metadata stays read-only. All write-capable routes authorize as
+    // metastore owner (see the @AuthorizeExpression on each), so READ_WRITE creds are only handed
+    // to principals that could already write through the UC API.
+    Map<String, String> config =
+        tableInfo.getDataSourceFormat() == DataSourceFormat.ICEBERG
+            ? tableConfigService.getTableConfig(tableMetadata, READ_WRITE)
```
yili-db: For `loadTable` of a native Iceberg table, it should start with `READ_WRITE` which is good. But if the user doesn't have the access to the table (authorization expression of this entire service will be properly implemented later), it should fallback to `READ_ONLY`.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -183,14 +243,298 @@ public LoadTableResponse loadTable(
...
+    TableType tableType;
+    if (location == null || location.isEmpty()) {
+      // No client-supplied location: assign one under the managed storage root (schema's, else
+      // catalog's, else the storage-root.tables server property) and create a MANAGED table,
+      // mirroring how staging locations are assigned for managed Delta tables. Clients like
+      // DuckDB rely on server-assigned locations.
+      tableType = TableType.MANAGED;
+      location =
+          tableRepository
+              .assignManagedIcebergTableLocation(catalog, namespace, UUID.randomUUID())
```
yili-db: Please use `StagingTableRepository.createStagingTable` instead.
Despite not being a first class concept in Iceberg, this staging registration works very well for it as a implementation detail. And in this repo the staging table is not limited to Delta. It does the same location allocation but additionally with registration.
Later a GC process will be implemented so that orphaned managed location will be cleaned up.

As a bonus point, it also does `validateIfAlreadyExists`

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -183,14 +243,298 @@ public LoadTableResponse loadTable(
...
+              .toString();
+    } else {
+      tableType = TableType.EXTERNAL;
+      location = location.replaceAll("/+$", "");
```
yili-db: Use `NormalizedURL` instead

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -21,4 +34,69 @@ public TableMetadata readTableMetadata(String metadataLocation) {
...
+   * directories. Object stores have no directories, so this is a no-op for them.
+   */
+  public void prepareTableLocation(TableMetadata tableMetadata) {
+    NormalizedURL location = NormalizedURL.from(tableMetadata.location());
+    UriScheme scheme = UriScheme.fromURI(location.toUri());
+    if (scheme != UriScheme.FILE && scheme != UriScheme.NULL) {
+      return;
+    }
+    String base = tableMetadata.location();
+    File baseDir = base.startsWith("file:") ? new File(URI.create(base).getPath()) : new File(base);
```
yili-db: `FileOperations.createStorageLocationDir` can help with this. It is no-op for cloud but mkdir for local.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -183,14 +243,298 @@ public LoadTableResponse loadTable(
...
+            .name(name)
+            .catalogName(catalog)
+            .schemaName(namespace)
+            .tableType(tableType)
+            .dataSourceFormat(DataSourceFormat.ICEBERG)
+            .columns(IcebergSchemaConverter.toColumnInfos(tableMetadata.schema()))
+            .storageLocation(tableMetadata.location())
+            .properties(properties);
+    try {
+      tableService.createTable(createTable);
```
yili-db: Prefer not calling tableService. Just call:
```
    TableInfo tableInfo = tableRepository.createTable(createTable);
    SchemaInfo schemaInfo =
        schemaRepository.getSchema(tableInfo.getCatalogName() + "." + tableInfo.getSchemaName());
    initializeHierarchicalAuthorization(tableInfo.getTableId(), schemaInfo.getSchemaId());
```

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -183,14 +243,298 @@ public LoadTableResponse loadTable(
...
+        }
+        throw new AlreadyExistsException("Table already exists: %s.%s", namespace, name);
+      }
+      throw e;
+    }
+    try {
+      metadataService.writeTableMetadata(tableMetadata, metadataLocation);
+      tableRepository.setIcebergMetadataLocation(catalog, namespace, name, null, metadataLocation);
+    } catch (RuntimeException e) {
+      // Roll the UC entry back so a failed create doesn't leave behind a table that can never be
```
yili-db: For rollback, it appears to me that the files should be deleted first (rolling back `writeTableMetadata`) then the deleteTable (rolling back `CreateTable`).

Or even better, why not `writeTableMetadata` first then `CreateTable`.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -183,14 +243,298 @@ public LoadTableResponse loadTable(
...
+          throw new CommitFailedException(
+              "Requirement failed: table already exists: %s.%s", namespace, name);
+        }
+        throw new AlreadyExistsException("Table already exists: %s.%s", namespace, name);
+      }
+      throw e;
+    }
+    try {
+      metadataService.writeTableMetadata(tableMetadata, metadataLocation);
+      tableRepository.setIcebergMetadataLocation(catalog, namespace, name, null, metadataLocation);
```
yili-db: `setIcebergMetadataLocation` can be folded with `CreateTable` in the same call and the same transaction. You can have a dedicated `createTable` call for Iceberg that takes an additional `metadataLocation`.
It's not available yet but you can add one for Iceberg. Just see this example:
```
  public DeltaLoadTableResponse createTableForDelta(
      CreateTable createTable, Optional<DeltaUniformUtils.UniformIcebergFields> uniformFields) {
    return createTableImpl(
```

yili-db:
I am still reviewing as it's a large PR implementing a major new feature.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -560,6 +562,133 @@ public String getTableUniformMetadataLocation(
...
+  }
+
+  /**
+   * Returns whether {@code location} sits under the managed-tables area that {@link
+   * #assignManagedIcebergTableLocation} assigns from, i.e. {@code <managed-root>/tables/}. The
+   * Iceberg REST commit path uses this to classify a staged create whose location was assigned by
+   * the server (MANAGED) versus supplied by the client (EXTERNAL). Returns false when no managed
+   * root is configured.
+   */
+  public boolean isManagedIcebergTableLocation(
```
yili-db: This can be done in a much simpler way: all managed table location always has a reserved prefix in the middle of the location:
```
public static final String MANAGED_STORAGE_PREFIX = "__unitystorage";
```
So if a location has "/__unitystorage/" it can not be external and that is guaranteed by `validateNotOverlapWithManagedStorage`.

As for the other question "is this really a assigned managed location?" this can only be answered by matching location against registered staging table as mentioned by the other comment.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -183,14 +243,298 @@ public LoadTableResponse loadTable(
...
+      @Param("table") String table,
+      UpdateTableRequest request) {
+    String fullName = catalog + "." + namespace + "." + table;
+    boolean isCreateCommit =
+        request.requirements().stream()
+            .anyMatch(r -> r instanceof UpdateRequirement.AssertTableDoesNotExist);
+    if (isCreateCommit) {
+      return commitStagedCreate(catalog, namespace, table, request);
+    }
+    TableInfo tableInfo = getIcebergTableInfo(catalog, namespace, table);
```
yili-db: For non-create table update, please refer `updateTableForDelta` as reference.
1. Just like `updateTableForDelta` it's OK to put some the entire transaction orchestration in `TableRepository.java` instead. Then Iceberg specific code can live in dedicated Iceberg helper/utils and `TableRepository` can call them.
2. Use a single transaction before reading/writing table metadata and commit, then another transaction just for `setIcebergMetadataLocation`.
3. Prefer `findTableOrThrow` that returns `dao` (which has `getUniformIcebergMetadataLocation`) instead of `TableInfo`
4. After writing out the table metadata, besides setting the metadata location, the `dao` should also be updated with new schema / properties etc just like `createTable` sets them.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -183,14 +243,298 @@ public LoadTableResponse loadTable(
...
+      metadataService.deleteTableMetadata(newMetadataLocation);
+      if (e instanceof BaseException
+          && ((BaseException) e).getErrorCode() == ErrorCode.UPDATE_REQUIREMENT_CONFLICT) {
+        throw new CommitFailedException(
+            "Commit conflict on %s: %s", fullName, ((BaseException) e).getErrorMessage());
+      }
+      throw e;
+    }
+
+    TableMetadata committed = metadataService.readTableMetadata(newMetadataLocation);
```
yili-db: Can `updated` be reused?

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -183,14 +243,298 @@ public LoadTableResponse loadTable(
...
+      throw new BadRequestException(
+          "Table %s was not created through the Iceberg REST catalog; drop it through the Unity"
+              + " Catalog API instead.",
+          fullName);
+    }
+    // EXTERNAL tables leave data and metadata files in place (purgeRequested is accepted for
+    // spec compatibility but does not delete files); MANAGED tables have their storage directory
+    // removed by the repository's delete path.
+    tableService.deleteTable(fullName);
+    // 200 instead of the spec's 204: Armeria stalls body-less 204 responses on HTTP/1.1
```
yili-db: > Armeria stalls body-less 204 responses on HTTP/1.1

Is there a source for that? I search on internet and couldn't find it.
It seems that the apache/iceberg-rest-fixture returns 200 because it simply forgot to return 204.
Clients do accept 200 as well so 200 is also fine. But this "Armeria stalls" claim seems inaccurate.

server/src/main/java/io/unitycatalog/server/service/iceberg/IcebergSchemaConverter.java
```diff
@@ -0,0 +1,161 @@
...
+      case TIMESTAMP ->
+          ((Types.TimestampType) type).shouldAdjustToUTC() ? "timestamp" : "timestamp_ntz";
+      case STRING, UUID -> "string";
+      case FIXED, BINARY -> "binary";
+      case DECIMAL -> {
+        Types.DecimalType decimal = (Types.DecimalType) type;
+        yield String.format("decimal(%d,%d)", decimal.precision(), decimal.scale());
+      }
+      case VARIANT -> "variant";
+      default ->
```
yili-db, line 135: Add a TODO about supporting GEOMETRY/GEOGRAPHY, or a type specific error throwing.

server/src/main/java/io/unitycatalog/server/service/iceberg/IcebergSchemaConverter.java
```diff
@@ -0,0 +1,161 @@
...
+              ? ColumnTypeName.TIMESTAMP
+              : ColumnTypeName.TIMESTAMP_NTZ;
+      case STRING, UUID -> ColumnTypeName.STRING;
+      case FIXED, BINARY -> ColumnTypeName.BINARY;
+      case DECIMAL -> ColumnTypeName.DECIMAL;
+      case VARIANT -> ColumnTypeName.VARIANT;
+      case STRUCT -> ColumnTypeName.STRUCT;
+      case LIST -> ColumnTypeName.ARRAY;
+      case MAP -> ColumnTypeName.MAP;
+      default ->
```
yili-db, line 163: Add a TODO about supporting GEOMETRY/GEOGRAPHY, or a type specific error throwing.

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -21,4 +34,69 @@ public TableMetadata readTableMetadata(String metadataLocation) {
     return CompletableFuture.supplyAsync(() -> TableMetadataParser.read(fileIO, metadataLocation))
         .join();
   }
+
+  public void writeTableMetadata(TableMetadata tableMetadata, String metadataLocation) {
+    FileIO fileIO =
+        fileIOFactory.getFileIO(NormalizedURL.from(metadataLocation), CredentialContext.READ_WRITE);
+
+    CompletableFuture.runAsync(
```
yili-db: I've seen this pre-existing pattern of `runAsync` then `join` in this repo. But it is pointless.

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -21,4 +34,69 @@ public TableMetadata readTableMetadata(String metadataLocation) {
...
+          fileIOFactory.getFileIO(
+              NormalizedURL.from(metadataLocation), CredentialContext.READ_WRITE);
+      CompletableFuture.runAsync(() -> fileIO.deleteFile(metadataLocation)).join();
+    } catch (Exception e) {
+      // Orphaned metadata files are harmless; the commit outcome is decided by the catalog.
+    }
+  }
+
+  /** Builds the location of the next metadata file, following the common Iceberg naming scheme. */
+  public static String newMetadataLocation(TableMetadata tableMetadata, int version) {
```
yili-db: Return a NormalizedURL to explicitly normalize the location.

yili-db (changes requested):
Thanks for taking this on. This is a feature a lot of people have been asking for and we'd really like to get it in, so thanks for pushing it forward.

It's a big change and a lot of solid work went into it, so please don't be discouraged by the number of comments below. That's just what comes with something this size, not a sign anything's wrong.

I left 18 inline comments, a mix of small stuff and a few worth discussing. Happy to talk through any of them. If you can work through these and refresh the PR, I'll keep reviewing in passes until we get it merged.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
yili-db: Or is it actually caused by the bug being fixed by https://github.com/unitycatalog/unitycatalog/pull/1716?

server/src/main/java/io/unitycatalog/server/service/iceberg/FileIOFactory.java
iamkiayun: > Just FYI I have a PR to consolidate this file into `FileOperations` so there will be conflict.

Thank you for the heads-up. I kept the overlapping `FileIOFactory` surface narrow and did not duplicate the open #1687 consolidation. I’ll reconcile any remaining conflict after that PR lands.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > This (service calling another service) is the pre-existing anti-pattern and it even complicates things.

Updated in 0cbbdf9. Namespace operations now call the repositories directly and initialize hierarchical authorization inside the Iceberg service, removing the response aggregation and JSON round-trip.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > Catching this exception should be unnecessary. `IcebergRestExceptionHandler` should handle it already.

Updated in 0cbbdf9. The endpoint-local duplicate translation is removed, and repository error codes are now mapped centrally by `IcebergRestExceptionHandler`.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > Doesn't `IcebergRestExceptionHandler` already do this? If not, it should be fix instead.

Updated in 0cbbdf9. The local not-found wrapper is removed; the centralized handler now produces the Iceberg-shaped table/schema not-found and conflict responses.

server/src/main/java/io/unitycatalog/server/service/iceberg/TableConfigService.java
iamkiayun: > Make it explicit that privileges is always provided.

Updated in 0cbbdf9. The implicit overload was removed, so every `getTableConfig` caller now supplies its credential privileges explicitly.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > For `loadTable` of a native Iceberg table, it should start with `READ_WRITE` ... [and] fallback to `READ_ONLY`.

Implemented in 0cbbdf9. Native Iceberg loads receive read/write credentials for the table owner or callers with `SELECT + MODIFY`; other callers fall back to read-only. UniForm loads remain read-only. I also added focused tests for these cases.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > Please use `StagingTableRepository.createStagingTable` instead.

Implemented in 0cbbdf9. Managed Iceberg creates now use `createStagingTable`, preserve the staging UUID through final registration, and mark the staging row committed in the table-creation transaction.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > Use `NormalizedURL` instead.

Updated in 0cbbdf9. Create locations are normalized with `NormalizedURL` at the request boundary and remain normalized through the metadata and repository paths.

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
iamkiayun: > `FileOperations.createStorageLocationDir` can help with this.

Updated in 0cbbdf9. Table-location preparation now uses `FileOperations.createStorageLocationDir`, with idempotent handling when a local directory already exists.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > Prefer not calling tableService.

Updated in 0cbbdf9. The Iceberg service no longer depends on `TableService`; it calls `TableRepository` directly and performs the corresponding authorization hierarchy initialization/removal locally.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > Or even better, why not `writeTableMetadata` first then `CreateTable`.

Implemented in 0cbbdf9. Creation writes the initial metadata file first, then atomically registers the UC row and metadata pointer. If registration fails, only the unreferenced metadata file is cleaned up.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > `setIcebergMetadataLocation` can be folded with `CreateTable` in the same call and the same transaction.

Implemented in 0cbbdf9 with `createTableForIceberg`. The table row, columns, properties, staging commit, and initial metadata pointer are now persisted in one database transaction.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
iamkiayun: > Is this really an assigned managed location? This can only be answered by matching location against registered staging table.

Updated in 0cbbdf9. Managed classification now requires an exact match to an uncommitted staging-table location. Existing reserved managed-path overlap validation continues to prevent external locations from using the managed prefix.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > Besides setting the metadata location, the `dao` should also be updated with new schema / properties.

Implemented in 0cbbdf9. Commit now reads a detached repository state, then under a row lock performs the compare-and-swap and atomically updates the metadata pointer, UC columns, UC properties, and audit fields. A concurrent-commit regression test covers the conflict path.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > Can `updated` be reused?

Updated in 0cbbdf9. The commit response now reuses the already-built updated metadata rather than reading the newly written file back.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
iamkiayun: > Or is it actually caused by the bug being fixed by #1716?

Yes. After rebasing the merged #1716 change, I confirmed the issue was the no-body transcoder path; my earlier broad Armeria claim was inaccurate. The drop endpoints now return bodyless 204 responses, covered by the server tests.

server/src/main/java/io/unitycatalog/server/service/iceberg/IcebergSchemaConverter.java
iamkiayun: > Add a TODO about supporting GEOMETRY/GEOGRAPHY, or a type specific error throwing.

Updated in 0cbbdf9. This conversion direction now explicitly documents that Iceberg has no standard GEOMETRY/GEOGRAPHY primitive and keeps unsupported extensions rejected.

server/src/main/java/io/unitycatalog/server/service/iceberg/IcebergSchemaConverter.java
iamkiayun: > Add a TODO about supporting GEOMETRY/GEOGRAPHY, or a type specific error throwing.

Updated in 0cbbdf9. The UC type-name conversion path now has the same explicit unsupported handling for GEOMETRY/GEOGRAPHY.

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
iamkiayun: > The pre-existing pattern of `runAsync` then `join` ... is pointless.

Updated in 0cbbdf9. The async wrappers were removed; metadata reads, writes, and cleanup now use the synchronous Iceberg/FileIO operations directly.

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
iamkiayun: > Return a `NormalizedURL` to explicitly normalize the location.

Updated in 0cbbdf9. `newMetadataLocation` now returns `NormalizedURL`, and the metadata helper boundaries use it consistently.

iamkiayun:
@yili-db Thank you for the detailed review and for taking the time to work through this large feature with me. Your comments helped align the implementation more closely with the existing Unity Catalog architecture and strengthened the transaction and consistency guarantees.

I rebased the branch on the current `main` and pushed the review updates in `0cbbdf9`. The main changes are:

- replaced service-to-service calls with direct repository operations and local authorization hierarchy updates;
- centralized Iceberg REST exception translation;
- moved managed Iceberg creates onto the shared staging-table lifecycle;
- made initial table registration and metadata-pointer persistence atomic;
- made commits atomically synchronize the Iceberg pointer, UC columns, UC properties, and audit fields under a row lock;
- added permission-aware read/write credential fallback while keeping UniForm read-only;
- adopted `NormalizedURL` and shared file-operation helpers, removed async-then-block metadata wrappers, and restored bodyless `204` responses after rebasing #1716;
- expanded lifecycle, staging, concurrent-commit, metadata, schema/property synchronization, and credential tests.

Validation completed with 349 server tests: 347 passed, 0 failed, and 2 Docker-dependent suites were skipped because Docker is unavailable in my local environment. Formatting, checkstyle, shaded assembly, and `git diff --check` also pass.

I replied to each inline thread with the corresponding change and left the threads open for your confirmation. I appreciate your effort in helping review this contribution, and I’ll continue working through any further concerns or comments you raise in the next pass.

yili-db:
Thank you @iamkiayun . Will take another look.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -580,7 +714,20 @@ private TableInfoDAO findTableOrThrow(
...
   public TableInfo createTable(CreateTable createTable) {
-    return createTableImpl(createTable, Optional.empty(), (session, dao, tableInfo) -> tableInfo);
+    return createTableImpl(
+        createTable, Optional.empty(), Optional.empty(), (session, dao, tableInfo) -> tableInfo);
+  }
+
+  public TableInfo createTableForIceberg(CreateTable createTable, NormalizedURL metadataLocation) {
+    if (createTable.getDataSourceFormat() != DataSourceFormat.ICEBERG) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT, "createTableForIceberg requires ICEBERG format.");
```
yili-db: nit: this indicates an implementation bug not invalid input. The only caller is expected to set it and it does.
Use `ErrorCode.INTERNAL` instead.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -580,7 +714,20 @@ private TableInfoDAO findTableOrThrow(
   }
 
   public TableInfo createTable(CreateTable createTable) {
-    return createTableImpl(createTable, Optional.empty(), (session, dao, tableInfo) -> tableInfo);
+    return createTableImpl(
```
yili-db, line 703: Note that you have `if (createTable.getDataSourceFormat() != DataSourceFormat.ICEBERG)` for Iceberg call path. The other two callers of `createTableImpl` should check that `getDataSourceFormat` is **not** `ICEBERG`.

yili-db (commented):
Can you please add a server property `server.iceberg-table.enabled` and default to `false`? Then guard the create/write endpoints with it. This way we treat it as an experimental feature and promote it to production ready later.
For example:
https://github.com/unitycatalog/unitycatalog/blob/697c2fcfe4abcb1ac39184d284deb0c29d3819fb/server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java#L203

server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
```diff
@@ -36,6 +36,31 @@ private static Throwable getOriginalException(BaseException exception) {
 
   @Override
   protected BaseException toBaseException(Throwable cause) {
+    if (cause instanceof BaseException baseException) {
```
yili-db, line 44: This function `toBaseException` is meant to convert **other** exceptions to `BaseException`.
The whole conversion is simply `createErrorResponse(toBaseException(cause))`. `toBaseException` makes sure all exceptions are converted to `BaseException` if they aren't already one. Then `createErrorResponse` can uniformly write out error response according to the `BaseException`.

To correctly handle exceptions for Iceberg:

1. For `BaseException` thrown by UC server code, we need to eventually map them into Iceberg `ErrorResponse`:
  a. Make a new `enum IcebergErrorType { NO_SUCH_TABLE(404, "NoSuchTableException"), ...}`
  b. Each `ErrorCode` has a `DeltaErrorType deltaErrorType` defined. A new field `IcebergErrorType icebergErrorType` can be added.
  c. Finally `IcebergRestExceptionHandler.createErrorResponse` can use `BaseException` and the `ErrorCode` inside to generate the correct response (existing impl is not correct).
 
2. For Iceberg exceptions thrown by Iceberg code:
  a. Convert them into corresponding `BaseException` with correct `ErrorCode` (and correct `IcebergErrorType` inside).
  b. `IcebergRestExceptionHandler.createErrorResponse` writes out correct response the same way it does to server code thrown exceptions.
  c. The current implementation of `IcebergRestExceptionHandler.getOriginalException` is lame as it simply finds a non-`BaseException` even if it's not a Iceberg exception at all.

This can be done in a separate/concurrent PR.

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
```diff
@@ -40,6 +40,17 @@ private StagingTableDAO findByStagingLocation(Session session, NormalizedURL sta
     return query.uniqueResult(); // Returns null if no result is found
   }
 
+  public boolean isUncommittedStagingLocation(NormalizedURL stagingLocation) {
```
yili-db: This is TOCTOU-shaped read. The access to DB is not required since all this function does is to figure out "MANAGED or EXTERNAL". A simple `if (url.toString().contains(Constants.MANAGED_STORAGE_PREFIX))` could do the trick. The actual validation and ownership check is done by:
```
            // Find and commit staging table with the same staging location
            StagingTableDAO stagingTableDAO =
                repositories
                    .getStagingTableRepository()
                    .commitStagingTable(session, callerId, storageLocation);
```

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -662,33 +811,31 @@ private <T> T createTableImpl(
...
-                createTable.getStorageLocation());
             ExternalLocationUtils.validateNotOverlapWithManagedStorage(session, storageLocation);
             tableUUID = UUID.randomUUID();
           } else if (tableType == TableType.MANAGED) {
             storageLocation = NormalizedURL.from(createTable.getStorageLocation());
             serverProperties.checkManagedTableEnabled();
-            if (createTable.getDataSourceFormat() != DataSourceFormat.DELTA) {
+            if (createTable.getDataSourceFormat() != DataSourceFormat.DELTA
+                && createTable.getDataSourceFormat() != DataSourceFormat.ICEBERG) {
               throw new BaseException(
```
yili-db, line 679: why remove this comment?

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -383,6 +389,28 @@ private static void requireDeltaTable(
     }
   }
 
+  /**
+   * Acquire {@code SELECT ... FOR UPDATE} on the table row and refresh in-memory state, so two
+   * concurrent Iceberg REST commits serialize. Lock-wait timeouts and deadlock victims surface as
+   * {@code UPDATE_REQUIREMENT_CONFLICT} (409) instead of a generic 500.
+   */
+  private static void lockTableForUpdate(
```
yili-db: An identical function `lockTableForCommit` exists in `DeltaCommitRepository`.
Please move it to `RepositoryUtils` so that you can reuse it.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -56,23 +89,32 @@ public class IcebergRestCatalogService {
...
+          Endpoint.V1_LIST_TABLES,
+          Endpoint.V1_CREATE_NAMESPACE,
+          Endpoint.V1_CREATE_TABLE,
+          Endpoint.V1_UPDATE_TABLE,
+          Endpoint.V1_DELETE_TABLE);
 
-  private final SchemaService schemaService;
   private final TableConfigService tableConfigService;
   private final MetadataService metadataService;
+  private final CatalogRepository catalogRepository;
```
openinx, line 109: Please add a server property `server.iceberg-table.enabled` (default `false`) and gate the create/write/delete endpoints behind it. Also stop advertising `V1_CREATE_NAMESPACE` / `V1_CREATE_TABLE` / `V1_UPDATE_TABLE` / `V1_DELETE_TABLE` here when the flag is off, so clients do not discover write routes that will fail. Treat this as experimental until we promote it.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -580,7 +714,20 @@ private TableInfoDAO findTableOrThrow(
   }
 
   public TableInfo createTable(CreateTable createTable) {
-    return createTableImpl(createTable, Optional.empty(), (session, dao, tableInfo) -> tableInfo);
+    return createTableImpl(
+        createTable, Optional.empty(), Optional.empty(), (session, dao, tableInfo) -> tableInfo);
+  }
+
+  public TableInfo createTableForIceberg(CreateTable createTable, NormalizedURL metadataLocation) {
```
openinx, line 707: `createTableForIceberg` already asserts `DataSourceFormat.ICEBERG`. The other two `createTableImpl` callers (`createTable` / `createTableForDelta`) should reject `ICEBERG` — OpenAPI now exposes the enum value, so UC/Delta create can otherwise register a native Iceberg row without a metadata pointer.

server/src/main/java/io/unitycatalog/server/persist/StagingTableRepository.java
openinx: This is TOCTOU-shaped: a DB read only to decide MANAGED vs EXTERNAL. Prefer `url.toString().contains(Constants.MANAGED_STORAGE_PREFIX)` instead. Ownership / already-committed validation already happens in `commitStagingTable` when the create path runs.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
openinx: An identical helper already exists as `DeltaCommitRepository.lockTableForCommit`. Please move it to `RepositoryUtils` and reuse it from both call sites instead of duplicating.

openinx (dismissed):
Left comments on the remaining blockers.

Also: please rebase onto current `main`. [#1726](https://github.com/unitycatalog/unitycatalog/pull/1726) added the `isCloudStorageRoot` guard on EXTERNAL creates; this branch is missing it relative to `main` (merge-base predates that change, so it does not show as a deletion in the PR diff). Restoring / picking that up should be part of the rebase.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -566,6 +594,112 @@ public String getTableUniformMetadataLocation(
...
+                column.setId(UUID.randomUUID());
+                column.setTable(dao);
+              });
+          dao.getColumns().clear();
+          session.flush();
+          dao.getColumns().addAll(newColumns);
+          dao.setColumnCount(newColumns.size());
+
+          MutablePropertyMap properties = MutablePropertyMap.load(session, dao.getId());
+          properties.removeAll(new ArrayList<>(properties.asMap().keySet()));
```
yili-db, line 674: This replaces the whole UC property set from the Iceberg metadata instead of applying the commit's `SetProperties` / `RemoveProperties` incrementally like `DeltaUpdateTableMapper` does. That's fine as it keeps the UC row a faithful projection and won't drift if Iceberg adds new update types.

Worth a comment though, because it means UC should never persist a server-owned property on a native Iceberg table: the next commit erases anything not in the client's metadata. The Delta path is the opposite as that's the intentional design, so it's an easy thing to generalize wrongly later.

Non-blocking.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -566,6 +594,112 @@ public String getTableUniformMetadataLocation(
...
+          properties.putAll(tableProperties);
+          properties.flush(session, dao.getId());
+
+          dao.setUniformIcebergMetadataLocation(newMetadataLocation.toString());
```
yili-db, line 678: `uniform_iceberg_metadata_location` now holds two different things:
- the UniForm-derived pointer (Delta is authoritative)
- and the native Iceberg pointer (Iceberg is authoritative), told apart only by `data_source_format`.

Safe today, since the they belong to two different types of tables. But a comment on the DAO field would help, so nobody later reads non-null as "this is a UniForm table" and adds an unguarded reader.

Non-blocking, and **not** suggesting a rename.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db, line 679: why remove this comment?

server/src/main/java/io/unitycatalog/server/service/iceberg/IcebergSchemaConverter.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+    return columns;
+  }
+
+  private static ObjectNode structFieldJson(Types.NestedField field) {
+    ObjectNode node = MAPPER.createObjectNode();
+    node.put("name", field.name());
+    node.set("type", typeJson(field.type()));
+    node.put("nullable", field.isOptional());
+    node.set("metadata", MAPPER.createObjectNode());
```
yili-db: Maybe save the `comment` like this?
```
  ObjectNode metadata = MAPPER.createObjectNode();
  if (field.doc() != null) {
    metadata.put("comment", field.doc());
  }
  node.set("metadata", metadata);
```

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -1,24 +1,112 @@
...
+    }
+  }
+
+  /** Builds the location of the next metadata file, following the common Iceberg naming scheme. */
+  public static NormalizedURL newMetadataLocation(TableMetadata tableMetadata, int version) {
+    String metadataDirectory =
+        tableMetadata
+            .properties()
+            .getOrDefault(
+                TableProperties.WRITE_METADATA_LOCATION, tableMetadata.location() + "/metadata");
```
yili-db, line 95: I see that here both the value from `WRITE_METADATA_LOCATION` or `tableMetadata.location()` could be set by client request and be set to any random path.
The metadata location should be validated against the table location which is peristed in table dao. For example UniForm tables have this:
https://github.com/unitycatalog/unitycatalog/blob/0f1445227bd251b386420c90136515daefa4e03d/server/src/main/java/io/unitycatalog/server/service/delta/DeltaUniformUtils.java#L300
This is true for all Iceberg endpoints where clients can specify a metadata location in request.

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -1,24 +1,112 @@
...
+      }
+    }
+    return -1;
+  }
+
+  /**
+   * Returns the location form Iceberg's local FileIO reports after a read. Cloud locations remain
+   * normalized URIs; local file URIs become paths.
+   */
+  public static String toIcebergMetadataLocation(NormalizedURL metadataLocation) {
```
yili-db, line 154: Please add a unit test.

server/src/main/java/io/unitycatalog/server/service/iceberg/SimpleLocalFileIO.java
```diff
@@ -13,11 +17,29 @@ public InputFile newInputFile(String path) {
...
-    throw new UnsupportedOperationException();
+    File file = toLocalFile(path);
+    if (file.exists() && !file.delete()) {
+      throw new UncheckedIOException(new IOException("Failed to delete: " + path));
+    }
+  }
+
+  // Locations arrive both as plain paths and as file: URIs (matching what
+  // org.apache.iceberg.Files.localInput accepts).
+  private static File toLocalFile(String path) {
```
yili-db: Please unit test it.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -172,26 +226,297 @@ public LoadTableResponse loadTable(
...
+            request.schema(), spec, writeOrder, location.toString(), properties);
+
+    if (request.stageCreate()) {
+      // The permanent table is not registered and no metadata file is written. Managed creates
+      // retain their staging row so duplicate validation, temporary credentials, and later
+      // lifecycle management use the same path as other managed tables.
+      if (ucTableExists(catalog, namespace, request.name())) {
+        throw new AlreadyExistsException("Table already exists: %s.%s", namespace, request.name());
+      }
+      metadataService.prepareTableLocation(tableMetadata);
```
yili-db: This is pre-existing pattern:
```
void function(TableMetadata tableMetadata) {
  doSomething(tableMetadata.location());
```
This is not good:
1. From call site it's hard to understand that this function only needs the location, not anything else from `TableMetadata`
2. For the same reason as https://github.com/unitycatalog/unitycatalog/pull/1696/changes#r3727123916, it's best to have this function explicitly take a `table location` instead so that all these path and cred operations are explicitly limited to persisted table dao location.

So here it would be `.addAllConfig(tableConfigService.getTableConfig(location, READ_WRITE))` and much easier to reason about.

While you are here, please fix existing function as well. Thanks!

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -172,26 +226,297 @@ public LoadTableResponse loadTable(
...
+      String catalog,
+      String namespace,
+      String name,
+      TableType tableType,
+      TableMetadata tableMetadata,
+      boolean fromCommit) {
+    NormalizedURL metadataLocation = MetadataService.newMetadataLocation(tableMetadata, 0);
+    TableMetadata committed =
+        TableMetadata.buildFrom(tableMetadata)
+            .discardChanges()
```
yili-db, line 354: Please comment why `.discardChanges()` is called here

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -172,26 +226,297 @@ public LoadTableResponse loadTable(
...
+
+    NormalizedURL metadataLocation = NormalizedURL.from(state.metadataLocation());
+    TableMetadata base = metadataService.readTableMetadata(metadataLocation);
+    request.requirements().forEach(requirement -> requirement.validate(base));
+
+    TableMetadata.Builder builder = TableMetadata.buildFrom(base);
+    request.updates().forEach(update -> update.applyTo(builder));
+    TableMetadata updatedWithoutLocation = builder.build();
+    if (updatedWithoutLocation.changes().isEmpty()) {
+      return LoadTableResponse.builder()
```
yili-db, line 429: Please comment "No-op update"

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -172,26 +226,297 @@ public LoadTableResponse loadTable(
...
+          .withTableMetadata(base)
+          .addAllConfig(tableConfigService.getTableConfig(base, READ_WRITE))
+          .build();
+    }
+
+    NormalizedURL newMetadataLocation =
+        MetadataService.newMetadataLocation(
+            updatedWithoutLocation, MetadataService.parseMetadataVersion(metadataLocation) + 1);
```
yili-db: This is another case of https://github.com/unitycatalog/unitycatalog/pull/1696/changes#r3727123916 that the user input location could come from persisted Iceberg commit file as well.
The same principal applies here too: the only location server should trust client having access to in all the Iceberg endpoints is the location persisted in table dao.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -172,26 +226,297 @@ public LoadTableResponse loadTable(
...
+  private LoadTableResponse commitStagedCreate(
+      String catalog, String namespace, String table, UpdateTableRequest request) {
+    String fullName = catalog + "." + namespace + "." + table;
+    for (UpdateRequirement requirement : request.requirements()) {
+      if (!(requirement instanceof UpdateRequirement.AssertTableDoesNotExist)) {
+        throw new BadRequestException(
+            "Invalid requirement for a create commit: %s", requirement.getClass().getSimpleName());
+      }
+    }
+    if (ucTableExists(catalog, namespace, table)) {
```
yili-db: This is redundant as it's handled by https://github.com/unitycatalog/unitycatalog/pull/1696/changes#diff-cd833ce691a9b116826023c1b4d2d8a7baf1faab9178a20ce03a1ccff89c3eacR350 already

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -172,26 +226,297 @@ public LoadTableResponse loadTable(
...
+        request.writeOrder() == null ? SortOrder.unsorted() : request.writeOrder();
+    TableMetadata tableMetadata =
+        TableMetadata.newTableMetadata(
+            request.schema(), spec, writeOrder, location.toString(), properties);
+
+    if (request.stageCreate()) {
+      // The permanent table is not registered and no metadata file is written. Managed creates
+      // retain their staging row so duplicate validation, temporary credentials, and later
+      // lifecycle management use the same path as other managed tables.
+      if (ucTableExists(catalog, namespace, request.name())) {
```
yili-db: For managed table, this is already checked by `stagingTableRepository.createStagingTable`

yili-db (commented):
Left comments. Please take a look.

For a major feature implementation like this, multiple rounds of review with many comments are very much expected. You contribution is very welcomed! Thank you!

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
yili-db: Also a related PR https://github.com/unitycatalog/unitycatalog/pull/1749

iamkiayun:
Thanks for the detailed review. I addressed the actionable feedback in commit 976592f:

- Added the opt-in `server.iceberg-table.enabled=false` flag, gated Iceberg write operations, and stopped advertising write endpoints while the flag is disabled.
- Corrected the table-format guards: `createTableForIceberg` reports an internal caller bug for an impossible wrong-format call, while the generic and Delta create paths reject native `ICEBERG` tables.
- Removed the TOCTOU-shaped staging lookup. Managed/external classification now uses the reserved managed-storage prefix, while ownership and committed-state validation remain in the staging transaction. The redundant managed duplicate check was removed.
- Moved the shared commit lock helper into `RepositoryUtils` and reused it from the Delta and Iceberg paths.
- Validated metadata locations, table roots, and `WRITE_METADATA_LOCATION` against the location persisted in the table DAO. Credential vending now also uses the persisted location, with unit coverage for accepted and rejected paths.
- Preserved Iceberg field documentation in schema conversion and added comments explaining property replacement, the dual-use metadata pointer, staging validation, no-op updates, and discarded builder changes.
- Added local FileIO write coverage and enabled local metadata output support required by the Iceberg write path.

I intentionally left the broader `IcebergErrorType`/exception-handler redesign for a focused follow-up PR, as suggested, so this change stays scoped.

The focused local suite passes (48/48), compilation and checkstyle pass, and the DCO check is green. The full integration class is currently blocked in my local clean-build environment by a test-runtime/shaded-assembly issue, so I will rely on the upstream CI environment for that coverage.

Thanks again for the careful review and for helping improve the design.

yili-db:
Note that the local FS access is now happening from server and is unbounded. This is fine in this PR because it's gated by both the `server.iceberg-table.enabled` (default `false`) and `#authorize(#principal, #metastore, OWNER)`.
I will do a follow up to address this issue after this PR.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -578,7 +693,58 @@ private TableInfoDAO findTableOrThrow(
...
+  }
+
+  /**
+   * Renames a table within its schema. Looks up the source table first (throwing {@link
+   * ErrorCode#TABLE_NOT_FOUND} if absent) and rejects a collision with an existing table of the
+   * target name in the same schema (throwing {@link ErrorCode#TABLE_ALREADY_EXISTS}). Renaming a
+   * table to its current name is an idempotent no-op. Metadata-only: the table UUID, schema, and
+   * storage location are all unchanged.
+   */
+  public void renameTable(String catalog, String schema, String table, String newName) {
```
yili-db: Looks like `renameTable` has been moved accidentally within this file.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -607,17 +775,26 @@ public DeltaLoadTableResponse createTableForDelta(
                 createTable.getName()));
   }
 
+  private static void requireNonIcebergFormat(CreateTable createTable, String caller) {
+    if (createTable.getDataSourceFormat() == DataSourceFormat.ICEBERG) {
+      throw new BaseException(
+          ErrorCode.UNSUPPORTED_TABLE_FORMAT,
+          caller + " cannot create native ICEBERG tables; use createTableForIceberg instead.");
```
yili-db: nit: this error message could be more informative by saying "use IRC endpoints instead"

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -1,30 +1,181 @@
...
     this.fileOperations = fileOperations;
   }
 
-  /**
-   * Reads and parses the Iceberg {@link TableMetadata} at the given metadata-file location, using a
-   * credential-vended {@link org.apache.iceberg.io.FileIO} obtained from {@link FileOperations}.
-   *
-   * @param metadataLocation the normalized location of the {@code *.metadata.json} file
-   */
   public TableMetadata readTableMetadata(NormalizedURL metadataLocation) {
```
yili-db: Now this can be made private

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -1,30 +1,181 @@
...
+   */
+  public TableMetadata readTableMetadata(
+      NormalizedURL metadataLocation, NormalizedURL persistedTableLocation) {
+    validateMetadataLocation(metadataLocation, persistedTableLocation);
+    TableMetadata tableMetadata = readTableMetadata(metadataLocation);
+    validateTableMetadataLocation(tableMetadata, persistedTableLocation);
+    return tableMetadata;
+  }
+
+  public void writeTableMetadata(TableMetadata tableMetadata, NormalizedURL metadataLocation) {
```
yili-db: This function is no longer used by any prod code? Consider removing it.

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -1,30 +1,181 @@
...
+      TableMetadata tableMetadata, NormalizedURL persistedTableLocation) {
+    validateTableMetadataLocation(tableMetadata, persistedTableLocation);
+    NormalizedURL location = persistedTableLocation;
+    createStorageLocationDirIfAbsent(location);
+    createStorageLocationDirIfAbsent(NormalizedURL.from(location + "/metadata"));
+    createStorageLocationDirIfAbsent(NormalizedURL.from(location + "/data"));
+  }
+
+  /** Best-effort cleanup of a metadata file that lost a commit race or whose commit failed. */
+  public void deleteTableMetadata(NormalizedURL metadataLocation) {
```
yili-db: make it private or just fold with the two-arg version

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -1,30 +1,181 @@
...
+    try (FileIO fileIO = fileOperations.getFileIO(metadataLocation, CredentialContext.READ_WRITE)) {
+      TableMetadataParser.write(tableMetadata, fileIO.newOutputFile(metadataLocation.toString()));
+    }
+  }
+
+  /**
+   * For local file locations, pre-creates the table's directory layout so staged creates can write
+   * data and manifest files before the first metadata commit. Object stores have no directories.
+   */
+  public void prepareTableLocation(TableMetadata tableMetadata) {
```
yili-db: remove?

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -1,30 +1,181 @@
...
+
+  /** Best-effort cleanup constrained to the persisted table location. */
+  public void deleteTableMetadata(
+      NormalizedURL metadataLocation, NormalizedURL persistedTableLocation) {
+    validateMetadataLocation(metadataLocation, persistedTableLocation);
+    deleteTableMetadata(metadataLocation);
+  }
+
+  /** Builds the location of the next metadata file, following Iceberg's naming scheme. */
+  public static NormalizedURL newMetadataLocation(TableMetadata tableMetadata, int version) {
```
yili-db: make private

yili-db (changes requested):
This PR is close to be ready for merging. This round I was able to find only a few minor issues and nits.
Also please rebase and check the failing CI test.

Thank you!

yili-db:
Hi @iamkiayun, are you still going to finish this PR? I think it's very close to be getting merged. Just rebase and a few minor fixes.
Thanks!

yili-db (approved):

iamkiayun:
Hi @yili-db — apologies for the delayed reply; I was tied up over the weekend. Yes, I’m still committed to finishing this PR. I saw that you applied the remaining minor fixes, merged the latest `main` into the branch, and approved the current head. I’m monitoring the remaining Spark 4.0 check and will address any issue it surfaces. I’ll also stay responsive to any further comments.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -570,6 +576,115 @@ public String getTableUniformMetadataLocation(
...
+          session.merge(dao);
+          return null;
+        },
+        "Failed to update Iceberg metadata location for " + fullName,
+        /* readOnly = */ false);
+  }
+
   private TableInfoDAO findTableOrThrow(
       Session session, String catalogName, String schemaName, String tableName) {
     UUID schemaId =
```
tdas, line 690: [AI-assisted]

native iceberg metadata location lands in uniformIcebergMetadataLocation — the column now covers two concepts (delta-derived uniform vs native iceberg). can we at least doc the dual use on the DAO field?

'uniform' is a delta-UniForm concept; a future maintainer reads the name, assumes delta-only, and either mishandles native iceberg rows or adds a redundant column — the name lies about what it holds.

server/src/main/java/io/unitycatalog/server/service/iceberg/MetadataService.java
```diff
@@ -1,30 +1,169 @@
...
+    }
+  }
+
+  /**
+   * For local file locations, pre-creates the table's directory layout so staged creates can write
+   * data and manifest files before the first metadata commit. Object stores have no directories.
+   * Prepares the directories only after validating the table's client-supplied locations.
+   */
+  public void prepareTableLocation(
+      TableMetadata tableMetadata, NormalizedURL persistedTableLocation) {
```
tdas, line 65: [AI-assisted]

we should never eat an error without logging, even best-effort cleanup. log at warn before swallowing.

tdas (approved):
Two comments... not blockers. Can be addressed in follow up PRs.

yili-db:
@iamkiayun @dataders 
Thank you very much for your contribution! This is a major feature implementation and it took major effort to get it done. Your work is very much appreciated and we also look forward for more contribution from you if you are interested.

dataders:
@yili-db WOO HOOOOOO! LFG! Thanks for taking this over the line!

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: Both comments being followed up at https://github.com/unitycatalog/unitycatalog/pull/1802
Thank you!

iamkiayun:
Thank you for the review, fixes, and merge, @yili-db! I really appreciate your guidance and help getting this contribution across the finish line.

PR #1698, opened 2026-07-21
Fix JCasbinAuthorizer reading the wrong Hibernate username property

### LINKED ISSUE #1697: JCasbinAuthorizer reads wrong Hibernate property (hibernate.connection.user vs hibernate.connection.username)
**What happened**

When `server.authorization=enable` and Unity Catalog is backed by an external database that authenticates on the connection username (e.g. PostgreSQL), the server crashes at startup with only:

```
io.unitycatalog.server.UnityCatalogServer - Initializing JCasbinAuthorizer...
Exception in thread "main" io.unitycatalog.server.exception.BaseException: Problem initializing authorizer.
        at io.unitycatalog.server.UnityCatalogServer.initializeAuthorizer(UnityCatalogServer.java:152)
        at io.unitycatalog.server.UnityCatalogServer.initializeServer(UnityCatalogServer.java:122)
        at io.unitycatalog.server.UnityCatalogServer.<init>(UnityCatalogServer.java:92)
        at io.unitycatalog.server.UnityCatalogServer$Builder.build(UnityCatalogServer.java:435)
        at io.unitycatalog.server.UnityCatalogServer.main(UnityCatalogServer.java:356)
                                                                                                                                               
```
There is no Caused by: — the actual root cause is invisible.

It looks like `JCasbinAuthorizer` reads the wrong Hibernate property name for the DB username. at line 41 (server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java) 

The standard Hibernate property is `hibernate.connection.username`. Because `JCasbinAuthorizer` reads `hibernate.connection.user`, it gets null for any normal config. It then passes that null straight to the casbin JDBC adapter. This causes the JDBC driver to fallback to a default (e.g. the OS/process username for the PostgreSQL driver). 

The main Hibernate SessionFactory connects fine (it reads hibernate.connection.username correctly and uses HikariCP), so the DB is clearly reachable — but the casbin adapter's separate connection is made under the wrong identity.

Also, `initializeAuthorizer` swallows the exception cause, making the above error — and any other authorizer init failure — undiagnosable. (server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java) (lines 146-153):
```
try {
    LOGGER.info("Initializing JCasbinAuthorizer...");
    UnityCatalogAuthorizer authorizer = new JCasbinAuthorizer(hibernateConfigurator);
    new UnityAccessUtil(repositories).initializeAdmin(authorizer);
    return authorizer;
  } catch (Exception e) {
    throw new BaseException(ErrorCode.INTERNAL, "Problem initializing authorizer.");
  } 
```
                                                                                                                                                
The caught `e` is never chained into the BaseException (no cause arg, no initCause), so the underlying SQL/auth exception is lost.

**To Reproduce**
1. Configure UC with an external PostgreSQL DB in hibernate.properties using hibernate.connection.username=<user> (the standard property) and username-sensitive auth (e.g. cert clientcert=verify-full, or scram/md5 with a pg_hba rule keyed to the role).
2. Set server.authorization=enable.
3. Start the server → crashes with Problem initializing authorizer and no cause. DB logs show the casbin connection attempting to authenticate as the process/OS user instead of the configured username.

**Expected behavior**

1. Error from `JCasbinAuthorizer` should be properly raised back up to caller
2. SIngle use of `hibernate.connection.username` or `hibernate.connection.user` (not both)

mvanhorn:
## What this fixes

`JCasbinAuthorizer` reads the database username from `hibernate.connection.user`, but the standard Hibernate property (used by the main session factory configuration and by the project's own tests) is `hibernate.connection.username`. With a standard configuration the casbin JDBC adapter received a `null` username, and the JDBC driver then silently fell back to a process default, crashing authorizer startup against an external database that authenticates on the username. Reported in #1697.

## Root cause

Two independent property spellings are in play in the codebase:

- `hibernate.connection.username` — the standard Hibernate key, set by the main session-factory config and by the existing tests.
- `hibernate.connection.user` — a non-standard key the deployment docs and Helm chart document.

`JCasbinAuthorizer` only read the second one, so any deployment using the standard spelling handed the JDBC adapter a `null` username. Separately, `UnityCatalogServer.initializeAuthorizer` swallowed the caught exception when rethrowing, so the original bug report had no `Caused by:` to point at the real failure.

## The fix

- Resolve the username from `hibernate.connection.username` first and fall back to `hibernate.connection.user`, so both spellings work.
- Preserve the caught exception as the cause when rethrowing from `initializeAuthorizer`, so the underlying failure is no longer hidden.
- Add unit tests covering username resolution: standard key, legacy fallback, and standard-wins-when-both-present.

All 13 tests in `JCasbinAuthorizerTest` pass locally via `build/sbt "server / testOnly io.unitycatalog.server.auth.JCasbinAuthorizerTest"`.

Fixes #1697

---
AI was used for assistance.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
+++ b/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
@@ -149,7 +149,7 @@ public class UnityCatalogServer {
         new UnityAccessUtil(repositories).initializeAdmin(authorizer);
         return authorizer;
       } catch (Exception e) {
-        throw new BaseException(ErrorCode.INTERNAL, "Problem initializing authorizer.");
+        throw new BaseException(ErrorCode.INTERNAL, "Problem initializing authorizer.", e);
       }
     } else {
       LOGGER.info("Authorization disabled. Using AllowingAuthorizer.");
--- a/server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
@@ -38,7 +38,7 @@ public class JCasbinAuthorizer implements UnityCatalogAuthorizer {
     Properties properties = hibernateConfigurator.getHibernateProperties();
     String driver = properties.getProperty("hibernate.connection.driver_class");
     String url = properties.getProperty("hibernate.connection.url");
-    String user = properties.getProperty("hibernate.connection.user");
+    String user = resolveConnectionUsername(properties);
     String password = properties.getProperty("hibernate.connection.password");
     JDBCAdapter adapter = new JDBCAdapter(driver, url, user, password);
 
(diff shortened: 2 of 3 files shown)


yili-db (approved):

mvanhorn:
Thank you @yili-db - JCasbinAuthorizer reading the right Hibernate username property fixes authz on renamed configs.

Issue #1694, opened 2026-07-20
Hadoop connector credential support for multi-storage location tables.


foss-contributor:
With the introduction of the Delta REST API, there can now be a table credential vend that returns a list of credentials. Currently, the connector discards any additional credentials that do not fall under the table’s own storage location. However, in the future, there may be such tables that require more than one storage location, hence more than one credential. The Hadoop connector must be updated to support multiple credentials in the future.

This change also adds stronger API response validation (for both UC REST API and Delta REST API) to ensure that there will always be at least one valid credential vended. 

An additional behavioral change applies to the Delta REST API credential vending path. If there is only one credential vended, this credential will be used regardless of whether or not it actually matches the accessed storage location. This is a relaxation of the constraints for the Delta REST API code path. However, the justification is that the client has not choice over which credential to use (only one credential) and the server is the authoritative source of credentials (and therefore the credential must work). Note that this applies to all code paths, including whether or not if `CredScopedFileSystem` is enabled.

To enable support for multi-location tables, please enable `CredScopedFileSystem` and `renewable credential`. This is the user's responsibility.

PR #1701, opened 2026-07-22
[hadoop]: change credential fetcher response type to list instead of single credential

foss-contributor:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1701/files) to review incremental changes.
- [**stack/select-vended-credential-by-location**](https://github.com/unitycatalog/unitycatalog/pull/1701) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1701/files)] ← _this PR_
  - [stack/scoped-credential-namespaces](https://github.com/unitycatalog/unitycatalog/pull/1704) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1704/files/cc071433385b61e553486e11715f1c15785b65ae..f8de7d7e9a2b33da67ec6fd74e1478a02cae1468)]
    - [stack/encode-scoped-vended-credentials](https://github.com/unitycatalog/unitycatalog/pull/1705) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1705/files/f8de7d7e9a2b33da67ec6fd74e1478a02cae1468..a048e325dd999a74cef5e8cfeffa9d8a74bbe31e)]

---------
## Summary

- Change credential fetchers from returning one `GenericCredential` to returning `List<GenericCredential>`. Standard UC responses remain singleton lists, while Delta table and staging-table fetchers preserve all vended credentials.
- Select the longest location-covering credential when multiple credentials are returned, and reject empty responses, missing selection locations, or responses without a covering credential.
- Generalize credential caches for list-valued initial credentials and selected provider credentials. Renewal refetches the list and repeats location selection.
- Cache the selected provider credential using the location-aware delegate filesystem identity while driver-side property construction continues requiring exactly one credential.

## Testing

- Added provider coverage for longest-prefix selection, shared caching, cache bypass, renewal/reselection, and invalid responses.
- Added selection coverage for path boundaries and cloud-scheme aliases.
- Updated UC and Delta fetcher tests for ordered list-valued responses and updated dependent cache, property, and Spark fixtures.

## Issue


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -188,9 +192,14 @@ private static Map<String, String> createCredProps(
       // Unsupported scheme: skip the credential fetch entirely and return no props.
       return Collections.emptyMap();
     }
-    GenericCredential cred =
-        fetchGenericCredential(
+    List<GenericCredential> credentials =
+        fetchGenericCredentials(
             hadoopConf, apiClient, catalogUri, tokenProvider, appVersions, credId);
+    Preconditions.checkState(
```
openinx, line 199: Rejecting `size != 1` breaks today’s Delta behavior. Fetchers used to `selectForLocation(credId.location(), …)` before returning one credential; now a multi-cred response fails planning. Issue #1694 says multi-cred responses already happen and the connector currently picks the covering one.

Why fail here instead of selecting with the Delta/staging `location` already on `CredId`, until the encode PR lands? If this layer must not ship alone, please say that explicitly in the PR.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
```diff
@@ -82,10 +85,25 @@ private GenericCredential renewCredential() throws ApiException {
...
+      return fetchAndSelectCredential();
     }
   }
+
+  private GenericCredential fetchAndSelectCredential() throws ApiException {
+    List<GenericCredential> credentials = genericCredentialFetcher().createCredentials();
+    Preconditions.checkState(!credentials.isEmpty(), "No vended credential was returned.");
+    if (credentials.size() == 1) {
+      return credentials.get(0);
+    }
```
openinx, line 108: Singleton responses skip coverage checks (`singleCredentialIsUsedWithoutMatchingLocation`). Old Delta path failed when the only credential did not cover the location. Why relax that for Delta too?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
```diff
@@ -82,10 +85,25 @@ private GenericCredential renewCredential() throws ApiException {
...
+    Preconditions.checkState(
+        cacheKey.prefix() != null,
+        "Multiple credentials were vended but no location is set to select one.");
+    return CredentialUtil.selectForLocation(cacheKey.prefix(), credentials);
```
openinx, line 115: Renewal selects via `FileSystemCredId.prefix()` (`UC_CREDENTIAL_PREFIX_KEY`), not `DeltaTableCredId.location()` / `UC_DELTA_LOCATION_KEY`. What breaks if multi-cred renewal runs with Delta location set but prefix unset?

openinx (commented):
Left comments on the incremental stacked range (`8a4c0ff1..436d76d5`).

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/CredPropsUtilTest.java
```diff
@@ -27,6 +31,31 @@ void credContextIdVariesByCatalogUriSchemeAndAuth() {
         .isNotEqualTo(base);
   }
 
+  @Test
+  void multipleVendedCredentialsAreRejectedOnDriver() {
```
openinx: nit: rename this test so it does not say "driver".

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
```diff
@@ -56,34 +56,31 @@ public static GenericCredential toGenericCredential(DeltaStorageCredential cred)
...
     int bestLen = -1;
-    for (DeltaStorageCredential c : creds) {
-      if (c == null || c.getPrefix() == null || !prefixCovers(location, c.getPrefix())) {
+    for (GenericCredential cred : creds) {
+      String prefix = cred.prefix();
+      if (prefix == null || !prefixCovers(location, prefix)) {
         continue;
       }
-      int len = stripTrailingSlashes(c.getPrefix()).length();
+      int len = stripTrailingSlashes(prefix).length();
```
openinx, line 77: The prefixCovers already stripped the trailing slash for the `prefix`, and here we do another trailing slash again, which seems unnecessary.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
```diff
@@ -74,41 +78,43 @@ int size() {
...
-   * later readers observe the same renewal behavior.
-   */
-  public static class RenewableCredential {
+  /** Decides whether {@code value} is close enough to expiry that it should be renewed. */
+  @FunctionalInterface
+  public interface Staleness<T> {
+    boolean readyToRenew(T value, Clock clock, long renewalLeadTimeMillis);
+  }
+
+  public static class RenewableCredential<T> {
```
openinx: Do you think making this class as `abstracted` and provide the `readyToRenew()` will actually simply the interface that we want to expose to the caller ?  the current PR i see two problems here: 

1.  we are passing the dup `Clock` and `renewalLeadTimeMillis` for both `Staleness` interface and `RenewalableCredential` class, and what if those are not matching ?? then it's a risk to be buggy. 

2.  What if the `Stateless<T>` does not match the `RenewableCredential` constructor ?  that's also a risk.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialFetcher.java
```diff
@@ -24,7 +25,11 @@
  * <p><b>Internal API — not for external use. May change without notice.</b>
  */
 public interface GenericCredentialFetcher {
-  GenericCredential createCredential() throws ApiException;
+  /**
+   * Fetches the vended credentials for this fetcher's scope. A single-credential API returns a
```
openinx: > A single-credential API returns a
   * one-element list.

This is unncessary and seems to be verbose.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
```diff
@@ -81,10 +85,25 @@ private GenericCredential renewCredential() throws ApiException {
...
-      return genericCredentialFetcher().createCredential();
+      return fetchAndSelectCredential();
     }
   }
+
+  private GenericCredential fetchAndSelectCredential() throws ApiException {
+    List<GenericCredential> credentials = genericCredentialFetcher().createCredentials();
+    Preconditions.checkState(!credentials.isEmpty(), "No vended credential was returned.");
+    if (credentials.size() == 1) {
+      return credentials.get(0);
```
openinx, line 107: the intellij actually suggest to use `credentials.getFirst()`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaGenericCredentialFetcher.java
```diff
@@ -39,7 +41,8 @@ public GenericCredential createCredential() throws ApiException {
         id.catalog(),
         id.schema(),
         id.table());
-    return CredentialUtil.toGenericCredential(
-        CredentialUtil.selectForLocation(credId.location(), response.getStorageCredentials()));
+    return response.getStorageCredentials().stream()
+        .map(CredentialUtil::toGenericCredential)
```
openinx, line 53: few comments for the CredentialUtil.toGenericCredential, but unrelated to this PR:

1. `toGenericCredential`, there is no `isGcsCred` check when we assume it's a `GcsCredential`.   and for those unexpected cases, we should throw. 

2.  `requireSingleCloudConfig`,  we'd better to reuse all the existing `isS3Config -> isS3Cred` / `isAzureConfig -> isAzureCred`/ `isGcsCred` methods. 

3. `toGenericCredential`,  i suggest to define those `isXXCred` for `TemporaryCredentials` also.  that help us to use the same methods as the `DeltaStorageCredential`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
```diff
@@ -6,14 +6,15 @@
...
+import java.util.List;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 /** Adapts the UC Delta staging table credentials SDK API for Hadoop token providers. */
 final class UCDeltaStagingTableCredentialFetcher implements GenericCredentialFetcher {
 
   private final DeltaTemporaryCredentialsApi api;
   private final UUID stagingTableId;
-  private final String stagingTableLocation;
```
openinx, line 16: If you are removing the `stagingTableLocation` here, why don't we also remove the `location` in the `DeltaStagingTableCredId` and even the `location` parameter in `createDeltaStagingTableCredProps` ?  

and that will even bring us a new issue, do we still need to support the public API that `unitycatalog-hadoop` module provides since `location` will be never used  ?  then we will need to deprecate the existing one, and add another one with only the `stagingTableId` parameter.

```
/**
     * Builds Hadoop properties for a <em>Delta staging table</em> using the UC Delta staging table
     * credentials API ({@code GET /delta/v1/staging-tables/{table_id}/credentials}).
     *
     * @param stagingTableId the staging table UUID
     * @param location the staging table storage location, used to select the matching storage
     *     credential from the response
     * @return unmodifiable map; empty if the scheme is unrecognized
     * @throws IllegalArgumentException if {@code stagingTableId}, {@code location}, or {@code
     *     tokenProvider} is missing
     * @throws ApiException if the credential fetch from the UC Delta API fails
     */
    public Map<String, String> buildForStagingTable(String stagingTableId, String location)
        throws ApiException {
```

consider all the context,  I suggest you open another separate PR to finish this removal.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
openinx: If you do want to make the 1 cred support in this PR, pls explicitly add the TODO in those line.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/AbfsVendedTokenProviderTest.java
```diff
@@ -20,12 +20,22 @@ protected AbfsVendedTokenProvider createTestProvider(
...
+  @Override
+  protected AbfsVendedTokenProvider createTestProvider(
+      Configuration conf, GenericCredentialFetcher fetcher) {
+    return new TestAbfsVendedTokenProvider(conf, fetcher);
+  }
+
   static class TestAbfsVendedTokenProvider extends AbfsVendedTokenProvider {
     private final GenericCredentialFetcher credentialFetcher;
 
     TestAbfsVendedTokenProvider(Configuration conf, TemporaryCredentialsApi mockApi) {
```
openinx: why don't you remove this constructor directly, and making all the mock method reuse the below override constructor directly ?  as the `TestAbfsVendedTokenProvider(Configuration conf, GenericCredentialFetcher credentialFetcher)` is always more abstracted to implement any kind of mocking credential ?

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/CredPropsUtilTest.java
openinx: This test sets `CredPropsUtil.genericCredFetcherFactory` and never resets it. `CredPropsBaseTest` / `CredPropsCacheTest` / `CredPropsRoundTripTest` all restore the factory in `@AfterEach`. Please add the same reset, or later tests in the same JVM can get poisoned.

nit: also rename so the method does not say "driver".

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/CredentialUtilTest.java
```diff
@@ -150,20 +110,27 @@ void selectorNormalizesTrailingSlashes() {
...
+    List<GenericCredential> creds =
+        Arrays.asList(credAt(null), credAt("s3://other"), credAt("s3://bucket/sibling"));
     assertThatThrownBy(() -> CredentialUtil.selectForLocation("s3://bucket/t", creds))
         .isInstanceOf(IllegalArgumentException.class)
-        .hasMessageContaining("No UC Delta credential matched");
+        .hasMessageContaining("No vended credential covers location");
+    assertThatThrownBy(() -> CredentialUtil.selectForLocation("S3A://bucket/t", creds))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessageContaining("No vended credential covers location");
   }
```
openinx: We dropped several `selectForLocation` edge-case tests that still matter for the new `List<GenericCredential>` API:
- empty list (and null list if that can reach it)
- single credential that does not cover the location (or cover this only via the provider shortcut, intentionally)
- null / prefix-less entries in a multi-cred list (old code skipped nulls; new code NPEs on null `cred`)

Please restore the ones that still match intended behavior, or add provider/fetcher tests that replace them.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/CredentialUtilTest.java
```diff
@@ -150,20 +110,27 @@ void selectorNormalizesTrailingSlashes() {
...
-    List<DeltaStorageCredential> creds =
-        Arrays.asList(credAt("s3://other"), credAt("s3://bucket/sibling"));
+  void selectForLocationThrowsWhenNoPrefixCovers() {
+    List<GenericCredential> creds =
+        Arrays.asList(credAt(null), credAt("s3://other"), credAt("s3://bucket/sibling"));
     assertThatThrownBy(() -> CredentialUtil.selectForLocation("s3://bucket/t", creds))
         .isInstanceOf(IllegalArgumentException.class)
-        .hasMessageContaining("No UC Delta credential matched");
+        .hasMessageContaining("No vended credential covers location");
+    assertThatThrownBy(() -> CredentialUtil.selectForLocation("S3A://bucket/t", creds))
```
openinx: The `S3A://…` case does not prove scheme non-aliasing: this fixture has no covering `s3://bucket/t` credential, so failure is expected even with scheme normalization. Did you mean something like one cred at `s3://bucket/t` and location `s3a://bucket/t/file`, expecting no match?

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/BaseTokenProviderTest.java
```diff
@@ -70,6 +84,119 @@ public void after() {
...
+  }
+
+  @Test
+  public void singleCredentialIsUsedWithoutMatchingLocation() {
+    GenericCredential only = newGenericCred("only", Long.MAX_VALUE, location("/other"));
+
+    T provider = provider(location("/t/file"), only);
+
+    assertThat(provider.accessCredentials()).isSameAs(only);
+  }
```
openinx, line 147: This locks in “singleton skips coverage.” Is that the intended contract for Delta too, or only UC REST? If intentional, fine; if not, the test is encoding the wrong behavior.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/UCDeltaGenericCredentialFetcherTest.java
```diff
@@ -44,17 +45,70 @@ void createCredentialCallsDeltaApiWithCredIdFieldsAndReturnsCredential() throws
...
     assertThat(cred.secretAccessKey()).isEqualTo("sk");
     assertThat(cred.sessionToken()).isEqualTo("st");
     assertThat(cred.expirationTimeMillis()).isEqualTo(789L);
+    assertThat(cred.prefix()).isEqualTo("s3://bucket/events");
     verify(api)
         .getTableCredentials(DeltaCredentialOperation.READ_WRITE, "main", "default", "events");
   }
 
+  @Test
+  void createCredentialsReturnsAllVendedCredentialsInOrder() throws Exception {
```
openinx, line 66: nit: table vs staging `createCredentialsReturnsAllVendedCredentialsInOrder` tests are near-duplicates; a small shared helper would shrink noise.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/UCDeltaGenericCredentialFetcherTest.java
```diff
@@ -44,17 +45,70 @@ void createCredentialCallsDeltaApiWithCredIdFieldsAndReturnsCredential() throws
...
+    AwsCredential cred1 = (AwsCredential) creds.get(0);
+    assertThat(cred1.accessKeyId()).isEqualTo("ak1");
+    assertThat(cred1.prefix()).isEqualTo("s3://bucket/events");
+    AwsCredential cred2 = (AwsCredential) creds.get(1);
+    assertThat(cred2.accessKeyId()).isEqualTo("ak2");
+    assertThat(cred2.prefix()).isEqualTo("s3://bucket/events/child");
+  }
+
   @Test
   void createCredentialRejectsMissingDeltaCredentialsResponse() throws Exception {
```
openinx: nit: method still named `createCredentialRejects…` after the API rename to `createCredentials`.

Also missing negative coverage for null/empty `getStorageCredentials()` (pairs with the production null-check gap).

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcherTest.java
```diff
@@ -30,17 +31,64 @@ void createCredentialCallsDeltaStagingApiAndReturnsCredential() throws Exception
...
+    AwsCredential cred1 = (AwsCredential) creds.get(0);
+    assertThat(cred1.accessKeyId()).isEqualTo("ak1");
+    assertThat(cred1.prefix()).isEqualTo(LOCATION);
+    AwsCredential cred2 = (AwsCredential) creds.get(1);
+    assertThat(cred2.accessKeyId()).isEqualTo("ak2");
+    assertThat(cred2.prefix()).isEqualTo(LOCATION + "/child");
+  }
+
   @Test
   void createCredentialRejectsNullResponse() throws Exception {
```
openinx: nit: rename `createCredentialRejectsNullResponse` to match `createCredentials`.

Same gap as the table fetcher: please add coverage for null/empty `getStorageCredentials()`.

openinx (commented):
Test-focused follow-up comments.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
openinx: Pls explicitly add the comment to attach the context that we've discussed offline. I mean for multip cred cases, we should set the `fs.unitycatalog.credential.prefix` explicitly in the `CredPropsBuilder.build()`.  while for 1 single cred, we also should set it  but we want to skip the prefix selection directly and use the only one cred.

openinx: if possible,  pls also context in the `UC_CREDENTIAL_PREFIX_KEY`

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaGenericCredentialFetcher.java
foss-contributor: I'll file this as follow up PR

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
foss-contributor: Let's do a follow up PR for future release.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/BaseTokenProviderTest.java
foss-contributor: Yes, locking in contract that when there is only credential, we use it regardless of if it actually covers the requested storage location. I'll update the GitHub Issue to reflect this.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
foss-contributor: This is the intended new behavior, updated it in the GitHub issue.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
foss-contributor: Requirement is if multi cred then must set, otherwise optional. Updated comments. I'll update the GitHub issue as well.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
foss-contributor: Let's keep each PR self contained; Will fix

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/auth/AbfsVendedTokenProviderTest.java
foss-contributor: Done, cleaned up for GCP and AWS as well

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/CredentialCache.java
openinx: Hi @foss-contributor ,  can you take a look at this comment also ?

foss-contributor: Fixing it right now

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/GenericCredentialProvider.java
```diff
@@ -80,11 +84,34 @@ private GenericCredential renewCredential() throws ApiException {
     if (credCacheEnabled) {
       return globalCache.access(
           cacheKey,
-          () ->
-              new RenewableCredential(
-                  renewalLeadTimeMillis, clock, genericCredentialFetcher().createCredential()));
+          () -> {
+            GenericCredential credential = fetchAndSelectCredential();
+            return new RenewableCredential<GenericCredential>(credential) {
```
openinx: nit:  this `GenericCredential` can be removed in the class new, since it will automatically infer the class type.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -242,14 +251,21 @@ private static GenericCredential fetchGenericCredential(
...
-        () ->
-            new RenewableCredential(
-                renewalLeadTimeMillis,
-                ClockUtil.resolveClock(hadoopConf),
-                createCredential(apiClient, catalogUri, tokenProvider, appVersions, credId)));
+        () -> {
+          Clock clock = ClockUtil.resolveClock(hadoopConf);
+          List<GenericCredential> credentials =
+              createCredentials(apiClient, catalogUri, tokenProvider, appVersions, credId);
+          return new RenewableCredential<List<GenericCredential>>(credentials) {
```
openinx: nit: This `List<GenericCredential>` can also be removed now.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
openinx: then for this PR, we will need to revert the `stagingTableLocation` deletion,  pls don't do the unnecessary removal in this PR, and let's defer to the follow-up PR.

openinx (approved):
Looks good to me now, but pls address the new 3 minor comments.  I will merge it once everything is fine.   Thanks @foss-contributor for the contribution.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/auth/UCDeltaStagingTableCredentialFetcher.java
foss-contributor: yes

PR #1702, opened 2026-07-22
[Spark] Fix ClassCastException on namespace ops when DeltaCatalog is absent

yili-db:
**PR Checklist**

**Description of changes**
When `DeltaCatalog` is not on the classpath (e.g. Spark 4.2, which has no matching Delta release yet), namespace commands like `SHOW`/`CREATE`/`DESC`/`DROP SCHEMA` failed with:
```
  java.lang.ClassCastException: class io.unitycatalog.spark.UCProxy cannot be cast to class org.apache.spark.sql.connector.catalog.DelegatingCatalogExtension
```
The namespace methods cast `delegate` to `DelegatingCatalogExtension`, which only holds when DeltaCatalog is loaded. Without it, `delegate` is `ucProxy`, a plain `TableCatalog with SupportsNamespaces`, so the cast throws.

Route namespace ops through a `namespaceCatalog` accessor that casts to `SupportsNamespaces` instead. Both possible delegates satisfy it: DeltaCatalog via CatalogExtension, and ucProxy directly.

Add a regression test that forces the no-Delta path and reproduces the failure without the fix.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -118,6 +118,16 @@ class UCSingleCatalog
     }
   }
 
+  /**
+   * The catalog to route `SupportsNamespaces` operations to.
+   *
+   * Both possible values of `delegate` are a `SupportsNamespaces`: when DeltaCatalog is on the
+   * classpath `delegate` is a `DelegatingCatalogExtension` (a `SupportsNamespaces` via
+   * `CatalogExtension`), and when it is absent `delegate` is `ucProxy` itself (a plain
+   * `TableCatalog with SupportsNamespaces`). So we can cast unconditionally.
+   */
+  private def namespaceCatalog: SupportsNamespaces = delegate.asInstanceOf[SupportsNamespaces]
+
   /** See [[DeltaVersionUtils.isDeltaRestApiReady]] for the predicate. */
   private def shouldUseDeltaAPI: Boolean =
     DeltaVersionUtils.isDeltaRestApiReady(deltaCatalogLoaded, deltaRestApiEnabled)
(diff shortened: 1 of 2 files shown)


connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -118,6 +118,16 @@ class UCSingleCatalog
...
 
+  /**
+   * The catalog to route `SupportsNamespaces` operations to.
+   *
+   * Both possible values of `delegate` are a `SupportsNamespaces`: when DeltaCatalog is on the
+   * classpath `delegate` is a `DelegatingCatalogExtension` (a `SupportsNamespaces` via
+   * `CatalogExtension`), and when it is absent `delegate` is `ucProxy` itself (a plain
+   * `TableCatalog with SupportsNamespaces`). So we can cast unconditionally.
+   */
+  private def namespaceCatalog: SupportsNamespaces = delegate.asInstanceOf[SupportsNamespaces]
```
openinx, line 129: do you think we should name it as `asNamespaceCatalog` ?

openinx: okay, that's not a strong opinion.  I saw the following code usage, actually it also makes sense for use the use the `namespaceCatalog` directly.

connectors/spark/src/test/java/io/unitycatalog/spark/SchemaOperationsTest.java
```diff
@@ -12,25 +12,12 @@
...
 import org.apache.spark.sql.catalyst.analysis.NamespaceAlreadyExistsException;
 import org.apache.spark.sql.catalyst.analysis.NoSuchNamespaceException;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 public class SchemaOperationsTest extends BaseSparkIntegrationTest {
-  @Test
-  public void testCreateSchema() {
```
yili-db, line 19: This is now consolidated as the new testNamespaceOps

openinx (approved):

PR #1637, opened 2026-06-29
Add VIEW table type support in server

TheThow:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [x] If there is a related issue, make sure it is linked to this PR.
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

Adds type VIEW as Table type to support views.
Result of the suggestion in my previous PR: https://github.com/unitycatalog/unitycatalog/pull/1311
Related issue: https://github.com/unitycatalog/unitycatalog/issues/1279

view can be created via CLI
```
bin/uc table create \
  --full_name unity.default.my_view \
  --columns "as_int INT, as_string STRING" \
  --table_type VIEW \
  --view_definition "SELECT as_int, as_string FROM unity.default.numbers" \
  --view_dependencies '[{"table": {"table_full_name": "unity.default.numbers"}}]' \
  --output jsonPretty
```

View dependencies are validated whether they exist, however there is currently no validation whether any or all view dependencies were specified. I'd avoid this for now as this requires SQL parsing.

This implementation works with the spark connector integration follow-up

Follow-Ups:
* spark connector integration: https://github.com/unitycatalog/unitycatalog/pull/1639
* iceberg integration: TBD
* Improve dependency validation: TBD

PR #1703, opened 2026-07-22
[Docs] Add user-facing docs for metric views (Spark 4.2)

chenwang-databricks:
## What this PR does

Adds user-facing documentation for **metric views**, now supported by the Unity Catalog Spark
connector on Apache Spark 4.2 (created via `CREATE VIEW ... WITH METRICS LANGUAGE YAML`).

- **New page** `docs/usage/views.md` (**Usage → Metric Views**) covering the full lifecycle:
  - Prerequisites (Spark 4.2+, `unitycatalog-spark_4.2_2.13`)
  - Launching Spark SQL against UC with the 4.2 connector
  - Creating a source table and a metric view (YAML `dimensions` / `measures`)
  - `SHOW VIEWS` / `SHOW TABLES` (a metric view appears on both surfaces on Spark 4.2)
  - `DESCRIBE` / `DESCRIBE EXTENDED` (columns + `METRIC_VIEW` type)
  - Querying a measure with `measure(...)` + `GROUP BY`
  - `DROP VIEW`, and notes that rename is not yet supported
- Wires the page into the mkdocs nav.
- Cross-links the new page from the **Apache Spark** integration page.

The SQL examples and console output mirror the behavior exercised by `MetricViewE2ETest`
(added in #1647): source `events(region, cnt)` → `measure(cnt_sum) GROUP BY region` → `eu=3, us=3`.

## Notes

- Metric views require Apache Spark 4.2+; the page states this in its prerequisites and the
  Spark integration page links to it with the same caveat.
- Docs only — no code changes.

### DIFF
--- a/docs/integrations/unity-catalog-spark.md
+++ b/docs/integrations/unity-catalog-spark.md
@@ -28,6 +28,11 @@ data stored in Unity Catalog through a controlled mechanism.
     | Apache Spark 4.0.x | `io.unitycatalog:unitycatalog-spark_4.0_2.13` | `io.delta:delta-spark_4.0_2.13` |
     | Apache Spark 4.1.x | `io.unitycatalog:unitycatalog-spark_4.1_2.13` | `io.delta:delta-spark_4.1_2.13` |
 
+!!! tip "Metric views"
+    On Apache Spark 4.2 and later, the connector also supports Unity Catalog **metric views** —
+    reusable dimensions and measures defined over a source table or SQL query. See
+    [Metric Views](../usage/metric-views.md).
+
 ## Download and Configure Unity Catalog for Apache Spark
 
 The following steps are required to download and configure Unity Catalog for Apache Spark.
--- a/docs/usage/metric-views.md
+++ b/docs/usage/metric-views.md
@@ -1,0 +1,212 @@
+# Unity Catalog Metric Views
+
+This page shows you how to create, query, and manage **metric views** in Unity Catalog using
+Apache Spark™.
+
+A metric view is a view-like object that defines reusable **dimensions** and **measures** (named
+aggregations) over a source — either a table/view or a SQL query — expressed in YAML. Consumers
+query the measures with the `measure(...)` function and group by the dimensions, so the aggregation
+logic lives in one governed place in Unity Catalog instead of being copy-pasted across queries.
+
+!!! warning "Prerequisites"
+    Metric views require **Apache Spark 4.2 or later** (the `CREATE VIEW ... WITH METRICS` SQL DDL
+    was introduced in Spark 4.2) and the **Unity Catalog 0.6.0** Spark 4.2 connector. Use the
+    Spark 4.2 connector artifact:
+
+    | Spark version | UC Spark connector artifact |
+    | --- | --- |
+    | Apache Spark 4.2.x | `io.unitycatalog:unitycatalog-spark_4.2_2.13` |
+
+    Metric views are not available on Spark 4.0.x / 4.1.x. For general Spark setup (downloading
+    Spark, configuring cloud storage, and the shell launch commands), see the
+    [Apache Spark integration](../integrations/unity-catalog-spark.md) page.
+
+## Launch Spark SQL against Unity Catalog
+
+Start a Spark SQL shell wired to your local Unity Catalog server with the Spark 4.2 connector:
+
+```sh
+export CATALOG_NAME=unity
+export UC_URI=http://localhost:8080
+export UC_TOKEN=
+
+bin/spark-sql --name "metric-view-test" \
+    --master "local[*]" \
+    --packages "io.unitycatalog:unitycatalog-spark_4.2_2.13:0.6.0" \
+    --conf "spark.sql.catalog.$CATALOG_NAME=io.unitycatalog.spark.UCSingleCatalog" \
+    --conf "spark.sql.catalog.$CATALOG_NAME.uri=$UC_URI" \
+    --conf "spark.sql.catalog.$CATALOG_NAME.token=$UC_TOKEN" \
+    --conf "spark.sql.defaultCatalog=$CATALOG_NAME"
+```
+
+!!! note "No Delta package here"
+    Unlike the examples on the [Apache Spark integration](../integrations/unity-catalog-spark.md)
+    page, this command deliberately omits the `delta-spark` package and the
+    `spark.sql.catalog.spark_catalog=...DeltaCatalog` conf, because the walkthrough below uses a
+    parquet source table. The connector logs a benign `WARN ... DeltaCatalog is not available in
+    the classpath`, which is expected. A managed Delta source table is not an option on Spark 4.2
+    yet, since no `delta-spark_4.2_2.13` artifact is published.
+
+## Create a source table
+
+A metric view computes over an existing source. Create a parquet external table to use as the
+source (this needs no Delta package and works with the launch command above):
+
+```sql
+CREATE TABLE unity.default.events (region STRING, cnt INT)
+USING parquet LOCATION '/tmp/uc_events_src';
+
+INSERT INTO unity.default.events VALUES ('us', 1), ('us', 2), ('eu', 3);
+```
+
+!!! note
+    A `CREATE TABLE` without a `LOCATION` (a managed table) fails with `Unity Catalog does not
+    support non-Delta managed table.` — use an external (`LOCATION`) parquet table as shown, or a
+    managed Delta table on a Spark version where `delta-spark` is available.
+
+## Create a metric view
+
+Use `CREATE VIEW ... WITH METRICS LANGUAGE YAML` to define the metric view. The YAML body declares:
+
+- `source` — the data the metrics are computed over. This can be a **table or view**
+  (`unity.default.events`) or a **SQL query** (`SELECT ... FROM ...`).
+- `filter` *(optional)* — a row filter applied to the source, like a `WHERE` clause.
+- `dimensions` — the grouping columns.
+- `measures` — the named aggregations.
+
+The example below computes metrics over the `events` table created above:
+
+```sql
+CREATE VIEW unity.default.events_metrics
+WITH METRICS
+LANGUAGE YAML
+AS $$
+version: "0.1"
+source: unity.default.events
+filter: cnt > 0
+dimensions:
+  - name: region
+    expr: region
+measures:
+  - name: cnt_sum
+    expr: sum(cnt)
+$$;
+```
+
+To compute metrics over a query instead of a table, set `source` to a SQL statement, for example
+`source: SELECT region, cnt FROM unity.default.events WHERE region <> 'unknown'`.
+
+## List metric views
+
+A metric view appears on the **view** surface via `SHOW VIEWS`:
+
+```sql
+SHOW VIEWS IN unity.default;
+```
+
+```console
++---------+--------------+-----------+
+|namespace|      viewName|isTemporary|
++---------+--------------+-----------+
+|  default|events_metrics|      false|
++---------+--------------+-----------+
+```
+
+It also appears on the **table** surface via `SHOW TABLES`, alongside the source table:
+
+```sql
+SHOW TABLES IN unity.default;
+```
+
+```console
++---------+--------------+-----------+
+|namespace|     tableName|isTemporary|
++---------+--------------+-----------+
+|  default|        events|      false|
+|  default|events_metrics|      false|
++---------+--------------+-----------+
+```
+
+## Inspect a metric view
+
+`DESCRIBE` returns the metric view's dimension and measure columns:
+
+```sql
+DESCRIBE unity.default.events_metrics;
+```
+
+```console
++--------+---------+-------+
+|col_name|data_type|comment|
++--------+---------+-------+
+|  region|   string|   NULL|
+| cnt_sum|   bigint|   NULL|
++--------+---------+-------+
+```
+
+`DESCRIBE EXTENDED` additionally reports the detailed view information, including that the object
+is a `METRIC_VIEW`:
+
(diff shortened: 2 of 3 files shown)


docs/integrations/unity-catalog-spark.md
```diff
@@ -28,6 +28,11 @@ data stored in Unity Catalog through a controlled mechanism.
     | Apache Spark 4.0.x | `io.unitycatalog:unitycatalog-spark_4.0_2.13` | `io.delta:delta-spark_4.0_2.13` |
     | Apache Spark 4.1.x | `io.unitycatalog:unitycatalog-spark_4.1_2.13` | `io.delta:delta-spark_4.1_2.13` |
 
+!!! tip "Metric views"
+    On Apache Spark 4.2 and later, the connector also supports Unity Catalog **metric views** —
+    reusable dimensions and measures defined over a source table. See
```
chenwang-databricks: Not just a source table, but it can also be a SQL statement, a SQL query.

docs/integrations/unity-catalog-spark.md
```diff
@@ -28,6 +28,11 @@ data stored in Unity Catalog through a controlled mechanism.
...
+    [Metric Views](../usage/views.md).
```
chenwang-databricks: Change the name of the file to Metric views.md.

docs/usage/views.md
```diff
@@ -0,0 +1,167 @@
+# Unity Catalog Metric Views
+
+This page shows you how to create, query, and manage **metric views** in Unity Catalog using
+Apache Spark™.
+
+A metric view is a view-like object that defines reusable **dimensions** and **measures** (named
+aggregations) over a source table, expressed in YAML. Consumers query the measures with the
```
chenwang-databricks: Same here. It can be a source table or a SQL query.

docs/usage/views.md
```diff
@@ -0,0 +1,167 @@
...
+`measure(...)` function and group by the dimensions, so the aggregation logic lives in one governed
+place in Unity Catalog instead of being copy-pasted across queries.
+
+!!! warning "Prerequisites"
+    Metric views require **Apache Spark 4.2 or later** (the `CREATE VIEW ... WITH METRICS` SQL DDL
+    was introduced in Spark 4.2 via [SPARK-56920](https://issues.apache.org/jira/browse/SPARK-56920))
```
chenwang-databricks: Don't link Spark PR here. If you could find public documentation for Spark Metric views, link it here. If not, just don't link anything. Just say it was introduced in Spark before too.

docs/usage/views.md
```diff
@@ -0,0 +1,167 @@
...
+    and **Unity Catalog 0.6.0 or later**. Use the Spark 4.2 connector artifact:
```
chenwang-databricks: It should be 0.5.0: link to this doc: https://github.com/unitycatalog/unitycatalog/releases#release-v0.5.0

docs/usage/views.md
```diff
@@ -0,0 +1,167 @@
...
+A metric view reads from an existing table. The metric-view definition resolves its `source:`
+relation eagerly, so create the source first:
+
+```sql
+CREATE TABLE unity.default.events (region STRING, cnt INT)
+USING parquet
+LOCATION '/tmp/tables/events';
+
+INSERT INTO unity.default.events VALUES ('us', 1), ('us', 2), ('eu', 3);
+```
```
chenwang-databricks, line 60: Remove

docs/usage/views.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Create a metric view
+
+Use `CREATE VIEW ... WITH METRICS LANGUAGE YAML` to define the metric view. The YAML body declares
+the `source`, the `dimensions` (grouping columns), and the `measures` (named aggregations):
```
chenwang-databricks: Point out that the source can be a table or a SQL, and we also support filter field as well (works like a where condition in a query.)

docs/usage/views.md
```diff
@@ -0,0 +1,167 @@
...
+
+```console
++---------+--------------+-----------+
+|namespace|      viewName|isTemporary|
++---------+--------------+-----------+
+|  default|events_metrics|      false|
++---------+--------------+-----------+
+```
+
+It also appears on the **table** surface via `SHOW TABLES` (Spark 4.2 lists tables and views
```
chenwang-databricks: Remove "Spark 4.2 lists tables and views together, alongside source table"

docs/usage/metric-views.md
```diff
@@ -0,0 +1,167 @@
...
+```console
++--------+---------+-------+
+|col_name|data_type|comment|
++--------+---------+-------+
+|  region|   string|   NULL|
+| cnt_sum|   bigint|   NULL|
++--------+---------+-------+
+```
+
+`DESCRIBE EXTENDED` additionally reports the detailed view information, including that the object
```
chenwang-databricks, line 147: Get an example output

chenwang-databricks:
Thanks for the review — addressed all the comments in the latest push:

- **Source can be a table or a SQL query**: updated the intro (both here and the Spark-integration cross-link) and the "Create a metric view" section to state that `source` accepts a table/view **or** a SQL query, with an example of the SQL form.
- **`filter` field**: documented the optional `filter` field (acts like a `WHERE` on the source) and added it to the example YAML.
- **Renamed the file**: `docs/usage/views.md` → `docs/usage/metric-views.md` (nav + cross-link updated).
- **Removed the Spark PR/JIRA link**: there's no public Apache Spark documentation page for metric views yet, so the doc now just says the DDL was introduced in Spark 4.2 with no link.
- **Removed the standalone "Create a source table" section** and folded a minimal note into the metric-view example.
- **Removed** the "Spark 4.2 lists tables and views together..." phrasing from the `SHOW TABLES` section.
- **Added example output** for `DESCRIBE EXTENDED` (showing `Type: METRIC_VIEW` and the view details).

**On the version** (you asked me to verify): `0.5.0` is the current UC **server** release, but I checked and neither the server-side VIEW support (#1637) nor the Spark 4.2 connector (#1675) is in any released tag — both `v0.5.0` and `v0.5.1` ship only `unitycatalog-spark_4.0`/`_4.1`, and no `unitycatalog-spark_4.2_2.13` exists on Maven yet. Those land in **0.6.0** (current `version.sbt` is `0.6.0-SNAPSHOT`). So the metric-view connector requires **UC 0.6.0**, which is what the prerequisites and the `--packages` coordinate now say. Happy to adjust if 4.2 support is planned for a 0.5.x instead.

docs/usage/metric-views.md
```diff
@@ -0,0 +1,181 @@
...
+Start a Spark SQL shell wired to your local Unity Catalog server with the Spark 4.2 connector:
+
+```sh
+export CATALOG_NAME=unity
+export UC_URI=http://localhost:8080
+export UC_TOKEN=
+
+bin/spark-sql --name "metric-view-test" \
+    --master "local[*]" \
+    --packages "io.unitycatalog:unitycatalog-spark_4.2_2.13:0.6.0" \
```
openinx, line 35: Does it work as we don't add the delta related artifacts in this `--packages` ?  I'm not sure, just want to confirm.

openinx: Ideally,  I suggest to have an e2e integration tests to covert this basic user story, so that we can verify the artifacts and catalog configs we set here are actually working and wont' be broken in future. 

that may need a separate github workflow to run the pipeline.

docs/usage/metric-views.md
```diff
@@ -0,0 +1,181 @@
...
+
+Use `CREATE VIEW ... WITH METRICS LANGUAGE YAML` to define the metric view. The YAML body declares:
+
+- `source` — the data the metrics are computed over. This can be a **table or view**
+  (`unity.default.events`) or a **SQL query** (`SELECT ... FROM ...`).
+- `filter` *(optional)* — a row filter applied to the source, like a `WHERE` clause.
+- `dimensions` — the grouping columns.
+- `measures` — the named aggregations.
+
+The example below computes metrics over an `events(region STRING, cnt INT)` table in
```
openinx: The page never shows how to create the `events` source table, and the obvious guess fails. Following the page as written, a reader runs `CREATE TABLE unity.default.events (region STRING, cnt INT);` and hits:

```
io.unitycatalog.client.ApiException: Unity Catalog does not support non-Delta managed table.
```

Please add an explicit source-table step before this example. A parquet external table (the same approach `MetricViewE2ETest.createEventsSource` uses to keep the source setup independent of the Delta create path) works with the exact launch command shown above and needs no Delta:

```sql
CREATE TABLE unity.default.events (region STRING, cnt INT)
USING parquet LOCATION '/tmp/uc_events_src';
INSERT INTO unity.default.events VALUES ('us', 1), ('us', 2), ('eu', 3);
```

I verified the full walkthrough end-to-end on Spark 4.2.0 after adding this step.

docs/usage/metric-views.md
openinx, line 35: nit: this launch command intentionally omits the Delta package and `spark.sql.catalog.spark_catalog=...DeltaCatalog` conf that the other examples on the integration page include. That is correct for a parquet/external source (the connector just logs a benign `WARN ... DeltaCatalog is not available`). Worth a one-line note that this is deliberate -- and that a managed Delta source table is not an option on Spark 4.2 yet, since no `delta-spark_4.2_2.13` artifact is published.

docs/usage/metric-views.md
```diff
@@ -0,0 +1,181 @@
...
+|col_name                   |data_type         |comment|
++---------------------------+------------------+-------+
+|region                     |string            |NULL   |
+|cnt_sum                    |bigint            |NULL   |
+|                           |                  |       |
+|# Detailed View Information|                  |       |
+|Catalog                    |unity             |       |
+|Database                   |default           |       |
+|Table                      |events_metrics    |       |
+|Type                       |METRIC_VIEW       |       |
```
openinx: nit: this `DESCRIBE EXTENDED` sample is trimmed -- the real output also includes a `View` row and a `Properties [metric_view.from.name=..., metric_view.where=cnt > 0]` row. Fine as an illustration; just flagging that it is not verbatim.

openinx (commented):
I built the `0.6.0-SNAPSHOT` Spark 4.2 connector and ran this whole page against a local UC server on Spark 4.2.0. The feature behaves exactly as documented: `SHOW VIEWS` / `SHOW TABLES` (dual surface), `DESCRIBE [EXTENDED]` -> `METRIC_VIEW`, `SELECT ... measure(cnt_sum) GROUP BY region` -> `eu=3` / `us=3`, and `DROP VIEW`. Nice work. A couple of doc gaps that block a first-time reader are inline below.

Also, unrelated to the file contents: please drop the `Co-authored-by` trailer from the commit message before merge.

chenwang-databricks:
Thanks for building the connector and running the whole page end-to-end, @openinx — and for confirming the 0.6.0 / Spark 4.2.0 behavior. Addressed the doc gaps in the latest push:

- **Missing source-table step** (the blocker): added an explicit **Create a source table** section before the metric-view example, using the parquet external table you suggested (matching `MetricViewE2ETest.createEventsSource`):
  ```sql
  CREATE TABLE unity.default.events (region STRING, cnt INT) USING parquet LOCATION '/tmp/uc_events_src';
  INSERT INTO unity.default.events VALUES ('us', 1), ('us', 2), ('eu', 3);
  ```
  Also added a note that a managed (`no LOCATION`) `CREATE TABLE` fails with `Unity Catalog does not support non-Delta managed table.`, so readers don't hit that.
- **Deliberate no-Delta launch command**: added a note under the launch command explaining that omitting the `delta-spark` package and `spark_catalog=...DeltaCatalog` conf is intentional for a parquet source (benign `WARN ... DeltaCatalog is not available`), and that a managed Delta source isn't an option on Spark 4.2 yet (no `delta-spark_4.2_2.13` published).
- **`DESCRIBE EXTENDED` sample**: made it more faithful (added the `View` row and the `Properties [metric_view.from.name=..., metric_view.where=cnt > 0]` row that your `filter: cnt > 0` produces) and flagged it as illustrative / trimmed, not verbatim.
- **Commit message**: dropped the `Co-authored-by` trailer.

On the **packaged-artifact e2e integration test** — good idea to guard the documented `--packages` / catalog config against future breakage. I'd like to keep this docs PR focused and do that as a **separate follow-up** (a dedicated workflow that smoke-tests the metric-view story against the published `unitycatalog-spark_4.2_2.13` once 0.6.0 is released). The connector behavior itself is already covered by `MetricViewE2ETest` (which you validated against your local 0.6.0-SNAPSHOT build).

openinx (approved):
Looks good to me now, thanks @chenwang-databricks  !

docs/usage/metric-views.md
```diff
@@ -0,0 +1,212 @@
...
+    --conf "spark.sql.catalog.$CATALOG_NAME.token=$UC_TOKEN" \
+    --conf "spark.sql.defaultCatalog=$CATALOG_NAME"
+```
+
+!!! note "No Delta package here"
+    Unlike the examples on the [Apache Spark integration](../integrations/unity-catalog-spark.md)
+    page, this command deliberately omits the `delta-spark` package and the
+    `spark.sql.catalog.spark_catalog=...DeltaCatalog` conf, because the walkthrough below uses a
+    parquet source table. The connector logs a benign `WARN ... DeltaCatalog is not available in
+    the classpath`, which is expected. A managed Delta source table is not an option on Spark 4.2
```
tdas: does it have to be managed? cant we define metrics views on any Delta table?

tdas:
```suggestion
    the classpath`, which is expected. A Delta source table is not an option on Spark 4.2
```

tdas (approved):
approved but please fix my one comment.

PR #1704, opened 2026-07-23
[Hadoop] Select namespaced credentials by location

foss-contributor:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1704/files) to review incremental changes.
- [**stack/scoped-credential-namespaces**](https://github.com/unitycatalog/unitycatalog/pull/1704) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1704/files)] ← _this PR_
  - [stack/encode-scoped-vended-credentials](https://github.com/unitycatalog/unitycatalog/pull/1705) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1705/files/fbc9dc5a47c525d311d5e11a8f62b629494f913e..9733d97a6a10df04b74cadffce1415c6fa0957d2)]

---------
## Summary

- Add indexed Hadoop configuration namespaces for scoped credentials.
- Select the namespace whose location is the longest covering prefix for the requested URI, then overlay only that credential onto the delegate filesystem configuration.
- Preserve legacy and single-credential behavior while isolating delegate filesystem cache entries by the selected credential location.
- Validate namespace counts and reject incomplete or non-covering multi-credential configurations.

## Testing

- Added unit coverage for longest-prefix selection, legacy and single-credential layouts, multi-credential overlays, malformed configurations, source configuration immutability, and cache isolation.

## Issue

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
@@ -11,44 +11,30 @@ import io.unitycatalog.hadoop.internal.auth.GcsCredential;
 import io.unitycatalog.hadoop.internal.auth.GenericCredential;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.List;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /** Internal utilities for building and selecting {@link GenericCredential}s. */
 public final class CredentialUtil {
   private static final Charset UTF_8 = StandardCharsets.UTF_8;
+  private static final Base64.Encoder ENCODER = Base64.getEncoder();
+  private static final Base64.Decoder DECODER = Base64.getDecoder();
 
   private CredentialUtil() {}
 
   public static String[] encodeMultiCredPrefixes(List<String> prefixes) {
-    Preconditions.checkArgument(
-        prefixes != null && !prefixes.isEmpty(),
-        "List of credential prefixes cannot be null or empty");
-    List<String> encodedPrefixes = new ArrayList<>(prefixes.size());
-    Base64.Encoder encoder = Base64.getEncoder();
-    for (String prefix : prefixes) {
-      Preconditions.checkArgument(
-          prefix != null && !prefix.isEmpty(), "Credential prefixes cannot be null or empty");
-      encodedPrefixes.add(encoder.encodeToString(prefix.getBytes(UTF_8)));
-    }
-    return encodedPrefixes.toArray(String[]::new);
+    return prefixes.stream()
+        .map(prefix -> ENCODER.encodeToString(prefix.getBytes(UTF_8)))
+        .toArray(String[]::new);
   }
 
   public static List<String> decodeMultiCredPrefixes(String[] prefixes) {
-    Preconditions.checkArgument(
-        prefixes != null && prefixes.length > 0,
-        "Encoded credential prefixes cannot be null or empty");
-    Base64.Decoder decoder = Base64.getDecoder();
-    List<String> decodedPrefixes = new ArrayList<>(prefixes.length);
-    for (String prefix : prefixes) {
-      Preconditions.checkArgument(prefix != null, "Encoded credential prefixes cannot be null");
-      String decodedPrefix = new String(decoder.decode(prefix), UTF_8);
-      Preconditions.checkArgument(!decodedPrefix.isEmpty(), "Credential prefixes cannot be empty");
-      decodedPrefixes.add(decodedPrefix);
-    }
-    return decodedPrefixes;
+    return Arrays.stream(prefixes)
+        .map(prefix -> new String(DECODER.decode(prefix), UTF_8))
+        .collect(Collectors.toList());
   }
 
   /** Converts a UC SDK {@link TemporaryCredentials} into an internal {@link GenericCredential}. */
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
@@ -15,6 +15,10 @@ import java.util.List;
 public final class CredentialUtil {
   private CredentialUtil() {}
 
+  public static String hadoopConfNamespaceForIndex(int index) {
+    return UCHadoopConfConstants.UC_MULTI_CRED_NAMESPACE_PREFIX + index + ".";
+  }
+
   /** Converts a UC SDK {@link TemporaryCredentials} into an internal {@link GenericCredential}. */
   public static GenericCredential toGenericCredential(TemporaryCredentials tempCred) {
     long expiry =
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
@@ -9,15 +9,46 @@ import io.unitycatalog.hadoop.internal.auth.AwsCredential;
 import io.unitycatalog.hadoop.internal.auth.AzureCredential;
 import io.unitycatalog.hadoop.internal.auth.GcsCredential;
 import io.unitycatalog.hadoop.internal.auth.GenericCredential;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Base64;
 import java.util.List;
 import java.util.function.Function;
 
 /** Internal utilities for building and selecting {@link GenericCredential}s. */
 public final class CredentialUtil {
+  private static final Charset UTF_8 = StandardCharsets.UTF_8;
+
   private CredentialUtil() {}
 
-  public static String multiCredPrefixKeyForIndex(int index) {
-    return UCHadoopConfConstants.UC_MULTI_CRED_PREFIX_KEY + index;
+  public static String[] encodeMultiCredPrefixes(List<String> prefixes) {
+    Preconditions.checkArgument(
+        prefixes != null && !prefixes.isEmpty(),
+        "List of credential prefixes cannot be null or empty");
+    List<String> encodedPrefixes = new ArrayList<>(prefixes.size());
+    Base64.Encoder encoder = Base64.getEncoder();
+    for (String prefix : prefixes) {
+      Preconditions.checkArgument(
+          prefix != null && !prefix.isEmpty(), "Credential prefixes cannot be null or empty");
+      encodedPrefixes.add(encoder.encodeToString(prefix.getBytes(UTF_8)));
+    }
+    return encodedPrefixes.toArray(String[]::new);
+  }
+
+  public static List<String> decodeMultiCredPrefixes(String[] prefixes) {
+    Preconditions.checkArgument(
+        prefixes != null && prefixes.length > 0,
+        "Encoded credential prefixes cannot be null or empty");
+    Base64.Decoder decoder = Base64.getDecoder();
+    List<String> decodedPrefixes = new ArrayList<>(prefixes.length);
+    for (String prefix : prefixes) {
+      Preconditions.checkArgument(prefix != null, "Encoded credential prefixes cannot be null");
+      String decodedPrefix = new String(decoder.decode(prefix), UTF_8);
+      Preconditions.checkArgument(!decodedPrefix.isEmpty(), "Credential prefixes cannot be empty");
+      decodedPrefixes.add(decodedPrefix);
+    }
+    return decodedPrefixes;
   }
 
   /** Converts a UC SDK {@link TemporaryCredentials} into an internal {@link GenericCredential}. */
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
@@ -96,10 +96,8 @@ public final class CredentialUtil {
   public static GenericCredential selectForLocation(
       String location, List<GenericCredential> creds) {
     Preconditions.checkArgument(
-        creds != null && !creds.isEmpty(), "Credential response has no storage credentials.");
-    if (creds.size() == 1) {
-      Preconditions.checkNotNull(creds.get(0), "Credential response contains null.");
-    }
+        creds != null && creds.size() > 1,
+        "Credential selection requires multiple storage credentials.");
     int selected = longestCoveringIndex(location, creds, GenericCredential::prefix);
     Preconditions.checkArgument(
         selected >= 0, "No vended credential covers location '%s'.", location);
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
@@ -16,8 +16,8 @@ import java.util.function.Function;
 public final class CredentialUtil {
   private CredentialUtil() {}
 
-  public static String hadoopConfNamespaceForIndex(int index) {
-    return UCHadoopConfConstants.UC_MULTI_CRED_NAMESPACE_PREFIX + index + ".";
+  public static String multiCredPrefixKeyForIndex(int index) {
+    return UCHadoopConfConstants.UC_MULTI_CRED_PREFIX_KEY + index;
   }
 
   /** Converts a UC SDK {@link TemporaryCredentials} into an internal {@link GenericCredential}. */
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
@@ -85,6 +85,27 @@ public final class CredentialUtil {
     return best;
   }
 
+  /**
+   * Returns the index of the prefix that covers {@code location} by the longest match, or {@code
+   * -1} if none does. Null prefixes are skipped.
+   */
+  public static int longestCoveringIndex(String location, List<String> prefixes) {
+    int best = -1;
+    int bestLen = -1;
+    for (int i = 0; i < prefixes.size(); i++) {
+      String prefix = prefixes.get(i);
+      if (prefix == null || !prefixCovers(location, prefix)) {
+        continue;
+      }
+      int len = stripTrailingSlashes(prefix).length();
+      if (len > bestLen) {
+        best = i;
+        bestLen = len;
+      }
+    }
+    return best;
+  }
+
   static boolean prefixCovers(String location, String prefix) {
     String l = stripTrailingSlashes(location);
     String p = stripTrailingSlashes(prefix);
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
@@ -10,6 +10,7 @@ import io.unitycatalog.hadoop.internal.auth.AzureCredential;
 import io.unitycatalog.hadoop.internal.auth.GcsCredential;
 import io.unitycatalog.hadoop.internal.auth.GenericCredential;
 import java.util.List;
+import java.util.function.Function;
 
 /** Internal utilities for building and selecting {@link GenericCredential}s. */
 public final class CredentialUtil {
@@ -68,25 +69,10 @@ public final class CredentialUtil {
     if (creds.size() == 1) {
       Preconditions.checkNotNull(creds.get(0), "Credential response contains null.");
     }
-    GenericCredential best = null;
-    int bestLen = -1;
-    for (GenericCredential cred : creds) {
-      if (cred == null) {
-        continue;
-      }
-      String prefix = cred.prefix();
-      if (prefix == null || !prefixCovers(location, prefix)) {
-        continue;
-      }
-      int len = stripTrailingSlashes(prefix).length();
-      if (len > bestLen) {
-        best = cred;
-        bestLen = len;
-      }
-    }
+    int selected = longestCoveringIndex(location, creds, GenericCredential::prefix);
     Preconditions.checkArgument(
-        best != null, "No vended credential covers location '%s'.", location);
-    return best;
+        selected >= 0, "No vended credential covers location '%s'.", location);
+    return creds.get(selected);
   }
 
   /**
@@ -94,10 +80,20 @@ public final class CredentialUtil {
    * -1} if none does. Null prefixes are skipped.
    */
   public static int longestCoveringIndex(String location, List<String> prefixes) {
+    Preconditions.checkNotNull(prefixes, "List of prefixes cannot be null.");
+    return longestCoveringIndex(location, prefixes, Function.identity());
+  }
+
+  private static <T> int longestCoveringIndex(
+      String location, List<T> values, Function<T, String> prefixExtractor) {
     int best = -1;
     int bestLen = -1;
-    for (int i = 0; i < prefixes.size(); i++) {
-      String prefix = prefixes.get(i);
+    for (int i = 0; i < values.size(); i++) {
+      T value = values.get(i);
+      if (value == null) {
+        continue;
+      }
+      String prefix = prefixExtractor.apply(value);
       if (prefix == null || !prefixCovers(location, prefix)) {
         continue;
       }
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
@@ -25,13 +25,13 @@ public final class CredentialUtil {
 
   private CredentialUtil() {}
 
-  public static String[] encodeMultiCredPrefixes(List<String> prefixes) {
+  public static String[] encodeCredPrefixes(List<String> prefixes) {
     return prefixes.stream()
         .map(prefix -> ENCODER.encodeToString(prefix.getBytes(UTF_8)))
         .toArray(String[]::new);
   }
 
-  public static List<String> decodeMultiCredPrefixes(String[] prefixes) {
+  public static List<String> decodeCredPrefixes(String[] prefixes) {
     return Arrays.stream(prefixes)
         .map(prefix -> new String(DECODER.decode(prefix), UTF_8))
         .collect(Collectors.toList());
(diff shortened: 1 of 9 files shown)


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -54,6 +59,12 @@ public class CredScopedFileSystem extends FilterFileSystem {
       "unitycatalog.credScopedFs.cache.maxSize";
   private static final int CRED_SCOPED_FS_CACHE_MAX_SIZE_DEFAULT = 100;
 
+  // Maximum number of namespaced credentials read from the configuration to prevent OOM.
+  private static final String MAX_MULTI_CRED_COUNT_PROPERTY = "unitycatalog.multi.cred.maxCount";
```
openinx: Should we move all those property keys and default values into a separate `PropsUtil` ?

openinx: even includes the 

```java
  private static final String CRED_SCOPED_FS_CACHE_MAX_SIZE =
      "unitycatalog.credScopedFs.cache.maxSize";
  private static final int CRED_SCOPED_FS_CACHE_MAX_SIZE_DEFAULT = 100;
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
```diff
@@ -64,24 +69,40 @@ public static GenericCredential selectForLocation(
     if (creds.size() == 1) {
```
openinx: nit: The `size() == 1` branch only null-checks, then still requires a covering prefix. That contradicts #1694 ("single credential used regardless of match") and `GenericCredentialProvider.fetchAndSelectCredential`, which already early-returns the sole credential. Prefer an early `return creds.get(0)` here (and update `selectorRejectsSingleWithoutPrefixMatch`), or drop the special-case if `selectForLocation` is multi-only.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystemNamespaceTest.java
```diff
@@ -0,0 +1,257 @@
...
+    assertThat(snapshot(conf)).isEqualTo(before);
+  }
+
+  // --- count > 1 requires selection and overlay
+  // ----------------------------------------------------------
+
+  /**
+   * Across credential sets of varying size, the one whose location is the longest prefix covering
+   * the URI is selected, and exactly that credential's keys are lifted to the top level (overriding
+   * any stale top-level placeholder), while the source configuration is left unchanged.
```
openinx: nit: Comment claims selected keys override stale top-level placeholders, but the parameterized cases never set a colliding top-level key. Add one assertion that a pre-existing top-level value is overwritten by the overlay.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/FileSystemCredId.java
```diff
@@ -37,18 +37,29 @@ public static FileSystemCredId create(Configuration conf) {
...
 
   private static String getCredPrefix(Configuration conf) {
     return conf.get(UCHadoopConfConstants.UC_CREDENTIAL_PREFIX_KEY);
   }
 
+  private static String getCredPrefix(Configuration conf, String namespace) {
+    String prefixKey =
+        namespace == null
+            ? UCHadoopConfConstants.UC_CREDENTIAL_PREFIX_KEY
+            : namespace + UCHadoopConfConstants.UC_CREDENTIAL_PREFIX_KEY;
```
openinx: wait, as I knew,   we won't be able to pass those config keys which does not starts with the `fs.*`, just as the comment said: https://github.com/unitycatalog/unitycatalog/blob/f9211f1f2b51b487c66124ca9472dd78f5c02833/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConfConstants.java#L40-L45

So I'm curious who did you successfully pass the test cases if we cannot succesfully pass the `fs.*` config keys from the spark driver to spark executors ?  we need to carefully add the tests for this.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -82,8 +93,50 @@ FileSystem getDelegate() {
...
+  }
+
+  /**
+   * Returns the namespace key for the credential that covers the given URI, or null if no namespace
+   * is needed (when there is only one credential). When there are multiple credentials, each
+   * credential's details in the Hadoop configuration are prefixed with the namespace key to avoid
+   * key collisions. These namespaced keys cannot be picked up downstream; they must be extracted
+   * and restored to top-level keys prior to file system initialization.
+   */
+  private static String selectNamespace(URI uri, Configuration conf) {
```
openinx: As the table with multi creds , is a 10% of the total user cases,  I'm a bit hesitated to over-complicate the hadoop conf settings for reusing the initial credential ?  do you think will it possible for us to reuse the inital credential only for the table with one single cred, for those tables with multi creds we won't reuse the initial cred ?  do you think will it simplify the entire fs design ?

If so, I'd prefer to remove the reuse for table with multi creds.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -82,8 +93,50 @@ FileSystem getDelegate() {
...
+    int count = conf.getInt(UCHadoopConfConstants.UC_MULTI_CRED_COUNT_KEY, 0);
```
openinx: You will need well-covered test cases to cover all those edge cases for this method, if you want to add this.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
foss-contributor: Wait, yes, this should line should be removed. `selectForLocation` is only called by cred providers, which already short circuit: 

```
private GenericCredential fetchAndSelectCredential() throws ApiException {
    ...
    if (credentials.size() == 1) {
      return credentials.get(0);
    }
...
    return CredentialUtil.selectForLocation(cacheKey.prefix(), credentials);
  }
```

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystemNamespaceTest.java
foss-contributor: I think this comment is outdated, should be updated as we don't set top level keys for multi cred encodings.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
foss-contributor: I thought about this. Here is some thoughts:
1. Yes, we can force multi cred = credscoped + renewal required. Then, we only encode list of prefixes and set FileSystemCredId and make executor fetch initial creds.
2. We can simplify this initialize method as such (if we still want current solution):

```
initialize:
    if multi_cred_marker is set:
               // FileSystemCredId = multi cred namespaced retrieval
               // this.fs = CACHE.(...multi cred version)
    FileSystemCredId key = FileSystemCredId.create(conf, uri); // unchanged
    this.fs = CACHE.getOrLoad(key, () -> newFileSystem(uri, conf)); // unchanged
```
This way code paths are isolated for the 90% single location tables.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
foss-contributor: maybe follow up PR?

openinx: I'm okay to be separate PR.  could you pls file a separate issue to track this ?

foss-contributor: done: https://github.com/unitycatalog/unitycatalog/issues/1757

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -82,10 +93,42 @@ FileSystem getDelegate() {
 
   @Override
   public void initialize(URI uri, Configuration conf) throws IOException {
+    int multiCredCount = conf.getInt(UCHadoopConfConstants.UC_MULTI_CRED_COUNT_KEY, 0);
```
openinx: Instead of using such a composited conf key to set the hadoop conf,  can we just use a single key to pass the prefix list directly? 

you change: 
*  `fs.unitycatalog.multi.cred.count`
*  `fs.unitycatalog.multi.cred.prefix.<number>`

can we just simplify the entire work with only one key ?
* `fs.unitycatalog.multi.cred.prefixes=<prefix1>;<prefix2>;<prefix3>`

foss-contributor: Ok yes, we can do that, this will greatly simplify the design. 

I was thinking about using Hadoop conf built in `setStrings` but it cannot handle commas as that is what Hadoop uses to serialize. Workaround is to use Base64URL encoding (or similar), which doesn't seem worth it. I don't think cloud URIs should have commas, but this is a safety precaution.

Perhaps we can use JSON instead, as it can handle raw strings for the prefixes.

WDYT?

openinx: JSON is actually not so necessary,  I think the `setStrings` will automatically handle the escape chars ?

openinx: okay, confirmed, the hadoop cnofiguration setStrings does not support the escaping.  we have to do it by our self.

openinx: anyway, I will prefer the hadoop `setStrings` with a based64 or escaping.

foss-contributor: I thought about it some more, `setStrings` should be sufficient as it encodes all URI characters fine. Commas should not come from any UC server URL. I'll go ahead and implement this and eliminate the redundant key.

foss-contributor: Update: I didn't refresh the page and missed the latest comments. Ok yes, will implement base64 encoding then use `setStrings`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -82,10 +93,48 @@ FileSystem getDelegate() {
...
+      selectPrefixAndInitializeFileSystem(uri, conf, multiCredPrefixes);
+    }
+  }
+
+  private static List<String> getMultiCredPrefixes(Configuration conf) {
+    String encodedMultiCredPrefixes = conf.get(UCHadoopConfConstants.UC_MULTI_CRED_PREFIXES_KEY);
+    if (encodedMultiCredPrefixes == null || encodedMultiCredPrefixes.isEmpty()) {
+      return Collections.emptyList();
+    }
+    String[] encodedPrefixes = encodedMultiCredPrefixes.split(",", -1);
```
openinx: Why not use the `conf.getStrings(UCHadoopConfConstants.UC_MULTI_CRED_PREFIXES_KEY)` which the hadoop configuration already provided direclty ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -54,6 +59,12 @@ public class CredScopedFileSystem extends FilterFileSystem {
       "unitycatalog.credScopedFs.cache.maxSize";
   private static final int CRED_SCOPED_FS_CACHE_MAX_SIZE_DEFAULT = 100;
 
+  // Maximum number of credential prefixes read from the configuration to prevent OOM.
+  private static final String MAX_MULTI_CRED_COUNT_PROPERTY = "unitycatalog.multi.cred.maxCount";
+  private static final int MAX_MULTI_CRED_COUNT_DEFAULT = 10;
+  private static final int MAX_MULTI_CRED_COUNT =
+      Integer.getInteger(MAX_MULTI_CRED_COUNT_PROPERTY, MAX_MULTI_CRED_COUNT_DEFAULT);
```
openinx: Is it worth for us to add this limitation in this PR ?  I assume not, since we are pretty sure it's not a common cases for us to have so much prefixes.  I'd defer to add this unless we really hit this problem in our real cases.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -82,10 +93,48 @@ FileSystem getDelegate() {
...
+
+  private static List<String> getMultiCredPrefixes(Configuration conf) {
+    String encodedMultiCredPrefixes = conf.get(UCHadoopConfConstants.UC_MULTI_CRED_PREFIXES_KEY);
+    if (encodedMultiCredPrefixes == null || encodedMultiCredPrefixes.isEmpty()) {
+      return Collections.emptyList();
+    }
+    String[] encodedPrefixes = encodedMultiCredPrefixes.split(",", -1);
+    Preconditions.checkArgument(
+        encodedPrefixes.length > 1 && encodedPrefixes.length <= MAX_MULTI_CRED_COUNT,
+        "Number of credentials must be between 2 and %s: got %s",
```
openinx: It's confusing for others to understand why the num of creds have to be greater than 2 ? 
I think it's allowed to set the `UC_MULTI_CRED_PREFIXES_KEY` with only one prefix, but we just don't use it, since we are pretty confident that cred is for the filesystem.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -95,40 +144,50 @@ private static void restoreImpl(Configuration fsConf, String key, String default
...
-      // SecureAzureBlobFileSystem if not set (registered via the Java service loader).
-      restoreImpl(fsConf, "fs.abfs.impl", "org.apache.hadoop.fs.azurebfs.AzureBlobFileSystem");
-      restoreImpl(
-          fsConf, "fs.abfss.impl", "org.apache.hadoop.fs.azurebfs.SecureAzureBlobFileSystem");
-      restoreImpl(fsConf, "fs.AbstractFileSystem.abfs.impl", "org.apache.hadoop.fs.azurebfs.Abfs");
-      restoreImpl(
-          fsConf, "fs.AbstractFileSystem.abfss.impl", "org.apache.hadoop.fs.azurebfs.Abfss");
-      fsConf.set("fs.abfs.impl.disable.cache", "true");
-      fsConf.set("fs.abfss.impl.disable.cache", "true");
+    Configuration fsConf = new Configuration(conf);
```
openinx, line 133: Is it necessary to touch this in this PR? if not,  then let's keep as it is unless we have to, that actually help us to be focused on the change that we want to get in.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/id/FileSystemCredId.java
```diff
@@ -34,17 +34,17 @@ public static FileSystemCredId create(Configuration conf) {
...
-   * the configuration carries no Unity Catalog credential type, falls back to a {@link
-   * DefaultCredId} derived from the URI's scheme and authority.
-   */
   public static FileSystemCredId create(Configuration conf, URI uri) {
     CredId credId = CredId.create(conf, () -> new DefaultCredId(uri, conf));
     String prefix = getCredPrefix(conf);
     return new FileSystemCredId(credId, prefix);
   }
 
+  public static FileSystemCredId create(Configuration conf, URI uri, String prefix) {
```
openinx, line 37: I'd prefer to entirely remove the `FileSystemCredId create(Configuration conf, URI uri)` directly,  after introduce this 3-arg method, as we can always use this one in the upper caller ?

openinx: The reason is,  it's pretty easy to people to confuse which method we should use considering one provides the default `prefix` parsing from the conf while another one need to intentionally set it. 

I understand this context,  the 1st one actually is for `XXTokenProvider` as those token providers only handle one cred,  while the other one is for `FileSystem` constructor.  

but I don't think others will easily get this context.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
```diff
@@ -9,12 +9,48 @@
...
+  public static String[] encodeMultiCredPrefixes(List<String> prefixes) {
+    Preconditions.checkArgument(
+        prefixes != null && !prefixes.isEmpty(),
+        "List of credential prefixes cannot be null or empty");
+    List<String> encodedPrefixes = new ArrayList<>(prefixes.size());
+    Base64.Encoder encoder = Base64.getEncoder();
+    for (String prefix : prefixes) {
+      Preconditions.checkArgument(
+          prefix != null && !prefix.isEmpty(), "Credential prefixes cannot be null or empty");
+      encodedPrefixes.add(encoder.encodeToString(prefix.getBytes(UTF_8)));
```
openinx: Are you sure we can reuse the same `encoder` to encode multiple strings without messing up the current string encoding from the last string encoding process which possibly retain state in the shared Encoder ? 

If so, then that's a serious bug.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
```diff
@@ -9,12 +9,48 @@
...
+    }
+    return encodedPrefixes.toArray(String[]::new);
+  }
+
+  public static List<String> decodeMultiCredPrefixes(String[] prefixes) {
```
openinx: I think we can use this java chained methods to do it direclty ?

```
Arrays.stream(prefixes)
                .map(str -> Base64.getEncoder().encodeToString(str.getBytes(StandardCharsets.UTF_8)))
                .collect(Collectors.toList());
```

openinx: Same as the decode process.

openinx (changes requested):
Left few comments.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
foss-contributor: Updated to use it, my concern is we do not set it via `setStrings` so encode and decode are not symmetrical.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
foss-contributor: I think it should be safe: Instances of [Base64.Encoder](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/util/Base64.Encoder.html) class are safe for use by multiple concurrent threads.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConfConstants.java
```diff
@@ -61,6 +61,12 @@ private UCHadoopConfConstants() {}
    */
   public static final String UC_CRED_CONTEXT_ID_KEY = "fs.unitycatalog.cred.context.id";
 
+  /**
+   * The list of credential prefixes for multi-credential responses. This should not be set when
+   * only one credential is vended. It must be set when there are multiple credentials.
+   */
+  public static final String UC_MULTI_CRED_PREFIXES_KEY = "fs.unitycatalog.multi.cred.prefixes";
```
openinx: I just looked at the `fs.unitycatalog.credential.prefix` ([link](https://github.com/unitycatalog/unitycatalog/blob/4e2569967ddb852c9c12c74af41b94f9b200a28b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/UCHadoopConfConstants.java#L50-L54)), and it says: 

```
  /**
   * The storage prefix that a UC credential covers. When there are multiple credentials, each
   * credential must set this key so the correct credential can be selected.
   */
  public static final String UC_CREDENTIAL_PREFIX_KEY = "fs.unitycatalog.credential.prefix";
```

This is unclear for me,  why for the multiple credential cases we will need to set this key ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -82,8 +87,34 @@ FileSystem getDelegate() {
...
-    FileSystemCredId key = FileSystemCredId.create(conf, uri);
-    this.fs = CACHE.getOrLoad(key, () -> newFileSystem(uri, conf));
+    List<String> multiCredPrefixes = getMultiCredPrefixes(conf);
+    if (multiCredPrefixes.size() > 1) {
+      FileSystemCredId key = getMultiCredFileSystemKey(uri, conf, multiCredPrefixes);
+      this.fs = CACHE.getOrLoad(key, () -> copyConfAndCreateNewFileSystem(uri, conf, key.prefix()));
+      return;
+    }
+
+    String prefix = conf.get(UCHadoopConfConstants.UC_CREDENTIAL_PREFIX_KEY);
```
openinx: This is pretty strange,  as here we are sure that `multiCredPrefixes` only have one element, but we are turning to parse the `UC_CREDENTIAL_PREFIX_KEY` even although we pretty sure the prefix exists in the  `multiCredPrefixes` ?

Zoom out,   what are the correct semantic for the `fs.unitycatalog.multi.cred.prefixes` and `fs.unitycatalog.credential.prefix` ?  I don't think it's clear enough for people to understand this.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -82,8 +87,33 @@ FileSystem getDelegate() {
...
+    } else {
+      FileSystemCredId key = getMultiCredFileSystemKey(uri, conf, credPrefixes);
+      this.fs = CACHE.getOrLoad(key, () -> copyConfAndCreateNewFileSystem(uri, conf, key.prefix()));
+    }
+  }
+
+  private static List<String> getCredPrefixes(Configuration conf) {
+    String[] encodedCredPrefixes =
+        conf.getStrings(UCHadoopConfConstants.UC_CREDENTIAL_PREFIXES_KEY);
+    if (encodedCredPrefixes == null) {
```
openinx:
```suggestion
    if (encodedCredPrefixes == null || encodedCredPrefixes.isEmpty()) {
```

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -82,8 +87,33 @@ FileSystem getDelegate() {
...
+      return Collections.emptyList();
+    }
+    return CredentialUtil.decodeMultiCredPrefixes(encodedCredPrefixes);
+  }
+
+  private static FileSystemCredId getMultiCredFileSystemKey(
+      URI uri, Configuration conf, List<String> multiCredPrefixes) {
```
openinx: `multiCredPrefixes` ->  `credPrefixes`.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -82,8 +87,33 @@ FileSystem getDelegate() {
 
   @Override
   public void initialize(URI uri, Configuration conf) throws IOException {
-    FileSystemCredId key = FileSystemCredId.create(conf, uri);
-    this.fs = CACHE.getOrLoad(key, () -> newFileSystem(uri, conf));
+    List<String> credPrefixes = getCredPrefixes(conf);
+    if (credPrefixes.size() <= 1) {
+      String prefix = credPrefixes.isEmpty() ? "" : credPrefixes.get(0);
```
openinx: Q:  should we use empty string or `null` for the `prefix` if the `credPrefixes` is empty ?  what's the difference ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -82,8 +87,33 @@ FileSystem getDelegate() {
...
+    }
+  }
+
+  private static List<String> getCredPrefixes(Configuration conf) {
+    String[] encodedCredPrefixes =
+        conf.getStrings(UCHadoopConfConstants.UC_CREDENTIAL_PREFIXES_KEY);
+    if (encodedCredPrefixes == null) {
+      return Collections.emptyList();
+    }
+    return CredentialUtil.decodeMultiCredPrefixes(encodedCredPrefixes);
```
openinx: `decodeMultiCredPrefixes` -> `decodeCredPrefixes`, since the `CredPrefixes` already implicate those are multi cred prefixes.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/fs/CredScopedFileSystem.java
```diff
@@ -82,8 +87,33 @@ FileSystem getDelegate() {
...
+  }
+
+  private static FileSystemCredId getMultiCredFileSystemKey(
```
openinx: `getMultiCredFileSystemKey` -> `getFileSystemCredId`

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredentialUtil.java
```diff
@@ -9,12 +9,34 @@
...
 
 /** Internal utilities for building and selecting {@link GenericCredential}s. */
 public final class CredentialUtil {
+  private static final Charset UTF_8 = StandardCharsets.UTF_8;
+  private static final Base64.Encoder ENCODER = Base64.getEncoder();
+  private static final Base64.Decoder DECODER = Base64.getDecoder();
+
   private CredentialUtil() {}
 
+  public static String[] encodeMultiCredPrefixes(List<String> prefixes) {
```
openinx: nit: remove the `Multi`.  same as the `decodeMultiCredPrefixes`

openinx (approved):
Took another look, generally looks good to me now.  I left few nit comments, please address them then I will merge.  Thanks for the patient work.

PR #1705, opened 2026-07-23
[Hadoop] Encode credentials for multi-location table support

### LINKED ISSUE #1694: Hadoop connector credential support for multi-storage location tables.

foss-contributor:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1705/files) to review incremental changes.
- [stack/encode-scoped-vended-credentials](https://github.com/unitycatalog/unitycatalog/pull/1705) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1705/files)]

---------
## Summary

- Encode vended prefixes, whether single, or multiple credential
- Preserve the legacy top-level layout for a single credential.

## Testing

Added Spark integration test for multi location table.

## Issue

Fixes https://github.com/unitycatalog/unitycatalog/issues/1694

### DIFF
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
@@ -5,7 +5,6 @@ import io.unitycatalog.client.ApiException;
 import io.unitycatalog.client.auth.TokenProvider;
 import io.unitycatalog.client.internal.ApiClientUtils;
 import io.unitycatalog.client.internal.Clock;
-import io.unitycatalog.client.internal.Preconditions;
 import io.unitycatalog.hadoop.UCCredentialHadoopConfs;
 import io.unitycatalog.hadoop.internal.auth.CredentialCache;
 import io.unitycatalog.hadoop.internal.auth.CredentialCache.RenewableCredential;
@@ -195,19 +194,15 @@ public class CredPropsUtil {
     List<GenericCredential> credentials =
         fetchGenericCredentials(
             hadoopConf, apiClient, catalogUri, tokenProvider, appVersions, credId);
-    // TODO: Support encoding multiple vended credentials in CredPropsBuilder.
-    Preconditions.checkState(
-        credentials.size() == 1,
-        "Only single credential responses are supported, got %s",
-        credentials.size());
-    GenericCredential cred = credentials.get(0);
     CredPropsBuilder builder =
         CredPropsBuilder.forCloud(cloudType.get(), hadoopConf)
             .credScopedFsEnabled(credScopedFsEnabled)
-            .initialCredential(cred);
+            .initialCredentials(credentials);
+
     if (renewCredEnabled) {
       builder.enableRenewCred(catalogUri, tokenProvider, credId, appVersions);
     }
+
     return builder.build();
   }
 
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
@@ -116,7 +116,7 @@ public abstract class CredPropsBuilder {
       }
     } else {
       // Do not set initial credentials when there are multiple credentials to minimize the
-      // number of key that need to be encoded. Instead, token providers fetch credentials
+      // number of keys that need to be encoded. Instead, token providers fetch credentials
       // at file system initialization time.
       Preconditions.checkState(
           credScopedFsEnabled,
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
@@ -110,9 +110,9 @@ public abstract class CredPropsBuilder {
     if (initialCredentials.size() == 1) {
       GenericCredential credential = initialCredentials.get(0);
       if (renewCredEnabled) {
-        setRenewableCredKeys(credential);
+        setInitRenewableCredKeys(credential);
       } else {
-        setFixedCredKeys(credential);
+        setInitFixedCredKeys(credential);
       }
     } else {
       // Do not set initial credentials when there are multiple credentials to minimize the
@@ -166,8 +166,8 @@ public abstract class CredPropsBuilder {
   protected abstract void setVendedProviderKeys();
 
   /** Writes the renewable-path credential secrets (the {@code init.*} keys) for this cloud. */
-  protected abstract void setRenewableCredKeys(GenericCredential cred);
+  protected abstract void setInitRenewableCredKeys(GenericCredential cred);
 
   /** Writes the fixed-path credential secrets for this cloud. */
-  protected abstract void setFixedCredKeys(GenericCredential cred);
+  protected abstract void setInitFixedCredKeys(GenericCredential cred);
 }
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
@@ -3,6 +3,7 @@ package io.unitycatalog.hadoop.internal.props;
 import io.unitycatalog.client.auth.TokenProvider;
 import io.unitycatalog.client.internal.Preconditions;
 import io.unitycatalog.hadoop.internal.CloudType;
+import io.unitycatalog.hadoop.internal.CredentialUtil;
 import io.unitycatalog.hadoop.internal.UCHadoopConfConstants;
 import io.unitycatalog.hadoop.internal.auth.GenericCredential;
 import io.unitycatalog.hadoop.internal.fs.CredScopedFileSystem;
@@ -10,7 +11,9 @@ import io.unitycatalog.hadoop.internal.fs.CredScopedFs;
 import io.unitycatalog.hadoop.internal.id.CredId;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import org.apache.hadoop.conf.Configuration;
 
 /** Builds the cloud-provider specific Hadoop configuration credential properties. */
@@ -29,7 +32,7 @@ public abstract class CredPropsBuilder {
   private TokenProvider tokenProvider;
   private CredId credId;
   private Map<String, String> appVersions;
-  private GenericCredential initialCredential;
+  private List<GenericCredential> initialCredentials;
 
   protected CredPropsBuilder(Configuration hadoopConf) {
     Preconditions.checkNotNull(hadoopConf, "hadoopConf is required");
@@ -73,19 +76,19 @@ public abstract class CredPropsBuilder {
     return this;
   }
 
-  /** Records the fetched credential whose secrets are written by {@link #build()}. */
-  public CredPropsBuilder initialCredential(GenericCredential initialCredential) {
-    this.initialCredential = initialCredential;
+  /** Records the fetched credentials used by {@link #build()}. */
+  public CredPropsBuilder initialCredentials(List<GenericCredential> initialCredentials) {
+    this.initialCredentials = initialCredentials;
     return this;
   }
 
   public Map<String, String> build() {
-    Preconditions.checkNotNull(initialCredential, "initialCredential is required");
+    Preconditions.checkState(
+        initialCredentials != null && !initialCredentials.isEmpty(),
+        "Initial credentials cannot be null or empty");
+
     if (credScopedFsEnabled) {
       setCredScopedFsKeys();
-    }
-    if (initialCredential.prefix() != null) {
-      set(UCHadoopConfConstants.UC_CREDENTIAL_PREFIX_KEY, initialCredential.prefix());
     }
     if (renewCredEnabled) {
       setVendedProviderKeys();
@@ -97,10 +100,45 @@ public abstract class CredPropsBuilder {
       credId.props().forEach(this::set);
       appVersions.forEach(
           (key, value) -> set(UCHadoopConfConstants.UC_ENGINE_VERSION_PREFIX + key, value));
-      setRenewableCredKeys(initialCredential);
-    } else {
-      setFixedCredKeys(initialCredential);
     }
+    List<String> credPrefixes =
+        initialCredentials.stream()
+            .map(GenericCredential::prefix)
+            .filter(prefix -> prefix != null && !prefix.isEmpty())
+            .collect(Collectors.toList());
+
+    if (initialCredentials.size() == 1) {
+      GenericCredential credential = initialCredentials.get(0);
+      if (renewCredEnabled) {
+        setRenewableCredKeys(credential);
+      } else {
+        setFixedCredKeys(credential);
+      }
+    } else {
+      // Do not set initial credentials when there are multiple credentials to minimize the
+      // number of key that need to be encoded. Instead, token providers fetch credentials
+      // at file system initialization time.
+      Preconditions.checkState(
+          credScopedFsEnabled,
+          "%s credentials were vended but the credential-scoped filesystem is disabled.",
+          initialCredentials.size());
+      Preconditions.checkState(
+          renewCredEnabled,
+          "%s credentials were vended but credential renewal is disabled.",
+          initialCredentials.size());
+      // Since there are multiple credentials, each credential's prefix must be non-null
+      // and non-empty to differentiate between credentials.
+      Preconditions.checkArgument(
+          credPrefixes.size() == initialCredentials.size(),
+          "Credential prefixes cannot be null or empty when multiple credentials are vended");
+    }
+
+    if (!credPrefixes.isEmpty()) {
+      String commaSeparatedPrefixes =
+          String.join(",", CredentialUtil.encodeCredPrefixes(credPrefixes));
+      set(UCHadoopConfConstants.UC_CREDENTIAL_PREFIXES_KEY, commaSeparatedPrefixes);
+    }
+
     return Collections.unmodifiableMap(new HashMap<>(props));
   }
 
--- a/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
+++ b/connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
@@ -83,13 +83,12 @@ public abstract class CredPropsBuilder {
   }
 
   public Map<String, String> build() {
-    Preconditions.checkState(
-        initialCredentials != null && !initialCredentials.isEmpty(),
-        "Initial credentials cannot be null or empty");
-
+    // Set the props when credential scoped filesystem enabled.
     if (credScopedFsEnabled) {
       setCredScopedFsKeys();
     }
+
+    // Set the props when renewal credential enabled.
     if (renewCredEnabled) {
       setVendedProviderKeys();
       set(UCHadoopConfConstants.UC_URI_KEY, catalogUri);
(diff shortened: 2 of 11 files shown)


connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -79,14 +82,16 @@ public CredPropsBuilder initialCredential(GenericCredential initialCredential) {
     return this;
   }
 
+  public CredPropsBuilder setCredPrefixes(List<String> prefixes) {
```
openinx: oh,  I think this method can be `credPrefixes`, since all those above methods don't have a `setXXX`, it will be good for us to align the method naming.

foss-contributor: agree, oops

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -97,9 +102,24 @@ public Map<String, String> build() {
       credId.props().forEach(this::set);
       appVersions.forEach(
           (key, value) -> set(UCHadoopConfConstants.UC_ENGINE_VERSION_PREFIX + key, value));
-      setRenewableCredKeys(initialCredential);
-    } else {
-      setFixedCredKeys(initialCredential);
+    }
+    if (initialCredential != null) {
+      if (renewCredEnabled) {
+        setRenewableCredKeys(initialCredential);
```
openinx: nit:  should we rename this as `setInitRenewableCredKey` ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -97,9 +102,24 @@ public Map<String, String> build() {
...
+      } else {
+        setFixedCredKeys(initialCredential);
+      }
+      if (initialCredential.prefix() != null) {
+        set(UCHadoopConfConstants.UC_CREDENTIAL_PREFIX_KEY, initialCredential.prefix());
```
openinx: wait... this key `UC_CREDENTIAL_PREFIX_KEY` will only be set by the the CredScopeFileSystem, as we discussed.  the for the `CredPropsBuilder`, we will only set the `UC_CREDENTIAL_PREFIXES_KEY` props ?

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -97,9 +102,24 @@ public Map<String, String> build() {
...
+        set(
+            UCHadoopConfConstants.UC_CREDENTIAL_PREFIXES_KEY,
+            CredentialUtil.encodeCredPrefixes(List.of(initialCredential.prefix()))[0]);
```
openinx: If the `initialCredential` have multiple credentials,  then why do we only keep the 1st prefix in this conf?  

THIS IS A BUG, which should be a bug. I think you need to use the string join with the commas to populate the conf.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
foss-contributor: Yes, I haven't pushed this change yet, but it will be deleted.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
openinx: wait, if you are populating the `initialCredential`, then why don't you just hide all those into the `setRenewableCredKeys` and `setFixedCredKeys` ?  essentially, the `prefix` is a part of the initial credential. and we shouldn't set part of the initial cred via the  `setRenewableCredKeys`/`setFixedCredKeys`, while set others outside of the method.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -97,9 +102,24 @@ public Map<String, String> build() {
...
+      }
+    }
+    if (credPrefixes != null) {
+      String[] encodedPrefixes = CredentialUtil.encodeCredPrefixes(credPrefixes);
+      String commaSeparatedPrefixes = String.join(",", encodedPrefixes);
+      set(UCHadoopConfConstants.UC_CREDENTIAL_PREFIXES_KEY, commaSeparatedPrefixes);
```
openinx: If the builder set the `initialCredential` and `credPrefixes` at the same time,  then this `UC_CREDENTIAL_PREFIXES_KEY` will be set twice ?  please think about this, and don't make the fragile meaning of this same key.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
foss-contributor: I've fixed this locally, just waiting for tests to pass before pushing. This is outdated code.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -195,22 +196,57 @@ private static Map<String, String> createCredProps(
...
-    Preconditions.checkState(
-        credentials.size() == 1,
-        "Only single credential responses are supported, got %s",
-        credentials.size());
-    GenericCredential cred = credentials.get(0);
+    Preconditions.checkState(!credentials.isEmpty(), "No credentials were vended.");
     CredPropsBuilder builder =
         CredPropsBuilder.forCloud(cloudType.get(), hadoopConf)
-            .credScopedFsEnabled(credScopedFsEnabled)
-            .initialCredential(cred);
```
openinx, line 207: Let's still keep the `initialCredential` code, regardless of single credential or multiple credentials.  and let's set the hadoop conf prefixes during the `CredPropsBuilder.build`,  which will greatly simplify the code. 

and we will use the `CredPropsBuilder.build()` as the source of truth to set the hadoop conf, includes `initialCredential` for a single credential, and `prefixes` for multiple credentials.

openinx: Let's be more clear. 

* for initial cred, we will only set the initial cred into the hadoop conf if  it's the single cred. 
* for `prefixes`, we will set it for both single cred and multiple credentials case.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/CredPropsUtil.java
```diff
@@ -195,22 +196,57 @@ private static Map<String, String> createCredProps(
...
+          credentials.size());
+
+      // Since there are multiple credentials, each credential's prefix must be non-null
+      // and non-empty to differentiate between credentials.
+      List<String> prefixes =
+          credentials.stream()
+              .map(CredPropsUtil::requireNonEmptyPrefix)
+              .collect(Collectors.toList());
+      builder.setCredPrefixes(prefixes);
+    }
```
openinx: this will be removed, if you think this comment makes sense. https://github.com/unitycatalog/unitycatalog/pull/1705/changes#r3739041036

connectors/spark/src/test/java/io/unitycatalog/spark/BaseSparkIntegrationTest.java
```diff
@@ -103,6 +104,7 @@ protected static boolean isSparkAtLeast(int major, int minor) {
   @BeforeEach
   @Override
   public void setUp() {
+    CredScopedFileSystem.clearCacheForTesting();
```
openinx: why do you want to change this ?  I mean with and without this PR, will the cache affect your test stability ? why ?

openinx (changes requested):
Left several comments, please take a look. thanks @foss-contributor for the contribution.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -73,20 +76,20 @@ public CredPropsBuilder credScopedFsEnabled(boolean credScopedFsEnabled) {
...
-  /** Records the fetched credential whose secrets are written by {@link #build()}. */
-  public CredPropsBuilder initialCredential(GenericCredential initialCredential) {
-    this.initialCredential = initialCredential;
+  /** Records the fetched credentials used by {@link #build()}. */
+  public CredPropsBuilder initialCredentials(List<GenericCredential> initialCredentials) {
+    this.initialCredentials = initialCredentials;
     return this;
   }
 
   public Map<String, String> build() {
```
openinx, line 85: Hi @foss-contributor ,  I just re-org the build code, so that it looks more clear and easier to track all the details. 

```java
public Map<String, String> build() {
    // Set the props when credential scoped filesystem enabled.
    if (credScopedFsEnabled) {
      setCredScopedFsKeys();
    }

    // Set the props when renewal credential enabled.
    if (renewCredEnabled) {
      setVendedProviderKeys();
      set(UCHadoopConfConstants.UC_URI_KEY, catalogUri);
      // Only 'fs.*' properties propagate to the FileSystem, so prefix the auth configs.
      tokenProvider
          .configs()
          .forEach((key, value) -> set(UCHadoopConfConstants.UC_AUTH_PREFIX + key, value));
      credId.props().forEach(this::set);
      appVersions.forEach(
          (key, value) -> set(UCHadoopConfConstants.UC_ENGINE_VERSION_PREFIX + key, value));
    }

    // Set the props for initial credentials.
    Preconditions.checkState(
        initialCredentials != null && !initialCredentials.isEmpty(),
        "Initial credentials cannot be null or empty");
    if (initialCredentials.size() == 1) {
      // Only set the initial credential for single case. for multiple credentials, we don't want to
      // do the initial credential optimization with over-complicating the code and hadoop conf for
      // uncommon cases.
      GenericCredential credential = initialCredentials.get(0);
      if (renewCredEnabled) {
        setInitRenewableCredKeys(credential);
      } else {
        setInitFixedCredKeys(credential);
      }
    }

    // Set the props for initial credential prefixes.
    List<String> credPrefixes = initialCredentials.stream()
        .map(GenericCredential::prefix)
        .filter(prefix -> prefix != null && !prefix.isEmpty())
        .collect(Collectors.toList());
    if (!credPrefixes.isEmpty()) {
      set(UCHadoopConfConstants.UC_CREDENTIAL_PREFIXES_KEY,
          String.join(",", CredentialUtil.encodeCredPrefixes(credPrefixes)));
    }

    return Collections.unmodifiableMap(new HashMap<>(props));
  }
```

openinx (approved):
the prod code looks good to me, few comments for the test code. please take a look.  thank you.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/CredPropsBaseTest.java
```diff
@@ -139,6 +142,56 @@ void credPropsMatchExactlyAcrossMatrix(
...
+  @MethodSource("testCaseMatrix")
+  void multipleCredPropsMatchExactlyAcrossMatrix(
+      CredKind kind,
+      boolean renew,
+      boolean credScoped,
+      boolean customImpl,
+      boolean expiring,
+      Map<String, String> appVersions)
+      throws Exception {
+    for (int prefixCount : List.of(2, 3, 5)) {
```
openinx, line 156: This multiplies the full `testCaseMatrix` (~128 combos) by `{2,3,5}` prefix counts, and `malformedCredentialPrefixesSkippedDoesNotThrow` below does the same × 6 prefix lists.

For multi-cred, `expiring` never affects the expected map (no init keys), and `renew=false` also writes **no secrets**—only prefixes—so many cells are vacuous "empty map matches empty map."

Please collapse to the dims that actually change encoding (cloud, renew, scoped, one multi size, a couple null/empty prefix cases). If `size > 1 && !renew` (or `!credScoped`) is unsupported, assert rejection instead of encoding an empty secret map.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -97,10 +99,35 @@ public Map<String, String> build() {
...
+    if (initialCredentials.size() == 1) {
+      // Seed the initial credential only in the common single-credential case. In the uncommon
+      // multiple-credential case, skip this optimization to keep the code and configuration simple.
+      GenericCredential credential = initialCredentials.get(0);
+      if (renewCredEnabled) {
+        setInitRenewableCredKeys(credential);
+      } else {
+        setInitFixedCredKeys(credential);
+      }
+    }
```
openinx, line 117: Issue #1694 / the PR Testing section say multi-cred requires credential scoping (and the old `multipleDeltaCredentialsAreRejected` test was removed), but `build()` does not reject `size > 1 && !credScopedFsEnabled` (or `!renewCredEnabled`).

With multiple credentials and no init seeding, a non-scoped / non-renew path only gets prefixes and no secrets—failure is deferred to `GenericCredentialProvider` needing `UC_CREDENTIAL_PREFIX_KEY`.

Please fail fast at encode time for unsupported combos, or drop that claim from the issue/PR description and document the deferred failure.

connectors/spark/src/test/java/io/unitycatalog/spark/PerPathCredentialTestFileSystem.java
```diff
@@ -0,0 +1,70 @@
...
+    GenericCredentialFetcher fetcher = () -> new ArrayList<>(EXPECTED_CREDENTIALS.values());
+    try (MockedStatic<GenericCredentialFetcher> mockedFetcher =
+        mockStatic(GenericCredentialFetcher.class)) {
+      mockedFetcher.when(() -> GenericCredentialFetcher.create(conf)).thenReturn(fetcher);
+
+      AwsSessionCredentials credentials =
+          (AwsSessionCredentials) new AwsVendedTokenProvider(conf).resolveCredentials();
+      assertThat(credentials.accessKeyId())
+          .as("expected access key %s for %s", expectedAccessKey, f)
+          .isEqualTo(expectedAccessKey);
```
openinx, line 58: This mock returns `EXPECTED_CREDENTIALS`, not what `vend()` / `CredPropsUtil` wrote into the conf. The Spark test mainly checks prefix selection via `UC_CREDENTIAL_PREFIX_KEY`, not that the encoded secrets/prefixes from the encoder are what the FS uses.

Prefer asserting against the delegate conf / real encoded props (as `CredPropsRoundTripTest` does), and keep a fetcher mock only for renewal.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/CredPropsUtilTest.java
```diff
@@ -39,16 +38,10 @@ void credContextIdVariesByCatalogUriSchemeAndAuth() {
         .isNotEqualTo(base);
   }
 
-  // TODO: Remove this test once CredPropsBuilder supports multiple vended credentials.
   @Test
-  void multipleDeltaCredentialsAreRejected() {
+  void noVendedCredentialsIsRejected() {
```
openinx, line 42: nit: The PR Testing section still says `CredPropsUtilTest` covers multi-encode / single top-level layout / rejecting multi-cred when scoping is disabled. Those live in `CredPropsBaseTest` / round-trip (and the multi-reject case is gone). Please update the description to match.

openinx (commented):
**Left comments.** Focused on test design.

**Summary**
- Encodes multi-cred prefixes via `CredPropsBuilder`, seeds init secrets only for single-cred, adds Hadoop round-trip + Spark scoped-FS coverage.
- Risk: matrix tests largely assert the encoder’s current map shape (including vacuous cases); the Spark FS double re-mocks the fetcher and can miss what was actually encoded.

Inline comments cover the blocking items and nits.

connectors/hadoop/src/test/java/io/unitycatalog/hadoop/internal/CredPropsBaseTest.java
```diff
@@ -139,6 +142,56 @@ void credPropsMatchExactlyAcrossMatrix(
...
+      Map<String, String> expected =
+          expectedForMultipleCredentials(kind, renew, credScoped, conf, prefixes, appVersions);
+      assertThat(created).containsExactlyInAnyOrderEntriesOf(expected);
+    }
+  }
+
+  @ParameterizedTest(
+      name = "{0} renew={1} credScoped={2} customImpl={3} expiring={4} appVersions={5}")
+  @MethodSource("testCaseMatrix")
+  void malformedCredentialPrefixesSkippedDoesNotThrow(
```
openinx, line 170: nit: The name undersells the exact-map asserts, and null/empty prefixes aren’t really "malformed." Something like `nullOrEmptyPrefixesAreOmittedFromEncodedProps` would be clearer.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
```diff
@@ -97,10 +99,35 @@ public Map<String, String> build() {
...
+
+    // Set the props for initial credential prefixes.
+    List<String> credPrefixes =
+        initialCredentials.stream()
+            .map(GenericCredential::prefix)
+            .filter(prefix -> prefix != null && !prefix.isEmpty())
```
openinx, line 123: nit: Why silently drop null/empty prefixes instead of failing when that leaves zero usable prefixes for a multi-cred response? Silent filtering can hide a bad vend.

connectors/spark/src/test/java/io/unitycatalog/spark/MultiCredentialScopedFsTest.java
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Verifies that one Spark query can read a UC-registered Parquet table whose files span two
+ * credential-scoped paths in the same bucket.
+ *
+ * <pre>
+ * s3://bucket/location-a/    credential A
+ * ├── part-*.parquet
+ * └── location-b/            credential B (more specific)
+ *     └── part-*.parquet
+ * </pre>
```
openinx, line 31: nit: Coverage is nested `location-b` under `location-a` only. A sibling-prefix case would better match `PerPathCredentialTestFileSystem`'s claimed layout support.

connectors/spark/src/test/java/io/unitycatalog/spark/PerPathCredentialTestFileSystem.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  @Override
+  protected String scheme() {
+    return "s3:";
+  }
+
+  private static boolean covers(String uri, String prefix) {
+    return (uri + "/").startsWith(prefix + "/");
```
openinx, line 68: nit: `covers()` reimplements longest-prefix matching beside `CredentialUtil`. Prefer sharing to avoid drift.

connectors/hadoop/src/main/java/io/unitycatalog/hadoop/internal/props/CredPropsBuilder.java
openinx: ignore this comment.

connectors/spark/src/test/java/io/unitycatalog/spark/PerPathCredentialTestFileSystem.java
openinx: ignore this comment.

PR #1706, opened 2026-07-23
Remove redundant setup.py from Python client build

yili-db:
**PR Checklist**

**Description of changes**
The Python client builds with hatchling via pyproject.toml, so setup.py was an unused second copy of the version string. Drop the file, its copy step, and the version-sync logic so version.sbt -> pyproject.toml is the single source of truth.


timothyw553 (approved):

openinx (approved):
Looks good to me,  thanks @yili-db for the cleaning.

tdas (commented):
`clients/python/build/.openapi-generator-ignore:7`
[AI-assisted]

you dropped setup.py everywhere except here — the ignore still lists it, and you also edited the readme to say only pyproject.toml is excluded. so does the generator emit its own setup.py? if yes, this line has to stay to suppress it and the readme is now wrong; if no, drop this stale line too. right now the two disagree.

the next person reconciling them either 'cleans up' this line thinking it's stale and the generator starts emitting the very setup.py you just removed, or keeps trusting a readme that no longer matches the ignore rules.

tdas (approved):
other than that small comment... LGTM.

PR #1708, opened 2026-07-23
[Spark] Preserve generated and identity columns in catalog schemas

timothyw553:
## What

- Add a small `CatalogV2UtilWithColumnMetadata` that delegates to Spark’s version-specific converter, then preserves generated-expression and identity-column metadata from each V2 `Column`.
- Convert once inside the existing `UCSingleCatalog.createTable(Array[Column], ...)` override, retain its routing logic, and pass the resulting `StructType` to the delegate.
- Override the three `Array[Column]` staging defaults so they use the same converter before calling UC’s existing `StructType` staging methods.
- Keep comments, default values, metadata JSON, and Spark 4.2 column IDs owned by Spark’s native converter. The converter writes only Spark metadata keys; downstream formats remain responsible for their own metadata.

## Why

Spark 4.0–4.2 can drop generated expressions and identity specifications when `CatalogV2Util.v2ColumnsToStructType` converts V2 columns. Downstream catalogs therefore receive an incomplete schema when requests pass through `UCSingleCatalog`. Preserving both representations in one shim also lets Delta’s central create path translate them into Delta feature metadata without format-specific logic in UC.

## Testing

- `SparkCatalogCompatibilitySuite` on Spark 4.0.0 and 4.1.0
- Focused `SparkCatalogCompatibilitySuite` on Spark 4.2.0 with the unrelated, unpublished Delta 4.2 test dependency excluded
- Delta + UC generated-column integration suite on Spark 4.2 (19 tests)

### DIFF
--- a/connectors/spark/src/main/scala/org/apache/spark/sql/connector/catalog/CatalogV2UtilShim.scala
+++ b/connectors/spark/src/main/scala/org/apache/spark/sql/connector/catalog/CatalogV2UtilShim.scala
@@ -1,0 +1,20 @@
+package org.apache.spark.sql.connector.catalog
+
+import org.apache.spark.sql.catalyst.util.GeneratedColumn
+import org.apache.spark.sql.types.{MetadataBuilder, StructType}
+
+/** Converts V2 columns to a schema while preserving generated-column metadata. */
+object CatalogV2UtilShim {
+
+  def v2ColumnsToStructType(columns: Array[Column]): StructType = {
+    val schema = CatalogV2Util.v2ColumnsToStructType(columns)
+    StructType(schema.fields.zip(columns).map { case (field, column) =>
+      Option(column.generationExpression()).fold(field) { expression =>
+        field.copy(metadata = new MetadataBuilder()
+          .withMetadata(field.metadata)
+          .putString(GeneratedColumn.GENERATION_EXPRESSION_METADATA_KEY, expression)
+          .build())
+      }
+    })
+  }
+}
--- a/connectors/spark/src/main/scala/org/apache/spark/sql/connector/catalog/CatalogV2UtilShim.scala
+++ b/connectors/spark/src/main/scala/org/apache/spark/sql/connector/catalog/CatalogV2UtilShim.scala
@@ -1,20 +1,27 @@
 package org.apache.spark.sql.connector.catalog
 
-import org.apache.spark.sql.catalyst.util.GeneratedColumn
+import org.apache.spark.sql.catalyst.util.{GeneratedColumn, IdentityColumn}
 import org.apache.spark.sql.types.{MetadataBuilder, StructType}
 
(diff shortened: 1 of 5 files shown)


openinx (approved):

connectors/spark/src/main/scala/org/apache/spark/sql/connector/catalog/CatalogV2UtilShim.scala
```diff
@@ -0,0 +1,27 @@
+package org.apache.spark.sql.connector.catalog
+
+import org.apache.spark.sql.catalyst.util.{GeneratedColumn, IdentityColumn}
+import org.apache.spark.sql.types.{MetadataBuilder, StructType}
+
+/** Converts V2 columns to a schema while preserving generated and identity column metadata. */
+object CatalogV2UtilShim {
```
openinx: Since this is not under the shim directory, let's not name it as `CatalogV2UtilShim`.

connectors/spark/src/test/scala/io/unitycatalog/spark/compat/SparkCatalogCompatibilitySuite.scala
```diff
@@ -3,11 +3,67 @@ package io.unitycatalog.spark.compat
...
+import org.apache.spark.sql.connector.catalog.{CatalogV2UtilShim, Column, ColumnDefaultValue, IdentityColumnSpec}
+import org.apache.spark.sql.connector.expressions.Expressions
+import org.apache.spark.sql.types.DataTypes
 import org.junit.jupiter.api.Assertions.{assertEquals, assertNotNull, assertTrue}
 import org.junit.jupiter.api.Test
 
 class SparkCatalogCompatibilitySuite {
 
+  @Test
+  def v2ColumnsToStructTypePreservesGenerationExpression(): Unit = {
```
openinx: Please move those tests out of this suite, and file a `CatalogV2UtilSuite` to cover it, that make more sense.

connectors/spark/src/main/scala/org/apache/spark/sql/connector/catalog/CatalogV2UtilShim.scala
timothyw553: Good point. Renamed it to `CatalogV2UtilWithColumnMetadata` to describe the shared converter behavior without implying that it lives under a version-selected shim directory.

connectors/spark/src/test/scala/io/unitycatalog/spark/compat/SparkCatalogCompatibilitySuite.scala
timothyw553: Done. I moved the generated- and identity-column conversion tests into a dedicated `CatalogV2UtilSuite`; `SparkCatalogCompatibilitySuite` now only contains its storage-format compatibility test.

PR #1710, opened 2026-07-24
closed

foss-contributor:


api/delta.yaml
```diff
@@ -544,6 +544,7 @@ components:
       enum:
         - MANAGED
         - EXTERNAL
+        - MANAGED_SHALLOW_CLONE # Not yet fully implemented
```
yili-db, line 547: run `sbt generate` and include the generated changes.

foss-contributor: DeltaTableType seems to be empty and doesn't generate after running it?

yili-db: Ah I see. It's a separate problem then.

yili-db (approved):

api/delta.yaml
foss-contributor: Yea I ran `JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64 build/sbt generate` and no docs were generated.

openinx (approved):
LGTM,  thanks @foss-contributor for the contribution !   @tdas   this will requires your approval to merge, pls take a look, thank you.

PR #1716, opened 2026-07-26
[Server] Fix URL transcoder hanging on responses with no body

### LINKED ISSUE #1046: tableExists requests are hanging on the Iceberg endpoints
**Describe the bug**
Newer Iceberg clients 1.8+ are using HEAD requests to check table existence: https://github.com/apache/iceberg/commit/bc36f5e33fd71335e91f04aa70a199243fd15897

For some reason the table existence endpoint defined in `IcebergRestCatalogService` does not respond to the client request and hangs.

**To Reproduce**
- I tested with the 0.3 branch and the official 0.3 docker image.
- Can be reproduced with `curl`

```
$ curl -v --head "http://localhost:8080/api/2.1/unity-catalog/iceberg/v1/catalogs/unity/namespaces/default/tables/marksheet_uniform"

* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> HEAD /api/2.1/unity-catalog/iceberg/v1/catalogs/unity/namespaces/default/tables/marksheet_uniform HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
<hanging>
```
**Expected behavior**
Expecting an `HTTP/1.1 200 OK` or exception from the server, similar to the `GET` request on the same endpoint.

- Tried the docker image: 0.3 on MacOS 15.1.1
- Built unitycatalog from scratch on MacOS 15.1.1: `branch-0.3`

**Additional context**
- Started debugging it, looks like request reaches `IcebergRestCatalogService.tableExists(...)`, but then the client does not receive a response.
- Tried upgrading `armeria` to 1.32.3, but the behavior did not change.
- Hanging even when the table does not exist.

SreeramaYeshwanthGowd:
**PR Checklist**

**Description of changes**

Fixes #1046

Any response with no body hangs forever through `URLTranscoderVerticle`. `bodyAsBuffer()` is null when there is no body, and `end(null)` throws inside the compose chain, so nothing is written and the connection stays open.

Requests arrive through the transcoder because `main` puts Armeria on `port + 1` and `compose.yaml` publishes only 8080.

Iceberg and Delta `HEAD` hung before and now answer. Filed #1717 for the failure handler.

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/URLTranscoderVerticleTest.java
+++ b/server/src/test/java/io/unitycatalog/server/URLTranscoderVerticleTest.java
@@ -1,0 +1,140 @@
+package io.unitycatalog.server;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.web.client.HttpResponse;
+import io.vertx.ext.web.client.WebClient;
+import java.io.IOException;
+import java.net.ServerSocket;
+import java.net.Socket;
+import java.util.concurrent.TimeUnit;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+public class URLTranscoderVerticleTest {
+
+  private static final String HOST = "127.0.0.1";
+  private static final int TIMEOUT_SECONDS = 10;
+
+  private static Vertx vertx;
+  private static WebClient client;
+  private static int transcodePort;
+
+  @BeforeAll
+  public static void setUp() throws Exception {
+    vertx = Vertx.vertx();
+    client = WebClient.create(vertx);
+    int servicePort = findAvailablePort();
+    transcodePort = findAvailablePort();
+    startService(servicePort);
+    vertx
+        .deployVerticle(new URLTranscoderVerticle(transcodePort, servicePort))
+        .toCompletionStage()
+        .toCompletableFuture()
+        .get(TIMEOUT_SECONDS, TimeUnit.SECONDS);
+    awaitPort(transcodePort);
+  }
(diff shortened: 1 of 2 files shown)


server/src/test/java/io/unitycatalog/server/URLTranscoderVerticleTest.java
```diff
@@ -0,0 +1,138 @@
...
+public class URLTranscoderVerticleTest {
+
+  private static final String HOST = "127.0.0.1";
+  private static final int TIMEOUT_SECONDS = 10;
+
+  private Vertx vertx;
+  private WebClient client;
+  private int transcodePort;
+
+  @BeforeEach
```
yili-db: Should be `@BeforeAll/@AfterAll` to avoid unnecessary test server startup and teardown.

server/src/test/java/io/unitycatalog/server/URLTranscoderVerticleTest.java
```diff
@@ -0,0 +1,138 @@
...
+          .toCompletableFuture()
+          .get(TIMEOUT_SECONDS, TimeUnit.SECONDS);
+    }
+  }
+
+  @Test
+  public void testNoContentResponseIsTranscoded() throws Exception {
+    HttpResponse<Buffer> response = send(HttpMethod.GET, "/no-content");
+
+    assertThat(response.statusCode()).isEqualTo(204);
```
yili-db, line 60: Also validate that body is empty?

yili-db (commented):
This is good fix!
Just two minor comments. Please address them. Thank you!

SreeramaYeshwanthGowd:
Thank you @yili-db - both addressed and pushed: setup moved to @BeforeAll/@AfterAll, and the empty body assertions added.

yili-db (approved):

PR #1719, opened 2026-07-26
[Server] Fix metastore-wide model listing page token

### LINKED ISSUE #1718: Metastore-wide GET /models returns a next_page_token that the server rejects
**Describe the bug**

`GET /api/2.1/unity-catalog/models` with no `catalog_name` or `schema_name` returns a
`next_page_token` that the same endpoint then rejects with HTTP 400.

That branch of `ModelRepository.listRegisteredModels` does not go through
`PagedListingHelper.listEntity` like the schema scoped branch does. It calls
`getAllRegisteredModelsDao`, which pages by row id:

```java
tokenToUse = UUID.fromString(token.get());
String hql = "FROM RegisteredModelInfoDAO t WHERE t.id > :token ORDER BY t.id ASC";
```

but the token returned to the client comes from `PagedListingHelper.getNextPageToken`, which returns
the last entity `name`. So `UUID.fromString("<model name>")` throws, `BaseExceptionHandler` maps it
to `INVALID_ARGUMENT`, and the caller gets a 400 for a token the server issued.

The same branch also skips the negative `max_results` check that `listEntity` performs.
`api/all.yaml` says "when set to a value less than 0, an invalid parameter error is returned", but
`GET /models?max_results=-1` returns 200 while the schema scoped form correctly returns 400.

**To Reproduce**

1. Create more registered models across the metastore than one page holds.
2. `GET /api/2.1/unity-catalog/models?max_results=2` and note `next_page_token`.
3. `GET /api/2.1/unity-catalog/models?max_results=2&page_token=<that token>` returns 400
   `Invalid UUID string: <model name>`.
4. `GET /api/2.1/unity-catalog/models?max_results=-1` returns 200 instead of 400.

**Expected behavior**

The token returned by the metastore wide listing can be passed back to fetch the next page, and
`max_results` below zero is rejected the same way the schema scoped listing rejects it.

- main at `ed504dee`

**Additional context**

Model names are only unique per schema, so the id ordering in this branch is deliberate. A name keyed
cursor would skip models whenever two schemas hold the same model name, so the fix is to make the
token match the id the query already pages by. Introduced in #404.

SreeramaYeshwanthGowd:
**PR Checklist**

**Description of changes**

Fixes #1718

`GET /models` with no catalog or schema pages by row id but returned the last model name as `next_page_token`, so passing that token back failed `UUID.fromString` and returned 400. That branch also skipped the negative `max_results` check.

Model names are not unique across a metastore, so the id ordering stays and the token now matches it.

`testPagingCoversModelsThatShareAName` walks three models sharing a name across a page, then both validations in the same test. Fails without the fix.

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/sdk/models/SdkModelCRUDTest.java
+++ b/server/src/test/java/io/unitycatalog/server/sdk/models/SdkModelCRUDTest.java
@@ -1,13 +1,30 @@
 package io.unitycatalog.server.sdk.models;
 
+import static io.unitycatalog.server.utils.TestUtils.CATALOG_NAME;
+import static io.unitycatalog.server.utils.TestUtils.CATALOG_NAME2;
+import static io.unitycatalog.server.utils.TestUtils.COMMENT;
+import static io.unitycatalog.server.utils.TestUtils.MODEL_NAME;
+import static io.unitycatalog.server.utils.TestUtils.SCHEMA_NAME;
+import static io.unitycatalog.server.utils.TestUtils.SCHEMA_NAME2;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.unitycatalog.client.ApiException;
+import io.unitycatalog.client.api.RegisteredModelsApi;
+import io.unitycatalog.client.model.CreateRegisteredModel;
+import io.unitycatalog.client.model.ListRegisteredModelsResponse;
+import io.unitycatalog.client.model.RegisteredModelInfo;
 import io.unitycatalog.server.base.ServerConfig;
 import io.unitycatalog.server.base.catalog.CatalogOperations;
 import io.unitycatalog.server.base.model.BaseModelCRUDTest;
 import io.unitycatalog.server.base.model.ModelOperations;
 import io.unitycatalog.server.base.schema.SchemaOperations;
+import io.unitycatalog.server.exception.ErrorCode;
 import io.unitycatalog.server.sdk.catalog.SdkCatalogOperations;
 import io.unitycatalog.server.sdk.schema.SdkSchemaOperations;
 import io.unitycatalog.server.utils.TestUtils;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.Test;
 
 public class SdkModelCRUDTest extends BaseModelCRUDTest {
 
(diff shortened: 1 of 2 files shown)


server/src/test/java/io/unitycatalog/server/sdk/models/SdkModelListingPaginationTest.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * Covers listing registered models across the whole metastore, with no catalog or schema filter.
+ */
+public class SdkModelListingPaginationTest extends BaseCRUDTest {
```
yili-db: Can this test file be folded into `SdkModelCRUDTest`? It's a base class of `BaseModelCRUDTest` where tests are actually defined. So currently itself has only factory methods. But it doesn't have to remain so. See `SdkTableCRUDTest` for example, you can definitely add additional tests in `SdkModelCRUDTest` and they can freely use registeredModelsApi as well. By doing this it can share the setup with other model crud tests.

SreeramaYeshwanthGowd:
Thank you @yili-db - folded into `SdkModelCRUDTest` and pushed, reusing `createCommonResources` so it shares setup with the other model tests.

server/src/test/java/io/unitycatalog/server/sdk/models/SdkModelCRUDTest.java
```diff
@@ -23,6 +42,94 @@ protected SchemaOperations createSchemaOperations(ServerConfig config) {
...
   }
+
+  /**
+   * ModelOperations drops the next page token, so the listing tests below use a direct
+   * `RegisteredModelsApi` client to inspect the response object.
+   */
+  private RegisteredModelsApi localRegisteredModelsApi;
+
+  @Test
+  public void testNextPageTokenFetchesTheRemainingModels() throws ApiException {
```
yili-db: This test can be folded into the next test `testPagingCoversModelsThatShareAName`:
1. run existing `testPagingCoversModelsThatShareAName` tests.
2. create one more model `createModel(CATALOG_NAME, SCHEMA_NAME, "uc_testmodel_two");` to make them three.
3. listAllModels again to check that all 3 can be listed.

The motivation is to reduce setup/teardown cost.

server/src/test/java/io/unitycatalog/server/sdk/models/SdkModelCRUDTest.java
```diff
@@ -23,6 +42,94 @@ protected SchemaOperations createSchemaOperations(ServerConfig config) {
...
+    ListRegisteredModelsResponse secondPage = listAllModels(1, firstPage.getNextPageToken());
+
+    assertThat(collectFullNames(firstPage, secondPage))
+        .containsExactlyInAnyOrder(
+            CATALOG_NAME + "." + SCHEMA_NAME + "." + MODEL_NAME,
+            CATALOG_NAME2 + "." + SCHEMA_NAME2 + "." + MODEL_NAME);
+  }
+
+  @Test
+  public void testNegativeMaxResultsIsRejected() {
```
yili-db: These two error cases can be folded into `testPagingCoversModelsThatShareAName` as well: just exercise the normal validation, then error asserts.

yili-db (changes requested):

SreeramaYeshwanthGowd:
Thank you @yili-db - folded into a single `testPagingCoversModelsThatShareAName`, added the third model and appended both error asserts as suggested. Pushed.

yili-db (approved):

PR #1721, opened 2026-07-27
[Server] Escape GCS paths in credential access boundary conditions

timothyw553:
**PR Checklist**

- [ ] If you have added or modified a feature, documentation in `docs` is updated

**Description of changes**

`GcpCredentialVendor` inserted decoded GCS paths directly into CEL conditions, so a quotation mark could inject an always-true expression and widen a downscoped token to the bucket. For example, `gs://bucket/team/x') || ('1'=='1` could turn the intended prefix check into an always-true CEL condition.

Render both path values with the official CEL string-quoting algorithm. Tests cover the reported payload, both quote delimiters, preceding backslashes, control characters, escape-looking text, Unicode, and ordinary paths. The focused suite passes all 10 tests; restoring raw interpolation makes 7 fail.


server/src/main/java/io/unitycatalog/server/service/credential/gcp/GcpCredentialVendor.java
```diff
@@ -85,6 +85,41 @@ private GcpCredentialGenerator createGenerator(GcsStorageConfig storageConfig) {
     return new ServiceAccountCredentialGenerator(jsonKeyFilePath);
   }
 
+  /**
+   * Converts text into a double-quoted CEL string literal.
```
yili-db, line 89: Worthy documenting:
```
emits a double-quoted literal because the escaper escapes \", not '; do not change the delimiter
```
Just in case someone in the future sees this as inconsistent and casually changes the `'"'` to `'\''`

timothyw553: Documented this invariant in the method Javadoc. It now calls out that the literal uses double quotes because those are the quotation marks the encoder escapes, and that changing only the delimiter would make apostrophes unsafe.

yili-db (approved):

PR #1017, opened 2025-05-23
[SC-193285][PROTOTYPE] Read databricks ST/MVs

xiaonanyang-db:
**PR Checklist**

- [ ] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

PR #1722, opened 2026-07-27
[Spark] Keep SHOW TABLES metadata-only on Spark 4.2

chenwang-databricks:
## Description of changes

On Spark 4.2, `SHOW TABLES` against `UCSingleCatalog` fails when the target schema contains a table that cannot be read from outside the compute environment (e.g. a managed / `TABLE_STANDARD` table). A metadata-only listing operation eagerly vends per-table storage credentials, and a single non-externalizable table aborts the whole command:

```
generateTemporaryTableCredentials call failed with: 400 - INVALID_PARAMETER_VALUE
Table with id <id> cannot be accessed from outside of Databricks Compute Environment due to its
kind being TABLE_STANDARD. Only 'TABLE_EXTERNAL', 'TABLE_DELTA_EXTERNAL' and 'TABLE_DELTA'
table kinds can be accessed externally.
reason: EXTERNAL_ACCESS_NOT_ALLOWED_FOR_TABLE
```

### Root cause

Spark 4.2 routes `SHOW TABLES` through `RelationCatalog.listRelationSummaries`, whose default implementation unions `TableCatalog.listTableSummaries` (tables) with `listViews` (views). The default `listTableSummaries` calls `loadTable` on **every** table just to read its `table_type`, and `UCProxy.loadTable` → `loadV1Table` unconditionally vends storage credentials via `buildForTable(READ_WRITE)`. For a non-externalizable table the credential vend returns 400 and the whole `SHOW TABLES` fails.

`SHOW VIEWS` is unaffected because the connector already overrides `listViews` to build identifiers straight from the list response, with no `loadTable` and no credentials. The table-listing path lacked the equivalent override.

This is Spark 4.2 only: on Spark 4.0/4.1 `UCSingleCatalog` is a plain `TableCatalog` (the shim is an empty marker trait), so `ShowTablesExec` takes the cheap `listTables` branch. Only the Spark 4.2 shim mixes in `RelationCatalog`.

### Fix

Override `listTableSummaries` on `UCProxy` (Spark-4.2 shim) so table listing stays metadata-only, mirroring the existing `listViews` override — each `TableSummary` is built from the single `listTables` RPC row using its `getTableType()`, with **no `loadTable` and no credential vend**. A delegating override is added on `UCSingleCatalog` next to the existing `listViews` delegate so it bypasses the `delegate`/Delta chain. Spark's default `listRelationSummaries` composes this with `listViews`.

The UC-type → Spark `TableSummary` type-string mapping lives in `UCViewTypes` (as string literals, so it stays free of Spark-4.2-only types and compiles on all supported Spark versions), reproducing the default's `FOREIGN` fallback.

Loading such a table for a real read (e.g. `SELECT`) still correctly fails, since `TABLE_STANDARD` data genuinely cannot be read externally. This change only stops metadata listing from triggering that path.

## Test plan

- [x] Spark 4.2.0 (GA) connector compile clean (main + test); checkstyle clean.
- [x] `UCViewProxySuite` passes **22/22** (18 pre-existing + 4 new), verified against Spark 4.2.0 GA.
- New Spark-4.2 unit tests assert the listing is tables-only, maps types correctly (`MANAGED`/`EXTERNAL`/`STREAMING_TABLE`→`EXTERNAL`/unknown→`FOREIGN`), unions tables + views, delegates through `UCSingleCatalog`, and — the core regression guard — **never calls `getTable`/`loadTable`**.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
@@ -101,14 +101,20 @@ private[spark] object UCViewTypes {
    * (not `TableSummary.*` constants) so this stays free of Spark-4.2-only types and compiles on
    * all supported Spark versions, mirroring the `"METRIC_VIEW"` literal in `viewLikeUcTypes`.
    *
-   * The mapping deliberately reproduces the semantics of Spark's default
-   * `TableCatalog.listTableSummaries`, which reads each table's `PROP_TABLE_TYPE` and falls back to
-   * `FOREIGN` when it is absent -- except it derives the type from the list-RPC row's
-   * `getTableType()` instead of a per-table `loadTable`, so no storage credentials are vended:
+   * The mapping stays consistent with what the credential-vending load path
+   * (`getUCTableLike` -> `loadV1Table`) would produce, so a table's summary type matches the type
+   * it would resolve to under the default `listTableSummaries` (which reads `PROP_TABLE_TYPE` off
+   * the loaded table and falls back to `FOREIGN` when absent) -- except it derives the type from
+   * the list-RPC row's `getTableType()` instead of a per-table `loadTable`, so no storage
+   * credentials are vended:
    *
    *   - `MANAGED`          -> `"MANAGED"`   (`TableSummary.MANAGED_TABLE_TYPE`)
+   *   - `STREAMING_TABLE`  -> `"MANAGED"`   -- `getUCTableLike` fetches with
+   *                                            `readStreamingTableAsManaged = true`, so the load
+   *                                            path sees UC `MANAGED` and `loadV1Table` builds a
+   *                                            `CatalogTableType.MANAGED` V1 table (see #1017);
+   *                                            classify it the same way here
    *   - `EXTERNAL`         -> `"EXTERNAL"`  (`TableSummary.EXTERNAL_TABLE_TYPE`)
-   *   - `STREAMING_TABLE`  -> `"EXTERNAL"`  (loaded as an EXTERNAL V1 table by `loadV1Table`)
    *   - anything else      -> `"FOREIGN"`  (`TableSummary.FOREIGN_TABLE_TYPE`), matching the
    *                                         default's absent-property fallback
    *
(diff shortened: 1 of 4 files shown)


connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
```diff
@@ -95,6 +95,33 @@ private[spark] object UCViewTypes {
...
+   * all supported Spark versions, mirroring the `"METRIC_VIEW"` literal in `viewLikeUcTypes`.
+   *
+   * The mapping deliberately reproduces the semantics of Spark's default
+   * `TableCatalog.listTableSummaries`, which reads each table's `PROP_TABLE_TYPE` and falls back to
+   * `FOREIGN` when it is absent -- except it derives the type from the list-RPC row's
+   * `getTableType()` instead of a per-table `loadTable`, so no storage credentials are vended:
+   *
+   *   - `MANAGED`          -> `"MANAGED"`   (`TableSummary.MANAGED_TABLE_TYPE`)
+   *   - `EXTERNAL`         -> `"EXTERNAL"`  (`TableSummary.EXTERNAL_TABLE_TYPE`)
+   *   - `STREAMING_TABLE`  -> `"EXTERNAL"`  (loaded as an EXTERNAL V1 table by `loadV1Table`)
```
yili-db: This contradicts with:
```
  private[spark] def getUCTableLike(ident: Identifier): Option[UCTableInfo] = {
    val fullName = UCSingleCatalog.fullTableNameForApi(this.name, ident)
    try {
      Some(tablesApi.getTable(
        fullName,
        /* readStreamingTableAsManaged = */ true,
        /* readMaterializedViewAsManaged = */ true))
```
Why?

chenwang-databricks: Good catch — you're right, that was inconsistent. `getUCTableLike` fetches with `readStreamingTableAsManaged = true`, so the load path sees UC `MANAGED` and `loadV1Table` builds a `CatalogTableType.MANAGED` V1 table (per #1017). Mapping `STREAMING_TABLE` to `EXTERNAL` in the summary contradicted that, and my rationale comment ("loaded as an EXTERNAL V1 table") was just wrong.

Fixed in 4dcee690: `STREAMING_TABLE -> MANAGED`, with the doc comment and test (`testListTableSummariesMapsStreamingTableToManagedAndUnknownToForeign`) updated to match. The summary type now matches what the credential-vending load path would have produced.

yili-db (approved):

PR #1723, opened 2026-07-27
[Spark] Fix Spark tests against Delta master-SNAPSHOT

yili-db:
**PR Checklist**

**Description of changes**

Align the Spark connector tests with current Delta `master` behavior:

- `ServerSidePlanningTest`: SSP-enabled `loadTable()` no longer forces a `_delta_log` read, so it succeeds without throwing on all supported Delta versions.
- `AwsCredRenewITTest` / `AbfsCredRenewITTest` / `GcsCredRenewITTest`: register the tracking filesystem as `spark.hadoop.fs.<scheme>.impl` so `CredScopedFileSystem` captures it as `fs.<scheme>.impl.original` and delegates to it (`credScopedFs.enabled` stays at its default `true`).

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AbfsCredRenewITTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AbfsCredRenewITTest.java
@@ -30,7 +30,13 @@ public class AbfsCredRenewITTest extends BaseCredRenewITTest {
 
   @Override
   protected Map<String, String> catalogExtraProps() {
-    return Map.of("fs.abfs.impl", AbfsCredFileSystem.class.getName());
+    // Register the tracking filesystem as a `spark.hadoop.fs.<scheme>.impl` property (not a bare
+    // catalog option) so it lands in the session Hadoop configuration. With credScopedFs enabled
+    // (the default), CredPropsUtil reads it there and saves it under `fs.<scheme>.impl.original`
+    // before installing CredScopedFileSystem, which then restores and delegates to it.
+    return Map.of(
+        "spark.hadoop.fs.abfs.impl", AbfsCredFileSystem.class.getName(),
+        "spark.hadoop.fs.abfss.impl", AbfsCredFileSystem.class.getName());
   }
 
   public static class AzureCredGenerator extends TimeBasedCredGenerator<AzureCredential>
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AbfsCredRenewITTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AbfsCredRenewITTest.java
@@ -30,10 +30,6 @@ public class AbfsCredRenewITTest extends BaseCredRenewITTest {
 
   @Override
   protected Map<String, String> catalogExtraProps() {
-    // Register the tracking filesystem as a `spark.hadoop.fs.<scheme>.impl` property (not a bare
-    // catalog option) so it lands in the session Hadoop configuration. With credScopedFs enabled
-    // (the default), CredPropsUtil reads it there and saves it under `fs.<scheme>.impl.original`
-    // before installing CredScopedFileSystem, which then restores and delegates to it.
     return Map.of(
         "spark.hadoop.fs.abfs.impl", AbfsCredFileSystem.class.getName(),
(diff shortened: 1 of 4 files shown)


connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AbfsCredRenewITTest.java
```diff
@@ -30,7 +30,13 @@ protected String scheme() {
 
   @Override
   protected Map<String, String> catalogExtraProps() {
-    return Map.of("fs.abfs.impl", AbfsCredFileSystem.class.getName());
+    // Register the tracking filesystem as a `spark.hadoop.fs.<scheme>.impl` property (not a bare
+    // catalog option) so it lands in the session Hadoop configuration. With credScopedFs enabled
+    // (the default), CredPropsUtil reads it there and saves it under `fs.<scheme>.impl.original`
+    // before installing CredScopedFileSystem, which then restores and delegates to it.
```
openinx: This comment is verbose, which I think we can remove it directly.   the following change looks good to me.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AwsCredRenewITTest.java
```diff
@@ -35,7 +35,13 @@ protected String scheme() {
 
   @Override
   protected Map<String, String> catalogExtraProps() {
-    return Map.of("fs.s3.impl", S3CredFileSystem.class.getName());
+    // Register the tracking filesystem as a `spark.hadoop.fs.<scheme>.impl` property (not a bare
```
openinx: ditto.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/GcsCredRenewITTest.java
```diff
@@ -31,9 +31,12 @@ protected String scheme() {
 
   @Override
   protected Map<String, String> catalogExtraProps() {
-    // Override fs.gs.impl to use our testing filesystem that tracks credential renewals
-    // Note: fs.gs.impl.disable.cache is already set by CredPropsUtil.GcsPropsBuilder
-    return Map.of("fs.gs.impl", GcsCredFileSystem.class.getName());
+    // Register the tracking filesystem as a `spark.hadoop.fs.gs.impl` property (not a bare catalog
```
openinx: ditto.

connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AbfsCredRenewITTest.java
openinx: one question: how will it works before without the `spark.hadoop.` prefix  ? and do you know what is the exact change that break the previous success ?

openinx (approved):
The code changes look good to me, but left one question.

timothyw553 (approved):

connectors/spark/src/test/java/io/unitycatalog/spark/auth/storage/AbfsCredRenewITTest.java
yili-db: This is rather complicated issue. Took many rounds of AI to figure out. It was triggered by turning on CredScopedFileSystem in test in Delta code base.

But why did fs.s3.impl worked before? Because without CredScopedFileSystem, it's consumed by  code like this:
```
val fs = path.getFileSystem(spark.sessionState.newHadoopConfWithOptions(options))
```

But with CredScopedFileSystem, it's like this:
```
    val hadoopConfSupplier: Supplier[Configuration] = () =>
      SparkSession.getActiveSession
        .map(_.sparkContext.hadoopConfiguration)
        .getOrElse(new Configuration())
...

return UCCredentialHadoopConfs.builder(baseUri, scheme)
...
.hadoopConf(hadoopConfSupplier.get());
```

- `newHadoopConf/newHadoopConfWithOptions` can access `fs.s3.impl` but they are session specific.
- `sparkContext.hadoopConfiguration` doesn't have access to `fs.s3.impl` but only `spark.hadoop.fs.<scheme>.impl`. But clients built by `UCTokenBasedRestClientFactory` is cached and shared between sessions so it can only use `hadoopConfiguration`.

PR #1725, opened 2026-07-28
Upgrade Delta from 4.2 to 4.3.1

yili-db:
**PR Checklist**

**Description of changes**

Upgrade Delta from 4.2 to 4.3.1

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
@@ -811,6 +811,16 @@ public abstract class BaseTableReadWriteTest extends BaseSparkIntegrationTest {
                 ColumnTypeName.VARIANT,
                 "variant",
                 "\"variant\""));
+    if (!isSparkAtLeast(4, 1)) {
+      // VARIANT round-trips through a managed Delta table only when Delta's `variantType` writer
+      // table feature is supported. The Delta build for Spark 4.0 does not support it (writes fail
+      // with DELTA_UNSUPPORTED_FEATURES_FOR_WRITE); the Spark 4.1+ build does. Gate the column on
+      // the running Spark version so the rest of the type coverage still runs on Spark 4.0.
+      cols =
+          cols.stream()
+              .filter(col -> !col.getName().equals("col_variant"))
+              .collect(Collectors.toList());
+    }
 
     session = createSparkSessionWithCatalogs(SPARK_CATALOG, CATALOG_NAME);
     String tableName = TEST_TABLE + "_complex_type";
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -40,7 +40,7 @@ jobs:
     runs-on: ubuntu-latest
 
     # Build a human-readable job name from the matrix entry, e.g.:
-    #   "All tests (Spark 4.0.0, Delta 4.2.0)"
+    #   "All tests (Spark 4.0.0, Delta 4.3.1)"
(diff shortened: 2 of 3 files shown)


connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -811,6 +811,16 @@ public void testTableWithSupportedDataTypes() throws ApiException {
...
                 "\"variant\""));
+    if (!isSparkAtLeast(4, 1)) {
+      // VARIANT round-trips through a managed Delta table only when Delta's `variantType` writer
+      // table feature is supported. The Delta build for Spark 4.0 does not support it (writes fail
+      // with DELTA_UNSUPPORTED_FEATURES_FOR_WRITE); the Spark 4.1+ build does. Gate the column on
+      // the running Spark version so the rest of the type coverage still runs on Spark 4.0.
+      cols =
+          cols.stream()
+              .filter(col -> !col.getName().equals("col_variant"))
+              .collect(Collectors.toList());
```
openinx, line 822: nit:  we can replace the `collect(Collectors.toList())` with the `.toList` directly ?

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
```diff
@@ -811,6 +811,16 @@ public void testTableWithSupportedDataTypes() throws ApiException {
                 ColumnTypeName.VARIANT,
                 "variant",
                 "\"variant\""));
+    if (!isSparkAtLeast(4, 1)) {
```
openinx, line 814: Why the delta version upgrade will affect the spark related test ?

openinx (approved):

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: Spark connector still builds with Java 11 for 4.1:
```
      // Spark 4.2+ connector tests reference Java records (e.g. Dependency.table); require --release 17.
      val testRelease =
        if (CrossSparkVersions.getSparkVersionSpec().isAtLeast(4, 2)) {
          javacRelease17
        } else {
          javacRelease11
        }
```

openinx: Okay,  then we are fine.

connectors/spark/src/test/java/io/unitycatalog/spark/BaseTableReadWriteTest.java
yili-db: Good question — the Delta bump doesn't change VARIANT support itself, it just un-skips this test.

`testTableWithSupportedDataTypes` is `@EnabledIf("canUpdateColumnsToUC")`, which for managed Delta tables requires Delta >= 4.3 (`MIN_DELTA_VERSION_FOR_UC_DELTA_API`). Under the old default (Delta 4.2) the test was skipped for managed Delta, so the `col_variant` write never ran. Bumping the default to 4.3.1 makes it run for the first time, and it then hits a Spark-axis limitation: the `delta-spark_4.0` build doesn't support the `variantType` writer feature (`DELTA_UNSUPPORTED_FEATURES_FOR_WRITE`), while `delta-spark_4.1` does — which is why only the Spark 4.0.0 job failed and 4.1.0 passed. So the gate is on the Spark version, where the feature is actually available.

tdas (approved):

PR #1729, opened 2026-07-28
fix: add terminal failure handler to URL transcode Future chain

### LINKED ISSUE #1717: URL transcoder leaves the request unanswered when the backend call fails
**Describe the bug**

`URLTranscoderVerticle` builds its response inside a `Future` chain that has no terminal failure
handler, so any failure in that chain leaves the request unanswered and the client waits until it
times out.

The chain is `transcodeRequest.body().compose(...).compose(...)` at
`server/src/main/java/io/unitycatalog/server/URLTranscoderVerticle.java`. If either stage fails, for
example because the request to the Armeria backend cannot be made, nothing writes to
`transcodeRequest.response()` and the connection stays open.

This matters because the transcoder is on the port users normally reach. `UnityCatalogServer.main`
starts Armeria on `port + 1` and the transcoder on `port`, and `compose.yaml` publishes only 8080.

This is the residual case noted in #1716. That PR fixed the null body path, which was the cause of
#1046. It deliberately did not add a failure handler, since that is a new wire behaviour rather than
a guard.

**To Reproduce**

1. Run the transcoder logic with the null body fix from #1716 already applied, pointing at a backend
   port with nothing listening on it.
2. Send any request to the transcoder port.
3. The request never completes.

```console
$ curl -s -o /dev/null -w "http_code=%{http_code} time=%{time_total}s\n" --max-time 8 \
    "http://127.0.0.1:19080/api/2.1/unity-catalog/catalogs"
http_code=000 time=8.002384s
$ echo $?
28
```

**Expected behavior**

A failure in the chain returns a response to the client, so the request completes instead of hanging.
Something in the 502 or 500 family seems reasonable, but the status is worth agreeing on before
implementing.

- main at `ed504dee`
- Vert.x 4.3.5, the version pinned in `build.sbt`

**Additional context**

Happy to send a PR once there is agreement on the status code to return.

waterWang:
## Problem

`URLTranscoderVerticle` builds its response inside a `Future` chain that has no terminal failure handler, so any failure in that chain leaves the request unanswered and the client waits until it times out.

The chain is `transcodeRequest.body().compose(...).compose(...)` at `server/src/main/java/io/unitycatalog/server/URLTranscoderVerticle.java`. If either stage fails, for example because the request to the Armeria backend cannot be made, nothing writes to `transcodeRequest.response()` and the connection stays open.

## Fix

Add an `.otherwise()` handler to the Future chain that:
1. Logs the error
2. Sets a 502 status code
3. Ends the response with the error message
4. Checks `headWritten()` to avoid double-write in edge cases

Fixes #1717

### DIFF
--- a/api/all.yaml
+++ b/api/all.yaml
@@ -1528,7 +1528,7 @@ components:
         comment:
           type: string
           maxLength: 65536
-          minLength: 1
+          minLength: 0
           description: The comment attached to the volume
         storage_location:
           type: string
@@ -1556,7 +1556,7 @@ components:
         comment:
           type: string
           maxLength: 65536
-          minLength: 1
+          minLength: 0
           description: The comment attached to the volume
         new_name:
           description: New name for the volume.
@@ -1576,7 +1576,7 @@ components:
         comment:
           type: string
           maxLength: 65536
-          minLength: 1
+          minLength: 0
           description: The comment attached to the volume
         owner:
           description: The identifier of the user who owns the volume
--- a/server/src/main/java/io/unitycatalog/server/URLTranscoderVerticle.java
+++ b/server/src/main/java/io/unitycatalog/server/URLTranscoderVerticle.java
@@ -58,6 +58,17 @@ class URLTranscoderVerticle extends AbstractVerticle {
                     // anything is written, leaving the client waiting on a response it never gets.
                     Buffer body = resp.bodyAsBuffer();
                     return body == null ? transcodeResp.end() : transcodeResp.end(body);
+                  })
+              .otherwise(
+                  err -> {
+                    LOGGER.error("URL transcode failed", err);
+                    HttpServerResponse transcodeResp = transcodeRequest.response();
+                    if (!transcodeResp.headWritten()) {
+                      transcodeResp.setStatusCode(502);
+                      transcodeResp.putHeader("Content-Type", "text/plain");
+                      transcodeResp.end("Backend error: " + err.getMessage());
+                    }
+                    return null;
                   });
         });
 


api/all.yaml
```diff
@@ -1528,7 +1528,7 @@ components:
         comment:
           type: string
           maxLength: 65536
-          minLength: 1
+          minLength: 0
```
yili-db, line 1531: This seems to belong to another PR

server/src/main/java/io/unitycatalog/server/URLTranscoderVerticle.java
```diff
@@ -71,4 +82,4 @@ public void start() {
           }
         });
   }
-}
+}
```
yili-db, line 85: Please do not remove new line at end of file

server/src/main/java/io/unitycatalog/server/URLTranscoderVerticle.java
```diff
@@ -58,6 +58,17 @@ public void start() {
...
                     return body == null ? transcodeResp.end() : transcodeResp.end(body);
+                  })
+              .otherwise(
+                  err -> {
+                    LOGGER.error("URL transcode failed", err);
+                    HttpServerResponse transcodeResp = transcodeRequest.response();
+                    if (!transcodeResp.headWritten()) {
+                      transcodeResp.setStatusCode(502);
+                      transcodeResp.putHeader("Content-Type", "text/plain");
+                      transcodeResp.end("Backend error: " + err.getMessage());
```
yili-db, line 69: `err.getMessage` may leak internal info. Something like `transcodeResp.end("Backend request failed");` is safer.

yili-db (changes requested):

yili-db (changes requested):
Needs a test for the new failure path. The existing `URLTranscoderVerticleTest` is already set up for this — deploy a transcoder pointed at a dead backend port (use `findAvailablePort()` for a port with nothing listening) and assert the request returns 502 instead of hanging. The bounded `.get(TIMEOUT_SECONDS, ...)` in `send(...)` means a regression fails as a timeout.

Thank you!

PR #1730, opened 2026-07-29
build: replace deprecated asciitable with freva ascii-table

roeap:
## Why

The CLI renders human-readable table output with `de.vandermeer:asciitable:0.3.2`, last released in **2017** and effectively abandoned. It transitively pulls the StringTemplate/ANTLR chain:

```
de.vandermeer:asciitable:0.3.2
  → de.vandermeer:skb-interfaces
      → org.antlr:ST4:4.0.8 → org.antlr:antlr-runtime:3.5.2   (ANTLR 3)
      → org.antlr:antlr4:4.5.1
```

This collides with the Spark connector's `org.antlr:antlr4`/`antlr4-runtime` **4.13.1**, and the build has had to special-case `org.antlr:ST4` in the `spark` project's `licenseDepExclusions` purely because asciitable puts ST4 on the classpath.

## What

- Swap `de.vandermeer:asciitable:0.3.2` → **`com.github.freva:ascii-table:1.12.1`** (MIT, actively maintained, **zero runtime dependencies** — only a compile-scoped `org.jspecify` annotations jar). This removes the entire ANTLR chain from the `cli` classpath (`sbt cli/dependencyTree` shows no `ST4`/`antlr-runtime`/`antlr4:4.5.1`).
- Drop the now-unnecessary `org.antlr:ST4` case from the `spark` `licenseDepExclusions`.
- Rewrite the three call sites (`CliUtils`, `DeltaKernelUtils`, `DeltaKernelReadUtils`) onto freva's `Column[]` + builder model.

## Behavior preserved

- **Same grid look** — `BASIC_ASCII` border keeps a rule between every row, matching the old output.
- **Same width fitting** — the existing `calculateFieldWidths`/`adjustColumnWidths` logic (including the id/name-full-width special case) is retained and feeds freva's per-column `maxWidth`; `maxTableWidth(outputWidth)` reproduces `render(outputWidth)`.
- Uppercased, centered headers; multi-line KEY/VALUE array cells wrap via `OverflowBehaviour.NEWLINE`.

## Behavior differences (minor)

- Long row cells truncate via `OverflowBehaviour.ELLIPSIS_RIGHT`, replacing the manual `preprocess()` helper. freva uses the single-char Unicode ellipsis `…` where the old code produced `...`.
- The Delta `table read` dump uses freva's default (right) alignment for cells (it had no explicit alignment before either).

## Testing

- `sbt cli/compile` / `sbt cli/Test/compile` — clean (checkstyle passes).
- `sbt cli/testOnly …CliCatalogCRUDTest` — green (CLI tests run with `--output json` so they exercise the JSON path, not table rendering).
- Table rendering was verified manually against sample data for all three paths (rows / KEY-VALUE / Delta). Reviewers should eyeball `bin/uc catalog list`, `bin/uc catalog get`, wide `table list`, `function get` (multi-line values), and `table read`.

This pull request was AI-assisted by Isaac.


examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
```diff
@@ -521,47 +513,56 @@ private static void processOutputAsRows(AsciiTable at, JsonNode node, int output
...
-    CWC_FixedWidth cwc = new CWC_FixedWidth();
+    Column[] columns = new Column[fieldNames.size()];
     for (int i = 0; i < fieldNames.size(); i++) {
-      cwc.add(columnWidths[i]);
+      columns[i] =
+          new Column()
+              .header(fieldNames.get(i).toUpperCase())
+              .headerAlign(HorizontalAlign.CENTER)
+              .dataAlign(HorizontalAlign.LEFT)
+              .maxWidth(columnWidths[i], OverflowBehaviour.ELLIPSIS_RIGHT);
```
openinx: freva's content limit is `maxWidth - 2*PADDING` (PADDING=1 each side). Feeding content string lengths here makes every column ~2 chars too narrow, so list output over-truncates (`unity` → `un…`) and narrow terminals throw `Range [0, -N) out of bounds` (repro: `UC_OUTPUT_WIDTH=60 bin/uc table list`).

Please compensate (e.g. `maxWidth(columnWidths[i] + 2, ...)`) or stop pairing content widths with freva `maxWidth`/`ELLIPSIS_RIGHT` this way. Same padding assumption applies if KEY/VALUE widths are meant to match the old fitting behavior.

Also: please add unit tests for row tables, KEY/VALUE wrapping, truncation, and narrow `UC_OUTPUT_WIDTH` — current CLI tests force `--output json` and never hit this path.

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelUtils.java
```diff
@@ -183,27 +184,27 @@ public static String readDeltaTable(
...
+      List<String[]> rows = new ArrayList<>();
       for (Row row : rowData) {
-        Object[] rowValues =
+        String[] rowValues =
             IntStream.range(0, schema.length)
-                .mapToObj(colOrdinal -> DeltaKernelReadUtils.getValue(row, colOrdinal))
-                .toArray();
-        at.addRow(rowValues);
-        at.addRule();
+                .mapToObj(ordinal -> String.valueOf(DeltaKernelReadUtils.getValue(row, ordinal)))
```
openinx: `String.valueOf(getValue(...))` turns SQL null into the literal `"null"`. Freva already renders a real `null` cell as blank — keep `Object[]`/`null` (or map null → `""`) instead of `String.valueOf`.

build.sbt
```diff
@@ -728,7 +728,6 @@ lazy val spark = (project in file("connectors/spark"))
       case DepModuleInfo("oro", "oro", _) => true
       case DepModuleInfo("org.glassfish", "javax.json", _) => true
       case DepModuleInfo("org.glassfish.hk2.external", "jakarta.inject", _) => true
-      case DepModuleInfo("org.antlr", "ST4", _) => true
```
openinx, line 731: Why drop the `org.antlr:ST4` exclusion here? `spark` never depended on asciitable (`spark` → `client`/`hadoop` only). `ST4` still comes from `antlr4:4.13.1` on the spark tree. This looks like an unrelated drive-by — please restore unless there is a separate, documented reason (CI green alone is not enough).

examples/cli/src/main/java/io/unitycatalog/cli/delta/DeltaKernelReadUtils.java
```diff
@@ -2,7 +2,7 @@
 
 import static io.delta.kernel.internal.util.Utils.singletonCloseableIterator;
 
-import de.vandermeer.asciitable.AsciiTable;
+import com.github.freva.asciitable.AsciiTable;
```
openinx: nit: this file only needs the new import for `{@link AsciiTable}` in the javadoc. Prefer an FQN in the javadoc so this is not a fake call-site change.

openinx (changes requested):
**Left comments.**

Swap to freva looks directionally right (drops the abandoned ANTLR/ST4 chain from the CLI), but list rendering is broken at real terminal widths and there is no coverage of the table path.

Verified locally:
- `UC_OUTPUT_WIDTH=60/80 bin/uc table list` → `Range [0, -N) out of bounds` then JSON fallback
- Normal-width `catalog list` over-truncates (`unity` → `un…`) because freva uses `limit = maxWidth - 2*PADDING`
- Spark 4.1 examples against this branch are fine (freva is CLI-only)

build.sbt
roeap: that was a mistake, i thoght this could now be dropped. sorry.

roeap:
@openinx - addressed the comments. There seems to have been a potential for div by 0 as well, which was adressed.

roeap:
cc @yili-db - this is the first pr.

examples/cli/src/main/java/io/unitycatalog/cli/utils/CliUtils.java
```diff
@@ -521,47 +528,57 @@ private static void processOutputAsRows(AsciiTable at, JsonNode node, int output
...
-    CWC_FixedWidth cwc = new CWC_FixedWidth();
+    Column[] columns = new Column[fieldNames.size()];
     for (int i = 0; i < fieldNames.size(); i++) {
-      cwc.add(columnWidths[i]);
+      columns[i] =
+          new Column()
+              .header(fieldNames.get(i).toUpperCase())
+              .headerAlign(HorizontalAlign.CENTER)
+              .dataAlign(HorizontalAlign.LEFT)
+              .maxWidth(columnWidths[i] + 2 * CELL_PADDING, OverflowBehaviour.ELLIPSIS_RIGHT);
```
openinx, line 538: padding compensation alone is not enough. freva’s maxTableWidth re-shrinks columns below the intended widths (fixed name/*id have no minWidth), so default bin/uc catalog list still shows unity → un… at width 120. At UC_OUTPUT_WIDTH=200 (no shrink) it is fine; at 60 with table list freva still errors (Max table width 60 is too small…) and falls back to JSON. Set minWidth/maxWidth to the fitted size (so freva cannot re-shrink), and/or stop double-fitting with maxTableWidth after adjustColumnWidths.

PR #1739, opened 2026-07-30
[Server] Restrict SCIM user reads to metastore admins

timothyw553:
**PR Checklist**

- [ ] If you have added or modified a feature, documentation in `docs` is updated

**Description of changes**

Before this change, any signed-in user could list all Unity Catalog users or look up another user by ID. For example, a regular user could read the admin user's email address.

This change allows those requests only for a metastore owner, which is an administrator. Regular users now receive `403 Forbidden`.

The new tests check that administrators still have access and regular users do not. If the two permission checks are removed, both tests for regular users fail.

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/sdk/access/SdkScim2UserAccessControlTest.java
+++ b/server/src/test/java/io/unitycatalog/server/sdk/access/SdkScim2UserAccessControlTest.java
@@ -1,0 +1,58 @@
+package io.unitycatalog.server.sdk.access;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import io.unitycatalog.control.ApiException;
+import io.unitycatalog.control.api.UsersApi;
+import io.unitycatalog.control.model.UserResource;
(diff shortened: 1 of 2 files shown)


server/src/test/java/io/unitycatalog/server/sdk/access/SdkScim2UserAccessControlTest.java
```diff
@@ -0,0 +1,58 @@
...
+  @BeforeEach
+  public void setUpUsers() throws Exception {
+    createTestUser(REGULAR_USER, "Regular User");
+    otherUser = createTestUser(OTHER_USER, "Other User");
+
+    ServerConfig regularUserConfig = createTestUserServerConfig(REGULAR_USER);
+    regularUsersApi = new UsersApi(createControlApiClient(regularUserConfig));
+  }
+
+  @Test
```
yili-db, line 32: can you please fold these tests into one test case?

timothyw553: done

yili-db (approved):

PR #1740, opened 2026-07-30
[Server] Require ownership to delete catalogs

timothyw553:
**PR Checklist**

**Description of changes**

Before, a user with only `USE CATALOG` could force-delete that catalog. This change requires catalog ownership (or existing metastore-admin access) instead.

For example, a browse-only user calling `DELETE /catalogs/my_catalog?force=true` is now denied, and the catalog and its schema remain.

Tests: the catalog access-control and CRUD suites pass. Restoring the old rule makes the new test fail because the delete succeeds.


server/src/test/java/io/unitycatalog/server/sdk/access/SdkCatalogAccessControlCRUDTest.java
```diff
@@ -36,6 +36,26 @@
  */
 public class SdkCatalogAccessControlCRUDTest extends SdkAccessControlBaseCRUDTest {
 
+  @Test
+  @SneakyThrows
+  public void useCatalogCannotForceDeleteCatalog() {
```
yili-db: Can you please fold this test into testCatalogAccess?

timothyw553: done

server/src/test/java/io/unitycatalog/server/sdk/access/SdkCatalogAccessControlCRUDTest.java
yili-db, line 166: Add a comment here:
```
// delete a catalog -> catalog owner -> allowed
```

yili-db (approved):

server/src/test/java/io/unitycatalog/server/sdk/access/SdkCatalogAccessControlCRUDTest.java
timothyw553: added, similar to the one above

openinx (approved):

PR #1741, opened 2026-07-30
Resolve view dependencies if not provided by Spark

TheThow:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [x] If there is a related issue, make sure it is linked to this PR.
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

Resolved the view dependencies based on parsing the sql.

Related PR: https://github.com/unitycatalog/unitycatalog/pull/1639
Related issue: https://github.com/unitycatalog/unitycatalog/issues/1279

| KEY | VALUE |
| --- | --- |
| NAME | my_view |
| CATALOG_NAME | unity |
| SCHEMA_NAME | default |
| TABLE_TYPE | VIEW |
| DATA_SOURCE_FORMAT | null |
| COLUMNS | `{"name":"as_int","type_text":"int","type_json":"{\"name\":\"as_int\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}}","type_name":"INT","type_precision":null,"type_scale":null,"type_interval_type":null,"position":0,"comment":null,"nullable":true,"partition_index":null}` <br> `{"name":"as_double","type_text":"double","type_json":"{\"name\":\"as_double\",\"type\":\"double\",\"nullable\":true,\"metadata\":{}}","type_name":"DOUBLE","type_precision":null,"type_scale":null,"type_interval_type":null,"position":1,"comment":null,"nullable":true,"partition_index":null}` |
| STORAGE_LOCATION | null |
| COMMENT | null |
| PROPERTIES | `{"view.sqlConfig.spark.sql.parquet.fieldId.write.enabled":"true","view.sqlConfig.spark.sql.parquet.fieldId.read.enabled":"true","view.sqlConfig.spark.sql.session.timeZone":"Europe/Berlin","view.sqlConfig.spark.sql.catalog.spark_catalog":"io.unitycatalog.spark.UCSingleCatalog","view.sqlConfig.spark.databricks.delta.variant.disableVariantTableFeatureForSpark40":"true","view.sqlConfig.spark.sql.ansi.enabled":"true","view.sqlConfig.spark.sql.defaultCatalog":"unity"}` |
| OWNER | null |
| CREATED_AT | 1785404222292 |
| CREATED_BY | null |
| UPDATED_AT | 1785404222292 |
| UPDATED_BY | null |
[...]

PR #1742, opened 2026-07-30
Persist view query column names and creation context on Spark 4.2

s-maling-celonis:
## Summary

Spark 4.2's `ViewCatalog` passes query-output column names and the creation-time catalog/namespace as typed `View` fields. The UC connector was dropping both on write and re-deriving them on read from the stored columns and the view's own schema. That breaks:

- **Explicit column lists** — `CREATE VIEW v (total) AS SELECT ... AS my_count` fails on read with `INCOMPATIBLE_VIEW_SCHEMA_CHANGE`
- **Unqualified references** — a view created from schema A with `SELECT * FROM t` resolves `t` against the view's own schema on read, which can return wrong data silently when a same-named table exists elsewhere

This PR persists both fields in the existing UC properties map using Spark's v1 key names (`view.query.out.*`, `view.catalogAndNamespace.*`), restores them in `toView`, and falls back to the old derivation when the keys are absent.

## Test plan

- [x] `spark/testOnly io.unitycatalog.spark.UCViewDDLIntegrationTest` on Spark 4.2.0 — 4 passed, 0 failed
- [x] Before fix: 2 new tests failed as expected (column list → `INCOMPATIBLE_VIEW_SCHEMA_CHANGE`; creation context → wrong row)
- [ ] CI Spark 4.2 matrix row (`unit-tests.yml`, non-blocking)

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
@@ -149,6 +149,9 @@ private[spark] object UCViewTypes {
   def addQueryColumnNames(
       properties: util.Map[String, String],
       queryColumnNames: Array[String]): Unit = {
+    if (queryColumnNames.isEmpty) {
+      return
+    }
     properties.put(
       CatalogTable.VIEW_QUERY_OUTPUT_NUM_COLUMNS,
       queryColumnNames.length.toString)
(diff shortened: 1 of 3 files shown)


connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
```diff
@@ -144,4 +144,64 @@ private[spark] object UCViewTypes {
...
+
+  /**
+   * Returns the persisted query-output column names when present. Absent keys mean the view was
+   * written by an older connector and the read path should fall back to the stored column names.
+   */
+  def extractQueryColumnNames(properties: util.Map[String, String]): Option[Array[String]] = {
+    Option(properties.get(CatalogTable.VIEW_QUERY_OUTPUT_NUM_COLUMNS)).map { numCols =>
+      val count = numCols.toInt
+      (0 until count)
+        .map(index => properties.get(s"${CatalogTable.VIEW_QUERY_OUTPUT_COLUMN_NAME_PREFIX}$index"))
```
yili-db:
```
        Option(properties.get(key)).getOrElse(
          throw new IllegalStateException(
            s"Corrupted view metadata: expected $numCols query-output columns but $key is missing"))
```
Throwing this error for corrupt view properties so that it's a clear error message rather than NPE.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
```diff
@@ -144,4 +144,64 @@ private[spark] object UCViewTypes {
...
+    }
+  }
+
+  /** Returns the persisted creation-time catalog and namespace when present. */
+  def extractCreationContext(
+      properties: util.Map[String, String]): Option[(String, Array[String])] = {
+    Option(properties.get(CatalogTable.VIEW_CATALOG_AND_NAMESPACE)).map { numParts =>
+      val count = numParts.toInt
+      val parts = (0 until count)
+        .map(index => properties.get(s"${CatalogTable.VIEW_CATALOG_AND_NAMESPACE_PART_PREFIX}$index"))
```
yili-db: ditto

connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
```diff
@@ -144,4 +144,64 @@ private[spark] object UCViewTypes {
...
+        .toSeq
+      (parts.head, parts.tail.toArray)
```
yili-db: `count` could be zero? then `parts` is empty.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
```diff
@@ -144,4 +144,64 @@ private[spark] object UCViewTypes {
     }
     configs
   }
+
+  /** Persists Spark's query-output column names into UC view properties. */
+  def addQueryColumnNames(
+      properties: util.Map[String, String],
+      queryColumnNames: Array[String]): Unit = {
+    properties.put(
```
yili-db, line 155: Should skip if `queryColumnNames` is empty?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
s-maling-celonis: Done — `extractQueryColumnNames` now throws `IllegalStateException` with the expected count and missing key when an indexed property is absent.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
s-maling-celonis: Same treatment applied to `extractCreationContext`.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
s-maling-celonis: Good catch — if the count is zero we now return `None` so the read path falls back instead of calling `parts.head` on an empty sequence.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewTypes.scala
s-maling-celonis: Agreed — `addQueryColumnNames` now returns early when the array is empty, leaving the properties unset so the read path keeps using the legacy fallback.

s-maling-celonis:
Thanks for the review @yili-db! Addressed all four comments in c1b7b4de:

- explicit `IllegalStateException` when indexed view metadata keys are missing on read
- skip persisting empty query-output column name lists on write
- treat a zero part count as absent metadata (fallback) instead of failing on `parts.head`

yili-db:
https://github.com/unitycatalog/unitycatalog/pull/1741 is merged and introduced some merge conflict. Please rebase main.

s-maling-celonis:
Rebased onto latest \`main\` and resolved the conflict in \`UCViewDDLIntegrationTest.java\` from #1741 (kept both the dependency tests and the view-resolution tests).

Verified locally on Spark 4.2.0: \`spark/testOnly io.unitycatalog.spark.UCViewDDLIntegrationTest\` — 6 passed, 0 failed.

yili-db (approved):

PR #1749, opened 2026-08-03
[Server] Validate Iceberg table location before vending credentials

timothyw553:
**PR Checklist**

**Description of changes**

The server stores a location for each table. An Iceberg metadata file also contains a table location. These two locations should describe the same table before the server creates temporary storage credentials.

Before this change, `loadTable` trusted the location inside the metadata file. For example, the registered table could point to `/tables/orders` while the metadata file pointed to `/tables/other`. The server could then create credentials for `/tables/other`.

This change compares the two normalized locations and rejects the request when they do not match. When they do match, credentials are created only for the registered table location.

The existing Iceberg table test now covers the mismatch case and checks that the request is rejected.

Tests: `sbt -batch "server/testOnly io.unitycatalog.server.service.IcebergRestCatalogTest"`

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
+++ b/server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
@@ -25,9 +25,12 @@ import io.unitycatalog.server.sdk.catalog.SdkCatalogOperations;
 import io.unitycatalog.server.sdk.schema.SdkSchemaOperations;
 import io.unitycatalog.server.sdk.tables.SdkTableOperations;
 import io.unitycatalog.server.service.iceberg.IcebergObjectMapper;
+import io.unitycatalog.server.utils.NormalizedURL;
 import io.unitycatalog.server.utils.TestUtils;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


yili-db (changes requested):
Please rebase

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
```diff
@@ -51,6 +51,7 @@ public class IcebergRestCatalogTest extends BaseServerTest {
 
   private static final String TEST_BASE_PREFIX = "/v1/catalogs/" + TestUtils.CATALOG_NAME;
   private static final String TEST_BASE_NON_PREFIX = "/v1";
+  private static final String ICEBERG_TABLE_LOCATION = "/tmp/uniform_iceberg_table";
```
yili-db: Does it actually write anything out here? Prefer to use a `@TempDir Path icebergTableLocation;` instead.

yili-db (commented):
Mostly good. Just a nit

yili-db (approved):

PR #1753, opened 2026-08-04
Clear Casbin hierarchy and policies on force catalog delete

s-maling-celonis:
## Summary
- On force cascade delete (`deleteCatalog(force=true)` and `deleteSchema(force=true)`), repositories return `DeletedResource` entries collected during the existing delete transaction; services clear Casbin policies and g2 hierarchy from that list.
- Replaces pre-delete listing in `CatalogService` — IDs are sourced from the delete itself (no duplicate tree traversal, no cross-transaction race).

## Problem
Force catalog/schema delete cascaded metadata removal via `CatalogRepository` / `SchemaRepository` but skipped per-child auth cleanup that `TableService` et al. perform on individual deletes, leaving orphaned Casbin rules.

## Design
- `DeletedResource(SecurableType type, String id, String parentId)` — recorded in persist layer as rows are removed.
- `AuthorizedService.clearDeletedResourceAuthorizations(List<DeletedResource>)` — shared cleanup for catalog and schema force deletes.

## Test plan
- [x] `sbt "server/testOnly io.unitycatalog.server.sdk.access.SdkCatalogAccessControlCRUDTest io.unitycatalog.server.sdk.access.SdkSchemaAccessControlCRUDTest io.unitycatalog.server.auth.JCasbinAuthorizerTest"` (Java 21)
- [ ] Full server access-control SDK suite


yili-db (changes requested):
The bug is real and the cleanup logic is correct. The bug also exists for `deleteSchema` in the same way. My concern is how the subtree is discovered.

To find the ids to clean, this lists five repositories with pagination. But `CatalogRepository.deleteCatalog(force=true)` later will traverse the whole tree again to delete it (one transaction), so it knows every deleted id — this PR constructs that list independently.
Downsides: it's an extra traversal of the entire tree and not in the same transaction. So entities added in another thread while the deleting thread runs between `collectSchemaAuthorizationSubtrees` and `catalogRepository.deleteCatalog` can be ignored.

Suggestion: have the repo delete functions return the ids they deleted and then the service clear authz from that list:
```
    public List<DeletedResource> deleteCatalog(String name, boolean force);  // {SecurableType type, String id}. Includes child schema and tables etc.
```

No re-listing, no per-type service code, and it's sourced from the delete itself — so no drift, and cleanup runs after the single commit (same ordering as SchemaService.deleteSchema today).

s-maling-celonis:
> The bug is real and the cleanup logic is correct. The bug also exists for `deleteSchema` in the same way. My concern is how the subtree is discovered.
> 
> To find the ids to clean, this lists five repositories with pagination. But `CatalogRepository.deleteCatalog(force=true)` later will traverse the whole tree again to delete it (one transaction), so it knows every deleted id — this PR constructs that list independently. Downsides: it's an extra traversal of the entire tree and not in the same transaction. So entities added in another thread while the deleting thread runs between `collectSchemaAuthorizationSubtrees` and `catalogRepository.deleteCatalog` can be ignored.
> 
> Suggestion: have the repo delete functions return the ids they deleted and then the service clear authz from that list:
> 
> ```
>     public List<DeletedResource> deleteCatalog(String name, boolean force);  // {SecurableType type, String id}. Includes child schema and tables etc.
> ```
> 
> No re-listing, no per-type service code, and it's sourced from the delete itself — so no drift, and cleanup runs after the single commit (same ordering as SchemaService.deleteSchema today).

Thanks @yili-db — agreed, returning deleted IDs from the repo delete path is the better approach here.

Reworked accordingly:

DeletedResource (SecurableType, id, parentId) is collected in SchemaRepository / CatalogRepository during the existing cascade delete transaction — no pre-listing in the service layer.
AuthorizedService.clearDeletedResourceAuthorizations drives Casbin cleanup from that list (children → schemas → catalog).
Removed the extra repository dependencies and pagination helpers from CatalogService.
Applied the same pattern to SchemaService.deleteSchema(force=true), which had the same orphaned child-policy issue.
Added forceDeleteSchemaClearsChildAuthorizations alongside the existing catalog test.

server/src/main/java/io/unitycatalog/server/service/CatalogService.java
```diff
@@ -124,10 +125,10 @@ public HttpResponse updateCatalog(
   public HttpResponse deleteCatalog(
       @Param("name") @AuthorizeResourceKey(CATALOG) String name,
       @Param("force") Optional<Boolean> force) {
-    CatalogInfo catalogInfo = catalogRepository.getCatalog(name);
-    catalogRepository.deleteCatalog(name, force.orElse(false));
-    removeAuthorizations(catalogInfo.getId());
+    catalogRepository.getCatalog(name);
```
yili-db: This seems useless now.

server/src/main/java/io/unitycatalog/server/service/CatalogService.java
```diff
@@ -72,8 +73,8 @@ public CatalogService(
   @AuthorizeResourceKey(METASTORE)
   public HttpResponse createCatalog(
       @AuthorizeResourceKey(value = EXTERNAL_LOCATION, key = "storage_root")
-      @AuthorizeKey(key = "storage_root")
-      CreateCatalog createCatalog) {
+          @AuthorizeKey(key = "storage_root")
```
yili-db: Avoid cosmetic churn.

server/src/main/java/io/unitycatalog/server/service/CatalogService.java
```diff
@@ -91,8 +92,8 @@ public HttpResponse createCatalog(
   public HttpResponse listCatalogs(
       @Param("max_results") Optional<Integer> maxResults,
       @Param("page_token") Optional<String> pageToken) {
-    ListCatalogsResponse listCatalogsResponse = catalogRepository.listCatalogs(
-        maxResults, pageToken);
+    ListCatalogsResponse listCatalogsResponse =
```
yili-db: Avoid cosmetic churn.

yili-db (commented):
Nice update. Just couple of nits. Very close to approval. Thank you!

server/src/main/java/io/unitycatalog/server/service/CatalogService.java
s-maling-celonis: Agreed — leftover from the old `removeAuthorizations(catalogInfo.getId())` path. Removed in 83da87a4; `deleteCatalog` already throws if the catalog is missing.

server/src/main/java/io/unitycatalog/server/service/CatalogService.java
s-maling-celonis: Reverted in 83da87a4.

server/src/main/java/io/unitycatalog/server/service/CatalogService.java
s-maling-celonis: Reverted in 83da87a4.

yili-db (approved):

PR #1754, opened 2026-08-04
Vend UC path credentials for bare delta cloud-path reads and writes

s-maling-celonis:
## Summary

Extends path credential vending to bare delta cloud-path reads and writes.

**Depends on #1653** — this branch is stacked on `feat/spark-path-credential-vending-upstream`. Please review only the commits after that branch; merge after #1653.

Delta analysis drops UnresolvedRelation options before DeltaLog accesses cloud storage. This change:

- Lets the existing hint-resolution rule inject credentials onto delta.`path` UnresolvedRelations and early-resolve them (before ResolveSQLOnFile) so options reach DeltaTableV2
- Early-resolves INSERT OVERWRITE / INSERT INTO delta.`path` targets to DataSourceV2Relation so Spark's V2 write path can run (the parquet analog of INSERT OVERWRITE DIRECTORY)
- Attaches PATH_READ_WRITE options on CREATE TABLE / CREATE TABLE AS SELECT / REPLACE TABLE AS SELECT for delta.`path`
- Adds DeltaPathCredentialSupport (reflection-based, no compile-time delta-spark dependency) with a post-resolution fallback on Delta plan nodes
- Extracts PathCredentialOptions for shared fs.* and option.fs.* credential merging
- Keeps the analyzer fixed-point idempotence guard, including the option.fs.* keys Delta stores

The parser stays untouched. INSERT OVERWRITE DIRECTORY USING delta is not a Delta path-table write (Spark requires a FileFormat; DeltaDataSource is not one).

## Test plan

- [x] PathCredentialReadWriteTest#testWriteAndReadBareDeltaPath — local CREATE+read; cloud CREATE TABLE AS SELECT, INSERT OVERWRITE, and SELECT on delta.`s3://...`
- [x] PathCredentialReadWriteTest#testRepeatedApplicationOnBareDeltaPathVendsOnce — single vend and a fixed point with the credential cache disabled
- [x] Full PathCredentialReadWriteTest suite (24/24)

### DIFF
--- a/connectors/spark/src/main/java/io/unitycatalog/spark/utils/OptionsUtil.java
+++ b/connectors/spark/src/main/java/io/unitycatalog/spark/utils/OptionsUtil.java
@@ -16,4 +16,7 @@ public class OptionsUtil {
   public static final boolean DEFAULT_SERVER_SIDE_PLANNING_ENABLED = false;
   public static final String DELTA_API_ENABLED = "deltaRestApi.enabled";
   public static final boolean DEFAULT_DELTA_API_ENABLED = true;
+
+  public static final String VEND_PATH_CREDENTIALS_ENABLED = "vendPathCredentials.enabled";
+  public static final boolean DEFAULT_VEND_PATH_CREDENTIALS_ENABLED = true;
 }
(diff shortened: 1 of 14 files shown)


yili-db:
I think Delta path table should be supported via Delta code change. Delta can (and already do) talk to UC directly via UC-Client and this is the direction we are going towards.

s-maling-celonis:
> I think Delta path table should be supported via Delta code change. Delta can (and already do) talk to UC directly via UC-Client and this is the direction we are going towards.

Alright, then i will close this PR

PR #1756, opened 2026-08-05
[Spark] Make plain-view dependencies null-compatible and harden view round-trip

yili-db:
**PR Checklist**

**Description of changes**

- Connector (Spark 4.2): `UCViewDependencies.derive` now returns `Option`; on a failed derivation `createView` leaves `view_dependencies` unset (null) instead of sending an empty list, so a lineage-validating server can derive them itself.
- Server: `TableRepository` accepts null/absent `view_dependencies` for a plain view (metric views still require at least one entry).
- Connector (Spark 4.0/4.1): `buildV1ViewTable` prefers the persisted `view.query.out.*` / `view.catalogAndNamespace.*` over regenerating them, so a view created on 4.2 reads back correctly.
- Moved the view-property (de)serialization helpers into `UCViewProperties`, leaving `UCViewTypes` as pure type mapping.
- Tests: broadened the shared view integration fixture (cross-catalog join; int/string/array/struct columns; declared-vs-query-output names; creation-context round-trip) and always bind `spark.sql.catalog.spark_catalog` to a Delta-capable catalog so Delta operations work in every test session.

### DIFF
--- a/connectors/spark/src/main/scala-shims/spark-4.0-4.1/io/unitycatalog/spark/UCProxyViewSupport.scala
+++ b/connectors/spark/src/main/scala-shims/spark-4.0-4.1/io/unitycatalog/spark/UCProxyViewSupport.scala
@@ -36,20 +36,27 @@ trait UCProxyViewSupport { self: UCProxy =>
     // Spark 4.2 surfaces these through the View API (withQueryColumnNames / withSchemaMode); on v1
     // they are read from properties (viewQueryColumnNames / viewSchemaModeFromProperties), so
     // populate them here for parity. The `view.sqlConfig.*` keys are already carried in `base`.
+    // Query-output names can differ from declared names (e.g. `CREATE VIEW v(a,b) AS SELECT x,y`),
+    // and Spark matches by them; prefer the persisted `view.query.out.*` in `base`, synthesizing
+    // from declared columns only when absent (deriving would risk INCOMPATIBLE_VIEW_SCHEMA_CHANGE).
     val queryOut =
-      if (fields.nonEmpty) {
+      if (base.contains(CatalogTable.VIEW_QUERY_OUTPUT_NUM_COLUMNS) || fields.isEmpty) {
+        Map.empty[String, String]
+      } else {
         Map(CatalogTable.VIEW_QUERY_OUTPUT_NUM_COLUMNS -> fields.length.toString) ++
           fields.zipWithIndex.map { case (f, i) =>
             s"${CatalogTable.VIEW_QUERY_OUTPUT_COLUMN_NAME_PREFIX}$i" -> f.name
           }
-      } else {
-        Map.empty[String, String]
       }
     val schemaModeDefault =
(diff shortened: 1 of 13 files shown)


connectors/spark/src/main/scala/io/unitycatalog/spark/UCViewProperties.scala
```diff
@@ -0,0 +1,113 @@
+package io.unitycatalog.spark
```
yili-db, line 1: Moved here from `UCViewTypes.scala`

connectors/spark/src/main/scala-shims/spark-4.0-4.1/io/unitycatalog/spark/UCProxyViewSupport.scala
```diff
@@ -36,20 +36,27 @@ trait UCProxyViewSupport { self: UCProxy =>
     // Spark 4.2 surfaces these through the View API (withQueryColumnNames / withSchemaMode); on v1
     // they are read from properties (viewQueryColumnNames / viewSchemaModeFromProperties), so
     // populate them here for parity. The `view.sqlConfig.*` keys are already carried in `base`.
+    // Query-output names can differ from declared names (e.g. `CREATE VIEW v(a,b) AS SELECT x,y`),
+    // and Spark matches by them; prefer the persisted `view.query.out.*` in `base`, synthesizing
+    // from declared columns only when absent (deriving would risk INCOMPATIBLE_VIEW_SCHEMA_CHANGE).
     val queryOut =
-      if (fields.nonEmpty) {
+      if (base.contains(CatalogTable.VIEW_QUERY_OUTPUT_NUM_COLUMNS) || fields.isEmpty) {
```
openinx, line 43: Hi @chenwang-databricks ,  can you help to take a look at this PR ?

openinx: is this a bug before this PR ?  for my understanding,  if the `base` has the `VIEW_QUERY_OUTPUT_NUM_COLUMNS`  but fields is not empty,  then we will still concat to the `properties` in line 67 ?  

but the map should automatically dedup it,  so why do we still need to add this ?

yili-db: The reader should take the `base` as source of truth. The re-generated `VIEW_QUERY_OUTPUT_NUM_COLUMNS` may be off.
The dedup later may not perfectly remove the re-generated items and favor the original ones.  And even if it can perfectly dedup, this is wasted computation work here.
btw I think this is VIEW specific, not impacting metric view.

yili-db: > if the base has the VIEW_QUERY_OUTPUT_NUM_COLUMNS

That means base also has the rest of the keys too. But checking for this leading key should be sufficient.

chenwang-databricks (approved):
LGTM

openinx (approved):

PR #1766, opened 2026-08-10
[Server] Refresh Casbin policy across server instances

s-maling-celonis:
## Motivation / Problem

Unity Catalog's `JCasbinAuthorizer` uses jCasbin with a JDBC adapter: every grant/revoke is **persisted** to the shared `casbin_rule` table, but authorization decisions are made against an **in-memory** policy loaded into the enforcer.

Today that in-memory policy is loaded **once at process start**. There is no watcher, no pub/sub, and no periodic reload across instances. `SyncedEnforcer` only makes a single process safe under concurrent `enforce()` / policy mutations on *that* JVM — it does not synchronize memory across processes.

So when multiple UC server replicas share one database:

- A grant/revoke written through replica A is visible in the DB immediately.
- Replica B keeps serving from its stale in-memory copy until restart.
- Result: spurious `403 Permission Denied` after a grant, and — worse — continued access after a revocation.

This is unsafe for any multi-replica deployment (HA, rolling updates, horizontal scale).

### Concrete example

1. Deploy two UC servers (`uc-1`, `uc-2`) against the same Postgres/MySQL database, behind a load balancer.
2. An admin calls `GRANT SELECT ON catalog.schema.table TO user@example.com` and the request lands on **`uc-1`**.
   - `uc-1` writes the policy row to `casbin_rule` and updates its local enforcer.
3. Moments later, `user@example.com` reads that table; the request lands on **`uc-2`**.
   - `uc-2`'s enforcer still lacks the grant → **403**, even though the grant is already in the DB.
4. The inverse is worse: revoke on `uc-1`, then a request on `uc-2` can still **succeed** until `uc-2` restarts — stale allow.

A related path is create-then-read across instances (e.g. create a securable / grant ownership on one replica, immediately access it via another): the deny path looks like a false negative until policy is reloaded.

### Why jCasbin / the current authorizer does not handle this

| What exists today | What it covers | What it does *not* cover |
|---|---|---|
| `JDBCAdapter` + `enableAutoSave(true)` | Persist local grant/revoke to `casbin_rule` | Notify other processes that policy changed |
| `SyncedEnforcer` | Thread-safe enforce/mutate **within one JVM** | Cross-process / cross-replica consistency |
| Startup `SyncedEnforcer(model, adapter)` | Initial `loadPolicy()` from DB | Subsequent reloads when another instance writes |

jCasbin does not ship a built-in multi-node policy invalidation for the JDBC adapter. Without an external refresh mechanism, each UC replica is effectively a permanently stale cache of authorization state.

## Solution

This PR adds `CasbinPolicyRefresher`, which:

1. **Polls** `casbin_rule` with `select count(*), coalesce(max(id), 0)` and reloads when either value changes (inserts/deletes with auto-increment ids change at least one). Probe, reload, and version stamp run under one lock with a `checkSeq` so concurrent callers coalesce; waiters skip after a consistent check. The stamped version is the one **probed before** reload, so the cache matches what was loaded (not a fresher post-reload re-read).
2. On a **denied** evaluation in `UnityAccessEvaluator.evaluate`, triggers one **debounced** `checkAndReload` (same version-aware path as the poller) and re-evaluates before returning 403. `refreshAuthorizations` returns true only if the current enforcer instance changed, including when a coalesced concurrent check performed the swap.

Reload does **not** call `loadPolicy()` on the current enforcer. It builds a fresh enforcer from the store and swaps it in under the exclusive side of a `ReadWriteLock`:

- `authorize` / `enforce` keep using the previous instance (no lock).
- Local grants/revokes take the shared side so they cannot land on the outgoing instance, but do not wait for each other (`SyncedEnforcer` still serializes mutations on one instance). The read lock is on the current-enforcer pointer, not Casbin read vs write.
- Reload takes the exclusive side for the whole `newEnforcer()` so the swap is not a stop-the-world for reads.

Refresh is **off by default**. Single-instance deployments are unchanged; multi-instance shared-DB setups must set `server.authorization.policy-refresh=true`. See [docs/server/auth.md](https://github.com/unitycatalog/unitycatalog/blob/fix/casbin-cross-replica-policy-refresh/docs/server/auth.md#multiple-server-instances).

## Alternatives considered

More efficient push-based options exist; we deliberately did not take them:

- **Redis (or similar) pub/sub** — lower latency invalidation, but would add another third-party runtime dependency and operational surface for every multi-replica deployment.
- **Native DB notify** (e.g. Postgres `LISTEN`/`NOTIFY`) — event-driven and cheap, but ties the server to Postgres-specific APIs and would introduce compile-time / dialect-specific database dependencies. UC supports more than one DB backend via JDBC/Hibernate, so policy refresh should stay portable.

Polling the shared `casbin_rule` table keeps the solution DB-agnostic and dependency-free, at the cost of up-to-interval propagation delay (tunable) and a cheap version query per interval.

## Configuration

| Property | Default | Purpose |
|---|---|---|
| `server.authorization.policy-refresh` | `false` | Kill switch (`true`/`false`). Off disables poll and deny-path reload. |
| `server.authorization.policy-refresh-interval` | `PT1M` | How often to poll `casbin_rule` for changes |
| `server.authorization.policy-refresh-debounce-interval` | `PT1S` | Min gap between deny-triggered checks (avoids stampede on legitimate 403s) |

No Helm change needed — defaults apply via existing `extraProperties`. Set `server.authorization.policy-refresh=true` for multi-instance shared-DB deployments. Documented in `docs/server/auth.md`.

## Test plan

- [x] `JCasbinAuthorizerMultiInstanceTest` — grant/revoke/hierarchy propagation, change detection, debounce (including after the window), kill switch, `close()`, authorize stays live during exclusive reload lock
- [x] `JCasbinAuthorizerTest` — existing grant/revoke/hierarchy/list behavior
- [x] `UnityAccessEvaluatorTest` — deny-path refresh + retry in `evaluate`
- [ ] `server/test` passes in CI

## Limitations

- Refresh is opt-in. Multi-instance deployments must enable it.
- Revocations propagate asynchronously (up to poll interval); deny-path refresh only helps stale *denies*.
- A write that lands during `newEnforcer()` may wait until the next check (version stamp is the pre-reload probe).
- `count(*)` cost grows with policy size; raise interval for very large policy sets.

### DIFF
--- a/etc/conf/server.properties
+++ b/etc/conf/server.properties
@@ -34,6 +34,29 @@ server.audiences=
 server.cookie-timeout=P5D
 server.access-token-timeout=PT24H
 
+# Authorization policy refresh.
+# Each server instance keeps the Casbin policy set in memory, so an instance must poll the shared
+# database to see grants and revocations made through another instance. Leave this enabled whenever
+# more than one instance runs against one database: with it disabled, a revocation made through one
+# instance is never honoured by the others, which keep allowing access with no error.
+# Disabling this turns off both the periodic poll and the refresh on a denied request, restoring the
+# behaviour from before either existed. It is a full kill switch, not just a way to stop polling.
+# Default: enable
+# server.authorization.policy-refresh=enable
+#
+# How often to check for changes. The check is a single aggregate over the casbin_rule table, whose
+# cost grows with the number of policy rows; deployments with very large policy sets may want to
+# raise this. Revocations take effect within roughly this interval.
+# Default: PT1S
+# server.authorization.policy-refresh-interval=PT1S
+#
+# Shortest interval between two refreshes triggered by a denied request. A denied request refreshes
+# once and retries before returning 403, so a privilege granted through another instance moments
+# earlier is honoured immediately rather than after the next poll. This bounds the work an
+# unauthorized caller can force by repeatedly hitting a denied endpoint.
+# Default: PT1S
+# server.authorization.policy-refresh-debounce=PT1S
+
 # Enable MANAGED table
 # Default: true (enabled)
 # server.managed-table.enabled=true
--- a/etc/conf/server.properties
+++ b/etc/conf/server.properties
@@ -34,27 +34,9 @@ server.audiences=
 server.cookie-timeout=P5D
 server.access-token-timeout=PT24H
 
-# Authorization policy refresh.
-# Each server instance keeps the Casbin policy set in memory, so an instance must poll the shared
-# database to see grants and revocations made through another instance. Leave this enabled whenever
-# more than one instance runs against one database: with it disabled, a revocation made through one
-# instance is never honoured by the others, which keep allowing access with no error.
-# Disabling this turns off both the periodic poll and the refresh on a denied request, restoring the
-# behaviour from before either existed. It is a full kill switch, not just a way to stop polling.
-# Default: enable
+# Authorization policy refresh (required for multiple instances on one database).
 # server.authorization.policy-refresh=enable
-#
-# How often to check for changes. The check is a single aggregate over the casbin_rule table, whose
-# cost grows with the number of policy rows; deployments with very large policy sets may want to
-# raise this. Revocations take effect within roughly this interval.
-# Default: PT1S
 # server.authorization.policy-refresh-interval=PT1S
-#
-# Shortest interval between two refreshes triggered by a denied request. A denied request refreshes
-# once and retries before returning 403, so a privilege granted through another instance moments
-# earlier is honoured immediately rather than after the next poll. This bounds the work an
-# unauthorized caller can force by repeatedly hitting a denied endpoint.
-# Default: PT1S
 # server.authorization.policy-refresh-debounce=PT1S
 
 # Enable MANAGED table
--- a/etc/conf/server.properties
+++ b/etc/conf/server.properties
@@ -35,19 +35,20 @@ server.cookie-timeout=P5D
 server.access-token-timeout=PT24H
 
 # Authorization policy refresh across server instances that share one database.
-# Each instance keeps Casbin policy in memory; without refresh, grants/revokes made
-# through another instance are invisible until restart.
+# Required for any multi-instance / HA deployment: each instance keeps Casbin
+# policy in memory, and without refresh, grants/revokes made through another
+# instance are invisible until restart. See docs/server/auth.md.
 #
 # server.authorization.policy-refresh=false
-#   Kill switch. "false" (default) turns off background polling and deny-path reload
-#   (single-instance deployments). Set to "true" for multi-instance shared-DB setups.
+#   Kill switch. "false" (default) is for single-instance deployments. Set to
+#   "true" whenever more than one UC server process shares this database.
 #
 # server.authorization.policy-refresh-interval=PT1M
 #   How often this instance polls casbin_rule for changes and reloads if needed.
 #   Lower this for faster revoke propagation across replicas; raise it for very large
 #   policy tables to reduce count(*)/max(id) load.
 #
-# server.authorization.policy-refresh-debounce=PT1S
+# server.authorization.policy-refresh-debounce-interval=PT1S
 #   Minimum time between deny-triggered reloads. On 403, we may reload once and
 #   re-check (covers stale denies across instances); debounce avoids a reload stampede
 #   under many legitimate denials.
--- a/etc/conf/server.properties
+++ b/etc/conf/server.properties
@@ -38,9 +38,9 @@ server.access-token-timeout=PT24H
 # Each instance keeps Casbin policy in memory; without refresh, grants/revokes made
 # through another instance are invisible until restart.
 #
-# server.authorization.policy-refresh=disable
-#   Kill switch. "disable" (default) turns off background polling and deny-path reload
-#   (single-instance deployments). Set to "enable" for multi-instance shared-DB setups.
+# server.authorization.policy-refresh=false
+#   Kill switch. "false" (default) turns off background polling and deny-path reload
+#   (single-instance deployments). Set to "true" for multi-instance shared-DB setups.
 #
 # server.authorization.policy-refresh-interval=PT1M
 #   How often this instance polls casbin_rule for changes and reloads if needed.
--- a/etc/conf/server.properties
+++ b/etc/conf/server.properties
@@ -34,10 +34,22 @@ server.audiences=
 server.cookie-timeout=P5D
 server.access-token-timeout=PT24H
 
-# Authorization policy refresh (required for multiple instances on one database).
+# Authorization policy refresh across server instances that share one database.
+# Each instance keeps Casbin policy in memory; without refresh, grants/revokes made
+# through another instance are invisible until restart.
+#
 # server.authorization.policy-refresh=enable
+#   Kill switch. "enable" (default) turns on background polling and deny-path reload.
+#   Set to "disable" to turn both off (single-instance only; multi-instance is unsafe).
+#
 # server.authorization.policy-refresh-interval=PT1S
+#   How often this instance polls casbin_rule for changes and reloads if needed.
+#   Raise this for very large policy tables to reduce count(*)/max(id) load.
+#
 # server.authorization.policy-refresh-debounce=PT1S
+#   Minimum time between deny-triggered reloads. On 403, we may reload once and
+#   re-check (covers stale denies across instances); debounce avoids a reload stampede
+#   under many legitimate denials.
 
 # Enable MANAGED table
 # Default: true (enabled)
--- a/etc/conf/server.properties
+++ b/etc/conf/server.properties
@@ -38,13 +38,14 @@ server.access-token-timeout=PT24H
 # Each instance keeps Casbin policy in memory; without refresh, grants/revokes made
 # through another instance are invisible until restart.
 #
-# server.authorization.policy-refresh=enable
-#   Kill switch. "enable" (default) turns on background polling and deny-path reload.
-#   Set to "disable" to turn both off (single-instance only; multi-instance is unsafe).
+# server.authorization.policy-refresh=disable
+#   Kill switch. "disable" (default) turns off background polling and deny-path reload
+#   (single-instance deployments). Set to "enable" for multi-instance shared-DB setups.
 #
-# server.authorization.policy-refresh-interval=PT1S
+# server.authorization.policy-refresh-interval=PT1M
 #   How often this instance polls casbin_rule for changes and reloads if needed.
-#   Raise this for very large policy tables to reduce count(*)/max(id) load.
+#   Lower this for faster revoke propagation across replicas; raise it for very large
+#   policy tables to reduce count(*)/max(id) load.
 #
 # server.authorization.policy-refresh-debounce=PT1S
 #   Minimum time between deny-triggered reloads. On 403, we may reload once and
--- a/server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
@@ -1,0 +1,206 @@
+package io.unitycatalog.server.auth;
+
+import java.time.Duration;
+import java.util.Optional;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.casbin.jcasbin.main.SyncedEnforcer;
+import org.hibernate.SessionFactory;
+import org.hibernate.query.NativeQuery;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Keeps one server instance's in-memory Casbin policy set in step with the shared database.
+ *
+ * <p>jCasbin evaluates every decision against an in-memory copy of the policy set that {@link
+ * org.casbin.jcasbin.main.CoreEnforcer} loads once, in its constructor. Writes go to the writing
+ * instance's memory and to {@code casbin_rule}, but no other instance ever learns about them. With
+ * more than one replica and no session affinity that means a grant made through one replica is
+ * invisible to the others, and — more seriously — a revocation made through one replica is not
+ * honoured by the others, which keep allowing access and vending credentials with no error and
+ * nothing in the log. This class closes that gap by reloading when the table changes.
+ *
+ * <h2>Change detection</h2>
+ *
+ * <p>Detection is {@code select count(*), coalesce(max(id), 0) from casbin_rule}: when either value
+ * differs from the previously observed pair, the policy set is reloaded. That pair cannot miss a
+ * change, given two properties of the casbin JDBC adapter:
+ *
+ * <ul>
+ *   <li>{@code casbin_rule.id} is an auto-incrementing primary key in every dialect the adapter
+ *       supports, so an insert always moves {@code max(id)}.
+ *   <li>The adapter only ever inserts and deletes — it issues no {@code UPDATE} against the table,
+ *       and even {@code updatePolicy} is a delete followed by an insert.
+ * </ul>
+ *
+ * <p>So an insert moves {@code max(id)}, a delete moves {@code count(*)}, and a delete paired with
+ * an insert leaves the count unchanged but still moves {@code max(id)}. No in-place update exists
+ * that could alter a row's contents without moving one of the two. The comparison is for inequality
+ * rather than for growth, so even a sequence reset triggers a reload.
+ *
+ * <p>Deliberately plain standard SQL: it works on every backend the adapter supports and needs no
+ * schema change, no extra table, and no database-specific notification mechanism.
+ *
+ * <h2>Concurrency</h2>
+ *
+ * <p>Reloading is safe to do while requests are being served, but only because the enforcer is a
+ * {@link SyncedEnforcer}: it overrides {@code loadPolicy()} to hold a write lock for the whole
+ * clear-then-repopulate cycle, while {@code enforce()} holds the matching read lock. A plain {@code
+ * Enforcer} would let concurrent {@code enforce()} calls observe the momentarily empty model that
+ * {@code loadPolicy()} passes through, denying valid requests.
+ */
+public class CasbinPolicyRefresher implements AutoCloseable {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(CasbinPolicyRefresher.class);
+
+  /**
+   * Reuses Hibernate's connection pool rather than opening a second one. {@code casbin_rule} is not
+   * a Hibernate-managed entity, so this has to be a native query.
+   */
+  private static final String VERSION_QUERY =
+      "select count(*), coalesce(max(id), 0) from casbin_rule";
+
+  private final SyncedEnforcer enforcer;
+  private final SessionFactory sessionFactory;
+
+  /**
+   * Last observed {@code (count, maxId)}. Guarded by {@code this} along with every reload, so the
+   * scheduled poll and a deny-triggered refresh can never interleave.
+   *
+   * <p>Deliberately left at a value the table can never report, so the first check always reloads.
+   * Seeding this from the table in the constructor would be a race: the enforcer loads the policy
+   * set before the refresher is built, so a write landing between the two would already be included
+   * in the seeded version while never having reached the enforcer — and would then never be picked
+   * up. One redundant reload at startup is a cheap price for closing that.
+   */
+  private long lastCount = -1;
+
+  private long lastMaxId = -1;
+
+  private ScheduledExecutorService executor;
+
+  public CasbinPolicyRefresher(SyncedEnforcer enforcer, SessionFactory sessionFactory) {
+    this.enforcer = enforcer;
+    this.sessionFactory = sessionFactory;
+  }
+
+  /**
+   * Starts polling for changes made by other instances.
+   *
+   * @param interval delay between the end of one check and the start of the next
+   */
+  public synchronized void start(Duration interval) {
+    if (executor != null) {
+      return;
+    }
+    executor =
+        Executors.newSingleThreadScheduledExecutor(
+            runnable -> {
+              Thread thread = new Thread(runnable, "casbin-policy-refresher");
+              // Daemon so a forgotten refresher can never hold the JVM open.
+              thread.setDaemon(true);
+              return thread;
+            });
+    long millis = Math.max(1, interval.toMillis());
+    // Fixed delay rather than fixed rate: a reload slower than the interval must not cause runs to
+    // queue up back to back.
+    executor.scheduleWithFixedDelay(this::pollQuietly, millis, millis, TimeUnit.MILLISECONDS);
+    LOGGER.info("Casbin policy refresh enabled, checking every {}ms", millis);
+  }
+
+  /** Never lets a failure escape into the executor, which would silently cancel the schedule. */
+  private void pollQuietly() {
+    try {
+      checkAndReload();
+    } catch (Throwable t) {
+      LOGGER.warn("Casbin policy refresh check failed; will retry on the next interval", t);
+    }
+  }
+
+  /**
+   * Reloads the policy set if the table has changed since the last observation.
+   *
+   * @return true if a reload happened
+   */
+  public synchronized boolean checkAndReload() {
+    long[] version = readVersion().orElse(null);
+    if (version == null) {
+      // Could not read the table. Leave the last observed values alone so the next check compares
+      // against a real observation rather than treating the failure as a change.
+      return false;
+    }
+    if (version[0] == lastCount && version[1] == lastMaxId) {
+      return false;
+    }
+    LOGGER.debug(
+        "Casbin policy changed (count {} -> {}, maxId {} -> {}); reloading",
+        lastCount,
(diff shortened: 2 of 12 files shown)


openinx:
Hi @s-maling-celonis ,  can you explain more about the context ?  and what's the problem that you want to solve ?  Thanks for the contribution.

etc/conf/server.properties
```diff
@@ -34,6 +34,11 @@ server.audiences=
 server.cookie-timeout=P5D
 server.access-token-timeout=PT24H
 
+# Authorization policy refresh (required for multiple instances on one database).
```
openinx: I suggest to add a more detailed comment for those three keys, seems like it's not easy to get what're the meaning of the latter two.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
```diff
@@ -252,7 +255,11 @@ public void beforeRequest(
         Map<SecurableType, UUID> resourceIds,
         Map<String, Object> nonResourceValues) {
       if (!evaluator.evaluate(principal, expression, resourceIds, nonResourceValues)) {
-        throw new BaseException(ErrorCode.PERMISSION_DENIED, "Access denied.");
+        if (!authorizer.refreshAuthorizations()
```
openinx: Interesting,  so this line is the key change, right ?   If we are going to deny the request, then we'd refresh the cache if possible, and then re-evaluate it again to see if it's a false positive check.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.casbin.jcasbin.main.SyncedEnforcer;
+import org.hibernate.SessionFactory;
+import org.hibernate.query.NativeQuery;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Reloads the in-memory Casbin policy when {@code casbin_rule} changes in the shared database.
```
openinx, line 15: What is the `casbin_rule` ? can you give me more context about this ?

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
```diff
@@ -0,0 +1,136 @@
...
+ * <p>Change detection uses {@code select count(*), coalesce(max(id), 0) from casbin_rule}. The JDBC
+ * adapter only inserts and deletes rows with auto-incrementing ids, so any write changes at least
+ * one of the two values. Requires a {@link SyncedEnforcer} so {@code loadPolicy()} is safe under
+ * concurrent {@code enforce()} calls.
+ */
+public class CasbinPolicyRefresher implements AutoCloseable {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(CasbinPolicyRefresher.class);
+
+  private static final String VERSION_QUERY =
```
openinx, line 31: How is the cost of this query in a large table ?  will it be the bottleneck ?

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
```diff
@@ -0,0 +1,136 @@
...
+    try {
+      checkAndReload();
+    } catch (Throwable t) {
+      LOGGER.warn("Casbin policy refresh check failed; will retry on the next interval", t);
+    }
+  }
+
+  /** @return true if a reload happened */
+  public synchronized boolean checkAndReload() {
+    long[] version = readVersion().orElse(null);
```
openinx: Since this `readVersion` is in the scope of `synchronized`,  I'd be very cautious about the latency of `readVersion()`. please check if it will be bottleneck, otherwise that will cause a huge contention.

openinx:
Hi @yili-db , do you want to take a look a this PR ?

s-maling-celonis:
> Hi @s-maling-celonis , can you explain more about the context ? and what's the problem that you want to solve ? Thanks for the contribution.

Thanks for taking a look — I updated the PR description with more detail. Let me know if you have any more questions.

etc/conf/server.properties
s-maling-celonis: Good call — I've expanded the comments for all three keys so interval vs debounce is clearer. Thanks!

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
s-maling-celonis: Yes, exactly. On a deny we try one debounced reload (in case another instance just wrote a grant), then re-evaluate. If it still fails — or refresh is disabled / debounced — we return 403 as before. Note this only helps stale *denies*; stale *allows* after a revoke still wait for the background poll (called out in the Limitations section).

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
s-maling-celonis: `casbin_rule` is the table owned by jCasbin's JDBC adapter — it stores the persisted `p`/`g` policy rows (grants and hierarchy). UC's authorizer writes there on grant/revoke, but each process only loads it into memory at startup; this class detects changes to that table and calls `loadPolicy()`. I added a short note in the class javadoc as well.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
s-maling-celonis: Good question. Revoke deletes rows from `casbin_rule` (JDBC adapter + auto-save), so the table size tracks the *current* policy set (grants + hierarchy), not historical churn. `count(*)` therefore scales with how large the live ACL is, not with how long the cluster has been running; `max(id)` keeps increasing across insert/delete churn but is cheap with the primary key.

At the default 1s poll this should be fine for typical UC deployments. If someone had a very large active policy table, they can raise `policy-refresh-interval`. We kept this simple/portable instead of adding a side version counter.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
s-maling-celonis: Agreed — holding the lock across the DB read was unnecessary. I moved `readVersion()` outside `synchronized` so we only hold the monitor while comparing/updating the cached version and calling `loadPolicy()`. The background poller is single-threaded and deny-path reload is debounced, so contention should stay low; this still removes the DB latency from the critical section.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
yili-db: I suggest moving this refresh and retry logic into `evaluate` itself:
- It happens to have the `authorizer`
- It can retry just the portion of `parser.parseExpression` and reuse everything else
- `evaluate` can be called in other places, like `RequestEvaluationAction`, that this PR missed.

Also please catch exception of `authorizer.refreshAuthorizations` and return the result before retry if it fails.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -191,6 +191,14 @@ public enum Property {
     SERVER_ENV("server.env", "dev", new EnumValidator(true, "dev", "prod", "test")),
     AUTHORIZATION_ENABLED(
         "server.authorization", "disable", new EnumValidator(true, "enable", "disable")),
+    POLICY_REFRESH_ENABLED(
+        "server.authorization.policy-refresh",
+        "enable",
+        new EnumValidator(true, "enable", "disable")),
+    POLICY_REFRESH_INTERVAL(
+        "server.authorization.policy-refresh-interval", "PT1S", DURATION_VALIDATOR),
```
yili-db: This polls the entire Casbin policy list every second by default. I think frequency should be reduced.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -191,6 +191,14 @@ public enum Property {
     SERVER_ENV("server.env", "dev", new EnumValidator(true, "dev", "prod", "test")),
     AUTHORIZATION_ENABLED(
         "server.authorization", "disable", new EnumValidator(true, "enable", "disable")),
+    POLICY_REFRESH_ENABLED(
+        "server.authorization.policy-refresh",
+        "enable",
+        new EnumValidator(true, "enable", "disable")),
```
yili-db: Use `BOOLEAN_VALIDATOR` with `true` and `false`.

`server.authorization` is a bad example that introduced before server properties is formalized.

yili-db: I think this should `false` by default.
This is a production behavior change. I haven't seen any real multi-UC deployment yet and we haven't tested in that mode. I suppose by default it should still be a single instance for now.
If later it becomes clear that it's quite common, changing this from `false` to `true` is no worse than doing it now.
At least for now I would consider this feature highly experimental.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
```diff
@@ -0,0 +1,151 @@
...
+      if (version[0] == lastCount && version[1] == lastMaxId) {
+        return false;
+      }
+      LOGGER.debug(
+          "Casbin policy changed (count {} -> {}, maxId {} -> {}); reloading",
+          lastCount,
+          version[0],
+          lastMaxId,
+          version[1]);
+      enforcer.loadPolicy();
```
yili-db: Even with the `version` filter, this `enforcer.loadPolicy()` still blocks all RPCs coming into the server at the authorization stage when any policy changes. It holds the enforcer's **write lock** across the full `SELECT * FROM casbin_rule` + role-link rebuild, and every `enforce()` takes the **read lock**, so a reload blocks all authorization (i.e. all RPCs) on that instance for its whole duration.
This periodic **stop-the-world** problem is still concerning.

I propose a workaround that still blocks all writes to casbin (creating a table etc) when refreshing, but allows all reads (all other RPCs):
- Build a fresh `enforcer` off-lock and atomically swap an `AtomicReference` that `authorize()` reads, so reads never block.
- Then there's problem with writes: concurrent local writes happening during refresh can be discarded until next refresh. To fix this, all local writes have to be locked against the refresh with one lock (write targets `current.get()` inside the lock) so no local write is lost.
- Optional later: journal-and-replay local writes to avoid blocking writes too.

yili-db (changes requested):

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessDecorator.java
s-maling-celonis: Done — deny-path refresh now lives in `UnityAccessEvaluator.evaluate`. It reuses the same evaluation context and only re-runs the SpEL expression, so `RequestEvaluationAction` and any other `evaluate` caller get the retry too.

If `refreshAuthorizations()` is disabled, debounced, or throws, we keep the original deny and do not retry.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
s-maling-celonis: Agreed. The default interval is now `PT1M`.

One clarification: the poll is only `select count(*), coalesce(max(id), 0) from casbin_rule`. The full policy load runs when that version changes, not on every tick.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
s-maling-celonis: Done — `server.authorization.policy-refresh` now uses `BOOLEAN_VALIDATOR` with `true` / `false`.

s-maling-celonis: Agreed on the default — it's now `false`, so single-instance deployments are unchanged. Multi-instance shared-DB setups need `server.authorization.policy-refresh=true`.

We are running multiple replicas today; that's how we found this. Grants and revokes written on one instance were invisible on the others until restart, and enabling this refresh on those replicas fixed it. I'm still fine treating the feature as opt-in until it's more widely used.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
s-maling-celonis: Agreed that in-place `loadPolicy()` on the live enforcer was a stop-the-world for every `enforce()`.

Reload now builds a fresh enforcer and swaps it into an `AtomicReference`. `authorize` reads that reference and never takes the reload lock, so RPCs stay live during refresh.

Local writes take the shared side of a `ReadWriteLock`; reload takes the exclusive side for the whole `newEnforcer()`. That way a grant cannot land on the outgoing instance (and be dropped on swap). Concurrent grants no longer wait for each other — only for a reload. We did not add journal-and-replay.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
```diff
@@ -0,0 +1,152 @@
...
+      }
+      LOGGER.debug(
+          "Casbin policy changed (count {} -> {}, maxId {} -> {}); reloading",
+          lastCount,
+          version[0],
+          lastMaxId,
+          version[1]);
+    }
+    reloader.run();
+    recordLatestVersion();
```
yili-db: I think the process can be improved:

```
  // Keep an `AtomicLong checkSeq`
  long seenCheck = checkSeq;
  lock;
  try {
    if (checkSeq != seenCheck) return;        // leader already did a consistent check; recorded is current -> skip probe + rebuild
    long[] db = readVersion().orElse(null);
    if (db == null) return;                   // probe failed: advance nothing, others retry
    if (db[0] == lastCount && db[1] == lastMaxId) { checkSeq++; return; }  // no change
    current.set(newEnforcer());               // may throw: advance nothing, others retry
    recordVersion(db[0], db[1]);
    checkSeq++;                               // only after a consistent load
  } finally { unlock; }
```

Note that it's beneficial to include `readVersion` inside the locked section, because concurrent `readVersion` is not useful so we'd better block it until the current update finishes and short circuit the waiter.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
```diff
@@ -0,0 +1,152 @@
...
+    return true;
+  }
+
+  /** Unconditional reload for callers that cannot wait for the next poll. */
+  public void forceReload() {
+    reloader.run();
+    recordLatestVersion();
```
yili-db: Shouldn't `forceReload` also check for version instead of blindly reload?
This function can be replaced by `checkAndReload`.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  private void recordLatestVersion() {
```
yili-db: recording a fresher version, after reloading, is not ideal.

server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
```diff
@@ -73,48 +110,87 @@ static String resolveConnectionUsername(Properties properties) {
...
+    reloadLock.writeLock().lock();
+    try {
+      current.set(newEnforcer());
+    } finally {
+      reloadLock.writeLock().unlock();
+    }
+  }
+
+  private <T> T mutate(Function<SyncedEnforcer, T> action) {
+    reloadLock.readLock().lock();
```
yili-db, line 143: This is correct, but to avoid future confusion it's worth commenting on "why mutate, a write operation, should have a readLock"

server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
```diff
@@ -154,11 +232,38 @@ public List<Privileges> listAuthorizations(UUID principal, UUID resource) {
...
+    }
+    long now = System.nanoTime();
+    long last = lastRefreshNanos.get();
+    if (now - last < refreshDebounceNanos) {
+      return false;
+    }
+    if (!lastRefreshNanos.compareAndSet(last, now)) {
+      return false;
+    }
+    refresher.forceReload();
```
yili-db: With https://github.com/unitycatalog/unitycatalog/pull/1766/changes#r3787724062, not all `forceReload` would produce a fresh enforcer. Here you can save `live()` locally before this line, then compare `live()` again and return false if it hasn't changed.

yili-db (commented):
PR is much better shape. The stop-the-world problem is mitigated. A few suggestion to further optimize the flow.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
s-maling-celonis: Agreed — we no longer re-read after reload. `checkAndReload` stamps the version probed before `reloader.run()`, so the cache matches what was loaded (`8ed54846` / clarified in `86bbbe74`).

server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
s-maling-celonis: Done — added a comment on `mutate` explaining that the read lock is on the live-enforcer pointer (many grants may share it; reload takes the write side), not Casbin read vs write (`86bbbe74`).

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
s-maling-celonis: Done in `8ed54846` — `checkAndReload` uses `checkSeq`, probes under the lock, stamps the probed version, and waiters skip after a consistent check.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
s-maling-celonis: Done in `dc910e5d` — `forceReload` is gone; deny-path goes through `checkAndReload`.

server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
s-maling-celonis: Done in `dc910e5d` — `refreshAuthorizations` snapshots `live()` before `checkAndReload` and returns `true` only if the enforcer instance changed (covers coalesced waiters).

yili-db:
Just made a diagram to show the locking mechanism
[casbin-locking-diagram.html](https://github.com/user-attachments/files/31152846/casbin-locking-diagram.html)

yili-db (approved):

etc/conf/server.properties
```diff
@@ -34,6 +34,24 @@ server.audiences=
 server.cookie-timeout=P5D
 server.access-token-timeout=PT24H
 
+# Authorization policy refresh across server instances that share one database.
+# Each instance keeps Casbin policy in memory; without refresh, grants/revokes made
+# through another instance are invisible until restart.
+#
+# server.authorization.policy-refresh=false
```
tdas, line 42: its not obvious that any one setting up multi-server set up needs to enable this policy refresh. So i think there are 2 approaches
1. always have this refresh enabled - is there any downside to having a policy refresh always enabled? if not, then no one in a multi-server scenario has to change anything. 

2. else, we document this in the docs (can be a follow up PR)

which one do you want to go with? 

In addition, we can also consider having a meta server property called `server.multi-server.enabled=true (default = false)` that is easier to understand+document and automaticaly enables other stuff like this policy-refresh and others in the future.

server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
```diff
@@ -73,48 +110,95 @@ static String resolveConnectionUsername(Properties properties) {
     return username;
   }
 
+  private SyncedEnforcer live() {
```
tdas: live() is a confusing name. live what? also..not directly use current (and rename to currentEnforcer)? why both live() and current exist?

server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
```diff
@@ -154,11 +240,43 @@ public List<Privileges> listAuthorizations(UUID principal, UUID resource) {
...
                     l -> Privileges.fromValue(l.get(PRIVILEGE_INDEX)), Collectors.toList())));
   }
+
+  /**
+   * Rate-limited policy check before returning 403, for cross-instance create-then-read.
+   *
+   * @return true if the live enforcer was replaced (by this call or a coalesced concurrent check)
+   */
+  @Override
+  public boolean refreshAuthorizations() {
```
tdas, line 258: shouldnt we log stuff here so that its easier to track in production when things do get updated?

etc/conf/server.properties
```diff
@@ -34,6 +34,24 @@ server.audiences=
...
+#   Kill switch. "false" (default) turns off background polling and deny-path reload
+#   (single-instance deployments). Set to "true" for multi-instance shared-DB setups.
+#
+# server.authorization.policy-refresh-interval=PT1M
+#   How often this instance polls casbin_rule for changes and reloads if needed.
+#   Lower this for faster revoke propagation across replicas; raise it for very large
+#   policy tables to reduce count(*)/max(id) load.
+#
+# server.authorization.policy-refresh-debounce=PT1S
```
tdas: shouldnt this be named as a `-interval` or `-timeout` as it take a time interval?

server/src/test/java/io/unitycatalog/server/auth/JCasbinAuthorizerMultiInstanceTest.java
```diff
@@ -0,0 +1,320 @@
...
+    assertThat(reader.authorize(principal, granted, Privileges.SELECT)).isTrue();
+    assertThat(reader.authorize(principal, revoked, Privileges.SELECT)).isFalse();
+  }
+
+  @Test
+  void denyTriggeredRefreshIsDebounced() throws Exception {
+    JCasbinAuthorizer replica = startReplica();
+
+    replica.refreshAuthorizations();
+    assertThat(replica.refreshAuthorizations()).isFalse();
```
tdas: how are you testing the debounce interval/timeout? that after the timeout refresh is allowed?

tdas (changes requested):
left a bunch of suggstions.

server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
s-maling-celonis: Agreed — `live()` is gone. The field is now `currentEnforcer`; callers use `currentEnforcer.get()` directly.

etc/conf/server.properties
s-maling-celonis: Went with option 2: keep the default `false` and document the opt-in.

yili-db [asked for that default](https://github.com/unitycatalog/unitycatalog/pull/1766#discussion_r3781540093) because this is still experimental: multi-instance shared-DB hasn’t been the assumed production mode, and turning refresh on is a behavior change we haven’t broadly tested. Flipping `false` → `true` later is cheaper than surprising single-instance deployments now.

So: default stays off (single-instance unchanged). Multi-instance shared-DB setups need `server.authorization.policy-refresh=true`. That’s now in a “Multiple server instances” section in `docs/server/auth.md`, and the `server.properties` comments call it out.

server/src/test/java/io/unitycatalog/server/auth/JCasbinAuthorizerMultiInstanceTest.java
s-maling-celonis: The test now covers both sides: a second call inside the window is skipped, and after the window a grant on another replica makes `refreshAuthorizations()` return `true`. The poller is set to `PT1H` in that test so it can’t steal the reload.

etc/conf/server.properties
s-maling-celonis: Renamed to `server.authorization.policy-refresh-debounce-interval` so it’s clearly a duration (the poller already owns `…-interval`).

server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
s-maling-celonis: Done. After a successful swap, `CasbinPolicyRefresher` logs at info with the old/new `count` and `maxId` (covers both the poller and the deny path). Failures stay at warn; no-ops and debounce skips stay quiet.

etc/conf/server.properties
tdas: i am good with that.

tdas (approved):
LGTM. thank you for addressign the comments.

PR #1772, opened 2026-08-11
Render errors in the dialect of the service that was called

yili-db:
**PR Checklist**

**Description of changes**

Armeria's `@ExceptionHandler` only wraps the annotated-service invocation, but authorization runs in route decorators outside it. A denial thrown there was always formatted with the UC envelope, so a 403 on a Delta or Iceberg route came back in a shape those clients do not expect.

- Each service names its dialect via `RegisteredService.exceptionHandler()`, replacing 21 per-service `@ExceptionHandler` annotations. `register` feeds that one value to both Armeria and the decorator, so they cannot disagree.
- `ExceptionHandlingDecorator` splits into `ServiceExceptionHandlingDecorator` (per service, records the dialect) and `GlobalExceptionHandlingDecorator` (outermost, the only one that can catch a route-decorator denial; reads the dialect off the route's chain).
- `RegisteredService` is an interface, so a service's protocol is independent of its base class. `annotate` is overloaded per service type, so the argument picks the protocol.
- `assertPermissionDenied` splits into UC and Delta variants that parse the body and reject the other dialect. This caught eight call sites asserting the UC shape against Delta endpoints.
- Also corrects the `ErrorInfo` detail's `@type` to the full protobuf Any type URL `type.googleapis.com/google.rpc.ErrorInfo` (the previous bare `google.rpc.ErrorInfo` was not a valid type URL)

Note: re-adding an `@ExceptionHandler` annotation will now fail at startup, since the handlers are singletons with private constructors. The dialect belongs in `exceptionHandler()`.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/exception/ExceptionHandlingDecorator.java
+++ b/server/src/main/java/io/unitycatalog/server/exception/ExceptionHandlingDecorator.java
@@ -32,8 +32,14 @@ public abstract class ExceptionHandlingDecorator extends SimpleDecoratingHttpSer
 
   @Override
   public final HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
+    // An exception can surface two ways: thrown while serve() runs (the catch), or delivered later
+    // as an error signal on the returned response when the service completes asynchronously (the
+    // recover). recover can only substitute a response that has not started flushing yet; a failure
+    // after headers are on the wire is unrecoverable by anyone.
     try {
-      return unwrap().serve(ctx, req);
+      return unwrap()
+          .serve(ctx, req)
+          .recover(Throwable.class, cause -> resolveHandler(ctx).handleException(ctx, req, cause));
     } catch (Exception e) {
       return resolveHandler(ctx).handleException(ctx, req, e);
     }
--- a/server/src/main/java/io/unitycatalog/server/exception/ExceptionHandlingDecorator.java
+++ b/server/src/main/java/io/unitycatalog/server/exception/ExceptionHandlingDecorator.java
@@ -2,37 +2,40 @@ package io.unitycatalog.server.exception;
 
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
-import com.linecorp.armeria.server.DecoratingHttpServiceFunction;
 import com.linecorp.armeria.server.HttpService;
 import com.linecorp.armeria.server.ServiceRequestContext;
+import com.linecorp.armeria.server.SimpleDecoratingHttpService;
 import com.linecorp.armeria.server.annotation.ExceptionHandlerFunction;
 
 /**
- * Catch exceptions and send response
+ * Turns an exception into a formatted error response, in the dialect of the service that was
+ * called.
  *
- * <p>The normal ExceptionHandler annotations do not catch exceptions from Decorators. This can be
- * used to catch those and return the desired response.
+ * <p>Armeria's own {@code @ExceptionHandler} support only wraps the annotated-service invocation,
+ * while authentication and PARAM/SYSTEM-source authorization run in route decorators outside it. An
+ * exception can therefore surface at two depths, so there are two subclasses registered at two
+ * depths: {@link ServiceExceptionHandlingDecorator} per service, and {@link
+ * GlobalExceptionHandlingDecorator} outermost. They differ only in how they answer {@link
+ * #resolveHandler}.
  *
- * <p>This should be set at the bottom of the decorator chain to ensure that it will catch
- * exceptions from upstream decorators.
- *
- * <p>.decorator(decorator1) .decorator(decorator2) .decorator(new ExceptionHandlingDecorator())
+ * <p>The global one must be registered last so it ends up at the bottom of the decorator chain,
+ * where it can catch exceptions from the decorators above it.
  */
-public class ExceptionHandlingDecorator implements DecoratingHttpServiceFunction {
+public abstract class ExceptionHandlingDecorator extends SimpleDecoratingHttpService {
 
-  private final ExceptionHandlerFunction exceptionHandlerFunction;
-
-  public ExceptionHandlingDecorator(ExceptionHandlerFunction exceptionHandlerFunction) {
-    this.exceptionHandlerFunction = exceptionHandlerFunction;
+  protected ExceptionHandlingDecorator(HttpService delegate) {
+    super(delegate);
   }
 
+  /** The dialect to render an exception from this request in. */
+  protected abstract ExceptionHandlerFunction resolveHandler(ServiceRequestContext ctx);
+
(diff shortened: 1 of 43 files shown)


server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -67,6 +66,11 @@ public class IcebergRestCatalogService {
   private final TableRepository tableRepository;
   private final SessionFactory sessionFactory;
 
+  @Override
+  public ExceptionHandlerFunction exceptionHandler() {
+    return IcebergRestExceptionHandler.INSTANCE;
+  }
```
timothyw553, line 77: maybe some tests for Iceberg and Delta auth errors?

server/src/test/java/io/unitycatalog/server/sdk/access/SdkStagingTableAccessControlTest.java
```diff
@@ -145,7 +153,7 @@ public void testManagedTableCreationByDifferentUserShouldFail() throws Exception
     grantPermissions(
         userCEmail, SecurableType.CATALOG, TestUtils.CATALOG_NAME, Privileges.USE_CATALOG);
     ServerConfig userCConfig = createTestUserServerConfig(userCEmail);
-    TestUtils.assertPermissionDenied(
+    assertDeniedInSurfaceFormat(
```
timothyw553, line 163: right below here there's another `TestUtils.assertPermissionDenied`. do we want to change that too?

timothyw553 (commented):
Two findings from first pass

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
yili-db: `assertDeniedInSurfaceFormat` is testing for both UC REST and Delta.
Iceberg doesn't have the auth yet (only metastore owner can access) so it can't be tested.

server/src/main/java/io/unitycatalog/server/exception/ExceptionHandlingDecorator.java
```diff
@@ -2,37 +2,40 @@
...
+  /** The dialect to render an exception from this request in. */
+  protected abstract ExceptionHandlerFunction resolveHandler(ServiceRequestContext ctx);
+
   @Override
-  public HttpResponse serve(HttpService delegate, ServiceRequestContext ctx, HttpRequest req)
-      throws Exception {
+  public final HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
     try {
-      return delegate.serve(ctx, req);
+      return unwrap().serve(ctx, req);
```
timothyw553: (ai assisted) we should check the httpresponse of `.serve()` via `response.recover()` for delayed/async crashes.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
timothyw553: > only metastore owner can access

right so is it possible to create a test to validate this? non-owner calls an Iceberg endpoint and verify a 403 occurred.

timothyw553: (ai assisted) also try to parse the body of the response with `ErrorResponseParser.fromJson(response.body())` to verify it is in the Iceberg format.

timothyw553 (approved):
second pass, generally looks good. I didn't see any pressing issues, except for the async/delayed response handling.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
yili-db: Testing Iceberg errors at this point is meaningless. I would postpone it to the PR that actually implements the Iceberg auth.

timothyw553 (approved):

server/src/main/java/io/unitycatalog/server/exception/GlobalExceptionHandler.java
```diff
@@ -36,7 +41,7 @@ protected HttpResponse createErrorResponse(BaseException exception) {
     }
 
     Map<String, Object> details = new HashMap<>();
-    details.put("@type", "google.rpc.ErrorInfo");
+    details.put("@type", "type.googleapis.com/google.rpc.ErrorInfo");
```
openinx, line 44: `@type` changed from `google.rpc.ErrorInfo` to `type.googleapis.com/google.rpc.ErrorInfo` here (follow-up commit), with no PR description callout and no assertion on the value. That is a user-facing envelope change for every UC error, independent of dialect routing.

Why here? Prefer revert + separate PR (or document as intentional compatibility and pin it in tests).

server/src/main/java/io/unitycatalog/server/exception/GlobalExceptionHandlingDecorator.java
```diff
@@ -0,0 +1,27 @@
...
+import com.linecorp.armeria.server.HttpService;
+import com.linecorp.armeria.server.ServiceRequestContext;
+import com.linecorp.armeria.server.annotation.ExceptionHandlerFunction;
+
+/**
+ * Registered globally, making it the outermost element of every chain and so the only thing that
+ * can catch a denial thrown by a route decorator, before the service is reached at all.
+ *
+ * <p>It tries to delegate to the {@link ServiceExceptionHandlingDecorator} of the calling service,
+ * or falls back to the default {@link GlobalExceptionHandler} it it finds none.
```
openinx: nit: typo — `it it finds` → `if it finds`.

openinx (commented):
Left comments.

openinx (approved):
Generally,  looks good to me, left two comments.

server/src/main/java/io/unitycatalog/server/exception/GlobalExceptionHandler.java
yili-db: Keeping it here, because the old value was simply wrong and this PR is what governs the error envelope. `@type` is the type URL of a `google.rpc.ErrorInfo` packed as a protobuf `Any`, and a type URL is `<prefix>/<fully-qualified-name>` with the standard prefix being `type.googleapis.com/`. The bare `google.rpc.ErrorInfo` was not a valid type URL, so a client matching on `@type` (the whole point of ErrorInfo being machine-readable) would not have recognized it. Since this PR is already touching how errors are rendered and shipping the envelope, correcting the detail rather than shipping it knowingly malformed avoids a second round-trip on the same field.

It is no longer silent: I added an assertion pinning `@type` to `type.googleapis.com/google.rpc.ErrorInfo` in the shared UC-envelope check (assertUcErrorEnvelope), so every UC error assertion now verifies it, and I called it out in the PR description.

PR #1339, opened 2026-02-04
Rename CredentialsGenerator to AwsCredentialGenerator

yili-db:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [ ] If there is a related issue, make sure it is linked to this PR.
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**
Rename CredentialsGenerator to AwsCredentialGenerator

And also rename *CredentialsGenerator to *CredentialGenerator to keep consistent with AwsCredentialVendor.
This addresses https://github.com/unitycatalog/unitycatalog/pull/1313#discussion_r2739045751

PR #1774, opened 2026-08-13
[Server] Grant KMS permissions in vended AWS session policy

### LINKED ISSUE #917: Temporary Table Credentials Lack `kms:Decrypt` Permission, Blocking Access to SSE-KMS Encrypted S3 Tables
**Describe the bug**

The temporary session credentials which can be obtained via the endpoint `POST /api/2.1/unity-catalog/temporary-table-credentials` prevent the client to perform the `kms:Decrypt` operation for an external table that lives within an SSE-KMS encrypted S3 bucket. This is due to the inline session policy set in [`AwsCredentialVendor.java:46`](https://github.com/unitycatalog/unitycatalog/blob/3baacf98ccbf3a90a22f1d9d8064e916eec80836/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialVendor.java#L46-L47): The boundary policy does not contain `kms:Decrypt` entries.

This makes it harder to directly use an AWS RDS to S3 export from Unity Catalog since Amazon mandates the use of SSE-KMS for the export. The current workaround is to re-encrypt the export prefix to SSE-S3, which does not require access to a KMS-key to read the data.

**To Reproduce**

1. Create an S3 Object with SSE-KMS encryption.
2. Create a role that allows to `kms:Decrypt` the used encryption key and set it in the `etc/conf/server.properties` S3 section: `s3.awsRoleArn.0=arn:aws:iam::*****:role/UnityCatalogS3Role`
3. Register the S3 object in Unity Catalog with `bin/uc table create --storage_location "s3://..."`. 
4. Try to access the table from a connected system (I used [the Apache Spark integration](https://docs.unitycatalog.io/integrations/unity-catalog-spark/#optional-running-spark-sql-for-cloud-object-stores)) and observe the error: `com.amazonaws.services.s3.model.AmazonS3Exception: User: arn:aws:sts::*****:assumed-role/UnityCatalogS3Role/***** is not authorized to perform: kms:Decrypt on resource: arn:aws:kms:eu-central-1:*****:key/***** because no session policy allows the kms:Decrypt action`

**Expected behavior**

The inline session policy should not prevent the decryption of SSE-KMS encrypted S3 objects.

**System:**

- Running the official [`unitycatalog/unitycatalog:v0.2.1`](https://hub.docker.com/layers/unitycatalog/unitycatalog/v0.2.1/images/sha256-b1d1758728f704b5cdd7c8cc8068965996f6f0d13eef5406d6b16a65dd611355) docker image in Amazon ECS.
- Connected to Apache Spark v3.5.4 [as described in the documentation](https://docs.unitycatalog.io/integrations/unity-catalog-spark/#optional-running-spark-sql-for-cloud-object-stores).

**Additional context**

N/A

moomindani:
**PR Checklist**

**Description of changes**

Fixes #917.

Tables stored in an S3 bucket with SSE-KMS encryption cannot be accessed through vended credentials. `AwsPolicyGenerator` grants only S3 actions, so the session policy attached to the vended credentials has no KMS permissions:

```java
static final List<String> SELECT_ACTIONS = List.of("s3:GetO*");
static final List<String> UPDATE_ACTIONS = List.of(
    "s3:GetO*", "s3:PutO*", "s3:DeleteO*", "s3:*Multipart*");
```

Since a session policy can only narrow what the assumed role is allowed to do, the missing `kms:Decrypt` makes the request fail even when the role itself has access to the key. (The issue references `AwsCredentialVendor.java:46`, which moved to `AwsPolicyGenerator` in #1339. The behaviour is unchanged on `main`.)

**Change**

A KMS statement is added to the generated policy:

- `SELECT`: `kms:Decrypt`
- `UPDATE`: `kms:Decrypt`, `kms:GenerateDataKey*`

It is appended after the per-bucket statements, so the index of the existing S3 statements within the policy does not change. For `s3://my-bucket/path1/table1` it generates as:

```json
{
  "Effect": "Allow",
  "Action": ["kms:Decrypt"],
  "Resource": ["*"],
  "Condition": {
    "StringLike": {
      "kms:ViaService": "s3.*.amazonaws.com",
      "kms:EncryptionContext:aws:s3:arn": [
        "arn:aws:s3:::my-bucket",
        "arn:aws:s3:::my-bucket/path1/table1/*",
        "arn:aws:s3:::my-bucket/path1/table1"
      ]
    }
  }
}
```

**On the resource being `*`**

Unity Catalog does not know which KMS key a bucket is configured with, so the statement cannot name one. It is narrowed by condition instead:

- `kms:ViaService` limits the permissions to KMS calls S3 makes on the caller's behalf, so the credentials cannot use the key for anything else.
- `kms:EncryptionContext:aws:s3:arn` limits them to the S3 ARNs this policy already grants access to. The encryption context reuses the same ARN strings the S3 statement's `Resource` is built from, added through one helper so the two lists cannot drift apart, IAM escaping included.

The bucket ARN is listed alongside the object ARNs because a bucket with S3 Bucket Keys enabled encrypts under the bucket ARN rather than the object ARN. That case cannot be narrowed to a path, since a single data key covers every object in the bucket.

A session policy is also intersected with the role's own permissions, so a role without KMS access still receives none. Both properties are verified against real AWS below.

**Testing**

Unit tests added to `AwsPolicyGeneratorTest`, confirmed to fail before the change and pass after it, covering both conditions, the escaping of IAM special characters in the encryption context, multiple buckets, and one test pinning the position of the S3 statements so a future reordering does not silently break the existing index-based assertions.

Verified against real AWS with an SSE-KMS bucket and a role allowed to use the key, passing the generated policy to `sts:AssumeRole --policy`:

| case | before | after |
|---|---|---|
| SELECT / GetObject | AccessDenied | success |
| UPDATE / PutObject | AccessDenied | success |
| GetObject on an object encrypted with bucket keys | AccessDenied | success |

The denial before the change names KMS, not S3:

```
is not authorized to perform: kms:Decrypt on resource: arn:aws:kms:...
because no session policy allows the kms:Decrypt action
```

Two negative cases confirm the statement grants nothing beyond the objects the policy already covers:

| case | result |
|---|---|
| role with S3 but no KMS permissions | denied, `no identity-based policy allows the kms:Decrypt action` |
| encryption context scoped to another path, S3 access unchanged | denied, `no session policy allows the kms:Decrypt action` |

The first shows the open resource adds nothing to what the role itself may do; the second shows the encryption context condition is actually enforced.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
+++ b/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
@@ -49,15 +49,25 @@ public class AwsPolicyGenerator {
       Resource: []
       """;
 
+  // The condition key S3 populates when it calls KMS on the caller's behalf. Its value is the ARN
+  // of the object being encrypted or decrypted, or the ARN of the bucket when the bucket has S3
+  // Bucket Keys enabled, in which case the data key is shared across the objects in the bucket.
+  static final String KMS_ENCRYPTION_CONTEXT_KEY = "kms:EncryptionContext:aws:s3:arn";
+
   // Unity Catalog doesn't know which KMS key a bucket is configured with, so the resource stays
-  // open. This grants nothing extra in practice: a session policy can only narrow what the
+  // open and the statement is narrowed by condition instead: to KMS calls made through S3, and to
+  // the S3 ARNs this policy already grants access to. A session policy can only narrow what the
   // assumed role is already allowed to do, so a role without KMS access still gets none.
   static final String KMS_STATEMENT = """
       Effect: Allow
       Action: []
       Resource:
         - "*"
-      """;
+      Condition:
+        StringLike:
+          "kms:ViaService": "s3.*.amazonaws.com"
+          "%s": []
+      """.formatted(KMS_ENCRYPTION_CONTEXT_KEY);
 
   private static final ObjectMapper JSON_MAPPER = new ObjectMapper();
   private static final ObjectMapper YAML_MAPPER = new ObjectMapper(new YAMLFactory());
@@ -89,6 +99,8 @@ public class AwsPolicyGenerator {
               privileges, locations));
     }
 
+    ArrayNode kmsEncryptionContexts = (ArrayNode) kmsStatement.findPath(KMS_ENCRYPTION_CONTEXT_KEY);
+
     // Group each location by s3 bucket it's located in, then for each
     // bucket, add the bucket arn for the listBucket and operations statements,
     // then add each path as a conditional prefix
@@ -100,6 +112,11 @@ public class AwsPolicyGenerator {
       ArrayNode operationsResource = (ArrayNode) operationsStatement.findPath("Resource");
       bucketResource.add(String.format("arn:aws:s3:::%s", bucketName));
 
+      // A bucket with S3 Bucket Keys enabled encrypts under the bucket arn rather than the object
+      // arn, so the bucket has to be allowed as an encryption context on its own. That case can't
+      // be scoped to a path: the same data key covers every object in the bucket.
+      kmsEncryptionContexts.add(String.format("arn:aws:s3:::%s", bucketName));
+
       ArrayNode conditionalPrefixes = (ArrayNode) listStatement.findPath("s3:prefix");
       paths.forEach(path -> {
         // remove any preceding forward slashes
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -43,6 +49,16 @@ public class AwsPolicyGenerator {
       Resource: []
       """;
 
+  // Unity Catalog doesn't know which KMS key a bucket is configured with, so the resource stays
+  // open. This grants nothing extra in practice: a session policy can only narrow what the
+  // assumed role is already allowed to do, so a role without KMS access still gets none.
+  static final String KMS_STATEMENT = """
```
yili-db, line 61: Can please you add `ViaService` condition:
```
  Condition:
    StringLike:
      "kms:ViaService": "s3.*.amazonaws.com"
```

As well as a path scoped down condition:
```
  StringLike:
    "kms:EncryptionContext:aws:s3:arn":
      - "arn:aws:s3:::my-bucket"            # bucket-key case
      - "arn:aws:s3:::my-bucket/path1/table1/*"  # per-object case
```
You can find existing code in this file that generates the bucket and path resource strings.

Thank you!

moomindani: Added both conditions, thank you. For `s3://my-bucket/path1/table1` the KMS statement now generates as:

```json
{
  "Effect": "Allow",
  "Action": ["kms:Decrypt"],
  "Resource": ["*"],
  "Condition": {
    "StringLike": {
      "kms:ViaService": "s3.*.amazonaws.com",
      "kms:EncryptionContext:aws:s3:arn": [
        "arn:aws:s3:::my-bucket",
        "arn:aws:s3:::my-bucket/path1/table1/*",
        "arn:aws:s3:::my-bucket/path1/table1"
      ]
    }
  }
}
```

The encryption context reuses the very same ARN strings the S3 statement's `Resource` is built from, added through one helper so the two lists can't drift apart, IAM escaping included.

The bucket ARN is listed alongside the object ARNs because a bucket with S3 Bucket Keys enabled encrypts under the bucket ARN rather than the object ARN. That case can't be narrowed to a path, since a single data key covers every object in the bucket.

Re-verified against real AWS, since a condition that fails to match would break access rather than widen it (SSE-KMS bucket, role allowed to use the key, generated policy passed to `sts:AssumeRole --policy`):

| case | result |
|---|---|
| SELECT / GetObject | success |
| UPDATE / PutObject | success |
| GetObject on an object encrypted with bucket keys | success |
| role with S3 but no KMS permissions | denied, `no identity-based policy allows the kms:Decrypt action` |
| encryption context scoped to another path, S3 access unchanged | denied, `no session policy allows the kms:Decrypt action` |

The last row confirms the condition is actually enforced. Unit tests cover both conditions.

yili-db (approved):

PR #1777, opened 2026-08-13
Authorize PAYLOAD-source keys against the bound request body

yili-db:
**PR Checklist**

**Description of changes**

PAYLOAD-source authorization read the body through Armeria's `peekData`, which only sees chunks and has to guess when the body is complete. Anything it did not recognise -- no body, a non-JSON content-type, malformed JSON, trailing whitespace -- skipped the check and trigger a failure.

- `AuthorizationGateConverter` binds the body first, then authorizes the bound object, so authorization reads exactly what the handler receives. A body that cannot bind is a client error (400) before authorization; a bound body with no usable key is a denial (403). The handler runs in neither case.
- `UnityAccessDecorator` stashes a `PayloadAuthorizer` for the gate to invoke, replacing `peekData`, `PeekDataHandler` and `AUTH_PENDING`. Both branches set one, so a missing value is never read as "already authorized".
- `EvaluationAction` becomes an abstract class owning the single response-splitting path, skipping the split when authorization already ran inline.
- The gate is installed only when authorization is enabled, since nothing produces a `PayloadAuthorizer` otherwise.

One test covers the body shapes the SDK cannot produce: body-less, non-JSON content-type, malformed JSON, JSON literal `null`, trailing newline, charset-qualified content-type, and a body split across two chunks mid-token.

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/utils/TestUtils.java
+++ b/server/src/test/java/io/unitycatalog/server/utils/TestUtils.java
@@ -305,8 +305,8 @@ public class TestUtils {
 
   /**
    * Sends a body-less POST to the given path. The SDK always attaches a serialized body, so raw
-   * HTTP is the only way to reach the body-less code path (used to exercise {@link
-   * io.unitycatalog.server.auth.decorator.AuthorizationGateConverter}'s silent-skip denial).
+   * HTTP is the only way to reach the body-less code path -- where the body cannot bind and the
+   * request fails as a 400 during binding, before authorization runs.
    */
   public static HttpResponse<String> sendRawEmptyPost(ServerConfig config, String path)
       throws Exception {
--- a/server/src/test/java/io/unitycatalog/server/utils/TestUtils.java
+++ b/server/src/test/java/io/unitycatalog/server/utils/TestUtils.java
@@ -18,9 +18,13 @@ import java.net.URI;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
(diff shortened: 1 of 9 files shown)


server/src/test/java/io/unitycatalog/server/utils/TestUtils.java
```diff
@@ -151,10 +304,81 @@ public static void assertHttpApiException(
    */
   public static HttpResponse<String> sendRawEmptyPost(ServerConfig config, String path)
       throws Exception {
+    return sendRawPost(config, path, HttpRequest.BodyPublishers.noBody(), null);
```
openinx, line 313: nit: the javadoc above still says "silent-skip denial"; body-less / non-JSON now fail in binding as 400. Update the comment to match.

openinx (approved):
**Verdict:** LGTM. Minor suggestion.

**Summary**
- Replaces `peekData` / `AUTH_PENDING` with bind-then-authorize via `PayloadAuthorizer`, so PAYLOAD auth sees the same object the handler gets; gate installed only when auth is enabled.
- Risk is contained: good shape coverage (400 vs 403), fail-closed when the authorizer is missing; stacked on #1772.

**Nits**
- `TestUtils.java` (`sendRawEmptyPost` javadoc): still says "silent-skip denial"; body-less / non-JSON now fail in binding as 400. Update the comment to match.

**For human reviewer**
- Auth is security-sensitive; worth a quick look from whoever owns UC server auth, especially the `/auth/tokens` exclusion + parameter `@RequestConverter` pairing (covered by `AuthServiceTest`, but easy to break later).

PR #1778, opened 2026-08-14
File content based commit idempotency for CCv2

yili-db:
**PR Checklist**

**Description of changes**

Once a commit version is backfilled, its staged file name is no longer tracked in DB, so the DB alone cannot tell a replay from a conflict. Resolve such a version by comparing the incoming staged file against the published commit file, outside the transaction.

- `classifyCommitReplay` is now the total factory `exceptionForAlreadyTakenVersion`, returning the exception to throw: replay, conflict, or a new `CommitContentCheckRequiredException` for a purged version.
- The entry points catch `CommitContentCheckRequiredException` and call `verifyContentReplayOrThrowConflict` out of the transaction: identical content is an idempotent no-op success, a definitive difference is a 409 conflict.
- Reads fail open: if the staged or published commit file cannot be read, return the new `COMMIT_STATE_UNKNOWN` (500, retriable) instead of a false conflict.
- `hasSameFileContent` streams both files in lockstep with a fixed buffer, so a client-influenced staged file is never read wholesale into memory (DoS-safe).
- `Repositories` now owns the `CloudCredentialVendor` -> `StorageCredentialVendor` -> `FileOperations` chain, giving the repository layer credentialed reads of table storage.

Tested on both entry points (REST `postCommit`, Delta `updateTableForDelta`): match, differ, and unreadable-published cases.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/exception/ErrorCode.java
+++ b/server/src/main/java/io/unitycatalog/server/exception/ErrorCode.java
@@ -48,9 +48,6 @@ public enum ErrorCode {
   UNIMPLEMENTED(12, 501, DeltaErrorType.NOT_IMPLEMENTED_EXCEPTION),
   INTERNAL(13, 500, DeltaErrorType.INTERNAL_SERVER_ERROR_EXCEPTION),
   DATA_LOSS(15, 500, DeltaErrorType.INTERNAL_SERVER_ERROR_EXCEPTION),
-  // CCv2 commit whose idempotency the server could not determine (e.g. the published or staged
-  // commit file could not be read for a content comparison). Retriable.
-  COMMIT_STATE_UNKNOWN(22, 500, DeltaErrorType.COMMIT_STATE_UNKNOWN_EXCEPTION),
   // UC-specific "already exists" error codes for backwards compatibility (all HTTP 400). But new
   // Delta API returns 409 as it has no backwards compatibility concern.
   RESOURCE_ALREADY_EXISTS(16, 400, DeltaErrorType.ALREADY_EXISTS_EXCEPTION, 409),
@@ -58,7 +55,10 @@ public enum ErrorCode {
   SCHEMA_ALREADY_EXISTS(18, 400, DeltaErrorType.ALREADY_EXISTS_EXCEPTION, 409),
   TABLE_ALREADY_EXISTS(19, 400, DeltaErrorType.ALREADY_EXISTS_EXCEPTION, 409),
   STORAGE_CREDENTIAL_ALREADY_EXISTS(20, 400, DeltaErrorType.ALREADY_EXISTS_EXCEPTION, 409),
-  EXTERNAL_LOCATION_ALREADY_EXISTS(21, 400, DeltaErrorType.ALREADY_EXISTS_EXCEPTION, 409);
+  EXTERNAL_LOCATION_ALREADY_EXISTS(21, 400, DeltaErrorType.ALREADY_EXISTS_EXCEPTION, 409),
+  // CCv2 commit whose idempotency the server could not determine (e.g. the published or staged
+  // commit file could not be read for a content comparison). Retriable.
+  COMMIT_STATE_UNKNOWN(22, 500, DeltaErrorType.COMMIT_STATE_UNKNOWN_EXCEPTION);
 
   // Canonical mapping from DeltaErrorType to HTTP status. Built at class-load time and validated
   // to ensure every ErrorCode that shares the same deltaErrorType also agrees on deltaHttpStatus.
--- a/server/src/main/java/io/unitycatalog/server/exception/ErrorCode.java
+++ b/server/src/main/java/io/unitycatalog/server/exception/ErrorCode.java
@@ -48,6 +48,9 @@ public enum ErrorCode {
   UNIMPLEMENTED(12, 501, DeltaErrorType.NOT_IMPLEMENTED_EXCEPTION),
   INTERNAL(13, 500, DeltaErrorType.INTERNAL_SERVER_ERROR_EXCEPTION),
   DATA_LOSS(15, 500, DeltaErrorType.INTERNAL_SERVER_ERROR_EXCEPTION),
+  // CCv2 commit whose idempotency the server could not determine (e.g. the published or staged
+  // commit file could not be read for a content comparison). Retriable.
+  COMMIT_STATE_UNKNOWN(22, 500, DeltaErrorType.COMMIT_STATE_UNKNOWN_EXCEPTION),
   // UC-specific "already exists" error codes for backwards compatibility (all HTTP 400). But new
   // Delta API returns 409 as it has no backwards compatibility concern.
   RESOURCE_ALREADY_EXISTS(16, 400, DeltaErrorType.ALREADY_EXISTS_EXCEPTION, 409),
--- a/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
@@ -35,6 +35,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -851,8 +852,13 @@ public class DeltaCommitRepository {
     }
     return new BaseException(
         ErrorCode.COMMIT_VERSION_CONFLICT,
-        "Commit version already accepted. Current table version is "
-            + lastCommitDAO.getCommitVersion());
+        "Commit version already accepted. Version "
+            + version
+            + " was accepted with commit file "
+            + existing.getCommitFilename()
+            + ", but the request carried "
+            + fileName
+            + ".");
   }
 
   /**
@@ -888,8 +894,11 @@ public class DeltaCommitRepository {
   static void verifyContentReplayOrThrowConflict(
       FileOperations fileOperations, CommitContentCheckRequiredException check) {
     String logDir = check.tableLocation + "/_delta_log";
-    String publishedPath = logDir + "/" + String.format("%020d.json", check.version);
-    String stagedPath = logDir + "/_staged_commits/" + check.stagedFileName;
+    // Locale.ROOT: the published file name is ASCII digits regardless of the server's locale, so it
+    // matches the actual _delta_log/<v>.json path (some locales render %d with non-ASCII digits).
+    String publishedPath = String.format(Locale.ROOT, "%s/%020d.json", logDir, check.version);
+    String stagedPath =
+        String.format(Locale.ROOT, "%s/_staged_commits/%s", logDir, check.stagedFileName);
     boolean sameContent;
     // getFileIO can vend credentials (cloud paths) and open resources, so it is acquired inside the
     // guarded block (and closed): a vend or read failure is equally "cannot determine" and must
--- a/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
@@ -18,6 +18,7 @@ import io.unitycatalog.server.persist.dao.ColumnInfoDAO;
 import io.unitycatalog.server.persist.dao.DeltaCommitDAO;
 import io.unitycatalog.server.persist.dao.PropertyDAO;
 import io.unitycatalog.server.persist.dao.TableInfoDAO;
+import io.unitycatalog.server.persist.utils.FileOperations;
 import io.unitycatalog.server.persist.utils.TransactionManager;
 import io.unitycatalog.server.service.delta.DeltaUniformUtils;
 import io.unitycatalog.server.service.delta.UcManagedDeltaContract;
@@ -27,6 +28,9 @@ import io.unitycatalog.server.utils.NormalizedURL;
 import io.unitycatalog.server.utils.ServerProperties;
 import io.unitycatalog.server.utils.ValidationUtils;
 import jakarta.persistence.PessimisticLockException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
@@ -36,6 +40,9 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import lombok.AllArgsConstructor;
+import org.apache.iceberg.io.FileIO;
+import org.apache.iceberg.io.InputFile;
 import org.hibernate.LockMode;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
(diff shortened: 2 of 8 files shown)


server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -832,6 +871,78 @@ private static Optional<DeltaCommitDAO> findCommitByVersion(
...
+   *       COMMIT_VERSION_CONFLICT} (409).
+   *   <li>either file unreadable -&gt; {@code COMMIT_STATE_UNKNOWN} (500, retriable), so the client
+   *       retries rather than receiving a false conflict.
+   * </ul>
+   */
+  static void verifyContentReplayOrThrowConflict(
+      FileOperations fileOperations, CommitContentCheckRequiredException check) {
+    String logDir = check.tableLocation + "/_delta_log";
+    String publishedPath = logDir + "/" + String.format("%020d.json", check.version);
+    String stagedPath = logDir + "/_staged_commits/" + check.stagedFileName;
```
timothyw553: should we add validation that `check.stagedFileName` is _just_ a single file name? Nothing like `../000002.json`. The name comes from a commit client so some server-side validation would be good

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
```diff
@@ -346,14 +350,28 @@ public void testBasicCoordinatedCommitsCRUD() throws ApiException {
...
+    verifyDeltaCommits(/* expectedLatestTableVersion= */ 4);
+
+    // A staged file that differs from the published commit means another writer won v2: 409.
+    byte[] otherContent = "different\n".getBytes(StandardCharsets.UTF_8);
+    writeTableFile(loc, "_delta_log/_staged_commits/file2", otherContent);
     assertApiException(
-        () -> deltaCommitsApi.commit(commit2),
-        ErrorCode.ALREADY_EXISTS,
-        "Commit version already accepted.");
+        () -> deltaCommitsApi.commit(commit2), ErrorCode.ALREADY_EXISTS, "already accepted");
```
timothyw553: perhaps we can harden the test assertion by checking the response body
```diff
- assertApiException(
-     () -> deltaCommitsApi.commit(commit2), ErrorCode.ALREADY_EXISTS, "already accepted");
+ ApiException conflict =
+     assertThrows(ApiException.class, () -> deltaCommitsApi.commit(commit2));
+ assertThat(conflict.getCode()).isEqualTo(409);
+ assertThat(conflict.getResponseBody()).contains("COMMIT_VERSION_CONFLICT");
```

server/src/test/java/io/unitycatalog/server/sdk/delta/SdkUpdateTableTest.java
```diff
@@ -1285,6 +1291,61 @@ h1, new DeltaSetLatestBackfilledVersionUpdate().latestPublishedVersion(5L)),
...
+
+      // A staged file whose content differs from the published commit is a definitive conflict.
+      writeTableFile(
+          loc,
+          "_delta_log/_staged_commits/00000002.json",
+          "different\n".getBytes(StandardCharsets.UTF_8));
+      TestUtils.assertDeltaApiException(
+          () -> updateTable(h3, commits[1]),
+          DeltaErrorType.COMMIT_VERSION_CONFLICT_EXCEPTION,
+          "already accepted");
```
timothyw553, line 1339: some thoughts on tests
  (1) the file exists -> byte-for-byte match vs no match
  (2) the file straight up doesn't exist
  (3) we cannot even read the file 

we have coverage for (1), could we add tests for (2) and (3)? there may be other cases too.

timothyw553 (approved):
straight forward solution but **really** important, and generally looks good to me. I left some small comments about testing and one security/validation concern.

server/src/test/java/io/unitycatalog/server/sdk/deltacommits/SdkDeltaCommitsCRUDTest.java
yili-db: `assertApiException` is the right shape to use. The error body parsing will be handled in https://github.com/unitycatalog/unitycatalog/pull/1772. I am adding additional change there.

server/src/main/java/io/unitycatalog/server/exception/ErrorCode.java
```diff
@@ -48,6 +48,9 @@ public enum ErrorCode {
   UNIMPLEMENTED(12, 501, DeltaErrorType.NOT_IMPLEMENTED_EXCEPTION),
   INTERNAL(13, 500, DeltaErrorType.INTERNAL_SERVER_ERROR_EXCEPTION),
   DATA_LOSS(15, 500, DeltaErrorType.INTERNAL_SERVER_ERROR_EXCEPTION),
+  // CCv2 commit whose idempotency the server could not determine (e.g. the published or staged
+  // commit file could not be read for a content comparison). Retriable.
+  COMMIT_STATE_UNKNOWN(22, 500, DeltaErrorType.COMMIT_STATE_UNKNOWN_EXCEPTION),
```
openinx: Why do we place the error code with `22` in the mid of `15` and `16` ?  ca you move to the end of line 61 if we really want to add a new error code ?

yili-db: It's put here because it's one of the 500.
The GRPC error code currently isn't used at all. It just happened to exist in this code repo.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -832,6 +871,78 @@ private static Optional<DeltaCommitDAO> findCommitByVersion(
...
+      sameContent =
+          hasSameFileContent(fileIO.newInputFile(publishedPath), fileIO.newInputFile(stagedPath));
+    } catch (Exception e) {
+      throw new BaseException(
+          ErrorCode.COMMIT_STATE_UNKNOWN,
+          "Could not determine whether commit version "
+              + check.version
+              + " is a replay: unable to read the staged or published commit file for the table at "
+              + check.tableLocation
+              + "; retry the request.",
```
openinx, line 916: should we need to be more clear to say that only retry when we are pretty confident the stage commit file isn't matching the published file ?

server/src/test/java/io/unitycatalog/server/persist/DeltaCommitContentCheckTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    // 2. Published and staged files are byte-identical: a recognized replay -> no exception.
+    putPublished("tbl_match", "{\"commitInfo\":{\"version\":2}}\n");
+    putStaged("tbl_match", "{\"commitInfo\":{\"version\":2}}\n");
+    assertThatCode(() -> runContentCheck("tbl_match")).doesNotThrowAnyException();
+
+    // 3. Published and staged files differ: another writer won this version -> 409 conflict.
+    putPublished("tbl_conflict", "published-commit\n");
+    putStaged("tbl_conflict", "a-different-commit\n");
+    assertThatThrownBy(() -> runContentCheck("tbl_conflict"))
```
openinx, line 75: This is basically verify the two file equality,  which only cover the `verifyContentReplayOrThrowConflict` itself.   How could we simulate the e2e test cases that the client can really build a mismatched file which should trigger either direct retry or pre-check before retry ?  I think that's the part we will need to take a look.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: No. With this PR it's **always** safe to retry the same RPC as is.
And only rebase-then-retry when it's a clear conflict.

This error message is about the former.

server/src/test/java/io/unitycatalog/server/persist/DeltaCommitContentCheckTest.java
yili-db: This is covered in the CRUD test https://github.com/unitycatalog/unitycatalog/pull/1778/changes#diff-52d156946b0141e68a091d61230e2264a6a1acac4860d8a9e1cef5d3350e0804R372-R378
B

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -832,6 +871,78 @@ private static Optional<DeltaCommitDAO> findCommitByVersion(
...
+   *   <li>a definitive difference -&gt; another writer won this version; throws {@code
+   *       COMMIT_VERSION_CONFLICT} (409).
+   *   <li>either file unreadable -&gt; {@code COMMIT_STATE_UNKNOWN} (500, retriable), so the client
+   *       retries rather than receiving a false conflict.
+   * </ul>
+   */
+  static void verifyContentReplayOrThrowConflict(
+      FileOperations fileOperations, CommitContentCheckRequiredException check) {
+    String logDir = check.tableLocation + "/_delta_log";
+    String publishedPath = logDir + "/" + String.format("%020d.json", check.version);
```
timothyw553: should we use Locale.ROOT?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -832,6 +871,78 @@ private static Optional<DeltaCommitDAO> findCommitByVersion(
...
+              + check.tableLocation
+              + "; retry the request.",
+          e);
+    }
+    if (!sameContent) {
+      throw new BaseException(
+          ErrorCode.COMMIT_VERSION_CONFLICT,
+          "Commit version already accepted. Version " + check.version + " is already published.");
+    }
+  }
```
timothyw553, line 924: (ai assisted)  The specification currently says: "Re-sending an add-commit that was already accepted fails with CommitVersionConflictException."

What is wrong: The implementation can now return success for a purged commit, while the normative specification says that a resend fails.

does this warrant a protocol wording update? if so, we can do in a follow up
```diff
- Re-sending an accepted add-commit fails with CommitVersionConflictException.
+ Re-sending an accepted add-commit is an idempotent success.
+ For a purged commit record, the server compares the staged and published files.
+ Equal content returns success, different content returns CommitVersionConflictException,
+ and an unreadable file returns CommitStateUnknownException.
```

server/src/main/java/io/unitycatalog/server/exception/ErrorCode.java
openinx: Will it make more sense to move it to the end of the list ? as we are ordering them by the `<code, httpStatus>` , where the `code` is ordering first.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
openinx: nit: Can you unify to always use the `String.format(...)` to concat all the parts to the complete string ?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -1120,6 +1231,17 @@ private static void validateCommitInfo(DeltaCommitInfo commitInfo) {
...
         "Field can not be empty: %s",
         DeltaCommitInfo.JSON_PROPERTY_FILE_NAME);
+    // The file name is client-supplied and the server later builds the staged-commit path it reads
+    // from it (_delta_log/_staged_commits/<fileName>). Require a single path segment so a name like
+    // "../<v>.json" cannot traverse out of that directory.
+    String fileName = commitInfo.getFileName();
+    ValidationUtils.checkArgument(
+        !fileName.contains("/")
+            && !fileName.contains("\\")
+            && !fileName.equals(".")
```
openinx, line 1250: The staging commit file name is liking the following, so why it won't have the DOT in its file name ? 

```
├── _delta_log
│   ├── _last_checkpoint
│   ├── _sidecars
│   │   └── 00000000000000000200.checkpoint.0000000001.0000000001.29b137f1-a2da-43fe-bc96-36d56c135f59.parquet
│   ├── _staged_commits
│   │   ├── 00000000000000000001.449c6f2f-f0c7-4bd0-97c0-dc70f4594a09.json
│   │   ├── 00000000000000000002.a9407a08-47f5-4465-9df8-2f7ce3747f4b.json
```

should we remove this DOT precondition check ?

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
```diff
@@ -793,27 +831,28 @@ private static boolean isAcceptedCommitReplay(
...
-    // The file name embeds a client-generated, per-commit-unique UUID, used as the dedup handle: a
-    // match means the same commit replayed; a difference means another writer won that version.
-    return existing.getCommitFilename().equals(fileName);
+    if (existing.getCommitFilename().equals(fileName)) {
+      return new CommitAlreadyAcceptedException();
+    }
+    return new BaseException(
+        ErrorCode.COMMIT_VERSION_CONFLICT,
+        "Commit version already accepted. Current table version is "
+            + lastCommitDAO.getCommitVersion());
```
openinx: I will suggest to be more clear in the error message,  let's say `Commit version {} already accepted, and the expected commit file is {}, but the accepted one is {}`.   I think the commit file is even more important than the commit version in this error message.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -309,6 +309,12 @@ public DeltaLoadTableResponse updateTableForDelta(
       // Idempotent replay: the transaction rolled back to a no-op. Return the current table state
       // (a fresh read, since the rolled-back transaction produced no response).
       return loadTableForDelta(catalog, schema, table);
+    } catch (DeltaCommitRepository.CommitContentCheckRequiredException e) {
+      // Version was purged, so the DB couldn't decide: settle it out of the (rolled-back)
+      // transaction by comparing file content. A match is a no-op success (return current state);
+      // a difference throws a conflict.
+      DeltaCommitRepository.verifyContentReplayOrThrowConflict(repositories.getFileOperations(), e);
+      return loadTableForDelta(catalog, schema, table);
```
openinx, line 317: I think the one concurrent problem is: even if we succesfully verified the matching contents of staging commit file and real commit file.  if someone committed another txn between line316 and line317,  then the `loadTableForDelta` will actually just load that latest version,  rather than the version that we expected.   essentially that will cause semantic problem if the connector is relying on the updateTable response.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: It is `equals(".")`, not `contains(".")`

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
yili-db: Spec doc is being updated in https://github.com/unitycatalog/unitycatalog/pull/1666/changes

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
yili-db: From protocol spec there is no guarantee that response to a `updateTable` will never pick up changes after that commit. It only says `Response: 200 See Load a Table response for the full response shape.`.
So it's OK if a response of `updateTable` contains newer changes.

In practice it's impossible to remember table metadata snapshots of each commit. The only result that can be returned by `updateTable` is the latest table status by the time of response.

server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java
openinx: okay, sounds good.  I misread this.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
openinx: > It only says Response: 200 See Load a Table response for the full response shape..
So it's OK if a response of updateTable contains newer changes.

This is confusing actually,  the `200` should means it's the current rest call that successfully updated the table, not other concurrent one.  so the `updateTable` should be strictly returned as the current updated version of table. I will suggest to strength the semantic of the spec, so that it will always give us a deterministic response. 

> In practice it's impossible to remember table metadata snapshots of each commit. The only result that can be returned by updateTable is the latest table status by the time of response.

If the `updateTable` can successfully return a response,  that means we are successfully update the table in this call, right ?  then as we willl always has the table row write lock to guarantee the database txn isolation,  the table loaded by the current session should always return the correct version of the table. 

Will it possible to move this `loadTableForDelta` under the table row lock context, so that we can accomplish the semantic ?

openinx: Discussed offline,  as we cannot restore the complete committed table info that exactly matches the commit version,  so it seems impossible to get the exact version of table info as the table response.  then  please help us to clearly state the semantic in the `updateTable` specification doc,  @yili-db .

openinx (approved):
Looks good to me now, pls take a look at the doc follow-up https://github.com/unitycatalog/unitycatalog/pull/1778#discussion_r3832250758

PR #1780, opened 2026-08-17
[Server] Reject the bootstrap `admin` principal in OAuth token exchange

yili-db:
**PR Checklist**

**Description of changes**

- The `/auth/tokens` exchange treated the literal `admin` subject as always allowed, letting an attacker-controlled token assume the bootstrap metastore-`OWNER` user.
- Reject `admin` in `findEnabledUser` (`TokenExchangeSubjectTokenHandler`), covering both the email-claim and OAuth-client (`external_id`) paths, and compare the resolved user's email so case variants like `ADMIN` are rejected on every backend.
- Added unit and end-to-end tests for these rejections; exchange as a real ENABLED user in the happy-path tests.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/AuthService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/AuthService.java
@@ -237,17 +237,15 @@ public class AuthService implements RegisteredService {
 
     LOGGER.debug("Validating principal: {}", subject);
 
-    // "admin" is the internal service-token principal (metastore OWNER) and must not be
-    // assumable via external token exchange. An explicit reject is required because "admin" is a
-    // real ENABLED user that would otherwise pass the lookup below.
-    if (subject.equals("admin")) {
-      throw new OAuthInvalidRequestException(
-          ErrorCode.INVALID_ARGUMENT, "User not allowed: " + subject);
-    }
-
     try {
       User user = userRepository.getUserByEmail(subject);
-      if (user != null && user.getState() == User.StateEnum.ENABLED) {
+      // The bootstrap "admin" user (metastore OWNER) is provisioned only for the internal service
+      // token and must not be assumable via external token exchange. Comparing the resolved user's
+      // email -- rather than the incoming subject -- also rejects case variants such as "ADMIN"
+      // that a case-insensitive database collation would otherwise resolve back to the admin user.
+      if (user != null
+          && !"admin".equals(user.getEmail())
+          && user.getState() == User.StateEnum.ENABLED) {
         LOGGER.debug("Principal {} is enabled", subject);
         return;
       }
--- a/server/src/main/java/io/unitycatalog/server/service/AuthService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/AuthService.java
@@ -237,9 +237,12 @@ public class AuthService implements RegisteredService {
 
     LOGGER.debug("Validating principal: {}", subject);
 
+    // "admin" is the internal service-token principal (metastore OWNER) and must not be
+    // assumable via external token exchange. An explicit reject is required because "admin" is a
+    // real ENABLED user that would otherwise pass the lookup below.
     if (subject.equals("admin")) {
-      LOGGER.debug("admin always allowed");
-      return;
+      throw new OAuthInvalidRequestException(
+          ErrorCode.INVALID_ARGUMENT, "User not allowed: " + subject);
     }
 
     try {
--- a/server/src/test/java/io/unitycatalog/server/service/AuthServiceTest.java
+++ b/server/src/test/java/io/unitycatalog/server/service/AuthServiceTest.java
@@ -249,17 +249,40 @@ public class AuthServiceTest extends BaseAuthCRUDTest {
   }
 
   @Test
-  public void testTokenExchangeRejectsDisallowedPrincipals() {
-    // The reserved "admin" principal must never be exchangeable (it is the internal service-token,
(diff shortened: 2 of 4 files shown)


server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -237,9 +237,12 @@ private void verifyPrincipal(DecodedJWT decodedJWT) {
...
     LOGGER.debug("Validating principal: {}", subject);
 
+    // "admin" is the internal service-token principal (metastore OWNER) and must not be
+    // assumable via external token exchange. An explicit reject is required because "admin" is a
+    // real ENABLED user that would otherwise pass the lookup below.
     if (subject.equals("admin")) {
-      LOGGER.debug("admin always allowed");
-      return;
+      throw new OAuthInvalidRequestException(
+          ErrorCode.INVALID_ARGUMENT, "User not allowed: " + subject);
```
timothyw553: MySQL’s normal text collation is case-insensitive. so `ADMIN` does not match `"admin"` in Java, but MySQL can resolve `ADMIN` to the stored admin user. what if instead, below we check based on email as the user's principal key?
```diff
-    if (subject.equals("admin")) {
-      throw new OAuthInvalidRequestException(
-          ErrorCode.INVALID_ARGUMENT, "User not allowed: " + subject);
-    }
-
     try {
       User user = userRepository.getUserByEmail(subject);
-      if (user != null && user.getState() == User.StateEnum.ENABLED) {
+      if (user != null
+          && !"admin".equals(user.getEmail())
+          && user.getState() == User.StateEnum.ENABLED) {
         LOGGER.debug("Principal {} is enabled", subject);
         return;
       }
```

server/src/test/java/io/unitycatalog/server/service/AuthServiceTest.java
```diff
@@ -211,14 +236,32 @@ public void testTokenExchangeWithWrongIssuerAndCorrectAudience() throws NoSuchAl
...
     // The issuer is not in the allowlist → 403 Forbidden
     assertThat(response.status()).isEqualTo(HttpStatus.UNAUTHORIZED);
   }
 
+  @Test
+  public void testTokenExchangeRejectsDisallowedPrincipals() {
+    // The reserved "admin" principal must never be exchangeable (it is the internal service-token,
+    // metastore-OWNER identity), and a correctly-signed token for any non-enabled user must fail
+    // closed. Both return the same generic 400 so "admin" is indistinguishable as a reserved name.
+    for (String subject : List.of("admin", "nobody@example.com")) {
```
timothyw553: add tests for 
  - `sub=opaque-id`, `email=admin`
  - `sub=ADMIN` on MySQL
  - An unknown user

for example
```java
  @Test
  public void rejectsCaseVariantOfBootstrapPrincipal() throws Exception {
    String token =
        createIdentityToken(
            "ADMIN", testIssuer, TEST_AUDIENCE, testIssuerAlgorithm, testIssuerKeyId);

    AggregatedHttpResponse response = exchangeToken(token);

    assertThat(response.status()).isEqualTo(HttpStatus.BAD_REQUEST);

    JsonNode error = MAPPER.readTree(response.contentUtf8());
    assertThat(error.get("error_code").asText()).isEqualTo("INVALID_ARGUMENT");
    assertThat(error.has("access_token")).isFalse();
  }
```

timothyw553 (commented):
some comments.

timothyw553:
@yili-db there is a merge conflict here. can you please fix it?

server/src/main/java/io/unitycatalog/server/service/TokenExchangeSubjectTokenHandler.java
```diff
@@ -110,12 +110,23 @@ private Optional<User> findEnabledUser(Supplier<User> lookup, String lookupType,
...
 
+  /**
+   * The bootstrap "admin" user (metastore OWNER) is provisioned only for the internal service token
+   * and must never be assumable through external token exchange, by either the email-claim or the
+   * OAuth-client resolution path. Comparing the resolved user's email -- rather than the incoming
+   * subject -- also rejects case variants such as "ADMIN" that a case-insensitive database
+   * collation would otherwise resolve back to the admin user.
+   */
+  private static boolean isAssumablePrincipal(User user) {
+    return !"admin".equals(user.getEmail());
```
openinx, line 128: > Comparing the resolved user's email -- rather than the incoming
> * subject -- also rejects case variants such as "ADMIN" that a case-insensitive database
> * collation would otherwise resolve back to the admin user.

This is very confusing.  as the `user.getEmail()` will has the format `<name>@<domain>`, then how will the `admin` has the chance to be equals to `user.getEmail()` ?

openinx: And how will `ADMIN` upper-case word turn us to use the email comparison ?

yili-db: The bootstrap admin isn't a real email — `UnityAccessUtil.initializeAdmin` provisions it with the literal email `admin` (and metastore OWNER), so `user.getEmail()` can equal `"admin"`.

The incoming `ADMIN` is just the lookup key: on a case-insensitive DB collation `getUserByEmail("ADMIN")` returns that same bootstrap user whose *stored* email is `admin`, so comparing the resolved email still rejects it. I've reworded the comment to make this clear.

PR #1781, opened 2026-08-17
Invalidate access tokens on logout

yili-db:
**PR Checklist**

**Description of changes**

- Add `TokenRevocationDAO` (`uc_token_revocations` table, keyed by token `jti`) and `TokenRevocationRepository` to record revoked tokens in the database, with an index on `expires_at` and best-effort asynchronous cleanup of expired rows.
- On `/logout`, revoke the presented `ACCESS` token's `jti` until its `exp`; `SERVICE` tokens are left untouched since they are not login sessions.
- In `AuthDecorator`, reject a revoked `jti` after signature/expiry verification, checked in the same transaction as the user lookup.
- Register `TokenRevocationDAO` in `HibernateConfigurator` and wire `TokenRevocationRepository` into `Repositories`.
- Add `TokenRevocationRepositoryTest` and extend `AuthServiceTest` to cover `ACCESS` token revocation and the `SERVICE` token no-op.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/TokenRevocationRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TokenRevocationRepository.java
@@ -1,5 +1,6 @@
 package io.unitycatalog.server.persist;
 
+import io.unitycatalog.server.exception.BaseException;
 import io.unitycatalog.server.persist.dao.TokenRevocationDAO;
 import io.unitycatalog.server.persist.utils.TransactionManager;
 import java.util.Date;
@@ -32,9 +33,15 @@ public class TokenRevocationRepository {
   }
 
   /**
-   * Records a token's jti as revoked until {@code expiresAt}. Idempotent: revoking a jti that is
-   * already present just refreshes its retained expiry. This touches only the token's own row, so
-   * concurrent revocations of different sessions do not contend.
+   * Records a token's jti as revoked. A {@code null} {@code expiresAt} means the token has no
+   * expiry, so its revocation is permanent; otherwise the row is kept until {@code expiresAt}. This
+   * touches only the token's own row, so concurrent revocations of different sessions do not
+   * contend.
+   *
+   * <p>Idempotent under concurrency: if the same token is revoked twice at once, one insert wins
+   * and the other fails on the jti primary key; that is treated as success, since the token is
+   * revoked either way. The expiry is derived from the token, so both inserts carry the same value
+   * and there is nothing to reconcile.
    *
    * <p>After the revocation commits, an asynchronous, best-effort cleanup of expired rows is
    * triggered in a separate transaction. It is deliberately fire-and-forget: a failed sweep (e.g.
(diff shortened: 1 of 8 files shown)


server/src/main/java/io/unitycatalog/server/service/AuthService.java
```diff
@@ -205,7 +212,25 @@ public HttpResponse grantToken(
...
+    if (decodedJWT != null
+        && JwtTokenType.ACCESS
+            .name()
+            .equals(decodedJWT.getClaim(JwtClaim.TOKEN_TYPE.key()).asString())) {
+      // Retain the denylist entry until the token would have expired anyway. Access tokens always
+      // carry an exp; fall back to the configured access-token lifetime if one is somehow absent.
+      Date expiresAt =
+          decodedJWT.getExpiresAt() != null
+              ? decodedJWT.getExpiresAt()
+              : Date.from(Instant.now().plus(serverProperties.getAccessTokenTimeout()));
```
timothyw553: > fall back to the configured access-token lifetime if one is somehow absent.

but `null` could mean no expiry right? if that's the case then tokens with no expiry will remain and continue working.
```diff  
- Date expiresAt =
-     decodedJWT.getExpiresAt() != null
-         ? decodedJWT.getExpiresAt()
-         : Date.from(Instant.now().plus(serverProperties.getAccessTokenTimeout()));
+ // null means this token never expires, so its revocation must be permanent.
+ Date expiresAt = decodedJWT.getExpiresAt();
```

server/src/main/java/io/unitycatalog/server/persist/TokenRevocationRepository.java
```diff
@@ -0,0 +1,101 @@
...
+    TransactionManager.executeWithTransaction(
+        sessionFactory,
+        session -> {
+          TokenRevocationDAO existing = session.get(TokenRevocationDAO.class, jti);
+          if (existing == null) {
+            session.persist(TokenRevocationDAO.builder().jti(jti).expiresAt(expiresAt).build());
+          } else {
+            // existing is managed, so the updated expiry is flushed on commit.
+            existing.setExpiresAt(expiresAt);
+          }
```
timothyw553: so to my understanding: someone logs out -> we need to revoke their credentials -> when we revoke, we check this TokenRevocationDAO table and add that credential to the table. future API request checks said table. if its in it, DENIED

not a blocker, but suppose two logouts happen at the same time. both check the table before either adds the row. both try to add the same token -- one insert succeeds, while the other fails with HTTP 500.

timothyw553: actually maybe the simplest solution is just to do an upsert: Insert the token ID. If it already exists, update it instead.
something like:
```diff
 session -> {
-  TokenRevocationDAO existing =
-      session.get(TokenRevocationDAO.class, jti);
-  if (existing == null) {
-    session.persist(
-        TokenRevocationDAO.builder()
-            .jti(jti)
-            .expiresAt(expiresAt)
-            .build());
-  } else {
-    existing.setExpiresAt(expiresAt);
-  }
+  MutationQuery upsert =
+      session.createMutationQuery(
+          "INSERT INTO TokenRevocationDAO (jti, expiresAt) "
+              + "VALUES (:jti, :expiresAt) "
+              + "ON CONFLICT (jti) DO UPDATE "
+              + "SET expiresAt = excluded.expiresAt");
+  upsert.setParameter("jti", jti);
+  upsert.setParameter("expiresAt", expiresAt);
+  upsert.executeUpdate();
+
   return null;
 }
```

timothyw553 (commented):
two comments.

server/src/main/java/io/unitycatalog/server/persist/TokenRevocationRepository.java
yili-db: HQL on conflict is a Hibernate 7.0 feature — 6.5 doesn't have it yet.
I am fixing it using a try-catch and read back method.

timothyw553:
@yili-db there is a conflict

timothyw553:
@yili-db bump on the merge conflict here

PR #1787, opened 2026-08-18
Make CreateTable.columns optional for metric views

openinx:
**PR Checklist**

**Description of changes**

`CreateTable.columns` was marked required in the OpenAPI spec, so the generated Python client rejected metric-view creates that omit `columns` (Pydantic `Field required`), even though the REST API accepts those requests — metric views derive schema from `view_definition` (dimensions/measures), not an explicit column list.

Changes:
- Remove `columns` from `CreateTable.required` in `api/all.yaml` and document when it is optional vs expected.
- Tolerate a null `columns` list in `TableRepository.createTableImpl` so clients that omit the field do not NPE.
- Java: extend `BaseMetricViewCRUDTest` / `SdkMetricViewCRUDTest` to assert empty columns on create/get, plus parameterized coverage for `columns=null` and `columns=[]`.
- Python: add metric-view create/get/list/delete coverage in `clients/python/tests/test_table.py` for omitted and empty `columns`, plus negative cases for missing `view_definition` / `view_dependencies`.

Regenerating the Python client (`pythonClient/generate`) produces `Optional[List[ColumnInfo]]` for `columns`.


clients/python/tests/test_table.py
```diff
@@ -101,3 +105,121 @@ async def test_table_create(tables_api):
 
     finally:
         await tables_api.delete_table("unity.default.mytable")
+
+
+_METRIC_VIEW_DEFINITION = """\
```
openinx: we will suggest to make it to be two separate `test_metric_view` and `test_view` py files, so that we can cover both view and metric view for this change.

openinx: Done. Split into two files: `clients/python/tests/test_metric_view.py` (metric view create with omitted / empty columns, plus negative cases for missing `view_definition` / `view_dependencies`) and `clients/python/tests/test_view.py` (plain SQL view create with omitted columns and with/without `view_dependencies`). Both cover the optional-columns change. Reverted `test_table.py` back to its original table cases.

yili-db (approved):

yili-db (changes requested):
It seems that it's required by Spark.

openinx:
Thanks Yi,   We had a double-check with the metric view.  the `columns` does a required field for the entire metric view lifecycle.  let's close this PR.

PR #1789, opened 2026-08-19
Do not send vended credential properties to the server in createTable

shenavaa:
UCSingleCatalog injects vended filesystem credentials (fs.s3a.access.key,  fs.s3a.secret.key, fs.s3a.session.token, fs.s3a.aws.credentials.provider,  fs.unitycatalog.*, plus their option.-prefixed duplicates) into the table  property map so the local Spark session can write to the table location.
  UCProxy.createTable forwarded these properties to the server as ordinary  table properties: it only filtered V2_TABLE_PROPERTIES. This breaks  external table creation on S3 whenever credential vending is configured,
  the ~1100-char STS session token overflows the server's
  uc_properties.property_value VARCHAR(255) column and createTable fails:

    io.unitycatalog.client.ApiException: createTable call failed with: 500 -
    {"error_code":"INTERNAL","message":"Error creating table: ... Value too
    long for column \"PROPERTY_VALUE CHARACTER VARYING(255)\" ..."}

  It is also a security issue: temporary AWS credentials were being  persisted into the catalog database as table metadata.

  Fix: strip fs.* and option.fs.* keys from the properties sent to the  server. They are session-scoped Hadoop configs for the local write path,  not table metadata.

**PR Checklist**

**Description of changes**

strip fs.* and option.fs.* keys from the properties sent to the server. They are session-scoped Hadoop configs for the local write path, not table metadata.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -916,8 +916,14 @@ private[spark] class UCProxy(
     createTable.setColumns(columns)
     createTable.setDataSourceFormat(convertDatasourceFormat(format))
     // Do not send the V2 table properties as they are made part of the `createTable` already.
-    val propertiesToServer =
-      properties.view.filterKeys(!UCTableProperties.V2_TABLE_PROPERTIES.contains(_)).toMap
+    // Also strip the vended filesystem credential properties (fs.* and their option.-prefixed
+    // duplicates, e.g. fs.s3a.session.token): they are session-scoped Hadoop configs injected by
+    // UCSingleCatalog for the local write path, not table metadata, and must never be persisted
+    // in the catalog.
+    val propertiesToServer = properties.view
+      .filterKeys(!UCTableProperties.V2_TABLE_PROPERTIES.contains(_))
+      .filterKeys(k => !k.startsWith("fs.") && !k.startsWith(TableCatalog.OPTION_PREFIX + "fs."))
+      .toMap
     createTable.setProperties(propertiesToServer)
     try {
       tablesApi.createTable(createTable)
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
@@ -16,12 +16,15 @@ import java.util.Optional;
 import java.util.stream.Stream;
 import lombok.SneakyThrows;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.spark.sql.connector.catalog.TableCatalog;
 import org.apache.spark.sql.types.DataTypes;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledIf;
 import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 public abstract class ExternalTableReadWriteTest extends BaseTableReadWriteTest {
   @TempDir protected File dataDir;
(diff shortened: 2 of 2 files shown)


openinx:
Hi @shenavaa , thanks for the reporting this and help us to cover this. for context,  I remembered we do have a PR to remove those `option.fs.*` and `fs.*` for those delta tables. https://github.com/delta-io/delta/pull/6300

Seems like your case is for non-delta table, right ?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -916,8 +916,14 @@ private[spark] class UCProxy(
...
     // Do not send the V2 table properties as they are made part of the `createTable` already.
-    val propertiesToServer =
-      properties.view.filterKeys(!UCTableProperties.V2_TABLE_PROPERTIES.contains(_)).toMap
+    // Also strip the vended filesystem credential properties (fs.* and their option.-prefixed
+    // duplicates, e.g. fs.s3a.session.token): they are session-scoped Hadoop configs injected by
+    // UCSingleCatalog for the local write path, not table metadata, and must never be persisted
+    // in the catalog.
+    val propertiesToServer = properties.view
+      .filterKeys(!UCTableProperties.V2_TABLE_PROPERTIES.contains(_))
+      .filterKeys(k => !k.startsWith("fs.") && !k.startsWith(TableCatalog.OPTION_PREFIX + "fs."))
```
openinx, line 925: This fix seems correct,  but I will suggest to add an end-to-end integration tests for this, so that we can clearly see there is no `fs.*` props propagated to the uc server. 

1. create the table via spark sql. 
2. use the sdk to access the table on the uc server via API directly. 
3. assert there is no `fs.*` and `options.fs.*` props in those returned table info, maybe also no `V2_TABLE_PROPERTIES`. 

If we already have a test coverage which cover the `V2_TABLE_PROPERTIES`, then perhaps we can just add the assertion in that test.

yili-db: The check can be added here:
https://github.com/unitycatalog/unitycatalog/blob/7b29af1a3e882730e728f2ac9f4733d7f715a8ad/connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java#L120

shenavaa: Thanks @openinx for the pointer to delta-io/delta#6300. I re-tested to confirm exactly how it interacts with this fix. My original report was actually a Delta table (CREATE TABLE ... USING delta LOCATION 's3://...'), and I retested with a Spark distribution that bundles Delta 4.0, which predates #6300. Verified matrix, same (unpatched) connector build and same UC server with per-bucket credential vending:
`
| Delta on classpath | Format  | createTable                                                                  |
|--------------------|---------|------------------------------------------------------------------------------|
| 4.3.1 (has #6300)  | delta   | ✅ succeeds, no fs.* / option.fs.* stored on the server                      |
| 4.0 (pre-#6300)    | delta   | ❌ 500 Value too long for PROPERTY_VALUE VARCHAR(255) at UCProxy.createTable |
| 4.3.1              | parquet | ❌ same 500 — DeltaTableV2 isn't on the non-Delta path                       |`

So you're right and it is already covered, and the remaining exposure is non-Delta formats. But engines bundling older Delta (several distros still ship ≤ 4.2) hit this on Delta tables too. 

This PR strips the credentials at the source and it will covers both cases regardless of which version, and also stops credentials from ever being sent and persisted.

Agreed on the end-to-end test. I'll add assertions in ExternalTableReadWriteTest as @yili-db suggested: create via Spark SQL (delta and parquet), fetch the table via the SDK, and assert the returned properties contain no fs.*, no option.fs.*, and no V2_TABLE_PROPERTIES keys.

shenavaa: Added the end-to-end test in ExternalTableReadWriteTest as suggested (976fda5): 

create the external table via Spark SQL, fetch it back through the SDK, assert the server-stored properties contain no fs.*, no option.fs.*, and no V2_TABLE_PROPERTIES keys. It's parameterized over cloudParameters (file/s3/gs/abfs, incl. credential-renewal and cred-scoped-fs variants) and lives in the abstract class, so it runs for both the Delta and Parquet suites.

 On the delta#6300 question, I re-tested to confirm the interaction. My original repro was a Delta table, but on a Spark + Delta 4.0.  Delta 4.3.1 + parquet fails identically, since the non-Delta path never goes through DeltaTableV2. So #6300 covers current-Delta delta tables, and this PR covers non-Delta formats plus engines shipping older Delta and filters the credentials at the source, in the same connector code that injects them.

openinx (approved):
Looks good to me now. Thank you for the contribution, @shenavaa  ! and also thanks @yili-db for the reviewing.  I will merge it.

openinx:
Perhaps it's worth to take a look at the DCO github check. https://github.com/unitycatalog/unitycatalog/pull/1789/checks?check_run_id=96674579316,  thank you !

yili-db:
> Perhaps it's worth to take a look at the DCO github check. https://github.com/unitycatalog/unitycatalog/pull/1789/checks?check_run_id=96674579316, thank you !

It's OK. It's just `git commit -s` so I can override this one. It doesn't matter when merged.

PR #1791, opened 2026-08-19
[Server] List every namespace and table from the Iceberg REST endpoints

### LINKED ISSUE #1790: Iceberg REST listNamespaces and listTables silently return only the first 100 entries
**Describe the bug**

`GET /v1/{prefix}/namespaces` and `GET /v1/{prefix}/namespaces/{namespace}/tables` return at most 100 entries and give the client no way to learn that more exist.

Both ask the repository for `Integer.MAX_VALUE` results, intending to fetch everything:

```java
// IcebergRestCatalogService.java:127
.listSchemas(catalog, Optional.of(Integer.MAX_VALUE), Optional.empty())
// IcebergRestCatalogService.java:263
catalog, namespace, Optional.of(Integer.MAX_VALUE), Optional.empty(), false, false);
```

but `PagedListingHelper` caps the page size:

```java
public static Integer DEFAULT_PAGE_SIZE = 100;

public static Integer getPageSize(Optional<Integer> maxResults) {
  return maxResults.filter(x -> (x > 0)).map(x -> Math.min(x, DEFAULT_PAGE_SIZE)).orElse(DEFAULT_PAGE_SIZE);
}
```

The `nextPageToken` that comes back is discarded, and the endpoints set no `next-page-token` of their own, so the truncation is silent.

`listTables` is affected more severely. It filters the page it received down to the tables that have uniform Iceberg metadata, and tables are listed in ascending name order. A schema whose first 100 tables are not uniform therefore returns an empty list even though uniform tables exist.

**To Reproduce**

1. Create a catalog with 101 schemas, then `GET /v1/catalogs/{catalog}/namespaces` — 100 namespaces come back and the last one is missing.
2. In one schema, create 100 non-uniform tables that sort before a uniform Iceberg table (`delta_000` … `delta_099`, `uniform_table`), then `GET /v1/catalogs/{catalog}/namespaces/{schema}/tables` — the response is `[]`.

Both reproduce against a running server. As a test, the second case reads:

```
Expecting actual:
  []
to contain exactly (and in same order):
  [uc_testschema.uniform_table]
```

**Expected behavior**

Both endpoints list every namespace in the catalog, and every uniform Iceberg table in the schema.

-   OS: macOS 26.5 (Darwin 25.5.0), JDK 17
-   Unity Catalog: `main` at 958fded, H2 backend. Not environment specific — the page size cap is unconditional.

**Additional context**

The Iceberg REST spec also defines `pageToken` / `pageSize` request parameters and `next-page-token` in the responses, which these endpoints accept and return neither of. Returning the complete listing is enough to be spec compliant: the spec says a server that does not support pagination should ignore `pageToken` and return all results in a single response with `next-page-token` omitted. I would like to fix the truncation on that basis first, and I have the fix and tests ready to submit.

Supporting client-driven pagination would be a reasonable follow-up, but it needs a decision beyond these endpoints:

- `pageSize` cannot be honoured above 100 without changing `PagedListingHelper.getPageSize`, which every Unity Catalog list API shares.
- For `listTables`, the `next-page-token` would have to be the repository cursor, i.e. the last row of the unfiltered page, rather than the last identifier returned. Using the latter would skip the tables in between, which is the same class of silent loss as this bug.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1790.

`listNamespaces` and `listTables` asked the repositories for `Integer.MAX_VALUE` results, intending to fetch everything, but `PagedListingHelper.getPageSize` caps the page size at `DEFAULT_PAGE_SIZE` (100). Both endpoints therefore returned at most one page, discarded the `nextPageToken` that came with it, and returned no `next-page-token` of their own, so the truncation was invisible to the client.

`listTables` was affected more severely: it filters the page it received down to the tables that have uniform Iceberg metadata, and tables are listed in ascending name order, so a schema whose first 100 tables are not uniform returned an empty list even though uniform tables existed.

**Change**

Both endpoints now follow the repository's page token to the end.

The response shape is unchanged. Per the Iceberg REST spec, a server that does not support pagination should ignore `pageToken` and return all results in a single response with `next-page-token` omitted, which is what these endpoints do, so nothing changes for clients other than the missing entries appearing.

`pageToken` / `pageSize` passthrough is deliberately left out of this PR. The issue describes why it needs a separate decision: `pageSize` cannot be honoured above 100 without changing the shared `PagedListingHelper`, and for `listTables` the token would have to expose the repository cursor rather than the last identifier returned. Happy to follow up on that if maintainers want it.

**Testing**

Two tests added to `IcebergRestCatalogTest`, both confirmed to fail before the change and pass after it:

- `testListNamespacesReturnsEveryNamespace` — one namespace more than a single page. Before the change the last namespace is missing.
- `testListTablesReturnsTablesBeyondTheFirstPage` — a full page of non-uniform tables ahead of one uniform table. Before the change the response is `[]`.

Both derive their size from `PagedListingHelper.DEFAULT_PAGE_SIZE` rather than hardcoding 100.

The rest of the server test suite is unaffected: 422 tests, 22 skipped, 398 passing, and 2 failures that fail the same way on an unmodified tree (`CloudCredentialVendorTest.testGenerateGcpTemporaryCredentials`, which expects the absence of GCP credentials, and `SdkModelCRUDTest.testModelCRUD`).

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
@@ -25,6 +25,7 @@ import io.unitycatalog.server.service.iceberg.TableConfigService;
 import io.unitycatalog.server.utils.JsonUtils;
 import io.unitycatalog.server.utils.NormalizedURL;
 import io.unitycatalog.server.utils.ValidationUtils;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -122,19 +123,24 @@ public class IcebergRestCatalogService implements RegisteredService {
       // nested namespaces is not supported, so child namespaces will be empty
       namespaces = Collections.emptyList();
     } else {
-      String respContent =
-          schemaService
-              .listSchemas(catalog, Optional.of(Integer.MAX_VALUE), Optional.empty())
-              .aggregate()
-              .join()
-              .contentUtf8();
-      ListSchemasResponse resp =
-          JsonUtils.getInstance().readValue(respContent, ListSchemasResponse.class);
-      assert resp.getSchemas() != null;
-      namespaces =
-          resp.getSchemas().stream()
-              .map(schemaInfo -> Namespace.of(schemaInfo.getName()))
-              .collect(Collectors.toList());
+      namespaces = new ArrayList<>();
+      // This endpoint returns the whole listing, so follow the repository's page token to the end
+      Optional<String> pageToken = Optional.empty();
+      do {
+        String respContent =
+            schemaService
+                .listSchemas(catalog, Optional.empty(), pageToken)
+                .aggregate()
+                .join()
+                .contentUtf8();
+        ListSchemasResponse resp =
+            JsonUtils.getInstance().readValue(respContent, ListSchemasResponse.class);
+        assert resp.getSchemas() != null;
+        resp.getSchemas().stream()
+            .map(schemaInfo -> Namespace.of(schemaInfo.getName()))
+            .forEach(namespaces::add);
+        pageToken = nextPageToken(resp.getNextPageToken());
+      } while (pageToken.isPresent());
     }
 
     return ListNamespacesResponse.builder().addAll(namespaces).build();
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -258,13 +264,20 @@ public HttpResponse reportMetrics(
...
       throws JsonProcessingException {
-    ListTablesResponse tables =
-        tableRepository.listTables(
-            catalog, namespace, Optional.of(Integer.MAX_VALUE), Optional.empty(), false, false);
+    List<TableInfo> tables = new ArrayList<>();
+    // This endpoint returns the whole listing, so follow the repository's page token to the end
+    Optional<String> pageToken = Optional.empty();
+    do {
+      ListTablesResponse page =
+          tableRepository.listTables(catalog, namespace, Optional.empty(), pageToken, false, false);
```
yili-db: This is a pre-existing problem but worth a simple fix:
The `false, false` means `tableRepository.listTables` will attach columns and properties to the result but they are discarded later. Changing them to `true, true` is an easy win.

yili-db (commented):
Overall this is a clean fix.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
moomindani: Good catch, thank you -- done in c4b044b.

The endpoint reads nothing but `getName()` and `getSchemaName()` off each `TableInfo`, and `TableInfoDAO.toTableInfo` populates both regardless of the `fetchColumns` flag, so omitting the columns and the properties is safe. It also drops the per-table `attachProperties` query that ran for every table on every page.

Verified with the existing `IcebergRestCatalogTest` suite (6 tests, all passing), including `testListTablesReturnsTablesBeyondTheFirstPage`, which pages past a full first page of non-uniform tables.

yili-db (approved):

PR #1794, opened 2026-08-20
[Server] Use IAM partition in AssumeRole session-policy S3 and KMS ARNs

s-maling-celonis:
**PR Checklist**

**Description of changes**

Vended AWS credentials 403 on GovCloud (and would on China) because `AwsPolicyGenerator` hardcodes commercial S3 ARNs in the AssumeRole session policy:

```
arn:aws:s3:::bucket
```

Session permissions are the intersection of the role policy and that session policy. A GovCloud role policy uses `arn:aws-us-gov:s3:::...`, so the commercial ARNs grant nothing even though STS issued keys. The same strings are reused as `kms:EncryptionContext:aws:s3:arn` (from #1774), so KMS is wrong in those partitions too.

**Change**

IAM partition is resolved in this order:

1. Assumed role ARN (`arn:aws-us-gov:iam::...` → `aws-us-gov`) — the session is bound to that role's partition, even if `aws.region` / the STS client is commercial or unset.
2. STS region catalog (`us-gov-west-1` → `aws-us-gov`, `cn-north-1` → `aws-cn`).
3. Commercial `aws`, matching today's hardcoded ARNs, when both are missing or unrecognized.

That partition is applied to:

- S3 `Resource` ARNs
- KMS encryption-context ARNs (same helper, so the two lists cannot drift)
- `kms:ViaService` (`s3.*.amazonaws.com` for `aws` / `aws-us-gov`, `s3.*.amazonaws.com.cn` for `aws-cn`)

The two-arg `generatePolicy(privileges, locations)` overload is unchanged for callers and existing tests.

**Testing**

- [x] `AwsPolicyGeneratorTest` partition helpers, GovCloud/China policy ARNs, role-ARN precedence over region, China `kms:ViaService`
- [x] `CloudCredentialVendorTest.testGovCloudRoleArnUsesAwsUsGovS3ArnsInSessionPolicy` (GovCloud role + commercial master region still emits `arn:aws-us-gov:s3`)
- [x] `sbt "server/testOnly io.unitycatalog.server.service.credential.aws.AwsPolicyGeneratorTest io.unitycatalog.server.service.credential.CloudCredentialVendorTest"` (Java 17) — 48 passed

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialGenerator.java
+++ b/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialGenerator.java
@@ -61,7 +61,7 @@ public interface AwsCredentialGenerator {
     // contains a CredentialDAO and it will assume the role ARN in CredentialDAO instead.
     private final String staticAwsRoleArn;
     // Same region the STS client uses; used when the assumed role ARN does not carry a partition.
-    private final String awsRegion;
+    private final Region awsRegion;
 
     public StsAwsCredentialGenerator(StsClientBuilder builder, S3StorageConfig config) {
       // Get STS region
@@ -89,7 +89,7 @@ public interface AwsCredentialGenerator {
 
       this.stsClient = builder.region(region).credentialsProvider(credentialsProvider).build();
       this.staticAwsRoleArn = config.getAwsRoleArn();
-      this.awsRegion = region.id();
+      this.awsRegion = region;
     }
 
     @Override
--- a/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialGenerator.java
+++ b/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialGenerator.java
@@ -60,6 +60,8 @@ public interface AwsCredentialGenerator {
     // This is the role ARN to assume for the per-bucket config. Otherwise, the CredentialContext
     // contains a CredentialDAO and it will assume the role ARN in CredentialDAO instead.
     private final String staticAwsRoleArn;
+    // Same region the STS client uses; used when the assumed role ARN does not carry a partition.
+    private final String awsRegion;
 
     public StsAwsCredentialGenerator(StsClientBuilder builder, S3StorageConfig config) {
       // Get STS region
@@ -87,6 +89,7 @@ public interface AwsCredentialGenerator {
 
       this.stsClient = builder.region(region).credentialsProvider(credentialsProvider).build();
       this.staticAwsRoleArn = config.getAwsRoleArn();
+      this.awsRegion = region.id();
     }
 
     @Override
@@ -99,7 +102,9 @@ public interface AwsCredentialGenerator {
       // externalId is only from CredentialDAO. Per-bucket config does not need it.
       Optional<String> externalId = awsIamRole.map(AwsIamRoleResponse::getExternalId);
 
-      String awsPolicy = AwsPolicyGenerator.generatePolicy(ctx.getPrivileges(), ctx.getLocations());
+      String awsPolicy =
+          AwsPolicyGenerator.generatePolicy(
+              ctx.getPrivileges(), ctx.getLocations(), roleArn, awsRegion);
       String roleSessionName = "uc-%s".formatted(UUID.randomUUID());
 
       AssumeRoleRequest.Builder roleRequestBuilder =
--- a/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
+++ b/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
@@ -3,6 +3,7 @@ package io.unitycatalog.server.service.credential.aws;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import io.unitycatalog.server.service.credential.CredentialContext;
 import io.unitycatalog.server.utils.NormalizedURL;
@@ -15,6 +16,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
+import software.amazon.awssdk.regions.PartitionMetadata;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.regions.RegionMetadata;
 
 public class AwsPolicyGenerator {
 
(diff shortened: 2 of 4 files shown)


server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -78,11 +82,33 @@ public class AwsPolicyGenerator {
...
+      Set<CredentialContext.Privilege> privileges,
+      List<NormalizedURL> locations,
+      String roleArn,
+      String awsRegion) {
+    String partition = iamPartitionId(roleArn, awsRegion);
     JsonNode policyRoot = loadYaml(POLICY_STATEMENT);
     ArrayNode policyStatement = (ArrayNode) policyRoot.findPath("Statement");
     JsonNode operationsStatement = loadYaml(OPERATION_STATEMENT);
     policyStatement.add(operationsStatement);
     JsonNode kmsStatement = loadYaml(KMS_STATEMENT);
```
yili-db, line 106: In `KMS_STATEMENT` this `"kms:ViaService": "s3.*.amazonaws.com"` is always overwritten. Should remove it then.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -157,6 +183,66 @@ private static void addObjectArn(
...
+        return "aws";
+      }
+      String partitionId = metadata.partition().id();
+      return partitionId == null || partitionId.isBlank() ? "aws" : partitionId;
+    } catch (RuntimeException e) {
+      return "aws";
+    }
+  }
+
+  static String kmsViaService(String partition) {
```
yili-db: private

server/src/test/java/io/unitycatalog/server/service/credential/CloudCredentialVendorTest.java
```diff
@@ -320,4 +320,62 @@ public void testVendCredentialWithExternalLocationCredential() {
...
+      AwsCredentialVendor awsCredentialVendor = new AwsCredentialVendor(serverProperties);
+      credentialsOperations = new CloudCredentialVendor(awsCredentialVendor, null, null);
+
+      vendCredential(S3_PATH, Set.of(CredentialContext.Privilege.SELECT));
+
+      AssumeRoleRequest capturedRequest = requestCaptor.getValue();
+      assertThat(capturedRequest.policy())
+          .contains("arn:aws-us-gov:s3:::gov-bucket")
+          .doesNotContain("arn:aws:s3:::gov-bucket")
+          .contains("kms:ViaService");
```
yili-db: Should check its value to make it complete.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -157,6 +183,66 @@ private static void addObjectArn(
...
+    }
+    return partitionFromRegion(awsRegion);
+  }
+
+  static String partitionFromArn(String arn) {
+    if (arn == null || arn.isBlank()) {
+      return null;
+    }
+    // arn:partition:service:region:account:resource
+    String[] parts = arn.split(":", 6);
```
yili-db: Use `Arn.fromString(roleArn).partition()` instead of split?

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialGenerator.java
```diff
@@ -87,6 +89,7 @@ public StsAwsCredentialGenerator(StsClientBuilder builder, S3StorageConfig confi
 
       this.stsClient = builder.region(region).credentialsProvider(credentialsProvider).build();
       this.staticAwsRoleArn = config.getAwsRoleArn();
+      this.awsRegion = region.id();
```
yili-db: why not save the entire `Region` here and pass it along the call stack so that it doesn't have to call `Region.of(awsRegion)` later?

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -157,6 +183,66 @@ private static void addObjectArn(
...
+    if (awsRegion == null || awsRegion.isBlank()) {
+      return "aws";
+    }
+    try {
+      RegionMetadata metadata = Region.of(awsRegion).metadata();
+      if (metadata == null || metadata.partition() == null) {
+        return "aws";
+      }
+      String partitionId = metadata.partition().id();
+      return partitionId == null || partitionId.isBlank() ? "aws" : partitionId;
```
yili-db: If it is returned as `PartitionMetadata` then it doesn't have to call `PartitionMetadata.of(partition)` later.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
s-maling-celonis: Removed it from `KMS_STATEMENT`. The template now only has the encryption-context key; `kms:ViaService` is set from the partition when the policy is built.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
s-maling-celonis: Done — `kmsViaService` is private. The partition helpers stay package-private so the existing unit tests can still call them.

server/src/test/java/io/unitycatalog/server/service/credential/CloudCredentialVendorTest.java
s-maling-celonis: Updated. The GovCloud vendor test now asserts `kms:ViaService` is `s3.*.amazonaws.com` (GovCloud uses the commercial DNS suffix).

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
s-maling-celonis: Switched to `Arn.fromString(roleArn).partition()`. Malformed ARNs still return null so we fall back to the STS region, then commercial `aws`.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialGenerator.java
s-maling-celonis: Done — `StsAwsCredentialGenerator` now stores the `Region` and passes it into `generatePolicy`, so we don’t call `Region.of` again.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
s-maling-celonis: `iamPartition` / `partitionFromRegion` now return `PartitionMetadata`. S3 ARNs use `.id()` and `kms:ViaService` uses `.dnsSuffix()`. The role-ARN path still calls `PartitionMetadata.of` once after `Arn.fromString`.

yili-db (approved):

PR #1795, opened 2026-08-20
[Server] Drop KMS encryption-context ARNs from STS session policies

s-maling-celonis:
**PR Checklist**

**Description of changes**

#1774 added `kms:Decrypt` / `kms:GenerateDataKey*` so vended credentials can read SSE-KMS objects. It also listed every S3 ARN again under `kms:EncryptionContext:aws:s3:arn`. That duplication makes AssumeRole fail with `Packed policy consumes 100% of allotted space` on long managed-table paths.

This is not GovCloud-specific. STS packed-policy size is a general limit (plaintext cap 2048 characters). GovCloud and China hit it sooner because `arn:aws-us-gov:` / `arn:aws-cn:` are longer than `arn:aws:`, and those ARNs were copied into both S3 `Resource` and the KMS encryption-context list. Commercial AWS overflows too with a long enough path or several locations in one vend call.

**Change**

Keep the KMS actions and `kms:ViaService` (partition-aware DNS suffix from #1794). Drop the encryption-context ARN list. S3 `Resource` / prefix conditions already constrain which objects the session can Get/Put; a session policy cannot grant more than the assumed role already has.

**Testing**

- [x] `AwsPolicyGeneratorTest.testKmsStatementOmitsEncryptionContext`
- [x] `AwsPolicyGeneratorTest.testLongManagedTablePathFitsStsSessionPolicyLimit` (GovCloud ARNs, policy < 1600 chars)
- [x] `sbt "server/testOnly io.unitycatalog.server.service.credential.aws.AwsPolicyGeneratorTest io.unitycatalog.server.service.credential.CloudCredentialVendorTest"` (Java 17) — 47 passed

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
+++ b/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
@@ -22,6 +22,8 @@ import software.amazon.awssdk.regions.Region;
 
 public class AwsPolicyGenerator {
 
+  static final String AWS_PARTITION = "aws";
+
   static final List<String> SELECT_ACTIONS = List.of("s3:GetO*");
   static final List<String> UPDATE_ACTIONS = List.of(
       "s3:GetO*", "s3:PutO*", "s3:DeleteO*", "s3:*Multipart*");
@@ -66,8 +68,6 @@ public class AwsPolicyGenerator {
       Action: []
       Resource:
         - "*"
-      Condition:
-        StringLike: {}
       """;
 
   private static final ObjectMapper JSON_MAPPER = new ObjectMapper();
@@ -75,8 +75,12 @@ public class AwsPolicyGenerator {
 
   // This can support generating a policy across multiple buckets and paths, however, the assumed
   // role the policy is applied to for a scoped-session needs to have access across those buckets
+  /**
+   * Test helper: commercial {@link #AWS_PARTITION} (no role ARN or region). Production always
+   * calls {@link #generatePolicy(Set, List, String, Region)}.
+   */
   @SneakyThrows
-  public static String generatePolicy(
+  static String generatePolicy(
       Set<CredentialContext.Privilege> privileges,
       List<NormalizedURL> locations) {
     return generatePolicy(privileges, locations, null, null);
@@ -102,10 +106,10 @@ public class AwsPolicyGenerator {
     JsonNode operationsStatement = loadYaml(OPERATION_STATEMENT);
     policyStatement.add(operationsStatement);
     JsonNode kmsStatement = loadYaml(KMS_STATEMENT);
-    JsonNode stringLike = kmsStatement.findPath("StringLike");
-    if (stringLike instanceof ObjectNode stringLikeNode) {
-      stringLikeNode.put("kms:ViaService", kmsViaService(partition));
-    }
+    ((ObjectNode) kmsStatement)
+        .putObject("Condition")
+        .putObject("StringLike")
+        .put("kms:ViaService", kmsViaService(partition));
 
     // Add the appropriate S3 and KMS operations for the privileges requested
     ArrayNode actions = (ArrayNode) operationsStatement.findPath("Action");
@@ -209,7 +213,7 @@ public class AwsPolicyGenerator {
   }
 
   private static PartitionMetadata commercialPartition() {
-    return PartitionMetadata.of("aws");
+    return PartitionMetadata.of(AWS_PARTITION);
   }
 
   private static String s3Arn(PartitionMetadata partition, String resource) {
--- a/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
+++ b/server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
@@ -53,24 +53,22 @@ public class AwsPolicyGenerator {
       Resource: []
       """;
 
-  // The condition key S3 populates when it calls KMS on the caller's behalf. Its value is the ARN
-  // of the object being encrypted or decrypted, or the ARN of the bucket when the bucket has S3
-  // Bucket Keys enabled, in which case the data key is shared across the objects in the bucket.
-  static final String KMS_ENCRYPTION_CONTEXT_KEY = "kms:EncryptionContext:aws:s3:arn";
-
   // Unity Catalog doesn't know which KMS key a bucket is configured with, so the resource stays
-  // open and the statement is narrowed by condition instead: to KMS calls made through S3, and to
-  // the S3 ARNs this policy already grants access to. A session policy can only narrow what the
-  // assumed role is already allowed to do, so a role without KMS access still gets none.
+  // open and the statement is narrowed to KMS calls made through S3. A session policy can only
+  // narrow what the assumed role is already allowed to do, so a role without KMS access still
+  // gets none. Encryption-context ARNs are omitted on purpose: they duplicate every S3 resource
+  // in this policy, and on long managed-table paths that blows the STS packed-policy limit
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -66,8 +69,7 @@ public class AwsPolicyGenerator {
       Condition:
         StringLike:
           "kms:ViaService": "s3.*.amazonaws.com"
```
yili-db: Avoiding setting a default here so that the correct value is always set?

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -78,11 +80,33 @@ public class AwsPolicyGenerator {
...
+      String roleArn,
+      String awsRegion) {
+    String partition = iamPartitionId(roleArn, awsRegion);
     JsonNode policyRoot = loadYaml(POLICY_STATEMENT);
     ArrayNode policyStatement = (ArrayNode) policyRoot.findPath("Statement");
     JsonNode operationsStatement = loadYaml(OPERATION_STATEMENT);
     policyStatement.add(operationsStatement);
     JsonNode kmsStatement = loadYaml(KMS_STATEMENT);
+    JsonNode stringLike = kmsStatement.findPath("StringLike");
+    if (stringLike instanceof ObjectNode stringLikeNode) {
```
yili-db: Per https://github.com/unitycatalog/unitycatalog/pull/1795/changes#r3827823743
If the default value is removed, this line can unconditionally add a new `StringLike` node.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -147,14 +161,63 @@ public static String generatePolicy(
...
+    }
+    return partitionFromRegion(awsRegion);
+  }
+
+  static String partitionFromArn(String arn) {
+    if (arn == null || arn.isBlank()) {
+      return null;
+    }
+    // arn:partition:service:region:account:resource
+    String[] parts = arn.split(":", 6);
```
yili-db: Reuse `software.amazon.awssdk.arns.Arn.fromString(arn).partition()`?

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -147,14 +161,63 @@ public static String generatePolicy(
...
+  static String partitionFromRegion(String awsRegion) {
+    if (awsRegion == null || awsRegion.isBlank()) {
+      return "aws";
+    }
+    try {
+      RegionMetadata metadata = Region.of(awsRegion).metadata();
+      if (metadata == null || metadata.partition() == null) {
+        return "aws";
+      }
+      String partitionId = metadata.partition().id();
```
yili-db: Maybe just return the entire `PartitionMetadata` so that downstream doesn't need to do `PartitionMetadata.of` to convert it back?

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialGenerator.java
```diff
@@ -87,6 +89,7 @@ public StsAwsCredentialGenerator(StsClientBuilder builder, S3StorageConfig confi
 
       this.stsClient = builder.region(region).credentialsProvider(credentialsProvider).build();
       this.staticAwsRoleArn = config.getAwsRoleArn();
+      this.awsRegion = region.id();
```
yili-db: Safe the entire `Region` and pass it along so that down stream doesn't need to do `Region.of`?

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -78,11 +80,33 @@ public class AwsPolicyGenerator {
   public static String generatePolicy(
```
yili-db: This is no longer used for prod use. But it's only kept for test use.
Either redirect tests to the 4-arg function and remove this one, or make it clear (or possibly package local) that this is test only.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
```diff
@@ -147,14 +161,63 @@ public static String generatePolicy(
...
+    String[] parts = arn.split(":", 6);
+    if (parts.length < 2 || !"arn".equalsIgnoreCase(parts[0]) || parts[1].isBlank()) {
+      return null;
+    }
+    return parts[1];
+  }
+
+  static String partitionFromRegion(String awsRegion) {
+    if (awsRegion == null || awsRegion.isBlank()) {
+      return "aws";
```
yili-db: `"aws"` appears many times in prod code and test code. Worth make it a constant.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
s-maling-celonis: Yes — no default in the template. `kms:ViaService` is always set from the partition when the policy is built.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
s-maling-celonis: Done in #1794; this branch now uses `Arn.fromString(roleArn).partition()`.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsCredentialGenerator.java
s-maling-celonis: Done in #1794; `StsAwsCredentialGenerator` stores the `Region` and passes it through.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
s-maling-celonis: Done in #1794; `iamPartition` / `partitionFromRegion` return `PartitionMetadata`.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
s-maling-celonis: Dropped `Condition`/`StringLike` from the YAML. The code now always puts `Condition.StringLike.kms:ViaService`.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
s-maling-celonis: Made package-private and documented as the test/commercial fallback. Production still uses the 4-arg form.

server/src/main/java/io/unitycatalog/server/service/credential/aws/AwsPolicyGenerator.java
s-maling-celonis: Added `AWS_PARTITION` and used it in `commercialPartition()` and the partition-id assertions. CsvSource rows still use the literal `"aws"`.

yili-db:
https://github.com/unitycatalog/unitycatalog/pull/1794 is merged. Please rebase

s-maling-celonis:
Thanks @yili-db — rebased onto `main` now that #1794 is merged.

yili-db (approved):

PR #1797, opened 2026-08-21
[Server] Type Iceberg REST errors in Iceberg's vocabulary

### LINKED ISSUE #1796: Iceberg REST errors raised by Unity Catalog are typed "BaseException", so clients read a missing namespace as a missing table
**Describe the bug**

On the Iceberg REST endpoints, every error that comes from Unity Catalog's own code is rendered with `"type": "BaseException"`. That is the name of an internal Java class, and it is not a type any Iceberg client recognizes. Errors raised by Iceberg's own code are typed correctly, so the gap only shows on the Unity Catalog paths: a catalog, schema, or table that does not exist, or a name Unity Catalog rejects.

| request | status | `type` |
| --- | --- | --- |
| `GET /namespaces` (catalog missing) | 404 | `BaseException` |
| `GET /namespaces/{ns}` (schema missing) | 404 | `BaseException` |
| `GET /namespaces/{ns}/tables` (schema missing) | 404 | `BaseException` |
| `GET /namespaces/{ns}/tables/{table}` (schema missing) | 404 | `BaseException` |
| `GET /namespaces/{ns}/tables/{table}` (table missing) | 404 | `BaseException` |
| `POST /namespaces/{ns}/tables/{table}/metrics` (table missing) | 404 | `BaseException` |
| `GET /namespaces/{ns}` (multi-level name) | 400 | `BaseException` |
| `GET /namespaces/{ns}/tables/{table}` (table has no Iceberg metadata) | 404 | `NoSuchTableException` |

The type is not decoration. Iceberg's client reads it to tell one 404 from another: `ErrorHandlers.TableErrorHandler` raises `NoSuchNamespaceException` only when the type says `NoSuchNamespaceException`, and falls back to `NoSuchTableException` otherwise. With `BaseException` on the wire the client cannot tell a wrong namespace from a wrong table, and reports the wrong one.

**To Reproduce**

1. Start the server: `bin/start-uc-server`
2. Ask for a table in a schema that does not exist:

```
$ curl -s 'http://localhost:8080/api/2.1/unity-catalog/iceberg/v1/catalogs/unity/namespaces/noSuchSchema/tables/t'
{"error":{"message":"Schema not found: unity.noSuchSchema","type":"BaseException","code":404}}
```

3. Make the same request through Iceberg's own client:

```java
RESTCatalog catalog = new RESTCatalog();
catalog.initialize(
    "uc",
    Map.of("uri", "http://localhost:8080/api/2.1/unity-catalog/iceberg", "warehouse", "unity"));
catalog.loadTable(TableIdentifier.of(Namespace.of("noSuchSchema"), "t"));
```

```
org.apache.iceberg.exceptions.NoSuchTableException: Schema not found: unity.noSuchSchema
```

The schema is what is missing, but the client raises `NoSuchTableException`, because `BaseException` is not the type it looks for.

**Expected behavior**

Every error the Iceberg REST endpoints return names a type from Iceberg's vocabulary, whichever code raised it: `NoSuchNamespaceException` for a catalog or schema that does not exist, `NoSuchTableException` for a table that does not exist, `BadRequestException` for a name that is rejected, and so on. No internal Unity Catalog class name reaches the wire.

-   Unity Catalog: `main` at 91af0c9, H2 backend, verified with Iceberg client 1.11.0. Not environment specific.

**Additional context**

`ErrorCode` already carries the Delta dialect's error type for every code (`DeltaErrorType`, from #1772). The Iceberg dialect has no such mapping and falls back to the exception's Java class name, which is right for an exception Iceberg threw and wrong for Unity Catalog's own `BaseException`: it is not an Iceberg exception, and its cause chain ends at itself, so `IcebergRestExceptionHandler.getOriginalException` hands back the wrapper it was meant to look past. I have a fix and tests ready to submit.

Two adjacent spec deviations turned up while verifying this and are not part of this report: `GET /namespaces?parent=` answers 200 with an empty list for a parent that does not exist, where the spec calls for 404, and `HEAD /namespaces/{ns}/tables/{table}` answers 200 where the spec calls for 204. Happy to file those separately.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1796.

`IcebergRestExceptionHandler` names the error type after the original exception's Java class. For an exception Iceberg raised that is exactly right, and `getOriginalException` walks the cause chain so a wrapped one still names itself. Unity Catalog's own `BaseException` is neither: it is not an Iceberg exception, and when it is thrown directly it has no cause to walk to, so the walk returns the wrapper and every error that starts in Unity Catalog code -- a catalog, schema, or table that does not exist, a name that fails validation -- went out as `"type": "BaseException"`.

Iceberg's client switches on that field. `ErrorHandlers.TableErrorHandler` raises `NoSuchNamespaceException` on a 404 only when the type names it, and `NoSuchTableException` otherwise, so a request for a table in a schema that does not exist came back to the client as a missing table.

Name those errors from their error code instead, mapping each code to the Iceberg exception the client raises for that code's HTTP status. Exceptions that did not come from Unity Catalog keep naming themselves, so nothing changes for the errors Iceberg raises. The mapping is an exhaustive switch, so a new error code has to name its Iceberg counterpart rather than silently inherit a misleading one.

| request | status | `type` before | `type` after |
| --- | --- | --- | --- |
| `GET /namespaces` (catalog missing) | 404 | `BaseException` | `NoSuchNamespaceException` |
| `GET /namespaces/{ns}` (schema missing) | 404 | `BaseException` | `NoSuchNamespaceException` |
| `GET /namespaces/{ns}/tables` (schema missing) | 404 | `BaseException` | `NoSuchNamespaceException` |
| `GET /namespaces/{ns}/tables/{table}` (schema missing) | 404 | `BaseException` | `NoSuchNamespaceException` |
| `GET /namespaces/{ns}/tables/{table}` (table missing) | 404 | `BaseException` | `NoSuchTableException` |
| `POST /namespaces/{ns}/tables/{table}/metrics` (table missing) | 404 | `BaseException` | `NoSuchTableException` |
| `GET /namespaces/{ns}` (multi-level name) | 400 | `BaseException` | `BadRequestException` |
| `GET /namespaces/{ns}/tables/{table}` (no Iceberg metadata) | 404 | `NoSuchTableException` | `NoSuchTableException` |

Statuses and messages are untouched, so a client that reads only those sees no change.

Verified against a running server with Iceberg's own `RESTCatalog` 1.11.0: `loadTable` on a table in a schema that does not exist raised `NoSuchTableException: Schema not found: ...` before this change and raises `NoSuchNamespaceException: Schema not found: ...` after it. The new test fails on `main` and passes with the fix; the rest of the Iceberg suite is unchanged.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
+++ b/server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
@@ -8,15 +8,20 @@ import lombok.SneakyThrows;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.exceptions.BadRequestException;
 import org.apache.iceberg.exceptions.CommitFailedException;
+import org.apache.iceberg.exceptions.ForbiddenException;
 import org.apache.iceberg.exceptions.NamespaceNotEmptyException;
 import org.apache.iceberg.exceptions.NoSuchNamespaceException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
 import org.apache.iceberg.exceptions.NoSuchViewException;
+import org.apache.iceberg.exceptions.NotAuthorizedException;
+import org.apache.iceberg.exceptions.NotFoundException;
+import org.apache.iceberg.exceptions.RESTException;
+import org.apache.iceberg.exceptions.ServiceFailureException;
 import org.apache.iceberg.rest.responses.ErrorResponse;
 
 /**
  * Exception handler for the Iceberg REST Catalog API. Formats errors using Iceberg's {@link
- * ErrorResponse} with the original exception's class name as the error type (e.g.,
+ * ErrorResponse}, whose error type is named in Iceberg's own vocabulary (e.g.,
  * "NoSuchTableException"). Uses {@link IcebergObjectMapper} for Iceberg-specific serialization.
  */
 public class IcebergRestExceptionHandler extends BaseExceptionHandler {
@@ -37,6 +42,64 @@ public class IcebergRestExceptionHandler extends BaseExceptionHandler {
       current = current.getCause();
     }
     return current;
+  }
+
+  /**
+   * Names the error type in the vocabulary the Iceberg REST spec and its clients use.
+   *
+   * <p>An exception that came from Iceberg already carries such a name, so it is reported as it
+   * stands. A {@link BaseException} raised by Unity Catalog itself does not: reporting its class
+   * name would put "BaseException" on the wire, which is both an internal name and a type no client
+   * recognizes. Iceberg's own client reads the type to tell one 404 from another -- {@code
+   * ErrorHandlers.TableErrorHandler} raises {@code NoSuchNamespaceException} only when the type
+   * says so, and {@code NoSuchTableException} otherwise -- so an unrecognized type makes a missing
+   * catalog or schema surface at the client as a missing table. Those exceptions are named from
+   * their error code instead, which is also what the response status is derived from.
+   *
+   * <p>{@link #toBaseException} already re-raises four codes as the Iceberg exception that stands
+   * for them, which is what moves the two already-exists codes off their legacy 400; those arrive
+   * here carrying an Iceberg exception and keep naming themselves. This names what that switch does
+   * not reach, so no error code can reach a client as "BaseException".
+   */
+  private static String errorType(BaseException exception) {
+    Throwable original = getOriginalException(exception);
+    if (original instanceof BaseException) {
+      return icebergExceptionFor(exception.getErrorCode()).getSimpleName();
+    }
+    return original.getClass().getSimpleName();
+  }
+
+  /**
+   * The Iceberg exception that stands for a Unity Catalog error code, chosen as the one Iceberg's
+   * client raises for that code's HTTP status; {@code RESTException} covers the statuses the client
+   * has no exception of its own for. The switch is exhaustive so that a new error code has to name
+   * its Iceberg counterpart rather than silently inherit a misleading one.
+   */
+  private static Class<? extends Exception> icebergExceptionFor(ErrorCode errorCode) {
+    return switch (errorCode) {
+      case CATALOG_NOT_FOUND, SCHEMA_NOT_FOUND -> NoSuchNamespaceException.class;
+      case TABLE_NOT_FOUND -> NoSuchTableException.class;
+      case NOT_FOUND -> NotFoundException.class;
+      case ALREADY_EXISTS,
+          RESOURCE_ALREADY_EXISTS,
+          CATALOG_ALREADY_EXISTS,
+          SCHEMA_ALREADY_EXISTS,
+          TABLE_ALREADY_EXISTS,
+          STORAGE_CREDENTIAL_ALREADY_EXISTS,
+          EXTERNAL_LOCATION_ALREADY_EXISTS ->
+          AlreadyExistsException.class;
+      case COMMIT_VERSION_CONFLICT, UPDATE_REQUIREMENT_CONFLICT -> CommitFailedException.class;
+      // ABORTED is the generic abort, raised by callers that have nothing to do with a commit
+      // (token verification, model registration), so it must not claim a commit failed.
+      case ABORTED -> RESTException.class;
+      case INVALID_ARGUMENT, UNSUPPORTED_TABLE_FORMAT, FAILED_PRECONDITION, OUT_OF_RANGE ->
+          BadRequestException.class;
+      case UNAUTHENTICATED -> NotAuthorizedException.class;
+      case PERMISSION_DENIED -> ForbiddenException.class;
+      case UNIMPLEMENTED -> UnsupportedOperationException.class;
+      case INTERNAL, DATA_LOSS, COMMIT_STATE_UNKNOWN -> ServiceFailureException.class;
+      case RESOURCE_EXHAUSTED -> RESTException.class;
+    };
   }
 
   @Override
@@ -79,6 +142,15 @@ public class IcebergRestExceptionHandler extends BaseExceptionHandler {
     if (cause instanceof BadRequestException) {
       return wrapException(ErrorCode.INVALID_ARGUMENT, cause);
     }
+    if (cause instanceof IllegalArgumentException) {
+      // Iceberg's own request parsers reject a request they cannot read with
+      // IllegalArgumentException, and their message is already in Iceberg's terms; only the name is
+      // not. The status is the 400 the base handler would have chosen for it either way.
+      String message = cause.getMessage() != null ? cause.getMessage() : "Bad request";
+      BadRequestException badRequest = new BadRequestException("%s", message);
+      badRequest.initCause(cause);
+      return wrapException(ErrorCode.INVALID_ARGUMENT, message, badRequest);
+    }
     return super.toBaseException(cause);
   }
 
@@ -93,7 +165,7 @@ public class IcebergRestExceptionHandler extends BaseExceptionHandler {
             .writeValueAsString(
                 ErrorResponse.builder()
                     .responseCode(status.code())
-                    .withType(getOriginalException(exception).getClass().getSimpleName())
+                    .withType(errorType(exception))
                     .withMessage(exception.getErrorMessage())
                     .build()));
   }
--- a/server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
+++ b/server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
@@ -112,7 +112,7 @@ public class IcebergRestExceptionHandler extends BaseExceptionHandler {
         // status this surface wants, and its Iceberg name comes from the code, so nothing else
         // needs re-raising here.
         case SCHEMA_ALREADY_EXISTS, TABLE_ALREADY_EXISTS ->
-            new BaseException(ErrorCode.ALREADY_EXISTS, baseException.getErrorMessage());
+            wrapException(ErrorCode.ALREADY_EXISTS, baseException.getErrorMessage(), baseException);
         default -> baseException;
       };
     }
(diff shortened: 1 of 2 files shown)


yili-db:

moomindani:
Rebased onto `58d5c7b` (`main`), and reviewed again on top of that rebase.

#1696 landed in the same file and already names four codes -- `SCHEMA_NOT_FOUND`, `TABLE_NOT_FOUND`, the already-exists pair, and `UPDATE_REQUIREMENT_CONFLICT` -- by wrapping them in the Iceberg exception that stands for them, so the schema and table cases from #1796 are fixed on `main`. What its `default -> baseException` leaves is still untyped: measured on `58d5c7b`, `GET /v1/catalogs/noSuchCatalog/namespaces` answers `404 "type": "BaseException"` and `GET .../namespaces/nested.namespace` answers `400 "type": "BaseException"`.

This PR sits on top of that work rather than replacing it. `toBaseException` keeps #1696's switch, which is what moves the already-exists codes off their legacy 400, and the error type is named from the error code only for exceptions that reach `createErrorResponse` still being a `BaseException`. The javadoc now states that division of labour so the two are not read as duplicates. The switch over `ErrorCode` is exhaustive, so a code added later has to name its Iceberg counterpart instead of silently inheriting `BaseException`; `COMMIT_STATE_UNKNOWN`, added since I opened this, is mapped to `ServiceFailureException` for its 500.

Two things changed while re-reviewing the rebase:

- `ABORTED` no longer names `CommitFailedException`. It is the generic abort -- the only callers that raise it are `JwksOperations` (token verification) and `ModelRepository`, neither of which is a commit -- so it names `RESTException`, and only `COMMIT_VERSION_CONFLICT` and `UPDATE_REQUIREMENT_CONFLICT` claim a commit failed.
- A request Iceberg's own parsers reject is now named too. They raise `IllegalArgumentException`, which the base handler already answers 400 for and whose message is already in Iceberg's terms, but whose class name means nothing to a client: `POST .../metrics` with `{"report-type": {"not": "a string"}}` answered `400 "type": "IllegalArgumentException"`. It now answers `BadRequestException`, with the original failure kept underneath so the server's own logs still have it.

The commit message describes the remaining gap rather than the original one. `testErrorsAreNamedInIcebergTerms` covers the missing catalog, the rejected namespace name and the parser rejection; it fails on `58d5c7b` at the first of those and passes with the fix. `server/testOnly io.unitycatalog.server.service.IcebergRestCatalogTest` is green (10/10), and the full `server/test` suite shows no failure attributable to this change.

server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
```diff
@@ -39,6 +44,64 @@ private static Throwable getOriginalException(BaseException exception) {
...
+
+  /**
+   * The Iceberg exception that stands for a Unity Catalog error code, chosen as the one Iceberg's
+   * client raises for that code's HTTP status; {@code RESTException} covers the statuses the client
+   * has no exception of its own for. The switch is exhaustive so that a new error code has to name
+   * its Iceberg counterpart rather than silently inherit a misleading one.
+   */
+  private static Class<? extends Exception> icebergExceptionFor(ErrorCode errorCode) {
+    return switch (errorCode) {
+      case CATALOG_NOT_FOUND, SCHEMA_NOT_FOUND -> NoSuchNamespaceException.class;
```
yili-db, line 81: Since you have this function to convert into Iceberg error types, most of the `toBaseException` which does the same job is now unnecessary. Just the default `super.toBaseException` that simply wraps them into `ErrorCode.INTERNAL` is good enough because this function would unwrap.

server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
```diff
@@ -39,6 +44,64 @@ private static Throwable getOriginalException(BaseException exception) {
...
+   * catalog or schema surface at the client as a missing table. Those exceptions are named from
+   * their error code instead, which is also what the response status is derived from.
+   *
+   * <p>{@link #toBaseException} already re-raises four codes as the Iceberg exception that stands
+   * for them, which is what moves the two already-exists codes off their legacy 400; those arrive
+   * here carrying an Iceberg exception and keep naming themselves. This names what that switch does
+   * not reach, so no error code can reach a client as "BaseException".
+   */
+  private static String errorType(BaseException exception) {
+    Throwable original = getOriginalException(exception);
```
yili-db, line 66: The original exception can be considered 3 types:
1. `BaseException` after all: call `icebergExceptionFor`
2. Iceberg errors: `original.getClass().getSimpleName()`
3. Any other errors that aren't `BaseException` or Iceberg: They should still use the wrapper `BaseException` and call `icebergExceptionFor`

So, I propose something like this:
```
  private static String errorType(BaseException exception) {
    Throwable original = getOriginalException(exception);
    // an exception Iceberg itself raised already carries a name the client knows -- and it may be
    // more specific than the code's mapping (NoSuchViewException, NamespaceNotEmptyException, which
    // toBaseException folds into NOT_FOUND / ALREADY_EXISTS), so keep it.
    if (original.getClass().getName().startsWith("org.apache.iceberg.exceptions.")) {
      return original.getClass().getSimpleName();
    }
    // anything else -- a UC BaseException, or a stray runtime exception -- names itself nothing a
    // client knows, so name it from the error code.
    return icebergExceptionFor(exception.getErrorCode()).getSimpleName();
  }
```

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
```diff
@@ -953,6 +954,75 @@ public void testListTablesReturnsTablesBeyondTheFirstPage()
         .containsExactly(TableIdentifier.of(Namespace.of(TestUtils.SCHEMA_NAME), "uniform_table"));
   }
 
+  @Test
+  public void testErrorsAreNamedInIcebergTerms()
```
yili-db: Existing test already assert on failures like this:
```
    // not setting warehouse param should result in 400 BadRequestException
    resp = client.get("/v1/config").aggregate().join();
    assertThat(resp.status().code()).isEqualTo(400);
```
Can you replace them with `assertErrorType` instead of adding a separate test?

server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
moomindani: Done, with one exception. Three of the four codes it re-raised were only supplying names -- `SCHEMA_NOT_FOUND` and `TABLE_NOT_FOUND` already carry 404 (`ErrorCode.java:34-35`), and `UPDATE_REQUIREMENT_CONFLICT` already carries 409 (line 46), whose code mapping names `CommitFailedException` anyway -- so they are gone, along with the branch that rewrapped the `IllegalArgumentException` Iceberg's parsers raise, which the base handler already answers 400 for.

What is left is the one thing `errorType` cannot do: `SCHEMA_ALREADY_EXISTS` and `TABLE_ALREADY_EXISTS` answer 400 on the native REST surface for backward compatibility (`ErrorCode.java:55-56`, with 409 as their Delta status), and no Iceberg client reads a 400 as a conflict. They are still re-raised as `ALREADY_EXISTS`, but as a plain `BaseException` rather than a wrapped Iceberg exception.

Dropping the rest into `super.toBaseException` would not work, though: the branches below it map the exceptions Iceberg itself raises onto statuses, so a `NoSuchTableException` from the service would go out as 500 instead of 404.

server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
moomindani: Adopted; your version is better than what I had. Keying on the package rather than on "is it still a `BaseException`" also names an exception from a library we happen to call, where mine leaked its class name. The tests cover that now: a metrics body Iceberg's parser rejects raises `IllegalArgumentException` and comes back as `BadRequestException`.

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
moomindani: Done. The separate test is gone; its cases now sit with the endpoints they belong to, and `assertErrorType` replaced the bare status assertions as well as the three places that spelled out the same two assertions by hand. Three cases had no existing home, so they went into `testNamespaces`: a catalog that does not exist, a schema that does not exist, and the dotted name a nested namespace arrives as.

The coverage got stronger rather than just shorter -- on an unfixed tree `testTable`, `testNamespaces` and `testReportMetrics` now fail with "expected BadRequestException but was BaseException" / "was IllegalArgumentException".

server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
```diff
@@ -39,30 +44,75 @@ private static Throwable getOriginalException(BaseException exception) {
...
-            wrapException(
-                ErrorCode.ALREADY_EXISTS,
-                baseException.getErrorMessage(),
-                new AlreadyExistsException("%s", baseException.getErrorMessage()));
-        case UPDATE_REQUIREMENT_CONFLICT ->
-            wrapException(
-                ErrorCode.ALREADY_EXISTS,
-                baseException.getErrorMessage(),
-                new CommitFailedException("%s", baseException.getErrorMessage()));
+            new BaseException(ErrorCode.ALREADY_EXISTS, baseException.getErrorMessage());
```
yili-db: Use `wrapException(ErrorCode.ALREADY_EXISTS, baseException.getErrorMessage(), baseException);` to keep the original stack trace.
This was missed in some lines here before this PR. But please fix it.

moomindani:
Done -- `wrapException` keeps the original exception as the cause, and clears its own trace so `getRelevantStackTrace` falls back to it. Also rebased onto `4ee2b248`; #1799, #1836 and #1847 landing in the same two files is what made this conflict.

I checked the rest of the file for the same omission you mentioned: the three `wrapException` calls below already pass the cause, so this was the only one.

yili-db (approved):

PR #1799, opened 2026-08-22
[Server] List Iceberg tables without resolving each one again

### LINKED ISSUE #1798: Iceberg REST listTables fails with 404 when a table is dropped while it is listing
**Describe the bug**

`GET /v1/{prefix}/namespaces/{namespace}/tables` lists a schema's tables and then, for every table it
just listed, asks the repository whether that table carries UniForm Iceberg metadata
(`getTableUniformMetadataLocation` -> `findTableOrThrow`). The second resolution happens after the
page was read, so a table dropped in between makes the lookup throw `TABLE_NOT_FOUND`, and the whole
listing fails with a 404 that names a table the client never asked about. A listing is a read of a
schema; another session dropping an unrelated table should not be able to fail it.

**To Reproduce**

1. Start the server and create a schema with a few dozen tables.
2. From four threads, request `GET .../namespaces/{schema}/tables` in a loop.
3. From four more threads, repeatedly create and delete a table in the same schema, each thread using
   its own table names.
4. Watch the listings fail:

```
successful listings: 30172
failures: 6641
  x584  404 {"error":{"message":"Table not found: unity.race2.c1_1", ... ,"code":404}}
  x573  404 {"error":{"message":"Table not found: unity.race2.c3_2", ... ,"code":404}}
  ...
```

About 18% of the listings failed over a 20-second run, and every failure named one of the tables
being created and dropped, not anything the request referred to.

**Expected behavior**

The listing returns the uniform tables of the schema. A table created or dropped after a page was
read is simply absent from, or present in, the result; it does not fail the request.

-   Unity Catalog: `main` at c9ed6b1, H2 backend. Not environment specific -- the second resolution
    is unconditional.

**Additional context**

The same second resolution also costs two queries per listed table (a schema lookup and a table
lookup inside `findTableOrThrow`), on top of the query that read the page. Answering whether a table
is uniform from the row the page was already read from removes the failure and that cost together. I
have a fix and a test ready to submit.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1798.

`listTables` on the Iceberg REST endpoint listed a schema's tables and then asked the repository,
once per listed table, whether that table carries UniForm Iceberg metadata. That question was
answered by resolving the table by name again, after the page had been read, so a table dropped in
between made `findTableOrThrow` raise `TABLE_NOT_FOUND` and the endpoint answered 404 -- naming a
table the client never asked about -- instead of listing the schema.

`TableRepository.listUniformIcebergTables` now reports which tables of a page are uniform from the
very rows the page was read from, and the endpoint builds its response out of those pages. Nothing is
resolved twice, so a table created or dropped after a page was read cannot affect that page, and a
page costs one query rather than one plus two per table it holds.

Measured against a running server, four threads listing one schema while four more created and
dropped tables in it for 20 seconds:

| | before | after |
| --- | --- | --- |
| listings that succeeded | 30172 | 50311 |
| listings that failed with 404 | 6641 | 0 |

The response for a quiet schema is unchanged, and so is paging: the listing still follows the
repository's page token to the end, so a page whose tables are all non-uniform contributes nothing
without ending the walk.

`testListTablesToleratesATableDroppedWhileListing` drives 200 listings while a table is created and
dropped alongside them. It fails on `main` (the sixth listing came back 404 in my run) and passes with
this change. The rest of `IcebergRestCatalogTest` passes unchanged, as do the table CRUD suites.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
@@ -25,7 +25,6 @@ import io.unitycatalog.server.model.CreateStagingTable;
 import io.unitycatalog.server.model.CreateTable;
 import io.unitycatalog.server.model.DataSourceFormat;
 import io.unitycatalog.server.model.ListSchemasResponse;
-import io.unitycatalog.server.model.ListTablesResponse;
 import io.unitycatalog.server.model.SchemaInfo;
 import io.unitycatalog.server.model.StagingTableInfo;
 import io.unitycatalog.server.model.TableInfo;
@@ -35,6 +34,7 @@ import io.unitycatalog.server.persist.Repositories;
 import io.unitycatalog.server.persist.SchemaRepository;
 import io.unitycatalog.server.persist.StagingTableRepository;
 import io.unitycatalog.server.persist.TableRepository;
+import io.unitycatalog.server.persist.TableRepository.IcebergTablePage;
 import io.unitycatalog.server.persist.dao.TableInfoDAO;
 import io.unitycatalog.server.persist.model.Privileges;
 import io.unitycatalog.server.service.credential.CredentialContext;
@@ -47,11 +47,9 @@ import io.unitycatalog.server.utils.ServerProperties;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.iceberg.MetadataUpdate;
 import org.apache.iceberg.PartitionSpec;
@@ -76,8 +74,6 @@ import org.apache.iceberg.rest.responses.GetNamespaceResponse;
 import org.apache.iceberg.rest.responses.ListNamespacesResponse;
 import org.apache.iceberg.rest.responses.LoadTableResponse;
 import org.apache.iceberg.rest.responses.LoadViewResponse;
-import org.hibernate.Session;
-import org.hibernate.SessionFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -110,7 +106,6 @@ public class IcebergRestCatalogService extends AuthorizedService implements Regi
   private final SchemaRepository schemaRepository;
   private final StagingTableRepository stagingTableRepository;
   private final TableRepository tableRepository;
-  private final SessionFactory sessionFactory;
 
   @Override
   public ExceptionHandlerFunction exceptionHandler() {
@@ -130,7 +125,6 @@ public class IcebergRestCatalogService extends AuthorizedService implements Regi
     this.schemaRepository = repositories.getSchemaRepository();
     this.stagingTableRepository = repositories.getStagingTableRepository();
     this.tableRepository = repositories.getTableRepository();
-    this.sessionFactory = repositories.getSessionFactory();
   }
 
   // Config APIs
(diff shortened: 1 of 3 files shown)


yili-db:

moomindani:
Rebased onto `58d5c7b` (`main`).

The bug is still there after #1696: `listTables` reads a page of the schema's tables and then calls `getTableUniformMetadataLocation` for each listed name, which goes through `findTableOrThrow`, so a table dropped between the page read and the filter still fails the whole listing. On `58d5c7b` the new test fails with `listing 23 returned {"error":{"message":"Table not found: uc_testcatalog.uc_testschema.churned","type":"NoSuchTableException","code":404}}` where 200 was expected, and passes with the fix.

Two things changed in the rebase:

- The repository method is now `listIcebergTables` returning `IcebergTablePage`, not `listUniformIcebergTables`/`UniformIcebergTablePage`. #1696 made `uniform_iceberg_metadata_location` the metadata pointer for native Iceberg tables as well, so the old name no longer described what the predicate selects. The predicate itself is unchanged, and it is the same one `listTables` applies on `main`.
- `IcebergRestCatalogService` no longer opens a session of its own for this endpoint, so the now-unused `sessionFactory` field and the two Hibernate imports go with it.

`server/testOnly io.unitycatalog.server.service.IcebergRestCatalogTest io.unitycatalog.server.service.IcebergRestCatalogServiceTest` is green (10/10 and 4/4).

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -592,44 +586,20 @@ public HttpResponse reportMetrics(
...
     } while (pageToken.isPresent());
 
-    List<TableIdentifier> filteredTables;
-    try (Session session = sessionFactory.openSession()) {
-      filteredTables =
-          tables.stream()
-              .filter(
-                  tableInfo -> {
-                    String metadataLocation =
-                        tableRepository.getTableUniformMetadataLocation(
```
yili-db, line 619: `getTableUniformMetadataLocation` is no longer used by any code. You can remove it from `tableRepository`

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -592,44 +586,20 @@ public HttpResponse reportMetrics(
...
-              Optional.empty(),
-              pageToken,
-              /* omitProperties= */ true,
-              /* omitColumns= */ true);
-      tables.addAll(Objects.requireNonNull(page.getTables()));
-      pageToken = nextPageToken(page.getNextPageToken());
+      IcebergTablePage page = tableRepository.listIcebergTables(catalog, namespace, pageToken);
+      page.tableNames().stream()
+          .map(tableName -> TableIdentifier.of(Namespace.of(namespace), tableName))
+          .forEach(tables::add);
```
yili-db: Instead of adding to `tables`, it's better to add directly to `ListTablesResponse.builder`.

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
```diff
@@ -953,6 +955,55 @@ public void testListTablesReturnsTablesBeyondTheFirstPage()
         .containsExactly(TableIdentifier.of(Namespace.of(TestUtils.SCHEMA_NAME), "uniform_table"));
   }
 
+  @Test
+  public void testListTablesToleratesATableDroppedWhileListing() throws Exception {
```
yili-db: This test is heavy and non-deterministic. I think it's better to remove it.
Existing test still verifies that the list table functionality is working.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
moomindani: Removed. It existed only for the second resolution this change drops, and nothing else called it.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
moomindani: Done; the loop adds straight into `ListTablesResponse.builder()`, so the intermediate list is gone.

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
moomindani: Removed. Fair: it ran 200 listings against a churn thread to hit a window, which is slow and could only ever fail probabilistically.

One thing to flag: with it gone, this PR has no test that fails without the fix. `testListTablesReturnsTablesBeyondTheFirstPage` covers paging and the uniform filter deterministically, but not the dropped-table window, which cannot be reached without concurrency. The change is structural rather than conditional -- the per-table second resolution is gone, so there is no window left -- and the measurements in the description (6641 of 36813 listings failed before, 0 of 50311 after) stand as the evidence. Happy to add something slimmer if you would rather have a test in the file.

yili-db (approved):

PR #1801, opened 2026-08-22
[Server] Answer three Iceberg REST details the way the spec describes them

### LINKED ISSUE #1800: Three Iceberg REST responses deviate from the spec: parent listing, tableExists status, unreadable body
**Describe the bug**

Three small deviations on the Iceberg REST endpoints, all measured on `main`:

1. `GET /v1/{prefix}/namespaces?parent=<namespace that does not exist>` answers 200 with an empty
   list. The spec answers a listing under a namespace that does not exist with 404, and a client
   cannot otherwise tell "this namespace has no children" from "there is no such namespace".
2. `HEAD /v1/{prefix}/namespaces/{namespace}/tables/{table}` answers 200 for a table that exists,
   where the spec says 204 with no content. The response also advertises a six-byte `text/plain`
   body, which a HEAD response must not carry.
3. A request body the server cannot parse is reported with the Jackson exception's class name and
   the reader's source location in the message, rather than a description of what was wrong.

**To Reproduce**

1. Start the server and create a table with UniForm Iceberg metadata in `unity.default`.
2. List under a parent that does not exist:

```
$ curl -s -w ' [%{http_code}]' '.../iceberg/v1/catalogs/unity/namespaces?parent=noSuchSchema'
{"namespaces":[],"next-page-token":null} [200]
```

3. Ask whether the uniform table exists:

```
$ curl -sI '.../iceberg/v1/catalogs/unity/namespaces/default/tables/marksheet_uniform'
HTTP/1.1 200 OK
content-type: text/plain; charset=utf-8
content-length: 6
```

4. Post an empty metrics body:

```
$ curl -s -X POST -H 'content-type: application/json' -d '' \
    '.../iceberg/v1/catalogs/unity/namespaces/default/tables/marksheet_uniform/metrics'
{"error":{"message":"failed to parse a JSON document: com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input\n at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1]","type":"IllegalArgumentException","code":400}}
```

**Expected behavior**

1. 404 for a parent namespace that does not exist; 200 with an empty list for one that exists, since
   Unity Catalog has no nested namespaces.
2. 204 with no content when the table exists.
3. A message that describes the malformed request without naming the classes of the JSON reader.

-   Unity Catalog: `main` at c9ed6b1, H2 backend. Not environment specific.

**Additional context**

These are the two adjacent deviations I mentioned in #1796, plus the error message beside them. I
have a fix and tests ready to submit.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1800.

Three independent deviations on the Iceberg REST endpoints, each small:

- **Listing under a parent namespace.** `listNamespaces` returned an empty list for any `parent`,
  including one naming a namespace that does not exist, so a client could not tell an empty namespace
  from a missing one. The parent is now resolved before answering. Unity Catalog still has no nested
  namespaces, so a parent that exists keeps answering an empty list; one that does not exist answers
  404, as the spec requires.
- **tableExists status.** `HEAD .../tables/{table}` answered 200, where the spec says 204 with no
  content. It now answers 204, which also drops the six-byte `200 OK` body Armeria described on a
  response that must not carry one. Iceberg's own client accepts either, so this changes nothing for
  it; it does stop a stricter client from seeing a body on a HEAD.
- **Unreadable request body.** The message quoted the Jackson exception: its class, the Java types
  the body was being mapped onto, and the reader's position in the body; and the error type was the
  converter's `IllegalArgumentException`, which no Iceberg client knows. A wrapper around the Iceberg
  surface's request converter now answers one of three messages, typed `BadRequestException`:
  `Malformed request body: no content`, `... not valid JSON`, or `... not the structure this endpoint
  accepts`.

  Recognizing the failure at the converter rather than in the exception handler keeps it to request
  bodies: the same Jackson failures are raised when the server reads JSON of its own, such as a
  table's metadata file, and one of those would otherwise have been answered as a malformed request
  body. The wrapper is on the Iceberg arm alone, since wrapping the shared converter would also
  change what the Delta surface says about an unreadable body, which its clients read today; a
  follow-up for the other dialects is planned.

Verified against a running server:

| request | before | after |
| --- | --- | --- |
| `GET /namespaces?parent=noSuchSchema` | 200, `{"namespaces":[]}` | 404, `Schema not found: unity.noSuchSchema` |
| `GET /namespaces?parent=default` | 200, `{"namespaces":[]}` | unchanged |
| `HEAD .../tables/marksheet_uniform` | 200, `content-length: 6` | 204, no content |
| `POST .../metrics` with an empty body | 400, message names `com.fasterxml.jackson...` | 400, `Malformed request body: no content` |

Three tests cover these: the new `testListNamespacesUnderAParent` and
`testReportMetricsRejectsAnUnreadableBodyWithoutLeakingTheParser`, and the tableExists assertion in
`testTable`, which now expects 204 and an empty body. All three fail on `main` and pass with this
change; the remainder of `IcebergRestCatalogTest`, `SdkSchemaCRUDTest`, and `SdkReportMetricsTest`
pass unchanged.

A `parent` naming a multi-level namespace still answers 400 rather than 404: that path is the
namespace-separator handling in #182, which this PR leaves alone.


yili-db:

moomindani:
Rebased onto `58d5c7b` (`main`), and reviewed again on top of that rebase. All three deviations are still present there, and each is still fixed by the same small change; what moved is the code around them.

- `listNamespaces`: `main` now returns an empty list whenever `parent` is set. The parent is resolved first, via `schemaRepository.getSchema` rather than the `schemaService` HTTP call I used before, so a parent that does not exist is a 404 and a parent that does keeps answering an empty list.
- `tableExists`: rebased onto #1696's `getIcebergTableState` lookup, returning 204 instead of 200. This also updates the HEAD assertion #1696 added in `testIcebergTableWriteLifecycle`, which expects 200 today.
- The unreadable-body case needed more than the rebase. My original change sanitized the message but left the type alone, and the message still quoted the reader. On `58d5c7b` with this branch as it first stood, `POST .../namespaces` with `{"namespace": ["x"], "properties": 5}` answered:

  ```json
  {"error":{"message":"Malformed request body: Cannot deserialize value of type `java.util.LinkedHashMap<java.lang.String,java.lang.String>` from Integer value (token `JsonToken.VALUE_NUMBER_INT`)","type":"IllegalArgumentException","code":400}}
  ```

  The type is the request converter's, which no Iceberg client knows, and the message still names the Java types the body was being mapped onto. It now answers:

  ```json
  {"error":{"message":"Malformed request body: not the structure this endpoint accepts","type":"BadRequestException","code":400}}
  ```

  and a body that is not JSON answers `not valid JSON`. The reader's failure is kept as the cause of the Iceberg exception, so the server's own logs still have the reason. The detection is narrowed to the two Jackson failures that mean "this body cannot be read" (`JsonParseException`, `MismatchedInputException`); it previously matched any `JsonProcessingException` anywhere in the cause chain, which would have reported a failure to *write* a response as a bad request.

On `58d5c7b` four tests in `IcebergRestCatalogTest` fail (`testListNamespacesUnderAParent`, `testReportMetricsRejectsAnUnreadableBodyWithoutLeakingTheParser`, and the two carrying HEAD assertions, `testTable` and `testIcebergTableWriteLifecycle`); with the fix the class is green (11/11), and the full `server/test` suite shows no failure attributable to this change.

The tests pin more than they did: the unreadable-body test asserts the type and code and that the message contains neither the parser's classes, nor `java.util.`, nor `[Source:`, over three bodies (empty, invalid JSON, wrong shape); the parent test also covers an empty `parent`, which is not a parent, so the listing is the whole one, and a catalog that does not exist.

yili-db:
This PR has some overlap with your other PR #1797. Will wait til that one is merged first.

moomindani:
Rebased onto `4ee2b248`. Still waiting on #1797, as you asked.

moomindani:
Rebased onto `81663d1f`, now that #1797 is in.

server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
```diff
@@ -129,9 +132,45 @@ protected BaseException toBaseException(Throwable cause) {
...
+   * Why the request body could not be read, or null if the failure was something else. Only the two
+   * Jackson failures that mean "this body cannot be read" are recognized: a body that is not JSON
+   * ({@link JsonParseException}) and a body whose shape does not fit what the endpoint takes
+   * ({@link MismatchedInputException}). Every other Jackson failure -- writing a response, for one
+   * -- is left to the general handling, which does not call it a bad request.
+   */
+  @Nullable
+  private static String unreadableBodyReason(Throwable cause) {
+    for (Throwable current = cause; current != null; current = current.getCause()) {
+      if (current instanceof JsonParseException) {
```
yili-db: These two exceptions could also be thrown by server when reading other json strings like metadata. It's not necessarily request body.
It can be done inside `ArmeriaServerBuilder.java` to add a wrapper that catches the body parsing error and convert them into an explicit `BaseException`:
```
  public final class MalformedBodyRejectingConverter implements RequestConverterFunction {

    private final RequestConverterFunction delegate;

    public MalformedBodyRejectingConverter(RequestConverterFunction delegate) {
      this.delegate = Objects.requireNonNull(delegate);
    }

    @Override
    public Object convertRequest(
        ServiceRequestContext ctx,
        AggregatedHttpRequest request,
        Class<?> expectedResultType,
        @Nullable ParameterizedType expectedParameterizedResultType)
        throws Exception {
      try {
        return delegate.convertRequest(
            ctx, request, expectedResultType, expectedParameterizedResultType);
      } catch (IllegalArgumentException | JsonProcessingException e) {
        // Armeria's Jackson converter rethrows an unreadable body as IllegalArgumentException with
        // the Jackson failure as its cause; a converter that throws that failure directly is caught
        // by the JsonProcessingException arm. Everything else, including Armeria's
        // FallthroughException signal, is never caught and propagates untouched.
        Throwable reason = e instanceof JsonProcessingException ? e : e.getCause();
        if (reason instanceof JsonParseException) {
          throw new BaseException(
              ErrorCode.INVALID_ARGUMENT, "Malformed request body: not valid JSON");
        }
        if (reason instanceof MismatchedInputException) {
          throw new BaseException(
              ErrorCode.INVALID_ARGUMENT,
              "Malformed request body: not the structure this endpoint accepts");
        }
        throw e;
      }
    }
  }

  private RequestConverterFunction bodyConverter(ObjectMapper mapper) {
    RequestConverterFunction jackson =
        new MalformedBodyRejectingConverter(new JacksonRequestConverterFunction(mapper));
    return authorizationEnabled ? new AuthorizationGateConverter(jackson, mapper) : jackson;
  }

```

moomindani: Moved, thank you -- this was a real false positive: the same Jackson failures come from the server reading JSON of its own, and a corrupt table metadata file would have been answered as "Malformed request body". The converter is the right place because only a body it could not read reaches it.

One deviation from your snippet: the wrapper is on the `ICEBERG` arm of `register` rather than inside `bodyConverter`. Wrapping the shared converter also changes what the Delta surface says about an unreadable body, and `RawCreateTableTest.rawHttpRejectsMalformedColumnRequests` pins that message today (it asserts Jackson's own "Cannot construct instance of `DeltaStructField` ..."). Widening it there looks right to me, but it changes released behaviour on another surface, so I left it out of this PR rather than editing that test.

Also rebased onto `eb52af2e`; the branch predated #1848.

yili-db:
@moomindani Please update the PR message as now the content of PR has changed.

server/src/main/java/io/unitycatalog/server/ArmeriaServerBuilder.java
```diff
@@ -212,7 +220,9 @@ private void register(ServiceProtocol protocol, String relativePath, RegisteredS
     RequestConverterFunction requestConverter =
         switch (protocol) {
           case AUTH, UC, SCIM -> bodyConverter(ucMapper);
-          case ICEBERG -> bodyConverter(icebergMapper);
+          // Only the Iceberg surface rejects an unreadable body in its own words; the other
```
yili-db: Actually I think this wrapper is also needed by other dialect. It's just not done in this PR and that's OK since this PR is focusing on IRC.
Can you please tune this comment like this:
```
  // The Jackson converter leaks its own error text (exception class, mapped types, source
  // position) on any surface. Only on the Iceberg surface is an unreadable body rewritten into
  // a clean message, since its REST spec fixes the error contract clients parse; the other
  // dialects still surface the raw converter message.
```
Later I can do a follow up for other dialect.

moomindani: Done, comment replaced with your wording.

Agreed on the follow-up. While pinning the messages in the test I found one that was wrong for the caller: an empty body raises `MismatchedInputException` too ("No content to map due to end-of-input"), so it was answered "not the structure this endpoint accepts", which sends someone who sent nothing to look at their schema. It now says `Malformed request body: no content`, and the test asserts each of the three messages exactly rather than just the prefix.

Also rebased onto `4e8e52e0` and updated the PR description.

yili-db (approved):

PR #1802, opened 2026-08-24
Follow up on Iceberg REST catalog review comments

yili-db:
**PR Checklist**

**Description of changes**

- Log at warn before swallowing best-effort Iceberg metadata cleanup failures, so a persistent credential or permission problem stays diagnosable instead of being dropped silently.
- Document that uniform_iceberg_metadata_location holds either a Delta UniForm projection or a native Iceberg pointer: the name is a misnomer, a non-null value does not imply a Delta table, and the field must be read together with data_source_format.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
@@ -83,8 +83,11 @@ public class TableInfoDAO extends IdentifiableDAO {
 
   /**
    * Iceberg metadata pointer for either a Delta UniForm projection or a native Iceberg table. The
-   * meaning is distinguished by {@link #dataSourceFormat}: Delta is authoritative for UniForm;
-   * Iceberg is authoritative for native Iceberg tables.
+   * {@code uniform_iceberg_} name predates native Iceberg support and is a misnomer here: a
+   * non-null value does NOT imply a Delta/UniForm table. The owning format is distinguished by
+   * {@link #dataSourceFormat} -- Delta is authoritative for UniForm, Iceberg is authoritative for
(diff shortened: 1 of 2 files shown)


timothyw553 (approved):
this is a follow up to https://github.com/unitycatalog/unitycatalog/pull/1696#discussion_r3845913836. LGTM.

server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
```diff
@@ -83,8 +83,11 @@ public class TableInfoDAO extends IdentifiableDAO {
 
   /**
    * Iceberg metadata pointer for either a Delta UniForm projection or a native Iceberg table. The
-   * meaning is distinguished by {@link #dataSourceFormat}: Delta is authoritative for UniForm;
-   * Iceberg is authoritative for native Iceberg tables.
+   * {@code uniform_iceberg_} name predates native Iceberg support and is a misnomer here: a
```
openinx: Can we rename it to the `icebergMetadataLocation` ?  for now,  I think it's better to rename it for now rather than leaving to some time in future ?

openinx: at least we can rename the class member name to the `icebergMetadataLocation`,  even if we want to keep the original `uniform_iceberg_metadata_location` in the database.

openinx (commented):
Left one comment , others look good to me.  Thanks @yili-db !

openinx (approved):

server/src/main/java/io/unitycatalog/server/persist/dao/TableInfoDAO.java
yili-db: Renamed

Issue #1576, opened 2026-05-28
Unity catalog adds credential check overhead when accessing S3 external table


yhwen:
**Describe the bug**

Every time when Unity Catalog runs a query to access the S3 external table data, it will add a check for the credential for the table access permission. Within the same application / job run, this becomes repeated credential checks among different queries, even for the same table. This leads to un-necessary credential check overhead and adds up to cause the overall application run slowdown. 

A workaround is to create a temp view at the beginning so the credential check becomes a one time set up. All the queries will use the same UC vended credential to access the table data. However, this would require the additional changes to the user application. We need to solution to enable the fast S3 filesystem listing, and avoid the repeated slow credential checks for every query.

**To Reproduce**

1. ...run the query multiple times, there's credential check overhead for each query
2. ...
3. ...

**Expected behavior**

**System [please complete the following information]:**

-   OS: Ubuntu 22.04

**Additional context**

tiennguyen-onehouse:
Since this was filed, the bulk of the overhead has been fixed by the connector-side credential cache (#1586, #1672, backported to 0.5.1, enabled by default): repeated queries on the same table now reuse the vended credential until it nears expiry, no temp-view workaround needed.

One residual remained on authorization-enabled servers: `UCProxy.loadV1Table` requests `READ_WRITE` for every load and falls back to `READ` on denial, and denials are not cacheable — so a `SELECT`-only principal still paid one denied round-trip per query before hitting the cached `READ` credential. #1803 fixes that by using Spark's declared write intent (`loadTable(ident, writePrivileges)`, available on all supported Spark versions) plus a per-table memory of `READ_WRITE` denials, and marks this issue closed.

tiennguyen-onehouse:
One addition for completeness: with the UC Delta REST API enabled (the default when the connector runs with Delta >= 4.3), table credentials are vended by delta-io's client rather than this connector, and that path has the same unconditional-READ_WRITE pattern. Filed and fixed upstream as delta-io/delta#7517 / delta-io/delta#7518; the connector-side E2E for that path (activates once the pinned Delta carries the fix) is #1806. So the full resolution of this issue is #1803 (legacy path, this repo) + delta-io/delta#7518 (Delta REST path, upstream).

PR #1803, opened 2026-08-24
fix(spark): honor Spark's declared write intent when vending table credentials

tiennguyen-onehouse:
## Summary

Partially addresses #1576 (does not close it — see "What remains" below).

`UCProxy.loadV1Table` requests `READ_WRITE` credentials for every load and falls back to `READ` on denial, regardless of what the statement is about to do. The TODO justifying that ("the server doesn't distinguish between READ and READ_WRITE credentials as of today") is stale: `AuthorizeExpressions.VEND_TABLE_CREDENTIAL` grants `READ_WRITE` only to `OWNER` or `SELECT+MODIFY`. For a declared write by a principal without that grant, the old fallback vended `READ` credentials and let the job fail later at the storage layer.

**Spark has been declaring intent all along.** `TableCatalog.loadTable(Identifier, Set<TableWritePrivilege>)` exists on every Spark version this connector supports (4.0/4.1/4.2, per `project/spark-versions.json`) and is how the analyzer resolves the target of INSERT / UPDATE / DELETE / MERGE. The connector never overrode it, so the default implementation silently dropped the privileges and forwarded to the intent-less `loadTable(ident)`.

Changes:

- **`UCSingleCatalog` overrides `loadTable(ident, writePrivileges)`** and records the intent in a thread-local for the vend site. A thread-local (not the argument) because the `DeltaCatalog` layer in between does not override the overload — its inherited default drops the privileges on the way to `UCProxy`, where the credential is actually vended. The privileges are still forwarded so a future `DeltaCatalog` that honors them receives them; resolution runs on the calling thread, and the flag is cleared in `finally`.
- **Declared write** → request `READ_WRITE`, and surface a denial immediately. Falling back to `READ` here (old behavior) only deferred the same authorization failure to the storage layer mid-job; now the caller gets a clean permission error at analysis time.
- **No declared intent** → unchanged: request `READ_WRITE` first, fall back to `READ`. Intent-less loads cannot be assumed to be reads: Delta's `OPTIMIZE`/`VACUUM`, every `ALTER TABLE` variant, and `writeStream.toTable()` all resolve their target through the intent-less overload, so defaulting them to `READ` would break writers who do hold the grant.
- One deliberate consequence: `EXPLAIN` of a write by a principal without write access now surfaces the permission error at analysis time (previously it analyzed against READ credentials and only the run failed).

## What remains of #1576

#1576 is about the redundant denied `READ_WRITE` round-trip that a SELECT-only principal pays on every query. This PR does **not** remove it: SELECT resolves through the intent-less overload, and intent-less loads keep the `READ_WRITE`-first probe for the reasons above. An earlier revision remembered denials per table to skip the probe; it was dropped per review (staleness for grants added after a denial, unbounded growth). A deterministic follow-up would be an opt-in catalog option for read-only sessions under which intent-less loads request `READ` directly.

Not changed: create/staging paths (always writes, already `READ_WRITE`), the view surface, path credentials, and server-side planning fallbacks (the `READ`+SSP fallback is extracted to `readCredentialsOrServerSidePlanning`, behavior identical).

**Scope — which credential path this governs.** With the UC Delta REST API enabled (`deltaRestApi.enabled`, on by default with Delta >= 4.3), table loads vend credentials through delta-io's own client (`DeltaTableCredId`), not `UCProxy.loadV1Table` — verified empirically while writing the E2E suite. That path currently requests `READ_WRITE` on every load too, but its operation choice lives in delta-io, so bringing intent selection there is upstream work (delta-io/delta#7518). This PR fixes the path this connector owns: non-Delta formats, Delta < 4.3, and `deltaRestApi.enabled=false`.

## Blast radius

For most deployments this change is a wire-level no-op: with authorization disabled, or for any principal holding write access, the request pattern is byte-identical to before (READ_WRITE first, granted, done). Every behavioral difference is confined to declared writes whose READ_WRITE request the server denies — and for them the underlying operation was already going to fail; what changes is when and how cleanly.

One scenario deserves explicit mention: if a server variant vends credentials that are not actually operation-scoped (the same storage credential for READ and READ_WRITE), the old fallback allowed a catalog-denied write to silently succeed at the storage layer. This change enforces the catalog's denial at analysis time. That is intentional: the catalog's authorization decision, not an accident of credential scoping, should decide whether a write proceeds.

## How we tested

**End-to-end — `WriteIntentE2ETest`** (real Spark SQL -> `UCSingleCatalog` -> real `DeltaCatalog` delegation -> `UCProxy` against a live UC server; table-credential fetches recorded, and selectively denied, at the hadoop connector's `CredPropsUtil.genericCredFetcherFactory` seam, which sees the requested `TableOperation` and the intent flag):

- `sqlStatementsRequestCredentialsMatchingTheirIntent` — `INSERT`, `UPDATE`, `DELETE`, `MERGE` each reach the vend site with declared write intent; `SELECT` stays intent-less; `MERGE` resolves its target with intent and its source without **in the same statement**; the thread-local is clear after every statement (no leak across the `DeltaCatalog` delegation).
- `readOnlyPrincipalReadsViaFallbackAndWritesFailFast` — with `READ_WRITE` denied for one table: `SELECT` = denied `READ_WRITE` then `READ` (query returns data); an `INSERT` fails fast with the denial message and no `READ` fallback fetch; after the "grant" returns, the write succeeds.
- `parquetV1WritePathDeclaresWriteIntent` — the V1 (non-Delta) write path also declares intent.

**Unit — `WriteIntentCredentialSuite`** (mock-backed `UCProxy`, same seam): intent-less denial falls back to `READ`; declared write fails fast without `READ` fallback; granted declared write requests `READ_WRITE` only.

```
build/sbt "spark/testOnly io.unitycatalog.spark.WriteIntentE2ETest io.unitycatalog.spark.WriteIntentCredentialSuite io.unitycatalog.spark.UCProxySuite"
[info] Passed: Total 16, Failed 0, Errors 0, Passed 16
```

**Delta REST API path**: the matching fix on delta's side is delta-io/delta#7518, and the connector E2E for that path (gated on the Delta version) is split into #1806 so this PR carries no dependency on the Delta work.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -220,11 +220,12 @@ class UCSingleCatalog
       ident: Identifier,
       writePrivileges: util.Set[TableWritePrivilege]): Table = {
     requireAddressableTableNameOrPathTable(ident)
+    val previousIntent = UCSingleCatalog.WRITE_INTENT.get()
     UCSingleCatalog.WRITE_INTENT.set(true)
     try {
       delegate.loadTable(ident, writePrivileges)
     } finally {
-      UCSingleCatalog.WRITE_INTENT.remove()
+      UCSingleCatalog.WRITE_INTENT.set(previousIntent)
     }
   }
 
@@ -1049,12 +1050,12 @@ private[spark] class UCProxy(
       credBuilder.buildForTable(tableId, TableOperation.READ_WRITE)
     } catch {
       case e: ApiException =>
-        logWarning(
-          s"READ_WRITE credential generation failed for table $identifier: ${e.getMessage}")
         if (UCSingleCatalog.WRITE_INTENT.get()) {
           // Fail fast: READ credentials would only defer the denial to the storage layer mid-job.
           throw e
         }
+        logWarning(
+          s"READ_WRITE credential generation failed for table $identifier: ${e.getMessage}")
         readCredentialsOrServerSidePlanning(credBuilder, tableId, identifier)
     }
 
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -215,10 +215,7 @@ class UCSingleCatalog
     delegate.loadTable(ident, timestamp)
   }
 
-  // Spark resolves the target of INSERT / UPDATE / DELETE / MERGE through this overload, so it is
-  // the one place write intent is declared. Record it for `UCProxy.loadV1Table` (see
-  // `WRITE_INTENT` for why a thread-local and not the `writePrivileges` argument itself), and
-  // still forward the privileges so a future `DeltaCatalog` that honors them receives them.
+  // Spark resolves write targets (INSERT / UPDATE / DELETE / MERGE) through this overload.
   override def loadTable(
       ident: Identifier,
       writePrivileges: util.Set[TableWritePrivilege]): Table = {
@@ -745,12 +742,9 @@ object UCSingleCatalog {
   val LOAD_DELTA_CATALOG = ThreadLocal.withInitial[Boolean](() => true)
   val DELTA_CATALOG_LOADED = ThreadLocal.withInitial[Boolean](() => false)
 
-  // True while the current thread resolves a table Spark declared as a write target via
-  // `loadTable(ident, writePrivileges)`. Carried out-of-band because the `DeltaCatalog` layer
-  // between `UCSingleCatalog` and `UCProxy` does not override that overload -- its inherited
-  // default drops the privileges and forwards to the intent-less `loadTable(ident)` -- so the
-  // vend site in `UCProxy.loadV1Table` could not see the intent otherwise. Set/cleared only in
-  // `UCSingleCatalog.loadTable(ident, writePrivileges)`; resolution runs on the calling thread.
+  // True while the current thread resolves a declared write target. Thread-local because
+  // DeltaCatalog does not override `loadTable(ident, writePrivileges)` and drops the argument
+  // before it reaches the vend site in `UCProxy.loadV1Table`.
   private[spark] val WRITE_INTENT = ThreadLocal.withInitial[Boolean](() => false)
 
   /**
@@ -997,13 +991,9 @@ private[spark] class UCProxy(
     }
   }
 
-  // Table ids whose READ_WRITE credential request was denied, so intent-less loads (SELECT
-  // resolution) go straight to READ instead of re-paying the denied round-trip on every query.
-  // An entry is cleared when a declared write for the table succeeds; per catalog instance.
+  // Table ids denied READ_WRITE, so intent-less loads skip the doomed round-trip per query.
   private[spark] val writeDeniedTables = util.concurrent.ConcurrentHashMap.newKeySet[String]()
 
-  // READ-credential fallback shared by the intent-less load paths: on failure, fall back to
-  // server-side planning when enabled, otherwise propagate.
   private def readCredentialsOrServerSidePlanning(
       credBuilder: UCCredentialHadoopConfs.Builder,
       tableId: String,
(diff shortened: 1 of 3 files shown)


tiennguyen-onehouse:
Upstream counterpart for the Delta REST API path (where loads vend through delta-io's client and this connector's fix does not apply): delta-io/delta#7518 (issue delta-io/delta#7517).

yili-db:
@openinx @timothyw553 Please take a look. Thank you!

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -766,6 +783,47 @@ private[spark] class UCProxy(
     }
   }
 
+  // Table ids whose READ_WRITE credential request was 403-denied, mapped to the deadline until
+  // which intent-less loads skip the doomed round-trip. Entries expire (so a later grant reaches
+  // intent-less paths like OPTIMIZE without a session restart) and are cleared early by a
+  // successful declared write.
+  private[spark] val writeDeniedTables =
```
timothyw553: not really a blocker: `writeDeniedTables` is unbounded, and an expired entry is removed only if the same table is loaded again. so a long-running process can keep entries for many tables forever. can we use a bounded cache with a 5 minute TTL -- maybe with LRU policy. i think this can be done in a follow up since it is an extreme case where this cache will have 100k+ entries in-memory.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -800,27 +858,36 @@ private[spark] class UCProxy(
...
+      } catch {
+        case e: ApiException =>
+          logWarning(s"READ_WRITE credential generation failed for declared write on table " +
+            s"$identifier: ${e.getMessage}")
+          if (e.getCode == HttpURLConnection.HTTP_FORBIDDEN) markWriteDenied(tableId)
+          throw e
+      }
+      writeDeniedTables.remove(tableId)
+      props
+    } else if (isWriteDenied(tableId)) {
```
timothyw553: on the Delta connector side, `OPTIMIZE` uses this intent-less path. here's a bit of an edge case from a client:
 1. The user first lacks write permission.
 2. An administrator grants write permission.
 3. The user runs intent-less OPTIMIZE within five minutes.

then that step 3 will fail because it will ask for `READ_WRITE` but then this cache makes it only give it back `READ` permission. if some step requires `WRITE` (which it most likely will), then it the OPTIMIZE command will fail.

somehow we need to acknowledge that this can happen cc: @yili-db

timothyw553 (commented):
I don't have any major blocking comments. The intent of this optimization makes sense.

yili-db:
@tiennguyen-onehouse Thank you for this PR.
- The `WRITE_INTENT` mechanism is a nice signal that is picked up from Spark and passed into `UCProxy`. It solves most of the useless "readwrite-then-read" fallback. This is very useful.
- The `writeDeniedTables`, OTOH, is a negative response cache that can easily go stale. With the `WRITE_INTENT` implemented, I don't think there's much value can be provided by `writeDeniedTables`. And it can easily introduce staleness problem as @timothyw553 pointed out.

So my suggestion is to keep only `WRITE_INTENT` in this PR.

tiennguyen-onehouse:
Thanks both. Dropped `writeDeniedTables` in 7e6261bb; the PR now carries only the `WRITE_INTENT` override plus fail-fast on a denied declared write. Intent-less loads are back to the unchanged `READ_WRITE`-then-`READ` probe.

One thing worth being explicit about: with the memory gone, this no longer changes the request pattern for SELECT-only principals. Their loads are intent-less, so they still pay the denied `READ_WRITE` round-trip per query. Intent-less cannot default to `READ` because `OPTIMIZE`/`VACUUM`, every `ALTER TABLE`, and `writeStream.toTable()` resolve through the intent-less overload (Delta's `DeltaCatalog` does not implement the privileged one). So what remains is a correctness/fail-fast change; I retitled the PR and rewrote the description accordingly.

If there is appetite for the read-side savings, the deterministic option is an opt-in catalog setting for read-only sessions (e.g. intent-less loads request `READ`, default keeps today's probe), alongside the existing `renewCredential.enabled` / `deltaRestApi.enabled` style options. Happy to open that as a separate issue/PR rather than widen this one.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
tiennguyen-onehouse: Removed the map entirely in 7e6261bb rather than bounding it, per @yili-db's comment below.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
tiennguyen-onehouse: Agreed, and this scenario is also why intent-less loads cannot default to READ. The denial memory is gone in 7e6261bb; OPTIMIZE now takes the same READ_WRITE-first path it did before this PR, so a grant is picked up on the next load.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -800,27 +835,27 @@ private[spark] class UCProxy(
...
-            } else {
-              throw e
-            }
-        }
+    // Intent-less loads still request READ_WRITE first: write paths such as Delta's OPTIMIZE,
+    // ALTER TABLE and streaming sinks resolve their target through the intent-less overload.
+    val extraSerdeProps = if (UCSingleCatalog.WRITE_INTENT.get()) {
+      // Fail fast: READ credentials would only defer the denial to the storage layer mid-job.
+      try {
+        credBuilder.buildForTable(tableId, TableOperation.READ_WRITE)
```
yili-db: This and the `else` branch is nearly identical except whether they fallback or not.
They can be folded to dedup and the `if (UCSingleCatalog.WRITE_INTENT.get()) {` can be moved to the `catch` block.

yili-db (commented):
This is in a much better shape.
But now it doesn't completely fix #1576. Please adjust the PR message.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
tiennguyen-onehouse: Folded in 319d2b4e: one try around the READ_WRITE request, and the catch decides between rethrow (declared write) and the READ fallback.

tiennguyen-onehouse:
Rebased onto main (conflict with #1653/#1810 in `UCSingleCatalog.scala`; the new override now goes through `requireAddressableTableNameOrPathTable` like the sibling `loadTable` overloads) and folded the vend-site branches per the inline comment. PR description now says *partially addresses* #1576 and has a **What remains of #1576** section spelling out that the SELECT-side probe is untouched here.

timothyw553 (approved):
LGTM. Thanks for tracking the follow up work.

yili-db (changes requested):

yili-db (approved):

PR #1805, opened 2026-08-24
Reject metric view creation with an empty column list

chenwang-databricks:
## What

Reject creation of a view or metric view whose `columns` list is null or empty.

A view (plain or metric) exposes its output as columns, and a reader resolves queries against that column list. The server previously accepted a view or metric view with no columns and persisted an entity that no engine can read. 

## Changes

- Reject a null or empty `columns` list in `validateViewLike`, so the rule covers both `VIEW` and `METRIC_VIEW`, with `INVALID_ARGUMENT`. The message carries the entity label: a plain view reports `columns must contain at least one entry for view`, a metric view reports `... for metric view`. The check runs ahead of the dependency-existence lookups, so a request-shape error is returned without first spending those lookups.
- Normalize `columns` before streaming it in `createTableImpl`. `columns` deserializes to null when a client omits the field, so a request carrying `"columns": null` previously failed with a `NullPointerException` and an internal error instead of `INVALID_ARGUMENT`.

## Scope

`VIEW` and `METRIC_VIEW` only. Other table types keep their existing behavior; for them the normalization only converts the former `NullPointerException` on a null `columns` field into the same handling as an empty list, which they already accept.

The Spark connector is unaffected: it always sends columns for both plain views and metric views, resolved from the view output schema, so no valid connector request is rejected.

## Tests

- The metric-view and plain-view happy-path requests now supply columns; the metric-view test columns also carry the required `metric_view.type` and `metric_view.expr` metadata, and the test asserts `type_json` round-trips verbatim.
- Added empty-list and null-list negative cases to both the view and metric-view suites.
- Full `server/test` passes (424).

---

**PR Checklist**

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -622,8 +622,12 @@ public class TableRepository {
     ValidationUtils.validateSqlObjectName(createTable.getName());
     String callerId = IdentityUtils.findPrincipalEmailAddress();
     DataSourceFormat format = createTable.getDataSourceFormat();
+    // `columns` arrives as null (not an empty list) when a client omits the field entirely, so
+    // normalize before streaming. Whether an empty column list is acceptable is a per-table-type
+    // rule decided in the create branches below, and those checks must be the ones that reject the
+    // request so the caller gets INVALID_ARGUMENT instead of an NPE here.
     List<ColumnInfo> columnInfos =
-        createTable.getColumns().stream()
+        Optional.ofNullable(createTable.getColumns()).orElse(List.of()).stream()
             .map(
                 c -> {
                   ColumnUtils.validateTypeJson(c, format);
@@ -767,6 +771,17 @@ public class TableRepository {
   }
 
   private void validateMetricView(Session session, CreateTable createTable) {
+    // A metric view exposes its dimensions and measures as columns, and engines resolve queries
+    // against that column list. A metric view with no columns is therefore accepted by the server
+    // but unusable by any reader, so reject it rather than persisting an unreadable entity.
+    // Column-level validation beyond this (that each column matches the view_definition YAML) needs
+    // the YAML semantics and stays with the engine that owns them.
+    // Checked ahead of validateViewLike so this request-shape rule is reported without first
+    // spending dependency-existence lookups on a request that cannot be accepted either way.
+    if (Optional.ofNullable(createTable.getColumns()).orElse(List.of()).isEmpty()) {
+      throw new BaseException(
+          ErrorCode.INVALID_ARGUMENT, "columns must contain at least one entry for metric view");
+    }
     validateViewLike(session, createTable, "metric view");
     // A metric view's dependencies are always client-supplied; require at least one.
     if (Optional.ofNullable(createTable.getViewDependencies())
--- a/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
@@ -771,17 +771,6 @@ public class TableRepository {
   }
 
   private void validateMetricView(Session session, CreateTable createTable) {
-    // A metric view exposes its dimensions and measures as columns, and engines resolve queries
-    // against that column list. A metric view with no columns is therefore accepted by the server
-    // but unusable by any reader, so reject it rather than persisting an unreadable entity.
-    // Column-level validation beyond this (that each column matches the view_definition YAML) needs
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -743,34 +607,27 @@ public DeltaLoadTableResponse createTableForDelta(
...
    */
   private <T> T createTableImpl(
       CreateTable createTable,
       Optional<DeltaUniformUtils.UniformIcebergFields> uniformFields,
-      Optional<NormalizedURL> nativeIcebergMetadataLocation,
       CreateResultMapper<T> mapper) {
     ValidationUtils.validateSqlObjectName(createTable.getName());
     String callerId = IdentityUtils.findPrincipalEmailAddress();
     DataSourceFormat format = createTable.getDataSourceFormat();
+    // `columns` arrives as null (not an empty list) when a client omits the field entirely, so
```
yili-db, line 625: Is it never null or empty because `validateMetricView` doesn't accept?
Then change in this function is redundant.

server/src/test/java/io/unitycatalog/server/base/table/BaseMetricViewCRUDTest.java
```diff
@@ -110,12 +111,41 @@ private static DependencyList makeDependencyList(String... tableFullNames) {
...
+   * a non-empty column list, so every valid request below carries these.
+   */
+  protected static final List<ColumnInfo> METRIC_VIEW_COLUMNS =
+      List.of(
+          new ColumnInfo()
+              .name("event_day")
+              .typeText("DATE")
+              .typeJson(
+                  "{\"name\":\"event_day\",\"type\":\"date\","
+                      + "\"nullable\":true,\"metadata\":{}}")
```
chenwang-databricks: Please follow this rule for test data:

Each metric-view column needs this metadata:

  {
    "metric_view.type": "dimension",
    "metric_view.expr": "region"
  }

  or:

  {
    "metric_view.type": "measure",
    "metric_view.expr": "sum(count)"
  }

  Rules:

  - Both values are JSON strings.
  - metric_view.type is dimension or measure.
  - metric_view.expr is the original YAML expression, limited to 1024 Scala string characters.
  - Both keys are emitted and declared required. No other metric-view-specific column keys are required.

  For a UC-style type_json, this is nested in the standard StructField shape:

  {
    "name": "region",
    "type": "string",
    "nullable": true,
    "metadata": {
      "metric_view.type": "dimension",
      "metric_view.expr": "region"
    }
  }

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
chenwang-databricks: Not redundant — the two checks run at different points in the request.

`createTableImpl` builds `columnInfos` *before* the transaction opens, while `validateMetricView` runs *inside* it, in the `METRIC_VIEW` branch. `columns` deserializes to `null` (not an empty list) when a client omits the field, so for a request carrying `"columns": null` the stream here is reached first and throws `NullPointerException` before any validation runs. The caller then sees an internal error instead of `INVALID_ARGUMENT`.

The `missing columns` negative test covers exactly this case, and it fails without the normalization here.

One caveat worth naming: the normalization applies to every table type, while the non-empty rule is `METRIC_VIEW`-only. For other table types `null` now behaves the same as an empty list, which they already accept. If you would rather keep this PR scoped strictly to metric views, I can move the null guard into the `METRIC_VIEW` branch and leave the `NullPointerException` for other table types to a separate fix.

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -920,6 +771,17 @@ private interface CreateResultMapper<T> {
   }
 
   private void validateMetricView(Session session, CreateTable createTable) {
+    // A metric view exposes its dimensions and measures as columns, and engines resolve queries
```
yili-db: I think the same rule applies perfectly to views too. A plain view without columns isn't usable by Spark.
Can you please move this validation into `validateViewLike`?

chenwang-databricks: Done, moved into `validateViewLike` so it now applies to both plain views and metric views. The message uses the entity label — a plain view reports `columns must contain at least one entry for view`, a metric view reports `... for metric view`.

Kept it ahead of the dependency-existence lookups so a request-shape error is still returned without first spending those. Added empty-list and null-list negative cases to the view suite to match the metric-view suite; the plain-view happy path already supplies columns, so it is unaffected. In `e3dfc150`.

yili-db (approved):

PR #1806, opened 2026-08-24
test(spark): E2E for the Delta REST API credential path (gated on delta-io/delta#7518)

tiennguyen-onehouse:
## Summary

Follow-up to #1803, split out so that PR is not coupled to the Delta-side work. **Stacked on #1803** — only the last commit (`test(spark): E2E for the Delta REST API credential path...`) is new here; review after #1803 merges.

#1803 fixes intent-aware credential selection on the connector's legacy path. On the Delta REST API path (default with Delta >= 4.3), credentials are vended by delta-io's client, and the matching fix is delta-io/delta#7518. This PR adds the connector-side E2E for that path: `WriteIntentE2ETest.deltaRestApiPathHonorsDeclaredWriteIntent` drives real SQL through `UCSingleCatalog` -> `DeltaCatalog` -> delta's UC client against a live UC server with the Delta REST API **enabled**, asserting on the recorded credential operations:

- a SELECT under a READ_WRITE denial requests READ_WRITE, is denied, and is served READ;
- a declared INSERT fails fast with no READ fallback — the write intent must survive both #1803's forwarding and delta-io/delta#7518's plumbing to produce this;
- once the grant returns, the declared write requests READ_WRITE only and succeeds.

The test is gated on a reflection check for Delta's intent-carrying `loadTable(ident, writePrivileges)` overload: on the pinned Delta 4.3.1 it aborts with a named JUnit assumption, and it activates automatically when the pin picks up a Delta release carrying delta-io/delta#7518. No CI impact until then.

## How we tested

On stock Delta 4.3.1 (what CI runs) the REST-path case is cleanly skipped via assumption; the other three cases pass. The active case was last exercised against a 4.4.0-SNAPSHOT of delta-io/delta#7518 before both PRs dropped their per-table denial memory per review; the remaining assertions are a strict subset of what passed then.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -220,11 +220,12 @@ class UCSingleCatalog
       ident: Identifier,
       writePrivileges: util.Set[TableWritePrivilege]): Table = {
     requireAddressableTableNameOrPathTable(ident)
+    val previousIntent = UCSingleCatalog.WRITE_INTENT.get()
     UCSingleCatalog.WRITE_INTENT.set(true)
     try {
       delegate.loadTable(ident, writePrivileges)
     } finally {
-      UCSingleCatalog.WRITE_INTENT.remove()
+      UCSingleCatalog.WRITE_INTENT.set(previousIntent)
     }
   }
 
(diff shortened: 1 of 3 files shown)


tiennguyen-onehouse:
Rebased onto the current #1803 head (7cfad83a) — this branch is again just #1803 plus one test commit. The REST-path case no longer asserts a per-table denial memory, since both #1803 and delta-io/delta#7518 dropped it per review; it now checks SELECT = denied READ_WRITE then READ, declared INSERT fails fast with no READ fallback, and a re-granted write requests READ_WRITE only. Still gated by assumption on Delta carrying delta-io/delta#7518 (skips on the pinned 4.3.1).

PR #1808, opened 2026-08-25
Recover schema from Delta log when DeltaCatalog delegates createTable with empty columns

sambhav8695:
**PR Checklist**

**Description of changes**

## Problem

When creating a Delta table through `UCSingleCatalog` with `DeltaCatalog` on the classpath
(e.g. `CREATE TABLE ... USING DELTA LOCATION '...'`), the table is registered in Unity Catalog
with **zero columns**, even though the `CREATE TABLE` statement specifies a full schema.

Repro:
```sql
CREATE TABLE catalog.schema.users (id INT, name STRING, email STRING) USING DELTA LOCATION 's3://...'
```
Before this change, `TablesApi.getTable(...)` for this table returns an empty `columns` list.

## Root cause

`UCSingleCatalog.createTable` builds the correct schema from the caller's columns and delegates
to `DeltaCatalog` (`delegate.createTable(ident, schema, partitions, newProps)`). `DeltaCatalog`
commits that schema into the table's `_delta_log` itself — from Delta's point of view, the
transaction log is the source of truth for schema, not the catalog. As its last step it calls
back into the underlying delegate catalog (`UCProxy`, via the legacy
`TableCatalog.createTable(Identifier, StructType, ...)` overload) purely to register a
location/provider pointer, passing an **empty** `StructType` since it doesn't expect the
underlying catalog to need the columns.

`UCProxy.createTable` faithfully persists whatever schema it's given, so UC ends up with a
zero-column table record. This is invisible to Spark reads in the same session (Delta resolves
schema from its own log at read time), but breaks any UC-native consumer of table metadata —
direct `TablesApi.getTable` callers, `SHOW COLUMNS`, other non-Spark engines reading UC
directly, etc.

## Fix

In `UCProxy.createTable(ident, schema: StructType, ...)`, detect the empty-schema +
delta-provider case and recover the real schema by reading it back from the table's
`_delta_log` (already committed by `DeltaCatalog` at this point) before building the
`ColumnInfo`s sent to `TablesApi.createTable`.

- New method `UCProxy.readSchemaFromDeltaLog(properties)`: lists `_delta_log/*.json` in commit
  order, parses each commit's `metaData.schemaString`, and keeps the last one seen (handles the
  rare case of multiple commits/schema evolution within the same create).
- Credentials for reading the log are fetched fresh via the existing
  `UCCredentialHadoopConfs.builder(...).buildForPath(...)` chain, rather than relying on
  `properties` — by the time `DeltaCatalog` calls back into `UCProxy`, it may have
  rebuilt/stripped the property map, so any `fs.*` / `option.fs.*` credential overrides
  `UCSingleCatalog` originally vended upstream aren't reliably present here.
- The recovery is **best-effort**: any failure (missing log, credential failure, malformed
  commit line) is logged and swallowed, falling back to the original (possibly empty) schema.
  This can never turn a previously-working `CREATE TABLE` into a failure — worst case, behavior
  is unchanged from before this fix.
- The recovery is only attempted when the incoming schema is empty **and** the provider is
  `delta`; a non-empty schema is never overridden, and non-Delta providers are left untouched.

## Scope / non-goals

- This only affects the *external-table-with-location* Delta create path that goes through the
  legacy `TableCatalog.createTable(StructType, ...)` overload (the `hasLocationClause` branch in
  `UCSingleCatalog.createTable`). It does not touch the newer UC Delta REST API path
  (`shouldUseDeltaAPI` / `managedDeltaCreatePropsForDelegate`), which is the longer-term intended
  mechanism for schema agreement between Delta and UC.
- No behavior change when `DeltaCatalog` is not on the classpath, for non-Delta providers, or
  when the incoming schema is already non-empty.

## Testing

Added `UCProxyDeltaSchemaRecoveryTest` (JUnit5 / Mockito / AssertJ, matching the conventions in
`UCSingleCatalogStagingTableTest`):

- `readSchemaFromDeltaLog`: recovers schema from a single commit; picks up the latest schema
  across multiple commits (schema evolution); tolerates realistic multi-action commit files
  (protocol/metaData/commitInfo lines); returns empty when `_delta_log` is missing; returns
  empty when no location property is present; fails soft (returns empty) on a malformed commit
  line.
- `createTable`: recovers and sends the correct columns when the incoming schema is empty and
  the provider is `delta`; never overrides an already-populated incoming schema, even when the
  Delta log at the same location disagrees; does not attempt recovery for non-Delta providers
  even when a valid Delta log happens to be present at the location.

Also manually verified end-to-end against a local UC server: `CREATE TABLE ... USING DELTA
LOCATION ...` now returns the correct column count/types from `TablesApi.getTable(...)` after
this change, where it previously returned zero columns.

### DIFF
--- a/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
+++ b/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
@@ -1,5 +1,6 @@
 package io.unitycatalog.spark
 
+import java.io.{BufferedReader, InputStreamReader}
 import java.net.URI
 import java.util
 import java.util.Locale
@@ -9,6 +10,7 @@ import scala.collection.convert.ImplicitConversions._
 import scala.collection.mutable.ArrayBuffer
 import scala.language.existentials
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import io.unitycatalog.client.api.{SchemasApi, TablesApi}
 import io.unitycatalog.client.auth.TokenProvider
 import io.unitycatalog.client.model.{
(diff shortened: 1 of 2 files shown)


yili-db:
Hi @sambhav8695 . This is largely a Delta problem and `UCSingleCatalog` is meant to be generic connector that shouldn't read from Delta commit log directly.
Eventually this problem should be fixed on Delta side so that Delta should send the columns for external tables as well and via the new UC Delta API (it already does so for managed tables).
Delta intentionally doesn't send the columns for external tables via the `UCProxy` today, because it hasn't implemented the mechanism to **update** columns for `ALTER TABLE` or other operations. So it's better to remain unset than being outdated/out-of-sync later.

This gap is recognized by Delta project and will be fixed on Delta side.

Thank you for checking this issue! You are very welcomed to contribute!

PR #1809, opened 2026-08-26
Upgrade google-java-format so javafmt actually checks modern Java

s-maling-celonis:
## Summary

We maintain a private fork that has to use a separate build (Maven / Spotless) alongside this sbt tree. The current formatter is **google-java-format 1.7** via `sbt-java-formatter` 0.8.0. That release cannot parse text blocks or switch expressions, so `javafmtCheck` silently skips those files, and GJF 1.7 cannot be run at all on JDK 17 by modern format plugins (Spotless requires ≥ 1.10.0). The two builds therefore cannot share a formatter.

This PR bumps to the oldest GJF the current `sbt-java-formatter` will actually run on the JDK 17 CI image: **1.28.0** (`sbt-java-formatter` 0.13.1, `javafmtFormatterCompatibleJavaVersion := 17`). That is a version a Maven/Spotless build can pin as well, so both trees can stay format-identical. 0.13.1 only changes the Java 21 formatter line (GJF 1.36.1); this tree stays on 17 / 1.28.0.

- Plugin + Checkstyle + README in the first commit. Checkstyle `Indentation` / `CommentsIndentation` are dropped, not retuned. They are Spark wrap-2; google-java-format is Google wrap-4. They only coexisted because GJF 1.7 never parsed switch expressions or text blocks. Checkstyle 10/11 still reject that layout, so indent stays owned by GJF.
- Mechanical `javafmtAll` in the second commit (review with `git show -w`).
- Plugin bump 0.12.0 → 0.13.1 in a follow-up commit; no extra reformat.

## Test plan
- [x] `build/sbt javafmtCheck` (same command as `.github/workflows/lint-check.yml`)
- [x] `build/sbt javafmtCheckAll`
- [ ] Confirm the `java-fmt` GitHub Actions job is green

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -28,7 +28,10 @@ addSbtPlugin("org.openapitools" % "sbt-openapi-generator" % "7.9.0")
 
 addSbtPlugin("com.github.sbt" % "sbt-license-report" % "1.6.1")
 
-addSbtPlugin("com.lightbend.sbt" % "sbt-java-formatter" % "0.8.0")
+// 0.8.0 bundled google-java-format 1.7, which cannot parse text blocks or switch
+// expressions and silently skipped those files. 0.12.0 forks GJF; the Java 17
+// line (see build.sbt) is google-java-format 1.28.0.
+addSbtPlugin("com.github.sbt" % "sbt-java-formatter" % "0.12.0")
 
 addSbtPlugin("software.purpledragon" % "sbt-checkstyle-plugin" % "4.0.1")
 // By default, sbt-checkstyle-plugin uses checkstyle version 6.15, but we should set it to use the
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -28,9 +28,8 @@ addSbtPlugin("org.openapitools" % "sbt-openapi-generator" % "7.9.0")
 
 addSbtPlugin("com.github.sbt" % "sbt-license-report" % "1.6.1")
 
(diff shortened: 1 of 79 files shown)


project/plugins.sbt
```diff
@@ -28,7 +28,10 @@ addSbtPlugin("org.openapitools" % "sbt-openapi-generator" % "7.9.0")
 
 addSbtPlugin("com.github.sbt" % "sbt-license-report" % "1.6.1")
 
-addSbtPlugin("com.lightbend.sbt" % "sbt-java-formatter" % "0.8.0")
+// 0.8.0 bundled google-java-format 1.7, which cannot parse text blocks or switch
+// expressions and silently skipped those files. 0.13.1 forks GJF; the Java 17
+// line (see build.sbt) is still google-java-format 1.28.0.
```
yili-db: `still` is the word that doesn't make sense once this PR is squashed and merged because this whole PR is bumping up GJF from 1.7 to 1.28.0 directly.

yili-db (commented):
Please rebase

project/plugins.sbt
s-maling-celonis: Dropped `still` and shortened this to state the Java 17 GJF version and why 0.8.0/1.7 was insufficient.

s-maling-celonis:
@yili-db Rebased onto `main` — thanks. Ready for re-review.

tdas (approved):

yili-db (approved):

PR #1810, opened 2026-08-26
[Spark] Bypass catalog lookup for non-Delta paths

### LINKED ISSUE #1075: Spark Parquet file path syntax is not accepted

s-maling-celonis:
**PR Checklist**

- [x] If you've added or modified a feature, documentation in `docs` is updated (not applicable: bug fix only)

**Description of changes**

Fixes #1075.

When a Unity Catalog is current, Spark probes it before resolving a path-based data source such as a Parquet path (`s3://bucket/path`). With Delta 4.3.1, the outer Delta catalog delegate tries to load that identifier through the UC REST API. UC rejects the path as an invalid table name with HTTP 400, preventing Spark's SQL-on-file fallback.

This change rejects non-Delta path identifiers in `UCSingleCatalog` before calling the delegate, allowing Spark to resolve them as file paths. Bare Delta paths remain delegated to `DeltaCatalog`.

Tests cover:
- all three `loadTable` overloads rejecting non-Delta paths without invoking the delegate;
- Delta path identifiers continuing to reach the delegate;
- an end-to-end Parquet path read while a Unity Catalog is current.

Tested with the repository defaults (Spark 4.1.0 and Delta 4.3.1):
- `./build/sbt -J-Xmx2G 'spark/compile'`
- `./build/sbt -J-Xmx2G 'spark/testOnly io.unitycatalog.spark.UCSingleCatalogStagingTableTest io.unitycatalog.spark.PathTableReadTest'` (25 tests passed)


connectors/spark/src/test/java/io/unitycatalog/spark/PathTableReadTest.java
```diff
@@ -0,0 +1,30 @@
...
+/** Integration tests for path-based data source queries while a Unity Catalog is current. */
+public class PathTableReadTest extends BaseSparkIntegrationTest {
+
+  @TempDir protected File dataDir;
+
+  @Test
+  public void testBareParquetPathWithCurrentUcCatalog() throws IOException {
+    session = createSparkSessionWithCatalogs(SPARK_CATALOG, CATALOG_NAME);
+    String location = new File(dataDir, "parquet-path").getCanonicalPath();
+    sql("INSERT OVERWRITE DIRECTORY '%s' USING parquet SELECT 1 AS id", location);
```
yili-db: https://github.com/unitycatalog/unitycatalog/pull/1653 has similar test `PathCredentialReadWriteTest` that exercises both `INSERT OVERWRITE DIRECTORY` and `SELECT` already. Is that good enough?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -200,11 +200,38 @@ class UCSingleCatalog
...
 
-  override def loadTable(ident: Identifier, timestamp:  Long): Table = delegate.loadTable(ident, timestamp)
+  /**
+   * Prevents non-Delta path identifiers (for example, `parquet`.`s3://bucket/path`) from reaching
+   * catalog delegates that interpret them as Unity Catalog table names. Reporting these identifiers
+   * as missing lets Spark's SQL-on-file resolution handle them instead. Delta paths remain delegated
+   * because DeltaCatalog resolves those directly.
+   */
+  private def requireAddressableTableNameOrDeltaPath(ident: Identifier): Unit = {
+    val isDeltaPath =
```
yili-db: AbstractDeltaCatalog.scala in Delta actually also handles iceberg tables:
```
          } else if (isIcebergPathIdentifier(ident)) {
            newIcebergPathTable(ident)
```
I am not sure how/if it works, but it's better to also pass through the Iceberg tables.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -200,11 +200,38 @@ class UCSingleCatalog
...
+      ident.namespace().length == 1 &&
+        !ident.namespace()(0).contains("/") &&
+        !ident.name().contains("/")
+    if (!isAddressableTableName && !isDeltaPath) {
+      throw new NoSuchTableException(ident)
+    }
+  }
 
   override def tableExists(ident: Identifier): Boolean = {
     delegate.tableExists(ident)
```
yili-db: Cover this as well?

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
```diff
@@ -200,11 +200,38 @@ class UCSingleCatalog
...
+  /**
+   * Prevents non-Delta path identifiers (for example, `parquet`.`s3://bucket/path`) from reaching
+   * catalog delegates that interpret them as Unity Catalog table names. Reporting these identifiers
+   * as missing lets Spark's SQL-on-file resolution handle them instead. Delta paths remain delegated
+   * because DeltaCatalog resolves those directly.
+   */
+  private def requireAddressableTableNameOrDeltaPath(ident: Identifier): Unit = {
+    val isDeltaPath =
+      ident.namespace().length == 1 && ident.namespace()(0).equalsIgnoreCase("delta")
+    val isAddressableTableName =
```
yili-db: `isAddressableTableName` is now a function introduced by https://github.com/unitycatalog/unitycatalog/pull/1653

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
yili-db: The UCProxy has `UCSingleCatalog.checkUnsupportedNestedNamespace` check. This function now short circuits for this case.
It's better to call `UCSingleCatalog.checkUnsupportedNestedNamespace` from here as well to throw an meaningful error.

connectors/spark/src/test/java/io/unitycatalog/spark/PathTableReadTest.java
s-maling-celonis: Dropped `PathTableReadTest`. `PathCredentialReadWriteTest` already covers `SET CATALOG` plus parquet path reads/writes; the catalog-guard coverage stays in `UCSingleCatalogStagingTableTest`.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: Agreed. Iceberg path identifiers (`iceberg.`path``) are now delegated the same way as Delta so `AbstractDeltaCatalog` can still construct `IcebergTablePlaceHolder`.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: `tableExists` now uses the same lookup policy: parquet-style paths return `false` without hitting the delegate; Delta/Iceberg paths still go through.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: Switched to `UCSingleCatalog.isAddressableTableName` from #1653 instead of duplicating the check.

connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
s-maling-celonis: Non-path lookups now call `checkUnsupportedNestedNamespace` first, so nested names still get `Nested namespaces are not supported` instead of `NoSuchTableException`.

s-maling-celonis: Follow-up: `loadTable` still calls `checkUnsupportedNestedNamespace` so nested names are not reported as missing tables.

`tableExists` does **not** throw that `ApiException`. Nested identifiers go to the Delta delegate so Delta 4.3+ can reject `DROP TABLE a.b.c.d` with `IllegalArgumentException` (`UC table identifier must be one of…`), which is what `BaseTableReadWriteTest` expects. Parquet-style path `tableExists` still returns `false` without hitting the delegate.

yili-db (approved):

PR #1812, opened 2026-08-27
[API] Return allowed Delta maintenance operations

timothyw553:
**PR Checklist**

- [x] A description of the changes is added to this PR.
- [x] No related issue exists.
- [x] Tests cover the new behavior.
- [x] The protocol specification and generated API docs are updated.

**Description of changes**

Catalog-managed Delta tables require catalog permission for maintenance operations. Before this change, the UC Delta API had no field for this permission. An external Delta engine could not learn which operations the catalog permits.

This change:

- Adds a typed `DeltaMaintenanceOperation` enum.
- Adds the optional `allowed-maintenance-operations` field to load-table responses.
- Returns `DATA_REORGANIZATION`, `DATA_CLEANUP`, and `METADATA_CLEANUP` for managed Delta tables.
- Returns no additional operations for external tables.
- Maps unknown Java client enum values to `UNKNOWN_DEFAULT_OPEN_API`.

A missing or empty list grants no additional permission. The new field is optional, so an old client can ignore it. The API protocol version stays at `1.0`. This change needs no database migration or stored table property.

**Tests**

- `build/sbt javafmtAll`
- `build/sbt "client/testOnly io.unitycatalog.client.delta.DeltaModelSerializationTest"`
- `build/sbt "server/testOnly io.unitycatalog.server.service.delta.DeltaModelSerializationTest"`
- `build/sbt "server/testOnly io.unitycatalog.server.sdk.delta.SdkLoadTableTest"`


api/delta.yaml
```diff
@@ -802,6 +809,14 @@ components:
           type: integer
           format: int64
           description: The latest ratified table version tracked by the server, including data-only commits. Compare with metadata.last-commit-version which only tracks metadata-changing commits.
+        client-maintenance-operations:
+          type: array
+          description: |
+            Additional maintenance operations that the catalog permits for this table. A missing
```
yili-db: ... besides what Delta spec allows by default.

timothyw553: done

api/delta.yaml
```diff
@@ -555,6 +555,13 @@ components:
         - READ
         - READ_WRITE
 
+    DeltaMaintenanceOperation:
+      type: string
+      enum:
```
yili-db, line 561: Add description for each value. Check for `x-enum-descriptions:` in this file for example.

api/delta.yaml
```diff
@@ -802,6 +809,14 @@ components:
           type: integer
           format: int64
           description: The latest ratified table version tracked by the server, including data-only commits. Compare with metadata.last-commit-version which only tracks metadata-changing commits.
+        client-maintenance-operations:
```
yili-db: The protocol version currently is `1.0`. Adding a new field should bump it up to `1.1`.
Update in this file and in `DeltaApiService.java`

api/delta.yaml
timothyw553: done. added `x-enum-descriptions` for each operation.

api/delta.yaml
timothyw553: yeah i initially had it as 1.0 because the field is actually optional on the client side (empty means non allowed), so this could keep backwards compatibility. but since this is the server-side I've updated version to 1.1

yili-db (approved):

api/delta.yaml
```diff
@@ -96,7 +96,7 @@ paths:
                   - "POST /v1/catalogs/{catalog}/schemas/{schema}/tables/{table}/metrics"
                   - "GET /v1/staging-tables/{table_id}/credentials"
                   - "GET /v1/temporary-path-credentials"
-                protocol-version: "1.0"
+                protocol-version: "1.1"
```
tdas: do we need to upgrade the protocol version? @yili-db

tdas: and if we do need to upgrade, what is the impact on the clients.. do we break existing clients?

yili-db: It's not strictly needed. The minor version upgrade gives the client info about what kind of extra fields they can expect. It's very much informational and in this case an absent `client-maintenance-operations` means the same behavior as 1.0. So either way it works.
As a convention, it's just a good practice to bump up the minor version for non-breaking so that we aren't silently evolving the RPC.

api/delta.yaml
```diff
@@ -555,6 +555,19 @@ components:
...
+    DeltaMaintenanceOperation:
+      type: string
+      description: A maintenance operation that a client can run on a Delta table.
+      enum:
+        - DATA_REORGANIZATION
+        - DATA_CLEANUP
+        - METADATA_CLEANUP
+      # Keep enum order aligned with x-enum-descriptions below
+      x-enum-descriptions:
+        - 'DATA_REORGANIZATION: Rewrites data files that the current table version uses. It changes how the files store and group the data. It does not change the current table rows.'
```
tdas: group/cluster/layout the data

api/delta.yaml
```diff
@@ -555,6 +555,19 @@ components:
...
+        - 'DATA_CLEANUP: Deletes old data files that no retained table version needs. It deletes the files only after the retention period.'
```
tdas: I don't think we need to qualify which files to delete because now it's not clear what retention period you are talking about. Rather it's better to clarify that this must not change the data of the latest version of the table.

api/delta.yaml
```diff
@@ -555,6 +555,19 @@ components:
...
+        - 'METADATA_CLEANUP: Deletes expired files from _delta_log. It keeps the files that are necessary to read retained table versions.'
```
tdas: Deletes old metadata files from the Delta log directory or other metadata related directories. It must not modify the content of the latest version of the table.

api/delta.yaml
tdas: Why does it have to say "client-". Also this doesn't convey that this is about "allowed" operations. Why not "allowed_maintenance_operations"

spec/protocols/ManagedTablesSpec.md
```diff
@@ -108,7 +108,7 @@ A Unity Catalog managed table needs to be registered and used according to the f
...
...em, with checkpoints, log compaction, and version checksum as the only default exceptions. Per the same protocol section, checkpoints and log compaction files may only target versions that are already published in `_delta_log`, while version checksum files may also be written for versions not yet published. The `client-maintenance-operations` field in a [`loadTable`](#load-a-table) response lists the additional operations that Unity Catalog permits for that table. Unity Catalog permits `DATA_REORGANIZATION`, `DATA_CLEANUP`, and `METADATA_CLEANUP` for managed tables. `DATA_REORGANIZATION` rewrites data files that the current table version uses, without changing the current table's rows. `DATA_CLEANUP` deletes old data files that the current table version does not use, after the retention period. `METADATA_CLEANUP` deletes expired files from `_delta_log` and keeps the files that are necessary to read retained table versions. A missing or empty list permits no additional operations. A client must not perform an operation that it does not recognize. Statistics collection that writes results into the catalog (e.g., ANALYZE) is unsupported, as this protocol version defines no API for it.
```
tdas: Update names accordingly.

tdas (changes requested):

api/delta.yaml
timothyw553: updated to group, cluster, layout

api/delta.yaml
timothyw553: makes sense. removed that last part about the retention period

api/delta.yaml
timothyw553: done. made it more general/expanded

spec/protocols/ManagedTablesSpec.md
timothyw553: updated to `allowed-maintenance-operations`, `getAllowedMaintenanceOperations`, etc..

tdas (approved):

tdas (approved):

PR #1815, opened 2026-08-30
[AI] [#1816] Add service credentials and secrets support to UC Python UDFs

grundprinzip:
Unity Catalog Python UDFs can declare service credentials and secrets that their body reads at execution time, but `DatabricksFunctionClient` had no way to emit either clause. Functions needing an external credential or an API key could not be created through this library at all.

Changes:

- Add `credentials` and `secrets` to `create_python_function` and `create_wrapped_function`. Credentials are declared as a credential name or a `ServiceCredential`, which also carries an alias and a default marker; secrets are three-part `catalog.schema.key` names.
- Validate both client-side, mirroring the server's own rules: at most one default credential, no duplicate references, three-part secret names, and an `environment_version` of at least 6 for secrets. The version Unity Catalog picks implicitly is below that minimum, so an explicit version is required.
- Backtick-quote every emitted identifier, so hyphenated credential names and secret keys parse.
- Validate dependencies, which are interpolated into a single-quoted SQL string literal: reject embedded single quotes, and check the shape of `/Volumes/` paths so a malformed one fails locally rather than at create time. The same quoting check now covers `environment_version`.
- Add `get_function_environment`, `get_function_credentials`, and `get_function_secrets`, which read these declarations back out of the JSON-encoded `FunctionInfo.properties` field. Unity Catalog exposes no typed field for them, and the values are themselves JSON-encoded.
- Reject local sandbox execution of a function that declares credentials or secrets. The sandbox has no Unity Catalog securable service, so the body otherwise dies with an opaque signal rather than a usable error.

`generate_sql_function_body` is now called with keyword arguments, as its sibling already was.

Verified end to end against a workspace: a function declaring a dependency, an aliased default credential, and a secret was created, read back with all three declarations intact, and executed on serverless.

Tests: `pytest tests/core` (352 passed; `test_local_executor.py` fails identically on an unmodified tree in this environment, and `tests/core/oss` needs a local server).

**PR Checklist**

**Description of changes**

### DIFF
--- a/ai/core/src/unitycatalog/ai/core/utils/callable_utils.py
+++ b/ai/core/src/unitycatalog/ai/core/utils/callable_utils.py
@@ -26,6 +26,10 @@ from unitycatalog.ai.core.utils.type_utils import (
 FORBIDDEN_PARAMS = ["self", "cls"]
 PRIMARY_INDENT = " " * 4
 WRAPPED_INDENT = " " * 6
+# Mirrors the server-side minimum environment version that ships the `databricks.secrets`
+# module. Unity Catalog is the authority on this value; it is duplicated here only to fail
+# fast with an actionable message instead of after a round trip.
+SECRETS_MIN_ENVIRONMENT_VERSION = 6
 
 
 class FunctionBodyExtractor(ast.NodeVisitor):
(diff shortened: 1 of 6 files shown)


ai/core/src/unitycatalog/ai/core/utils/callable_utils.py
```diff
@@ -389,6 +415,84 @@ def process_parameter(
...
+    Args:
+        environment_version: The environment version to validate.
+
+    Raises:
+        ValueError: If the version contains a single quote or a backslash.
+    """
+    for character, name in (("'", "a single quote"), ("\\", "a backslash")):
+        if character in environment_version:
+            raise ValueError(
+                f"environment_version cannot contain {name}: {environment_version!r}. It is "
```
smurching, line 491: Super nit suggestion on the error message just to make the causal relationship here clear (I got a bit confused by it on first read). Or we can just remove the reason so users can focus on the actionable piece (remove the offending char(s))

```suggestion
                f"environment_version cannot contain {name}: {environment_version!r}, because it is "
```

smurching (approved):
LGTM, thanks @grundprinzip !

tdas (approved):
LGTM. Thank you @grundprinzip and @smurching for reviewing.

tdas:
@grundprinzip just waiting for you to resolve this comment from @smurching - https://github.com/unitycatalog/unitycatalog/pull/1815#discussion_r3905819003

tdas:
@yili-db can you review why this the spark unit tests are failing?

yili-db:
> @yili-db can you review why this the spark unit tests are failing?

The error was: 
> X The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
> Spark tests (Spark 4.1.0, Delta 4.3.1): .github#1

The "server" is the github actions backend server. So this is very likely a github infra flaky. I'll rerun this test.

yili-db:
@tdas Yes it's a github infra flakiness indeed. The test passed when retried.

PR #1821, opened 2026-09-01
[Server] Validate OIDC discovery responses

SauronShepherd:
## Summary

This PR validates OIDC discovery responses before constructing a JWKS provider and converts malformed discovery data into controlled OAuth errors.

## Problem

`JwksOperations` previously consumed discovery responses without consistently validating the HTTP status, JSON shape, and required metadata. Non-success responses, invalid JSON, missing or incorrectly typed `issuer`/`jwks_uri`, blank metadata, or an issuer mismatch could result in uncontrolled parsing, null/type, or downstream failures.

## Changes

- Check for a successful discovery HTTP response before parsing the body.
- Convert invalid discovery JSON into `OAuthInvalidRequestException`.
- Require non-empty string values for `issuer` and `jwks_uri`.
- Require the discovery issuer to exactly match the configured issuer.
- Preserve existing issuer semantics; this PR does not add URL normalization, audience-policy changes, or provider caching.

## Compatibility and rationale

Valid discovery configurations continue to work unchanged. The exact issuer comparison is consistent with the current documented configuration contract and existing exact-match behavior for non-wildcard issuers. Equivalent URL normalization, such as trailing-slash handling, is intentionally left for a separate review if maintainers want to define that behavior.

## Tests

Executed on Windows:

```text
build\sbt "server/testOnly io.unitycatalog.server.utils.JwksOperationsTest"
build\sbt "server/compile"
```

The tests cover valid metadata, missing fields, wrong types, blank values, issuer mismatch, HTTP 500, and invalid JSON. Both commands completed successfully.

## Scope

Files changed:

- `server/src/main/java/io/unitycatalog/server/utils/JwksOperations.java`
- `server/src/test/java/io/unitycatalog/server/utils/JwksOperationsTest.java`

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/utils/JwksOperations.java
+++ b/server/src/main/java/io/unitycatalog/server/utils/JwksOperations.java
@@ -8,6 +8,7 @@ import com.auth0.jwk.JwkProviderBuilder;
 import com.auth0.jwt.JWT;
 import com.auth0.jwt.JWTVerifier;
 import com.auth0.jwt.algorithms.Algorithm;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.linecorp.armeria.client.WebClient;
@@ -26,7 +27,7 @@ import org.slf4j.LoggerFactory;
 
 public class JwksOperations {
 
-  private final WebClient webClient = WebClient.builder().build();
+  private final WebClient webClient;
   private static final ObjectMapper mapper = new ObjectMapper();
   private final SecurityContext securityContext;
 
@@ -34,6 +35,7 @@ public class JwksOperations {
 
   public JwksOperations(SecurityContext securityContext) {
     this.securityContext = securityContext;
+    this.webClient = WebClient.builder().build();
   }
 
   @SneakyThrows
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/unitycatalog/server/utils/JwksOperations.java
```diff
@@ -95,35 +97,55 @@ public JwkProvider loadJwkProvider(String issuer) {
...
 
-      // TODO: Or maybe just cache the provider for reuse.
+      String configJwksUri = validateIssuerConfiguration(configMap, issuer);
+
+      // Do not cache this provider: its rate limiter is scoped to the provider instance.
       return new JwkProviderBuilder(URI.create(configJwksUri).toURL()).cached(false).build();
     }
   }
+
+  static String validateIssuerConfiguration(Map<String, Object> configMap, String issuer) {
```
yili-db, line 134: This function should be names as `getValidatedJwksUri` for example as it returns the URI.

server/src/main/java/io/unitycatalog/server/utils/JwksOperations.java
```diff
@@ -95,35 +97,55 @@ public JwkProvider loadJwkProvider(String issuer) {
...
       }
 
-      String configIssuer = (String) configMap.get("issuer");
-      String configJwksUri = (String) configMap.get("jwks_uri");
+      String responseBody = response.contentUtf8();
 
-      if (!configIssuer.equals(issuer)) {
-        throw new OAuthInvalidRequestException(ErrorCode.ABORTED,
-            "Issuer doesn't match configuration");
+      // Discovery is intentionally performed per verifier until metadata caching can be made
```
yili-db, line 112: This new comment and L129 worry me — they state a *why* that doesn't match what the code does. `.cached(false)` disables jwk **key** caching inside the provider; it's not "don't cache the provider" (we already rebuild a new one every call), and it's unrelated to the rate limiter — that's the separate `.rateLimited()` decorator, still default-on here. and metadata caching doesn't really couple to key rotation.
Can we just keep the old TODOs, or say only what's true? A comment asserting the wrong mechanism is worse than no comment — the next person designs around a constraint that isn't there.

Side note: since the rate limiter is per-instance, building a fresh provider each call resets its bucket every time, so the 10/min limit is basically not enforced. Probably worth a follow-up rather than this PR.

server/src/test/java/io/unitycatalog/server/utils/JwksOperationsTest.java
```diff
@@ -0,0 +1,105 @@
...
+        "/.well-known/openid-configuration",
+        exchange -> {
+          exchange.sendResponseHeaders(500, -1);
+          exchange.close();
+        });
+
+    assertThatThrownBy(
+            () ->
+                new JwksOperations(null)
+                    .loadJwkProvider("http://localhost:" + server.getAddress().getPort()))
```
yili-db, line 84: No happy path test for `loadJwkProvider`?

server/src/main/java/io/unitycatalog/server/utils/JwksOperations.java
```diff
@@ -95,35 +97,55 @@ public JwkProvider loadJwkProvider(String issuer) {
...
 
-      // TODO: We should cache this. No need to fetch it each time.
-      Map<String, Object> configMap = mapper.readValue(response, new TypeReference<>() {});
-
-      if (configMap == null || configMap.isEmpty()) {
-        throw new OAuthInvalidRequestException(ErrorCode.ABORTED,
-            "Could not get issuer configuration");
+      if (!response.status().isSuccess()) {
+        throw new OAuthInvalidRequestException(
+            ErrorCode.ABORTED, "Could not get issuer configuration: HTTP " + response.status());
```
yili-db, line 107: The original error is worth debug logging to help diagnosis of problem by service admin.
Then the HTTP response code of the provider isn't the best thing to expose to RPC callers.

yili-db (commented):
This is a nice change. A few minor comments. And also please rebase.
Thank you!

PR #1824, opened 2026-09-02
[Server] Widen uc_properties.property_value beyond varchar(255)

s-maling-celonis:
## Summary
- Hibernate maps an unannotated `String` as `varchar(255)`. Any `uc_properties.property_value` longer than that fails with `ERROR: value too long for type character varying(255)` on `PropertyDAO` insert.
- [#1830](https://github.com/unitycatalog/unitycatalog/pull/1830) stopped Spark from persisting Hive-style schema JSON (`spark.sql.sources.schema.part.N`) on `createTable` / `createView`. That does **not** cover every long value: Spark still forwards `view.sqlConfig.*` and other user `TBLPROPERTIES`. Those still hit `varchar(255)`.
- Map `property_value` with the same `length = 16777215` used for `ColumnInfoDAO.typeJson`.
- **Existing deployments are not updated by this change.** The server uses `hibernate.hbm2ddl.auto=update`, which creates missing tables/columns but does **not** change the type or length of a column that already exists. Upgrading the server JAR alone leaves `varchar(255)` in place. Operators must `ALTER` `uc_properties.property_value` (statements in `docs/server/deployment.md`).

## Example that fails today

Delta `CREATE TABLE` with a wide schema succeeds: table schema is stored on `uc_columns` (`type_json`), not as a table property.

`CREATE VIEW` then calls `TablesApi.createTable` with Spark view properties (including `view.sqlConfig.*` and user `TBLPROPERTIES`). A value longer than 255 characters 500s the insert even though the view SQL itself is stored on `uc_tables.view_definition` (already a wide column).

```text
io.unitycatalog.client.ApiException: createTable call failed with: 500
{"error_code":"INTERNAL","message":"Error creating table: cat.sch.activity_union: could not execute statement [ERROR: value too long for type character varying(255)] [/* insert for io.unitycatalog.server.persist.dao.PropertyDAO */ insert into uc_properties (entity_id,entity_type,property_key,property_value,id) values (?,?,?,?,?)]"}
```

The same 255-char limit still applies to any REST client that sends a long table/view property.

## Test plan
- [ ] `SdkViewCRUDTest` / `BaseViewCRUDTest.testCreateViewAcceptsPropertyValueLongerThanDefaultVarchar` persists a user property longer than 255 characters
- [ ] Confirm a **new** database gets the wider column from Hibernate
- [ ] Confirm an **existing** database still has `varchar(255)` until the documented `ALTER` is run, then `CREATE VIEW` with a long property succeeds

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableCRUDTest.java
+++ b/server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableCRUDTest.java
@@ -99,6 +99,49 @@ public class SdkTableCRUDTest extends BaseTableCRUDTest {
         .withMessageContaining("s3://bucket/");
   }
 
+  /**
+   * Spark stores datasource schema JSON in {@code spark.sql.sources.schema.part.N}. That value is
+   * larger than Hibernate's default {@code varchar(255)} as soon as the table has more than a
+   * couple of columns.
+   */
+  @Test
+  public void testCreateTableAcceptsPropertyValueLongerThanDefaultVarchar() throws Exception {
+    String schemaJson =
+        "{\"type\":\"struct\",\"fields\":["
+            + "{\"name\":\"col1\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}},"
+            + "{\"name\":\"col2\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}},"
+            + "{\"name\":\"col3\",\"type\":\"double\",\"nullable\":true,\"metadata\":{}},"
+            + "{\"name\":\"col4\",\"type\":\"long\",\"nullable\":true,\"metadata\":{}},"
+            + "{\"name\":\"col5\",\"type\":\"boolean\",\"nullable\":true,\"metadata\":{}},"
+            + "{\"name\":\"col6\",\"type\":\"timestamp\",\"nullable\":true,\"metadata\":{}},"
+            + "{\"name\":\"col7\",\"type\":\"decimal(18,2)\",\"nullable\":true,\"metadata\":{}}"
+            + "]}";
+    assertThat(schemaJson.length()).isGreaterThan(255);
+
+    String tableName = "long_property_table";
+    TableInfo created =
+        tableOperations.createTable(
+            new CreateTable()
+                .name(tableName)
+                .catalogName(TestUtils.CATALOG_NAME)
+                .schemaName(TestUtils.SCHEMA_NAME)
+                .columns(COLUMNS)
+                .properties(Map.of("spark.sql.sources.schema.part.0", schemaJson))
+                .comment(TestUtils.COMMENT)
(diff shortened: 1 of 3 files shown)


yili-db:
The property of `spark.sql.sources.schema.part.N` seems a dup of `ColumnInfo.typeJson`. And in UC tables it doesn't seem to be useful at all.
An easier fix is to strip `spark.sql.sources.schema.*` properties before creating the table.

s-maling-celonis:
Superseded by https://github.com/unitycatalog/unitycatalog/pull/1830.

The parquet CREATE TABLE failure was Spark persisting Hive-metastore schema JSON (`spark.sql.sources.schema.part.N`) into `uc_properties`, not a column-width problem. #1830 strips those keys (and the other non-metadata properties) in the Spark connector instead of widening `property_value`.

s-maling-celonis:
@yili-db reopening this.

#1830 was the right fix for the parquet `CREATE TABLE` case (`spark.sql.sources.schema.part.N`). The same `varchar(255)` overflow just hit again on a normal Spark `CREATE VIEW`: `UCProxyViewSupport.createView` still persists `view.sqlConfig.*` (and other view properties) into `uc_properties`, and a value longer than 255 characters 500s the insert even though the view SQL itself is stored on `uc_tables.view_definition`.

Delta `CREATE TABLE` for the same wide schema succeeded; only the subsequent `CREATE VIEW` failed. Stripping schema JSON does not cover that path, so we still need a wider `property_value`.

yili-db (approved):

server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableCRUDTest.java
```diff
@@ -99,6 +99,49 @@ public void testCreateExternalTableRejectsCloudStorageRoot() {
         .withMessageContaining("s3://bucket/");
   }
 
+  /**
```
yili-db: I think this test should be folded into existing test of `BaseViewCRUDTest`

yili-db (changes requested):

server/src/test/java/io/unitycatalog/server/sdk/tables/SdkTableCRUDTest.java
s-maling-celonis: Agreed. The parquet `spark.sql.sources.schema.part.N` test was the old #1830 story; Spark no longer persists that key.

Moved the coverage into `BaseViewCRUDTest.testCreateViewAcceptsPropertyValueLongerThanDefaultVarchar`: create/get a view with a `user.note` property longer than 255 characters. Also updated `docs/server/deployment.md` so the ALTER is justified by table/view properties (`TBLPROPERTIES`, `view.sqlConfig.*`) rather than schema JSON.

yili-db (approved):

PR #1826, opened 2026-09-03
[Python] Remove OpenAI upper version bound

B-Step62:
**PR Checklist**

- [x] If there is a related issue, make sure it is linked to this PR. (No related issue.)
- [x] If you've fixed a bug or added code that should be tested, add tests! (Dependency metadata only; lockfile validation is listed below.)
- [x] If you've added or modified a feature, documentation in `docs` is updated. (No feature change.)

**Description of changes**

Remove the `openai<=2.24.0` upper bound from `unitycatalog-openai` while retaining the existing minimum supported version. The upper bound reflected a point-in-time dependency cutoff rather than a known compatibility constraint, and now prevents consumers from using newer OpenAI SDK releases.

The existing `exclude-newer` setting and lockfile continue to provide reproducible development and CI resolution.

Validation:

- `UV_DEFAULT_INDEX=https://pypi.org/simple UV_SYSTEM_CERTS=true uv lock --check`
- `git diff --check`

### DIFF
--- a/ai/integrations/openai/pyproject.toml
+++ b/ai/integrations/openai/pyproject.toml
@@ -9,7 +9,7 @@ readme = "README.md"
 license = { text="Apache-2.0" }
 requires-python = ">=3.10"
 dependencies = [
-    "openai>=1.46.1,<=2.24.0",
+    "openai>=1.46.1",
     "unitycatalog-ai>=0.4.0,<0.5.0",
     "pydantic<3,>=2",
 ]
--- a/ai/integrations/openai/pyproject.toml
+++ b/ai/integrations/openai/pyproject.toml
@@ -74,7 +74,7 @@ docstring-code-line-length = 88
 convention = "google"
 
 [tool.uv]
-exclude-newer = "2026-03-10T00:00:00Z"
+exclude-newer = "30d"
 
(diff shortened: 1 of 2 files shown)


B-Step62:
@openinx Can you TAL this PR? The stale cap is causing dependency conflict for a customer.

openinx:
Hi @B-Step62, thanks for the contribution.  I think we will need several changes to advance the version while continue to preventing us from the supply chain attack. 

* we will need to use `uv` to lock the deps, and sync the deps if we want to upgrade the `openai` versions. 
* advance the `exclude-newer` to recent 30days cooldown,  so that we can advance the existing dep versions while get ride of supply chain attack. 
```
[tool.uv]
exclude-newer = "30d"
```

* ideally, we should use the safe pipe repository which already went through the daily security scanning.  but that can be a long term change.

B-Step62:
@openinx Thanks for the response. The https://github.com/unitycatalog/unitycatalog/tree/main/ai/integrations/openai directory already contain uv lock and `exclude-newer` guardrail. The `exclude-newer` is hard coded to a fixed date, which can be updated to `30d` as you suggested. Let me update two files.

PR #1828, opened 2026-09-03
[Server] Map commit lock contention to `COMMIT_STATE_UNKNOWN`, not a conflict

yili-db:
**PR Checklist**

**Description of changes**

- `RepositoryUtils.lockTableForCommit` now throws `COMMIT_STATE_UNKNOWN` (retryable 500) instead of `UPDATE_REQUIREMENT_CONFLICT` (409) when the row lock can't be acquired.
  - The failed request rolls back, but the lock holder may be this same commit's own earlier attempt (a slow client retry) that goes on to land, so the commit's outcome is unknown.
  - A 409 would let the client rebase and re-commit the same change; instead it resends the identical request, which dedup resolves to the true outcome.
- Also catch `jakarta.persistence.LockTimeoutException`, which a lock-wait can raise instead of `PessimisticLockException`.
- Keeps a 409 meaning the commit definitely did not land.
- Add `RepositoryUtilsTest` covering the lock-failure mapping and non-lock passthrough.

### DIFF
--- a/connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
+++ b/connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
@@ -78,7 +78,9 @@ public class PathCredentialReadWriteTest extends BaseSparkIntegrationTest {
     return createPathCredSession(false, false, true, ucCatalogs);
   }
 
-  /** Same layout as {@link #createPathCredSession} but omits {@code vendPathCredentials.enabled}. */
+  /**
+   * Same layout as {@link #createPathCredSession} but omits {@code vendPathCredentials.enabled}.
+   */
   private SparkSession createPathCredSessionLeavingVendFlagUnset(String... ucCatalogs) {
     SparkSession.Builder builder =
         SparkSession.builder()
--- a/server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
+++ b/server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
@@ -30,7 +30,16 @@ public class RepositoryUtils {
 
   /**
    * Acquires the table row lock shared by Delta and Iceberg commit paths. Lock waits and deadlock
-   * victims are reported as retryable requirement conflicts instead of leaking ORM exceptions.
+   * victims are reported as {@code COMMIT_STATE_UNKNOWN} (a retryable, unknown-outcome error)
+   * instead of leaking ORM exceptions.
+   *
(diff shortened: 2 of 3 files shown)


connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -78,7 +78,9 @@ private SparkSession createPathCredSession(String... ucCatalogs) {
     return createPathCredSession(false, false, true, ucCatalogs);
   }
 
-  /** Same layout as {@link #createPathCredSession} but omits {@code vendPathCredentials.enabled}. */
+  /**
+   * Same layout as {@link #createPathCredSession} but omits {@code vendPathCredentials.enabled}.
```
yili-db, line 82: `sbt javafmtAll` insisted reformatting this line.

server/src/main/java/io/unitycatalog/server/persist/utils/RepositoryUtils.java
```diff
@@ -30,19 +30,29 @@ public class RepositoryUtils {
...
+   * resolves to the true outcome.
    */
   public static void lockTableForCommit(
       Session session, TableInfoDAO dao, UUID tableId, Optional<String> tableFullNameForLogging) {
     try {
       session.refresh(dao, LockMode.PESSIMISTIC_WRITE);
     } catch (RuntimeException e) {
       if (!(e instanceof org.hibernate.PessimisticLockException)
-          && !(e instanceof jakarta.persistence.PessimisticLockException)) {
+          && !(e instanceof jakarta.persistence.PessimisticLockException)
```
timothyw553, line 50: we should also handle Hibernate’s native lock exceptions:
```diff
 if (!(e instanceof org.hibernate.PessimisticLockException)
+    && !(e instanceof org.hibernate.exception.LockAcquisitionException)
+    && !(e instanceof org.hibernate.exception.LockTimeoutException)
     && !(e instanceof jakarta.persistence.PessimisticLockException)
     && !(e instanceof jakarta.persistence.LockTimeoutException)) {
```
and add tests.

timothyw553 (changes requested):
one comment. and also should we add a protocol-specific mapping in [IcebergRestExceptionHandler](https://github.com/unitycatalog/unitycatalog/blob/main/server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java#L61) such as:
```diff
+case COMMIT_STATE_UNKNOWN ->
+    wrapException(
+        ErrorCode.COMMIT_STATE_UNKNOWN,
+        baseException.getErrorMessage(),
+        new CommitStateUnknownException(baseException));
 case UPDATE_REQUIREMENT_CONFLICT ->
```

PR #1830, opened 2026-09-04
[Spark] Do not persist Spark schema JSON, credentials, or reserved V2 keys

s-maling-celonis:
## Summary
- Spark packages `StructType` JSON into table properties (`spark.sql.sources.schema`, `spark.sql.sources.schema.part.N`, partition/bucket lists, and `spark.sql.partitionSchema*`) so Hive metastore can reconstruct schema. UC already stores schema on `uc_columns`, and `schema.part.N` routinely exceeds `uc_properties.property_value` (`varchar(255)`), so `CREATE TABLE … USING PARQUET` 500s for anything but a tiny schema.
- One deny-list, `UCTableProperties.shouldPersistProperty`, is used by `UCProxy.createTable` and Spark 4.2 `createView`. It drops reserved V2 keys, vended `fs.*` credentials, and Spark schema JSON, including the `option.`-prefixed copies Spark adds from OPTIONS.
- Load still rebuilds Spark schema from columns. View-only metadata (`view.sqlConfig.*`, schema mode, query column names) is still written after the filter.
- Alternative to widening `property_value` for this Spark path ([#1824](https://github.com/unitycatalog/unitycatalog/pull/1824)). User `TBLPROPERTIES` and REST clients with long values are unchanged.

## Test plan
- [x] `UCTablePropertiesTest` for schema keys, `option.` copies, `fs.*`, reserved V2 keys, and keys that must still persist (user props, `view.sqlConfig.*`)
- [x] `ParquetExternalTableReadWriteTest.testCreateParquetTableDoesNotPersistSparkDatasourceSchema` — seven-column parquet CREATE (the case that overflowed 255 chars), columns persisted, no schema property keys, `SELECT *` works
- [x] `ExternalTableReadWriteTest.testCreateExternalTableSendsNoCredentialProperties` asserts every stored key passes `shouldPersistProperty`
- [ ] CI `spark` tests (4.0 / 4.1 / 4.2)

### DIFF
--- a/connectors/spark/src/main/java/io/unitycatalog/spark/UCTableProperties.java
+++ b/connectors/spark/src/main/java/io/unitycatalog/spark/UCTableProperties.java
@@ -35,4 +35,23 @@ public class UCTableProperties {
           TableCatalog.PROP_OWNER,
           TableCatalog.PROP_PROVIDER,
           PROP_TABLE_TYPE);
+
+  // Spark HiveExternalCatalog packages StructType JSON into TBLPROPERTIES so Hive metastore
+  // (which has no first-class nested schema) can reconstruct the table. Keys are
+  // `spark.sql.sources.schema`, `spark.sql.sources.schema.numParts`,
+  // `spark.sql.sources.schema.part.N`, partition/bucket column lists, and the split
+  // `spark.sql.partitionSchema*` form. UC already stores schema on `uc_columns`, so these
+  // are redundant -- and `schema.part.N` routinely exceeds the server's varchar(255)
+  // property value. Strip them on create; do not persist them.
+  public static final String SPARK_DATASOURCE_SCHEMA = "spark.sql.sources.schema";
+  public static final String SPARK_DATASOURCE_SCHEMA_PREFIX = SPARK_DATASOURCE_SCHEMA + ".";
+  public static final String SPARK_PARTITION_SCHEMA = "spark.sql.partitionSchema";
+  public static final String SPARK_PARTITION_SCHEMA_PREFIX = SPARK_PARTITION_SCHEMA + ".";
+
+  public static boolean isSparkDatasourceSchemaProperty(String key) {
+    return key.equals(SPARK_DATASOURCE_SCHEMA)
+        || key.startsWith(SPARK_DATASOURCE_SCHEMA_PREFIX)
+        || key.equals(SPARK_PARTITION_SCHEMA)
+        || key.startsWith(SPARK_PARTITION_SCHEMA_PREFIX);
+  }
 }
--- a/connectors/spark/src/main/java/io/unitycatalog/spark/UCTableProperties.java
+++ b/connectors/spark/src/main/java/io/unitycatalog/spark/UCTableProperties.java
@@ -48,10 +48,38 @@ public class UCTableProperties {
   public static final String SPARK_PARTITION_SCHEMA = "spark.sql.partitionSchema";
   public static final String SPARK_PARTITION_SCHEMA_PREFIX = SPARK_PARTITION_SCHEMA + ".";
 
+  /**
+   * Spark copies USING/OPTIONS entries both bare and with {@link TableCatalog#OPTION_PREFIX}
+   * ({@code option.}). Credential and schema keys must match in either form.
+   */
+  private static String withoutOptionPrefix(String key) {
+    if (key.startsWith(TableCatalog.OPTION_PREFIX)) {
+      return key.substring(TableCatalog.OPTION_PREFIX.length());
+    }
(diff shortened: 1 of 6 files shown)


connectors/spark/src/main/java/io/unitycatalog/spark/UCTableProperties.java
```diff
@@ -35,4 +35,51 @@ private UCTableProperties() {}
...
+  }
+
+  /**
+   * Keys that must not be forwarded on {@code createTable} / {@code createView}. Reserved V2 fields
+   * are first-class columns on the UC payload; {@code fs.*} is session-scoped credential config;
+   * Spark datasource schema JSON duplicates {@code uc_columns} and overflows varchar(255).
+   */
+  public static boolean shouldPersistProperty(String key) {
+    String bare = withoutOptionPrefix(key);
+    if (V2_TABLE_PROPERTIES.contains(key) || V2_TABLE_PROPERTIES.contains(bare)) {
```
yili-db: `V2_TABLE_PROPERTIES.contains(key)` is dead code because `V2_TABLE_PROPERTIES` does not contain any key with `option.` prefix and `withoutOptionPrefix` returns the input `key` if it doesn't start with `option.`

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
```diff
@@ -167,10 +167,7 @@ public void testCreateExternalTableSendsNoCredentialProperties(
     TableInfo tableInfo = tableOperations.getTable(fullTableName);
     java.util.Map<String, String> serverProperties =
         tableInfo.getProperties() == null ? java.util.Map.of() : tableInfo.getProperties();
-    assertThat(serverProperties.keySet())
-        .noneMatch(key -> key.startsWith("fs."))
-        .noneMatch(key -> key.startsWith(TableCatalog.OPTION_PREFIX + "fs."))
-        .noneMatch(UCTableProperties.V2_TABLE_PROPERTIES::contains);
+    assertThat(serverProperties.keySet()).allMatch(UCTableProperties::shouldPersistProperty);
```
yili-db: This test becomes circular: it calls the same function that does the real work in test.
The original check is better. You can add extra checks on keys that would otherwise appear in the `serverProperties`: `spark.sql.partitionSchema.something` etc.

connectors/spark/src/test/java/io/unitycatalog/spark/ParquetExternalTableReadWriteTest.java
```diff
@@ -1,8 +1,52 @@
...
+                        Pair.of("col4", "BIGINT"),
+                        Pair.of("col5", "BOOLEAN"),
+                        Pair.of("col6", "TIMESTAMP"),
+                        Pair.of("col7", "DECIMAL(18, 2)"))));
+
+    TableInfo tableInfo = tableOperations.getTable(fullTableName);
+    Map<String, String> serverProperties =
+        tableInfo.getProperties() == null ? Map.of() : tableInfo.getProperties();
+    assertThat(serverProperties.keySet())
+        .noneMatch(UCTableProperties::isSparkDatasourceSchemaProperty);
```
yili-db: ditto.

connectors/spark/src/main/java/io/unitycatalog/spark/UCTableProperties.java
s-maling-celonis: Good catch — `contains(key)` was redundant once we normalize with `withoutOptionPrefix`. Dropped it so the V2 check is only `V2_TABLE_PROPERTIES.contains(bare)`.

connectors/spark/src/test/java/io/unitycatalog/spark/ExternalTableReadWriteTest.java
s-maling-celonis: Agreed — asserting `shouldPersistProperty` here was circular. Restored the original `fs.*` / `option.fs.*` / reserved V2 checks, and added explicit prefix matchers for `spark.sql.sources.schema*` and `spark.sql.partitionSchema*` (including `option.` copies) so this IT does not go through the persist filter.

connectors/spark/src/test/java/io/unitycatalog/spark/ParquetExternalTableReadWriteTest.java
s-maling-celonis: Same here: this test now asserts those schema keys with literal prefixes instead of `isSparkDatasourceSchemaProperty`.

yili-db (approved):

PR #1836, opened 2026-09-05
[Server] Answer unrouted Iceberg requests with an Iceberg error document

### LINKED ISSUE #1835: Iceberg REST endpoints answer unimplemented paths with a body Iceberg clients cannot parse
**Describe the bug**

A request to an Iceberg REST path this server does not implement never reaches
`IcebergRestCatalogService`, so `IcebergRestExceptionHandler` never sees it and Armeria answers with
its own plain-text body -- a `Status:` line followed by a `Description:` line, sent as
`text/plain`. Iceberg's `ErrorResponseParser` cannot read that, so where the REST spec puts an error
document the client gets an unparsable body and fails with a parse error instead of the
`NotFoundException` or `RESTException` the status calls for.

Clients that read the `endpoints` list from `GET /v1/config` never send such a request. Clients
written before that field existed, or that do not consult it, do: `dropNamespace` and `listViews` are
both unimplemented here and neither is advertised.

**To Reproduce**

1. Start the server and create a catalog, a schema, and a table exposed through the Iceberg REST
   catalog.
2. `GET /api/2.1/unity-catalog/iceberg/v1/catalogs/{catalog}/namespaces/{schema}/views` -- a path
   the server does not serve. Response:
   `404`, `Content-Type: text/plain; charset=utf-8`, body `Status: 404\nDescription: Not Found\n`
3. `DELETE /api/2.1/unity-catalog/iceberg/v1/catalogs/{catalog}/namespaces/{schema}` -- a method
   that path does not accept. Response:
   `405`, `Content-Type: text/plain; charset=utf-8`, body `Status: 405\nDescription: Method Not Allowed\n`
4. Hand either body to `org.apache.iceberg.rest.responses.ErrorResponseParser.fromJson`, which is
   what the Iceberg client does with an error response: it throws instead of producing an
   `ErrorResponse`.

**Expected behavior**

An error response from a path under the Iceberg REST API is an Iceberg `ErrorResponse` document, as
it already is for every request that reaches the service, e.g.
`{"error":{"message":"Not Found","type":"NotFoundException","code":404}}`. Paths outside the Iceberg
API are unaffected.

**Additional context**

Verified on `main` at 58d5c7b. The statuses and messages Armeria chooses are correct; only the
rendering is wrong, so this is about the body, not about implementing the missing endpoints.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1835.

A request to an Iceberg REST path this server does not implement never reaches
`IcebergRestCatalogService`, so `IcebergRestExceptionHandler` never sees it and Armeria renders the
status itself as `text/plain` -- `Status: 404` and a `Description:` line. Iceberg's
`ErrorResponseParser` cannot read that, so a client that does not consult the advertised `endpoints`
list gets an unparsable body where the REST spec puts an error document.

This registers a `ServerErrorHandler` on the Armeria server that renders those status-only responses
as an Iceberg `ErrorResponse`, and only for paths under the Iceberg API:

- `onServiceException` returns null, so an exception raised by a service is still rendered by that
  service's own handler; nothing about the existing error paths changes.
- `renderStatus` returns null unless the request path starts with the Iceberg API prefix, so
  non-Iceberg paths and `/` keep Armeria's rendering.
- The prefix is built from the same `ICEBERG_RELATIVE_PATH` constant `UnityCatalogServer` mounts the
  service with, so the mount point and the handler cannot drift apart. It matches on a slash
  boundary, and matches the mount point itself as well as the paths under it, so
  `GET /api/2.1/unity-catalog/iceberg` is covered while a sibling path that merely starts with the
  same characters is not.
- 404 is named `NotFoundException`; the remaining statuses have no Iceberg exception of their own and
  are named `RESTException`, which is what Iceberg's client raises for them.

For users: an Iceberg client that calls an endpoint this catalog does not implement now fails with
the exception the status means, instead of a parse error. Responses from implemented endpoints are
byte-for-byte unchanged.

Verified on `main` at 58d5c7b. Before, `GET .../namespaces/{ns}/views` answered
`text/plain` `Status: 404\nDescription: Not Found\n` and `DELETE .../namespaces/{ns}` answered
`Status: 405\nDescription: Method Not Allowed\n`; after, they answer
`{"error":{"message":"Not Found","type":"NotFoundException","code":404}}` and
`{"error":{"message":"Method Not Allowed","type":"RESTException","code":405}}` as
`application/json`. `testUnroutedIcebergRequestsAnswerWithAnIcebergError` fails on an unfixed tree
(`expected: application/json but was: text/plain; charset=utf-8`) and passes with the fix. It pins the
error `type` and `message` as well as the status -- `NotFoundException` for the 404 and
`RESTException` for the 405 -- the mount point itself, and two paths outside the Iceberg API: another
UC endpoint, and `/`, which still answers `Hello, Unity Catalog!`.
`IcebergRestCatalogTest` (10/10) and `ArmeriaServerBuilderTest` are green, and the full
`server/test` suite shows no failure attributable to this change.

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
+++ b/server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
@@ -62,6 +62,8 @@ import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.exceptions.BadRequestException;
 import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
+import org.apache.iceberg.exceptions.NotFoundException;
+import org.apache.iceberg.exceptions.RESTException;
 import org.apache.iceberg.expressions.Expressions;
 import org.apache.iceberg.metrics.CommitMetrics;
 import org.apache.iceberg.metrics.CommitMetricsResult;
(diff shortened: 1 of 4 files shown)


server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
```diff
@@ -953,6 +955,65 @@ public void testListTablesReturnsTablesBeyondTheFirstPage()
         .containsExactly(TableIdentifier.of(Namespace.of(TestUtils.SCHEMA_NAME), "uniform_table"));
   }
 
+  @Test
+  public void testUnroutedIcebergRequestsAnswerWithAnIcebergError() throws Exception {
+    createUniformIcebergTable();
```
yili-db: This setup looks unnecessary since none of the RPC calls below hits any actual endpoint.

yili-db (commented):
This is a good fix. Just a minor comment.

Thank you!

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
moomindani: Removed; you are right, every request there is answered before a service is reached, so nothing needs to exist. A comment says so instead.

yili-db (approved):

PR #1844, opened 2026-09-05
[Server] Honour the Iceberg snapshots parameter on loadTable

### LINKED ISSUE #1840: Iceberg REST loadTable ignores the snapshots parameter
**Describe the bug**

`GET /v1/{prefix}/namespaces/{namespace}/tables/{table}` ignores the spec's `snapshots` parameter. The
spec defines two values: `all`, the default, returns every snapshot the table metadata holds, and
`refs` returns only the snapshots the table's refs point at. `loadTable` takes no such parameter, so
the response is the whole metadata whatever the client asked for, and a misspelled value is accepted
just as silently.

An Iceberg client configured with `rest-snapshot-loading-mode=refs` sends `snapshots=refs` to keep a
table with a long history from arriving in full on every load. Against this server it gets the full
history anyway.

**To Reproduce**

1. Start the server and expose a table through the Iceberg REST catalog whose metadata holds a
   snapshot no ref points at.
2. `GET /api/2.1/unity-catalog/iceberg/v1/catalogs/{catalog}/namespaces/{schema}/tables/{table}?snapshots=refs`
   answers 200 with every snapshot in `metadata.snapshots`, including the unreferenced one.
3. `...?snapshots=some` answers 200 as well, rather than rejecting a value the spec does not define.

**Expected behavior**

`snapshots=refs` returns metadata with only the snapshots the refs point at, `snapshots=all` and an
absent parameter return everything, and any other value is a 400.

**Additional context**

Iceberg's own handler builds the `refs` response with
`TableMetadata.buildFrom(metadata).withMetadataLocation(...).suppressHistoricalSnapshots().build()`,
and rejects an unknown value with `IllegalArgumentException`.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1840.

`loadTable` took no `snapshots` parameter, so a client asking for `refs` was answered with the whole
metadata, and a value the spec does not define was accepted without a word.

The parameter is now read. For `refs` the response is built from metadata with the historical snapshots
suppressed, which is how Iceberg's own handler answers that mode; `all` and an absent parameter answer
as before. Any other value is a bad request, so a client asking for something this server does not
implement is told rather than handed a response that does not match its request.

For users: a client configured with `rest-snapshot-loading-mode=refs` gets what it asked for, which on
a table with a long history is a much smaller response. Clients that do not send the parameter see no
change.

`testLoadTableSnapshotsParameter` pins all three cases against a table whose metadata holds a snapshot
no ref points at (a new test fixture): `all` and no parameter return both snapshots, `refs` returns
only the referenced one while the rest of the metadata is unchanged, and an unknown value is a 400
typed `BadRequestException`. It fails on an unfixed tree and passes with the fix; the full
`server/test` suite shows no failure attributable to this change.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
@@ -85,6 +85,9 @@ public class IcebergRestCatalogService extends AuthorizedService implements Regi
 
   private static final String PREFIX_BASE = "catalogs/";
 
+  private static final String SNAPSHOTS_ALL = "all";
+  private static final String SNAPSHOTS_REFS = "refs";
+
   private static final List<Endpoint> READ_ENDPOINTS =
       List.of(
           Endpoint.V1_LIST_NAMESPACES,
@@ -276,7 +279,8 @@ public class IcebergRestCatalogService extends AuthorizedService implements Regi
   public LoadTableResponse loadTable(
       @Param("catalog") String catalog,
       @Param("namespace") String namespace,
-      @Param("table") String table) {
+      @Param("table") String table,
+      @Param("snapshots") Optional<String> snapshots) {
     TableRepository.IcebergTableState state =
         tableRepository.getIcebergTableState(catalog, namespace, table);
     if (state.metadataLocation() == null) {
@@ -287,6 +291,13 @@ public class IcebergRestCatalogService extends AuthorizedService implements Regi
     TableMetadata tableMetadata =
         metadataService.readTableMetadata(
             NormalizedURL.from(state.metadataLocation()), tableLocation);
+    if (refsOnly(snapshots)) {
+      tableMetadata =
+          TableMetadata.buildFrom(tableMetadata)
+              .withMetadataLocation(tableMetadata.metadataFileLocation())
+              .suppressHistoricalSnapshots()
+              .build();
+    }
     Map<String, String> config =
         tableConfigService.getTableConfig(tableLocation, getLoadCredentialPrivileges(state));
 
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -87,6 +87,9 @@ public class IcebergRestCatalogService extends AuthorizedService implements Regi
 
   private static final String PREFIX_BASE = "catalogs/";
 
+  private static final String SNAPSHOTS_ALL = "all";
```
yili-db: Are they `RESTCatalogProperties.SnapshotMode`?

yili-db (commented):
Nice change. Only a minor comment.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
moomindani: They are. Switched to `RESTCatalogProperties.SnapshotMode` and `RESTCatalogProperties.SNAPSHOTS_QUERY_PARAMETER`, so the parameter name and its values come from Iceberg rather than from local literals, and an unknown value falls out of `valueOf` into the same bad request. The mode travels lowercased, which is what the client sends, so the message names it that way too.

yili-db:
Wait for https://github.com/unitycatalog/unitycatalog/pull/1844 first.

moomindani:
I think this comment was meant for #1853: that one carries this change (its ETag has to cover the snapshot mode, so it uses the `SnapshotMode` added here), so #1853 is the PR that needs this one to go in first. Let me know if you meant something else.

yili-db:
> I think this comment was meant for #1853: that one carries this change (its ETag has to cover the snapshot mode, so it uses the `SnapshotMode` added here), so #1853 is the PR that needs this one to go in first. Let me know if you meant something else.

Right. This PR needs to go in first, then I'll review #1853 which depend on this one.

yili-db (approved):

yili-db (approved):
Please rebase.

moomindani:
Rebased onto `4ee2b248`.

PR #1845, opened 2026-09-05
[Server] Report an unreadable table metadata file as the failure it is

### LINKED ISSUE #1841: Iceberg REST answers an unreadable metadata file with a 500 typed NotFoundException
**Describe the bug**

When the metadata file a table points at cannot be read, the Iceberg REST endpoints answer 500 with
`"type": "NotFoundException"` and the reader's own message, which quotes the path the server keeps the
file at:

```json
{"error":{"message":"Failed to read file: /tmp/marksheet_uniform/metadata/00002-5b7aa739-d074-4764-b49d-ad6c63419576.metadata.json","type":"NotFoundException","code":500}}
```

The status and the type contradict each other. Iceberg raises `NotFoundException` for any file a table
points at, and `IcebergRestExceptionHandler` has no case for it, so the status comes from the internal
error the request became while the type comes from the cause. A client reading that type is told the
thing it asked for is missing -- something it could act on -- when the table is there and the server
could not read it. The message also puts the server's storage layout in a client-facing error.

**To Reproduce**

1. Start the server and expose a table through the Iceberg REST catalog.
2. Remove the metadata file the table's metadata pointer names (or point it at a file that is not
   there), leaving the table registered.
3. `GET /api/2.1/unity-catalog/iceberg/v1/catalogs/{catalog}/namespaces/{schema}/tables/{table}`
   answers the 500 above, typed `NotFoundException`.

**Expected behavior**

A 500 whose type says a server-side failure, and a message that says the table's Iceberg metadata could
not be read without quoting where the server keeps it. The reason belongs in the server's log.

**Additional context**

Iceberg's client picks the exception from the status, so a 500 raises `ServiceFailureException`
whatever the type says; the type is what makes the response contradict itself, and what a client
inspecting the body reads.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1841.

A table the catalog lists whose metadata file cannot be read came back as a 500 typed
`NotFoundException`, with the file reader's message and the path the server keeps the file at. Iceberg
raises that exception for any file a table points at, and this handler had no case for it, so the
status came from the internal error the request became while the type came from the cause.

`IcebergRestExceptionHandler` now answers it as a service failure, which is what the 500 already said,
with a message that says the table's Iceberg metadata could not be read. The original failure stays as
the cause, so the server's own logs still name the file and the reason.

For users: a client can tell "this table is gone" from "this server could not read this table", which
the previous response conflated, and the storage path no longer reaches the client.

`testLoadTableWhoseMetadataCannotBeRead` deletes the metadata file of a registered table and pins the
500, the `ServiceFailureException` type and the absence of the table's location from the message. It
fails on an unfixed tree and passes with the fix; the full `server/test` suite shows no failure
attributable to this change.


yili-db:
Wait for https://github.com/unitycatalog/unitycatalog/pull/1797 to get in first.

moomindani:
Rebased onto `4ee2b248`. Still waiting on #1797, as you asked.

moomindani:
Rebased onto `81663d1f`, now that #1797 is in.

server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
```diff
@@ -129,6 +129,16 @@ protected BaseException toBaseException(Throwable cause) {
     if (cause instanceof BadRequestException) {
       return wrapException(ErrorCode.INVALID_ARGUMENT, cause);
     }
+    if (cause instanceof NotFoundException) {
```
yili-db, line 139: This error from Iceberg is thrown for any file not found. Today only table metadata error can be triggered. But this is not future proof.
The error message can be weakened as `"Could not read this table"` instead.

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
```diff
@@ -1210,6 +1211,34 @@ private static List<Long> loadedSnapshotIds(AggregatedHttpResponse resp) throws
...
+
+    // A table whose metadata cannot be read is a server-side failure, not a missing table, so the
+    // type has to be one that means that: a 500 typed "NotFoundException" describes the failure as
+    // something the client could act on.
+    assertThat(resp.status().code()).isEqualTo(500);
+    ErrorResponse error = ErrorResponseParser.fromJson(resp.contentUtf8());
+    assertThat(error.code()).isEqualTo(500);
+    assertThat(error.type()).isEqualTo(ServiceFailureException.class.getSimpleName());
+    // Where the server keeps the file is not the client's business.
+    assertThat(error.message()).doesNotContain(icebergTableLocation.toString());
```
yili-db, line 1307: Please also assert on the mesage text.

server/src/main/java/io/unitycatalog/server/exception/IcebergRestExceptionHandler.java
moomindani: Weakened to `"Could not read this table"`, and the comment now says why: Iceberg raises this for any file, today only a table's metadata file reaches here, and naming that file would make the message wrong as soon as another read raises it.

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
moomindani: Added -- the test now asserts the message is exactly `"Could not read this table"`, alongside the existing check that it does not quote the server's storage path.

yili-db (approved):

Issue #1846, opened 2026-09-06
Iceberg REST catalog is missing dropNamespace, updateNamespaceProperties and renameTable


moomindani:
**Describe the bug**

Three operations the Iceberg REST spec defines are not served by this catalog, even though Unity
Catalog already implements each of them internally. An engine pointed at the Iceberg REST endpoint can
read and write tables but cannot run the matching DDL:

| Spec operation | Request | Answered today | Unity Catalog already has |
| --- | --- | --- | --- |
| `dropNamespace` | `DELETE /v1/{prefix}/namespaces/{namespace}` | 405, no route | `SchemaRepository.deleteSchema` |
| `updateProperties` | `POST /v1/{prefix}/namespaces/{namespace}/properties` | 405, no route | schema properties, updated via `SchemaService` |
| `renameTable` | `POST /v1/{prefix}/tables/rename` | 404, no route | `TableRepository.renameTable` |

So `DROP NAMESPACE`, `ALTER NAMESPACE ... SET PROPERTIES` and `ALTER TABLE ... RENAME TO` all fail from
Spark, Trino or any other Iceberg client, with a status that says the path or method does not exist
rather than that the catalog cannot do it. None of the three is advertised in the `endpoints` list from
`GET /v1/config` either, so a client that reads that list is at least not misled -- it simply has no
way to perform these operations.

**To Reproduce**

1. Start the server with `server.iceberg-table.enabled=true`, create a catalog `unity` and a schema
   `default`.
2. `DELETE /api/2.1/unity-catalog/iceberg/v1/catalogs/unity/namespaces/default` answers
   `405 Method Not Allowed`.
3. `POST /api/2.1/unity-catalog/iceberg/v1/catalogs/unity/namespaces/default/properties` with
   `{"removals": [], "updates": {"a": "b"}}` answers `405 Method Not Allowed`.
4. `POST /api/2.1/unity-catalog/iceberg/v1/catalogs/unity/tables/rename` with a `source` and
   `destination` answers `404 Not Found`.

**Expected behavior**

- `dropNamespace`: 204, 404 for a namespace that is not there, 409 for one that is not empty.
- `updateProperties`: 200 reporting which keys were `updated`, `removed` and `missing`, 404 for a
  namespace that is not there, 422 for a key named in both `updates` and `removals`.
- `renameTable`: 204, 404 for a table that is not there, 409 for a destination name that is taken.
- All three advertised in the config `endpoints` list.

[...]

PR #1847, opened 2026-09-06
[Server] Implement the Iceberg dropNamespace endpoint

moomindani:
**PR Checklist**

**Description of changes**

Part of #1846.

`DELETE` on a namespace had no route, so an engine's `DROP NAMESPACE` was answered by Armeria before
any service was reached, with a 405 saying the path takes no such method. Unity Catalog has dropped
schemas all along, through `SchemaRepository.deleteSchema`; the operation was only missing from the
Iceberg REST surface.

This serves it and advertises `V1_DELETE_NAMESPACE` with the other write endpoints, so it is gated the
same way the other Iceberg writes are.

- A namespace that still holds objects is refused. The repository reports that as a failed
  precondition, which would go out as 400; the spec answers a namespace that is not empty with 409, so
  it is raised as Iceberg's `NamespaceNotEmptyException`.
- Authorizations of everything deleted are cleared exactly as `SchemaService` clears them, so dropping
  a namespace here leaves nothing behind that dropping it through the Unity Catalog API would remove.

For users: `DROP NAMESPACE` works from Spark, Trino and any other Iceberg client, and a namespace that
still has tables is refused with the status the client understands as "not empty".

`testDropNamespace` pins the 409 while a table is there, the 204 and empty body once it is not, the
404 for the namespace afterwards, and the updated `endpoints` list. It fails on an unfixed tree and
passes with the fix; the full `server/test` suite shows no failure attributable to this change.

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
+++ b/server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
@@ -228,6 +228,30 @@ public class IcebergRestCatalogTest extends BaseServerTest {
       resp = client.get(TEST_BASE_NON_PREFIX + "/namespaces").aggregate().join();
       assertThat(resp.status().code()).isEqualTo(404);
     }
+
+    // DropNamespace
+    {
+      String namespacePath = TEST_BASE_PREFIX + "/namespaces/" + TestUtils.SCHEMA_NAME;
+
+      // A namespace that still holds a table cannot be dropped, and the spec answers that with 409
+      // rather than the failed precondition the repository reports.
+      createTable(TestUtils.TABLE_NAME);
+      AggregatedHttpResponse resp = client.delete(namespacePath).aggregate().join();
+      assertThat(resp.status().code()).isEqualTo(409);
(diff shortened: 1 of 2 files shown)


server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
```diff
@@ -953,6 +955,30 @@ public void testListTablesReturnsTablesBeyondTheFirstPage()
         .containsExactly(TableIdentifier.of(Namespace.of(TestUtils.SCHEMA_NAME), "uniform_table"));
   }
 
+  @Test
+  public void testDropNamespace() throws Exception {
```
yili-db: Please fold this test into `testNamespaces`

moomindani: Done. The cases are now a DropNamespace block in `testNamespaces`, after the get and list ones, reusing the catalog and schema that test already creates. Without the endpoint the block fails with 405 where 409 is expected, as the separate test did.

yili-db (approved):

PR #1848, opened 2026-09-06
[Server] Implement the Iceberg updateNamespaceProperties endpoint

moomindani:
**PR Checklist**

**Description of changes**

Part of #1846.

`POST` on a namespace's properties had no route, so an engine's `ALTER NAMESPACE ... SET PROPERTIES` was
answered with a 405 before any service was reached. Unity Catalog has kept schema properties all along;
the endpoint that patches them was missing from the Iceberg REST surface.

This serves it and advertises `V1_UPDATE_NAMESPACE` with the other write endpoints. The response
reports which keys were `updated`, which were `removed`, and which were asked to be removed but were
not there, as the spec describes.

Two things worth a reviewer's attention:

- The patch is applied by a new `SchemaRepository` method rather than through `updateSchema`.
  `updateSchema` replaces the whole property map and reads an empty one as "nothing to do", so
  removing a schema's last property could not be expressed through it; and the patch has to read and
  write in one transaction, so two callers changing different keys do not lose each other's change.
- A key named in both `updates` and `removals` is refused by Iceberg's own request validation, which
  raises `UnprocessableEntityException`. The spec answers that with 422, a status the shared
  `ErrorCode` enum had no entry for, so this adds one; the Delta API has no 422, so that dialect keeps
  reporting such a request as a bad one, which the enum already expresses with its per-dialect status.
  Say so if you would rather this answered 400 and left `ErrorCode` alone.

For users: namespace properties can be set and removed from any Iceberg client, and the response says
exactly what the catalog did.

`testUpdateNamespaceProperties` pins the three reported sets, that keys not mentioned are left alone,
the 422, and the 404 for a namespace that is not there. It fails on an unfixed tree and passes with the
fix; the full `server/test` suite shows no failure attributable to this change.

### DIFF
--- a/server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
+++ b/server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
@@ -65,6 +65,7 @@ import org.apache.iceberg.exceptions.NamespaceNotEmptyException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
 import org.apache.iceberg.exceptions.NotFoundException;
 import org.apache.iceberg.exceptions.RESTException;
+import org.apache.iceberg.exceptions.UnprocessableEntityException;
 import org.apache.iceberg.expressions.Expressions;
 import org.apache.iceberg.metrics.CommitMetrics;
 import org.apache.iceberg.metrics.CommitMetricsResult;
@@ -84,6 +85,7 @@ import org.apache.iceberg.rest.responses.GetNamespaceResponse;
 import org.apache.iceberg.rest.responses.ListNamespacesResponse;
 import org.apache.iceberg.rest.responses.ListTablesResponse;
 import org.apache.iceberg.rest.responses.LoadTableResponse;
+import org.apache.iceberg.rest.responses.UpdateNamespacePropertiesResponse;
 import org.apache.iceberg.types.Types;
 import org.hibernate.Session;
 import org.hibernate.Transaction;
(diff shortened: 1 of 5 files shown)


server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
```diff
@@ -953,6 +956,57 @@ public void testListTablesReturnsTablesBeyondTheFirstPage()
...
+            "{\"removals\": [\"drop\", \"absent\"], \"updates\": {\"added\": \"value\"}}");
+    assertThat(resp.status().code()).isEqualTo(200);
+    UpdateNamespacePropertiesResponse updated =
+        IcebergObjectMapper.mapper()
+            .readValue(resp.contentUtf8(), UpdateNamespacePropertiesResponse.class);
+    assertThat(updated.updated()).containsExactly("added");
+    assertThat(updated.removed()).containsExactly("drop");
+    assertThat(updated.missing()).containsExactly("absent");
+
+    // The namespace itself reflects the patch: keys not mentioned are left alone.
```
yili-db, line 1241: After this one, please also try to remove **all** properties and make sure no property is left for this schema.
The reason being that some future change might "optimize" setting empty property to not changing property at all.

moomindani: Good catch, and it is not hypothetical: `updateSchema` is already that shortcut -- it only touches properties when the map it is given is non-empty -- which is exactly why this endpoint patches through `applyPropertyChanges` instead of it.

The test now removes every remaining key and asserts the namespace reports no properties at all. Reading the namespace back moved into a small helper, since two assertions want it.

yili-db (approved):
Please rebase

moomindani:
Rebased onto `4ee2b248`.

yili-db:
I think https://github.com/unitycatalog/unitycatalog/pull/1844 caused conflict again.

moomindani:
Thanks for resolving this yourself -- I had the same rebase ready locally (same `UNPROCESSABLE_ENTITY` arm, same test resolution) and dropped it in favour of your merge. Happy to take the rebase on the remaining ones so it does not land on you.

PR #1849, opened 2026-09-06
[Server] Implement the Iceberg renameTable endpoint

moomindani:
**PR Checklist**

**Description of changes**

Part of #1846.

The rename endpoint had no route, so an engine's `ALTER TABLE ... RENAME TO` was answered with a 404
for a path this server does not serve. Unity Catalog has renamed tables all along, through
`TableRepository.renameTable`, which already refuses a name that is taken; the operation was only
missing from the Iceberg REST surface.

This serves it and advertises `V1_RENAME_TABLE` with the other write endpoints.

- The table is resolved the way `loadTable` resolves it, so a table this API does not expose cannot be
  renamed through it.
- A destination name that is taken is the spec's 409, which is what the repository already raises.
- A destination in a different namespace is refused as unimplemented (501). Unity Catalog has no way
  to move a table between schemas, and that status tells a client the catalog does not do it, rather
  than describing a well-formed request as malformed. The spec lists 406 for an unsupported rename;
  say so if you would rather have that.

For users: `ALTER TABLE ... RENAME TO` works from any Iceberg client, within a namespace.

`testRenameTable` pins the 204 and that the table is listed and loadable under its new name, the 404
for a table that is not there, the 409 for a name that is taken, and the 501 for a cross-namespace
rename. It fails on an unfixed tree and passes with the fix; the full `server/test` suite shows no
failure attributable to this change.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
@@ -588,6 +588,8 @@ public class IcebergRestCatalogService extends AuthorizedService implements Regi
   @AuthorizeResourceKey(METASTORE)
   public HttpResponse renameTable(@Param("catalog") String catalog, RenameTableRequest request) {
     serverProperties.checkIcebergTableEnabled();
+    // A request missing either identifier is a bad request, not the NPE reading it would raise.
+    request.validate();
     Namespace source = request.source().namespace();
     Namespace destination = request.destination().namespace();
     if (!source.equals(destination)) {
@@ -601,12 +603,17 @@ public class IcebergRestCatalogService extends AuthorizedService implements Regi
               + destination);
     }
     String namespace = source.toString();
-    // Only tables this API serves can be renamed through it; anything else is not a table it has.
+    String fullName = catalog + "." + namespace + "." + request.source().name();
+    // As with dropTable, only a table created through this API may be changed through it: a UniForm
+    // table is a Delta table that these endpoints read, and renaming it here would rename the Delta
+    // table under its own API.
     TableRepository.IcebergTableState state =
         tableRepository.getIcebergTableState(catalog, namespace, request.source().name());
-    if (state.metadataLocation() == null) {
-      throw new NoSuchTableException(
-          "Table does not exist: %s", namespace + "." + request.source().name());
+    if (state.dataSourceFormat() != DataSourceFormat.ICEBERG) {
+      throw new BadRequestException(
+          "Table %s was not created through the Iceberg REST catalog; rename it through the Unity"
+              + " Catalog API instead.",
+          fullName);
     }
     tableRepository.renameTable(
         catalog, namespace, request.source().name(), request.destination().name());
--- a/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
@@ -68,6 +68,7 @@ import org.apache.iceberg.exceptions.NoSuchViewException;
 import org.apache.iceberg.rest.Endpoint;
 import org.apache.iceberg.rest.requests.CreateNamespaceRequest;
 import org.apache.iceberg.rest.requests.CreateTableRequest;
+import org.apache.iceberg.rest.requests.RenameTableRequest;
 import org.apache.iceberg.rest.requests.ReportMetricsRequest;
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -548,6 +550,36 @@ private boolean ucTableExists(String catalog, String namespace, String table) {
...
+          "Renaming a table into another namespace is not supported: "
+              + source
+              + " to "
+              + destination);
+    }
+    String namespace = source.toString();
+    // Only tables this API serves can be renamed through it; anything else is not a table it has.
+    TableRepository.IcebergTableState state =
+        tableRepository.getIcebergTableState(catalog, namespace, request.source().name());
+    if (state.metadataLocation() == null) {
```
yili-db: `dropTable` has this:
```
    TableRepository.IcebergTableState state =
        tableRepository.getIcebergTableState(catalog, namespace, table);
    if (state.dataSourceFormat() != DataSourceFormat.ICEBERG) {
      throw new BadRequestException(
          "Table %s was not created through the Iceberg REST catalog; drop it through the Unity"
              + " Catalog API instead.",
          fullName);
    }
```

UniForm tables are supposedly readonly via IRC endpoints.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -548,6 +550,36 @@ private boolean ucTableExists(String catalog, String namespace, String table) {
     }
   }
 
+  @Post("/v1/catalogs/{catalog}/tables/rename")
+  @AuthorizeExpression("#authorize(#principal, #metastore, OWNER)")
+  @AuthorizeResourceKey(METASTORE)
+  public HttpResponse renameTable(@Param("catalog") String catalog, RenameTableRequest request) {
+    serverProperties.checkIcebergTableEnabled();
+    Namespace source = request.source().namespace();
```
yili-db, line 604: `request` is not validated.
A malformed `request` with null source/destination would throw NPE and 500, instead of 400.

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
```diff
@@ -953,6 +954,60 @@ public void testListTablesReturnsTablesBeyondTheFirstPage()
         .containsExactly(TableIdentifier.of(Namespace.of(TestUtils.SCHEMA_NAME), "uniform_table"));
   }
 
+  @Test
+  public void testRenameTable() throws Exception {
```
yili-db: Better fold this test into `testIcebergTableWriteLifecycle` and between some steps in that test.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
moomindani: Fixed. `renameTable` makes the same check `dropTable` makes and answers 400, with a message naming the rename rather than the drop: a UniForm table is a Delta table these endpoints read, so renaming it here would rename the Delta table under its own API.

That also changed where the test belongs, which lines up with your other comment: the rename now runs against the native table `testIcebergTableWriteLifecycle` creates, and the read-only case joins the commit and drop rejections `testTable` already makes against a UniForm table.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
moomindani: Fixed -- `request.validate()` before anything reads it, so a body missing either identifier is the 400 it should be instead of the 500 the NPE produced. `RegisterTableRequest` in #1851 was already validated this way; the rest of the write endpoints take no body or validate it, so this was the only one.

server/src/test/java/io/unitycatalog/server/service/IcebergRestCatalogTest.java
moomindani: Done. The rename runs in `testIcebergTableWriteLifecycle` between the commit and the drop -- against the native table that test creates, which is also what the read-only check above requires -- and renames back at the end so the drop still finds the table. The 404, 409, cross-namespace and malformed-request cases go with it. Without the endpoint they fail with 404 where 204 is expected.

yili-db (approved):
Please rebase

moomindani:
Rebased onto `4ee2b248`.

PR #1851, opened 2026-09-06
[Server] Implement the Iceberg registerTable endpoint

### LINKED ISSUE #1850: Iceberg REST catalog does not serve registerTable
**Describe the bug**

`POST /v1/{prefix}/namespaces/{namespace}/register` is not served, so a table whose Iceberg metadata
already exists cannot be brought into this catalog through the REST API. The request is answered before
any service is reached, with a 404 for a path this server does not have, and the endpoint is not in the
`endpoints` list from `GET /v1/config` either.

That is the path an engine takes to adopt a table written elsewhere -- Spark's
`CALL system.register_table`, or any client calling `Catalog.registerTable` -- and the path a migration
of existing Iceberg tables onto Unity Catalog goes through. Without it the only way in is to create the
table again and rewrite its data, or to insert rows into Unity Catalog by some other route and hope the
metadata pointer matches.

**To Reproduce**

1. Start the server with `server.iceberg-table.enabled=true` and create a catalog `unity` with a schema
   `default`.
2. Write, or take, an Iceberg table's metadata file somewhere the server can read.
3. `POST /api/2.1/unity-catalog/iceberg/v1/catalogs/unity/namespaces/default/register` with
   `{"name": "adopted", "metadata-location": "<that file>"}` answers `404 Not Found`.

**Expected behavior**

200 with the `LoadTableResponse` for the table the metadata file describes, and the table registered in
Unity Catalog from that point on: listed by `listTables`, loadable by `loadTable`. A name that is
already taken is 409 unless the request asks to overwrite; a namespace that does not exist is 404. The
metadata file itself is not modified. The endpoint is advertised in the config `endpoints` list.

**Additional context**

Verified on `main` at 58d5c7b against a running server. The rest of the write surface -- create, update,
drop -- landed in #1696; this is the one write the spec defines that is still missing, apart from
transactions and the view endpoints.

Related: #1846 covers three other operations the spec defines that this catalog does not serve
(`dropNamespace`, `updateProperties`, `renameTable`). This one is filed separately because those three
are already implemented inside Unity Catalog and only need wiring, while registering a table needs a
read of metadata for a table the catalog does not have yet.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1850.

The register endpoint was not served, so a table whose Iceberg metadata already exists could not be
brought into this catalog through the REST API: the request was answered with a 404 for a path this
server does not have. That is the path an engine takes to adopt a table written elsewhere, and the one a
migration onto Unity Catalog goes through.

This serves it and advertises `V1_REGISTER_TABLE` with the other write endpoints, so it is gated the
same way the other Iceberg writes are.

- The metadata file is read and never written. The table's columns, properties and location come from
  the document the request names, and because nothing is written there is nothing to clean up if the
  row cannot be created -- unlike `createTable`, which has to delete the metadata it wrote.
- The file has to sit inside the table location it declares, so a metadata file cannot register a table
  rooted somewhere it does not live. `MetadataService` gained a read for a table Unity Catalog does not
  have yet, because the location it validates against cannot come from a persisted row that is not
  there.
- A name that is already taken is the spec's 409. When the request asks to overwrite, the given file is
  committed over whatever the table points at now, through the same compare-and-swap a commit uses: a
  commit that lands in between is reported as a conflict rather than silently overwritten. Overwriting a
  table that is not a native Iceberg table is refused by that commit path, as it is for any other
  commit. Iceberg's client does send `overwrite`, which is why it is implemented rather than rejected.

For users: an existing Iceberg table can be adopted into Unity Catalog by naming its metadata file, and
the table is a full member of the catalog afterwards -- listed, loadable, and committable.

`testRegisterTable` pins the response's metadata pointer, that the table is then listed and loadable and
carries its columns in Unity Catalog, the 409 for a taken name, the 200 when overwrite is asked for, the
404 for a namespace that does not exist, and the rejection of a metadata file outside the location it
declares. It fails on an unfixed tree and passes with the fix; the full `server/test` suite shows no
failure attributable to this change (the one failure here is the GCP credential test, which needs
credentials this machine does not have).

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
@@ -20,6 +20,7 @@ import io.unitycatalog.server.exception.BaseException;
 import io.unitycatalog.server.exception.ErrorCode;
 import io.unitycatalog.server.exception.IcebergRestExceptionHandler;
 import io.unitycatalog.server.model.CatalogInfo;
+import io.unitycatalog.server.model.ColumnInfo;
 import io.unitycatalog.server.model.CreateSchema;
 import io.unitycatalog.server.model.CreateStagingTable;
 import io.unitycatalog.server.model.CreateTable;
@@ -68,6 +69,7 @@ import org.apache.iceberg.exceptions.NoSuchViewException;
 import org.apache.iceberg.rest.Endpoint;
 import org.apache.iceberg.rest.requests.CreateNamespaceRequest;
 import org.apache.iceberg.rest.requests.CreateTableRequest;
+import org.apache.iceberg.rest.requests.RegisterTableRequest;
 import org.apache.iceberg.rest.requests.ReportMetricsRequest;
 import org.apache.iceberg.rest.requests.UpdateTableRequest;
 import org.apache.iceberg.rest.responses.ConfigResponse;
@@ -102,7 +104,8 @@ public class IcebergRestCatalogService extends AuthorizedService implements Regi
           Endpoint.V1_CREATE_NAMESPACE,
           Endpoint.V1_CREATE_TABLE,
           Endpoint.V1_UPDATE_TABLE,
-          Endpoint.V1_DELETE_TABLE);
+          Endpoint.V1_DELETE_TABLE,
+          Endpoint.V1_REGISTER_TABLE);
 
   private final TableConfigService tableConfigService;
   private final MetadataService metadataService;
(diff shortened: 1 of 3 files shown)


yili-db:
Thanks for putting this together. Before we get into the code, I want to step back on whether `/register` fits our model at all — I'm not sure it does, at least not the fresh-register case.

Our UC Iceberg table model is: a table has one location, and governed by UC, and everything (data + metadata) lives under it. For managed tables it is assigned by UC, and for external tables it is chosen by client. Either way the location is governed by UC once the table is created.
`RegisterTableRequest` is `{name, metadata-location, overwrite}` — there's no table location in the request. So on a fresh register UC never gets to choose or approve or even know where the table lives and where is the storage boundary; it can only do `location = read(metadataFile).location` and trust it. And that is risk as there isn't any guarantee by spec that the location is indeed the storage boundary.

So I don't think we are ready to implement this endpoint in UC yet.

moomindani:
Fair enough -- feel free to close it. The intent across these PRs was spec coverage, filling in the Iceberg REST endpoints UC does not serve yet, and register turns out not to fit UC's model the way the others do; that is not something to settle quickly, so I would rather park it than force it.

One note for whenever it comes up again: besides the request carrying no location for UC to approve, the overwrite path is unsound on its own terms -- it commits the given metadata file over an existing table's pointer without checking the file is under the location UC governs for that table (`commitIcebergTable` never compares against the table's url; the normal commit path is safe only because the service derives the location from the table itself).

PR #1853, opened 2026-09-06
[Server] Answer conditional Iceberg loadTable requests

### LINKED ISSUE #1852: [Server] Iceberg loadTable ignores If-None-Match, so clients re-read metadata they already hold

`loadTable` on the Iceberg REST endpoints reads and parses the table's metadata file from storage on every single call, even when the caller already holds that exact version. The Iceberg REST spec has a mechanism for exactly this -- `If-None-Match` on `loadTable`, a `304 Not Modified` response, and an `ETag` on `LoadTableResponse` -- and Unity Catalog implements none of it, so no caller can avoid the round trip to object storage.

Every released Iceberg Java client is ready to use it. `RESTSessionCatalog` keeps a per-session table cache (5 minutes by default, `rest-table-cache.expire-after-write-ms`) and populates it only when a load response carries an `ETag`; on the next load within that window it sends the tag back in `If-None-Match` and, on a `304`, serves the table from the cache. Against Unity Catalog the cache is never populated, so a workload that loads the same table repeatedly -- a planner, a metadata browser, several short queries -- pays for one metadata read per load.

The information needed to answer is already in Unity Catalog's own row: `uniform_iceberg_metadata_location` names the table's current metadata version, and it moves on every commit. A conditional load can be answered from the row alone, without touching storage at all.

`loadTable` should return an `ETag` derived from the table's current metadata pointer, and answer a request whose `If-None-Match` names that tag with a bodiless `304` -- short-circuiting before the metadata file is read. Because `snapshots=refs` and `snapshots=all` return different bodies for the same version, the tag has to distinguish them, as the spec's `ETag` parameter notes.

One thing worth deciding along with it: released Iceberg clients keep response headers in a map keyed by the header name as they received it and look this one up as `ETag` (`RESTSessionCatalog` reads `HttpHeaders.ETAG` out of the `Maps.newHashMap()` that `HTTPClient` fills). Armeria writes HTTP/1 header names in their lowercase HTTP/2 form by default, so `etag` on the wire is invisible to those clients and the feature would have no effect on them.

**Additional context**

Iceberg REST spec 1.11.0 (`open-api/rest-catalog-open-api.yaml`): `If-None-Match` and the `304` response on `loadTable`, and the `ETag` parameter shared by `CreateTableResponse`, `LoadTableResponse` and `CommitTableResponse`.

Part of the ongoing work to close the gap between Unity Catalog's Iceberg REST endpoints and the spec.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1852.

`loadTable` on the Iceberg REST endpoints read and parsed the table's metadata file from storage on every call, even for a caller that already held that version, because Unity Catalog implemented none of the conditional-request mechanism the Iceberg REST spec defines for the endpoint (`If-None-Match`, `304 Not Modified`, `ETag`).

It now returns an `ETag` naming the version it served, and answers a request whose `If-None-Match` names that tag with a bodiless `304` -- decided from the row alone, so the metadata file is never read. Every released Iceberg Java client benefits without any configuration: `RESTSessionCatalog` caches a table per session for five minutes once a load response carries an `ETag`, sends the tag back on the next load, and serves the table from its cache on a `304`.

Technical changes:

- `IcebergRestCatalogService.loadTable` returns an `HttpResult<LoadTableResponse>` so it can carry the header, and takes the `If-None-Match` header. The tag is a weak validator over a SHA-256 of the table's current metadata pointer -- hashed so the tag does not hand callers a storage location, weak because it identifies the metadata version rather than one exact body. The snapshots mode is part of it, since `refs` and `all` return different bodies for the same version; the spec's `ETag` parameter calls for exactly that. An invalid `snapshots` value is still rejected before any of this, so a conditional request cannot mask a bad one.
- `ArmeriaServerBuilder` sets `http1HeaderNaming(Http1HeaderNaming.traditional())`. Without it the feature has no effect on the clients it is for: Armeria writes HTTP/1 header names in their lowercase HTTP/2 form, while `HTTPClient` collects response headers into a plain `HashMap` keyed by the name as received and `RESTSessionCatalog` looks this one up as `ETag`. The option only changes the casing of standard header names on HTTP/1 responses (`Content-Type`, `ETag`); HTTP/1.1 field names are case-insensitive, so this is invisible to any client that reads them correctly, and unknown names are left alone.

Only `loadTable` gained a tag. The spec also allows one on `createTable` and `commitTable` responses, but no released client seeds its cache from those, so there is nothing to serve yet.

Testing:

- `IcebergRestCatalogTest.testLoadTableConditionalGet` covers the tag's shape (weak, and not the metadata location), the `304` and its empty body, a multi-tag `If-None-Match`, a stale tag and `*` both getting the full response, `refs` carrying its own tag that does not match the default one, an invalid `snapshots` value still being a 400, a moved metadata pointer retiring the tag, and the header reaching the wire spelled `ETag` (read off a socket, since both Armeria's client and the JDK's lowercase header names while parsing). It fails on an unfixed tree, in both halves: without the service change there is no tag at all, and without the naming option the wire assertion fails.
- `build/sbt "server/testOnly io.unitycatalog.server.service.IcebergRestCatalogTest"` and `build/sbt -J-Xmx2G "server/test"`.
- Validated against a running server rather than only in tests. A real Iceberg 1.11.0 `RESTCatalog` loaded a table, then loaded it again after the metadata file was made unreadable, and succeeded -- the second load was answered `304` from the pointer and served from the client's cache. By hand: the tag is issued and echoed on the `304`, a tag list containing it still matches, a stale tag and `*` get the full response, `snapshots=refs` carries a different tag, a real commit retires the tag, and a conditional request succeeds while the metadata file is unreadable where a plain one fails.

Backward compatibility: additive. A caller that sends no `If-None-Match` sees the same response it always did, plus a header it may ignore. Nothing changes for the Delta, native or SCIM surfaces beyond the casing of standard HTTP/1 response header names.

Stacked on #1844 (the `snapshots` parameter), whose branch this one builds on -- the tag has to distinguish `refs` from `all`, which only that PR makes possible. Review the top commit, or merge #1844 first.

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/ArmeriaServerBuilder.java
+++ b/server/src/main/java/io/unitycatalog/server/ArmeriaServerBuilder.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.json.JsonMapper;
+import com.linecorp.armeria.common.Http1HeaderNaming;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.server.DecoratingHttpServiceFunction;
@@ -81,6 +82,11 @@ public class ArmeriaServerBuilder {
     this.armeriaServerBuilder =
         Server.builder()
             .localPort(port, SessionProtocol.HTTP)
+            // Armeria names HTTP/1 headers in their lowercase HTTP/2 form by default. Released
+            // Iceberg clients read our response headers out of a plain map keyed by the name as
+            // received, so a header they look up by its traditional spelling -- "ETag" for a
+            // conditional loadTable -- is invisible to them unless we write it that way.
+            .http1HeaderNaming(Http1HeaderNaming.traditional())
             .serviceUnder("/docs", new DocService());
     this.armeriaServerBuilder.service("/", (ctx, req) -> HttpResponse.of("Hello, Unity Catalog!"));
     this.basePath = basePath;
--- a/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
+++ b/server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
@@ -4,12 +4,18 @@ import static io.unitycatalog.server.model.SecurableType.METASTORE;
 import static io.unitycatalog.server.service.credential.CredentialContext.READ_ONLY;
 import static io.unitycatalog.server.service.credential.CredentialContext.READ_WRITE;
 
+import com.google.common.hash.Hashing;
+import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.common.ResponseHeaders;
 import com.linecorp.armeria.server.annotation.Delete;
 import com.linecorp.armeria.server.annotation.ExceptionHandlerFunction;
 import com.linecorp.armeria.server.annotation.Get;
 import com.linecorp.armeria.server.annotation.Head;
+import com.linecorp.armeria.server.annotation.Header;
+import com.linecorp.armeria.server.annotation.HttpResult;
 import com.linecorp.armeria.server.annotation.Param;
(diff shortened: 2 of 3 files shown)


moomindani:
Rebased on top of #1844's branch, so this PR is now the ETag commit alone once that one lands. Also squashed the empty CI re-run commit away.

yili-db:
https://github.com/unitycatalog/unitycatalog/pull/1844 is merged. Please rebase.

moomindani:
Rebased onto `81663d1f`, so this is the ETag change alone now that #1844 is in.

yili-db (approved):

PR #1855, opened 2026-09-06
[Server] Wait for the client-facing port before reporting the server started

### LINKED ISSUE #1854: [Server] The server reports itself started before the client-facing port is listening
**Describe the bug**

The server reports itself started before the port clients actually use is listening, and the test harnesses wait on the wrong port, so a suite that talks to the server immediately can fail with a connection refused.

`UnityCatalogServer.main` starts Armeria on `port + 1`, waits for it, and then deploys the Vert.x URL transcoder that serves the port clients connect to:

```java
unityCatalogServer.start();          // Armeria, port + 1 -- awaited
Vertx vertx = Vertx.vertx();
Verticle transcodeVerticle = new URLTranscoderVerticle(options.getPort(), options.getPort() + 1);
vertx.deployVerticle(transcodeVerticle);   // client-facing port -- not awaited
```

`URLTranscoderVerticle.start()` calls `server.listen(transcodePort, ar -> ...)` and returns without waiting for the bind, so the deployment is complete before the socket exists. Two consequences:

1. There is a window in which the internal port answers and the client-facing port refuses connections.
2. If the client-facing port cannot be bound at all, the failure is logged at INFO and dropped -- `deployVerticle`'s result is never inspected -- so the process keeps running, exits 0 on shutdown, and serves nothing on the port every client uses. `URLTranscoderVerticleTest` documents the first half of this in a helper comment: "The verticle binds its port asynchronously, so wait for it before sending any request."

The Python client suite hits window 1 in CI. Its fixture waits for the internal port and then talks to the client-facing one:

```python
response = requests.head("http://localhost:8081", timeout=60)   # internal
...
config = Configuration(host="http://localhost:8080/api/2.1/unity-catalog")   # what the tests use
```

It failed exactly that way on a recent run: `1 failed, 34 passed`, the one failure being the first test to issue a request.

```
FAILED clients/python/tests/test_catalog.py::test_catalog_list - aiohttp.client_exceptions.ClientConnectorError:
Cannot connect to host localhost:8080 ssl:default [Multiple exceptions: [Errno 111] Connect call failed ('::1', 8080, 0, 0),
[Errno 111] Connect call failed ('127.0.0.1', 8080)]
```

`tests/test_tutorial.py` and `tests/test_tarball_generation.py` check the same internal port and print "Server is running."; they are shielded only by fixed `time.sleep(60)` / `time.sleep(15)` calls that precede the probe.

**To Reproduce**

Start the server, and while Armeria is up but the transcoder has not bound yet, request the client-facing port. For the dropped-failure half, occupy port 8080 with another process and start the server: it logs the bind failure at INFO, keeps running with only 8081, and reports success.

**Expected behavior**

Deploying the transcoder should complete only once its port is listening, a bind failure there should stop the server rather than leave it serving only the internal port, and a readiness probe should test the port the client will use.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1854.

The server reported itself started before the port clients use was listening. `main` awaited Armeria on `port + 1` and then deployed the Vert.x URL transcoder for the client-facing port without awaiting it, and `URLTranscoderVerticle.start()` returned as soon as `server.listen` had been called. A bind failure on that port was logged at INFO and dropped, since `deployVerticle`'s result was never inspected, so the process kept running and exited 0 while serving nothing on the port every client connects to.

The Python client suite hit the first half of this in CI (`1 failed, 34 passed`, the failure being the first test to issue a request), because its fixture waits for the internal port and then talks to the client-facing one.

Technical changes:

- `URLTranscoderVerticle` takes the `Promise<Void>` form of `start` and completes it from the `listen` callback, so the deployment completes only once the port is bound and fails with the bind error otherwise.
- `UnityCatalogServer.main` waits for that deployment and, if it fails, logs which port could not be bound, closes the server and rethrows. A server that cannot serve the port its clients use now exits non-zero instead of running half-up.
- The three readiness probes that waited on the internal port -- `clients/python/tests/conftest.py`, `tests/test_tutorial.py`, `tests/test_tarball_generation.py` -- now probe the port those suites actually use. The fixed `time.sleep` calls that were masking the race in the latter two are left as they are; this change is about waiting for the right thing, not about retuning those waits.

Testing:

- `URLTranscoderVerticleTest.testDeploymentFailsWhenTheTranscodePortIsTaken` occupies a port and asserts the deployment fails with a `BindException`. It fails on an unfixed tree, where the deployment reports success. The test class's `awaitPort` polling helper is gone: deployment now means the port is listening, which is what its comment used to work around.
- `build/sbt "server/testOnly io.unitycatalog.server.URLTranscoderVerticleTest"` and `build/sbt -J-Xmx2G "server/test"`.
- Validated against a running server. `HEAD /` on the client-facing port answers 200, which is what the probes assert. With that port occupied by another process, the server now logs `Failed to start the URL transcoder on port 8080`, stops cleanly ("Unity Catalog server stopped.", pool cleanup) and exits 1, where before it stayed up on the internal port only.
- Ran the Python client suite against a server started by `bin/start-uc-server` with the updated fixture: it came up and ran in 9.7s (`3 failed, 32 passed`); the three failures are `bin/uc` invocations, which need the CLI jar I had not built locally.

Backward compatibility: a deployment that could not bind the client-facing port used to leave the process running; it now fails the startup. That is the point of the change -- such a server cannot serve any client -- but it is a behaviour change for anyone who was starting the server with that port already in use.

### DIFF
--- a/clients/python/tests/conftest.py
+++ b/clients/python/tests/conftest.py
@@ -37,7 +37,7 @@ def uc_server():
             print(">> Waiting for server to accept connections ...")
             for _ in range(90):
                 try:
-                    response = requests.head("http://localhost:8081", timeout=60)
+                    response = requests.head("http://localhost:8080", timeout=60)
                     if response.status_code == 200:
                         print("Server is running.")
                         break
--- a/server/src/main/java/io/unitycatalog/server/URLTranscoderVerticle.java
+++ b/server/src/main/java/io/unitycatalog/server/URLTranscoderVerticle.java
@@ -1,6 +1,7 @@
 package io.unitycatalog.server;
 
 import io.vertx.core.AbstractVerticle;
+import io.vertx.core.Promise;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.HttpMethod;
 import io.vertx.core.http.HttpServer;
@@ -25,7 +26,7 @@ class URLTranscoderVerticle extends AbstractVerticle {
   }
 
   @Override
-  public void start() {
+  public void start(Promise<Void> startPromise) {
     HttpServer server = vertx.createHttpServer();
     WebClient client = WebClient.create(vertx);
 
@@ -61,13 +62,16 @@ class URLTranscoderVerticle extends AbstractVerticle {
                   });
         });
 
+    // The port is bound asynchronously, so the deployment is only complete once it is listening:
+    // callers are told the server is ready when this verticle is deployed.
     server.listen(
         transcodePort,
         ar -> {
           if (ar.succeeded()) {
             LOGGER.info("URL transcoder started on port {}", transcodePort);
+            startPromise.complete();
           } else {
-            LOGGER.info("Failed to start URL transcoder: {}", String.valueOf(ar.cause()));
+            startPromise.fail(ar.cause());
           }
         });
   }
--- a/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
+++ b/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
@@ -48,6 +48,7 @@ import io.unitycatalog.server.utils.VersionUtils;
 import io.vertx.core.Verticle;
 import io.vertx.core.Vertx;
 import java.nio.file.Path;
+import java.util.concurrent.CompletionException;
 import org.apache.logging.log4j.core.config.Configurator;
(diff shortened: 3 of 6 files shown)


yili-db (approved):

tdas (approved):
thanks for fixing this corner case.

PR #1856, opened 2026-09-07
[spark] Add Apache Spark 4.3.0 RC1 support

timothyw553:
**PR Checklist**

- [ ] If you have added or modified a feature, documentation in docs is updated.

**Description of changes**

Add a Spark 4.3.0 RC1 build profile for the Unity Catalog Spark connector.

The profile uses the existing source-build support with the Apache Spark `v4.3.0-rc1` tag. Spark 4.3 RC1 uses the same relation and view interfaces as Spark 4.2. The Spark 4.2 shim directories are therefore renamed to `spark-4.2-4.3` and shared by both versions.

The unit-test workflow adds one Spark 4.3 matrix row. It keeps the existing unit-test action, build wrapper, and job-name logic unchanged.

Spark 4.3 also compares two analyzer implementations during tests. The path-credential rule needs more than one analyzer pass. The path-credential test session disables this comparison on Spark 4.3 so it can test the rule itself.

**Test plan**

- Spark connector production sources compiled against Apache Spark 4.3.0 RC1 commit `3db813416c8716996b0d4d2fbb94fed5dc121fd2`.
- `PathCredentialReadWriteTest` passed with Spark 4.3 RC1: 25 tests passed.
- Cross-Spark publication tests passed.
- Source-build metadata and deterministic cache-key tests passed.
- Checkstyle passed.
- Workflow, JSON, and Python validation passed.


build.sbt
```diff
@@ -675,10 +675,14 @@ lazy val spark = (project in file("connectors/spark"))
...
-      "com.fasterxml.jackson.core" % "jackson-databind" % "2.15.0",
-      "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.15.0",
-      "com.fasterxml.jackson.core" % "jackson-annotations" % "2.15.0",
-      "com.fasterxml.jackson.core" % "jackson-core" % "2.15.0",
+      "com.fasterxml.jackson.core" % "jackson-databind" %
+        (if (CrossSparkVersions.getSparkVersionSpec().isAtLeast(4, 3)) "2.21.2" else "2.15.0"),
+      "com.fasterxml.jackson.module" %% "jackson-module-scala" %
+        (if (CrossSparkVersions.getSparkVersionSpec().isAtLeast(4, 3)) "2.21.2" else "2.15.0"),
+      "com.fasterxml.jackson.core" % "jackson-annotations" %
+        (if (CrossSparkVersions.getSparkVersionSpec().isAtLeast(4, 3)) "2.21" else "2.15.0"),
```
yili-db, line 683: why this is `2.21` and others are `2.21.2`?

.github/workflows/unit-tests.yml
```diff
@@ -75,12 +75,20 @@ jobs:
             validation-mode: "spark-tests"
             non-blocking: true
             spark-source-build: false
+          - spark-version: "4.3.0-SNAPSHOT"
+            delta-version: "master-SNAPSHOT"
+            delta-ref: "pull/7624/head"
```
yili-db, line 80: Assuming this is only for test. Remember to revert this line.
This test will fail as `non-blocking` until Delta's 7624 is merged. And that's fine.

build.sbt
timothyw553: jackson-annotations has a separate release version. Spark 4.3 uses 2.21 for annotations and 2.21.2 for the other Jackson modules. There is no jackson-annotations 2.21.2 artifact.

connectors/spark/src/test/java/io/unitycatalog/spark/PathCredentialReadWriteTest.java
```diff
@@ -122,6 +122,9 @@ private SparkSession createPathCredSessionWithLayout(
             .master("local[*]")
             .config("spark.sql.shuffle.partitions", "4")
             .config("spark.sql.extensions", DELTA_AND_UC_EXTENSIONS);
+    if (isSparkAtLeast(4, 3)) {
+      builder.config("spark.sql.analyzer.singlePassResolver.dualRunWithLegacy", "false");
```
yili-db, line 126: FYI @s-maling-celonis Spark 4.3 seems to behave differently with the path based tables access.

s-maling-celonis: @yili-db Thanks for the heads-up.

This is the 4.3 test dual-run (fixed-point vs single-pass). Production still uses fixed-point, so path vending is unchanged. dualRunWithLegacy=false keeps these tests on that path. I’ll look at single-pass support for the rule over the next few days.

s-maling-celonis: @yili-db @timothyw553 I dug into this one — the difference is Spark's new single-pass analyzer, and the workaround on this line is the right call for now.

`spark.sql.analyzer.singlePassResolver.dualRunWithLegacy` defaults to `Utils.isTesting`, so from 4.3 it turns itself on in any JVM started with `spark.testing`, and only there. Dual-run analyzes every query with both the fixed-point and the single-pass analyzer and compares the two plans.

That breaks bare-path access specifically. Path credential vending relies on `injectHintResolutionRule`, the only extension point that runs before relation metadata is resolved — the fixed-point analyzer runs it in its "Hints" batch, ahead of `ResolveSQLOnFile`, so `ResolvePathCredentials` sees the `UnresolvedRelation` and can attach credentials. `HybridAnalyzer` forwards the other `singlePass*` extension points but has no hint resolution rule list, so under single-pass those rules are dropped and the path is listed before anything can rewrite it. The two halves then disagree and the query fails on the plan comparison instead of quietly skipping vending.

So this is a gap in Spark rather than in Unity Catalog, and nothing here is being masked: production does not enable the single-pass resolver, and setting `dualRunWithLegacy=false` just pins the analyzer that production already uses. I would only add a comment next to it pointing at the ticket, so it does not look arbitrary later:

- SPARK-59574: https://issues.apache.org/jira/browse/SPARK-59574
- apache/spark#58856: https://github.com/apache/spark/pull/58856

Once that lands, Unity Catalog needs no production change — `ResolvePathCredentials` already matches the right operators. The follow-up here is test coverage: running the bare-path vending tests across the fixed-point, single-pass and dual-run analyzers so both paths are proven rather than one being switched off.

PR #1859, opened 2026-09-09
[Server] Make deny-triggered Casbin refresh causally fresh

s-maling-celonis:
**PR Checklist**

**Description of changes**

Follow-up to [#1766](https://github.com/unitycatalog/unitycatalog/pull/1766).

Cross-replica create-then-read still 403s after that change. Two races in the deny-triggered refresh path:

1. A refresh that is already in flight claims the debounce timestamp before it checks the store. Concurrent denials keep the original 403 while the winner reloads.
2. Elapsed time since the last probe does not prove that probe could have seen a later grant. If a refresh finished just before OWNER was written, `getStagingTableCredentials` (and any other follow-up request) re-evaluates the stale enforcer inside the one-second window.

This PR replaces skip-if-recent debounce with causal freshness, and joins concurrent denials onto one probe:

- each authorized operation carries a monotonic barrier taken no later than its start (list filters stamp at `filter()`, after the list query);
- a successful policy probe publishes the timestamp from immediately before its database query;
- a completed probe covers every operation that started before that probe;
- concurrent uncovered requests coalesce behind one probe and at most one reload;
- an operation that started while a probe was already in flight triggers a subsequent probe;
- failed probes do not cover later denials (fail-closed).

The deny path probes `max(id)` only (a stale deny is a missing insert). The poller still uses `count(*)` and `max(id)` so delete-only revokes propagate.

**Config**

`server.authorization.policy-refresh-debounce-interval` is removed. Existing files that still set it start normally; the key is ignored and does not restore skip-if-recent behavior.

`server.authorization.policy-refresh-min-probe-interval` defaults to `PT1S`. Quiet traffic is unlikely to wait. Under many 403s per second, lower toward `PT0S` so each new request can probe immediately; that increases `casbin_rule` load.

Refresh remains off unless `server.authorization.policy-refresh=true`.

**Test plan**

- [x] `JCasbinAuthorizerMultiInstanceTest` — grant after a previous probe is visible to a new request; denials that share one request barrier do not probe again; concurrent denials wait and re-evaluate; min probe interval waits rather than skips; revoke of a non-max row still needs a full poller `count(*)`
- [x] `UnityAccessEvaluatorTest` — deny-path refresh + retry
- [x] `ServerPropertiesTest` — `PT0S` / `PT0.1S` / `PT1S` valid, `PT-1S` rejected
- [x] `JCasbinAuthorizerTest` — existing grant/revoke/hierarchy/list behavior
- [ ] `server/test` in CI

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
@@ -18,8 +18,16 @@ import org.slf4j.LoggerFactory;
  * auto-save enabled, but each server process only loads that table into its enforcer at startup
  * unless this refresher reloads it.
  *
- * <p>The poller uses {@code count(*)} and {@code max(id)}; deny-path checks use {@code max(id)}
- * only. Reload is delegated to {@link JCasbinAuthorizer} so a fresh enforcer can be swapped without
+ * <p>Two probes, for different use cases:
+ *
+ * <ul>
+ *   <li>The background poller uses {@link #VERSION_QUERY} ({@code count(*)} and {@code max(id)}) so
+ *       a revoke that deletes a non-max row still shows up.
+ *   <li>The deny path uses {@link #MAX_ID_QUERY} ({@code max(id)} only). A stale deny after a grant
+ *       is a missing insert; revokes are left to the poller.
+ * </ul>
+ *
+ * <p>Reload is delegated to {@link JCasbinAuthorizer} so a fresh enforcer can be swapped without
  * blocking {@code enforce()}.
  */
 public class CasbinPolicyRefresher implements AutoCloseable {
@@ -32,8 +40,10 @@ public class CasbinPolicyRefresher implements AutoCloseable {
     RELOADED
   }
 
+  // Poller: count(*) catches revokes that delete a non-max row; max(id) catches new grants.
   private static final String VERSION_QUERY =
       "select count(*), coalesce(max(id), 0) from casbin_rule";
+  // Deny path: a stale deny after a grant is a missing insert, which always raises max(id).
   private static final String MAX_ID_QUERY = "select coalesce(max(id), 0) from casbin_rule";
 
   private final Runnable reloader;
@@ -45,7 +55,7 @@ public class CasbinPolicyRefresher implements AutoCloseable {
   private long lastMaxId = -1;
 
   // System.nanoTime() of the last successful probe start; Long.MIN_VALUE if none.
-  private volatile long lastCompletedProbeAt = Long.MIN_VALUE;
+  private volatile long lastCompletedProbeAtNanos = Long.MIN_VALUE;
 
   private ScheduledExecutorService executor;
 
@@ -84,26 +94,27 @@ public class CasbinPolicyRefresher implements AutoCloseable {
   }
 
   /**
-   * Probes {@code casbin_rule} and reloads when the version changed.
+   * Poller probe: {@link #VERSION_QUERY} ({@code count(*)} and {@code max(id)}).
    *
    * @return true if a reload happened
    */
   public boolean checkAndReload() {
     synchronized (this) {
-      return probeAndReload(true) == ProbeResult.RELOADED;
+      return probeCountAndMaxId() == ProbeResult.RELOADED;
     }
   }
 
   /**
-   * Reloads if policy may have changed after {@code observedBefore} ({@link System#nanoTime()}).
+   * Reloads if policy may have changed after {@code operationStartNanos} ({@link System#nanoTime()}
+   * from the start of the denied operation).
    */
-  public boolean checkAndReloadAfter(long observedBefore) {
-    if (isCoveredByCompletedProbe(observedBefore)) {
+  public boolean checkAndReloadAfter(long operationStartNanos) {
+    if (isCoveredByCompletedProbe(operationStartNanos)) {
       return true;
     }
     synchronized (this) {
       while (true) {
-        if (isCoveredByCompletedProbe(observedBefore)) {
+        if (isCoveredByCompletedProbe(operationStartNanos)) {
           return true;
         }
         Duration remaining = remainingMinProbeInterval();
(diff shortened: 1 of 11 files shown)


server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -194,8 +201,8 @@ public enum Property {
     POLICY_REFRESH_ENABLED("server.authorization.policy-refresh", "false", BOOLEAN_VALIDATOR),
     POLICY_REFRESH_INTERVAL(
         "server.authorization.policy-refresh-interval", "PT1M", DURATION_VALIDATOR),
-    POLICY_REFRESH_DEBOUNCE_INTERVAL(
-        "server.authorization.policy-refresh-debounce-interval", "PT1S", DURATION_VALIDATOR),
+    POLICY_REFRESH_MIN_PROBE_INTERVAL(
+        "server.authorization.policy-refresh-min-probe-interval", "PT0S", DURATION_VALIDATOR),
```
yili-db: Can this be kept as `PT1S` to cap the polling frequency?

403 storm isn't a rare event. In the UC Spark connector we have loadTable vending a table credential with READ_WRITE-then-READ fallback. Every intent-less load by a SELECT-only principal issues an expected-403 READ_WRITE credential request.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessEvaluator.java
```diff
@@ -89,6 +86,27 @@ public boolean evaluate(
...
+   * @param resourceIds Map of resource types to their UUIDs
+   * @param nonResourceValues Map of parameter names to their values (from @AuthorizeKey)
+   * @param observedBefore {@link System#nanoTime()} from the start of this operation
+   */
+  public boolean evaluate(
+      UUID principal,
+      String expression,
+      Map<SecurableType, UUID> resourceIds,
+      Map<String, Object> nonResourceValues,
+      long observedBefore) {
```
yili-db: It's hard to understand the meaning of `observedBefore` without looking at the actual code that uses it. It could be renamed as `startBarrierNanos` or `operationStartNanos`.

Also `observedBefore` is `System#nanoTime` but a untyped `long`.
Please rename to `*Nanos` to carry this type info.
The same to `lastCompletedProbeAt`.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
```diff
@@ -19,36 +18,42 @@
...
 
   private static final Logger LOGGER = LoggerFactory.getLogger(CasbinPolicyRefresher.class);
 
+  private enum ProbeResult {
+    FAILED,
+    UNCHANGED,
+    RELOADED
+  }
+
   private static final String VERSION_QUERY =
```
yili-db, line 44: Please document the difference of the use cases of the two queries: One is the background poller that cares about revokes too, while the 2nd one only cares about new grants up on an initial denial.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
```diff
@@ -78,45 +86,106 @@ private void pollQuietly() {
...
+    try {
+      wait(Math.max(1, remaining.toMillis()));
+      return true;
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      return false;
+    }
+  }
+
+  private ProbeResult probeAndReload(boolean fullVersion) {
```
yili-db: It seems that the full version and id only version doesn't share anything inside this function. They are simply in difference sides of the if-else.
Also nothing calls `probeAndReload` with a variable `fullVersion` but only const false or true.
So it makes sense to split this function into two and avoid this `if`.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
```diff
@@ -78,45 +86,106 @@ private void pollQuietly() {
...
+      return ProbeResult.FAILED;
+    }
+    if (maxId == lastMaxId) {
+      recordCompletedProbe(probeStarted);
+      return ProbeResult.UNCHANGED;
     }
+    return reload(probeStarted, lastCount, maxId);
+  }
+
+  private ProbeResult reload(long probeStarted, long count, long maxId) {
```
yili-db: rename `probeStarted` to `probeStartedNanos`

yili-db:
FYI the failing CI has been fixed in main branch.

server/src/main/java/io/unitycatalog/server/auth/decorator/UnityAccessEvaluator.java
s-maling-celonis: Renamed to `operationStartNanos` / `lastCompletedProbeAtNanos` (and `probeStartedNanos` on the probe path). Left as `long` from `System.nanoTime()` rather than `Instant`, since the barrier is monotonic, not wall-clock.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
s-maling-celonis: Documented on the class and next to the query constants: the poller uses `count(*)` and `max(id)` so a revoke of a non-max row is visible; the deny path uses `max(id)` only because a stale deny after a grant is a missing insert.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
s-maling-celonis: Kept the default at `PT1S`.

Quiet traffic is unlikely to wait (the 1s gap is already gone unless a probe just ran). Under a 403 storm, lower toward `PT0S` if you need faster grant visibility; that costs more deny-path `max(id)` queries. Documented in `docs/server/auth.md` and `etc/conf/server.properties`.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
s-maling-celonis: Split into `probeCountAndMaxId()` (poller) and `probeMaxId()` (deny path). Shared reload / completed-probe bookkeeping stays in helpers.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
s-maling-celonis: Renamed to `probeStartedNanos`.

server/src/main/java/io/unitycatalog/server/auth/CasbinPolicyRefresher.java
```diff
@@ -76,47 +94,110 @@ private void pollQuietly() {
...
+    long probeStartedNanos = System.nanoTime();
+    Long maxId = readMaxId().orElse(null);
+    if (maxId == null) {
+      return ProbeResult.FAILED;
+    }
+    if (maxId == lastMaxId) {
+      recordCompletedProbe(probeStartedNanos);
+      return ProbeResult.UNCHANGED;
+    }
+    return reload(probeStartedNanos, lastCount, maxId);
```
yili-db: Here has a problem: the `lastCount` could be stale and doesn't match the updated `maxId`. This pair of non-matching values being recorded could shadow future changes. For example:
1. Initial state: (C, I)
2. A permission is granted, in DB it becomes (C+1, I+1). But in memory it's still (C, I)
3. A 403 happened so this function executes and reloads and it becomes (C, I+1) in memory which is half stale.
4. Any other permission is revoked. DB has (C, I+1). Memory also has (C, I+1).
5. Future reloads will be shadowed due to these two numbers in memory mistakenly match DB.

An easy fix could be replacing this line with delegating to `probeCountAndMaxId` when maxId is changed so that the count is also accurately updated.

s-maling-celonis: Good catch — fixed in e9dc529c.

When `max(id)` changes, `probeMaxId()` now delegates to `probeCountAndMaxId()` so recorded `(count, maxId)` stays consistent for later poller checks. Added `denyTriggeredReloadKeepsCountInSyncSoLaterRevokeIsVisibleToPoller` to cover the grant-then-non-max-revoke shadow case.

yili-db (approved):

tdas (approved):
LGTM on etc changes.

PR #1863, opened 2026-09-10
[Server] Give Hibernate and Casbin one HikariCP pool

s-maling-celonis:
**PR Checklist**

**Description of changes**

Hibernate currently uses its built-in non-production JDBC pool, and JCasbin opens a separate JDBC connection for policy storage. Those two paths have no shared bound, so total database connections are Hibernate's pool plus Casbin's extra connection.

This change puts both under one HikariCP pool so the cap is `hibernate.hikari.maximumPoolSize` (not Hibernate pool + 1). jdbc-adapter 2.7.0 still checks out one connection in its constructor and holds it for the authorizer lifetime, so this does not share connections with Hibernate and does not relieve reload-vs-catalog contention. Real sharing would need a per-operation-checkout adapter.

In detail:

- Adds HikariCP 7.1.0 and builds one pool in `HibernateConfigurator` from `hibernate.connection.*` plus optional `hibernate.hikari.*` keys.
- Hands that `DataSource` to Hibernate (`jakarta.nonJtaDataSource`) and to JCasbin's `JDBCAdapter`.
- Defaults pool autocommit to false so Hibernate can `rollback()` (MySQL and PostgreSQL reject rollback on autocommit connections). Casbin's one checkout is forced to autocommit so `loadPolicy` does not leave Postgres/MySQL idle-in-transaction.
- Closes the pool with the session factory, and closes the Casbin JDBC adapter with the authorizer.

Username resolution still prefers `hibernate.connection.username` and falls back to `hibernate.connection.user`.

**Test plan**

- [x] `HibernateConfiguratorTest` — pool size/name/autocommit, username fallback
- [x] `JCasbinAuthorizerTest` — grant/revoke/hierarchy/list; Casbin checkout is autocommit, Hibernate pool is not
- [x] `JCasbinAuthorizerMultiInstanceTest` — cross-replica refresh still works
- [x] `UnityAccessEvaluatorTest` — deny-path refresh
- [x] `TransactionManagerTest` — isolation restore and REPEATABLE_READ snapshot
- [x] `server/test` locally (498 passed, 2 skipped)
- [ ] `server/test` in CI

### DIFF
--- a/server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
+++ b/server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
@@ -8,7 +8,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.Properties;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
@@ -67,12 +66,7 @@ public class JCasbinAuthorizer implements UnityCatalogAuthorizer, AutoCloseable 
   public JCasbinAuthorizer(
       HibernateConfigurator hibernateConfigurator, ServerProperties serverProperties)
       throws Exception {
-    Properties properties = hibernateConfigurator.getHibernateProperties();
-    String driver = properties.getProperty("hibernate.connection.driver_class");
-    String url = properties.getProperty("hibernate.connection.url");
-    String user = resolveConnectionUsername(properties);
-    String password = properties.getProperty("hibernate.connection.password");
-    this.adapter = new JDBCAdapter(driver, url, user, password);
+    this.adapter = new JDBCAdapter(hibernateConfigurator.getDataSource());
 
     InputStream modelStream = this.getClass().getResourceAsStream("/jcasbin_auth_model.conf");
     this.modelText = IOUtils.toString(modelStream, StandardCharsets.UTF_8);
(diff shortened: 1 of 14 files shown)


server/src/main/java/io/unitycatalog/server/auth/JCasbinAuthorizer.java
```diff
@@ -67,12 +66,7 @@ public class JCasbinAuthorizer implements UnityCatalogAuthorizer, AutoCloseable
   public JCasbinAuthorizer(
       HibernateConfigurator hibernateConfigurator, ServerProperties serverProperties)
       throws Exception {
-    Properties properties = hibernateConfigurator.getHibernateProperties();
-    String driver = properties.getProperty("hibernate.connection.driver_class");
-    String url = properties.getProperty("hibernate.connection.url");
-    String user = resolveConnectionUsername(properties);
-    String password = properties.getProperty("hibernate.connection.password");
-    this.adapter = new JDBCAdapter(driver, url, user, password);
+    this.adapter = new JDBCAdapter(hibernateConfigurator.getDataSource());
```
yili-db: `jdbc-adapter` 2.7.0 grabs one connection in its constructor and holds it for the authorizer's whole lifetime, never returning it to the pool between operations. Given that, what does routing Casbin through the shared pool actually buy?
It doesn't share connections (Casbin monopolizes one; Hibernate can never use it) and it doesn't relieve reload-vs-catalog contention (reloads already ran on Casbin's own connection). The only real gain is a single unified cap on total DB connections instead of "Hibernate pool + 1".

And it adds a cost: the pool defaults `autoCommit` to false, but the adapter's `loadPolicy` reads never commit, so on Postgres/MySQL Casbin's held connection sits idle-in-transaction until the first write (permanently on a refresh-only follower). CI is H2-only, so this won't surface there. If the goal is just a unified connection bound, can we say that plainly in the description? Actual sharing would need a per-operation-checkout adapter, which this one isn't.

s-maling-celonis: You're right: jdbc-adapter 2.7.0 still takes one connection in the constructor and holds it, so this is a unified bound (`maximumPoolSize`), not connection sharing and not a fix for reload-vs-catalog contention. I updated the PR description and docs to say that. Per-operation checkout is not in any published adapter (including 2.13.0); that would be a follow-up.

The idle-in-transaction cost is real. The pool stays `autoCommit=false` for Hibernate's `rollback()`. Casbin's one `getConnection()` is forced to autocommit so `loadPolicy` does not leave a Postgres/MySQL transaction open on a refresh-only replica. Rather than adding an adapter, I opted for the simpler solution, adding a proxy around the pool.

yili-db (approved):

tdas (approved):
build.sbt change approved

PR #1858, opened 2026-09-08
[Server] Add managed storage cleanup task persistence

timothyw553:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1858/files/a59188b9961af2ce35920aa0c8ec62551b15c014..5c99e26b8596cd407a021d19936e9053bab201cf) to review incremental changes.
- [stack/managed-cleanup-schema](https://github.com/unitycatalog/unitycatalog/pull/1857) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1857/files)]
  - [**stack/managed-cleanup-repository**](https://github.com/unitycatalog/unitycatalog/pull/1858) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1858/files/a59188b9961af2ce35920aa0c8ec62551b15c014..5c99e26b8596cd407a021d19936e9053bab201cf)] ← _this PR_
    - [stack/managed-cleanup-fileio](https://github.com/unitycatalog/unitycatalog/pull/1865) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1865/files/5c99e26b8596cd407a021d19936e9053bab201cf..55108f3ba97921e33e457f6b1df48c18f8943bb1)]
      - [stack/managed-cleanup-worker](https://github.com/unitycatalog/unitycatalog/pull/1870) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1870/files/55108f3ba97921e33e457f6b1df48c18f8943bb1..7c9f901f059e326ba43cf3c41832a15cd16f4c46)]
        - [stack/managed-cleanup-drop](https://github.com/unitycatalog/unitycatalog/pull/1872) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1872/files/7c9f901f059e326ba43cf3c41832a15cd16f4c46..d08b9f3685c8bfbe176235754f8a2c5491dfef37)]
          - [stack/managed-cleanup-azure](https://github.com/unitycatalog/unitycatalog/pull/1895) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1895/files/d08b9f3685c8bfbe176235754f8a2c5491dfef37..cc07cbf623b2a29f68307c76fc928eb2386f9896)]
[...]

PR #1857, opened 2026-09-08
[Server] Add managed storage cleanup task schema

timothyw553:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1857/files) to review incremental changes.
- [**stack/managed-cleanup-schema**](https://github.com/unitycatalog/unitycatalog/pull/1857) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1857/files)] ← _this PR_
  - [stack/managed-cleanup-repository](https://github.com/unitycatalog/unitycatalog/pull/1858) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1858/files/a59188b9961af2ce35920aa0c8ec62551b15c014..5c99e26b8596cd407a021d19936e9053bab201cf)]
    - [stack/managed-cleanup-fileio](https://github.com/unitycatalog/unitycatalog/pull/1865) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1865/files/5c99e26b8596cd407a021d19936e9053bab201cf..55108f3ba97921e33e457f6b1df48c18f8943bb1)]
      - [stack/managed-cleanup-worker](https://github.com/unitycatalog/unitycatalog/pull/1870) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1870/files/55108f3ba97921e33e457f6b1df48c18f8943bb1..7c9f901f059e326ba43cf3c41832a15cd16f4c46)]
        - [stack/managed-cleanup-drop](https://github.com/unitycatalog/unitycatalog/pull/1872) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1872/files/7c9f901f059e326ba43cf3c41832a15cd16f4c46..d08b9f3685c8bfbe176235754f8a2c5491dfef37)]
          - [stack/managed-cleanup-azure](https://github.com/unitycatalog/unitycatalog/pull/1895) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1895/files/d08b9f3685c8bfbe176235754f8a2c5491dfef37..cc07cbf623b2a29f68307c76fc928eb2386f9896)]
[...]

PR #1865, opened 2026-09-12
[Server] Add cancellable local, S3, and GCS storage cleanup

timothyw553:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1865/files/5c99e26b8596cd407a021d19936e9053bab201cf..55108f3ba97921e33e457f6b1df48c18f8943bb1) to review incremental changes.
- [stack/managed-cleanup-schema](https://github.com/unitycatalog/unitycatalog/pull/1857) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1857/files)]
  - [stack/managed-cleanup-repository](https://github.com/unitycatalog/unitycatalog/pull/1858) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1858/files/a59188b9961af2ce35920aa0c8ec62551b15c014..5c99e26b8596cd407a021d19936e9053bab201cf)]
    - [**stack/managed-cleanup-fileio**](https://github.com/unitycatalog/unitycatalog/pull/1865) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1865/files/5c99e26b8596cd407a021d19936e9053bab201cf..55108f3ba97921e33e457f6b1df48c18f8943bb1)] ← _this PR_
      - [stack/managed-cleanup-worker](https://github.com/unitycatalog/unitycatalog/pull/1870) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1870/files/55108f3ba97921e33e457f6b1df48c18f8943bb1..7c9f901f059e326ba43cf3c41832a15cd16f4c46)]
        - [stack/managed-cleanup-drop](https://github.com/unitycatalog/unitycatalog/pull/1872) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1872/files/7c9f901f059e326ba43cf3c41832a15cd16f4c46..d08b9f3685c8bfbe176235754f8a2c5491dfef37)]
          - [stack/managed-cleanup-azure](https://github.com/unitycatalog/unitycatalog/pull/1895) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1895/files/d08b9f3685c8bfbe176235754f8a2c5491dfef37..cc07cbf623b2a29f68307c76fc928eb2386f9896)]
            - [stack/managed-cleanup-config-examples](https://github.com/unitycatalog/unitycatalog/pull/1904) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1904/files/cc07cbf623b2a29f68307c76fc928eb2386f9896..c80ca58ddda020a4d03b8ff33bffdc6e19aa3d0c)]

---------
**PR Checklist**

**Description of changes**

This combines local cleanup and the cloud wrapper from #1866. Local files, S3, and GCS use the same cooperative cancellation rules.

For example, a large directory should not need to fit in memory before deletion starts, and a large cloud prefix should stop starting new batches when its cleanup budget expires.

Local cleanup now deletes files while walking the directory tree, then deletes each directory after its children. Missing paths are already clean, so repeating cleanup is safe. Local directory listings stay lazy. Callers can close a listing early, and closing the FileIO also closes listings that were only partly read.

S3 and GCS cleanup use a fresh credentialed `ResolvingFileIO` wrapped by `InterruptiblePrefixOperations`. The wrapper uses one lazy listing and deletes batches of at most 1,000 objects. It is bound to one slash-delimited cleanup prefix. For example, it deletes objects matching `tables/id/`, including a directory marker if one is listed, but does not try to delete a separate object named `tables/id`. UC does not create that object, and trying to delete it when it is absent produces a misleading warning in GCS. Closing the wrapper closes the underlying FileIO.

S3 keeps the existing requirement for a configured region for each bucket; this PR does not add a global region fallback.

Both implementations use `CooperativeDeadline.checkCancelled()` between storage steps. It checks the attempt deadline and consumes a thread interrupt used for shutdown. The shared `NO_DEADLINE` instance disables deadline expiry but still checks interruption. Cloud calls use the provider's default request settings, with no custom socket timeout. Cancellation is checked between batches; it does not impose a hard deadline on an in-flight request.

This PR does not start a worker or change DROP. ADLS cleanup remains deferred.

Tests passed:

```sh
build/sbt javafmtAll
build/sbt "server/testOnly io.unitycatalog.server.persist.utils.SimpleLocalFileIOTest io.unitycatalog.server.persist.utils.InterruptiblePrefixOperationsTest io.unitycatalog.server.utils.FileOperationsTest"
```

All 51 tests passed. Coverage includes mid-traversal and between-batch deadline expiry, interrupt clearing, null prefixes, prefix boundaries, empty listings, directory markers, missing bucket regions with and without a global region setting, repeat-safe deletion, listing closure after partial reads or failure, fresh credentials, and S3/GCS configuration and closure through the resolving FileIO. GCS tests also exercise the real Iceberg implementation against a mocked storage service, including object-key handling and cancellation after one batch. These tests do not use live cloud buckets.


server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
```diff
@@ -104,44 +104,63 @@ public void deletePrefix(String prefix) {
...
-   * are returned in reverse lexicographic order (children before parents) so callers can delete
-   * bottom-up. Returns an empty iterable if the prefix does not exist. Callers must close the
-   * result to release the underlying directory stream.
+   * Walks the tree under {@code prefix} and returns a lazy, close-safe view of its regular files.
+   * Returns an empty iterable if the prefix does not exist. Callers must close the result to
+   * release the underlying directory stream.
    */
-  private static CloseableIterable<Path> walkPrefix(
-      String prefix, boolean includeDirectories, boolean bottomUp) {
+  private static CloseableIterable<Path> walkPrefix(String prefix) {
```
timothyw553, line 161: We simplify `walkPrefix` so it does one thing: lazily stream regular files under a prefix.Deletion now uses `walkFileTree` (see https://docs.oracle.com/javase/tutorial/essential/io/walk.html) separately.

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
```diff
@@ -91,9 +93,7 @@ public void deleteFiles(Iterable<String> pathsToDelete) throws BulkDeletionFailu
    */
   @Override
   public CloseableIterable<FileInfo> listPrefix(String prefix) {
-    return CloseableIterable.transform(
-        walkPrefix(prefix, false /* includeDirectories */, false /* bottomUp */),
-        SimpleLocalFileIO::toFileInfo);
+    return CloseableIterable.transform(walkPrefix(prefix), SimpleLocalFileIO::toFileInfo);
```
timothyw553, line 96: see below. this `listPrefix` now should just call the simplified `walkPrefix` method

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
```diff
@@ -175,11 +186,9 @@ private static FileInfo toFileInfo(Path path) {
     }
   }
 
-  private static void delete(Path path) {
```
timothyw553, line 178: we dont need `delete(Path)` anymore. the walkFileTree visitor now calls `Files.deleteIfExists(path);`

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
```diff
@@ -104,44 +104,63 @@ public void deletePrefix(String prefix) {
...
    * Recursively deletes everything under {@code prefix}, including the prefix directory itself.
    * Exposed as a static so callers with only a local path (and no FileIO instance) can reuse this
-   * logic; the instance {@link #deletePrefix(String)} delegates here.
-   *
-   * @throws UncheckedIOException wrapping a {@link FileNotFoundException} if the prefix does not
-   *     exist.
+   * logic; the instance {@link #deletePrefix(String)} delegates here. A missing prefix is already
+   * clean and returns successfully.
    */
   public static void deleteDirectory(String prefix) {
```
timothyw553, line 110: Old approach:
```java
paths = walkPrefix(prefix, includeDirectories=true, bottomUp=true);

for (Path path : paths) {
  delete(path);
}
```
It:
  1. Finds every path.
  2. Stores every path in memory.
  3. Sorts children before parents.
  4. Deletes them.
New approach:
```
Files.walkFileTree(prefix, visitor);
```
It deletes entries while walking:
```
Enter directory
    |
    +-- Find file       → delete file
    |
    +-- Visit children
    |
    +-- Leave directory → delete directory
```

timothyw553: The four callbacks mean:
- `preVisitDirectory`: check cancellation before entering.
- `visitFile`: check cancellation, then delete the file.
- `postVisitDirectory`: delete the directory after its children.
- `visitFileFailed`: ignore an entry that already disappeared; fail for other

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
```diff
@@ -104,44 +104,63 @@ public void deletePrefix(String prefix) {
...
+                throws IOException {
+              checkInterrupted();
+              if (failure instanceof NoSuchFileException) {
+                return FileVisitResult.CONTINUE;
+              }
+              throw failure;
+            }
+          });
+    } catch (NoSuchFileException e) {
+      // Cleanup is idempotent. Another attempt may have already removed the prefix.
```
yili-db: `NoSuchFileException` seems never thrown?

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
```diff
@@ -104,44 +104,63 @@ public void deletePrefix(String prefix) {
   /**
    * Recursively deletes everything under {@code prefix}, including the prefix directory itself.
    * Exposed as a static so callers with only a local path (and no FileIO instance) can reuse this
-   * logic; the instance {@link #deletePrefix(String)} delegates here.
-   *
-   * @throws UncheckedIOException wrapping a {@link FileNotFoundException} if the prefix does not
-   *     exist.
+   * logic; the instance {@link #deletePrefix(String)} delegates here. A missing prefix is already
+   * clean and returns successfully.
```
yili-db, line 133: Worth documenting the mechanics of clear/expect the interrupt flag.
Also is all other cloud FS `DelegateFileIO` implementations do it similarly?

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
timothyw553: ah yes the outer catch is redundant. removed

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
```diff
@@ -175,11 +184,9 @@ private static FileInfo toFileInfo(Path path) {
     }
   }
 
-  private static void delete(Path path) {
-    try {
-      Files.delete(path);
-    } catch (IOException e) {
-      throw new UncheckedIOException("Failed to delete " + path, e);
+  private static void checkInterrupted() {
+    if (Thread.currentThread().isInterrupted()) {
```
yili-db: Use `Thread.interrupted` so that it triggers only once. **If** the same thread is reused to call this list again it won't immediately blow up.

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
timothyw553: added docs. the check reads and clears the interrupt flag before throwing, so the thread can be reused. Cloud `DelegateFileIO` implementations do not reliably handle interruption, so cloud cleanup checks it between bounded requests.

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
timothyw553: done

yili-db (approved):

server/src/main/java/io/unitycatalog/server/cleanup/CleanupContext.java
```diff
@@ -0,0 +1,37 @@
+package io.unitycatalog.server.cleanup;
```
yili-db: CleanupContext lives in io.unitycatalog.server.cleanup, but it's now imported by SimpleLocalFileIO in persist.utils — a general-purpose local FileIO used for all local reads/writes, not just cleanup. So core storage IO now depends on the cleanup feature package, and it deepens a persist.utils ↔ cleanup package cycle (FileOperations.getCleanupFileIO → cleanup.InterruptiblePrefixOperations/CleanupContext, while cleanup.StorageCleanupWorker → persist.utils.FileOperations). The abstraction is fine; the name and location aren't — nothing about a (clock, deadline) cancellation check is cleanup-specific.

Move it to a neutral package (persist.utils, or a shared utils) and give it a neutral name (CooperativeDeadline / AttemptDeadline / CancellationCheck — "Context" is vague and "Cleanup" bakes in the feature). Then SimpleLocalFileIO depends on a generic cancellation concept instead of on cleanup, and the cycle shrinks.

server/src/main/java/io/unitycatalog/server/cleanup/InterruptiblePrefixOperations.java
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.iceberg.io.InputFile;
+import org.apache.iceberg.io.OutputFile;
+import org.apache.iceberg.io.SupportsPrefixOperations;
+
+/** Deletes one bound location in batches using the attempt's shared cancellation checks. */
+public final class InterruptiblePrefixOperations implements SupportsPrefixOperations {
+  private static final int MAX_BATCH_SIZE = 1000;
+
+  private final DelegateFileIO delegate;
+  private final String cleanupPrefix;
```
yili-db: Similar to https://github.com/unitycatalog/unitycatalog/pull/1865/changes/cd59dad42e0a0931fd8685e36bcd65f87de50b0b..2812ce42d2e0a3d7abc6183ce38d56b2c8f75130#r4030855689
`InterruptiblePrefixOperations` is named and shaped as a generic, reusable wrapper, not inherently about cleanup. 
A neutral `boundPrefix` / `targetPrefix` / or just `prefix` fits better.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -110,6 +120,34 @@ public FileIO getFileIO(NormalizedURL path, Set<CredentialContext.Privilege> pri
...
+    };
+  }
+
+  /**
+   * Builds a fresh S3 FileIO for one cleanup attempt. The socket timeout bounds a stalled request;
+   * the wrapper checks interruption and the attempt's deadline between list and delete requests.
+   */
+  private SupportsPrefixOperations getS3CleanupFileIO(
+      NormalizedURL path, Duration socketTimeout, CleanupContext cleanupContext) {
+    ValidationUtils.checkArgument(
```
yili-db: Move this into `getCleanupFileIO` so that later it doesn't have to be duped for Azure and GCP.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -110,6 +120,34 @@ public FileIO getFileIO(NormalizedURL path, Set<CredentialContext.Privilege> pri
...
+        socketTimeout.toMillis() > 0, "S3 socket timeout must be at least one millisecond");
+    Map<String, String> config = new HashMap<>(getFileIOConfig(path, CredentialContext.READ_WRITE));
+    config.put(
+        HttpClientProperties.APACHE_SOCKET_TIMEOUT_MS, Long.toString(socketTimeout.toMillis()));
+    S3FileIO fileIO = new S3FileIO();
```
yili-db: Or could this simply be similar to
```
      case S3, GS, ABFS, ABFSS -> {
        ResolvingFileIO fileio = new ResolvingFileIO();
        fileio.initialize(getFileIOConfig(path, privileges));
```

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
```diff
@@ -33,6 +39,17 @@
 public class SimpleLocalFileIO implements DelegateFileIO {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(SimpleLocalFileIO.class);
+  private final CleanupContext cleanupContext;
+
+  /** Creates local operations with interrupt checks and no cleanup deadline. */
+  public SimpleLocalFileIO() {
+    this(new CleanupContext(Clock.systemUTC(), Instant.MAX));
```
yili-db: `new CleanupContext(Clock.systemUTC(), Instant.MAX)` can be made a static instance inside `CleanupContext` class itself and named properly?

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
```diff
@@ -91,57 +108,80 @@ public void deleteFiles(Iterable<String> pathsToDelete) throws BulkDeletionFailu
    */
   @Override
   public CloseableIterable<FileInfo> listPrefix(String prefix) {
```
yili-db, line 109: This is pre-existing bug: the interface requires `Iterable<FileInfo>` but this function returns `CloseableIterable<FileInfo>` and expect callers to close it. This will result in leak.
It's fine today because no code actually calls `listPrefix`. With your PRs it will be called and this will be leaked. Please fix this pre-existing bug.

yili-db (changes requested):

server/src/main/java/io/unitycatalog/server/persist/utils/InterruptiblePrefixOperations.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.iceberg.io.OutputFile;
+import org.apache.iceberg.io.SupportsPrefixOperations;
+
+/** Deletes one bound location in batches with cooperative cancellation checks. */
+public final class InterruptiblePrefixOperations implements SupportsPrefixOperations {
+  private static final int MAX_BATCH_SIZE = 1000;
+
+  private final DelegateFileIO delegate;
+  private final String boundPrefix;
+  private final CooperativeDeadline deadline;
```
timothyw553, line 22: why do we need closeablegroup?

timothyw553: Because closing the FileIO must close every directory stream it opened. One FileIO can create multiple listings. 
`CloseableGroup` is Iceberg’s existing helper for that:
```
listings.addCloseable(walk); // Remember this stream.
listings.close();            // Close all remembered streams.
```

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
timothyw553: now using CloseableGroup see https://github.com/unitycatalog/unitycatalog/pull/1865/files/cd59dad42e0a0931fd8685e36bcd65f87de50b0b..45cbe9a2fc96e22088ec21091ff1e38f82e24b34#r4031409445 

and we close it.

server/src/main/java/io/unitycatalog/server/cleanup/CleanupContext.java
timothyw553: done . renamed it to `CooperativeDeadline`, moved it to `server.utils`

server/src/main/java/io/unitycatalog/server/cleanup/InterruptiblePrefixOperations.java
timothyw553: renamed it to `boundPrefix` and moved the wrapper to `persist.utils`. removed cleanup-specific wording too.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
timothyw553: yep done.

server/src/main/java/io/unitycatalog/server/persist/utils/SimpleLocalFileIO.java
timothyw553: Added CooperativeDeadline.NO_DEADLINE and reused it in the default constructor. It disables deadline expiry but still checks interruption.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
timothyw553: yeah just generic ResolvingFileIO and then that will get configured based on the path. do note that GS, ABFS are not supported right now

server/src/main/java/io/unitycatalog/server/persist/utils/InterruptiblePrefixOperations.java
```diff
@@ -0,0 +1,122 @@
...
+
+  /** Creates prefix operations bound to one location and sharing the given deadline. */
+  public InterruptiblePrefixOperations(
+      DelegateFileIO delegate, String boundPrefix, CooperativeDeadline deadline) {
+    if (!boundPrefix.endsWith("/")) {
+      throw new IllegalArgumentException("Bound prefix must end with '/'");
+    }
+    this.delegate = Objects.requireNonNull(delegate, "delegate");
+    this.boundPrefix = boundPrefix;
```
yili-db: nit: check null for this one as well, before checking for "/"

server/src/main/java/io/unitycatalog/server/persist/utils/InterruptiblePrefixOperations.java
```diff
@@ -0,0 +1,122 @@
...
+      while (batch.size() < MAX_BATCH_SIZE) {
+        deadline.checkCancelled();
+        if (!listing.hasNext()) {
+          break;
+        }
+        batch.add(listing.next().location());
+      }
+      deadline.checkCancelled();
+      if (batch.isEmpty()) {
+        delegate.deleteFile(boundPrefix.substring(0, boundPrefix.length() - 1));
```
yili-db: This is usually fine but it seems may generate spammy false positive warning logs like this:
```
  if (!clientForStoragePath(path).storage().delete(BlobId.fromGsUtilUri(path))) {
    LOG.warn("Failed to delete path: {}", path);
  }
```
Because in UC we never create that dummy directory object. Maybe this is something it can skip.

server/src/main/java/io/unitycatalog/server/persist/utils/FileOperations.java
```diff
@@ -171,11 +196,8 @@ private Map<String, String> getGCSConfig(GcpOauthToken gcpOauthToken, Long expir
 
   private Map<String, String> getS3Config(NormalizedURL path, AwsCredentials awsCredentials) {
     // TODO: if region isn't configured, use HEAD bucket to figure out
-    String s3Region = s3BucketRegionMap.get(path.getStorageBase());
+    String s3Region = s3BucketRegionMap.getOrDefault(path.getStorageBase(), defaultS3Region);
```
yili-db: This is change of behavior. We used to force region to be configured for each bucket. Now it's silently falling back to `AWS_REGION`

timothyw553: reverted.

server/src/main/java/io/unitycatalog/server/persist/utils/InterruptiblePrefixOperations.java
timothyw553: removed the extra root-object delete.

server/src/main/java/io/unitycatalog/server/persist/utils/InterruptiblePrefixOperations.java
timothyw553: added the null check

yili-db (approved):

PR #1866, opened 2026-09-12
[Server] Add interruptible S3 prefix cleanup

timothyw553:
Consolidated into #1865. This PR is closed to keep the stack smaller; its review history is preserved below.

## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1866/files/075ed77fa3812db64f6b4791809ce03b5ea0359e..f4dae760e5484b3cb46893f6b1b9801238d085e3) to review incremental changes.
- [stack/cleanup-persistence-schema-clean](https://github.com/unitycatalog/unitycatalog/pull/1857) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1857/files)]
  - [stack/cleanup-task-persistence-clean](https://github.com/unitycatalog/unitycatalog/pull/1858) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1858/files/4138f0ca295d403ebecd10f0ba09d0d6525b5d63..cd59dad42e0a0931fd8685e36bcd65f87de50b0b)]
    - [stack/storage-cleanup-attempt](https://github.com/unitycatalog/unitycatalog/pull/1865) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1865/files/cd59dad42e0a0931fd8685e36bcd65f87de50b0b..075ed77fa3812db64f6b4791809ce03b5ea0359e)]
      - [**stack/s3-storage-cleanup-adapter**](https://github.com/unitycatalog/unitycatalog/pull/1866) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1866/files/075ed77fa3812db64f6b4791809ce03b5ea0359e..f4dae760e5484b3cb46893f6b1b9801238d085e3)] ← _this PR_
        - [stack/storage-cleanup-task-worker](https://github.com/unitycatalog/unitycatalog/pull/1870) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1870/files/f4dae760e5484b3cb46893f6b1b9801238d085e3..eac58de9fb6991a16fcc66bcde96f89ec61c0ce8)]
          - [stack/storage-cleanup-polling](https://github.com/unitycatalog/unitycatalog/pull/1871) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1871/files/eac58de9fb6991a16fcc66bcde96f89ec61c0ce8..006bdc80e07b92f45b9fbe66c45edbc6d65a029d)]
[...]

PR #1870, opened 2026-09-12
[Server] Add background storage cleanup worker

timothyw553:
## 🥞 Stacked PR
Use this [link](https://github.com/unitycatalog/unitycatalog/pull/1870/files/55108f3ba97921e33e457f6b1df48c18f8943bb1..7c9f901f059e326ba43cf3c41832a15cd16f4c46) to review incremental changes.
- [stack/managed-cleanup-schema](https://github.com/unitycatalog/unitycatalog/pull/1857) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1857/files)]
  - [stack/managed-cleanup-repository](https://github.com/unitycatalog/unitycatalog/pull/1858) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1858/files/a59188b9961af2ce35920aa0c8ec62551b15c014..5c99e26b8596cd407a021d19936e9053bab201cf)]
    - [stack/managed-cleanup-fileio](https://github.com/unitycatalog/unitycatalog/pull/1865) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1865/files/5c99e26b8596cd407a021d19936e9053bab201cf..55108f3ba97921e33e457f6b1df48c18f8943bb1)]
      - [**stack/managed-cleanup-worker**](https://github.com/unitycatalog/unitycatalog/pull/1870) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1870/files/55108f3ba97921e33e457f6b1df48c18f8943bb1..7c9f901f059e326ba43cf3c41832a15cd16f4c46)] ← _this PR_
        - [stack/managed-cleanup-drop](https://github.com/unitycatalog/unitycatalog/pull/1872) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1872/files/7c9f901f059e326ba43cf3c41832a15cd16f4c46..d08b9f3685c8bfbe176235754f8a2c5491dfef37)]
          - [stack/managed-cleanup-azure](https://github.com/unitycatalog/unitycatalog/pull/1895) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1895/files/d08b9f3685c8bfbe176235754f8a2c5491dfef37..cc07cbf623b2a29f68307c76fc928eb2386f9896)]
            - [stack/managed-cleanup-config-examples](https://github.com/unitycatalog/unitycatalog/pull/1904) [[Files changed](https://github.com/unitycatalog/unitycatalog/pull/1904/files/cc07cbf623b2a29f68307c76fc928eb2386f9896..c80ca58ddda020a4d03b8ff33bffdc6e19aa3d0c)]

---------
**PR Checklist**

**Description of changes**

This adds one background `StorageCleanupWorker` per UC server. The server passes `ServerProperties` to its constructor and calls `start()` without arguments. The worker reads all cleanup timing settings when it is created. One scheduled thread polls for a task and performs its cleanup directly. `runOnce()` is package-private for direct unit testing. There is no separate poller class, attempt executor, future, or timeout interrupt.

Each run claims one ready task, validates its resource ID and path, creates fresh storage access, deletes the prefix, and checks that no descendants remain. Success removes the task. A storage failure, cancellation, or remaining file records a short error and schedules a retry. Reports and completion require the current lease token.

For example, if a large prefix reaches its 30-minute budget while processing batches, the next shared `CooperativeDeadline.checkCancelled()` stops the attempt. The worker reports the retry, and a later attempt continues from the remaining files. The deadline applies to storage work, not the database report or completion.

Server shutdown interrupts the polling thread and waits up to five seconds for it to exit. If it is still running, the worker logs a warning and returns so shutdown can continue. Tasks that were not completed remain eligible for retry after their leases expire. A restart cannot start another polling thread until the previous executor has terminated.

The defaults are a one-minute poll interval, a 30-minute cooperative attempt budget, a two-hour lease, a seven-day initial delay, and a one-hour retry delay. All are configurable in server properties. The initial delay postpones deletion; it does not provide table recovery. Invalid settings fail during startup; the lease must exceed the attempt timeout. Cloud calls use provider defaults, without a custom socket timeout. An in-flight call can outlast both the attempt budget and the shutdown grace period.

The worker uses the local, S3, and GCS cleanup support from its parent PR. This PR does not change DROP. ADLS remains unsupported.

The commented configuration examples in `etc/conf/server.properties` are reviewed separately in #1904.

Tests passed:

```sh
build/sbt javafmtAll
build/sbt "server/testOnly io.unitycatalog.server.cleanup.StorageCleanupWorkerTest io.unitycatalog.server.utils.ServerPropertiesTest"
```

All 36 tests passed. Coverage includes same-thread cleanup, deadline expiry, retry reporting, completion after the cleanup budget, configured polling, bounded shutdown when a call ignores interruption, interruption of the shutdown caller, repeated shutdown, and safe restart.


server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.function.BooleanSupplier;
+import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.io.FileInfo;
+import org.apache.iceberg.io.SupportsPrefixOperations;
+
+/**
+ * Claims and synchronously processes at most one ready storage cleanup task.
+ *
+ * <p>The time slice is cooperative: local and cloud implementations check it between deletion
+ * steps. An in-flight cloud request may continue until it returns or reaches its socket timeout.
```
timothyw553: in general, this can't really be enforced because a list or delete S3 request could just be in flight. that's fine for our internal purposes and it's just a best-effort approximation just so these cleanups dont hang forever.

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
```diff
@@ -0,0 +1,169 @@
...
+   */
+  public boolean runOnce() {
+    Optional<Claim> claimed = taskRepository.claim(leaseDuration, initialDelay);
+    if (claimed.isEmpty()) {
+      return false;
+    }
+
+    Claim claim = claimed.get();
+    Instant deadline = clock.instant().plus(attemptTimeSlice);
+    BooleanSupplier deadlineExpired = () -> !clock.instant().isBefore(deadline);
```
timothyw553: this is just a reusable check now() vs deadline. it is passed around many places including the delete loops

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
```diff
@@ -0,0 +1,169 @@
...
+
+  /** Deletes the claimed location and verifies that no descendants remain. */
+  private void cleanup(Claim claim, BooleanSupplier deadlineExpired) throws IOException {
+    checkCancelled(deadlineExpired);
+    NormalizedURL location = validateTask(claim);
+    String prefix = location + "/";
+    checkCancelled(deadlineExpired);
+    try (SupportsPrefixOperations prefixOperations =
+        fileOperations.getCleanupFileIO(location, socketTimeout, deadlineExpired)) {
+      prefixOperations.deletePrefix(prefix);
```
timothyw553, line 164: recall [deletePrefix from previous PR](https://github.com/unitycatalog/unitycatalog/pull/1866/files/6b600b680651997319ebf4e2d3d54fb2af5a8853..1f28baacc51438ce9666124032d4ed6f95d3fafd#diff-982a2096ce4a13c3170ed3ab0394ec597b0f7ab4963672b7be9d09f85b1d7a0eR70). it will try to delete until everything is gone

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
```diff
@@ -0,0 +1,169 @@
...
+      checkCancelled(deadlineExpired);
+      try (CloseableIterable<FileInfo> remaining =
```
timothyw553, line 166: but then i have this defensive check. if someone wrote to the location after deletePrefix finished, then its part of this check. 
```
deletePrefix() deletes what it sees
→ fresh listing finds a late file
→ throw “Storage cleanup did not empty the task prefix"
→ keep task and schedule retry
→ next attempt deletes the late file
→ finish only after a fresh listing is empty
```

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
```diff
@@ -0,0 +1,169 @@
...
+    this.clock = Objects.requireNonNull(clock, "clock");
+  }
+
+  /**
+   * Claims and processes at most one cleanup task on the calling thread.
+   *
+   * @return false if no task was ready; true if a task was claimed
+   */
+  public boolean runOnce() {
+    Optional<Claim> claimed = taskRepository.claim(leaseDuration, initialDelay);
```
yili-db: Naming like `claimed` vs `claim` isn't ideal. Maybe `maybeClaim`?

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
```diff
@@ -0,0 +1,169 @@
...
+      Duration socketTimeout,
+      Duration attemptTimeSlice,
+      Duration initialDelay,
+      Duration retryBackoff,
+      Clock clock) {
+    this.taskRepository = Objects.requireNonNull(taskRepository, "taskRepository");
+    this.fileOperations = Objects.requireNonNull(fileOperations, "fileOperations");
+    this.leaseDuration = Objects.requireNonNull(leaseDuration, "leaseDuration");
+    this.socketTimeout = Objects.requireNonNull(socketTimeout, "socketTimeout");
+    this.attemptTimeSlice = Objects.requireNonNull(attemptTimeSlice, "attemptTimeSlice");
```
yili-db: Is `attemptTimeSlice` always smaller than `leaseDuration`?

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
yili-db: I feel this can be simplified by introducing a small nested class for each cleanup effort. Then `deadline` can be a member variable and `deadlineExpired` can be an ordinary member function.

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+    return true;
+  }
+
+  /** Deletes the claimed location and verifies that no descendants remain. */
+  private void cleanup(Claim claim, BooleanSupplier deadlineExpired) throws IOException {
+    checkCancelled(deadlineExpired);
+    NormalizedURL location = validateTask(claim);
+    String prefix = location + "/";
+    checkCancelled(deadlineExpired);
```
yili-db: Don't need to call `checkCancelled` twice just for a simple sync call of `validateTask`

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
yili-db: Turn this into a comment.

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+    return location;
+  }
+
+  private static String safeError(Exception exception) {
+    return "Storage cleanup failed: " + exception.getClass().getSimpleName();
+  }
+
+  private static void checkCancelled(BooleanSupplier deadlineExpired) {
+    if (Thread.interrupted()) {
```
yili-db: I probably should raise this question in earlier PR already. But here is the same. Here are two mechanisms to stop a long running clean up here and in the FileIOs introduced in earlier PR:
1. `Thread.interrupted` which is triggered by some caller. But I suppose this is the caller, right?
2. deadline by time.

Why are there two?

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
timothyw553: done, renamed it to `maybeClaim`

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
timothyw553: yes. renamed it to `attemptTimeout` and now enforce that `leaseDuration > attemptTimeout + socketTimeout`

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
timothyw553: no more deadline plumbing

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
timothyw553: no more deadline plumbing

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
timothyw553: we have simplified to just using the interrupt signal. that can come from interrupt due to an attempt timing out or the server stopping. both are captured.

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
timothyw553: done, added this explanation above the fresh listing check.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -325,11 +364,13 @@ public static void main(String[] args) {
   public void start() {
     LOGGER.info("Starting Unity Catalog server...");
     server.start().join();
+    cleanupWorker.start(cleanupPollInterval);
```
timothyw553: point of interest

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -196,6 +218,16 @@ public enum Property {
...
         "server.authorization.policy-refresh-debounce-interval", "PT1S", DURATION_VALIDATOR),
+    STORAGE_CLEANUP_POLL_INTERVAL(
+        "server.storage-cleanup.poll-interval", "PT5S", POSITIVE_DURATION_VALIDATOR),
+    STORAGE_CLEANUP_ATTEMPT_TIMEOUT(
+        "server.storage-cleanup.attempt-timeout", "PT20S", POSITIVE_DURATION_VALIDATOR),
+    STORAGE_CLEANUP_LEASE_DURATION(
+        "server.storage-cleanup.lease-duration", "PT1M", POSITIVE_DURATION_VALIDATOR),
+    STORAGE_CLEANUP_INITIAL_DELAY(
+        "server.storage-cleanup.initial-delay", "PT1H5M", POSITIVE_DURATION_VALIDATOR),
+    STORAGE_CLEANUP_RETRY_BACKOFF(
```
yili-db, line 229: Backoff is too short? Consider this could be access error case.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -196,6 +218,16 @@ public enum Property {
         "server.authorization.policy-refresh-interval", "PT1M", DURATION_VALIDATOR),
     POLICY_REFRESH_DEBOUNCE_INTERVAL(
         "server.authorization.policy-refresh-debounce-interval", "PT1S", DURATION_VALIDATOR),
+    STORAGE_CLEANUP_POLL_INTERVAL(
+        "server.storage-cleanup.poll-interval", "PT5S", POSITIVE_DURATION_VALIDATOR),
+    STORAGE_CLEANUP_ATTEMPT_TIMEOUT(
+        "server.storage-cleanup.attempt-timeout", "PT20S", POSITIVE_DURATION_VALIDATOR),
```
yili-db: Both STORAGE_CLEANUP_ATTEMPT_TIMEOUT and STORAGE_CLEANUP_LEASE_DURATION are too short. Clean up of a bigger table can easily go beyond 1 minute. And this is a background task.
I propose something like 10 mins attempt and 15 mins lease.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -196,6 +218,16 @@ public enum Property {
...
+    STORAGE_CLEANUP_LEASE_DURATION(
+        "server.storage-cleanup.lease-duration", "PT1M", POSITIVE_DURATION_VALIDATOR),
+    STORAGE_CLEANUP_INITIAL_DELAY(
```
yili-db, line 227: Clean up 1 hour after table is deleted? That seems short too.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
```diff
@@ -196,6 +218,16 @@ public enum Property {
         "server.authorization.policy-refresh-interval", "PT1M", DURATION_VALIDATOR),
     POLICY_REFRESH_DEBOUNCE_INTERVAL(
         "server.authorization.policy-refresh-debounce-interval", "PT1S", DURATION_VALIDATOR),
+    STORAGE_CLEANUP_POLL_INTERVAL(
```
yili-db, line 221: 5 second polling is unnecessary. Something like 1 hour is good enough.

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
```diff
@@ -0,0 +1,209 @@
...
+  /** Stops polling and waits for an active attempt to finish. The worker may be started again. */
+  public synchronized void stop() {
+    if (executor == null) {
+      return;
+    }
+    executor.shutdownNow();
+    boolean interrupted = false;
+    while (!executor.isTerminated()) {
+      try {
+        executor.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
```
yili-db: This infinite wait is not ideal. Should have a bounded wait for some `gracePeriod`. Fixed value is fine.
The worker thread is a daemon (setDaemon(true)), so it dies at JVM exit regardless.
Infinite wait would cause the server to hang when shutting down.

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
```diff
@@ -0,0 +1,209 @@
...
+      Thread.currentThread().interrupt();
+    }
+  }
+
+  @Override
+  public void close() {
+    stop();
+  }
+
+  synchronized boolean isRunning() {
```
yili-db: `isRunning` called from anywhere?

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  /**
+   * Claims and processes at most one task on the calling thread. Deadline checks apply to storage
+   * work; database reporting and completion run after cleanup has returned. Do not call this
+   * directly while background polling is running.
+   *
+   * @return false if no task was claimed; true if a task was claimed, even if it needs a retry
+   */
+  public boolean runOnce() {
```
yili-db: should be private

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
```diff
@@ -167,10 +176,39 @@ private Server initializeServer(UnityCatalogServer.Builder unityCatalogServerBui
...
     return armeriaServerBuilder.build();
   }
 
+  private void initializeCleanup(ServerProperties serverProperties, Repositories repositories) {
+    cleanupWorker =
+        new StorageCleanupWorker(
+            repositories.getStorageCleanupTaskRepository(),
+            repositories.getFileOperations(),
+            Clock.systemUTC(),
+            serverProperties.getStorageCleanupLeaseDuration(),
```
yili-db: Might as well pass the entire `serverProperties` into this constructor.
Also, do the same to `getStorageCleanupPollInterval`

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
timothyw553: oh this was just test only... i can remove it.

server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java
timothyw553: yeah, i just pass the whole `serverProperties` now.

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
timothyw553: privatized.

server/src/main/java/io/unitycatalog/server/cleanup/StorageCleanupWorker.java
timothyw553: made it bounded.

server/src/main/java/io/unitycatalog/server/utils/ServerProperties.java
timothyw553: yes these were just placeholders. nwo the policy is as such:
Policy | Default
-- | --
Initial delay | 7 days
Polling | 1 minute between runs, one task at a time per UC server
Attempt budget | 30 minutes, checked between storage operations—not a hard request timeout
Lease | 2 hours from claiming; an abandoned task becomes eligible again after expiry
Failed or incomplete attempt | Retry after 1 hour, with no retry limit
Shutdown | Interrupt the worker and wait up to 5 seconds, then continue shutdown

yili-db (approved):

PR #1878, opened 2026-09-15
[Server] Paginate the Iceberg listing endpoints when the client asks

### LINKED ISSUE #1877: Iceberg REST listNamespaces and listTables ignore pageToken and never return next-page-token

The Iceberg REST endpoints `GET /v1/{prefix}/namespaces` and `GET /v1/{prefix}/namespaces/{namespace}/tables` ignore the `pageToken` and `pageSize` query parameters and never return `next-page-token`. They always answer with the whole listing: the handler walks every repository page (100 rows each) and buffers every identifier before responding.

This is legal -- the spec says a server that does not support pagination should ignore `pageToken` and return all results -- but it has costs. A schema with thousands of tables is one unbounded response that the server fully materializes, and a client that sets `rest-page-size` (Iceberg's `RESTCatalogProperties.PAGE_SIZE`) gets no benefit from it.

Honour the two parameters on both endpoints, gated exactly as the spec requires:

- Paginate when, and only when, the request carries `pageToken`. The spec is explicit that a server must return all results with a null `next-page-token` when `pageToken` is not set, and Iceberg's client sends `pageToken=""` on its first request, so the empty value counts as present.
- Treat `pageSize` as an upper bound; returning fewer is allowed.
- Return `next-page-token` while results remain, and null on the last page.
- A request without `pageToken` keeps today's behaviour exactly, so this is additive.

The gate matters: paginating unconditionally would make a client that ignores `next-page-token` read only the first page, which is the truncation bug #1790 fixed.

**Additional context**

Two implementation notes for whoever picks this up:

- The repository's page tokens are already name-based (`name > token`, ordered by name ascending), so they can serve directly as the opaque REST token.
- `listTables` filters out rows that carry no Iceberg metadata pointer *after* reading a page of rows, so a page of rows can yield fewer names than it holds. Filling a requested page therefore has to keep reading rows and derive the token from the last row consumed; deriving it from the last row of a repository page instead would either hand back short pages or re-scan rows.

Tests should cover: no `pageToken` returns everything with no token; `pageToken=` returns a first page plus a token; following tokens to the end yields every entry exactly once; `pageSize` respected as an upper bound; a schema mixing Iceberg and non-Iceberg tables across a page boundary; and an invalid `pageSize` answered 400.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1877.

`listNamespaces` and `listTables` ignored `pageToken` and `pageSize` and never returned
`next-page-token`, so every listing was the whole listing: the handler walked every repository page
and buffered every identifier before answering.

Both endpoints now honour the two parameters, gated the way the REST spec requires:

- A request that carries `pageToken` is answered one page plus a `next-page-token`; a request that
  does not keeps answering the whole listing with a null token. The gate is `pageToken`, not
  `pageSize`, because the spec says a server must return everything when `pageToken` is not set --
  and because paginating unconditionally would leave a client that ignores the token reading only the
  first page, which is the truncation #1790 fixed.
- `pageSize` is an upper bound, capped at the page the repository reads. A value below one, or one
  that is not a whole number, is a 400 naming the value.

`listTables` reads the Iceberg metadata pointer from the row rather than filtering in the query, so a
page of rows yields at most as many names as it holds and can yield none. Filling a requested page
therefore keeps reading rows and takes its token from the last row read, not from the last row of a
page whose remaining rows were never looked at. Without that, a schema whose Iceberg tables sort
after a page of Delta tables answers an empty page, or resumes by re-reading rows it already skipped.

Backward compatible: a client that sends neither parameter sees exactly what it sees today.

Verified against a running server (H2), with three namespaces, and a namespace holding three Delta
tables the Iceberg endpoints do not serve followed by two native Iceberg tables:

| request | answer |
| --- | --- |
| `GET /namespaces` | all three, `"next-page-token": null` |
| `GET /namespaces?pageSize=2` | all three -- no `pageToken`, so no pagination |
| `GET /namespaces?pageToken=&pageSize=2` | `s_a`, `s_b` and `"next-page-token": "s_b"` |
| `GET /namespaces?pageToken=s_b&pageSize=2` | `s_c`, token null |
| `GET /namespaces?pageToken=&pageSize=0` (also `-1`, `abc`) | 400 `Invalid pageSize: ... It must be a whole number of at least 1.` |
| `GET /namespaces?pageToken=&pageSize=` | the default page, all three |
| `GET /namespaces/s_a/tables` | the two Iceberg tables, token null |
| `GET /namespaces/s_a/tables?pageToken=&pageSize=1` | `z_ice_1` and a token, having read past the three Delta rows |
| `GET /namespaces/s_a/tables?pageToken=z_ice_1&pageSize=1` | `z_ice_2`, token null |

Tests: the two existing listing tests gained the paginated cases, reusing their fixtures -- the gate,
following tokens to the end, `pageSize` as an upper bound, the three rejected sizes, and a schema
whose Iceberg tables sort behind a full page of Delta tables. Both fail on `main` and pass here.
`server/test` is otherwise unchanged.

Not verified with Iceberg's own `RESTCatalog` client: the responses are built with Iceberg's
`ListNamespacesResponse` / `ListTablesResponse` builders and mapper, and the paginated requests above
are the exact query strings that client sends.


server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
```diff
@@ -1074,39 +1074,89 @@ public record IcebergTablePage(List<String> tableNames, Optional<String> nextPag
...
         },
         "Failed to list tables",
         /* readOnly= */ true);
   }
 
+  /**
+   * Collects the names {@link #listIcebergTables} answers with, and the name to resume after when
+   * the schema holds more rows than were read.
+   */
+  private IcebergTablePage collectIcebergTables(
```
yili-db: `collectIcebergTables` invents its own page token (`lastRowRead`) to stop partway through what `LISTING_HELPER` returns. It's correct (keyset is `name > :token`, names are dense), but it reaches around the opaque token and hardcodes the helper's scheme, and it fetches a full 100-row page to return maybe one — re-reading the overlap next request. That's the whole reason for the `moreRows`/`lastRowRead` bookkeeping.

Why not filter in the query? `uniform_iceberg_metadata_location` is a plain column, so `... IS NOT NULL` makes results == rows: `getNextPageToken` gives the token, `setMaxResults` becomes a real `LIMIT`, and this method collapses to one list call.

The Javadoc says the pointer is read from the row "rather than filtered in the query" to avoid re-resolving each table — but an `IS NOT NULL` predicate reads the same row in the same scan, so it keeps that and still paginates. Anything else forcing the app-side filter?

Rough shape: give the helper an optional `Function<Root<T>, Predicate>`, and pass `root -> root.get("uniformIcebergMetadataLocation").isNotNull()`. `collectIcebergTables` then deletes.

yili-db: However, even with this change, there is still no guarantee that each page will return exactly `maxResults`.
Check `applyResponseFilter(SecurableType.TABLE, listTablesResponse.getTables());` for example, which **will** be applied to Iceberg. It filters out tables according to permission. And in extreme case it can return an empty list but a valid page token and client is still required to call again.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -726,21 +741,37 @@ public HttpResponse reportMetrics(
...
+      // Without the parameter the whole listing is the answer, so follow the repository's page
+      // token to the end and say nothing about pages.
+      Optional<String> cursor = Optional.empty();
+      do {
+        IcebergTablePage page =
+            tableRepository.listIcebergTables(catalog, namespace, cursor, Optional.empty());
+        page.tableNames()
+            .forEach(table -> listed.add(TableIdentifier.of(Namespace.of(namespace), table)));
+        cursor = page.nextPageToken();
+      } while (cursor.isPresent());
```
yili-db: You can remove the `if` and fold two branches together, with only a small change:
```
    } while (requestedCursor.isPresent() && cursor.isPresent());
    return listed.nextPageToken(page.nextPageToken().orElse(null)).build();
```

server/src/main/java/io/unitycatalog/server/persist/TableRepository.java
moomindani: Nothing else forces the app-side filter -- you are right that `IS NOT NULL` reads the same row in the
same scan, so the property the Javadoc was defending (deciding from the row the page was read from,
rather than resolving each listed name a second time) survives the move into the query. Done in
0c60573d, in the shape you sketched: `PagedListingHelper.listEntity` takes an optional
`Function<Root<T>, Predicate>` and applies it next to the keyset and parent-id predicates, and
`listIcebergTables` passes `root -> root.get("uniformIcebergMetadataLocation").isNotNull()`.

`collectIcebergTables` is gone with its `moreRows` / `lastRowRead` bookkeeping, one page of rows is
one page of results again, and the net diff is smaller than before. The Javadoc now says the pointer
is decided in the query and keeps the note about not re-resolving each name.

moomindani: Agreed, and that is worth pinning rather than papering over: with the filter in the query a full page
always carries a token, so the last page of a paginated listing holds nothing even before
`applyResponseFilter` empties one. The listing makes no promise about how many entries a page holds --
only that a client follows the token until it is null. The updated test now does exactly that instead
of expecting the second page to be the last, so the empty-final-page shape is covered, and the
`IcebergTablePage` Javadoc no longer claims anything about page fullness.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
moomindani: Folded in 0c60573d, with the polarity the other way round: `} while (pageToken.isEmpty() &&
cursor.isPresent());`. A request that does not carry `pageToken` is the one answered with the whole
listing, so it is the case that has to keep following the repository's token; with
`requestedCursor.isPresent()` that request would stop after one page, which is the truncation #1790
fixed. `pageToken.isPresent()` then reads exactly one page and hands its token back, and the single
`return listed.nextPageToken(page.nextPageToken().orElse(null)).build()` you suggested covers both --
the whole-listing case ends with an empty token, so it renders as null.

server/src/main/java/io/unitycatalog/server/service/IcebergRestCatalogService.java
```diff
@@ -180,26 +181,40 @@ public ConfigResponse config(@Param("warehouse") Optional<String> catalogOpt) {
...
       // still resolved rather than assumed: per the REST spec, listing under a namespace that does
       // not exist is a 404, which an empty list would hide from the client.
       schemaRepository.getSchema(String.join(".", catalog, parent.get()));
-    } else {
-      // This endpoint returns the whole listing, so follow the repository's page token to the end.
-      Optional<String> pageToken = Optional.empty();
+      return ListNamespacesResponse.builder().build();
+    }
+
+    if (pageToken.isEmpty()) {
```
yili-db: The `listTables` below has a good unified implementation for both paged and non-paged listing. Can you please also follow that for schemas as well?

moomindani: Done in e4d57330 -- `listNamespaces` now has the same single loop, with the condition on the absence
of `pageToken`, and the local list the whole-listing branch accumulated is gone (the response builder
collects the namespaces directly). Both listings read the same way now.

PR #1884, opened 2026-09-15
[Spark] Align connector Jackson with Spark 4.1/4.2

s-maling-celonis:
## Summary
Spark-connector Jackson pins only (not the server Jackson 2.17.0 line):

- Spark **4.0**: Jackson **2.18.2**
- Spark **4.1+**: Jackson **2.21.6** (`jackson-annotations` **2.21**, major.minor-only since 2.20)

This also removes the old `libraryDependencies` (2.15.0) vs `dependencyOverrides` (2.19.2) mismatch.

lz4 is out of this PR: the connector does not reference it, and a `Provided` pin would not close CVE-2026-59949 on clusters that still ship Spark's 1.11.0.

## Test plan
- [ ] CI for `-DsparkVersion=4.0`, `4.1`, and `4.2`
- [ ] Spark connector tests compile against the per-line Jackson pins


yili-db:
Thanks for splitting these out. Built the connector locally for `4.0`/`4.1`/`4.2` — all compile, all versions resolve.

**Jackson:** Nice consolidation into `sparkJacksonVersion` / `sparkJacksonAnnotationsVersion`, and it fixes the old `libraryDependencies` (`2.15.0`) vs `dependencyOverrides` (`2.19.2`) mismatch. `jackson-annotations:2.21` is expected (annotations went major.minor-only in 2.20).

**lz4: please drop before merge.** The connector doesn't reference lz4, and it's `Provided` so it's never shipped — runtime still uses Spark's `1.11.0`, so CVE-2026-59949 isn't actually addressed (as your description notes). Net: build-time surface for no effect. Can we pull `sparkLz4Deps` out and keep this PR to the Jackson alignment? If we want that CVE closed, better at the Spark-image / runtime layer.

**Nits:**
- Keep the old `jackson-core >= 2.18` comment (Spark 4.2 `YAMLParser._updateToken`) — easy to lose the "why" on the next bump.
- One line on why annotations is `2.21` not `2.21.6`, so nobody "corrects" it.
- Description says "server Jackson 2.18.10 line"; server is `2.17.0` today.

s-maling-celonis:
Dropped `sparkLz4Deps` — agreed it is build-time surface with no runtime effect. Restored the `jackson-core >= 2.18` (`YAMLParser._updateToken`) comment and noted why annotations is `2.21` not `2.21.6`. PR description no longer claims a server Jackson 2.18.10 line.

PR #1887, opened 2026-09-15
[Server] Show a schema in the listing to the principal who owns it

### LINKED ISSUE #1105: Schema doesnt appear in schema list with auth enabled for user that created schema
**Describe the bug**

A user (other than admin), that has create schema rights, can create a schema. With auth enabled, when said user subsequently uses schema list, the schema doesn't appear in the list despite the user being the schema owner.

**To Reproduce**

1. enable authentication and start uc server
2. create a user: 
bin/uc --auth_token $(cat etc/conf/token.txt) user create --name "Your User" --email your@user.com
3. give user rights USE CATALOG and CREATE SCHEMA on catalog unity:
bin/uc --auth_token $(cat etc/conf/token.txt) permission create --securable_type catalog --name unity --privilege "USE CATALOG" --principal your@user.com
bin/uc --auth_token $(cat etc/conf/token.txt) permission create --securable_type catalog --name unity --privilege "CREATE SCHEMA" --principal your@user.com
4. Get authentication token for your new user: bin/uc auth login --output jsonPretty
5. create a new schema using token from step 4:
bin/uc --auth_token $user_token schema create --catalog unity --name test
6. list schemas with user:
bin/uc --auth_token $user_token schema list --catalog unity
7. Compare above list to what admin sees: 
bin/uc --auth_token $(cat etc/conf/token.txt) schema list --catalog unity
 
In step 6, the user should see the schema they created. In step 7 you confirm the schema has been created.

**Expected behavior**

In step 6, the user should see the schema they created in the list

-   OS: macOS Sequoia 15.7
-   Java 17
-   Using keycloak as idp
-   UC Commit b2d072e

**Additional context**

Granting USE SCHEMA on the newly created user lists the schema normally. Creating tables and viewing tables in the schema works normally, despite not granting any rights on that.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1105.

With authorization enabled, a principal who creates a schema does not see it in `listSchemas`, even
though the schema records them as its owner. The listing's authorization expression accepted only
`USE_SCHEMA` on the schema:

```
#authorize(#principal, #metastore, OWNER) ||
#authorize(#principal, #catalog, OWNER) ||
(#authorize(#principal, #schema, USE_SCHEMA) && ...)
```

`createSchema` grants the creator `OWNER` through `initializeHierarchicalAuthorization`, and
`getSchema` -- the method directly below it -- already accepts `#authorizeAny(#principal, #schema,
OWNER, USE_SCHEMA)`. So the two endpoints disagreed about the same principal and the same schema:
`GET /schemas/{full_name}` answered 200 while `GET /schemas?catalog_name=...` answered an empty list.

The fix makes the listing's schema clause match `getSchema`'s. The whole production diff:

```diff
-      (#authorize(#principal, #schema, USE_SCHEMA) &&
+      (#authorizeAny(#principal, #schema, OWNER, USE_SCHEMA) &&
            #authorizeAny(#principal, #catalog, OWNER, USE_CATALOG))
```

The catalog clause is untouched, so a principal still needs `OWNER` or `USE_CATALOG` on the catalog,
and still sees nothing they neither own nor hold `USE_SCHEMA` on. No new privilege is introduced and
no other endpoint changes.

**Why CI did not catch this.** `SdkSchemaAccessControlCRUDTest.testSchemaAccess` grants the creator
`USE_SCHEMA` on the schema it just created, immediately before listing it:

```java
SchemaInfo schemaRg2Info = regular2SchemasApi.createSchema(schemaRg2);
// give user USE SCHEMA on sch_rg2
grantPermissions(REGULAR_2, SecurableType.SCHEMA, "cat_pr1.sch_rg2", Privileges.USE_SCHEMA);
```

That extra grant is exactly the workaround the bug forces, so the existing coverage passed either
way. I left that test alone -- it still covers the `USE_SCHEMA` path -- and added a case that grants
ownership only.

**Tests.** `build/sbt "server/testOnly io.unitycatalog.server.sdk.access.*
io.unitycatalog.server.service.PermissionServiceTest io.unitycatalog.server.auth.*"` -- 57 passed.

The new `schemaCreatorSeesTheirOwnSchemaInTheListing` gives a principal `USE_CATALOG` and
`CREATE_SCHEMA` on a catalog, has them create a schema, and asserts the listing contains exactly
that schema -- with no `USE_SCHEMA` granted on top of ownership. `containsExactly` is deliberate: it
also pins that the principal still does not see `sch_pr1`, which belongs to someone else, so the fix
is not widening the listing. The test asserts `getSchema` succeeds too, which is what makes the old
inconsistency visible in one place. It fails on `main` (the listing comes back empty) and passes here.

Full `server/test`: 503 passed, 2 skipped, 2 failed, neither related --
`CloudCredentialVendorTest.testGenerateGcpTemporaryCredentials` fails the same way on an unmodified
`main`, and `SdkModelCRUDTest.testModelCRUD` is a timestamp collision (`updated_at` equal to
`created_at` inside the same millisecond) that passes 3/3 on this branch.

**Verified on a running server**, following the reporter's steps rather than only the test harness: an
auth-enabled server, a user created through SCIM2, `USE CATALOG` and `CREATE SCHEMA` granted on
`unity`, and that user's own token. Same server and same database state, only the build differs.

Before:

```
create schema as the user     201, "owner": "kavan@localhost"
list schemas as the user      []                                  <- their own schema is missing
get   schema as the user      200                                 <- but it is readable by name
list schemas as admin         ['default', 'kavan_test']
```

After:

```
list schemas as the user      ['kavan_test']                      <- and still not 'default'
get   schema as the user      200
list schemas as admin         ['default', 'kavan_test']
```

No `docs` change: this restores the documented intent (a schema owner can see their schema) rather
than changing a documented behaviour.

**Checked against managed Unity Catalog**, since this loosens an authorization check and the two
should not drift. In a workspace whose metastore is owned by an admin group, a non-admin service
principal was granted `USE CATALOG` and `CREATE SCHEMA` on a catalog and nothing at the schema level:

```
before creating anything      listing returns 1  (information_schema only)   <- the listing IS filtered
after it creates a schema     listing returns 2  (information_schema + its own schema)
explicit grants on it         {}                                            <- visibility from ownership alone
get by name                   200
schemas owned by others       not returned
```

So managed UC shows a schema to its owner on ownership alone, without `USE SCHEMA`, while still hiding
schemas owned by others. That is the behaviour this change gives the OSS listing, and the behaviour
`getSchema` here already had.


server/src/test/java/io/unitycatalog/server/sdk/access/SdkSchemaAccessControlCRUDTest.java
```diff
@@ -190,6 +190,30 @@ public void testSchemaAccess() {
     assertPermissionDenied(() -> principal1SchemasApi.createSchema(schemaWithLoc2));
   }
 
+  @Test
+  @SneakyThrows
+  public void schemaCreatorSeesTheirOwnSchemaInTheListing() {
```
yili-db: Can you fold this test into existing test here? Like around this line:
```
// get schema (principal-1) -> owner -> allowed
```
It seems to test schema owner can get. You can easily add a line after that to try list as well.

server/src/main/java/io/unitycatalog/server/service/SchemaService.java
```diff
@@ -96,7 +96,7 @@ public HttpResponse createSchema(
       """
       #authorize(#principal, #metastore, OWNER) ||
```
yili-db: Thanks for catching this.
This expression should be the same as getSchema. Please move it to `AuthorizeExpressions` as `GET_SCHEMA` and use it for both get and list so that they never diverge.

moomindani: Done in 2e03191c: `AuthorizeExpressions.GET_SCHEMA` is `getSchema`'s expression verbatim, and both
`getSchema` and `listSchemas` now name it. The two were equivalent once this fix landed -- the
listing's version carried catalog `OWNER` in its third clause, which the second clause
(`#authorize(#principal, #catalog, OWNER)`) already covers -- so nothing widened or narrowed in the
move. The Javadoc says why they share it: the listing's response filter asks the same question per
schema, so a narrower expression on either side hides a schema the caller can read one URL over.

server/src/test/java/io/unitycatalog/server/sdk/access/SdkSchemaAccessControlCRUDTest.java
moomindani: Folded into `testSchemaAccess` in 2e03191c, though a little earlier than you pointed at. After
`// get schema (principal-1) -> owner -> allowed`, principal-1 owns cat_pr1 as well, so its listing
passes on the catalog-owner clause and the assertion would hold with or without this fix (the
existing `list schemas (principal-1)` line above already covers that path).

The case that exercises the fix is regular-2 right after it creates `sch_rg2`: it owns that schema
and holds nothing else on it. So the listing assertion sits between the create and the
`USE SCHEMA` grant on the next line -- that grant is exactly what used to mask the bug. Checked by
restoring the old expression: the folded assertion fails.

PR #1890, opened 2026-09-16
[Server] Accept a capitalized boolean for the Iceberg dropTable's purgeRequested

moomindani:
**PR Checklist**

- [x] If you've added or modified a feature, documentation in `docs` is updated (nothing in `docs/` describes this parameter)

**Description of changes**

`pyiceberg` cannot drop an Iceberg table through Unity Catalog today. Both of its entry points fail with a 400:

```
>>> catalog.drop_table(("my_schema", "t"))    # sends purgeRequested=False
pyiceberg.exceptions.BadRequestError: BadRequestException: Can't convert 'False' to type 'Boolean'.
>>> catalog.purge_table(("my_schema", "t"))   # sends purgeRequested=True
pyiceberg.exceptions.BadRequestError: BadRequestException: Can't convert 'True' to type 'Boolean'.
```

pyiceberg hands a Python `bool` straight to `requests` (`params={"purgeRequested": purge_requested}` in `pyiceberg/catalog/rest/__init__.py`), and `requests` renders it `True` / `False`. Armeria converts a `Boolean` parameter through a fixed table of `true|TRUE|1` and `false|FALSE|0` (`StringUtil.stringToBoolean`), so neither spelling converts, and `AnnotatedValueResolver` is in an `internal` package, so the conversion is not pluggable. The Iceberg REST spec types this parameter `boolean` with `default: false`, so a capitalized spelling is a valid request and dropping the table is the correct answer.

This is the second of the two interop checks #1696 named as open when the native Iceberg write endpoints landed; the first (HEAD stalling) was fixed in #1801.

**Technical changes**

- `IcebergRestCatalogService.dropTable` binds `purgeRequested` as a `String` and validates it, reading `true` and `false` without regard to case.
- The spellings Armeria already accepted, `1` and `0` included, keep working, so the accepted set only grows. A value that is not a boolean at all is still a 400, now with a message that names the offending value (`Invalid purgeRequested: yes. It must be true or false.`) instead of Armeria's `Can't convert 'yes' to type 'Boolean'.`
- The value itself stays unused: purging is not implemented, and `dropTable` removes the table either way, as before.

**Two choices worth a reviewer's opinion**

- The helper validates rather than parses. Returning an `Optional<Boolean>` no caller reads would be new surface with nothing behind it, so the value is dropped after the check; when purge is implemented this becomes a parse.
- Only the Iceberg endpoint is fixed here. Ten `Optional<Boolean>` query parameters on the native surface (`force` on the catalog, schema, function, model, credential and external-location deletes, plus `include_browse`, `omit_properties`, `omit_columns`) reject `True` the same way -- measured on a local server. That is the native API's own contract rather than an Iceberg interop bug, and this repo's generated Python client is unaffected because it lowercases booleans, so I left it out and will file it separately.

**Tests**

- `testDropTableAcceptsCapitalizedPurgeRequested` walks `true, false, True, False, TRUE, FALSE, tRuE, 1, 0` and the empty value against a fresh table each, and asserts the table is gone afterwards.
- `testDropTableRejectsNonBooleanPurgeRequested` asserts `yes`, `purge` and `2` are a 400 with the exact message **and that the table is still there** -- a destructive endpoint must not act on input it refused.

Verified both directions: with the service file from `main` both new tests fail (`expected: 204 but was: 400` for `True`, and Armeria's message for `yes`); with the change all 19 tests in the class pass.

```
sbt "server/testOnly io.unitycatalog.server.service.IcebergRestCatalogTest"
```

End-to-end against a local server with pyiceberg 0.12.0: `drop_table` and `purge_table` both fail with the 400 above on `main` and both succeed on this branch, with the table gone from `list_tables` afterwards.

**Backward compatibility**: the set of accepted values grows and nothing that worked stops working. The only change for an existing caller is the message text on a non-boolean value (same 400).


moomindani:
Filed #1891 for the ten `Optional<Boolean>` query parameters on the native REST surface that reject `True` the same way. Deliberately out of scope here: that is the native API's own contract rather than an Iceberg interop bug, and ten call sites want one shared helper instead of ten copies of this check.

yili-db (approved):

PR #1892, opened 2026-09-16
Stop shipping a Log4j SLF4J binding on library artifacts

s-maling-celonis:
## Summary
- `commonSettings` declared `log4j-slf4j2-impl` (and a direct `log4j-api`) at compile scope on every module. That leaked an SLF4J *implementation* into published libraries: `unitycatalog-client`, `unitycatalog-hadoop`, and `unitycatalog-spark_*`.
- A client should only depend on `slf4j-api`. Adding the client to a Spring Boot / Logback app without exclusions produces two SLF4J bindings; with `log4j-to-slf4j` on the classpath Log4j refuses to start.
- Move Log4j2 to the processes that actually use it: server and CLI (`Configurator` + `*.log4j2.properties`). `log4j-core` is compile-scope there because `Configurator` lives in core. Direct `log4j-api` is unused in library (and server) source and is omitted.

## Test plan
- [x] `build/sbt client/compile server/compile cli/compile`
- [ ] Confirm generated POMs: client/hadoop/spark no longer list `log4j-slf4j2-impl`; server still does
- [ ] Add `io.unitycatalog:unitycatalog-client` to a Spring Boot app with Logback and no exclusions — app starts; no dual SLF4J binding warning
- [ ] Run the UC server and CLI and confirm Log4j2 config still loads


yili-db (approved):

tdas (approved):
good catch. thanks for fixing.

PR #1897, opened 2026-09-17
[Server] Accept a capitalized boolean for the native REST query parameters

### LINKED ISSUE #1891: [Server] Boolean query parameters reject the capitalized spelling a hand-written client sends
**Describe the bug**

Every `Optional<Boolean>` query parameter on the native REST surface rejects `True` / `False` with a 400, although the OpenAPI spec types them `boolean`. Armeria converts a `Boolean` parameter through a fixed table of `true|TRUE|1` and `false|FALSE|0` (`com.linecorp.armeria.internal.common.util.StringUtil.stringToBoolean`), so a capitalized value never converts. `AnnotatedValueResolver` is in an `internal` package, so the conversion cannot be replaced globally.

Ten parameters are affected:

| Endpoint | Parameter |
| --- | --- |
| `DELETE /catalogs/{name}` | `force` |
| `DELETE /schemas/{full_name}` | `force` |
| `DELETE /functions/{name}` | `force` |
| `DELETE /models/{full_name}` | `force` |
| `DELETE /credentials/{name}` | `force` |
| `DELETE /external-locations/{name}` | `force` |
| `GET /volumes` (both list handlers) | `include_browse` |
| `GET /tables` | `omit_properties`, `omit_columns` |

`force` is the one that matters: it is acted on, it decides whether a delete cascades, and a caller who spells it `True` gets a 400 instead of the delete they asked for.

**To Reproduce**

1. Start a server.
2. `curl -X DELETE 'http://localhost:8080/api/2.1/unity-catalog/catalogs/c?force=True'`

   ```
   400 {"error_code":"INVALID_ARGUMENT","message":"Can't convert 'True' to type 'Boolean'."}
   ```
3. The same request with `force=true` (or `force=1`) is accepted.

Measured the same way for `include_browse`, `omit_properties` and `omit_columns`: 400 on `True`, 200 on `true`.

`True` is what a hand-written Python client sends -- `requests.delete(url, params={"force": True})` renders the query as `force=True`. The generated Python client in this repo is not affected, because its `parameters_to_url_query` lowercases booleans, so this reaches scripts written directly against the REST API.

**Expected behavior**

A parameter the spec types `boolean` accepts `true` and `false` without regard to case. The values Armeria already converts, `1` and `0` included, keep working, so the accepted set only grows and no existing caller changes.

**System**

- Server built from `main` (0.7.0-SNAPSHOT), macOS, JDK 17.

**Additional context**

The same defect on the Iceberg REST surface (`purgeRequested` on `dropTable`) is fixed in #1890, where it also stops `pyiceberg`'s `drop_table` and `purge_table` from working at all. That PR validates the value as a `String` in the one handler it touches; doing the same for ten parameters wants one shared helper rather than ten copies, which is why this is filed separately. I am happy to send that patch.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1891.

Every query parameter the OpenAPI spec types as a `boolean` rejected `True` / `False` with a 400. Armeria converts a `Boolean` parameter through a fixed table of `true|TRUE|1` and `false|FALSE|0`, so a capitalized spelling never converted:

```
$ curl -X DELETE 'http://localhost:8080/api/2.1/unity-catalog/catalogs/c?force=True'
400 {"error_code":"INVALID_ARGUMENT","message":"Can't convert 'True' to type 'Boolean'."}
```

`True` is what a hand-written Python client sends -- `requests.delete(url, params={"force": True})` renders the query as `force=True`. `force` is the parameter that matters: it is acted on, it decides whether a delete cascades, and a caller who spells it that way gets a 400 instead of the delete they asked for. The generated Python client in this repo is unaffected, because its `parameters_to_url_query` lowercases booleans, so this reaches scripts written directly against the REST API.

**Technical changes**

- `ValidationUtils.parseBooleanParam` reads a query parameter typed as a boolean: `true` and `false` without regard to case, plus the `1` / `0` Armeria already converted, so the accepted set only grows and no existing caller changes. An empty value stays what it is today, absent. Anything else is a 400 naming the parameter and the offending value.
- Ten parameters bind as `Optional<String>` and convert through that helper: `force` on the six deletes (catalogs, schemas, functions, models, credentials, external locations), `include_browse` on both volume read handlers, and `omit_properties` / `omit_columns` on `listTables`.
- `TestUtils` gains `sendRawGet` / `sendRawDelete` over the existing raw-request sender.

Armeria's conversion cannot be replaced globally: `AnnotatedServiceTypeUtil` looks `Boolean` up in a fixed map of types and only falls back to a type's own `of` / `valueOf` / `fromString` for classes outside it, and `AnnotatedValueResolver` is in an `internal` package. So the conversion has to move into the handlers, and one shared helper keeps it in a single place rather than ten copies. The same defect on the Iceberg REST surface (`purgeRequested` on `dropTable`) is fixed in #1890, which is independent of this PR; whichever merges first, the other switches to the shared helper.

**Compatibility**: the spec is unchanged -- it types these parameters as `boolean` already, which is why a capitalized spelling is a valid request. Every spelling accepted before is still accepted, so this is purely additive.

**Tests** (`server/src/test/java/io/unitycatalog/server/service/BooleanQueryParamTest.java`)

The cases go over raw HTTP, because the generated client cannot send the values under test. They pin that the value is *read*, not merely accepted: `force=False` refuses to delete a catalog that holds a schema while `force=True` cascades it, and `omit_columns` / `omit_properties` decide what the table listing carries. All four tests fail on the unfixed handlers.

```
build/sbt "server/testOnly io.unitycatalog.server.service.BooleanQueryParamTest"   # 4 passed
build/sbt -J-Xmx2G "server/test"   # 507 passed, 1 pre-existing env failure (GCP credentials), 2 skipped
```


yili-db:
Should the native REST API accept `True`/`False` at all? I don't think it should.

For this surface `all.yaml` is the source of truth, and it types these as `type: boolean`, which serializes `true`/`false`. `True` isn't a spelling of that; it's a client handing Python's `str(True)` to `requests`. So this isn't "the spec allows a capitalized spelling," it's us accepting input the spec doesn't describe, permanently (we can't walk it back) and unwritten-down (OpenAPI can't express a case-insensitive boolean). Our own generated clients already send `true`/`false`; the only thing this helps is a client serializing booleans wrong, and accepting it just hides that bug until it hits a spec-conformant gateway or validator.

moomindani:
You are right, and I am closing this.

`all.yaml` types these as `boolean`, and `True` is not a spelling of that -- it is a client handing
Python's `str(True)` to `requests`. Accepting it would put an undocumented input into the API
permanently, in a form OpenAPI cannot express, and the only client it helps is one serializing
booleans wrongly; the repository's own generated clients already send `true`/`false`.

This is also the line I drew myself when I split it out of #1890: that one was an interop bug, where
pyiceberg cannot drop or purge a table at all because it is pyiceberg that spells the value, and the
Iceberg REST spec is the contract being met. Here the contract is ours, and it says `true`/`false`.

Closing #1891 with the same reasoning.

PR #1899, opened 2026-09-17
[Server] Answer a volume that does not exist with a not-found

### LINKED ISSUE #1898: [Server] A volume that does not exist answers 500 instead of 404
**Describe the bug**

`VolumeRepository.getVolume` never checks whether the lookup found anything: `getVolumeDAO` returns
`query.uniqueResult()`, which is `null` for a volume that does not exist, and the caller dereferences
it. The NPE is wrapped as `INTERNAL`, so the request is answered 500 -- and the message carries an
internal DAO class name.

Three surfaces reach it, all through that one call:

| Request | On `main` | Expected |
| --- | --- | --- |
| `GET /volumes/{full_name}` | 500 `INTERNAL` | 404 |
| `DELETE /volumes/{full_name}` | 500 `INTERNAL` | 404 |
| `GET /permissions/volume/{full_name}` | 500 `INTERNAL` | 404 |

`PATCH /volumes/{full_name}` is correct (404 `NOT_FOUND`, "Volume not found: ..."), because
`updateVolume` does its own null check -- as does `deleteVolume(Session, UUID, String)`. So the fix is
the check those two already have, in the one read path that is missing it.

Volumes are the only resource that behaves this way. Measured on the same server:

```
GET /tables/{missing}    404 TABLE_NOT_FOUND    Table not found: ...
GET /functions/{missing} 404 NOT_FOUND          Function not found: ...
GET /models/{missing}    404 NOT_FOUND          Registered model not found: ...
GET /schemas/{missing}   404 SCHEMA_NOT_FOUND   Schema not found: ...
GET /catalogs/{missing}  404 CATALOG_NOT_FOUND  Catalog not found: ...
```

**To Reproduce**

1. Start a server, then create a catalog and a schema:

   ```
   curl -X POST .../catalogs -H 'Content-Type: application/json' -d '{"name":"c"}'
   curl -X POST .../schemas  -H 'Content-Type: application/json' -d '{"name":"s","catalog_name":"c"}'
   ```

2. Ask for a volume that was never created:

   ```
   curl .../volumes/c.s.nope
   ```

3. The answer is 500, and the message names an internal class:

   ```
   500 {"error_code":"INTERNAL","message":"Failed to get volume: Cannot invoke
   \"io.unitycatalog.server.persist.dao.VolumeInfoDAO.toVolumeInfo(String, String)\" because the
   return value of \"io.unitycatalog.server.persist.VolumeRepository.getVolumeDAO(org.hibernate.Session,
   String, String, String)\" is null"}
   ```

   `DELETE /volumes/c.s.nope` and `GET /permissions/volume/c.s.nope` answer the same way.

**Expected behavior**

A volume that does not exist is a 404 with the message the other volume paths already produce
("Volume not found: c.s.nope"), and no internal class name in the response. A 500 also misleads a
client's retry logic: it reads as a server fault worth retrying, when nothing about the request will
ever succeed.

-   OS: macOS, JDK 17, server built from `main` (0.7.0-SNAPSHOT)

**Additional context**

With authorization enabled the same call sits in `KeyMapper` (it resolves a volume key through
`getVolume`), so an authorization check against a missing volume also lands on the 500 rather than a
404 or a denial.

I am happy to send the patch.

moomindani:
**PR Checklist**

**Description of changes**

Fixes #1898.

`VolumeRepository.getVolume` looked the DAO up and used it without checking that the lookup found anything. `getVolumeDAO` returns `query.uniqueResult()`, which is `null` for a volume that was never created, so the NPE was wrapped as `INTERNAL`: the request was answered 500, with an internal DAO class name in the message.

Three surfaces read the volume through that one call, and all three behaved that way. Measured against a local server, before:

```
GET    /volumes/c.s.nope             500 INTERNAL   Failed to get volume: Cannot invoke "io.unitycatalog.server.persist.dao.VolumeInfoDAO.toVolumeInfo(String, String)" because ... is null
DELETE /volumes/c.s.nope             500 INTERNAL   (same)
GET    /permissions/volume/c.s.nope  500 INTERNAL   (same)
```

after:

```
GET    /volumes/c.s.nope             404 NOT_FOUND  Volume not found: c.s.nope
DELETE /volumes/c.s.nope             404 NOT_FOUND  Volume not found: c.s.nope
GET    /permissions/volume/c.s.nope  404 NOT_FOUND  Volume not found: c.s.nope
```

**Technical changes**

`getVolume` now carries the same lookup check `updateVolume` and `deleteVolume(Session, UUID, String)` already had, with the same error code and message. That is why `PATCH /volumes/{full_name}` was already correct at 404, and why this is the missing half of a check the repository already makes everywhere else.

A 404 is also what every other resource type already answers for a name that does not exist -- tables, functions, models, schemas and catalogs -- so volumes were the only outlier. And a 500 misleads a client's retry logic: it reads as a server fault worth retrying, when nothing about the request will ever succeed.

**Compatibility**: the status for a missing volume changes from 500 to 404 on the three paths above, which is the fix. Nothing changes for a volume that exists, for the listing, or for `PATCH`.

**Tests**

- `SdkVolumeCRUDTest.testGetAndDeleteVolumeThatDoesNotExist` pins the code and message on the `GET` and `DELETE` paths.
- `PermissionServiceTest.permissionsOnAVolumeThatDoesNotExistAreNotFound` pins the permissions path, where the securable is resolved through the same call, with authorization enabled.

Both fail on the unfixed repository (`expected: 404 but was: 500`).

```
build/sbt "server/testOnly io.unitycatalog.server.sdk.volume.SdkVolumeCRUDTest io.unitycatalog.server.service.PermissionServiceTest"   # 8 passed
build/sbt -J-Xmx2G "server/test"   # 505 passed, 1 pre-existing env failure (GCP credentials), 2 skipped
```

Verified on a local server as well, with the requests shown above.


server/src/test/java/io/unitycatalog/server/service/PermissionServiceTest.java
```diff
@@ -129,6 +129,17 @@ public void unauthenticatedRequestIsRejected() {
         "authorization");
   }
 
+  @Test
+  public void permissionsOnAVolumeThatDoesNotExistAreNotFound() {
+    // The securable is resolved through VolumeRepository.getVolume, so this endpoint answered the
+    // same 500 as GET /volumes for a volume that was never created.
+    String missingVolume = SCHEMA_FULL_NAME + ".no_such_volume";
```
yili-db: Please fold this test into `testPermissionsServiceUseCases`

moomindani: Folded into `testPermissionsServiceUseCases` in b1791bc5, next to the other read-backs at the end of
it. One server start fewer, and the case reads as one more securable the owner queries.

PR #1901, opened 2026-09-17
fix(server): keep a duplicate table name from being persisted twice

### LINKED ISSUE #1837: Concurrent creates of the same table name persist duplicate rows, permanently breaking that name
**Describe the bug**

`uc_tables` has no unique constraint on `(schema_id, name)` -- `TableInfoDAO` declares only
`@Index(name = "idx_name", columnList = "name")` -- and `TableRepository.createTable` checks for an
existing name and then inserts in separate statements. Two concurrent creates of the same table name
therefore both pass the check and both insert, leaving two rows with the same `(schema_id, name)`.

Every later lookup of that name fails, because the repository resolves a table by name with
`Query.uniqueResult()`, which throws `NonUniqueResultException` on two rows: `getTable` answers 500,
and so does `deleteTable`, so the duplicate cannot be removed through the API and the name stays
broken. The Iceberg REST endpoints resolve the same way through `findTableOrThrow`.

The same check-then-insert shape, with no unique constraint behind the column it checks, appears
elsewhere: `SchemaRepository.createSchema` (`uc_schemas` declares no constraints at all),
`ModelRepository.createRegisteredModel` (`uc_registered_models` has a plain name index),
`StagingTableRepository`, and `UserRepository.createUser`, whose email check has no constraint behind
it -- `uc_users` constrains `external_id` only. I have only reproduced the table case.

**To Reproduce**

1. Start the server and create a catalog and a schema.
2. Send two `POST /api/2.1/unity-catalog/tables` requests for the same table name at the same moment
   (two threads released by a barrier). Both answer 200, with different `table_id`s -- reproduced on
   the first attempt against `main` at 58d5c7b:
   `[created:20b45618-2046-4c1e-83cd-8fcbcaf12cfa, created:8dc1d9a9-23e9-40ec-b936-31d44b94399f]`
3. `GET /api/2.1/unity-catalog/tables/{catalog}.{schema}.{table}` now answers
   `500 INTERNAL`, `"message": "Failed to get table: Query did not return a unique result: 2 results were returned"`
   (from `TableRepository.getTable`).
4. `DELETE` on the same name answers `500 INTERNAL`,
   `"message": "Failed to delete table: Query did not return a unique result: 2 results were returned"`
   (from `TableRepository.deleteTable`), so the rows cannot be removed through the API. The name stays
   unusable until the database is edited directly.

**Expected behavior**

One of the two creates succeeds and the other fails with `TABLE_ALREADY_EXISTS`, which is what a
single create against an existing name already returns. A duplicate `(schema_id, name)` should not be
representable in the schema.

-   OS: macOS 15, server built from `main` at 58d5c7b (H2 store, but the missing constraint is not store-specific)

**Additional context**

A unique constraint on `(schema_id, name)` would close it at the source, and would let
`createTable` translate the constraint violation into `TABLE_ALREADY_EXISTS` instead of relying on
the preceding read. That implies a schema change for existing databases, which is adjacent to the
"database schema upgrades" item on the roadmap, so I am filing this rather than sending a patch --
happy to implement whichever direction the maintainers prefer.

tiennguyen-onehouse:
## Summary

Fixes #1837.

`createTableImpl` resolves the schema, calls `findBySchemaIdAndName`, and inserts, all inside one transaction, and `uc_tables` declared only `@Index(name = "idx_name", columnList = "name")`. Under read-committed neither transaction sees the other's uncommitted row, so two concurrent creates of one name both pass the check and both commit.

The second row is what does the lasting damage. `findBySchemaIdAndName` resolves with `uniqueResult()`, so once two rows share `(schema_id, name)` every later lookup of that name throws `NonUniqueResultException`: `getTable` answers 500 and so does `deleteTable`, which leaves the name unusable and unremovable through the API. The Iceberg REST endpoints resolve the same way through `findTableOrThrow`.

Changes:

- **Constrain `(schema_id, name)` on `uc_tables`**, following the pattern `ColumnInfoDAO`, `DeltaCommitDAO`, `PropertyDAO` and `UserDAO` already use in the same package. The database, not the check, is what actually decides the race.
- **Translate the violation into `TABLE_ALREADY_EXISTS`** so the losing request gets a 409 instead of a 500. The insert is flushed inside the create block rather than at commit, because a violation raised at commit surfaces outside the block and `TransactionManager` wraps it as `INTERNAL`.
- The existing pre-check stays. It still gives the better message on the ordinary non-racing path, and it avoids a doomed insert.
- **`renameTable` has the same shape** and the same outcome: it checks whether the target name is taken and then writes, so four concurrent renames onto one name all land. The constraint already stops that write; this catches the violation there too so the losing rename reports a conflict rather than an internal error. Not mentioned in the issue, found while checking what else writes a name into `uc_tables`.

## Why the database rather than the application

Worth stating, since the check has been in application code since the beginning:

- Nothing suggests the omission was deliberate. `ColumnInfoDAO` and `PropertyDAO` have carried unique constraints since the first commit, and `UserDAO` gained one as recently as #1691, so this is the established direction rather than a departure. I found no comment, doc or review note arguing against constraints on this table.
- Two statements cannot be made atomic from the application side without either serializing creates or taking a lock on the parent schema. The constraint is the cheap correct answer, which is presumably why the other four tables use it.
- It should cost nothing on reads and plausibly helps. The only name-based query is `WHERE t.schemaId = :schemaId AND t.name = :name`, and listing filters by schema and orders by name. A composite index on `(schema_id, name)` serves both better than the existing `idx_name` on `name` alone, which no query matches on its own. I left `idx_name` in place rather than widen the change, but it now looks redundant.
- The write cost is one more index maintained on a table whose inserts are table creations, which are rare next to the lookups every query performs.

The violation is matched on the constraint name rather than caught wholesale, so a request with duplicate column names keeps reporting the column constraint it actually tripped. The match is by containment because dialects report it differently: Postgres names the constraint, H2 names the backing index and appends a generated suffix (`UC_TABLES_SCHEMA_ID_NAME_KEY_INDEX_9`).

## Upgrade behaviour

`hibernate.hbm2ddl.auto=update` decides whether an existing deployment picks the constraint up. I measured both cases by creating a database with a build of `main`, then opening it with this branch:

| Existing database | Server starts | Constraint added | Protected after upgrade |
|---|---|---|---|
| No duplicates | yes | yes, as `UC_TABLES_SCHEMA_ID_NAME_KEY` | yes |
| Already contains duplicates | yes | no | no |

So the upgrade is safe either way — no server fails to start — and a clean database is protected automatically. A database that already carries duplicate rows keeps them, silently stays unprotected, and needs those rows reconciled by hand before the constraint can be created. I did not add a startup check or a repair step for that; if you would like one, say which shape you prefer and I will add it here or in a follow-up.

## Scope

This fixes the reported table case only. #1837 notes the same check-then-insert shape in `SchemaRepository.createSchema`, `ModelRepository.createRegisteredModel`, `StagingTableRepository` and `UserRepository.createUser` (whose email check has no constraint behind it, though `external_id` does). Those look like the same one-line-constraint plus translation, but each needs its own reproduction and its own answer to the duplicates-already-present question above, so I would rather land the reproduced case and follow up than put five constraints into one PR.

## How we tested

Two new tests in `TableRepositoryTest`, both of which fail without the constraint with all four callers succeeding.

`concurrentCreatesOfTheSameNameKeepTheNameUsable`: four threads released by a barrier create one name, over eight independent names because whether two transactions actually overlap is a scheduling accident. It asserts exactly one create wins, that every loser reports `TABLE_ALREADY_EXISTS` rather than a server error, and then that the name still reads back and still deletes — the two operations the duplicate rows break.

Against `main` it fails the way the issue describes, with all four creates succeeding:

```
[exactly one create wins the race]
expected: 1
 but was: 4
```

`concurrentRenamesOntoTheSameNameKeepTheNameUsable` does the same for renames: four tables are renamed onto one target name, exactly one wins, and the target still reads back.

With the fix, `server/test` is 525 passed. One unrelated failure remains, `CloudCredentialVendorTest.testGenerateGcpTemporaryCredentials`, which fails identically on a clean checkout of `main` at e36e9ca0. An earlier run also hit `BaseModelCRUDTest.testModelCRUD`, which asserts two update timestamps differ and flakes when both land in the same millisecond; it passes when run alone and did not recur.


tiennguyen-onehouse:
Out of draft — all checks green, including the full Spark 4.0.0 lane, so the concurrent-create test holds up on CI hardware and not just locally.

@moomindani thanks for the writeup on #1837; the reproduction steps matched exactly, and the four-threads-on-one-name test lands four rows against `main`. Two questions in the description that are really yours to weigh in on: whether you'd want a startup check or repair path for databases that already carry duplicates (Hibernate silently declines to add the constraint there, measured table in the description), and whether to extend to the other four sites you identified in this PR or a follow-up.

tiennguyen-onehouse:
Pushed 6e0c6476, which extends this to `renameTable`.

I went looking for whether the missing constraint was a deliberate choice before leaning harder on it, and found no sign that it was: `ColumnInfoDAO` and `PropertyDAO` have had unique constraints since the first commit and `UserDAO` gained one in #1691, with no comment or review note anywhere arguing against them here. While tracing what else writes a name into `uc_tables` I found `renameTable` has the same check-then-write shape, and it fails the same way: without the constraint, four concurrent renames onto one target name all land, leaving the same unusable duplicate. The constraint was already stopping that write, so this just catches the violation there too and reports it as a conflict. There is a second test covering it.

On cost, since it is a fair thing to ask of a new index: the only name-based query filters on `(schemaId, name)` and listing filters by schema and orders by name, so the composite index serves both better than the existing `idx_name` on `name` alone, which no query matches by itself. I left `idx_name` alone rather than widen the change, but it looks redundant now if you would like that as a follow-up.

yili-db:
@tiennguyen-onehouse Thanks for this PR. This has been identified as a problem for some time. It really bothers.
I previously filed a PR trying to address it, however I did not have a good story on how to migrate existing deployments that they may have duplicate rows already. See comment https://github.com/unitycatalog/unitycatalog/pull/1380/changes/c00cc26683498b2efa08c5b9bda322b00e4f3611..f4609279fd2bb6415f2e66382dc308f2a3994cab#r2913725525
Currently what we lack is a good mechanism for database table schema migration like this, and an upgrade test that exercises it. We have planned for that but have not started the work.

tiennguyen-onehouse:
@yili-db thank you, that context is exactly what I was missing — I had not found #1380, and the review thread there is the useful part.

I went and measured the specific thing that blocked it. In [this comment](https://github.com/unitycatalog/unitycatalog/pull/1380#discussion_r2913725525) @tdas wrote that if a production database already has duplicate rows, `ALTER TABLE ADD CONSTRAINT` will fail and the server will not start. **That turns out not to happen.** I built a database with a build of `main`, seeded duplicates, then opened it with this branch, on PostgreSQL 16 and on H2:

| Existing database | Startup | Constraint after upgrade | Duplicate rows |
|---|---|---|---|
| Clean | starts normally | `uc_tables_schema_id_name_key` created | 0 |
| Already has duplicates | **starts normally** | not created | left as they were |

Hibernate's `hbm2ddl.auto=update` attempts the constraint, the attempt fails, and it swallows the failure. Nothing is logged at any level. So the upgrade is not dangerous in the way the thread assumed, on either database.

The real problem is the opposite one, and it is worse to reason about than a crash: **the failure is silent.** An operator upgrades, reasonably believes the race is now impossible, and on a database that was already bitten it is not. A crash at least tells you. That is the gap worth closing, and it is much smaller than a general migration framework.

Three things I can contribute here, whichever you prefer:

1. **Detection at startup.** After the schema update, check whether the constraints the mapping declares actually exist, and log a clear warning naming the table and the remediation when one is missing. Optionally a strict mode that refuses to start, off by default. This is maybe thirty lines and needs no migration tooling.
2. **The upgrade test you mentioned you lack.** The measurement above is one: stand up the database with a prior build, run the new build against it, assert on the resulting schema and data. I already have it working for PostgreSQL via Testcontainers and for H2, across two checkouts. I would be glad to turn it into a reusable harness so any future DDL change can be exercised the same way, rather than leaving it as something I ran by hand.
3. **A deduplication utility**, if you want operators to have a supported path rather than hand-written SQL.

On the rest of the #1380 thread, this PR happens to answer a few of those points by being narrower, which may make it a useful first instance of the pattern:

- tdas asked whether `REPEATABLE_READ` plus unique constraints had been considered instead of blanket `SERIALIZABLE`, and whether the cost had been measured. This PR is that lighter option — no isolation change at all — and I measured it on PostgreSQL 16 with 10,500 rows. Creates, lookups and listings all got slightly faster, because every create already does a name lookup first and the new composite index on `(schema_id, name)` serves it, and the only name query and the listing sort both match it, while the existing `idx_name` on `name` alone matches neither.
- He also asked that retries not fire on any `ConstraintViolationException`. This catches only the table-name constraint, by name, so a request with duplicate columns still reports its own violation.
- And he flagged that the concurrency test could pass vacuously if both writers succeed. Mine asserts the winner count explicitly, which is what fails first against `main` — four creates and four renames all land.

Happy to take any of the three on, in this PR or separately. I would rather not guess which fits your plans.

yili-db:
@tiennguyen-onehouse That makes perfect sense!

Now that `(schema_id, name)` is constrained, `idx_name` on `name` alone matches no query: both `findBySchemaIdAndName` and the listing (filter by `schema_id`, order by `name`) are served by the constraint's backing index. Can you drop its annotation in this PR rather than leave it redundant? It's one line and directly caused by this change.

One caveat worth a note in the description: `update` mode won't drop the physical index on existing DBs, so this only stops new databases from creating it. But that's OK.

moomindani:
Thanks for validating the `hbm2ddl` behaviour that blocked #1380 -- that measurement is the useful part
of the thread.

**Detection at startup: yes, and I would keep it out of this PR.** To be precise about which of your
three I mean: (1), a check that warns when a declared constraint is missing. Not a repair path --
deciding which duplicate row survives, and what becomes of its children and its grants, is a design
question, and it is one the migration work @yili-db described will have to answer anyway. Detection
needs none of that, and it matters because a deployment that already carries duplicates is exactly the
one whose operator cannot clean them up through the API: `getTable` and `deleteTable` both 500, which is
how I found #1837. (2) is the harness they said is planned but not started, so offering it is right and
designing it here is not. (3) I would leave until an operator asks.

**The other sites: follow-ups, and you are right that each needs its own reproduction -- they do not
fail alike.** Measured on `main` (H2, four threads released by a barrier, eight distinct names each):

| site | duplicates persisted | what the name does afterwards |
| --- | --- | --- |
| `uc_tables` | yes, four rows on every name | `GET` and `DELETE` both 500 `Query did not return a unique result` |
| `uc_schemas` | yes, on every name | `GET` answers 200; `DELETE` answers 200 and removes one row, so the name survives with a different `schema_id` -- four deletes to clear it in this run |
| `uc_registered_models` | yes, on six of eight names | `GET` answers 200 |
| `uc_users` (SCIM, one email) | yes, on seven of eight emails | no error at all |

`uc_tables` is the only one that fails loudly. The schema, model and user-email lookups each cap at one
row -- `RepositoryUtils.getSchemaDaoOpt`, `ModelRepository.getRegisteredModelDao` and
`UserRepository.getUserByEmail` all `setMaxResults(1)` -- so `uniqueResult` never sees two rows and
nothing raises; the name resolves to an unspecified row, since none of the three queries orders its
results.

I would put the email one first: it has no constraint behind it today (`UserDAO` constrains
`external_id` only) and it is the lookup the authentication and principal-resolution paths go through,
so a silent pick there is both harder to notice and worse placed than the 500 this PR fixes. After it,
`uc_schemas` (no constraints at all today), then `uc_registered_models`.

One correction to my own issue: I listed `StagingTableRepository` there, and that one I had wrong.
`validateIfAlreadyExists` deliberately allows two staging tables of the same name ("It's OK if a
staging table with the same name already exist") and enforces uniqueness on the staging location
instead, so four concurrent creates succeeding is the design working rather than a race. Nothing to
constrain there unless the intended uniqueness changes.

Happy to help with those follow-ups as separate PRs if that is useful to you.

